PR #586, opened 2018-03-13
Fix ignored tests

duncanp-sonar:
The approach being used by the tests to wait for the initial fetch to complete was unreliable leading to intermittent test failures.

### DIFF
--- a/src/Integration/Rules/QualityProfileProviderCachingDecorator.cs
+++ b/src/Integration/Rules/QualityProfileProviderCachingDecorator.cs
@@ -110,10 +110,6 @@ namespace SonarLint.VisualStudio.Integration.Rules
             }
 
             SynchronizeQualityProfiles();
-
-
-            Thread.Sleep(1000);
-
             OnInitialFetchComplete();
         }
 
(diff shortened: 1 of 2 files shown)


src/Integration/Rules/QualityProfileProviderCachingDecorator.cs
```diff
@@ -142,15 +150,25 @@ private void SynchronizeQualityProfiles()
         private bool isDisposed;
         public void Dispose()
         {
-            if (this.isDisposed)
+            this.Dispose(true);
+            GC.SuppressFinalize(this);
```
duncanp-sonar: I didn't think it was necessary to call _SuppressFinalize_ if the class doesn't have a finalizer, but our rule [S3881](https://rules.sonarsource.com/csharp/RSPEC-3881) complains if it isn't there.

duncanp-sonar:
I added a virtual _OnInitialFetchComplete_ method to the product class so the tests can reliably wait for the fetch to complete.

src/Integration/Rules/QualityProfileProviderCachingDecorator.cs
michalb-sonar: IIRC it is because the class is not sealed and child classes can have it.

Evangelink: @michalb-sonar is right. We had one user who made us the remark that if the class is not sealed and doesn't have the `GC.SuppressFinalize` then sub-classes with a dtor wouldn't have the right behavior.

PR #592, opened 2018-03-14
Remove the special handling for ruleset as it wasn't working properly

Evangelink:


src/Integration/MefServices/DeprecatedSonarRuleSetManager.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            if (this.projectSystemHelper.GetSolutionProjects().Any(HasAnySonarRule))
+            {
+                this.logger.WriteLine(Strings.ProjectWithSonarRules);
+            }
+        }
+
+        private bool HasAnySonarRule(EnvDTE.Project project)
+        {
+            // We only look for the "global" CodeAnalysisRuleSet property (i.e. not under any condition)
```
duncanp-sonar: It looks like the existing _ISolutionRuleSetsInformationProvider.GetProjectRuleSetsDeclarations_ will return all of the ruleset declarations for a given project i.e. taking the configuration into account.

src/Integration/MefServices/DeprecatedSonarRuleSetManager.cs
```diff
@@ -0,0 +1,171 @@
...
+                WarnIfAnyProjectHasSonarRuleSet();
+            }
+        }
+
+        private void OnActiveSolutionChanged(object sender, ActiveSolutionChangedEventArgs e)
+        {
+            if (e.IsSolutionOpen)
+            {
+                WarnIfAnyProjectHasSonarRuleSet();
+            }
```
duncanp-sonar: The event handlers will be called on the UI thread so it would be safer to have a _try-catch-squash-if-not-critical block_.

src/Integration.UnitTests/MefServices/DeprecatedSonarRuleSetManagerTests.cs
```diff
@@ -0,0 +1,353 @@
...
+            this.activeSolutionBoundTracker.SimulateSolutionBindingChanged(new ActiveSolutionBindingEventArgs(
+                new NewConnectedMode.BindingConfiguration(new Persistence.BoundSonarQubeProject(),
+                NewConnectedMode.SonarLintMode.LegacyConnected)));
+
+            // Assert
+            loggerMock.Verify(x => x.WriteLine(DeprecatedSonarRuleSetManager.DeprecationMessage), Times.Never);
+        }
+
+        [TestMethod]
+        public void OnSolutionBindingChanged_WhenStandaloneAndProjectHasSonarRules_DoWarn()
```
duncanp-sonar: Are there any tests where the ruleset exists but does not have Sonar rules?

src/Integration.UnitTests/MefServices/DeprecatedSonarRuleSetManagerTests.cs
```diff
@@ -0,0 +1,353 @@
...
+            var rulesetName = $"{Guid.NewGuid()}.ruleset";
+            var ruleset = TestRuleSetHelper.CreateTestRuleSetWithRuleIds(new[] { "S110" }, "SonarAnalyzer.Foo");
+            ruleset.WriteToFile(Path.Combine(Path.GetDirectoryName(projectPath), rulesetName));
+
+            var project = new ProjectMock(projectPath);
+            projectSystemHelper.Projects = new[] { project };
+            project.SetBuildProperty(Constants.CodeAnalysisRuleSetPropertyKey, rulesetName);
+
+            this.ruleSetProvider.RegisterProjectInfo(project, new RuleSetDeclaration(project, new PropertyMock("bar", null),
+                rulesetName, "", ""));
```
duncanp-sonar: This _create-and-register-ruleset_ code could be moved to a helper function (perhaps with a parameter to say whether the ruleset should contain a Sonar rule or not).

PR #593, opened 2018-03-26
Fixed failing unit test

duncanp-sonar:
Test was broken by the change to daylight saving time

### DIFF
--- a/src/Integration.UnitTests/Telemetry/TelemetryDataRepositoryTests.cs
+++ b/src/Integration.UnitTests/Telemetry/TelemetryDataRepositoryTests.cs
@@ -151,16 +151,24 @@ namespace SonarLint.VisualStudio.Integration.Tests
  <LastUploadDate>1999-12-31T23:59:59.9999999</LastUploadDate>
 </TelemetryData>");
 
+            // Calculate the expected result.
+            // Previously, this test started failing when daylight saving was applied on the test agent
+            // machine. Creating the local date first then converting it to a DateTimeOffset gives the
+            // expected result regardless of the local time zone or whether the test agent machine is
+            // automatically adjusting dor daylight saving time or not.
+            var expectedDate = new DateTime(1999, 12, 31, 23, 59, 59, 999, DateTimeKind.Local).AddTicks(9999);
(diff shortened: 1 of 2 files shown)


src/Integration.UnitTests/Telemetry/TelemetryDataRepositoryTests.cs
```diff
@@ -151,16 +151,24 @@ public void Can_Read_Old_TelemetryXml()
  <LastUploadDate>1999-12-31T23:59:59.9999999</LastUploadDate>
 </TelemetryData>");
 
+            // Calculate the expected result.
+            // Previously, this test started failing when daylight saving was applied on the test agent
+            // machine. Creating the local date first then converting it to a DateTimeOffset gives the
+            // expected result regardless of the local time zone or whether the test agent machine is
+            // automatically adjusting dor daylight saving time or not.
+            var expectedDate = new DateTime(1999, 12, 31, 23, 59, 59, 999, DateTimeKind.Local).AddTicks(9999);
+            var expectedDateTimeOffset = new DateTimeOffset(expectedDate);
```
duncanp-sonar: Actually, I've just looked at the other tests in this class and they are all creating the _DateTimeOffset_ by first creating a _DateTime_.

src/Integration/Telemetry/TelemetryData.cs
```diff
@@ -43,23 +43,29 @@ public sealed class TelemetryData
...
-            set { LastUploadDate = DateTimeOffset.ParseExact(value, "o", null); }
+            set { LastUploadDate = ParseSavedString(value); }
+        }
+
+        private static DateTimeOffset ParseSavedString(string data)
+        {
+            // ParseExact will throw an exception when value is invalid date, but
+            // XmlSerializer will swallow it and return default(TelemetryData)
+            return DateTimeOffset.ParseExact(data, "o", null,
+                System.Globalization.DateTimeStyles.AssumeLocal | System.Globalization.DateTimeStyles.AdjustToUniversal);
```
duncanp-sonar: I've added the _styles_ options to make it explicit how dates without timezones will be processed (although _AssumeLocal_ is the default so the actual behaviour won't change).

PR #603, opened 2018-04-05
Revert changes to save new config to .suo file

### LINKED ISSUE #596: [Technical] Modify new binding to serialization to create an sqconfig file
* drop the .suo serialization
* decide on some way to distinguish between old and new binding modes (e.g. a version number in the file)

duncanp-sonar:
* new connected mode config is now saved in a solution-level folder called ".sonarlint" in a file called "{solution name}.sqconfig"

Reverts the changes in cb9adc4af
Fixed #596


src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -305,9 +305,14 @@ private void RegisterLocalServices()
...
+                var solution = this.GetService<SVsSolution, IVsSolution>();
+                var store = new SecretStore(SolutionBindingSerializer.StoreNamespace);
                 var legacySerializer = new SolutionBindingSerializer(this);
-                return new ConfigurationProvider(legacySerializer, InMemoryConfigurationProvider.Instance);
-
+                // The SCC wrapper maintains a queue of file writes. For the new provider we want to the file write to be
+                // executed immediately, so we want our own SCC wrapper rather than sharing the one used by the legacy mode.
+                var sccFileSystem = new SourceControlledFileSystem(this);
+                var newConfigSerializer = new ConfigurationSerializer(solution, sccFileSystem, store, Logger);
+                return new ConfigurationProvider(legacySerializer, newConfigSerializer);
```
duncanp-sonar: This code will change again shortly once the new binding method is implemented (we'll be writing multiple files during binding again, so it should be ok to share the SCC wrapper again).

src/Integration/NewConnectedMode/ConfigurationProvider.cs
```diff
@@ -65,19 +68,18 @@ public bool WriteConfiguration(BindingConfiguration configuration)
...
-                return success;
+                case SonarLintMode.LegacyConnected:
+                    return legacySerializer.WriteSolutionBinding(configuration.Project) != null;
+                case SonarLintMode.Connected:
+                    return newConnectedModeSerializer.WriteSolutionBinding(configuration.Project) != null;
+                case SonarLintMode.Standalone:
+                    throw new InvalidOperationException(Strings.Bind_CannotSaveStandaloneConfiguration);
+                default:
+                    Debug.Fail("Unrecognised write mode");
+                    return false;
```
duncanp-sonar: Unfortunately, the removal of the _InMemoryConfigurationProvider_ means that every time a component asks for the current configuration, we'll attempt to read from disk (and in the case of standalone mode, we'll tried to read from two different locations, but find nothing).

I'll fix this if I have time, probably by making this component store the config in-memory, and adding a "refresh"  option that checks for the files on disk. The "refresh" would be called whenever a solution is opened or closed. That should be enough to keep the in-memory state in sync.

PR #605, opened 2018-04-09
Removed redundant interface method

duncanp-sonar:
- ISolutionBindingInformationProvider.GetBoundProjects was not used. Removed declaration and implementation and fixed up tests.
- Miscellaneous: added documentation to binding workflow

### DIFF
--- a/src/Integration.UnitTests/LocalServices/SolutionBindingInformationProviderTests.cs
+++ b/src/Integration.UnitTests/LocalServices/SolutionBindingInformationProviderTests.cs
@@ -26,6 +26,7 @@ using FluentAssertions;
 using Microsoft.VisualStudio.CodeAnalysis.RuleSets;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarLint.VisualStudio.Integration.NewConnectedMode;
+using SonarLint.VisualStudio.Integration.Persistence;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests
 {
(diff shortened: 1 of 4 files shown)


src/Integration.UnitTests/LocalServices/SolutionBindingInformationProviderTests.cs
```diff
@@ -65,143 +66,117 @@ public void SolutionBindingInformationProvider_ArgCheck()
         }
 
         [TestMethod]
-        public void SolutionBindingInformationProvider_GetBoundProjects_SolutionNotBound_StandaloneMode()
+        public void SolutionBindingInformationProvider_GetUnboundProjects_SolutionBound_EmptyFilteredProjects()
```
duncanp-sonar: I deleted the first two tests since there are already _unbound_ equivalents for them (father down in the file).

However, the _GetBoundProjects_ tests were more comprehensive than the tests for _GetUnBound_methods, so I've kept the rest of those tests but changed them to call _GetUnboundProjects_, and changed the assertions to look for the inverse set of projects.

src/Integration/Binding/BindingWorkflow.cs
```diff
@@ -138,24 +138,50 @@ private ProgressStepDefinition[] CreateWorkflowSteps(IProgressController control
...
             return new ProgressStepDefinition[]
             {
+                // Some of the steps are broken into multiple sub-steps, either
+                // because the work needs to be done on a different thread or
+                // to report progress separate.
+
+                //*****************************************************************
+                // Initialization
+                //*****************************************************************
+                // Show an initial message and check the solution isn't dirty
```
duncanp-sonar: Miscellaneous: added documentation to explain the binding steps at higher level.

src/Integration/LocalServices/SolutionBindingInformationProvider.cs
```diff
@@ -84,32 +79,18 @@ private IEnumerable<Project> GetUnboundProjects(BoundSonarQubeProject binding)
...
+            // method can be called from the UI thread, hence this short-lived cache
             Dictionary<string, RuleSet> cache = new Dictionary<string, RuleSet>(StringComparer.OrdinalIgnoreCase);
 
             // Note: we will still may end up analyzing the same project rule set
             // but that should in marginal since it will be already loaded into memory
 
             // Reuse the binding information passed in to avoid reading it more than once
             return projectSystem.GetFilteredSolutionProjects()
-                .Where(p => this.IsFullyBoundProject(cache, binding, p));
+                .Where(p => !this.IsFullyBoundProject(cache, binding, p));
```
duncanp-sonar: Previously, the _GetUnboundProjects_ method got all of the filtered projects then called _GetBoundProjects_ to work out which ones to exclude.  Now, it just gets the filtered projects and works out which ones are not fully bound directly.

duncanp-sonar:
Minor refactoring and cleanup. The commit doesn't change the binding at all, but it does get rid of some redundant code.

PR #606, opened 2018-04-10
Update all instances of SingleOrDefault to avoid crash in production

### LINKED ISSUE #594: InvalidOperationException thrown
### Description

We're seeing VS crashes being reported by Watson coming from SonarLint. An InvalidOperationException is being thrown by a call to `System.Linq.Enumerable.SingleOrDefault` somewhere in the **SonarLint.VisualStudio.Integration** dll with the error message *Sequence contains more than one matching element*.

Unfortunately, we cannot tell exactly where in the source code the issue occurs, only that it is in the SonarLint.VisualStudio.Integration module.

### Repro steps

N/A

### Known workarounds

A possible fix is to change the call from `SingleOrDefault` to `FirstOrDefault` which doesn't throw when there are duplicate values in the collection.

### Related information

* SonarLint for VS Version **unknown**
* Visual Studio Version **2015 and 2017**

If you have any questions, feel free to send an email to madsk@microsoft.com

Thanks,
Mads Kristensen
Visual Studio Extensibility

> PS. If you care to ship your .pdb files inside the VSIX container then we have a better chance at debugging the exact location of these types of issues. It will add to the overall weight of your .vsix file, so I understand if you are hesitant to do so.

Evangelink:
Fix #594


src/Integration/Helpers/RuleSetHelper.cs
```diff
@@ -95,9 +95,24 @@ public static RuleSetInclude FindInclude(RuleSet source, RuleSet target)
...
+                return null;
+            }
+            else if (matchingRuleSetIncludes.Count == 1)
+            {
+                return matchingRuleSetIncludes[0];
+            }
+            else
+            {
+                Debug.Fail("Not expecting to find multiple RuleSetInclude matching the filter");
+                return null;
```
duncanp-sonar: It looks like this methods is only used in _SolutionBindingInformationProvider.IsFullyBoundProject_, and this change will cause _IsFullyBound_ to returning _false_, which will trigger the Error List gold bar.
I don't think is the behaviour we want; I don't think it matters in this case if the ruleset is included multiple times, although it would be worth updating the method comment to make it clear that this is the behaviour.
What do you think?

Evangelink: So I keep the `Debug.Fail`, I add a comment and return the first element, does this sound OK?

duncanp-sonar: Looks good to me.

PR #610, opened 2018-04-10
Refactored binding workflow to extract NuGet code

duncanp-sonar:
I suggest starting with the _BindingWorkflow.cs_ file. I extracted code from that into a new _NuGetBindingOperation_ file, and then moved the tests into a new file too.
I'll call out all of the changed code.

### DIFF
--- a/src/Integration.UnitTests/Binding/BindingWorkflowTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingWorkflowTests.cs
@@ -314,10 +314,10 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             var bindingArgs = new BindCommandArgs("projectKey", "projectName", new ConnectionInformation(new Uri("http://connected")));
 
             Mock<INuGetBindingOperation> nugetMock = new Mock<INuGetBindingOperation>();
-            nugetMock.Setup(x => x.InstallPackages(It.IsAny<IProgressController>(),
-                It.IsAny<CancellationToken>(),
+            nugetMock.Setup(x => x.InstallPackages(It.IsAny<ISet<Project>>(),
+                It.IsAny<IProgressController>(),
                 It.IsAny<IProgressStepExecutionEvents>(),
-                It.IsAny<ISet<Project>>())).Returns(true);
+                It.IsAny<CancellationToken>())).Returns(true);
 
             var testSubject = new BindingWorkflow(this.host, bindingArgs, nugetMock.Object);
 
@@ -331,7 +331,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             testSubject.BindingOperationSucceeded = false;
 
             // Act
-            testSubject.InstallPackages(new ConfigurableProgressController(), cts.Token, progressEvents);
+            testSubject.InstallPackages(new ConfigurableProgressController(), progressEvents, cts.Token);
 
             // Assert
             testSubject.BindingOperationSucceeded.Should().BeTrue();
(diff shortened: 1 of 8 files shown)


src/Integration.UnitTests/Binding/BindingWorkflowTests.cs
```diff
@@ -298,184 +308,63 @@ public void BindingWorkflow_ArgChecks()
...
-        }
-
-        [TestMethod]
-        public void BindingWorkflow_InstallPackages_WhenAllProjectsAreVbNet_Succeed()
-        {
-            BindingWorkflow_InstallPackages_Succeed(ProjectSystemHelper.VbProjectKind, Language.VBNET);
-        }
-
-        private void BindingWorkflow_InstallPackages_Succeed(string projectKind, Language language)
+        public void BindingWorkflow_InstallPackages_Succeeds_SuccessPropertyIsTrue()
```
duncanp-sonar: This test and the next test are new. All of the other changes tests below have been moved to the new _NuGetBindingOperationTests.cs_ file.

src/Integration.UnitTests/Binding/BindingWorkflowTests.cs
```diff
@@ -94,7 +93,17 @@ public void BindingWorkflow_ArgChecks()
...
-            BindingWorkflow testSubject = this.CreateTestSubject("key", ProjectName);
+
+            // Record all of the calls to NuGetBindingOperation.ProcessExport
+            var actualProfiles = new List<Tuple<Language, RoslynExportProfileResponse>>();
+            Mock<INuGetBindingOperation> nuGetOpMock = new Mock<INuGetBindingOperation>();
+            nuGetOpMock.Setup(x => x.ProcessExport(It.IsAny<Language>(), It.IsAny<RoslynExportProfileResponse>()))
+                .Callback<Language, RoslynExportProfileResponse>((l, r) => actualProfiles.Add(new Tuple<Language, RoslynExportProfileResponse>(l, r)))
+                .Returns(true);
+
+            BindingWorkflow testSubject = this.CreateTestSubject("key", ProjectName, nuGetOpMock.Object);
```
duncanp-sonar: Fixed up this _BindingWorkflow_ test to use a mocked _INuGetBindingOperation_ rather than calling the real code. The test is just checking that _ProcessExport_ is called with the expected data.

src/Integration.UnitTests/Binding/NuGetBindingOperationTests.cs
```diff
@@ -0,0 +1,270 @@
...
+        public void TestInitialize()
+        {
+            serviceProvider = new ConfigurableServiceProvider();
+            logger = new TestLogger();
+        }
+
+        #region Tests
+
+        [TestMethod]
+        public void InstallPackages_WhenAllProjectsAreCSharp_Succeed()
```
duncanp-sonar: All of these tests were moved from _BindingWorkflowTests.cs_. I haven't added any new ones yet.

src/Integration.UnitTests/Binding/NuGetHelperTests.cs
```diff
@@ -37,33 +37,34 @@ public class NuGetHelperTests
...
         [TestMethod]
         public void NuGetHelper_HandleFailures()
         {
             // Arrange
-            ConfigurableServiceProvider sp = CreateServiceProvider();
-            var outputWindow = new ConfigurableVsOutputWindow();
-            var outputPane = outputWindow.GetOrCreateSonarLintPane();
-            sp.RegisterService(typeof(SVsOutputWindow), outputWindow);
+            var sp = CreateServiceProvider();
+            var testLogger = new TestLogger();
```
duncanp-sonar: Minor refactoring of the _NuGetHelper_ so that it accepts an _ILogger_ in the constructor.
Previously, it was using a static helper method to write to the VS Output pane.

src/TestInfrastructure/Framework/TestLogger.cs
```diff
@@ -0,0 +1,57 @@
...
+    public class TestLogger : ILogger
+    {
+        private IList<string> outputStrings = new List<string>();
+
+        public void AssertOutputStrings(int expectedOutputMessages)
+        {
+            this.outputStrings.Should().HaveCount(expectedOutputMessages);
+        }
+
+        public void AssertOutputStrings(params string[] orderedOutputMessages)
```
duncanp-sonar: There wasn't previously a simple implementation of _ILogger_ - the existing tests wrote used the _ConfigurableOutputWindowPane_ class (which is where I copied the assertions from).

src/Integration/Binding/NuGetBindingOperation.cs
```diff
@@ -0,0 +1,148 @@
...
+
+                // TODO: SVS-33 (https://jira.sonarsource.com/browse/SVS-33) Trigger a Team Explorer warning notification to investigate the partial binding in the output window.
+                overallSuccess &= isNugetInstallSuccessful;
+
+                progressNotifier.NotifyIncrementedProgress(string.Empty);
+            }
+            return overallSuccess;
+        }
+
+        public bool ProcessExport(Language language, RoslynExportProfileResponse exportProfileResponse)
```
duncanp-sonar: The code in this method was extracted from inside _BindingWorkflow.DownloadQualityProfileAsync_.

PR #611, opened 2018-04-11
Added no-op NuGet binder for the new connected mode

duncanp-sonar:


src/Integration/Binding/BindingController.cs
```diff
@@ -145,7 +146,7 @@ void IBindingWorkflowExecutor.BindProject(BindCommandArgs bindingArgs)
                     Strings.Bind_FirstTimeBinding :
                     Strings.Bind_UpdatingNewStyleBinding);
 
-            return new NewBindingWorkflow(host, bindingArgs, configProvider);
+            return new BindingWorkflow(host, bindingArgs, new NoOpNuGetBindingOperation(host.Logger));
```
duncanp-sonar: Instead of using a completely separate binding workflow class, we now use the same binding class but with different behaviours plugged in (I'll delete the redundant new binding workflow code separately).
The next step is to plug in a "solution binding operation" that saves the data to a different location and doesn't add the files to the solution folder.

src/Integration/Binding/BindingWorkflow.cs
```diff
@@ -85,6 +85,8 @@ public BindingWorkflow(IHost host, BindCommandArgs bindingArgs)
             this.nugetBindingOperation = nugetBindingOperation;
         }
 
+        internal /*for testing*/ INuGetBindingOperation NuGetBindingOperation {  get { return this.nugetBindingOperation; } }
```
duncanp-sonar: This isn't ideal: it's only there so the unit test can check that the workflow picked the right NuGet behaviour. I might create a factory class later for this and the solution binding operation, but it isn't critical at the moment.

PR #612, opened 2018-04-12
Updated SolutionRulesetsInformationProvider to handle new and legacy …

duncanp-sonar:
…connected modes

The new-new connected mode writes the shared rulesets and config file to a different solution-level folder.
This change updates the _ISolutionRuleSetsInformationProvider_ so it can handle the different modes.

### DIFF
--- a/src/Integration/Binding/BindingWorkflow.cs
+++ b/src/Integration/Binding/BindingWorkflow.cs
@@ -81,7 +81,9 @@ namespace SonarLint.VisualStudio.Integration.Binding
             this.solutionBindingOperation = new SolutionBindingOperation(
                     this.host,
                     this.bindingArgs.Connection,
-                    this.bindingArgs.ProjectKey);
+                    this.bindingArgs.ProjectKey,
+                    //TODO: duncanp pick the correct mode
+                    NewConnectedMode.SonarLintMode.LegacyConnected);
             this.nugetBindingOperation = nugetBindingOperation;
         }
 
--- a/src/Integration/Binding/SolutionBindingOperation.cs
+++ b/src/Integration/Binding/SolutionBindingOperation.cs
@@ -48,8 +48,9 @@ namespace SonarLint.VisualStudio.Integration.Binding
         private Dictionary<Language, SonarQubeQualityProfile> qualityProfileMap;
         private readonly ConnectionInformation connection;
         private readonly string ProjectKey;
+        private readonly SonarLintMode bindingMode;
 
-        public SolutionBindingOperation(IServiceProvider serviceProvider, ConnectionInformation connection, string ProjectKey)
+        public SolutionBindingOperation(IServiceProvider serviceProvider, ConnectionInformation connection, string ProjectKey, SonarLintMode bindingMode)
         {
             if (serviceProvider == null)
             {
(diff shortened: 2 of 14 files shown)


src/Integration/Binding/BindingWorkflow.cs
```diff
@@ -81,7 +81,9 @@ internal class BindingWorkflow : IBindingWorkflow
             this.solutionBindingOperation = new SolutionBindingOperation(
                     this.host,
                     this.bindingArgs.Connection,
-                    this.bindingArgs.ProjectKey);
+                    this.bindingArgs.ProjectKey,
+                    //TODO: duncanp pick the correct mode
```
duncanp-sonar: This will done in the next (and hopefully last) PR.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -255,7 +263,7 @@ private void PendBindingInformation(ConnectionInformation connInfo)
                 connInfo.Organization);
             bound.Profiles = map;
 
-            // TODO: choose solution binding mode
+            // TODO: duncanp choose solution binding mode
```
duncanp-sonar: The binding mode is being supplied in the constructor, so the next line should be using it. It'll be part of the next PR.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -48,8 +48,9 @@ internal class SolutionBindingOperation : ISolutionRuleStore
         private Dictionary<Language, SonarQubeQualityProfile> qualityProfileMap;
         private readonly ConnectionInformation connection;
         private readonly string ProjectKey;
+        private readonly SonarLintMode bindingMode;
 
-        public SolutionBindingOperation(IServiceProvider serviceProvider, ConnectionInformation connection, string ProjectKey)
+        public SolutionBindingOperation(IServiceProvider serviceProvider, ConnectionInformation connection, string ProjectKey, SonarLintMode bindingMode)
```
duncanp-sonar: I took the simple approach of making this class aware of the binding mode and put in conditional checks in a couple of places - I couldn't see an easy way to factor out the mode-specific code in the way I did for the _BindingWorkflow_.

src/Integration/Persistence/SolutionBindingSerializer.cs
```diff
@@ -76,7 +79,7 @@ private void AddSolutionItemFile(string configFile)
             var projectSystemHelper = this.serviceProvider.GetService<IProjectSystemHelper>();
             projectSystemHelper.AssertLocalServiceIsNotNull();
 
-            Project solutionItemsProject = projectSystemHelper.GetSolutionFolderProject(Constants.SonarQubeManagedFolderName, true);
+            Project solutionItemsProject = projectSystemHelper.GetSolutionFolderProject(Constants.LegacySonarQubeManagedFolderName, true);
```
duncanp-sonar: There are separate serializer classes for legacy and connected modes, so the binding mode can be hard-coded.

src/Integration/LocalServices/SolutionBindingInformationProvider.cs
```diff
@@ -127,7 +127,9 @@ private RuleSet FindSonarQubeSolutionRuleSet(Dictionary<string, RuleSet> cache,
 
             string expectedSolutionRuleSet = ruleSetInfoProvider.CalculateSolutionSonarQubeRuleSetFilePath(
                          binding.ProjectKey,
-                         Language.ForProject(project));
+                         Language.ForProject(project),
+                         // TODO: duncanp decide whether to support new connected mode
```
duncanp-sonar: I've added a couple of specific reminders for myself of additional features to look at before shipping (I had quick look at a couple of the ruleset/conflict manager classes and couldn't see anything that was specific to NuGet so they might work ok with the new connected mode).

PR #613, opened 2018-04-12
Updated binding save to handle new binding mode

duncanp-sonar:


src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -308,9 +308,7 @@ private void RegisterLocalServices()
                 var solution = this.GetService<SVsSolution, IVsSolution>();
                 var store = new SecretStore(SolutionBindingSerializer.StoreNamespace);
                 var legacySerializer = new SolutionBindingSerializer(this);
-                // The SCC wrapper maintains a queue of file writes. For the new provider we want to the file write to be
-                // executed immediately, so we want our own SCC wrapper rather than sharing the one used by the legacy mode.
```
duncanp-sonar: The new connected mode now writes multiple files to disk (shared ruleset, config file, and the project-level ruleset files) so there is no reason for it not to use the same ISourceControlledFileSystem to queue the changes.

src/Integration/Persistence/FileBindingSerializer.cs
```diff
@@ -29,26 +29,22 @@
 
 namespace SonarLint.VisualStudio.Integration.Persistence
 {
+    /// <summary>
+    /// Writes the binding configuration file to the source controlled file system
+    /// </summary>
+    /// <remarks>
+    /// The file will be enqued but not actually written.
```
duncanp-sonar: Typo - "enqued"

src/Integration/Persistence/FileBindingSerializer.cs
```diff
@@ -29,26 +29,22 @@
...
+    /// It is the responsibility of the caller to flush the queue.
+    /// This is to allow multiple other files to be written using the 
+    /// same instance of the SCC wrapper (e.g. ruleset files).
+    /// </remarks>
```
duncanp-sonar: An earlier version of the new connected mode needed the ability to commit a single file change directly to disk using the ISourceControlledFileSystem (I think it was when the only file being written was the config file).
That code is no longer used so I've removed it.

duncanp-sonar:
After this commit, it's possible to bind a project using the new workflow (i.e. same as the old workflow, but no NuGet files and the solution level folder and files have different names).

Still to do:
* delete the redundant NewBindingWorkflow class
* change the config file extension to _.slconfig_
* the ConfigurationProvider is throwing an exception when a solution is closed (the exception is suppressed, but it shouldn't be thrown in the first place)
* decide whether to action the remaining "TODOs"

PR #614, opened 2018-04-12
Removed unused code

duncanp-sonar:
Removed _DeprecatedSonarRuleSetManager_ and tests


src/Integration/MefServices/DeprecatedSonarRuleSetManager.cs
```diff
@@ -1,186 +0,0 @@
...
-
-            // 1. Collect all paths (current ruleset + includes)
-            var ruleSetIncludeFullPaths = projectRuleSet.RuleSetIncludes
-                .Select(include => GetFullPath(include.FilePath, projectDirectoryFullPath))
-                .ToList();
-
-            // 2. Look if any of the effective rules is from SonarAnalyzer and initialize dictionary with this result.
-            return projectRuleSet
-                .GetEffectiveRules(ruleSetIncludeFullPaths, new RuleInfoProvider[0])
-                .Any(rule => rule.AnalyzerId.StartsWith("SonarAnalyzer.", StringComparison.OrdinalIgnoreCase));
```
duncanp-sonar: FYI this line was throwing if the ruleset contained an included ruleset that could not be located...

src/Integration.Vsix/SonarLintIntegrationPackage.cs
```diff
@@ -80,8 +79,6 @@ protected override void Initialize()
 
             IServiceProvider serviceProvider = this;
 
-            this.deprecatedSonarRuleSetManager = this.GetMefService<IDeprecatedSonarRuleSetManager>();
```
duncanp-sonar: This line caused the _DeprecatedSonarRulesetManager_ to be created. If the constructor threw then it would crash VS.
The _SingleOrDefault_ call in _GetMefService_ would be on the call stack at the point of the crash. It could be that the unexplained crashes that have been reported are not to do with _SingleOrDefault_ finding multiple items, but errors in the constructors of objects being created via MEF.

I'll create a ticket for hardening the constructors and the package initialisation.

Evangelink: Branch needs to be rebased on master to avoid `SingleOrDefault` issues.

duncanp-sonar: @Evangelink  I don't think using _FirstOrDefault_ would fix this issue: the problem isn't that we are getting multiple objects returned by MEF, it's that object being created by MEF is throwing an exception which isn't caught.

PR #616, opened 2018-04-12
Changed new config file extension

duncanp-sonar:
... to .slconfig.

Fixed up legacy telemetry to detect new connected mode solutions.

### DIFF
--- a/src/Integration.Vsix/BoundSolutionAnalyzer.cs
+++ b/src/Integration.Vsix/BoundSolutionAnalyzer.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarLint for Visual Studio
  * Copyright (C) 2016-2018 SonarSource SA
  * mailto:info AT sonarsource DOT com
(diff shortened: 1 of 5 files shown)


src/Integration.Vsix/BoundSolutionAnalyzer.cs
```diff
@@ -94,6 +89,17 @@ private void OnSolutionBuilding()
             }
         }
 
+        private static string[] GetFiles(string rootDirectory, string folder, string pattern)
+        {
+            string[] matches = null;
```
Evangelink: [minor][personal taste] I'd prefer to return `null` directly when directory doesn't exist and return `GetFiles` otherwise.

src/Integration.Vsix/BoundSolutionAnalyzer.cs
```diff
@@ -91,13 +91,12 @@ private void OnSolutionBuilding()
 
         private static string[] GetFiles(string rootDirectory, string folder, string pattern)
         {
-            string[] matches = null;
             string searchDir = Path.Combine(rootDirectory, folder);
-            if (Directory.Exists(searchDir))
+            if (!Directory.Exists(searchDir))
```
Evangelink: My teacher in school would kill me because they say we should have only 1 IN and 1 OUT but I think this is so much more readable like this. I know that we exit there instead of having to remember state later on the method.

PR #618, opened 2018-04-13
Removed unnecessary call to WriteConfiguration

duncanp-sonar:


src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -85,11 +85,6 @@ public ActiveSolutionBoundTracker(IHost host, IActiveSolutionTracker activeSolut
             // An exception here will crash VS
             try
             {
-                if (!args.IsSolutionOpen)
-                {
-                    this.configurationProvider.WriteConfiguration(BindingConfiguration.Standalone);
```
duncanp-sonar: This call to reset the binding configuration was by the configuration provider that that tracked the binding state in memory. However, that code has been removed and the code is back to the previous state of checking for the config files on disk on every call to _GetConfiguration_.

Attempting to read the files isn't ideal, but changing that behaviour is riskier than just removing this code.

duncanp-sonar:
I've added this commit to the PR to push all of the changes from a _feature/connectedModeRb_ (a rebased version of _feature/connectedMode_) to master. I'll just close this PR. Thanks for the review.

PR #627, opened 2018-04-25
Re-added ruleset/QP check for new connected mode

duncanp-sonar:


src/Integration.UnitTests/LocalServices/ErrorListInfoBarController.QualityProfileBackgroundProcessorTests.cs
```diff
@@ -100,21 +100,6 @@ public void QualityProfileBackgroundProcessor_QueueCheckIfUpdateIsRequired_ArgCh
             Exceptions.Expect<ArgumentNullException>(() => testSubject.QueueCheckIfUpdateIsRequired(null));
         }
 
-        [TestMethod]
-        public void QualityProfileBackgroundProcessor_QueueCheckIfUpdateIsRequired_NoFilteredProjects()
```
duncanp-sonar: This test and the next one seemed to be testing the same condition, so I removed this one.

src/Integration.UnitTests/LocalServices/ErrorListInfoBarController.QualityProfileBackgroundProcessorTests.cs
```diff
@@ -182,7 +152,40 @@ public void QualityProfileBackgroundProcessor_QueueCheckIfUpdateIsRequired_NoPro
         }
 
         [TestMethod]
-        public void QualityProfileBackgroundProcessor_BackgroundTask_DifferentTimestamp_RequiresUpdate()
+        public void QualityProfileBackgroundProcessor_BackgroundTask_DifferentTimestamp_RequiresUpdate_Connected()
```
duncanp-sonar: Testing:
I've duplicated all of the legacy mode tests in _SolutionBindingInformationProviderTests_ so that they are also run for connected mode.

The _SolutionBindingInformationProvider_ is used by the higher-level _ErrorListInfoBarController_ and _QualityProfileBackgroundProcessor_. I haven't added comprehensive tests to _ErrorListInfoBarControllerTests_ and _QualityProfileBackgroundProcessorTests_ - I've just added a single test to each to make sure they are passing the binding mode through correctly.

src/Integration.UnitTests/LocalServices/SolutionBindingInformationProviderTests.cs
```diff
@@ -82,15 +87,33 @@ public void SolutionBindingInformationProvider_GetUnboundProjects_SolutionBound_
         }
 
         [TestMethod]
-        public void SolutionBindingInformationProvider_GetUnboundProjects_ValidSolution_SolutionRuleSetIsMissing()
+        public void GetUnboundProjects_Connected_SolutionBound_EmptyFilteredProjects()
```
duncanp-sonar: For now, I've just duplicated the existing _legacy_ tests with the binding mode set to _connected_.
I considered try to remove some of the duplication in the test code, but decided to do that later.

src/Integration/ProfileConflicts/ConflictsManager.cs
```diff
@@ -140,7 +140,6 @@ private RuleSetInformation[] GetAggregatedSolutionRuleSets(BoundSonarQubeProject
                 string baselineRuleSet = ruleSetInfoProvider.CalculateSolutionSonarQubeRuleSetFilePath(
                     bindingInfo.ProjectKey,
                     Language.ForProject(project),
-                    // TODO: duncanp decide whether to support new connected mode
```
duncanp-sonar: I decided not to enable the rule-conflict checking for the new connected mode (i.e. the complicated code that checks whether the user has modified the ruleset to weaken it), many due to time constraints, but also because I wasn't certain that this was a behaviour we wanted to keep. If necessary, we can add it later.

src/Integration/Resources/Strings.resx
```diff
@@ -359,12 +359,6 @@ ocalization note: '[try again]()' is a special markup used to translate the text
     <value>Could not acquire service: {0}</value>
     <comment>Exception message indicating the package could not acquire the given service. {0} service name</comment>
   </data>
-  <data name="ServerNuGetTrustWarningMessage" xml:space="preserve">
```
duncanp-sonar: Unrelated: the code that used this resource string has been deleted, but the resource was left behind.

PR #628, opened 2018-04-26
Added logging and exception handling to daemon

duncanp-sonar:


duncanp-sonar:
Christophe discovered a crash where the exception was being thrown inside _daemon.Start()_. I've added logging and exception handling round all of the calls to _Start_ that didn't already have it. I've also added exception handling in all of the daemon UI and callback methods.

src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml.cs
```diff
@@ -42,23 +45,35 @@ public GeneralOptionsDialogControl(ISonarLintSettings settings, ISonarLintDaemon
...
 
-            if (!daemon.IsInstalled)
+                if (!daemon.IsInstalled)
+                {
+                    settings.IsActivateMoreEnabled = false;
+                }
+
+                UpdateActiveMoreControls();
+            }
+            catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
```
duncanp-sonar: I have the nasty feeling that all of these lines of code will be treated as changed and that the quality gate will break as a result. Fingers crossed it doesn't, as none of this code is easy to test.

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemonInstaller.xaml.cs
```diff
@@ -48,65 +51,96 @@ public SonarLintDaemonInstaller(ISonarLintSettings settings, ISonarLintDaemon da
...
         private void Window_ContentRendered(object sender, EventArgs args)
         {
-            ProgressBar.Visibility = Visibility.Visible;
-            CompletedMessage.Visibility = Visibility.Collapsed;
+            try
+            {
+                ProgressBar.Visibility = Visibility.Visible;
+                CompletedMessage.Visibility = Visibility.Collapsed;
+
+                this.logger.WriteLine(Strings.Daemon_DownloadingDaemon);
```
duncanp-sonar: I'll move this logging statement and the one in line 76 - they don't make sense where they are.

duncanp-sonar:
Christophe discovered a crash where the exception was being thrown inside _daemon.Start()_. I've added logging and exception handling round all of the calls to _Start_ that didn't already have it. I've also added exception handling in all of the daemon UI and callback methods.

duncanp-sonar:
Fixes #629

PR #630, opened 2018-04-30
Fix leak period issues

duncanp-sonar:
Minor issues: removing unused fields, sealing a class and adding a call to _GC.SuppressFinalization(this)_

### DIFF
--- a/src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
+++ b/src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarLint for Visual Studio
  * Copyright (C) 2016-2018 SonarSource SA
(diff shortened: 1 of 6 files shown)


src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
```diff
@@ -53,13 +53,21 @@ public void TestInitialize()
         [TestMethod]
         public void ProjectSystemHelper_ArgCheck()
         {
-            Exceptions.Expect<ArgumentNullException>(() => new ProjectSystemHelper(null));
+            // Arrange
+            Action act = () => new ProjectSystemHelper(null);
+
+            // Act & Assert
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("serviceProvider");
```
duncanp-sonar: I added one new arg-check test to this class, so I changed the existing arg checks to use the Fluent library.

PR #651, opened 2018-05-31
SonarQube client

valhristov:
What's changed:

- the requests are no longer in a single class, but are in separate classes for each request and version
- the request/response messages are built into the request classes because they can be versioned too (properties, structure, etc. could change)
- each request class is registered into the RequestFactory, which takes care about versioning and what request to execute, e.g. no need for if-s in the code. If you want to - to handle a breaking change, either create a new request, or override an existing.
- the requests return models, each request handles errors and http response codes (because this could also change between the api versions)
- I did not write the task cancellation tests, nor tests that ensure we throw when we are not connected and call a service method; not sure if there is big benefit from them, but should be easy enough

New functionality:

- latest API changes are applied (see last few commits)

What could be improved:

- SonarQubeService.GetQualityProfileAsync is not good, it could be moved to a single Request class that does everything, or better yet - create a CompositeRequest class that accepts several requests and executes them. 

TODO (documentation)

- write a readme.txt that explains how to create a new request, how to handle a breaking change, etc.

TODO (test plan):

- connect to SonarQube
	- invalid user/pass
	- invalid url
	- correct user/pass and url

- connect to SonarQube
	- with SonarC# installed
	- without SonarC# installed

- connect to SonarQube
	- with existing projects
		- quality profile
			- with rules
			- without rules
	- without projects

- analyze a project with suppressed issues (suppressed in SQ should be hidden in VS)

- receive notification event

All connected servers:
- should be able to show project dashboard in the VS Team Explorer window

Repeat the above steps with:
- SQ 5.6.*
- SQ 6.7.*
- SQ 7.0
- SQ 7.1
- SQ 7.2 (when available)
- SonarCloud

Repeat the above steps with:
- VS 2015
- VS 2017


src/Integration/Service/VersionHelper.cs
```diff
@@ -28,6 +29,13 @@ internal static class VersionHelper
     {
         private const char PrereleaseSeparator = '-';
 
+        public static readonly string SonarLintVersion = GetSonarLintVersion();
```
Evangelink: Use a property instead of a field.

src/SonarQube.Client.Tests/Api/Requests/V6_30/GetPluginsRequest_FormatForSonarLint.cs
```diff
@@ -0,0 +1,46 @@
...
+            GetPluginsRequest.FormatForSonarLint("1").Should().Be("1.0.0");
+            GetPluginsRequest.FormatForSonarLint("1.2").Should().Be("1.2.0");
+            GetPluginsRequest.FormatForSonarLint("1.2.3").Should().Be("1.2.3");
+            GetPluginsRequest.FormatForSonarLint("1.2.3.4").Should().Be("1.2.3");
+
+            GetPluginsRequest.FormatForSonarLint("1 (Build 2345)").Should().Be("1.0.0");
+            GetPluginsRequest.FormatForSonarLint("1 (build 2345)").Should().Be("1.0.0"); // lower case B
+            GetPluginsRequest.FormatForSonarLint("1.2 (Build 3456)").Should().Be("1.2.0");
+            GetPluginsRequest.FormatForSonarLint("1.2.3 (Build 4567)").Should().Be("1.2.3");
+        }
```
Evangelink: Add a test with a version spanning across multiple digits (e.g. 10.11.12)

src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllOrganizationsAsync.cs
```diff
@@ -0,0 +1,192 @@
...
+  ""organizations"": [
+    {string.Join(",\n", Enumerable.Range(1001, 10).Select(i => $@"{{ ""key"": ""{i}"", ""name"": ""Name{i}"" }}"))}
+  ]
+}}");
+
+            var result = await service.GetAllOrganizationsAsync(CancellationToken.None);
+
+            messageHandler.VerifyAll();
+
+            result.Should().HaveCount(1010);
```
Evangelink: Fine by me but why did you chose 1010?

src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllOrganizationsAsync.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        [TestMethod]
+        public async Task GetOrganizations_Old_SonarQube()
+        {
+            await ConnectToSonarQube("6.1.0.0");
+
+            Func<Task> action = async () => await service.GetAllOrganizationsAsync(CancellationToken.None);
+
+            action.Should().ThrowExactly<InvalidOperationException>();
```
Evangelink: Are we not expecting any specific message?

src/SonarQube.Client.Tests/Api/SonarQubeService_GetNotificationEventsAsync.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        [TestMethod]
+        public async Task GetNotifications_Old_SonarQube()
+        {
+            await ConnectToSonarQube("6.5.0.0");
+
+            Func<Task> action = async () =>
+                await service.GetNotificationEventsAsync("my_project", new DateTimeOffset(), CancellationToken.None);
+
+            action.Should().ThrowExactly<InvalidOperationException>();
```
Evangelink: Don't we have any expected message for this issue?

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -0,0 +1,70 @@
...
+        public async Task GetSuppressedIssuesAsync_ExampleFromSonarQube()
+        {
+            await ConnectToSonarQube();
+
+            SetupRequest("batch/issues?key=project1",
+                new StreamReader(@"TestResources\IssuesProtobufResponse").ReadToEnd());
+
+            var result = await service.GetSuppressedIssuesAsync("project1", CancellationToken.None);
+
+            // TODO: create a protobuf file with more than one issue with different states
```
Evangelink: How complex is it to do? I think it would be worth doing it soon.

src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllOrganizationsAsync.cs
valhristov: The message is covered in the tests for RequestFactory, not sure if we should duplicate the message everywhere.

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
valhristov: That's what we used to have. It should not be very difficult.

src/SonarQube.Client/Api/RequestBase.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// the returned string response and throws when non-200 response code.
+    /// </summary>
+    /// <typeparam name="TResponse">The type of the request result.</typeparam>
+    public abstract class RequestBase<TResponse> : IRequest<TResponse>
+    {
+        /// <summary>
+        /// Override this property to specify the relative path of this request.
+        /// </summary>
+        [JsonIgnore]
+        protected abstract string Path { get; }
```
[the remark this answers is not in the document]
duncanp-sonar: @Evangelink I'm not sure I understand - isn't it part of the responsibility of this class to know what the version-specific API path is?

src/SonarQube.Client/Api/DefaultConfiguration.cs
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+namespace SonarQube.Client.Api.Requests
+{
+    public static class DefaultConfiguration
+    {
+        public static RequestFactory Configure(RequestFactory requestFactory)
+        {
+            requestFactory
+                .RegisterRequest<IGetPluginsRequest, V2_10.GetPluginsRequest>("2.1")
```
duncanp-sonar: Perhaps the version of SQ should be a property on the Request object (rather than being a separate parameter that is passed n RegisterRequest).
The request is version-specific so it has to know the SQ version number implicitly - there doesn't seem to be any advantage in having the caller to requestFactory know this too.

src/SonarQube.Client/Api/RequestBase.cs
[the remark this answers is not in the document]
Evangelink: @duncanp-sonar It is but except if I missed something there is no need to have this as a property (this could be a ctor argument).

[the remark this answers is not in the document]
duncanp-sonar: @Evangelink do you mean a non-public constructor i.e. the child class calls the base class constructor with a hard-coded value? That feels slightly odd to me,  and I'm trying to work out why.

src/SonarQube.Client/Services/RequestFactory.cs
```diff
@@ -0,0 +1,98 @@
...
+            where TRequest : IRequest
+            where TRequestImpl : TRequest
+        {
+            SortedList<Version, Func<object>> map;
+            if (!requestMappings.TryGetValue(typeof(TRequest), out map))
+            {
+                map = new SortedList<Version, Func<object>>();
+                requestMappings[typeof(TRequest)] = map;
+            }
+            map[Version.Parse(version)] = () => factory();
```
duncanp-sonar: We don't expect an entry to already exist, do we? If not, then it would be worth asserting that this is the case to help catch coding errors.

src/SonarQube.Client/Services/RequestFactory.cs
```diff
@@ -0,0 +1,98 @@
...
+            return this;
+        }
+
+        /// <summary>
+        /// Creates a new TRequest implementation for the specified SonarQube version.
+        /// </summary>
+        /// <typeparam name="TRequest">The type of the request implementation to create.</typeparam>
+        /// <param name="version">
+        /// SonarQube version to return a request implementation for. The default value returns the
+        /// latest registered implementation.QueryStringSerializer
```
duncanp-sonar: Should "QueryStringSerializer" be part of the comment? It's not clear why it's there.

src/SonarQube.Client/Services/RequestFactory.cs
```diff
@@ -0,0 +1,98 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SonarQube.Client
+{
+    public class RequestFactory
+    {
+        private readonly Dictionary<Type, SortedList<Version, Func<object>>> requestMappings =
+            new Dictionary<Type, SortedList<Version, Func<object>>>();
```
duncanp-sonar: Does this have to be object, or could it be IRequest?

src/SonarQube.Client/Services/RequestFactory.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        /// <summary>
+        /// Registers a request factory for the specified version of SonarQube.
+        /// </summary>
+        /// <typeparam name="TRequest">The type of the request interface to create implementation for.</typeparam>
+        /// <typeparam name="TRequestImpl">The type implementing TRequest.</typeparam>
+        /// <param name="version">The version of SonarQube which first implements the request.</param>
+        /// <param name="factory">Factory function to create new instances of TRequestImpl.</param>
+        /// <returns>Returns this.</returns>
+        public RequestFactory RegisterRequest<TRequest, TRequestImpl>(string version, Func<TRequestImpl> factory)
```
duncanp-sonar: Is it ok to supply a null/empty version?

src/SonarQube.Client/Services/RequestFactory.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <summary>
+        /// Creates a new TRequest implementation for the specified SonarQube version.
+        /// </summary>
+        /// <typeparam name="TRequest">The type of the request implementation to create.</typeparam>
+        /// <param name="version">
+        /// SonarQube version to return a request implementation for. The default value returns the
+        /// latest registered implementation.QueryStringSerializer
+        /// </param>
+        /// <returns>New TRequest implementation for the specified SonarQube version.</returns>
+        public TRequest Create<TRequest>(Version version = null)
```
duncanp-sonar: Is this method ever called with Version = null?
If not, it would be better to remove it until it's needed: it's just something else to test.
(Removing it would also mean I don't have to remember why there is a design guideline about why it is better to have overloads instead of optional parameters...)

src/SonarQube.Client/Api/SonarQubeService.cs
```diff
@@ -0,0 +1,252 @@
...
+            }
+        }
+
+        public void Disconnect()
+        {
+            IsConnected = false;
+            httpClient.Dispose();
+        }
+
+        private void EnsureIsConnected()
```
duncanp-sonar: "CheckIsConnected" might be better - to me "Ensure" suggests that the method will open a connection if not already connected, not that it will throw.

src/SonarQube.Client/Api/SonarQubeService.cs
```diff
@@ -0,0 +1,252 @@
...
+        {
+            EnsureIsConnected();
+
+            const string ProjectDashboardRelativeUrl = "dashboard/index/{0}";
+
+            return new Uri(httpClient.BaseAddress, string.Format(ProjectDashboardRelativeUrl, projectKey));
+        }
+
+        public async Task<SonarQubeQualityProfile> GetQualityProfileAsync(string projectKey, string organizationKey, SonarQubeLanguage language, CancellationToken token)
+        {
```
duncanp-sonar: Is it ok for any of the parameters to be null/empty?

src/SonarQube.Client/Api/SonarQubeService.cs
```diff
@@ -0,0 +1,252 @@
...
+                    request.OrganizationKey = organizationKey;
+                },
+                token);
+
+            // Consider adding the language filter to the request configuration above and removing this line
+            var profilesWithGivenLanguage = qualityProfiles.Where(x => x.Language == language.Key).ToList();
+
+            var qualityProfile = profilesWithGivenLanguage.Count > 1
+                ? profilesWithGivenLanguage.Single(x => x.IsDefault)
+                : profilesWithGivenLanguage.Single();
```
duncanp-sonar: This is going to fail if the language isn't installed on the server (e.g. we ask for VB.NET when running against a community edition of SQ server). Is that ok? i.e. is it up to the caller to check that the appropriate plugin exists first? If so, we should document this.

src/SonarQube.Client/Api/SonarQubeService.cs
```diff
@@ -0,0 +1,252 @@
...
+            await InvokeRequestAsync<IGetRoslynExportProfileRequest, RoslynExportProfileResponse>(
+                request =>
+                {
+                    request.QualityProfileName = qualityProfileName;
+                    request.LanguageKey = language.Key;
+                    request.OrganizationKey = organizationKey;
+                },
+                token);
+
+        public async Task<IList<SonarQubeIssue>> GetSuppressedIssuesAsync(string key, CancellationToken token)
```
duncanp-sonar: Can key be null/empty?

src/SonarQube.Client/Api/SonarQubeService.cs
```diff
@@ -0,0 +1,252 @@
...
+                    request.ProjectKey = key;
+                },
+                token);
+
+            return result
+                .Where(x => x.ResolutionState != SonarQubeIssueResolutionState.Unresolved)
+                .ToList();
+        }
+
+        public async Task<IList<SonarQubeNotification>> GetNotificationEventsAsync(string projectKey, DateTimeOffset eventsSince,
```
duncanp-sonar: Can projectKey be null/empty?

src/SonarQube.Client/Api/Requests/V5_20/GetQualityProfilesRequest.cs
```diff
@@ -0,0 +1,93 @@
...
+                .ToObject<QualityProfileResponse[]>()
+                .Select(FromResponse)
+                .ToArray();
+
+        private static SonarQubeQualityProfile FromResponse(QualityProfileResponse response) =>
+            new SonarQubeQualityProfile(
+                response.Key, response.Name, response.Language, response.IsDefault, response.LastRuleChange);
+
+        // This class MUST NOT change! If a breaking change in the API is introduced,
+        // create a new versioned request and reimplement the serialization there
```
duncanp-sonar: Is there something special about this class, or does the warning apply to all of the data classes?

src/SonarQube.Client/Api/Requests/V5_20/GetQualityProfileChangeLogRequest.cs
```diff
@@ -0,0 +1,62 @@
...
+    public class GetQualityProfileChangeLogRequest : PagedRequestBase<DateTime>, IGetQualityProfileChangeLogRequest
+    {
+        [JsonProperty("profileKey")]
+        public virtual string QualityProfileKey { get; set; }
+
+        [JsonProperty("qualityProfile", DefaultValueHandling = DefaultValueHandling.Ignore), DefaultValue(null)]
+        public virtual string QualityProfileName
+        {
+            get { return null; }
+            set { /* to prevent the property from serialization we always return the default */ }
```
duncanp-sonar: Why do we want to this property and LanguageName always to be null and not to be serialized?

src/SonarQube.Client/Api/Requests/V5_10/GetIssuesRequest.cs
```diff
@@ -0,0 +1,96 @@
...
+                case "":
+                    return SonarQubeIssueResolutionState.Unresolved;
+                case "WONTFIX":
+                    return SonarQubeIssueResolutionState.WontFix;
+                case "FALSE-POSITIVE":
+                    return SonarQubeIssueResolutionState.FalsePositive;
+                case "FIXED":
+                    return SonarQubeIssueResolutionState.Fixed;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(resolution));
```
duncanp-sonar: It might be better to assert and ignore (or treat it as "unknown") if we get an unrecognised value.
It's not likely that a new resolution state would be added to SonarQube/Cloud, but if one is then all existing client installations using that version of SQ will start failing until we roll out an update and the customer installs it.

src/SonarQube.Client.Tests/Services/RequestFactoryTests.cs
```diff
@@ -0,0 +1,98 @@
...
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SonarQube.Client.Tests.Services
+{
+    [TestClass]
+    public class RequestFactoryTests
+    {
+        [TestMethod]
+        public void Create_Throws_When_Not_Regitered()
```
duncanp-sonar: Typo - "Regitered"

src/SonarQube.Client.Tests/Services/RequestFactoryTests.cs
```diff
@@ -0,0 +1,98 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
```
duncanp-sonar: No tests for Create without a version being specified

src/SonarQube.Client.Tests/Services/RequestFactoryTests.cs
```diff
@@ -0,0 +1,98 @@
...
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SonarQube.Client.Tests.Services
```
duncanp-sonar: No tests for Register:
* missing/invalid version - what should happen?
* registering the same type/version twice
* registering the same version for different types -> correct type returned

src/SonarQube.Client/Api/DefaultConfiguration.cs
valhristov: Warning: this cannot be an instance property like `Path` because in the factory we don't have instances before resolve the versioned request and that's too late.

src/SonarQube.Client/Services/RequestFactory.cs
valhristov: It could be IRequest

src/SonarQube.Client/Api/SonarQubeService.cs
valhristov: If it is null or empty we will pass empty query string argument, which is ok, I think. I don't know what SQ will do in this case, though :)

src/SonarQube.Client/Api/SonarQubeService.cs
valhristov: If it is null or empty we will pass empty query string argument, which is ok, I think. I don't know what SQ will do in this case, though :)

src/SonarQube.Client/Api/Requests/V5_20/GetQualityProfilesRequest.cs
valhristov: It applies to all data classes. Perhaps I should remove the comment, because it should be copied on all classes, which is inconvenient.

src/SonarQube.Client/Api/Requests/V5_20/GetQualityProfileChangeLogRequest.cs
valhristov: Wt does not exist in this version of the API and there is no point sending it.

src/SonarQube.Client/Api/Requests/V5_10/GetIssuesRequest.cs
valhristov: I agree about this. However, asserting will happen only when debugging, e.g. when testing manually... Instead of asserting I would rather write a log entry, but since this exception was already there, I prefer to do it in another PR. WDYT?

src/SonarQube.Client/Api/Requests/V5_20/GetQualityProfileChangeLogRequest.cs
valhristov: I agree this adds some complexity, but unfortunately once a property is `JsonIgnore`d it cannot be un-ignored. Other options are:
- create completely new implementation (same code - duplication, not less complex)
- create base class without this property (more classes, not less complex)
- always send the property (not sure what will happen in future versions of SQ if we send invalid property and is generally not very clean imo)

src/SonarQube.Client/Api/SonarQubeService.cs
valhristov: I would prefer to fix this bug in a separate PR (with a corresponding ticket) because it already exists in SLVS (this is a copy of the existing code).

src/SonarQube.Client/Api/SonarQubeService.cs
valhristov: It is reasonable to at least log when there is organization key while SQ does not support organizations, because this would mean that we have inconsistency in our data... This is applicable for most of the API calls and I would prefer to implement this as a separate PR/ticket.

src/SonarQube.Client/Api/DefaultConfiguration.cs
valhristov: I could invert the code, for example in DefaultConfiguration:
```
```
GetQualiryProfilesRequest.Register(factory);
...
```

Bug imo this does not improve the code that much as we cannot use a base class to define the method and we will have to duplicate...

File: src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllOrganizationsAsync.cs (hunk 3 above)
Comment (valhristov) [reply src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllOrganizationsAsync.cs]: It could be 1001... First page has 500 items, second - 500, third 10 => 1010.

File: src/SonarQube.Client/Api/PagedRequestBase.cs [hunk 25]
```diff
@@ -0,0 +1,65 @@
...
+
+            Result<TResponseItem[]> pageResult;
+            do
+            {
+                pageResult = await InvokeImplAsync(httpClient, token);
+                pageResult.EnsureSuccess();
+
+                result.AddRange(pageResult.Value);
+
+                Page++;
```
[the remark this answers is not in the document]
valhristov: It seems easier to loop until the current page returns less items than the page size...

src/SonarQube.Client/Api/PagedRequestBase.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// Abstract implementation of IPagedRequest<typeparamref name="TResponseItem"/> that automatically downloads all pages
+    /// from the server.
+    /// </summary>
+    /// <typeparam name="TResponseItem">The type of the items returned by this request.</typeparam>
+    public abstract class PagedRequestBase<TResponseItem> : RequestBase<TResponseItem[]>, IPagedRequest<TResponseItem>
+    {
+        private const int FirstPage = 1;
+        private const int MaximumPageSize = 500;
+
+        [JsonProperty("p")]
```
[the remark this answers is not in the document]
valhristov: As far as I could tell we don't use requests that require different parameter name.

src/SonarQube.Client/Api/Requests/V5_10/GetIssuesRequest.cs
duncanp-sonar: Ok: you might want to put a TODO in so we don't forget.

src/SonarQube.Client/Api/RequestBase.cs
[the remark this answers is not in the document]
valhristov: I don't see big difference, except that each request class should define a ctor that calls the base with an argument...
```
public class Request : RequestBase, MyRequest
{

File: src/SonarQube.Client/Api/Requests/V5_20/GetQualityProfileChangeLogRequest.cs (hunk 20 above)
Comment (duncanp-sonar) [reply src/SonarQube.Client/Api/Requests/V5_20/GetQualityProfileChangeLogRequest.cs]: I think the current implementation is fine: the only problem was that it wasn't clear from the comment why it is like that.
I think adding your explanation above would be enough e.g. _Property does not exist in this version of the API so no point in sending it_

File: src/SonarQube.Client/Api/RequestBase.cs [hunk 27]
```diff
@@ -0,0 +1,111 @@
...
+            return await ReadResponseAsync(httpResponse);
+        }
+
+        /// <summary>
+        /// Override this method to change the way the response content is read and deserialized from the HttpResponseMessage
+        /// created while invoking the request.
+        /// </summary>
+        /// <param name="httpResponse">HttpResponseMessage created when invoking the request.</param>
+        /// <returns>Result instance that contains the HttpStatusCode and the deserialized request response.</returns>
+        protected virtual async Task<Result<TResponse>> ReadResponseAsync(HttpResponseMessage httpResponse)
```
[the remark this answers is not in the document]
valhristov: I suggest improving that as a separate PR/Ticket as the code was copied from the original as is.

src/SonarQube.Client/Api/Requests/V5_10/GetIssuesRequest.cs
```diff
@@ -0,0 +1,96 @@
...
+                    ReadFromProtobufStream(stream, ServerIssue.Parser).Select(ToSonarQubeIssue).ToArray());
+            }
+        }
+
+        protected override SonarQubeIssue[] ParseResponse(string response)
+        {
+            throw new NotImplementedException("This method will not be called because we override ReadResponse.");
+        }
+
+        private static IEnumerable<T> ReadFromProtobufStream<T>(Stream stream, MessageParser<T> parser)
```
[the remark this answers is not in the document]
valhristov: I suggest improving that as a separate PR/Ticket as the code was copied from the original as is.

src/SonarQube.Client/Api/RequestBase.cs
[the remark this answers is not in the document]
valhristov: +1 in general, however:
- no method of httpResponse.Content accepts a CancellationToken, hence the parameter would be useless in this method
- reading from protobuf could (should imo) be improved as a separate PR/Ticket

src/SonarQube.Client.Tests/Api/SonarQubeService_GetNotificationEventsAsync.cs
duncanp-sonar: I'd say leave the test as it is.

src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllOrganizationsAsync.cs
valhristov: Also it is a nice number - if `1010` is binary, it converts to `10` in decimal 💃 🔢

src/SonarQube.Client/Api/SonarQubeService.cs
valhristov: On a second thought... `Ensure*` seems in sync with `HttpResponseMessage.EnsureSuccessStatusCode()`, which also throws when the state of the object is not expected.

src/SonarQube.Client/Api/SonarQubeService.cs
duncanp-sonar: They are both required by the SQ API so presumably we'll get an error code back which will be handled as a failure by the client.
Is that the general approach that we want to take for the api calls - to leave the validation to the server? (seems ok to me)

src/SonarQube.Client/Api/SonarQubeService.cs
valhristov: #662

src/SonarQube.Client/Api/SonarQubeService.cs
valhristov: #663

src/SonarQube.Client/Api/Requests/V5_10/GetIssuesRequest.cs
valhristov: #663

src/SonarQube.Client/Api/DefaultConfiguration.cs
duncanp-sonar: It breaks encapsulation and information hiding, but since we're using static registration I guess that risk of us getting it wrong is low. I'm ok with leaving it as it is.

FYI I was trying to think of an alternative approach, and the only one I could come up with is to create a request factory interface for each type of request. That would mean that current _requestFactory_ wouldn't have to know the versions that are supported by each request type, and it would also mean each factory interface could have a strongly-typed _Create_ method with the parameters required for that API (instead of the current _configure_ lambda).

src/SonarQube.Client/Api/PagedRequestBase.cs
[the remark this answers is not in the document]
Evangelink: Thx for the check!

src/SonarQube.Client/Api/PagedRequestBase.cs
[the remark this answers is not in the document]
Evangelink: Ok but is it always true? I know they have strange logic on SQ api side.

src/SonarQube.Client/Api/RequestBase.cs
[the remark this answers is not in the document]
Evangelink: @duncanp-sonar @valhristov my suggestion seems off to you but the current one seems off to me so let's not change anything and we will see later on.

src/SonarQube.Client/Api/RequestBase.cs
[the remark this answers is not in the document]
Evangelink: I have seen you created the ticket so let's consider this point as handled.

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
duncanp-sonar: @Val - I think this comment still needs a response

src/SonarQube.Client/Api/RequestBase.cs
```diff
@@ -0,0 +1,111 @@
...
+using SonarQube.Client.Services;
+
+namespace SonarQube.Client
+{
+    /// <summary>
+    /// Abstract implementation of IRequest<typeparamref name="TResponse"/> that automatically deserializes
+    /// the returned string response and throws when non-200 response code.
+    /// </summary>
+    /// <typeparam name="TResponse">The type of the request result.</typeparam>
+    public abstract class RequestBase<TResponse> : IRequest<TResponse>
```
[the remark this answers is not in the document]
Evangelink: @valhristov Not sure how long fixing this point would take but if you think that's easy enough I think that would be a nice improvement.

src/Integration/Service/VersionHelper.cs
Evangelink: @valhristov This still needs to be fixed

src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllOrganizationsAsync.cs
Evangelink: @valhristov there too.

src/SonarQube.Client/Api/Requests/V5_20/GetQualityProfileChangeLogRequest.cs
```diff
@@ -38,11 +38,15 @@ public virtual string QualityProfileName
             set { /* to prevent the property from serialization we always return the default */ }
         }
 
+        /// <summary>
```
duncanp-sonar: ?Same comment applies to the _QualityProfileName_ immediately above?

src/SonarQube.Client/Api/RequestBase.cs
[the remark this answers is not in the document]
valhristov: Probably not much, I generally like the idea, but I am afraid the hierarchy will become very complex:
- RequestBase
	- JsonRequestBase
  		- PaginatedJsonRequestBase
   			- ... all paginated Json requests ...
  		- ... all Json requests ...
 	- ProtobufRequestBase
  		- the single Protobuf request we have

while the current is:
- RequestBase
	- all Json requests
	- the single Protobuf request with little hacky override
- PaginatedRequestBase
	- all paginated requests

I have a very "architectural" idea, but I'll need some time to validate it :)

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
valhristov: #669

src/SonarQube.Client/Api/SonarQubeService.cs
valhristov: IMO the client should follow the API specification and if an argument is missing in a particular version of SQ it should not be sent.

PR #654, opened 2018-06-05
Fix telemetry dates by handling properly cultures

### LINKED ISSUE #653: Fix telemetry dates by handling properly cultures
### Description

We are currently receiving some invalid dates from the telemetry. After some investigation it appears that we are not handling properly cultures.

Evangelink:
Fix #653

### DIFF
--- a/src/Integration/Telemetry/ShortIsoDateTimeOffsetConverter.cs
+++ b/src/Integration/Telemetry/ShortIsoDateTimeOffsetConverter.cs
@@ -24,7 +24,7 @@ using Newtonsoft.Json;
 namespace SonarLint.VisualStudio.Integration.Telemetry
 {
     /// <summary>
-    /// Converts DateTimeOffset to roundtrip (ISO 8061) but with just milliseconds, no ticks
+    /// Converts DateTimeOffset to roundtrip (ISO 8601) but with just milliseconds, no ticks
     /// </summary>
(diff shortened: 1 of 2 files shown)


src/Integration/Telemetry/ShortIsoDateTimeOffsetConverter.cs
```diff
@@ -34,7 +34,17 @@ internal class ShortIsoDateTimeOffsetConverter : JsonConverter
...
-            writer.WriteValue(((DateTimeOffset)value).ToString("yyyy-MM-ddTHH:mm:ss.fffzzz"));
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            // Use the standard format string for ISO 8601
+            var iso8601Date = ((DateTimeOffset)value).ToString("o");
+
+            // Remove the ticks part (not expected by our server side)
+            const int millisecondLastIndex = 23;
+            const int timeOffsetFirstIndex = 27;
+            var simplifiedDate = iso8601Date.Substring(0, millisecondLastIndex) + iso8601Date.Substring(timeOffsetFirstIndex);
```
valhristov: Perhaps write a test ensuring that the width of the date parts is always the same, e.g. with single digit month, day, hour, min, sec, etc.

PR #668, opened 2018-06-13
Update extension description to fit value proposition

n-b-sonar:


n-b-sonar:
You deserve an answer @valhristov ! (although I see you already merged) So it's a fair point, but for now we'll keep it as is as it allows us to have a common blurb that we can reuse as needed (+the current phrasing doesn't do much harm even if already in marketplace).

PR #685, opened 2018-06-28
Changed notifications package to be asynchronous

### LINKED ISSUE #679: Convert SonarLintNotificationsPackage to be asynchronous
### Background
Synchronously loaded packages delay Visual Studio startup and can result in a gold bar warning the user that the extension is slowing VS startup and suggesting that they disable it. In addition, synchronous packages will deprecated at some point. We need to convert all of our packages to be asynchronous.

See MSDN: [How to: Use AsyncPackage to Load VSPackages in the Background](https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-use-asyncpackage-to-load-vspackages-in-the-background)

### Related
#647

duncanp-sonar:
Resolved #679 

I followed the pattern Mads used for the [SonarLintDaemonPackage](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/SonarLintDaemonPackage.cs#L49) and the MSDN guidance [here](https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-diagnose-extension-performance).

I've manually tested with VS2015 and VS2017 and the notification package no longer appears in the _Visual Studio Performance_ entries in the activity log.
However, the telemetry package does, so that's next.

### DIFF
--- a/src/Integration.Vsix/Resources/Strings.resx
+++ b/src/Integration.Vsix/Resources/Strings.resx
@@ -169,4 +169,32 @@
     <value>Invalid VisualStudio version, expecting '14.0' or '15.0' got '{0}'.</value>
     <comment>Error message in output window when the VisualStudio version is not the expected one. {0} Actual VisualStudio version</comment>
   </data>
+  <data name="Notifications_Connected" xml:space="preserve">
+    <value>Connected: checking for notifications</value>
+    <comment>Output window message</comment>
+  </data>
+  <data name="Notifications_ERROR" xml:space="preserve">
+    <value>Notifications: error occurred: {0}</value>
+    <comment>Output window message</comment>
+  </data>
+  <data name="Notifications_InitializationComplete" xml:space="preserve">
+    <value>Finished initializing the notifications package</value>
+    <comment>Output window message</comment>
+  </data>
+  <data name="Notifications_Initializing" xml:space="preserve">
+    <value>Initializing the notifications package...</value>
+    <comment>Output window message</comment>
+  </data>
+  <data name="Notifications_LoadingSettings" xml:space="preserve">
+    <value>Loading notifcations settings...</value>
+    <comment>Output window message</comment>
+  </data>
+  <data name="Notifications_NotConnected" xml:space="preserve">
+    <value>Not connected: not checking for notifications</value>
+    <comment>Output window message</comment>
+  </data>
+  <data name="Notifications_SavingSettings" xml:space="preserve">
+    <value>Saving notifcations settings...</value>
+    <comment>Output window message</comment>
+  </data>
 </root>
--- a/src/Integration.Vsix/SonarLintNotificationsPackage.cs
+++ b/src/Integration.Vsix/SonarLintNotificationsPackage.cs
@@ -19,25 +19,28 @@
  */
 
 using System;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Binary;
+using System.Threading;
+using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
+using SonarLint.VisualStudio.Integration.NewConnectedMode;
 using SonarLint.VisualStudio.Integration.Notifications;
 using SonarQube.Client.Services;
 
 namespace SonarLint.VisualStudio.Integration.Vsix
 {
     [ExcludeFromCodeCoverage] // Not easily testable, this class should be kept as simple as possible
-    [PackageRegistration(UseManagedResourcesOnly = true)]
+    [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid(PackageGuidString)]
     [SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1650:ElementDocumentationMustBeSpelledCorrectly", Justification = "pkgdef, VS and vsixmanifest are valid VS terms")]
-    [ProvideAutoLoad(UIContextGuids.NoSolution)]
-    [ProvideAutoLoad(UIContextGuids.SolutionExists)]
-    public sealed class SonarLintNotificationsPackage : Package
+    [ProvideAutoLoad(UIContextGuids.SolutionExists, PackageAutoLoadFlags.BackgroundLoad)]
+    public sealed class SonarLintNotificationsPackage : AsyncPackage
     {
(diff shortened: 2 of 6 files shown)


src/Integration.Vsix/Resources/Strings.resx
```diff
@@ -169,4 +169,32 @@
     <value>Invalid VisualStudio version, expecting '14.0' or '15.0' got '{0}'.</value>
     <comment>Error message in output window when the VisualStudio version is not the expected one. {0} Actual VisualStudio version</comment>
   </data>
+  <data name="Notifications_Connected" xml:space="preserve">
+    <value>Connected: checking for notifications</value>
```
duncanp-sonar: I added some output messages to help track when the package is loaded.

src/Integration.Vsix/SonarLintNotificationsPackage.cs
```diff
@@ -46,40 +49,84 @@ public sealed class SonarLintNotificationsPackage : Package
         private const string NotificationDataKey = "NotificationEventData";
 
         private readonly IFormatter formatter = new BinaryFormatter();
-        private readonly NotificationIndicator notificationIcon = new NotificationIndicator();
+        private NotificationIndicator notificationIcon;
 
         private IActiveSolutionBoundTracker activeSolutionBoundTracker;
         private ISonarQubeNotificationService notifications;
-        private ILogger sonarLintOutput;
+        private ILogger logger;
```
duncanp-sonar: Miscellaneous - renamed.

src/Integration.Vsix/SonarLintNotificationsPackage.cs
```diff
@@ -46,40 +49,84 @@ public sealed class SonarLintNotificationsPackage : Package
         private const string NotificationDataKey = "NotificationEventData";
 
         private readonly IFormatter formatter = new BinaryFormatter();
-        private readonly NotificationIndicator notificationIcon = new NotificationIndicator();
+        private NotificationIndicator notificationIcon;
```
duncanp-sonar: The notification icon contains WPF elements so it has to be created on the UI thread.
The package is now loaded on a background thread so the field can't be created using a field initializer.

src/Integration.Vsix/SonarLintNotificationsPackage.cs
```diff
@@ -46,40 +49,84 @@ public sealed class SonarLintNotificationsPackage : Package
...
+        protected override System.Threading.Tasks.Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
         {
-            AddOptionKey(NotificationDataKey);
-            base.Initialize();
+            JoinableTaskFactory.RunAsync(Init);
+            return System.Threading.Tasks.Task.CompletedTask;
+        }
+
+        private async System.Threading.Tasks.Task Init()
+        {
```
duncanp-sonar: I've split the initialisation into two parts - the stuff that can be done on the background thread, and the stuff that has to be done on the UI thread. The UI initialisation is in a separate method.

src/Integration.Vsix/SonarLintNotificationsPackage.cs
```diff
@@ -46,40 +49,84 @@ public sealed class SonarLintNotificationsPackage : Package
...
 
             notifications = new SonarQubeNotificationService(sonarqubeService,
-                new NotificationIndicatorViewModel(), new TimerWrapper { Interval = 60000 }, sonarLintOutput);
+                new NotificationIndicatorViewModel(), new TimerWrapper { Interval = 60000 }, logger);
+
+            activeSolutionBoundTracker = await this.GetMefServiceAsync<IActiveSolutionBoundTracker>();
 
+            // A bound solution might already have completed loading. If so, we need to
+            // trigger the load of the options from the solution file
+            if (activeSolutionBoundTracker.CurrentConfiguration.Mode != SonarLintMode.Standalone)
```
duncanp-sonar: This is new code that wasn't required before.

src/Integration.Vsix/SonarLintNotificationsPackage.cs
```diff
@@ -46,40 +49,84 @@ public sealed class SonarLintNotificationsPackage : Package
...
+
+            // Ordering: don't register for solution change notifications
+            // until we've finished setting up everything else
             activeSolutionBoundTracker.SolutionBindingChanged += OnSolutionBindingChanged;
         }
 
         private void OnSolutionBindingChanged(object sender, ActiveSolutionBindingEventArgs binding)
         {
-            if (binding.Configuration.Mode != NewConnectedMode.SonarLintMode.Standalone)
+            SafePerformOpOnUIThread(() => Refresh(binding.Configuration));
```
duncanp-sonar: Miscellaneous: I've tried to pull methods called on a UI thread inside a _try-catch_ block so unhandled exceptions won't crash VS.

src/Integration/Resources/Strings.resx
```diff
@@ -723,4 +719,12 @@ This is the general format to use when writing errors to output window or when t
   <data name="Bind_NuGetAnalyzersNoLongerInstalled" xml:space="preserve">
     <value>INFO: SonarLint connected mode no longer installs the analyzers NuGet packages. The analyzers embedded in the VSIX are used instead.</value>
   </data>
+  <data name="Notifications_NotSupported" xml:space="preserve">
+    <value>Notifications are not supported on this version of SonarQube</value>
+    <comment>Output window message</comment>
+  </data>
+  <data name="OnlySupportedPluginHasNoProjectInSolution" xml:space="preserve">
```
duncanp-sonar: This message already existed  It has been relocated from line 649 - not sure why,

src/Integration.Vsix/SonarLintNotificationsPackage.cs
```diff
@@ -46,40 +49,84 @@ public sealed class SonarLintNotificationsPackage : Package
...
+        }
+        private void PerformUIInitialisation()
+        {
+            notificationIcon = new NotificationIndicator();
             notificationIcon.DataContext = notifications.Model;
 
-            activeSolutionBoundTracker = this.GetMefService<IActiveSolutionBoundTracker>();
+            // The package is now loaded asynchronously, so a solution might have
+            // finished loading before this package is initialized
+            Refresh(activeSolutionBoundTracker.CurrentConfiguration);
```
valhristov: I think the NotificationIndicator could be initialized as before, you only need add it to the VisualTree on the UI thread and probably set DataContext, but I am not sure for the latter.

duncanp-sonar: An exception was being thrown from the field initializer which is what prompted me to change it (like you, I expected the creation to be ok, but it seems it isn't).

PR #687, opened 2018-06-29
Convert telemetry package to load asynchronously

### LINKED ISSUE #681: Convert SonarLintTelemetryPackage to be asynchronous
See #679 for more information

Related to #647

duncanp-sonar:
Resolved #681

### DIFF
--- a/src/Integration.Vsix/SonarLintTelemetryPackage.cs
+++ b/src/Integration.Vsix/SonarLintTelemetryPackage.cs
@@ -86,11 +86,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix
             object isLoaded;
             var hresult = solution.GetProperty((int)__VSPROPID4.VSPROPID_IsSolutionFullyLoaded, out isLoaded);
 
-            if (ErrorHandler.Succeeded(hresult) && isLoaded is Boolean)
-            {
-                return (bool)isLoaded;
-            }
-            return false;
+            return ErrorHandler.Succeeded(hresult) && (isLoaded as bool?) == true;
         }
 
         protected override void Dispose(bool disposing)
--- a/src/Integration.Vsix/SonarLintTelemetryPackage.cs
+++ b/src/Integration.Vsix/SonarLintTelemetryPackage.cs
@@ -19,28 +19,73 @@
  */
 
 using System;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.InteropServices;
+using System.Threading;
+using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 
 namespace SonarLint.VisualStudio.Integration.Vsix
 {
-    [PackageRegistration(UseManagedResourcesOnly = true)]
+    [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid(PackageGuidString)]
-    [ProvideAutoLoad(UIContextGuids.NoSolution)]
+    [ProvideAutoLoad(UIContextGuids.NoSolution, PackageAutoLoadFlags.BackgroundLoad)]
     [ExcludeFromCodeCoverage] // Simple bootstrapper class relying on Visual Studio
-    public sealed class SonarLintTelemetryPackage : Package
+    public sealed class SonarLintTelemetryPackage : AsyncPackage
     {
         public const string PackageGuidString = "4E057B4B-E2B8-490D-95D8-2A1A4E7ACAED";
 
         private ITelemetryManager telemetryManager;
 
-        protected override void Initialize()
+        protected async override System.Threading.Tasks.Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
         {
-            base.Initialize();
+            Debug.Assert(!ThreadHelper.CheckAccess(), "Expecting the packatge to be initialized on the UI thread");
 
-            telemetryManager = this.GetMefService<ITelemetryManager>();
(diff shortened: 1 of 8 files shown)


src/Integration.Vsix/SonarLintTelemetryPackage.cs
```diff
@@ -19,28 +19,73 @@
...
+                logger.WriteLine(Resources.Strings.Telemetry_Initializing);
+                telemetryManager = await this.GetMefServiceAsync<ITelemetryManager>();
+                logger.WriteLine(Resources.Strings.Telemetry_InitializationComplete);
+
+                if (await IsSolutionFullyOpened())
+                {
+                    telemetryManager.Update();
+                }
+            }
+            catch(Exception ex) when (!ErrorHandler.IsCriticalException(ex))
```
duncanp-sonar: Added exception handling since on the UI thread.

src/Integration.Vsix/SonarLintTelemetryPackage.cs
```diff
@@ -19,28 +19,73 @@
...
+            // The ui and non-ui parts of the host should be split into separate classes.
+            try
+            {
+                logger.WriteLine(Resources.Strings.Telemetry_Initializing);
+                telemetryManager = await this.GetMefServiceAsync<ITelemetryManager>();
+                logger.WriteLine(Resources.Strings.Telemetry_InitializationComplete);
+
+                if (await IsSolutionFullyOpened())
+                {
+                    telemetryManager.Update();
```
duncanp-sonar: The solution could be loaded before the package is loaded so we could lose the first call to the telemetry manager. I was in two minds about whether to bother about losing the event, but in the end I decided it was simple so I should just add it.

src/Integration.Vsix/SonarLintTelemetryPackage.cs
```diff
@@ -19,28 +19,73 @@
...
+
+        private async System.Threading.Tasks.Task<bool> IsSolutionFullyOpened()
+        {
+            Debug.Assert(ThreadHelper.CheckAccess(), "Expecting to be called on the UI thread");
+            var solution = await this.GetServiceAsync(typeof(SVsSolution)) as IVsSolution;
+            Debug.Assert(solution != null, "Cannot find SVsSolution");
+
+            var hresult = solution.GetProperty((int)__VSPROPID4.VSPROPID_IsSolutionFullyLoaded, out object value);
+
+            return ErrorHandler.Succeeded(hresult) && value is bool isOpen && isOpen;
```
duncanp-sonar: There is a non-async method in our _IProjectSystemHelper_ that does this. However, I didn't want to add a dependency on that class just for this simple call, so I've added an async version here.

src/Integration.Vsix/SonarLintTelemetryPackage.cs
```diff
@@ -19,28 +19,78 @@
...
+                logger.WriteLine(Resources.Strings.Telemetry_Initializing);
+                telemetryManager = await this.GetMefServiceAsync<ITelemetryManager>();
+                logger.WriteLine(Resources.Strings.Telemetry_InitializationComplete);
+
+                if (await IsSolutionFullyOpened())
+                {
+                    telemetryManager.Update();
+                }
+            }
+            catch(Exception ex) when (!ErrorHandler.IsCriticalException(ex))
```
Evangelink: [cosmetic] You forgot the space before the parenthesis.

src/Integration.Vsix/SonarLintTelemetryPackage.cs
```diff
@@ -19,28 +19,78 @@
...
+        private async System.Threading.Tasks.Task<bool> IsSolutionFullyOpened()
+        {
+            Debug.Assert(ThreadHelper.CheckAccess(), "Expecting to be called on the UI thread");
+            var solution = await this.GetServiceAsync(typeof(SVsSolution)) as IVsSolution;
+            Debug.Assert(solution != null, "Cannot find SVsSolution");
+
+            object isLoaded;
+            var hresult = solution.GetProperty((int)__VSPROPID4.VSPROPID_IsSolutionFullyLoaded, out isLoaded);
+
+            if (ErrorHandler.Succeeded(hresult) && isLoaded is Boolean)
```
Evangelink: Let's rewrite this if/else as a return: `return ErrorHandler.Succeeded(hresult) && isLoaded is Boolean && isLoaded;`

src/Integration/Telemetry/TelemetryManager.cs
```diff
@@ -137,17 +138,27 @@ private TelemetryPayload GetPayload(TelemetryData telemetryData)
...
+            {
+                var lastAnalysisDate = telemetryRepository.Data.LastSavedAnalysisDate;
+                if (!DateTimeOffset.Now.IsSameDay(lastAnalysisDate))
+                {
+                    telemetryRepository.Data.LastSavedAnalysisDate = DateTimeOffset.Now;
+                    telemetryRepository.Data.NumberOfDaysOfUse++;
+                    telemetryRepository.Save();
+                }
+            }
+            catch(Exception ex) when (!ErrorHandler.IsCriticalException(ex))
```
Evangelink: Shouldn't we log in here too?

src/Integration.Vsix/SonarLintTelemetryPackage.cs
duncanp-sonar: Did you mean _...& isLoaded is Boolean boolIsLoaded && boolIsLoaded_?
That won't compile in VS2015.

Evangelink: arf VS2015.... what about `ErrorHandler.Succeeded(hresult) && (isLoaded as bool?) == true`?

Issue #696, opened 2018-07-24
SonarQubeService allows requests to be made before the initial connection is complete


duncanp-sonar:
### Description

When a connected solution is opened in VS, _SonarQubeService.ConnectAsync_ is called to establish the connection to the server. This makes a couple of server round-trips i.e. establishing the connection is not instantaneous and could fail.

However, the service is used from code running on different background threads (e.g. the daemon manager and the notification manager). Those managers can make requests against the service before the initial connection logic has completed.

This leads to a race condition between the connection logic and the background requests.

### Repro steps

1. Open VS
2. Open a bound project
3. Look at the messages in the output window.

### Expected behavior

The behaviour should be deterministic e.g. requests to the service are blocked until the connection logic has completed (although we must not block the UI thread).

### Actual behavior

Depends on timing whether this issue will occur or not, and the effects are difficult to predict.

One possible failure condition is that a background request is made before the server version has been retrieved, in which case the call will assume that the highest version number should be used. This will cause an exception if the user has an older, incompatible version of SonarQube.

It's difficult to know what impact this will have on the user experience as there are multiple potential background callers. The caller might simply output an error to the output window and retry later, in which case the impact would be low.

```

### Known workarounds

None

### Related information

* SonarLint for VS Version v4

**Comment by duncanp-sonar** [comment]
Example of the sort of issue this can lead to:

```
[Taint] Failed to synchronize taint vulnerabilities with the connected server: System.NullReferenceException: Object reference not set to an instance of an object.
   at SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintIssuesSynchronizer.IsFeatureSupported()
   at SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintIssuesSynchronizer.<SynchronizeWithServer>d__10.MoveNext().
```

#### Product code
https://github.com/SonarSource/sonarlint-visualstudio/blob/63a9400ef82f8e62298b8ab2eb433723dc17a8f8/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs#L148

The following code fails because `sonarQubeServuce.IsConnected` returns true, but `sonarQubeService.ServerInfo` is null -> null reference exception.

```
        private bool IsFeatureSupported()
        {
            if (sonarQubeService.ServerInfo.ServerType == ServerType.SonarCloud ||
                sonarQubeService.ServerInfo.Version >= MinimumRequiredSonarQubeVersion)
            {
                return true;
            }

[...]

PR #697, opened 2018-07-25
Fixed #689 - LoggingHttpClientHandler is disposed when disconnecting

### LINKED ISSUE #689: [Client] LoggingHttpClientHandler is disposed when a bound solution is closed
### Description

The _LoggingHttpClientHandler_ is disposed when a bound solution is closed, causing all subsequent calls to SonarQube to fail until VS is restarted.

### Repro steps

1. Open a bound solution -> connects correctly to SQ
2. Close the bound solution
3. Re-open the bound solution

### Expected behavior

The connection is re-opened successfully.

### Actual behavior

A gold bar appears with the message _Cannot connect to the SonarQube server.  Make sure you provided the ...._

The output window contains the following:
```
SonarQube request failed: Cannot access a disposed object.
Object name: 'SonarLint.VisualStudio.Integration.Helpers.LoggingHttpClientHandler'. 
```

### Known workarounds

Re-start VS

### Related information

* SonarLint for VS Version: 4.2 (unreleased)

duncanp-sonar:


src/SonarQube.Client.Tests/Api/SonarQubeService_Lifecycle.cs
```diff
@@ -23,12 +23,14 @@
...
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using Moq.Protected;
 using SonarQube.Client.Helpers;
 
 namespace SonarQube.Client.Tests.Api
 {
     [TestClass]
-    public class SonarQubeService_ConnectAsync : SonarQubeService_TestBase
+    public class SonarQubeService_Lifecycle : SonarQubeService_TestBase
```
duncanp-sonar: The tests for connecting, disconnecting and disposing are all in this class, so I renamed it (didn't seem worth creating a separate test class for testing _Dispose_ with only a single test in it).

src/SonarQube.Client/Services/RequestFactory.cs
```diff
@@ -98,7 +98,7 @@ public TRequest Create<TRequest>(Version version)
             SortedList<Version, Func<IRequest>> map;
             if (registrations.TryGetValue(typeof(TRequest), out map))
             {
-                Log($"Looking up implementation of '{typeof(TRequest).Name}' for version '{version}'.");
+                Log($"Looking up implementation of '{typeof(TRequest).Name}' for version '{version}' on thread '{System.Threading.Thread.CurrentThread.ManagedThreadId}'");
```
duncanp-sonar: Miscellaneous: extra debug information. It helped show a race condition that can occur while connecting (#696)

src/SonarQube.Client/Api/SonarQubeService.cs
```diff
@@ -252,5 +253,32 @@ public Uri GetProjectDashboardUrl(string projectKey)
...
+        }
+
+        // This code added to correctly implement the disposable pattern.
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in Dispose(bool disposing) above.
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+        #endregion
```
Evangelink: [minor] I usually like to put a comment close to the `endregion` to state what this is closing (e.g. `#endregion // IDisposable Support`).

Also I don't think we have regions globally in the code base of SLVS so I would remove it.

PR #698, opened 2018-07-25
Fixed #688 - pass organization key when requesting QP changelog

### LINKED ISSUE #688: [Client] api/qualityprofiles/changelog needs to provide organisation when fetching custom profile on SonarCloud
### Description

The call to `https://sonarcloud.io/api/qualityprofiles/changelog?qualityProfile=CustomProfile1&language=cs&p=1&ps=1` needs to provide the organisation parameter on SonarCloud.

We should check whether any the other APIs also need to be updated.

Note: the API call works ok when retrieve the Sonar Way profile by name, but for any other profile the organisation is required.

### Repro steps

1. Open a project that is bound to SonarCloud
2. Change the project profile on SonarCloud to be a custom profile
3. Open the SonarQube tab in the Team Explorer and update the project

### Expected behavior

The rulesets are updated

### Actual behavior

The update fails with a 404 error.

### Known workarounds

None

duncanp-sonar:


src/SonarQube.Client.Tests/Api/SonarQubeService_GetQualityProfileAsync.cs
```diff
@@ -323,7 +323,7 @@ public class SonarQubeService_GetQualityProfileAsync : SonarQubeService_TestBase
   }
 }");
 
-            SetupRequest("api/qualityprofiles/changelog?qualityProfile=Sonar+way&language=cs&p=1&ps=1", @"{
+            SetupRequest("api/qualityprofiles/changelog?qualityProfile=Sonar+way&language=cs&organization=my_organization&p=1&ps=1", @"{
```
duncanp-sonar: There is an existing test for pre-6.5 that checks that the organization parameter is not passed (although it wouldn't really matter in practice if it was since it would be ignored).

src/SonarQube.Client/Api/Requests/V5_20/GetQualityProfileChangeLogRequest.cs
```diff
@@ -53,6 +53,17 @@ public virtual string LanguageName
             set { /* not supported in this implementation */ }
         }
 
+        /// <summary>
+        /// This property does not exist before SonarQube 6.4 and is deliberately implemented to always return
+        /// the default value in order to prevent it from serialization.
+        /// </summary>
+        [JsonProperty("organization", DefaultValueHandling = DefaultValueHandling.Ignore), DefaultValue(null)]
```
duncanp-sonar: The new parameter has to be added to the interface (i.e. the interface is a superset of all the possible parameters for every version).
This is the v5_2 base class, which is the version in which the API was introduced.
The class for the later version of the API (6.5) inherits from this class and overrides the property so it can be set and serialized.

duncanp-sonar:
I've manually tested that the SLVS can retrieve the default and custom QPs in SonarCloud, and in SonarQube 6.5. I haven't manually tested older versions of SQ - I'm relying on the existing unit test for that.

PR #700, opened 2018-07-26
Fixed #699 - suppressions not being fetched in SQ7.2

### LINKED ISSUE #699: [7.2RC] Suppressed issues are not fetched from SonarCloud
### Description

An exception is thrown parsing the response from the server. The exception is suppressed so there is no indication that fetching the suppressions has failed.

### Repro steps

1. Bind an MSBuild project to a project on SonarCloud that has an Open issue (i.e. the resolution is empty)

### Expected behavior

The list of expected issues is retrieved, or an exception is recorded in the SonarLint output window.

### Actual behavior

The call fails silently because the exception thrown during parsing is suppressed.

### Known workarounds

None

### Related information

* SonarLint for VS Version: 7.2 RC i.e. currently unreleased version so will not currently be affecting users.
* SonarQube server version: unconfirmed, but almost certainly an issue with SQ v7.2 too.

### Background

The SonarClient calls a different API to fetch issues when connected to v7.2,
The old API (_batch/issues_) used protobuf to send the results.
The new API (_api/issues/search_) uses JSON.
The difference appears to be in how missing _resolution_ values are handled. Protobuf returns empty but the JSON deserializer returns null, which is not handled.

duncanp-sonar:
Miscellaneous: adding logging output to the suppression manager

### DIFF
--- a/src/Integration.UnitTests/Notifications/SonarQubeNotificationServiceTests.cs
+++ b/src/Integration.UnitTests/Notifications/SonarQubeNotificationServiceTests.cs
@@ -110,7 +110,7 @@ namespace SonarLint.VisualStudio.Integration.Notifications.UnitTests
             await notifications.StartAsync("should-throw", null);
 
             // Assert
-            outputMock.Verify(x => x.WriteLine($"Failed to fetch notifications: test exception"), Times.Once);
+            outputMock.Verify(x => x.WriteLine("Failed to fetch notifications: {0}", "test exception"), Times.Once);
         }
 
         [TestMethod]
--- a/src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
+++ b/src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
@@ -38,6 +38,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
         private Mock<ISonarQubeService> mockSqService;
         private Mock<ITimerFactory> mockTimerFactory;
         private Mock<ITimer> mockTimer;
+        private TestLogger testLogger;
 
         /// <summary>
         /// Wait handle that is set to signalled when the initial fetch task has started
@@ -53,7 +54,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
         public void TestInitialize()
         {
             mockSqService = new Mock<ISonarQubeService>();
-            
+
+            testLogger = new TestLogger();
+
             mockTimerFactory = new Mock<ITimerFactory>();
             mockTimer = new Mock<ITimer>();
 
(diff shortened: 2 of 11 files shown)


src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -38,6 +38,7 @@ public class SonarQubeIssueProviderTests
         private Mock<ISonarQubeService> mockSqService;
         private Mock<ITimerFactory> mockTimerFactory;
         private Mock<ITimer> mockTimer;
+        private TestLogger testLogger;
```
duncanp-sonar: I used the test logger rather than a mock because it's easier to check a series of expected messages.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -417,6 +440,7 @@ public void GetIssues_ErrorFetchingOnTimerElapsed_IsSuppressed()
             fetchCallCount.Should().Be(1);
 
             VerifyTimerStart(Times.Exactly(1)); // once, on construction
+            testLogger.AssertPartialOutputStrings("Checking for suppressions", "dummy error from mock");
```
duncanp-sonar: New assertion to check that the error message is logged.

src/Integration/Notifications/SonarQubeNotificationService.cs
```diff
@@ -119,7 +119,7 @@ public void Stop()
             }
             catch (Exception ex)
             {
-                logger.WriteLine($"Failed to fetch notifications: {ex.Message}");
+                logger.WriteLine(Strings.Notifications_ERROR_Fetching, ex.Message);
```
duncanp-sonar: Miscellaneous change: changing hard-coded string to a resource

src/Integration_2017/Integration.csproj
```diff
@@ -223,7 +223,7 @@
       <Generator>MSBuild:Compile</Generator>
       <Link>%(RecursiveDir)%(Filename)%(Extension)</Link>
     </Page>
-    <Compile Include="$(ReferencedDirectory)\Resources\Strings.Designer.cs">
+    <Compile Include="..\Integration\Resources\Strings.Designer.cs">
```
duncanp-sonar: Oops. Didn't mean to check this file in.

src/SonarQube.Client/Api/Requests/V7_20/GetIssuesRequest.cs
```diff
@@ -61,7 +61,10 @@ private static SonarQubeIssueResolutionState ParseResolutionState(string resolut
         {
             switch (resolution)
             {
-                case "":
+                // Issues with Status=OPEN will not have a resolution so we're
+                // expecting null, but we'll be defensive and handle empty too
+                case null:
```
duncanp-sonar: This was the bug - the JSON deserializer used with the non-deprecated API handles missing values differently from the protobuf deserializer.

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -96,7 +96,6 @@ public class SonarQubeService_GetSuppressedIssuesAsync : SonarQubeService_TestBa
         ""endOffset"": 107
     },
     ""flows"": [],
-    ""resolution"": """",
```
duncanp-sonar: I called the API on SonarCloud and the data returned didn't have a _resolution_ field if the status was _Open_
i.e. looks like this test JSON was wrong.

src/Integration.UnitTests/Notifications/SonarQubeNotificationServiceTests.cs
```diff
@@ -110,7 +110,7 @@ public void StartAsync_Throws_ArgumentNullException_For_ProjectKey()
             await notifications.StartAsync("should-throw", null);
 
             // Assert
-            outputMock.Verify(x => x.WriteLine($"Failed to fetch notifications: test exception"), Times.Once);
+            outputMock.Verify(x => x.WriteLine("Failed to fetch notifications: {0}", "test exception"), Times.Once);
```
Evangelink: Is this change just about readability or was there any problem?

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
Evangelink: I don't mind using test classes :)

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
Evangelink: What about a SonarQube instance like Next?

src/TestInfrastructure/Framework/TestLogger.cs
```diff
@@ -40,6 +40,22 @@ public void AssertOutputStrings(params string[] orderedOutputMessages)
...
 
+        public void AssertPartialOutputStrings(params string[] orderedPartialOutputMessages)
+        {
+            this.outputStrings.Should().Equal(orderedPartialOutputMessages, (actualValue, expectedValue) =>
+                actualValue.Contains(expectedValue));
+        }
+
+        public void AssertNoOutputMessages()
+        {
+            outputStrings.Count().Should().Be(0);
```
Evangelink: You could use `Should().HaveCount(0)` instead.

src/TestInfrastructure/Framework/TestLogger.cs
```diff
@@ -40,6 +40,22 @@ public void AssertOutputStrings(params string[] orderedOutputMessages)
...
+        }
+
+        public void Reset()
+        {
+            outputStrings = new List<string>();
```
Evangelink: [Minor] I'd rather have `outputStrings` readonly and call `Clear`.

src/Integration.UnitTests/Notifications/SonarQubeNotificationServiceTests.cs
duncanp-sonar: It was required since this test is using a mock. Previously the product code was calling _WriteLine(string)_, now it's calling _WriteLine(string, object[] args)_, so the test needs to verify a different overload.

duncanp-sonar:
_Next_ gives the same response i.e. no _resolution_ field if _status_ is _OPEN_

PR #703, opened 2018-08-02
Fixed #692: added SonarCloud to telemetry

### LINKED ISSUE #692: Add SonarCloud flag in telemetry
Telemetry data should show if SonarLint is connected to SonarCloud.
The parameter is: `connected_mode_sonarcloud.`
The value is a boolean: `true/false`

For example:
> "connected_mode_used": true
"connected_mode_sonarcloud": true

Obviously, the parameter can be true only when the connected mode is used.

duncanp-sonar:


src/Integration/Telemetry/TelemetryManager.cs
```diff
@@ -140,7 +140,7 @@ private void EnableAllEvents()
         private TelemetryPayload GetPayload(TelemetryData telemetryData)
         {
             return TelemetryHelper.CreatePayload(telemetryData, DateTimeOffset.Now,
-                solutionBindingTracker.CurrentConfiguration.Mode != NewConnectedMode.SonarLintMode.Standalone);
```
duncanp-sonar: I changed the _CreatePayload_ method to take the binding as a parameter, otherwise the caller would have to work out _isConnected/isSonarCloud_.

src/Integration/Telemetry/TelemetryHelper.cs
```diff
@@ -32,8 +33,25 @@ private static string GetSonarLintVersion()
             return FileVersionInfo.GetVersionInfo(typeof(TelemetryTimer).Assembly.Location).FileVersion;
         }
 
-        public static TelemetryPayload CreatePayload(TelemetryData telemetryData, DateTimeOffset now, bool isConnected)
+        public static bool IsSonarCloud(Uri sonarqubeUri) =>
+            sonarqubeUri?.ToString().Equals("https://sonarcloud.io/", StringComparison.OrdinalIgnoreCase) ?? false;
```
valhristov: Could this be `https://www.sonarcloud.io`?

src/Integration.UnitTests/Telemetry/TelemetryHelper_CreatePayload.cs
```diff
@@ -22,14 +22,26 @@
...
 {
     [TestClass]
     public class TelemetryHelper_CreatePayload
     {
         [TestMethod]
-        public void CreatePayload_Creates_Payload()
+        public void CreatePayload_InvalidArg_Throws()
+        {
+            Action action = () => TelemetryHelper.CreatePayload(null, DateTimeOffset.Now, BindingConfiguration.Standalone);
+            action.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("telemetryData");
```
Evangelink: It's usually better to use `ThowExactly` IMO.

src/Integration.UnitTests/Telemetry/TelemetryHelper_CreatePayload.cs
```diff
@@ -22,14 +22,26 @@
...
+
+            action = () => TelemetryHelper.CreatePayload(new TelemetryData(), DateTimeOffset.Now, null);
+            action.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("bindingConfiguration");
```
Evangelink: Same here (`ThrowExactly`).

src/Integration/Telemetry/TelemetryHelper.cs
Evangelink: Is there no better way than comparing the `ToString` of an URI?

duncanp-sonar: Good catch. The browser will redirect to _https://sonarcloud.io_, but if the user types in _https://www.sonarcloud.io_ then that's what we'll store in the file.

Evangelink:
@duncanp-sonar What happened?

duncanp-sonar:
@Evangelink an unrelated time-dependent test failed (one that expected the result to be _x_ plus or minus _y_ milliseconds)

PR #706, opened 2018-08-03
Fixed #686: connection information is ignored

### LINKED ISSUE #686: Connection information is ignored
### Description

If a wrong URL is entered when trying to connect a new application, after the URL is corrected, the old one is still used.

### Repro steps

1. Create new application
2. Open SonarQube pane in Team Explorer, click "Connect..."
3. Enter incorrect URL, such as "http://asdasdasd"
4. Enter user/pass
5. Click Connect, the connection should fail
6. Repeat the process with a correct URL
7. The connection will use the old, wrong URL until VS is restarted.

### Expected behavior

The entered URL after failure should be taken into account.

### Known workarounds

Restart VS

duncanp-sonar:


src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -359,10 +359,12 @@ private string GetPluginProjectMismatchErrorMessage(ICollection<EnvDTE.Project>
             return isPluginSupported;
         }
 
-        private static void AbortWorkflow(IProgressController controller, CancellationToken token)
+        private void AbortWorkflow(IProgressController controller, CancellationToken token)
         {
             bool aborted = controller.TryAbort();
             Debug.Assert(aborted || token.IsCancellationRequested, "Failed to abort the workflow");
+
+            this.host.SonarQubeService.Disconnect();
```
duncanp-sonar: If the first connection attempt failed (e.g. because the user entered the wrong URL or credentials), the SonarQubeService was still marked as being connected. Then, if the user tried to reconnect, the connection workflow would not attempt to connect again using the new url/credentials etc.

The simple fix is just to call disconnect whenever the workflow is aborted.

This bug could have been fixed in a different way i.e. by not marking the service as being connected unless the connection is successful. However, that is a more complicated fix, and will probably be a side effect of fixing #696.

PR #707, opened 2018-08-03
Fixed #642: double-click to bind

### LINKED ISSUE #642: New binding: double-clicking on the project in the list does nothing
This is a regression. Previously, double-clicking on the project would bind to that project.
Now, it is only possible to bind by using the context menu for the item.

### Related information

* SonarLint for VS Version: v4.0 - for both SLVS2015 and SLVS2017

duncanp-sonar:
This ticket wasn't originally part of the sprint, but I've added it since it was an easy fix and the bug was annoying me (I was binding to a lot of projects as part of testing the other fixes).

### DIFF
--- a/src/Integration.UnitTests/WPF/ProjectViewModelToBindingArgsConverterTests.cs
+++ b/src/Integration.UnitTests/WPF/ProjectViewModelToBindingArgsConverterTests.cs
@@ -1,0 +1,74 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2018 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Globalization;
+using System.Security;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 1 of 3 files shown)


src/Integration/TeamExplorer/ConnectSectionView.xaml
```diff
@@ -138,7 +139,7 @@
                             <Grid.InputBindings>
                                 <MouseBinding Command="{Binding Path=DataContext.BindCommand,
                                                                 RelativeSource={RelativeSource AncestorType=UserControl}}"
-                                              CommandParameter="{Binding .}"
+                                              CommandParameter="{Binding ., Converter={StaticResource ProjectViewModelToBindingArgsConverter}}"
```
duncanp-sonar: Some time ago I refactored the binding command to use strongly-typed arguments to make the binding code easier to understand. However, I failed to update this command binding.

The command now expects a parameter of type _BindCommandArgs_, but here the command is bound to a _ProjectViewModel_. Fortunately, the project view model contains all of the information needed to create a _BindCommandArgs_, so adding a simple converter is enough.

src/Integration.UnitTests/WPF/ProjectViewModelToBindingArgsConverterTests.cs
```diff
@@ -0,0 +1,74 @@
...
+    public class ProjectViewModelToBindingArgsConverterTests
+    {
+        [TestMethod]
+        public void Convert_NotProjectViewModel_ReturnsNull()
+        {
+            // Arrange
+            var converter = new ProjectViewModelToBindingArgsConverter();
+
+            // Act && Assert
+            converter.Convert(null, typeof(object), null, CultureInfo.CurrentCulture).Should().BeNull();
```
Evangelink: You are not using the target type nor the culture so you could pass `null` for both.

src/Integration.UnitTests/WPF/ProjectViewModelToBindingArgsConverterTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            var expectedPassword = new SecureString();
+            expectedPassword.AppendChar('x');
+            var serverViewModel = new ServerViewModel(new ConnectionInformation(expectedUri, "user1", expectedPassword));
+            var project = new SonarQubeProject("key1", "name1");
+            var projectViewModel = new ProjectViewModel(serverViewModel, project);
+
+            var converter = new ProjectViewModelToBindingArgsConverter();
+
+            // Act
+            var convertedObj = converter.Convert(projectViewModel, typeof(object), null, System.Globalization.CultureInfo.CurrentCulture);
```
Evangelink: Same here, no need for `typeof(object)` and `System.Globalization.CultureInfo.CurrentCulture`

PR #717, opened 2018-09-03
Use Jenkinsfile library v2 to free some node usage

drautureau-sonarsource:
The aim of the new version of the library is to send any notifications to BURGR, GitHub and Repox without requiring a slave.
With this, we can then free up the usage of some slaves.


Evangelink:
@drautureau-sonarsource What should we do about this PR?

drautureau-sonarsource:
@Evangelink, it is just to let you know of this change (description in 1st comment).
I can merge if it's OK for you

PR #745, opened 2018-10-16
 Refactor suppression matching algorithm not to rely on sonarqube modules

### LINKED ISSUE #574: Resolve as Won't Fix Issues not always syncing with Sonar Lint Connected Mode
### Description

This is similar to another issue I opened last year, however we are on later versions of Sonar, Sonar C# Analzyser now and the latest SonarLint. We have moved our sonar instance to a central one for the whole company (50+ teams), this instance is ultimately more up to date than the one we used for our office's five teams. The issue was resolved for the most part except I did not notice some problems with Won't Fix resolutions still being reported in Visual Studio at some times. However the problem now seems to be more persistent.

The problem reported last year was : [here](https://github.com/SonarSource/sonarlint-visualstudio/issues/445)

We have three code smells reported in Sonar Qube, two are genuine tech debt which we intend to resolve and a third which we marked as Resolved Won't Fix. When the analysis runs on the server only the two that we want reported are reported. In our IDE, when we perform a rebuild  the issue resolved as Won't Fix is reported in the Visual Studio error list. If I then open the file in question that warning is no longer reported unless I close the file and Rebuild again. Sometimes this does not occur, sometimes once the file has been opened any subsequent rebuilds do not cause the issue to re-appear but if we close the solution and re-open and build the issue always returns. The only way to get the issue to disappear again is to open the file.

Obviously we want to keep our IDE error list as clean as possible in terms of errors/warnings in the error list so to have the issues reported is adding extra noise, thankfully we only have this one issues marked as Won't Fix but if we had many more this would be a much bigger problem.

### Repro steps

1. Write code that violates a Sonar Rule to invoke an analysis build warning.

2. Execute Sonar Qube analysis to see the violation recorded as code smell in Sonar Qube.

3. Mark as Resolved Won't Fix.

4. Run analysis again to see the code smell no longer reported in Sonar Qube.

5. Perform a build/rebuild in Visual Studio. Issue still reported as build warning.

6. Open offending file, the issue is no longer reported.

7. Close file (and sometimes) close solution.

8. Perform a Rebuild and the issue is reported again.

9. Open offending file, the issue is no longer reported.

Note : Step 7 is intermittent, sometimes just closing the file causes the issue to be reported again, however closing and re-oping the solution always causes the issue to be reported again.

### Expected behavior

Resolved Marked as Won't Fix issues in Sonar Qube are no longer reported by SonarAnalyzer as long as the sync from SonarLint to the SonarQube instance has occurred which I according to "duncanp-sonar" in my previous reported issue occurs every ten minutes or following opening the solution.

### Actual behavior

Issues Marked as Won't Fix are reported in the IDE following a Rebuild unless the file is opened in the IDE.

### Known workarounds

Open the file.

### Related information

* SonarLint for VS Version 3.10.0.3095
* Visual Studio Version Enterprise 15.2 (26430.13)
* SonarQube server version Version 6.7 (build 33306)
* Languages in your solution - Just C#
* Sonar C# Analyzer Version 6.7.1

### LINKED ISSUE #739: Update SonarLint to handle dropping of support for modules
### Description

The concept of modules is being dropped from the Sonar ecosystem. See https://jira.sonarsource.com/browse/MMF-365 for more information. SLVS needs to be updated to handle the absence of modules in the future.

### Detail
The only place SLVS uses module information is in handling suppressions in connected mode. SLVS retrieves a list of issues from the server and attempts to match issues raised by the analyzers with an issue returned from the server.

The issue key returned by the server is currently in the following form:
`root_project_key:root_project_key:module_key:relative_file_path_from_module`
where the _module_key_ is the _ProjectGuid_ for the MSBuild project.

In future, the issue key will be in the following form:
`root_project_key:relative_file_path_from_root_project`

SLVS must handle issue keys in both old and new, partly to support users on older versions of SonarQube, and also because the projects on the new versions of SonarQube will still use the old issue key format until the project is analyzed for the first time following migration to the new version of SonarQube.

### LINKED ISSUE #741: Suppressed issues for files attached to root module doesn't work
### Description

Currently we don't retrieve suppressed issues for files attached to the root module.

Evangelink:
Fix #739, fix #741 and probably fix #574


src/Integration.Vsix/Suppression/LiveIssue.cs
```diff
@@ -33,7 +34,9 @@ public class LiveIssue
             string wholeLineText = "")
         {
             Diagnostic = diagnostic;
-            IssueFilePath = issueFilePath;
+            IssueFilePath = issueFilePath != string.Empty
```
Evangelink: I couldn't find cases where the file path was null but maybe I should try to be defensive here, WDYT?

src/Integration.Vsix/Suppression/LiveIssueFactory.cs
```diff
@@ -98,7 +97,6 @@ public LiveIssue Create(SyntaxTree syntaxTree, Diagnostic diagnostic)
             }
 
             var lineSpan = diagnostic.Location.GetLineSpan();
-            var relativeFilePath = FileUtilities.GetRelativePath(projectFilePath, lineSpan.Path);
```
Evangelink: We no longer want a path relative to the csproj.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -41,7 +42,7 @@ public sealed class SonarQubeIssuesProvider : ISonarQubeIssuesProvider, IDisposa
         private readonly ILogger logger;
         private readonly CancellationTokenSource initialFetchCancellationTokenSource;
 
-        private IList<SonarQubeIssue> cachedSuppressedIssues;
+        private List<IGrouping<string, SonarQubeIssue>> cachedSuppressedIssues;
```
Evangelink: I haven't seen any performance issue with this data-structure but it might be worth doing some manual test on a big project like Roslyn.

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -214,12 +214,8 @@ public async Task GetSuppressedIssuesAsync_From_7_20_ModuleLevelIssue()
   ""issues"": [{
     ""key"": ""AWQYvwc4pbnviuOCCX4g"",
     ""rule"": ""csharpsquid:S1116"",
-    ""severity"": ""MINOR"",
```
Evangelink: Removing extra data we don't use for this test.

src/SonarQube.Client/Api/Requests/V7_20/GetIssuesRequest.cs
```diff
@@ -86,22 +86,20 @@ private static SonarQubeIssueResolutionState ParseResolutionState(string resolut
 
         private class ServerIssue
         {
-            [JsonProperty("key")]
-            public string Key { get; set; }
```
Evangelink: This was unused so I have removed it.

src/SonarQube.Client/Api/Requests/V7_20/GetIssuesRequest.cs
```diff
@@ -86,22 +86,20 @@ private static SonarQubeIssueResolutionState ParseResolutionState(string resolut
...
+            [JsonProperty("rule")]
+            public string CompositeRuleKey { get; set; }
```
Evangelink: Re-ordered to match the json response order.

src/SonarQube.Client/Api/Requests/V7_20/GetIssuesRequest.cs
```diff
@@ -86,22 +86,20 @@ private static SonarQubeIssueResolutionState ParseResolutionState(string resolut
...
             [JsonProperty("component")]
             public string Component { get; set; }
+            [JsonProperty("subProject")]
+            public string SubProject { get; set; }
```
Evangelink: Same here.

src/SonarQube.Client/Api/Requests/V7_20/GetIssuesRequest.cs
```diff
@@ -86,22 +86,20 @@ private static SonarQubeIssueResolutionState ParseResolutionState(string resolut
...
             [JsonProperty("hash")]
             public string Hash { get; set; }
             [JsonProperty("line")]
-            public int Line { get; set; }
+            public int? Line { get; set; }
```
Evangelink: Changed into nullable because module level issues don't have a line (before it was defaulting to default(int)).

src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllModulesAsync.cs
```diff
@@ -0,0 +1,155 @@
...
+            result[0].Name.Should().Be("SonarQube Project Name");
+            result[0].RelativePathToRoot.Should().BeNull();
+
+            result[1].Key.Should().Be("sq-project-key:sq-project-key:50934C7A-2751-4675-91C4-CFD37C8BF57C");
+            result[1].Name.Should().Be("Project1");
+            result[1].RelativePathToRoot.Should().BeNull();            
+        }
+
+        [TestMethod]
+        public async Task GetModules_ExpectedNewSonarQubeApi()
```
duncanp-sonar: This test name is a bit confusing. What is the test checking that the previous two don't - that the API handles a response that only has a root project?

src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllModulesAsync.cs
```diff
@@ -0,0 +1,155 @@
...
+            messageHandler.VerifyAll();
+
+            result.Should().HaveCount(1);
+
+            result[0].Key.Should().Be("sq-project-key");
+            result[0].Name.Should().Be("SonarQube Project Name");
+            result[0].RelativePathToRoot.Should().BeNull();
+        }
+    }
+}
```
duncanp-sonar: Additional test case - response if project with the specified key is not found ( -> empty list, or null?)

src/SonarQube.Client/Api/Requests/V5_40/GetModulesRequest.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    public class GetModulesRequest : PagedRequestBase<SonarQubeModule>, IGetModulesRequest
+    {
+        protected override string Path => "api/components/tree";
+
+        [JsonProperty("qualifiers")]
+        public string Qualifiers
+        {
+            get { return "BRC"; } // Sub-projects
+            set { /* not supported in this implementation */ }
```
duncanp-sonar: Is our code ever expected to call the _set_ method? If not, maybe it would be better to throw to make it easy to detect if we do ever call it.

src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllModulesAsync.cs
Evangelink: According to Duarte, when a project will no longer have modules (only the root) we expect the API to return the following json (`baseComponent` filled and `components` empty).

src/SonarQube.Client/Api/Requests/V5_40/GetModulesRequest.cs
Evangelink: If the _ever_ means does the current codebase calls this setter, the answer is no. If the _ever_ means could we end up having to use this functionality later on, then the answer is maybe.

I'll keep things simple and throw for now.

src/Integration.Vsix/Suppression/LiveIssue.cs
duncanp-sonar: Hhhmm. It should be clear to someone reading the code later whether nulls are allowed or not, so there are three options: assert, throw, or be defensive and handle nulls. The simplest is probably to be defensive and use _IsNullOrEmpty_.

src/Integration.Vsix/Suppression/LiveIssueFactory.cs
duncanp-sonar: It looks like this was the only place that used _GetRelativePath_ so it could be deleted.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -106,15 +107,11 @@ public IEnumerable<SonarQubeIssue> GetSuppressedIssues(string projectGuid, strin
...
-                x.FilePath.Equals(filePath, StringComparison.OrdinalIgnoreCase) &&
-                x.ModuleKey.Equals(moduleKey, StringComparison.OrdinalIgnoreCase));
-        }
-
-        private string BuildModuleKey(string projectGuid)
-        {
-            return $"{sonarQubeProjectKey}:{sonarQubeProjectKey}:{projectGuid}";
+            string moduleKey = $"{sonarQubeProjectKey}:{sonarQubeProjectKey}:{projectGuid}";
+            return this.cachedSuppressedIssues
+                .Where(x => filePath.EndsWith(x.Key, StringComparison.OrdinalIgnoreCase)
```
duncanp-sonar: If we're only expecting one match, couldn't we use _FirstOrDefault_ rather than _Where_? (it would make the expectation clearer and would also be slightly faster in cases where a match is found).

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -106,15 +107,11 @@ public IEnumerable<SonarQubeIssue> GetSuppressedIssues(string projectGuid, strin
...
-            string moduleKey = BuildModuleKey(projectGuid);
-            return this.cachedSuppressedIssues.Where(x =>
-                x.FilePath.Equals(filePath, StringComparison.OrdinalIgnoreCase) &&
-                x.ModuleKey.Equals(moduleKey, StringComparison.OrdinalIgnoreCase));
-        }
-
-        private string BuildModuleKey(string projectGuid)
-        {
-            return $"{sonarQubeProjectKey}:{sonarQubeProjectKey}:{projectGuid}";
+            string moduleKey = $"{sonarQubeProjectKey}:{sonarQubeProjectKey}:{projectGuid}";
```
duncanp-sonar: This is the format of the legacy module key, so we need to match against this for old versions of SQ and newer versions where the module keys haven't been migrated to the new format.
However, I thought that the module key in the new module-less world will just be the project key for the root project. Don't we need to match against that value too?

src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllModulesAsync.cs
duncanp-sonar: The project could be deleted from the server by an admin after we've opened the bound solution. Not likely, but possible so we shouldn't crash if it does.

src/Integration.Vsix/Suppression/LiveIssue.cs
```diff
@@ -34,15 +34,17 @@ public class LiveIssue
...
+            IssueFilePath = !string.IsNullOrEmpty(issueFilePath)
                 ? Path.GetFullPath(issueFilePath)
                 : string.Empty; 
             ProjectGuid = projectGuid;
             StartLine = startLine;
             WholeLineText = wholeLineText;
 
             // SonarQube doesn't calculate hash for file-level issues
-            LineHash = wholeLineText != "" ? ChecksumCalculator.Calculate(WholeLineText) : "";
+            LineHash = wholeLineText != string.Empty
```
duncanp-sonar: Not something you've changed in this PR, but perhaps this logic should be inside the _ChecksumCalculator_.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -156,11 +171,12 @@ private async Task SynchronizeSuppressedIssuesAsync()
                 var moduleKeyToRelativePathToRoot = (await this.sonarQubeService.GetAllModulesAsync(sonarQubeProjectKey,
                         cancellationTokenSource.Token))
                     .ToDictionary(x => x.Key, x => x.RelativePathToRoot ?? string.Empty);
+                this.hasModules = moduleKeyToRelativePathToRoot.Keys.Count == 1;
```
duncanp-sonar: Isn't it the case that if there is only one module then that must be the root module -> _hasModules = false_?

If the code as-is is incorrect then there must a gap in the test coverage.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -156,11 +171,12 @@ private async Task SynchronizeSuppressedIssuesAsync()
...
 
                 this.cachedSuppressedIssues = (await this.sonarQubeService.GetSuppressedIssuesAsync(sonarQubeProjectKey,
                         cancellationTokenSource.Token))
-                    .Select(x => ToIssueWithFixedRelativePath(moduleKeyToRelativePathToRoot, x))
-                    .GroupBy(x => x.FilePath)
+                    .Select(x => new { Key = ProcessKey(moduleKeyToRelativePathToRoot, x), Issue = x })
+                    .GroupBy(x => x.Key, x => x.Issue)
                     .OrderBy(x => x.Key.Length)
```
duncanp-sonar: Don't we want to sort by key length in descending order? (default sort order is ascending)

If so, then there's a test hole: we need to test having the same file in different directories.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -156,11 +171,12 @@ private async Task SynchronizeSuppressedIssuesAsync()
                 var moduleKeyToRelativePathToRoot = (await this.sonarQubeService.GetAllModulesAsync(sonarQubeProjectKey,
                         cancellationTokenSource.Token))
                     .ToDictionary(x => x.Key, x => x.RelativePathToRoot ?? string.Empty);
+                this.hasModules = moduleKeyToRelativePathToRoot.Keys.Count == 1;
 
                 this.cachedSuppressedIssues = (await this.sonarQubeService.GetSuppressedIssuesAsync(sonarQubeProjectKey,
                         cancellationTokenSource.Token))
-                    .Select(x => ToIssueWithFixedRelativePath(moduleKeyToRelativePathToRoot, x))
-                    .GroupBy(x => x.FilePath)
+                    .Select(x => new { Key = ProcessKey(moduleKeyToRelativePathToRoot, x), Issue = x })
```
duncanp-sonar: If I understand correctly, the keys of items in the cache will be a mix of file-based keys and non-file based keys. When we're doing a lookup into the list we're walking the keys doing two comparisons per key, one to see if it's a file-based key and a second if it isn't.
It would be more efficient if we maintained separate lists of file-based and non-file based keys and then when looking for a match we only seached one list - we'd be searching through fewer items, and only doing one comparison per item.
Even if this change is worth making, I think it should be done in a separate PR i.e. stabilise and test the work in progress, then worry about optimisation.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -170,17 +184,23 @@ private async Task SynchronizeSuppressedIssuesAsync()
                 // TODO: Handle race conditions
                 var moduleKeyToRelativePathToRoot = (await this.sonarQubeService.GetAllModulesAsync(sonarQubeProjectKey,
                         cancellationTokenSource.Token))
-                    .ToDictionary(x => x.Key, x => x.RelativePathToRoot ?? string.Empty);
-                this.hasModules = moduleKeyToRelativePathToRoot.Keys.Count == 1;
+                    .ToDictionary(x => x.Key, x => x.RelativePathToRoot?.Trim('\\', '/').Replace('/', '\\') ?? string.Empty);
```
duncanp-sonar: The _Trim_ and _Replace_ is repeated on line 226 below. It would be better to have a _NormalizePath_ function that handled the null checking and transformation.

src/SonarQube.Client.Tests/Api/SonarQubeService_GetAllModulesAsync.cs
Evangelink: We will crash earlier in the code as it's not handled by the other methods.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -205,95 +205,250 @@ public void GetIssues_NoIssuesOnServer_ReturnsEmptyList()
...
-            matches.Should().BeEmpty();
-            VerifyServiceGetIssues(Times.Exactly(1), "sqKey"); // cached issues should be used
+            // Act
+            var matches = issuesProvider.GetSuppressedIssues("projectId", null);
 
-            // 2. File id doesn't match -> no issues
-            matches = issuesProvider.GetSuppressedIssues("projectID1", "folder1/filexxx");
-            matches.Should().BeEmpty();
+            // Assert
+            matches.Should().HaveCount(1);
```
duncanp-sonar: As discussed, this test doesn't prove that the product code is doing any filtering - there would need to be at least one other issue that is returned by the mocked server to test that filtering is happening.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -205,95 +205,250 @@ public void GetIssues_NoIssuesOnServer_ReturnsEmptyList()
...
+            var issuesProvider = new SonarQubeIssuesProvider(mockSqService.Object, "sqkey", mockTimerFactory.Object, testLogger);
+            WaitForInitialFetchTaskToStart();
+
+            VerifyServiceGetIssues(Times.Exactly(1)); // issues should be fetched on creation
+
+            // Act
+            var matches = issuesProvider.GetSuppressedIssues("projectId", null);
+
+            // Assert
+            matches.Should().HaveCount(1);
```
duncanp-sonar: Same: the mocked list of issues should contain an issue that should not be returned.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -205,95 +205,250 @@ public void GetIssues_NoIssuesOnServer_ReturnsEmptyList()
...
+        public void GetSuppressedIssues_WhenProjectHasModules_WhenIssueIsFileLevelAndFound_ReturnsExpectedIssue()
+        {
+            // Arrange
+            var sonarQubeIssue1 = new SonarQubeIssue("\\foo\\foo.cs", null, null, "message", "sqkey:sqkey:projectId",
+                SonarQubeIssueResolutionState.FalsePositive, "S1");
+            var sonarQubeIssue2 = new SonarQubeIssue("/foo/foo.cs", null, null, "message", "sqkey:sqkey:projectId",
+                SonarQubeIssueResolutionState.FalsePositive, "S2");
+            var sonarQubeIssue3 = new SonarQubeIssue("foo\\foo.cs", null, null, "message", "sqkey:sqkey:projectId",
+                SonarQubeIssueResolutionState.FalsePositive, "S3");
+            var sonarQubeIssue4 = new SonarQubeIssue("foo/foo.cs", null, null, "message", "sqkey:sqkey:projectId",
```
duncanp-sonar: Add an issue that won't be returned to check that filtering is happening.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -205,95 +205,250 @@ public void GetIssues_NoIssuesOnServer_ReturnsEmptyList()
...
+                new List<SonarQubeIssue> { sonarQubeIssue1, sonarQubeIssue2, sonarQubeIssue3, sonarQubeIssue4 },
+                new List<SonarQubeModule> { new SonarQubeModule("sqkey", "", ""), new SonarQubeModule("sqkey:sqkey:projectId", "", "src/bar") });
+
+            var issuesProvider = new SonarQubeIssuesProvider(mockSqService.Object, "sqkey", mockTimerFactory.Object, testLogger);
+            WaitForInitialFetchTaskToStart();
+
+            VerifyServiceGetIssues(Times.Exactly(1)); // issues should be fetched on creation
+
+            // Act
+            var matches = issuesProvider.GetSuppressedIssues("", "C:\\AwesomeProject\\src\\bar\\foo\\foo.cs");
```
duncanp-sonar: Case-insensitivity - maybe pass something like _C:\\AwesomeProject\\src\\bar\\Foo\\fOO.cs_

duncanp-sonar:
FYI I created a [new PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/749) with a re-written history: I kept the first "Add web service commit" and squashed the rest into a "Refactor suppression matching" commit (I know there were minor tweaks to the web service later but didn't think it was worth trying to separating them out completely).
Closing this PR.

PR #750, opened 2018-10-22
Update the way we retrieve issues to suppress

Evangelink:


src/Integration.Vsix.UnitTests/Suppression/SuppressionHandlerTests.cs
```diff
@@ -20,7 +20,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
```
Evangelink: Automatic change.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -705,7 +674,7 @@ private void SetupSolutionBinding(bool isConnected, IList<SonarQubeIssue> issues
             InitialFetchWaitHandle = new EventWaitHandle(false, EventResetMode.ManualReset);
 
             mockSqService.Setup(x => x.GetAllModulesAsync(It.IsAny<string>(), It.IsAny<CancellationToken>()))
-                .ReturnsAsync(modules
```
Evangelink: Automatic removal of trailing spaces.

src/SonarQube.Client/Api/Requests/V5_10/GetIssuesRequest.cs
```diff
@@ -37,6 +38,13 @@ public class GetIssuesRequest : RequestBase<SonarQubeIssue[]>, IGetIssuesRequest
         [JsonProperty("key")]
         public virtual string ProjectKey { get; set; }
 
+        [JsonProperty("statuses", DefaultValueHandling = DefaultValueHandling.Ignore), DefaultValue(null)]
```
Evangelink: This property doesn't exist on the old API so let's make sure it's not serialized nor used.

src/SonarQube.Client/Api/Requests/V7_20/GetIssuesRequest.cs
```diff
@@ -40,18 +42,12 @@ public class GetIssuesRequest : PagedRequestBase<SonarQubeIssue>, IGetIssuesRequ
...
-                moduleKey: ComputeModuleKey(issue),
-                resolutionState: ParseResolutionState(issue.Resolution),
-                ruleId: GetRuleKey(issue.CompositeRuleKey));
+            new SonarQubeIssue(ComputePath(issue), issue.Hash, issue.Line, issue.Message, ComputeModuleKey(issue), 
+                GetRuleKey(issue.CompositeRuleKey), issue.Status);
 
         private static string ComputePath(ServerIssue issue) =>
             // Component is "{SubProject}:Path"
-            issue.SubProject != null 
+            issue.SubProject != null
```
Evangelink: Automatic removal of trailing space.

src/SonarQube.Client/Api/Requests/V7_20/GetIssuesRequest.cs
```diff
@@ -97,9 +73,9 @@ private class ServerIssue
             [JsonProperty("line")]
             public int? Line { get; set; }
             [JsonProperty("message")]
-            public string Message { get; set; }            
-            [JsonProperty("resolution")]
-            public string Resolution { get; set; }
```
Evangelink: We don't care about the resolution state but only about the status for our work.

src/SonarQube.Client/Models/SonarQubeIssue.cs
```diff
@@ -22,24 +22,24 @@ namespace SonarQube.Client.Models
...
         }
 
         public string FilePath { get; }
         public string Hash { get; }
         public int? Line { get; }
         public string Message { get; }
         public string ModuleKey { get; }
-        public SonarQubeIssueResolutionState ResolutionState { get; }
         public string RuleId { get; }
+        public string Status { get; }
```
valhristov: This is a bit cosmetic, but having a string Status does not feel nice, because it couples the rest of the code with a server-side value... Since we care only about Resolved/Unresolved states, perhaps a bool property would have been better. This way the magic strings would remain in the request object.

Evangelink: I wanted the code to remain generic enough but I think the point is valid and will simplify current code .

src/SonarQube.Client/Api/SonarQubeService.cs
```diff
@@ -233,15 +233,17 @@ public async Task<SonarQubeQualityProfile> GetQualityProfileAsync(string project
...
                 request =>
                 {
                     request.ProjectKey = projectKey;
+                    request.Statuses = statusResolved;
                 },
                 token);
 
             return result
-                .Where(x => x.ResolutionState != SonarQubeIssueResolutionState.Unresolved)
+                .Where(x => x.Status == statusResolved)
```
duncanp-sonar: Why is this filter required? Is it because the old API doesn't support querying for specific statuses, so the results still need to be filtered? If so, it would be worth calling this out in a comment as the filter looks to be redundant.

duncanp-sonar: Also, if the filter is required for the old API, are there any unit tests that check it works as expected?

src/SonarQube.Client/Api/Requests/V5_10/GetIssuesRequest.cs
```diff
@@ -73,24 +81,7 @@ private static IEnumerable<T> ReadFromProtobufStream<T>(Stream stream, MessagePa
...
-                case "FALSE-POSITIVE":
-                    return SonarQubeIssueResolutionState.FalsePositive;
-                case "FIXED":
-                    return SonarQubeIssueResolutionState.Fixed;
-                default:
-                    throw new ArgumentOutOfRangeException(nameof(resolution));
-            }
-        }
+            new SonarQubeIssue(issue.Path, issue.Checksum, issue.Line, issue.Msg, issue.ModuleKey, issue.RuleKey, 
+                issue.Status == "RESOLVED");
```
duncanp-sonar: It looks like the old API can return unresolved issues, and it also returns the _status_ field. Why are we hard-coding all issues to be resolved?

PR #753, opened 2018-10-23
#752 - fix tooltips for non-Roslyn languages

### LINKED ISSUE #752: [WIP] [C++] Issue tooltips don't work if the file has been edited
### Description

Issues are underlined correctly with squigglies even when the file is dirty (i.e. the developer has edited the document but not saved the changes). However, hovering over the text underlined by the squiggly may not show a tooltip, or may show information about the wrong issue.

Related to #701 (also about tooltips / the "quick info" code) 

### Repro steps

1. Add an issue to a C++ project file
2. Save the file and check that the issue appears in the error list, and that error information appears in the tooltip when you hovering over the squiggly that highlights the issue.
3. Delete a line before the issue and hover over the squiggly again.

### Expected behavior

A tooltip appears showing information about the error

### Actual behavior

No tooltip appears.

Variant: if there are multiple tooltips in the file the tooltip information that appears might relate to a different issue in the file.

### Known workarounds

Save the file.

### Related information

* SonarLint for VS Version: 
* Languages in your solution: C++. Probably repros for JavaScript too.

### Notes

The list of issues is refreshed when the file is saved i.e. the line number for each issue is the line number in the saved file.
The quick info/tooltip implementation is using the current line number in the file being edit to try to work out the tooltip information to show. This clearly won't work reliably if the file has been edited.

In SonarLint, the tooltip information for C++ and JavaScript is provided by an implementation of [IQuickInfoSource](https://docs.microsoft.com/en-us/visualstudio/extensibility/walkthrough-displaying-quickinfo-tooltips?view=vs-2017).

However, it may not be necessary to implement this interface to show tool tips. The squigglies are provided by creating an instance of _IErrorTag_ for each span that contains an error. SonarLint has code to handle translating the location of those error spans as the document is edited (which is why the squigglies remain in the correct place when the document is edited). The constructor for [_ErrorTag_](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.text.tagging.errortag?view=visualstudiosdk-2017) has an overload that accepts the tooltip content to display. It should be possible to use this to display the tooltip text correctly too.

duncanp-sonar:
Removed the _QuickInfoSource_ implementation and instead rely on the tooltip support for text tagged with _IErrorTag_.

Fixed #752


src/Integration.Vsix/SonarLintTagger/IssueTagger.cs
```diff
@@ -218,15 +241,16 @@ private void UpdateMarkers(ITextSnapshot currentSnapshot, IssuesSnapshot snapsho
...
             {
                 return Enumerable.Empty<ITagSpan<IErrorTag>>();
             }
 
             return snapshot.IssueMarkers
-                .Select(issue => issue.Span)
-                .Where(spans.IntersectsWith)
-                .Select(span => new TagSpan<IErrorTag>(span, new ErrorTag(PredefinedErrorTypeNames.Warning)));
+                .Where(marker => spans.IntersectsWith(marker.Span))
+                .Select(marker => new TagSpan<IErrorTag>(marker.Span, new ErrorTag(PredefinedErrorTypeNames.Warning, marker.Issue.Message)));
```
duncanp-sonar: There is a built-in _QuickInfoSource_ implementation that will automatically include tooltip information from _IErrorTags_. Here, we're using a different override of the _ErrorTag_ constructor to pass in the issue message as the tooltip info.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -114,11 +114,14 @@ internal sealed class TaggerProvider : IViewTaggerProvider, ITableDataSource, II
             {
                 lock (taggers)
                 {
-                    if (!taggers.ExistsForFile(textDocument.FilePath))
+                    IssueTagger tagger = null;
+                    // Always return the same tagger for a given document.
+                    // If we don't, then we won't automatically get support for tooltips for issues.
+                    if (!taggers.TryGetValue(textDocument.FilePath, out tagger))
```
duncanp-sonar: The built-in _QuickInfoSource_ implementation makes a separate request for a tagger i.e. there are (at least) two request for a tagger for a given document/text buffer, from different sources. Previously we were only returning a tagger for the initial request, which was enough to make squigglies appear under issues, but not enough to have tooltips appearing too.
Strictly speaking, we should be returning separate tagger instances for each request since in theory they could be disposed of separately. However, it seems that none of the tagger-requesting callers are disposed until the document is closed. I'll try some more manually testing to see if I can get it to break (I suspect opening a second tab for the same document might do it).

src/Integration.Vsix.UnitTests/Integration.Vsix.UnitTests.csproj
```diff
@@ -93,10 +93,6 @@
       <HintPath>..\..\packages\Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime.14.3.25407\lib\Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="Microsoft.VisualStudio.Language.Intellisense, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Microsoft.VisualStudio.Language.Intellisense.14.0.23205\lib\net45\Microsoft.VisualStudio.Language.Intellisense.dll</HintPath>
```
duncanp-sonar: I deleted all of the _QuickInfoSource_ code and references, which included this reference.
The problem with the previous implementation was that when it was trying to find the issue message to display in the tooltip, it was using the current line number in the document. If the document is dirty (i.e. has unsaved edits), then the line numbers in the unsaved document wouldn't match with the line numbers in the saved version -> issues not found, or out of range problems if lines had been added to the live document.

The _tag_ mechanism already handles this problem - when a document is edited (i.e. text buffer is updated), the text spans in the "old" text buffer are mapped across to the corresponding text in the "new" text buffer. Since there is also support for providing tooltip information with _IErrorTag_, this means we can rely on the tagging framework code to show the correct tooltip text even when the document is dirty.

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
```diff
@@ -239,49 +212,150 @@ public void Should_not_crash_when_sink_subscriber_gone()
...
+            mockTextDocument.Setup(x => x.TextBuffer).Returns(textBuffer);
+
+            // Register the buffer-to-doc mapping for the factory service
+            dummyDocumentFactoryService.RegisterDocument(mockTextDocument.Object);
+
+            return mockTextDocument.Object;
         }
 
-        private void SetMockDocumentFilename(string filename)
+        private class DummyTextDocumentFactoryService : ITextDocumentFactoryService
```
duncanp-sonar: I originally tried to use a _Mock_, but handling the _out_ parameters of the _TryGetDocument_ method was complicated and it was easier just to provide a simple implementation of the interface.

src/Integration.Vsix/Helpers/IIssueConverter.cs
```diff
@@ -22,7 +22,7 @@
 
 namespace SonarLint.VisualStudio.Integration.Vsix.Helpers
 {
-    public interface IIssueConverter
+    internal interface IIssueConverter
```
duncanp-sonar: Changed to _internal_ for consistency (none of the tagger/tracker classes need to be public and most of them are _internal_).

src/Integration.Vsix/SonarLintTagger/IssueTagger.cs
```diff
@@ -20,213 +20,71 @@
...
+    /// <remarks>
+    /// See the README.md in this folder for more information
     /// </remarks>
     internal sealed class IssueTagger : ITagger<IErrorTag>, IDisposable
     {
-        private readonly DTE dte;
-        private readonly TaggerProvider provider;
-        private readonly ITextBuffer textBuffer;
-        private readonly IEnumerable<SonarLanguage> detectedLanguages;
+        private IssuesSnapshot issues;
```
duncanp-sonar: Most of the product changes are in this class. Essentially, most of the code from the old version of this class was moved to the new _TextBufferIssueTracker_ class, leaving this class as a very thin disposable tagger.

src/Integration.Vsix/SonarLintTagger/SinkManager.cs
```diff
@@ -45,17 +53,17 @@ public void Dispose()
             taggerProvider.RemoveSinkManager(this);
         }
 
-        internal void AddFactory(SnapshotFactory factory)
+        public void AddFactory(SnapshotFactory factory)
```
duncanp-sonar: The class is already _internal_ so I've changed the methods to _public_ to make it clearer that they are really part of the public API of the class.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -30,27 +30,29 @@
...
-    internal sealed class TaggerProvider : IViewTaggerProvider, ITableDataSource, IIssueConsumer
+    internal sealed class TaggerProvider : IViewTaggerProvider, ITableDataSource
     {
         internal readonly ITableManager errorTableManager;
         internal readonly ITextDocumentFactoryService textDocumentFactoryService;
         internal readonly DTE dte;
 
-        private readonly List<SinkManager> managers = new List<SinkManager>();
-        private readonly TrackerManager taggers = new TrackerManager();
+        private readonly ISet<SinkManager> managers = new HashSet<SinkManager>();
```
duncanp-sonar: Changed to _ISet_ to make it clearer that values will be unique.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -112,24 +115,17 @@ internal sealed class TaggerProvider : IViewTaggerProvider, ITableDataSource, II
...
-                lock (taggers)
-                {
-                    if (!taggers.ExistsForFile(textDocument.FilePath))
-                    {
-                        return new IssueTagger(dte, this, buffer, textDocument, detectedLanguages)
-                            as ITagger<T>;
-                    }
-                }
+                // Multiple views could have that buffer open simultaneously, so only create one instance of the tracker.
+                var issueTracker = buffer.Properties.GetOrCreateSingletonProperty(typeof(TextBufferIssueTracker),
```
duncanp-sonar: Copied from the VS Extensibility [_Error List_ ](https://github.com/Microsoft/VSSDK-Extensibility-Samples/blob/master/ErrorList/C%23/SpellCheckerProvider.cs#L61 )sample - guaranteed way to create one tracker per file.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -143,71 +139,21 @@ internal void Rename(string oldPath, string newPath)
 
         #endregion
 
-        private class TrackerManager
```
duncanp-sonar: This class used to maintain the mapping between the file path and the single _IssueTracker_ for that file. No longer necessary.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -112,24 +115,17 @@ internal sealed class TaggerProvider : IViewTaggerProvider, ITableDataSource, II
...
+                    () => new TextBufferIssueTracker(dte, this, textDocument, detectedLanguages));
+
+                // Always create a new tagger for each request
+                return new IssueTagger(issueTracker) as ITagger<T>;
             }
 
             return null;
         }
 
-        internal void Rename(string oldPath, string newPath)
```
duncanp-sonar: Now handled by _TextBufferIssueTracker_ directly.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -221,29 +167,15 @@ public void RequestAnalysis(string path, string charset, IEnumerable<SonarLangua
             }
         }
 
-        public void Accept(string path, IEnumerable<Issue> issues)
```
duncanp-sonar: _Accept_ is the method on _IIssueConsumer_, the callback interface used by the daemon to return results. I've made the _TextBufferIssueHandler_ implement the interface directly (previously this class used to receive the results then then forward them to the issue handler).

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -0,0 +1,261 @@
...
+    /// Tracks SonarLint errors for a specific buffer.
+    ///</summary>
+    /// <remarks>
+    /// <para>The lifespan of this object is tied to the lifespan of the taggers on the view. On creation of the first tagger,
+    /// it starts tracking errors. On the disposal of the last tagger, it shuts down.</para>
+    /// <para>
+    /// See the README.md in this folder for more information
+    /// </para>
+    /// </remarks>
+    internal sealed class TextBufferIssueTracker : IIssueConsumer
```
duncanp-sonar: Most of this code was factored out of the old _IssueTagger_ class.
The new version is based very closely on [the equivalent class](https://github.com/Microsoft/VSSDK-Extensibility-Samples/blob/master/ErrorList/C%23/SpellChecker.cs) in the VS Extensibility Error List sample.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -0,0 +1,261 @@
...
+                Provider.RemoveIssueTracker(this);
+            }
+        }
+
+        private void FileActionOccurred(object sender, TextDocumentFileActionEventArgs e)
+        {
+            if (e.FileActionType == FileActionTypes.DocumentRenamed)
+            {
+                FilePath = e.FilePath;
+                ProjectItem = dte.Solution.FindProjectItem(this.FilePath);
```
duncanp-sonar: File renaming is now handled here.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -0,0 +1,261 @@
...
+            SnapshotSpan? affectedSpan = CalculateAffectedSpan(LastIssues, newIssues);
+            foreach (var tagger in activeTaggers)
+            {
+                tagger.UpdateMarkers(newIssues, affectedSpan);
+            }
+
+            this.LastIssues = newIssues;
+        }
+
+        private SnapshotSpan? CalculateAffectedSpan(IssuesSnapshot oldIssues, IssuesSnapshot newIssues)
```
duncanp-sonar: This is about the only major difference between this implementation and the VS SDK sample.
The SDK sample repeated this method in each lightweight _tagger_ class, but the calculation is the same in each case so I moved it to this class so it's only calculated once.

duncanp-sonar:
FYI I've updated the _README.md_ file with a more detailed explanation of the architecture so that would be a good place to start.

PR #754, opened 2018-10-25
Added unit tests for TextBufferIssueTracker

duncanp-sonar:
Related to #752

### DIFF
--- a/src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
+++ b/src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
@@ -89,6 +89,8 @@ namespace SonarLint.VisualStudio.Integration.Vsix
             var projectName = this.ProjectItem?.ContainingProject.Name ?? "{none}";
             this.Factory = new SnapshotFactory(new IssuesSnapshot(projectName, this.FilePath, 0,
                 new List<IssueMarker>()));
+
+            document.FileActionOccurred += FileActionOccurred;
         }
 
         public void AddTagger(IssueTagger tagger)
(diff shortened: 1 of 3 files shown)


src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -89,6 +89,8 @@ internal sealed class TextBufferIssueTracker : IIssueConsumer
             var projectName = this.ProjectItem?.ContainingProject.Name ?? "{none}";
             this.Factory = new SnapshotFactory(new IssuesSnapshot(projectName, this.FilePath, 0,
                 new List<IssueMarker>()));
+
+            document.FileActionOccurred += FileActionOccurred;
```
duncanp-sonar: Minor product change to fix a bug around file renaming - we need to start listening to file change events immediately to pick up file renames...

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -133,8 +134,13 @@ private void FileActionOccurred(object sender, TextDocumentFileActionEventArgs e
             {
                 FilePath = e.FilePath;
                 ProjectItem = dte.Solution.FindProjectItem(this.FilePath);
+
+                // Update and publish a new snapshow with the existing issues so 
+                // that the name change propagates to items in the error list
+                RefreshIssues();
```
duncanp-sonar: ... and if a file name occurs we need to re-publish the data to the Error List (otherwise the issues will still show the old file name and double-clicking on an error won't navigate to the error).

PR #756, opened 2018-10-26
Connected mode does not work with multiple developers if anonymous access is disabled

### LINKED ISSUE #726: Connected mode does not work with multiple developers if anonymous access is disabled
### Description

SonarLint Connected Mode does not work well in team scenarios if the bound server does not all anonymous access to a project. Everything works for the first developer who sets up the binding, but team members who try to then open the solution will get an error saying the bound project could not be found (unless they have previous connected to the same server so their credentials have already been cached).

This issue is a blocker for using Connected Mode if anonymous access is disabled.

See #722

### Repro steps

1. Developer _A_ binds a solution to a SonarQube server that does not allow anonymous access.
2. _A_ is prompted for his credentials, completes the binding, and checks in the _.sonarlint_ binding files.
3. Developer _B_ fetches the latest version of the code and opens the solution

### Expected behavior

_B_ is prompted for credentials if he has not previously connected to that server

### Actual behavior

_B_ is not prompted and the binding fails with a misleading error:

image

### Known workarounds
See #722 

### Related information
* SonarLint for VS Version: all

valhristov:
Fix #726

### DIFF
--- a/src/Integration/Resources/Strings.resx
+++ b/src/Integration/Resources/Strings.resx
@@ -312,7 +312,7 @@ ocalization note: '[try again]()' is a special markup used to translate the text
     <comment>Exception message</comment>
   </data>
   <data name="BoundProjectNotFound" xml:space="preserve">
-    <value>Bound project with key '{0}' was not found on the connected server. Select a new SonarQube project and bind to it.</value>
+    <value>Bound project with key '{0}' was not found on the connected server. Select a new SonarQube project and bind to it, or [use different login]() that has access to the desired project.</value>
     <comment>Team explorer error message when a persisted bound project is not found. {0} - SonarQube project key.</comment>
   </data>
   <data name="ConnectingToSonarQubePrefixMessageFormat" xml:space="preserve">
--- a/src/Integration/TeamExplorer/SectionController.cs
+++ b/src/Integration/TeamExplorer/SectionController.cs
@@ -201,6 +201,7 @@ namespace SonarLint.VisualStudio.Integration.TeamExplorer
             get;
             private set;
         }
+        public ICommand ReconnectCommand { get; private set; }
 
         public ICommand<ConnectionInformation> RefreshCommand
         {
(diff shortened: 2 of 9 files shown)


src/Integration/Resources/Strings.resx
```diff
@@ -312,7 +312,7 @@ ocalization note: '[try again]()' is a special markup used to translate the text
     <comment>Exception message</comment>
   </data>
   <data name="BoundProjectNotFound" xml:space="preserve">
-    <value>Bound project with key '{0}' was not found on the connected server. Select a new SonarQube project and bind to it.</value>
+    <value>Bound project with key '{0}' was not found on the connected server. Select a new SonarQube project and bind to it, or [use different login]() that has access to the desired project.</value>
```
duncanp-sonar: "_use a different login_"


Suggested tweaking for the existing wording:

_Bound project with key '{0}' was not found on the connected server. Select a different SonarQube project, or [use a different login]() that has access to the desired project._

src/Integration/TeamExplorer/SectionController.cs
```diff
@@ -259,6 +260,12 @@ private void InitializeProvidedCommands()
             this.ToggleShowAllProjectsCommand = new RelayCommand<ServerViewModel>(this.ToggleShowAllProjects, this.CanToggleShowAllProjects);
             this.BrowseToUrlCommand = new RelayCommand<string>(this.ExecBrowseToUrl, this.CanExecBrowseToUrl);
             this.BrowseToProjectDashboardCommand = new RelayCommand<ProjectViewModel>(this.ExecBrowseToProjectDashboard, this.CanExecBrowseToProjectDashboard);
+            this.ReconnectCommand = new RelayCommand(() =>
```
duncanp-sonar: Should this new command also be referenced in the _CleanProvidedCommands_ method below?

valhristov: I don't think it is required. The `ConnectCommand, RefreshCommand and BindCommand` are all coming from different controllers (classes) and I suppose the references are cleaned to prevent memory leaks. For example the DisconnectCommand is not cleaned there...

valhristov: Perhaps the whole `Clean` method is not needed... The class instances that contain the commands are created internally and there is nothing that would keep references... I don't know, it is difficult to see what are the possible implications.

PR #757, opened 2018-10-26
Fix NRE in Debug.Assert

duncanp-sonar:


src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -231,7 +231,7 @@ private string ProcessKey(Dictionary<string, string> keyToPath, SonarQubeIssue i
 
         private string NormalizeSonarQubePath(string path)
         {
-            Debug.Assert(!path.Contains("\\"), 
+            Debug.Assert(path == null || !path.Contains("\\"),
```
duncanp-sonar: The product code handles _path_ being null but the assertion didn't. This made the suppression handling code fail in _debug_ builds.

PR #759, opened 2018-10-29
Save credentials on successful connection

valhristov:


src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -165,6 +168,10 @@ private void OnProjectsChanged(ConnectionInformation connection, IEnumerable<Son
 
                 this.ConnectedServer = connection;
 
+                var creds = new Credential(connection.UserName, connection.Password.ToUnsecureString());
+                var credentials = new SecretStore(SolutionBindingSerializer.StoreNamespace);
+                credentials.WriteCredentials(connection.ServerUri, creds);
+
```
duncanp-sonar: It's might be worth saving the credentials for the server even if the there aren't any compatible plugins/projects in the solution (i.e. before the check in line 162). What do you think?

duncanp-sonar: Also, does this make the code that currently writes the credentials redundant?

valhristov: It makes sense to store the credentials, I'll add a commit

PR #782, opened 2019-01-08
Converted SonarLintIntegrationPackage to an asynchronous package

### LINKED ISSUE #680: Convert SonarLintIntegrationPackage to be asynchronous
See #679 for more information.

This is the only remaining package that is loaded synchronously in SLVS. We need to fix this for the performance reasons referred to in #679, which are impacting users currently.

Looking ahead, supported for synchronously-loaded packages is being dropped in [Visual Studio 2019](https://blogs.msdn.microsoft.com/visualstudio/2018/05/16/improving-the-responsiveness-of-critical-scenarios-by-updating-auto-load-behavior-for-extensions/):
* users will be warned that a package is loading synchronously in VS2019.0
* support for synchronous loading will be dropped in VS2019.1
Deadline: a release date hasn't been published for VS2019, but a preview has been available for a few months, and it is likely that the release will be in Q1 of 2019.

duncanp-sonar:
Fixed #680

### DIFF
--- a/src/Integration.Vsix/Delegates/SonarAnalyzerManager.cs
+++ b/src/Integration.Vsix/Delegates/SonarAnalyzerManager.cs
@@ -116,11 +116,13 @@ namespace SonarLint.VisualStudio.Integration.Vsix
             switch (configuration?.Mode)
             {
                 case SonarLintMode.Standalone:
+                    this.logger.WriteLine(Resources.Strings.AnalyzerManager_InStandaloneMode);
                     this.currentWorklow = new SonarAnalyzerStandaloneWorkflow(this.workspace);
                     break;
 
                 case SonarLintMode.LegacyConnected:
                 case SonarLintMode.Connected:
+                    this.logger.WriteLine(Resources.Strings.AnalyzerManager_InConnectedMode);
                     var sonarQubeIssueProvider = new SonarQubeIssuesProvider(sonarQubeService, configuration.Project.ProjectKey,
                         new TimerFactory(), this.logger);
                     this.disposableObjects.Add(sonarQubeIssueProvider);
--- a/src/Integration.Vsix/SonarLintIntegrationPackage.cs
+++ b/src/Integration.Vsix/SonarLintIntegrationPackage.cs
@@ -22,8 +22,8 @@ using System;
 using System.ComponentModel.Design;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
-using System.Runtime.Serialization;
-using System.Runtime.Serialization.Formatters.Binary;
+using System.Threading;
+using Microsoft.VisualStudio;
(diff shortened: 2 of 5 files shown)


src/Integration.Vsix/Delegates/SonarAnalyzerManager.cs
```diff
@@ -116,11 +116,13 @@ private void RefreshWorkflow(BindingConfiguration configuration)
             switch (configuration?.Mode)
             {
                 case SonarLintMode.Standalone:
+                    this.logger.WriteLine(Resources.Strings.AnalyzerManager_InStandaloneMode);
```
duncanp-sonar: Changing the package to be asynchronous affects when we start monitoring whether the open solution is in connected more or not. I added some logging to make it easier to see when monitoring has started (although it would be very useful to finally implement the ticket about supporting different levels of logging so we have a verbose mode).

src/Integration.Vsix/SonarLintIntegrationPackage.cs
```diff
@@ -34,11 +34,11 @@
 namespace SonarLint.VisualStudio.Integration.Vsix
 {
     // Register this class as a VS package.
-    [PackageRegistration(UseManagedResourcesOnly = true)]
+    [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
```
duncanp-sonar: This change follows the same pattern that Mads Kristensen used for converting the SonarLintDaemon package [here](https://github.com/SonarSource/sonarlint-visualstudio/commit/fd2b0ab5b675faab60d2b36dc0d0ed82100f86d5). Basically, we queue up some work to be done on the UI thread later.

Components that use MEF or that call _IServiceProvider.GetService_ can't be initialised on a background thread. We could convert some of the components so they can be initialised in the background, but that would required more work.

src/Integration.Vsix/SonarLintTelemetryPackage.cs
```diff
@@ -41,7 +41,7 @@ public sealed class SonarLintTelemetryPackage : AsyncPackage
 
         protected async override System.Threading.Tasks.Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
         {
-            Debug.Assert(!ThreadHelper.CheckAccess(), "Expecting the packatge to be initialized on the UI thread");
+            Debug.Assert(!ThreadHelper.CheckAccess(), "Not expecting the package to be initialized on the UI thread");
```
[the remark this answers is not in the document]
duncanp-sonar: Yes. The message was the wrong way round before:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.threadhelper.checkaccess?view=visualstudiosdk-2017

Issue #785, opened 2019-01-09
[Test projects] Projects tagged with TestContainer capability are not recognized as test projects


duncanp-sonar:
### Description
New NET.Core test projects for NUnit, XUnit, or MSTest have a _PackageReference_ to _Microsoft.Net.Sdk_. Alternatively, the new style projects will also have project following capability set: 
```
    <ProjectCapability Include="TestContainer" />
```
This won't appear in the project file, but it will be in a _props_ file added by the reference test framework.
It can be detected with the following code:
```
Microsoft.VisualStudio.Shell.PackageUtilities.IsCapabilityMatch(hierarchy, "TestContainer")
```

### Known workarounds

* Add the standard test _ProjectTypeGuid_ to the project file (as suggested in the SO post referred to above)
or
* Add the _SonarQubeTestProject_ property to the project file, set to _true_.

### Related information

* SonarLint for VS Version: 4.7
* Languages affected: C#, VB.Net

### Background
For more information on how the Scanner for MSBuild detects test project, see [Analysis of product projects vs. test projects](https://github.com/SonarSource/sonar-scanner-msbuild/wiki/Analysis-of-product-projects-vs.-test-projects) in the S4MSB wiki

duncanp-sonar:
Won't fix: this bug is no longer relevant if #632 is fixed.

PR #786, opened 2019-01-09
Fixed #760: shared projects should be excluded from binding

### LINKED ISSUE #760: [Connected mode] "Rule sets out of date" gold bar shown for shared projects
### Description

Shared projects are incorrectly reported as being out of date because they do not reference the SQ ruleset file.

### Repro steps

1. Create a solution containing a shared project and bind it to the server

2. Close and re-open the solution

### Expected behavior

The solution is up to date so there should not be a gold bar in the _Error List_ saying that an update is required.

### Actual behavior

Gold bar appears in the _Error List_ with the message:
> One or more rule sets are out of date or not linked to the SonarQube quality profile rule set. See output window (SonarLint) for affected projects."

The output window contains the following:

> DEBUGONLY: Connect workflow finished, Execution result: Succeeded
Checking for unbound projects.
Warning: could not find CodeAnalysisRuleSet for project 'SharedCode\SharedCode.shproj'
1 unbound projects were found (SharedCode\SharedCode.shproj).

### Known workarounds

None.

### Related information

* SonarLint for VS Version: all (up to 4.6 so far)
* VS2017 v15.8.6

### Notes
Shared projects are ignored during binding (so the ruleset isn't added to the project). However, shared projects don't seem to be ignored when checking whether the solution is up to date or not.

duncanp-sonar:
There is a separate issue with regard to test projects not being correctly recognized.
I've created #785 to track that bug, and I'm investigating another test project issue.

### DIFF
--- a/src/Integration.UnitTests/LocalServices/ProjectSystemFilterTests.cs
+++ b/src/Integration.UnitTests/LocalServices/ProjectSystemFilterTests.cs
@@ -340,23 +340,28 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
         public void ProjectSystemFilter_IsAccepted_UnrecognisedProjectType_ReturnsFalse()
         {
             // Arrange
+            var testSubject = this.CreateTestSubject();
             var project = new ProjectMock("unsupported.vcxproj");
             project.SetBuildProperty(Constants.SonarQubeTestProjectBuildPropertyKey, "False"); // Should not matter
 
             // Act and Assert
-            CheckProjectIsNotAccepted(project);
+            testSubject.IsAccepted(project).Should().BeFalse();
         }
 
         [TestMethod]
         public void ProjectSystemFilter_IsAccepted_SharedProject_ReturnsFalse()
         {
+            // Arrange
+            var testSubject = this.CreateTestSubject();
+
             var project = new ProjectMock("shared1.shproj");
             project.SetCSProjectKind();
-            CheckProjectIsNotAccepted(project);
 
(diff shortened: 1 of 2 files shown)


src/Integration.UnitTests/LocalServices/ProjectSystemFilterTests.cs
```diff
@@ -336,6 +336,29 @@ public void ProjectSystemFilter_SetTestRegex_ArgCheck()
             Exceptions.Expect<ArgumentNullException>(() => testSubject.SetTestRegex(null));
         }
 
+        [TestMethod]
+        public void ProjectSystemFilter_IsAccepted_UnrecognisedProjectType_ReturnsFalse()
```
duncanp-sonar: This first test is not related to the fix. I could find any tests that covered non-VB/CS projects so I added this one.

src/Integration.UnitTests/LocalServices/ProjectSystemFilterTests.cs
```diff
@@ -346,6 +369,15 @@ private ProjectSystemFilter CreateTestSubject()
             return new ProjectSystemFilter(host);
         }
 
+        private void CheckProjectIsNotAccepted(ProjectMock project)
+        {
+            var testSubject = this.CreateTestSubject();
+
+            var result = testSubject.IsAccepted(project);
+
+            result.Should().BeFalse();
```
valhristov: I think this method is not generally needed, it saves two lines and IMO the code in the tests is harder to understand

PR #792, opened 2019-01-11
Suppressions in shared/linked files do not work in SQ7.4

### LINKED ISSUE #778: Suppressions in shared/linked files do not work in SQ7.4
### Description

Issues in linked files or files in shared projects that have been marked as fixed in SQ (FP, Won't Fix...) are not suppressed in the IDE.

### Repro steps

1. Create a solution with a normal project and a shared project
2. Bind the solution to a SQ project
3. Add a class file in the shared project that contains an issue
4. Analyse the project using the Scanner for MSBuild
5. Mark the issue as a FP in SQ
6. Re-open the solution in VS and open the shared file (NB it might be necessary to trigger a new analysis by editing the file due to a different bug)

### Expected behavior

The suppressed issue is not reported in the IDE.

### Actual behavior

The issue is still reported.

### TODO
Check whether the issue also occurs for versions of SQ prior to 7.2 (the format of the data returned by the issues API changed in 7.2)

### Known workarounds

None

### Related information

* Visual Studio Version: VS2017 (not checked in VS2015 but presumably the same)
* SonarQube server version: 7.4

### Notes
SLVS issues a call like the following to fetch the list of issues:
`http://localhost:7400/api/issues/search?projects=slvs1&statuses=RESOLVED&p=1&ps=500`

SQ7.4 returns the following snippet for suppressions in linked/shared files:
```
    {
      "key": "AWgZ-nNTuwjFIkoqzRiF",
      "rule": "csharpsquid:S1144",
      "severity": "MAJOR",
      "component": "slvs1:shared/SharedProject1/ClassInSharedProject.cs",
      "project": "slvs1",
      "line": 11,
      "hash": "56a6afcaeeed577bf7e56b505661fd0b",
      "textRange": {
        "startLine": 11,
        "endLine": 14,
        "startOffset": 8,
        "endOffset": 9
      },
      "flows": [],
      "resolution": "FALSE-POSITIVE",
      "status": "RESOLVED",
      "message": "Remove the unused private method 'UnusedMethodInClassInSharedProject'.",
      "effort": "5min",
      "debt": "5min",
      "author": "anon@sonarsource.com",
      "tags": [ "unused" ],
      "creationDate": "2019-01-04T10:34:48+0000",
      "updateDate": "2019-01-04T18:19:40+0000",
      "type": "CODE_SMELL",
      "organization": "default-organization",
      "fromHotspot": false
    }

```
The problem is the format of the _component_ data, which isn't handled correctly by [this code](https://github.com/SonarSource/sonarlint-visualstudio/blob/9c5f245ec4113f2838d1f2178775e42b1b2de47f/src/SonarQube.Client/Api/Requests/V7_20/GetIssuesRequest.cs#L48) in SLVS.

### LINKED ISSUE #779: Suppressions do not work in SQ7.6 [drop modules alpha version]
### Description

Very similar to #778, except even issues in "normal" files are not suppressed.
The root cause appears to be the same: the _component_ field is not in the format expected by SLVS so it doesn't calculate the file path correctly.

### TODO
Check whether the bug is on the SQ side or in SLVS.

### Related information

* SonarQube server version: private alpha version of 7.6 used for testing the drop of modules

valhristov:
Fix #778 
Fix #779 (with the change it just works)

I changed the way we compute the FilePath of SonarQubeIssue. It used to be substring of `component` or `subProject`, now we lookup the corresponding component in the response and take the file path from there.

```
{
    "issues": [
		{
			"component": "component1",
			...
		}
    ],
    "components": [
        {
            "key": "component1",
            "qualifier": "FIL",
            "path": "the actual path of the file where the issue is raised",
 			...
        }
    ]
}
```

### DIFF
--- a/src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
+++ b/src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
@@ -73,176 +73,229 @@ namespace SonarQube.Client.Tests.Api
         {
             await ConnectToSonarQube("7.2.0.0");
 
-            SetupRequest("api/issues/search?projects=simplcom&statuses=RESOLVED&p=1&ps=500", @"
+            SetupRequest("api/issues/search?projects=com.github.kevinsawicki:http-request&statuses=RESOLVED&p=1&ps=500", @"
 {
+  ""total"": 5,
+  ""p"": 1,
+  ""ps"": 100,
   ""paging"": {
     ""pageIndex"": 1,
     ""pageSize"": 100,
-    ""total"": 1
+    ""total"": 5
   },
-  ""issues"": [{
-    ""key"": ""AWQYvw-4pbnviuOCCX9e"",
-    ""rule"": ""csharpsquid:S1075"",
-    ""severity"": ""MINOR"",
-    ""component"": ""simplcom:simplcom:91A9DD1C-08F3-4F85-8728-BB30F55FDACB:Controllers/ThemeApiController.cs"",
-    ""project"": ""simplcom"",
-    ""subProject"": ""simplcom:simplcom:91A9DD1C-08F3-4F85-8728-BB30F55FDACB"",
-    ""line"": 36,
-    ""hash"": ""0dcbf3b077bacc9fdbd898ff3b587085"",
-    ""textRange"": {
-        ""startLine"": 36,
-        ""endLine"": 36,
-        ""startOffset"": 49,
-        ""endOffset"": 107
+  ""issues"": [
+    {
+      ""key"": ""AWg8bjekFPFMeKWzHZ_6"",
+      ""rule"": ""csharpsquid:S3990"",
+      ""severity"": ""MAJOR"",
+      ""component"": ""shared:shared:D111C902-E7B1-4665-927E-A52E1443B8C1"",
+      ""project"": ""shared"",
+      ""flows"": [],
+      ""status"": ""OPEN"",
+      ""message"": ""Mark this assembly with 'System.CLSCompliantAttribute'"",
+      ""effort"": ""1min"",
+      ""debt"": ""1min"",
+      ""author"": """",
+      ""tags"": [
+        ""api-design""
+      ],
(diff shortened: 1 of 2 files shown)


src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -73,176 +73,229 @@ public async Task GetSuppressedIssuesAsync_From_7_20()
         {
             await ConnectToSonarQube("7.2.0.0");
 
-            SetupRequest("api/issues/search?projects=simplcom&statuses=RESOLVED&p=1&ps=500", @"
+            SetupRequest("api/issues/search?projects=com.github.kevinsawicki:http-request&statuses=RESOLVED&p=1&ps=500", @"
```
valhristov: Created a new version of the response with more issues and containing all fields in the JSON object (e.g. `components`).

src/SonarQube.Client/Api/Requests/V7_20/GetIssuesRequest.cs
```diff
@@ -35,26 +36,41 @@ public class GetIssuesRequest : PagedRequestBase<SonarQubeIssue>, IGetIssuesRequ
...
+            new SonarQubeIssue(ComputePath(issue, componentKeyPathLookup), issue.Hash, issue.Line, issue.Message, ComputeModuleKey(issue),
                 GetRuleKey(issue.CompositeRuleKey), issue.Status == "RESOLVED");
 
-        private static string ComputePath(ServerIssue issue) =>
-            // Component is "{SubProject}:Path"
-            issue.SubProject != null
-                ? issue.Component.Substring(issue.SubProject.Length + 1)
-                : string.Empty;
+        private static string ComputePath(ServerIssue issue, ILookup<string, string> componentKeyPathLookup) =>
+            componentKeyPathLookup[issue.Component].FirstOrDefault() ?? string.Empty;
```
valhristov: The actual fix is here. Instead of substring-ing `issue.Component` we look up the SonarQube-provided path in the `components` collection of the response.

PR #793, opened 2019-01-11
Fix #777: guard against unhandled errors in daemon editor integration

### LINKED ISSUE #777: [C++] Crash in visual studio while removing code that corresponded to an issue location
### Repro steps

* Create a C++ file with a function with unused parameter.
* Select the parameter name
* Delete it

In my case, the issue was in a header,, which means it could not be highlighter, AFAIK... I don't know if it matters...

There is a message from visual studio telling it has deactivated a plugin, and in `C:\Users\XXX\AppData\Roaming\Microsoft\VisualStudio\15.0_4aadb425\ActivityLog.xml` we can see:
```
  <entry>
    <record>830</record>
    <time>2019/01/03 14:53:03.568</time>
    <type>Error</type>
    <source>Editor or Editor Extension</source>
    <description>System.NullReferenceException: La r&#x00E9;f&#x00E9;rence d&apos;objet n&apos;est pas d&#x00E9;finie &#x00E0; une instance d&apos;un objet.&#x000D;&#x000A;   &#x00E0; SonarLint.VisualStudio.Integration.Vsix.TextBufferIssueTracker.TranslateMarkerSpans()&#x000D;&#x000A;   &#x00E0; SonarLint.VisualStudio.Integration.Vsix.TextBufferIssueTracker.OnBufferChange(Object sender, TextContentChangedEventArgs e)&#x000D;&#x000A;   &#x00E0; Microsoft.VisualStudio.Text.Utilities.GuardedOperations.RaiseEvent[TArgs](Object sender, EventHandler`1 eventHandlers, TArgs args)&#x000D;&#x000A;--- Fin de la trace de la pile &#x00E0; partir de l&apos;emplacement pr&#x00E9;c&#x00E9;dent au niveau duquel l&apos;exception a &#x00E9;t&#x00E9; lev&#x00E9;e ---&#x000D;&#x000A;   &#x00E0; Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)</description>
  </entry>
```
### Related information

* SonarLint for VS Version: 4.7.0.3891
* Visual Studio Version: 15.9.4 (Entreprise)
* Languages in my solution: C++

duncanp-sonar:


src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -121,38 +125,56 @@ public void RemoveTagger(IssueTagger tagger)
...
                 Provider.RemoveIssueTracker(this);
             }
         }
 
-        private void FileActionOccurred(object sender, TextDocumentFileActionEventArgs e)
+        private void SafeOnFileActionOccurred(object sender, TextDocumentFileActionEventArgs e)
         {
-            if (e.FileActionType == FileActionTypes.DocumentRenamed)
+            // Handles callback from VS. Suppress non-critical errors to prevent them
+            // propagating to VS, which would display a dialogue and disable the extension.
```
duncanp-sonar: We didn't manage to repro the error @loic-joly-sonarsource reported in #777.
However, we can log more info about any errors and prevent them from bubbling up to VS, which otherwise would display a dialogue and disable the extension.

Suppress exception shouldn't cause any major issues here; worst case is that the issue tagging for the current file will disappear/be wrong until it is refreshed by the file being saved.

duncanp-sonar: I've added try-catch blocks round the only two callbacks from VS in this class.

PR #799, opened 2019-01-15
[DO NOT SQUASH]  Request suppressed issues by type

### LINKED ISSUE #776: SonarLint does not suppress issues over 10k SonarQube API limit
### Description

We have a legacy application with lot of technical debt. There are more then 10k resolved issues today, but SonarLint does not suppress a lot of resolved issues on server. Debbuging the source code we found that it hapends because of de 10k limit of Issues API, used to suppress issues. Are there some improvement planned to handle this limitation? Please, do you have any suggestion of workaround to resolve this scenario?

### Repro steps

Please provide the steps required to reproduce the problem

1. Create a project with more then 10k issues and run SonarQube analysis for it;

2. Resolve more then 10k issues, by configuring "Source File Exclusions" on project, for example;

3. Run a new SonarQube analysis for project;

4. Open the project on Visual Studio 2017 using SonarLint on Connected Mode;

### Expected behavior

All issues resolved at server should by suppressed by SonarLint.

### Actual behavior

Only 10k resolved issues are suppressed by SonarLint, not all of them.

### Known workarounds

No workarounds found.

### Related information

* SonarLint for VS Version - 4.7
* Visual Studio Version - Professional 2017
* SonarQube server version - 7.2.0.13530
* Languages in your solution - C#

valhristov:
Fix #776

In general I am not sure if this change brings large benefit, because we will start doing three times more requests for each sync.

Two commits, a bit related:
- Moving the suppressed issues filtering from the service to the request - this is request-specific, so there is no need to filter already filtered data and potentially put pressure on the memory and CPU when copying large collections.
- Request suppressed issues by type.

### DIFF
--- a/src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
+++ b/src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
@@ -73,7 +73,7 @@ namespace SonarQube.Client.Tests.Api
         {
             await ConnectToSonarQube("7.2.0.0");
 
-            SetupRequest("api/issues/search?projects=shared&statuses=RESOLVED&p=1&ps=500", @"
+            SetupRequest("api/issues/search?projects=shared&statuses=RESOLVED&types=CODE_SMELL&p=1&ps=500", @"
 {
   ""total"": 5,
   ""p"": 1,
@@ -104,7 +104,22 @@ namespace SonarQube.Client.Tests.Api
       ""updateDate"": ""2019-01-11T11:28:22+0100"",
       ""type"": ""CODE_SMELL"",
       ""organization"": ""default-organization""
-    },
+    }
+  ],
+  ""components"": [ ]
+}
+");
+            SetupRequest("api/issues/search?projects=shared&statuses=RESOLVED&types=BUG&p=1&ps=500", @"
+{
+  ""total"": 5,
+  ""p"": 1,
+  ""ps"": 100,
+  ""paging"": {
+    ""pageIndex"": 1,
+    ""pageSize"": 100,
+    ""total"": 5
+  },
+  ""issues"": [
     {
       ""key"": ""AWg8adcV_JurIR2zdSvR"",
       ""rule"": ""csharpsquid:S1118"",
@@ -132,9 +147,36 @@ namespace SonarQube.Client.Tests.Api
       ],
       ""creationDate"": ""2019-01-11T11:16:30+0100"",
       ""updateDate"": ""2019-01-11T11:26:39+0100"",
-      ""type"": ""CODE_SMELL"",
+      ""type"": ""BUG"",
       ""organization"": ""default-organization""
-    },
+    }
+  ],
+  ""components"": [
+    {
+      ""organization"": ""default-organization"",
+      ""key"": ""shared:shared:2B470B7D-D47B-4E41-B105-D3938E196082:Program.cs"",
+      ""uuid"": ""AWg8adNk_JurIR2zdSvM"",
+      ""enabled"": true,
+      ""qualifier"": ""FIL"",
+      ""name"": ""Program.cs"",
+      ""longName"": ""Program.cs"",
+      ""path"": ""Program.cs""
+    }
+  ]
+}
+");
+
+            SetupRequest("api/issues/search?projects=shared&statuses=RESOLVED&types=VULNERABILITY&p=1&ps=500", @"
+{
+  ""total"": 5,
+  ""p"": 1,
+  ""ps"": 100,
+  ""paging"": {
+    ""pageIndex"": 1,
+    ""pageSize"": 100,
+    ""total"": 5
+  },
+  ""issues"": [
     {
       ""key"": ""AWg9DV27DpKqrfA7luen"",
       ""rule"": ""csharpsquid:S1451"",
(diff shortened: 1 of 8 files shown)


src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -73,7 +73,7 @@ public async Task GetSuppressedIssuesAsync_From_7_20()
         {
             await ConnectToSonarQube("7.2.0.0");
 
-            SetupRequest("api/issues/search?projects=com.github.kevinsawicki:http-request&statuses=RESOLVED&p=1&ps=500", @"
```
valhristov: cosmetic, fixed the project key to match the response items

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -84,26 +84,6 @@ public async Task GetSuppressedIssuesAsync_From_7_20()
...
-      ""debt"": ""1min"",
-      ""author"": """",
-      ""tags"": [
-        ""api-design""
-      ],
-      ""creationDate"": ""2019-01-11T11:21:20+0100"",
-      ""updateDate"": ""2019-01-11T11:21:20+0100"",
-      ""type"": ""CODE_SMELL"",
-      ""organization"": ""default-organization""
-    },
```
valhristov: removed the open items from the responses to simulate the server-side filtering

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -263,31 +205,12 @@ public async Task GetSuppressedIssuesAsync_From_7_20()
...
-      ""qualifier"": ""TRK"",
-      ""name"": ""shared"",
-      ""longName"": ""shared""
     }
   ]
 }
 ");
 
-            var result = await service.GetSuppressedIssuesAsync("com.github.kevinsawicki:http-request", CancellationToken.None);
+            var result = await service.GetSuppressedIssuesAsync("shared", CancellationToken.None);
```
valhristov: cosmetic, fixed the project key to match the response items

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -263,31 +205,12 @@ public async Task GetSuppressedIssuesAsync_From_7_20()
...
-      ""path"": ""ConsoleApp1""
-    },
-    {
-      ""organization"": ""default-organization"",
-      ""key"": ""shared"",
-      ""uuid"": ""AWg8aLQfhQTYUbm7ZplB"",
-      ""enabled"": true,
-      ""qualifier"": ""TRK"",
-      ""name"": ""shared"",
-      ""longName"": ""shared""
```
valhristov: removed the unused components

src/SonarQube.Client/Api/SonarQubeService.cs
```diff
@@ -231,20 +231,15 @@ public async Task<SonarQubeQualityProfile> GetQualityProfileAsync(string project
...
-        {
-            const string statusResolved = "RESOLVED";
-            var result = await InvokeRequestAsync<IGetIssuesRequest, SonarQubeIssue[]>(
+        public async Task<IList<SonarQubeIssue>> GetSuppressedIssuesAsync(string projectKey, CancellationToken token) =>
+            await InvokeRequestAsync<IGetIssuesRequest, SonarQubeIssue[]>(
                 request =>
                 {
                     request.ProjectKey = projectKey;
-                    request.Statuses = statusResolved;
+                    request.Statuses = "RESOLVED"; // Resolved issues will be hidden in SLVS
```
valhristov: We should create an enumeration for this, removed the constant for consistency as we have other literals in this class

src/SonarQube.Client.Tests/Api/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -73,7 +73,7 @@ public async Task GetSuppressedIssuesAsync_From_7_20()
         {
             await ConnectToSonarQube("7.2.0.0");
 
-            SetupRequest("api/issues/search?projects=shared&statuses=RESOLVED&p=1&ps=500", @"
+            SetupRequest("api/issues/search?projects=shared&statuses=RESOLVED&types=CODE_SMELL&p=1&ps=500", @"
```
valhristov: The new request implementation is doing 3 times more requests...

valhristov:
We could do something even better: instead of retrieving issues by type (which only reduces the effect of this limitation) we can sort the issues by date. After the first request if there are 10k results we can request more issues with date larger than the last issue... This should completely fix the limitation.

valhristov:
Update: the suggestion is not very convenient since the API does not allow to nicely paginate the results by date (or something else). Both date filters are inclusive and this means that we could still miss issues or we could have duplicates in the result. On the other hand, if there are more than 10k issues with the same creationDate, we could enter in an endless loop... We could discuss this on standup or a dedicated 10-15min meeting (if Ammo wants to come).

PR #802, opened 2019-01-16
[DO NOT SQUASH] Fix #801: updating a bound solution failed in TfVC

### LINKED ISSUE #801: [TFS] Updating an existing binding fails
### Description

Updating an existing binding fails when the source control system is TfVC.
It works correctly for Git.

### Repro steps

1. Create a solution, commit it to TfVC, and bind the solution to SonarQube.

2. Go to the SonarQube pane in the _Team Explorer_ and select _Update_

### Expected behavior

The binding update succeeds.

### Actual behavior

The update fails. The following gold bar is displayed.

image

There is no useful information in the output window.

image

### Known workarounds

Manually check out the files under the _.sonarlint_ folder before updating the binding:

image

### Related information

* Visual Studio Version: VS2015 and VS2017, all SKUs
* Languages affected: C\, VB

### Notes
On debugging, the failure is happening when attempting to check the files for edit in [this line](https://github.com/SonarSource/sonarlint-visualstudio/blob/aadf9950c7d2eb3a6978738e071de752a4fcd1f5/src/Integration/LocalServices/SourceControlledFileSystem.cs#L159).
The error code that is returned is ok so the code does not throw.
However, the _verdict_ is not _QER_EditOK_ so the method returns 
The [_moreInfo_](https://docs.microsoft.com/cs-cz/dotnet/api/microsoft.visualstudio.shell.interop.tagvsqueryeditresultflags?view=visualstudiosdk-2017) value is 272 i.e. QER_ReadOnlyUnderScc | QER_NoisyCheckoutRequired

duncanp-sonar:
Fixed #801

### DIFF
--- a/src/Integration/LocalServices/SourceControlledFileSystem.cs
+++ b/src/Integration/LocalServices/SourceControlledFileSystem.cs
@@ -36,25 +36,31 @@ namespace SonarLint.VisualStudio.Integration
     internal class SourceControlledFileSystem : ISourceControlledFileSystem
     {
         private readonly IServiceProvider serviceProvider;
+        private readonly ILogger logger;
         private readonly IFileSystem fileSystemWrapper;
         private readonly HashSet<string> filesEdit = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
         private readonly HashSet<string> filesCreate = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
         private readonly Queue<Func<bool>> fileWriteOperations = new Queue<Func<bool>>();
         private IVsQueryEditQuerySave2 queryFileOperation;
 
-        public SourceControlledFileSystem(IServiceProvider serviceProvider)
-            : this(serviceProvider, null)
+        public SourceControlledFileSystem(IServiceProvider serviceProvider, ILogger logger)
+            : this(serviceProvider, logger, null)
         {
         }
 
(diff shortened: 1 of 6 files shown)


duncanp-sonar:
There are two commits, and I'd suggest reviewing them separately.
The first just adds logging of error messages from our source-controlled file system wrapper.
The second (much smaller change) fixes the bug.

src/Integration/LocalServices/SourceControlledFileSystem.cs
```diff
@@ -146,7 +152,10 @@ void IFileSystem.CreateDirectory(string path)
         #region Helpers
         private bool CheckoutForEdit(params string[] fileNames)
         {
-            VsQueryEditFlags flags = VsQueryEditFlags.SilentMode | VsQueryEditFlags.DetectAnyChangedFile;
+            VsQueryEditFlags flags = VsQueryEditFlags.SilentMode | VsQueryEditFlags.DetectAnyChangedFile |
+                // Force no prompting here to fix SLVS#801 (otherwise TFS would prompt about the slconfig
+                // and generated ruleset files in new connected mode that are not in the solution)
+                VsQueryEditFlags.ForceEdit_NoPrompting;
```
duncanp-sonar: This is the bug fix - passing an extra flag.

I've manually tested the change in VS2017 on solution that are not in SCC, in Git and in TfVC.

PR #810, opened 2019-02-06
 Adding SonarSource/sonarqube-webclient-dotnet as submodule

valhristov:
All modifications are due to some namespace changes I did in the client repo.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "sonarqube-webclient"]
+	path = sonarqube-webclient
+	url = https://github.com/sonarsource/sonarqube-webclient-dotnet
```
valhristov: That's the important change, check the path where the submodule is located

PR #831, opened 2019-03-19
Fixed build number var in ci-build

duncanp-sonar:


build/ci-build.ps1
```diff
@@ -90,7 +91,16 @@ try {
...
+    Write-Header "Build parameters:"
     Write-Host "Branch: ${branchName}"
+    Write-Host "Build number: ${buildNumber}"
+    
+    if ($isPullRequest) {
+      Write-Host "Build kind: PR"
+      Write-Host "PR: ${githubPullRequest}"
+      Write-Host "PR source: ${githubPRBaseBranch}"
+      Write-Host "PR target: ${githubPRTargetBranch}"
+    }
```
duncanp-sonar: I moved some but not all of the variable logging here. The logging about branch kind etc is inside nested conditional statements just after this code, so I thought I'd leave it there rather than repeating the conditional logic.

PR #833, opened 2019-03-19
Add UTs to ensure the right version of dependencies is included

Evangelink:


src/Integration.TeamExplorer.UnitTests/Integration.TeamExplorer.UnitTests.csproj
```diff
@@ -246,16 +246,7 @@
...
-    <Compile Include="Framework\ConfigurableTeamExplorer.cs" />
-    <Compile Include="Framework\ConfigurableTeamExplorerPage.cs" />
-    <Compile Include="HostedCommandControllerBaseTests.cs" />
-    <Compile Include="Progress\ProgressStepRunnerTests.cs" />
-    <Compile Include="Properties\AssemblyInfo.cs" />
-    <Compile Include="SectionControllerTests.cs" />
-    <Compile Include="SonarQubeNavigationItemTests.cs" />
-    <Compile Include="SonarQubePageTests.cs" />
-    <Compile Include="TeamExplorerControllerTests.cs" />
+    <Compile Include="**\*.cs" Exclude="**\obj\**\*" />
```
Evangelink: I can't say that's the easiest solution to work with but at least now this is consistent with how other projects are set up.

src/Integration.TeamExplorer.UnitTests/ReferencesTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace SonarLint.VisualStudio.Integration.UnitTests
+{
+    [TestClass]
+    public class ReferencesTests
+    {
+        [TestMethod]
+        public void MicrosoftTeamFoundationClient_EnsureCorrectVersion()
+        {
+            var assemblyVersion = typeof(Microsoft.TeamFoundation.PluginCatalog).Assembly.GetName().Version;
```
duncanp-sonar: This is checking the assembly referenced by the test project. A more robust check would be to check the assembly referenced by our product assembly e.g.
```
var sonarLintAssembly = typeof(some type in our assembly).Assembly;
var tfAssembly = sonarLintAssembly.GetReferencedAssemblies()
    .Where(ra => ra.Name == "Microsoft.TeamFoundation.Client.dll");
```

src/Integration.TeamExplorer.UnitTests/ReferencesTests.cs
```diff
@@ -0,0 +1,52 @@
...
+            var callingAssembly = Assembly.GetCallingAssembly().GetName().Version;
```
duncanp-sonar: I think it's worth calling out explicitly the assumption we're making here i.e. that the version of the test execution is the one that corresponds to the target version of VS.

src/Integration.TeamExplorer.UnitTests/Integration.TeamExplorer.UnitTests.csproj
duncanp-sonar: I'd left it with explicit references because it was easier to work with, but it doesn't matter too much either way.

PR #840, opened 2019-03-20
Update process so that our VS2017 CI machine produces the VS2019 artifact

### LINKED ISSUE #780: Add support for VS2019 RTM
Add support for the RTM version of VS2019

### TODO
* finalise decision on whether to release a separate extension for VS2019, or to support both VS2017 and VS2019 in a single extension
* testing
* update the marketplace
* product announcements/marketing

### LINKED ISSUE #838: Fixup CI to remove requirement to have the 3 VS installed
### Description

We currently need one machine with VS2015, one with VS2017 and one with VS2019 to be able to produce the 3 VSIX artifacts.

The idea is to simplify the process by embedding the few specific references we need directly in the repository.

Evangelink:
Fix #780 and fix #838

### DIFF
--- a/build/vs2017.bat
+++ b/build/vs2017.bat
@@ -1,2 +1,2 @@
 call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools\VsDevCmd.bat"
-msbuild SonarLint.VisualStudio.Integration.sln /t:rebuild /p:configuration=Release /p:DeployExtension=false /p:ZipPackageCompressionLevel=normal /v:m /p:defineConstants=SignAssembly /p:SignAssembly=true /p:AssemblyOriginatorKeyFile=%CERT_PATH%
+msbuild SonarLint.VisualStudio.Integration.sln /t:rebuild /p:configuration=Release /p:DeployExtension=false /p:ZipPackageCompressionLevel=normal /v:m /p:defineConstants=SignAssembly /p:SignAssembly=true /p:AssemblyOriginatorKeyFile=%CERT_PATH% /p:VsTargetVersion=%VsTargetVersion%
--- a/src/Integration.TeamExplorer.UnitTests/Integration.TeamExplorer.UnitTests.csproj
+++ b/src/Integration.TeamExplorer.UnitTests/Integration.TeamExplorer.UnitTests.csproj
@@ -31,7 +31,7 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <OutputPath>bin\Debug\</OutputPath>
+    <OutputPath>bin\Debug\VS$(VersionSpecificSuffix)\</OutputPath>
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
@@ -39,7 +39,7 @@
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
-    <OutputPath>bin\Release\</OutputPath>
+    <OutputPath>bin\Release\VS$(VersionSpecificSuffix)\</OutputPath>
     <DefineConstants>TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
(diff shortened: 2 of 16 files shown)


src/Integration.TeamExplorer.UnitTests/Integration.TeamExplorer.UnitTests.csproj
```diff
@@ -31,15 +31,15 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <OutputPath>bin\Debug\</OutputPath>
+    <OutputPath>bin\Debug\VS$(VersionSpecificSuffix)\</OutputPath>
```
duncanp-sonar: All of the projects that have version-specific references now have version-specific output folders.
This makes things simpler when building multiple version on the same machine - avoids MSBuild/VS getting confused about whether it need to rebuild a project or not.

I tried making all of the assembly names version-specific too, but that cause problems with the `InternalsVisibleTo` attributes, so I undid that change.

src/Integration/Integration.csproj
```diff
@@ -25,7 +25,7 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <OutputPath>bin\Debug\</OutputPath>
+    <OutputPath>bin\Debug\VS$(VersionSpecificSuffix)\</OutputPath>
```
Evangelink: Is this project version specific?

src/SonarLint.VSSpecificAssemblies.props
```diff
@@ -15,7 +16,60 @@
...
+  <Target Name="CheckVsTargetVersionSetting" BeforeTargets="PrepareForBuild">
+    <Error 
+      Condition=" $(BuildingInsideVisualStudio) != 'true' AND $(VsTargetVersion) == '' "
+      Text="Must set VsTargetVersion build property when building outside Visual Studio. Expected values: 2015, 2017, 2019"/>
+
+    <Error
+      Condition=" $(BuildingInsideVisualStudio) != 'true' AND $(VsTargetVersion) != '2015' AND $(VsTargetVersion) != '2017' AND $(VsTargetVersion) != '2019' "
+      Text="Invalid value for VsTargetVersion. Expected values: 2015, 2017, 2019. Actual value: $(VsTargetVersion)"/>
+    
+    <!-- Check build constraints -->
```
duncanp-sonar: I added a couple more checks to make it easier to diagnose build errors due to trying to build invalid combinations e.g. MSBuild 14 + VS2019.

src/Integration/Integration.csproj
duncanp-sonar: Yes; it imports the legacy VS ruleset assembly.

build/vs2017.bat
```diff
@@ -1,2 +1,2 @@
 call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools\VsDevCmd.bat"
-msbuild SonarLint.VisualStudio.Integration.sln /t:rebuild /p:configuration=Release /p:DeployExtension=false /p:ZipPackageCompressionLevel=normal /v:m /p:defineConstants=SignAssembly /p:SignAssembly=true /p:AssemblyOriginatorKeyFile=%CERT_PATH%
\ No newline at end of file
+msbuild SonarLint.VisualStudio.Integration.sln /t:rebuild /p:configuration=Release /p:DeployExtension=false /p:ZipPackageCompressionLevel=normal /v:m /p:defineConstants=SignAssembly /p:SignAssembly=true /p:AssemblyOriginatorKeyFile=%CERT_PATH% /p:VsTargetVersion=%vsTargetVersion%
```
duncanp-sonar: The MSBuild targets are expecting 2015, 2017 or 2019, not VS2015 etc.
I think it's ok just to have 2015 etc since the variable name is `vsTargetVersion`.

src/Integration.TeamExplorer.UnitTests/ReferencesTests.cs
```diff
@@ -55,14 +49,31 @@ public void MicrosoftTeamFoundationControls_EnsureCorrectVersion()
...
+            {
+                dllMajorVersion.Should().Be(15);
+            }
+            else if (executingAssemblyLocation.Contains("\\VS2019\\"))
+            {
+                dllMajorVersion.Should().Be(16);
+            }
+            else
+            {
+                Assert.Fail("Expecting the path to contain one of 'VS2015', 'VS2017' or 'VS2019'.");
```
duncanp-sonar: I'd prefix the message with something like "Test setup error: " to make it clear from the logs that we didn't get as far as actually checking the product i.e. we don't know whether the product code is bad or not.

Normally I'd use an `Assert.Inconclusive` for test setup failures but that doesn't fail the build, and in this case the consequences are serious enough that we wouldn't want to ship until we'd resolved the issue.

PR #844, opened 2019-03-21
Fix #835: initialisation should be on the UI thread

### LINKED ISSUE #835: Initialization should happen on the UI thread
### Description

Following the extraction of the code linked to TeamExplorer, some components are now created on the background thread, causing some `Debug.Assert` failures.

They don't appear to have any critical impact yet they should be fixed.

duncanp-sonar:
Fixed #835 

The refactoring of the Team Explorer plugins to a separate assembly changed the order of initialisation of components, which led to some of the UI-related components being created on background threads in some circumstances (i.e. timing related)

### DIFF
--- a/src/Integration.Vsix/SonarLintNotificationsPackage.cs
+++ b/src/Integration.Vsix/SonarLintNotificationsPackage.cs
@@ -76,21 +76,23 @@ namespace SonarLint.VisualStudio.Integration.Vsix
 
             notifications = new SonarQubeNotificationService(sonarqubeService,
                 new NotificationIndicatorViewModel(), new TimerWrapper { Interval = 60000 }, logger);
-
-            activeSolutionBoundTracker = await this.GetMefServiceAsync<IActiveSolutionBoundTracker>();
-
-            // A bound solution might already have completed loading. If so, we need to
-            // trigger the load of the options from the solution file
-            if (activeSolutionBoundTracker.CurrentConfiguration.Mode != SonarLintMode.Standalone)
-            {
-                var solutionPersistence = (IVsSolutionPersistence)await this.GetServiceAsync(typeof(SVsSolutionPersistence));
-                solutionPersistence.LoadPackageUserOpts(this, NotificationDataKey);
-            }
-
+            
             // Initialising the UI elements has to be on the main thread
             await JoinableTaskFactory.SwitchToMainThreadAsync();
             SafePerformOpOnUIThread(() =>
             {
+                // Creating the tracker might indirectly cause UI-related MEF components to be
+                // created, so this needs to be done on the UI thread just in case.
+                activeSolutionBoundTracker = this.GetMefService<IActiveSolutionBoundTracker>();
+
+                // A bound solution might already have completed loading. If so, we need to
+                // trigger the load of the options from the solution file
+                if (activeSolutionBoundTracker.CurrentConfiguration.Mode != SonarLintMode.Standalone)
+                {
+                    var solutionPersistence = (IVsSolutionPersistence)this.GetService(typeof(SVsSolutionPersistence));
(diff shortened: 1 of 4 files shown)


src/Integration.Vsix/SonarLintNotificationsPackage.cs
```diff
@@ -76,21 +76,23 @@ private async System.Threading.Tasks.Task InitAsync()
...
-            activeSolutionBoundTracker = await this.GetMefServiceAsync<IActiveSolutionBoundTracker>();
-
-            // A bound solution might already have completed loading. If so, we need to
-            // trigger the load of the options from the solution file
-            if (activeSolutionBoundTracker.CurrentConfiguration.Mode != SonarLintMode.Standalone)
-            {
-                var solutionPersistence = (IVsSolutionPersistence)await this.GetServiceAsync(typeof(SVsSolutionPersistence));
-                solutionPersistence.LoadPackageUserOpts(this, NotificationDataKey);
-            }
-
```
duncanp-sonar: This section hasn't changed - I just moved it inside the the block below that is executed on the UI thread.

src/Integration/Binding/BindingWorkflow.cs
```diff
@@ -336,7 +336,7 @@ private void UpdateDownloadedSonarQubeQualityProfile(RuleSet ruleSet, SonarQubeQ
 
         private void InitializeSolutionBindingOnUIThread(IProgressStepExecutionEvents notificationEvents)
         {
-            Debug.Assert(System.Windows.Application.Current?.Dispatcher.CheckAccess() ?? false, "Expected to run on UI thread");
```
duncanp-sonar: I've changed all of the direct usages of `System.Windows.Application.Current?.Dispatcher` to `host.UIDispatcher` for consistency.

src/Integration/LocalServices/ErrorListInfoBarController.cs
```diff
@@ -114,7 +114,9 @@ public void Refresh()
         // possible use #if DEBUG instead.
         private void AssertOnUIThread()
         {
-            Debug.Assert(this.host.UIDispatcher.CheckAccess(), "The controller needs to run on the UI thread");
+            // Note: "this.host.UIDispatcher.CheckAccess()" does not return the correct result when 
+            // VS is shutting down. However, ThreadHelper.CheckAccess() does.
+            Debug.Assert(ThreadHelper.CheckAccess(), "The controller needs to run on the UI thread");
```
duncanp-sonar: This is an odd case - I debugged, and `host.UIDispatcher.CheckAccess()` was returning false on shutdown, even though the _ManagedThreadId_ was 1. 
Using `ThreadHelper.CheckAccess()` returns the expected result, even in the case of shutdown.

src/Integration/WPF/RelayCommandBase.cs
```diff
@@ -32,7 +32,7 @@ public void RequeryCanExecute()
         {
             if (System.Windows.Application.Current != null)
             {
-                Debug.Assert(System.Windows.Application.Current.Dispatcher.CheckAccess(), "RequeryCanExecute should be called from the UI thread");
+                Debug.Assert(Microsoft.VisualStudio.Shell.ThreadHelper.CheckAccess(), "RequeryCanExecute should be called from the UI thread");
```
duncanp-sonar: This class doesn't have a reference to an `IHost` object so I've used `ThreadHelper.CheckAccess()` directly here.

duncanp-sonar:
Tactic fix to prevent the assertions on startup. I've made the smallest change possible to fix the issue.
The unit tests pass for all three versions of VS, and I've manually tested that the assertion don't occur in all three versions either.

Evangelink:
I confirm that the assertions have been fixed!

PR #845, opened 2019-03-21
Workaround the 10K projects limitation

### LINKED ISSUE #813: Error connecting to a server that has more than 10k projects
### Description

Configuring a SonarQube connection / updating the binding fails when more than 10K projects+modules are defined in SonarQube.

See the related tickets for other SonarLint IDEs
https://jira.sonarsource.com/browse/SLI-258
https://jira.sonarsource.com/browse/SLE-229
... and the ticket about adding a search feature: https://jira.sonarsource.com/browse/SLI-267

### Actual behavior

Error reported in the output window and binding fails:

d566985b7ec80e9b3419d80471347da6d074f120

### Known workarounds

None

### Related information

* SonarLint for VS Version: all (current is v4.8)

### Fix
The fix will be the same as for IntelliJ and Eclipse - we'll fetch the first 10,000 items, then stop, rather than failing.
If the project is in the first 10k they will be able to bind as normal.
If it is not, they will be able to manually work round the issue by manually editing/creating the binding file to specify the target project.

Evangelink:
Fix #813


src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -183,6 +183,16 @@ private void OnProjectsChanged(ConnectionInformation connection, IEnumerable<Son
                 var projects = await this.host.SonarQubeService.GetAllProjectsAsync(connection.Organization?.Key,
                     cancellationToken);
 
+                if (string.IsNullOrEmpty(connection.Password.ToUnsecureString()) &&
+                    projects.Count == 10000 &&
+                    !projects.Any(p => p.Key == this.host.VisualStateManager.BoundProjectKey))
+                {
+                    projects = new List<SonarQubeProject>
+                    {
+                        new SonarQubeProject(this.host.VisualStateManager.BoundProjectKey, this.host.VisualStateManager.BoundProjectName)
```
Evangelink: We need the project name otherwise the selected/bound project doesn't show-up correctly.

src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -183,6 +183,16 @@ private void OnProjectsChanged(ConnectionInformation connection, IEnumerable<Son
                 var projects = await this.host.SonarQubeService.GetAllProjectsAsync(connection.Organization?.Key,
                     cancellationToken);
 
+                if (string.IsNullOrEmpty(connection.Password.ToUnsecureString()) &&
```
Evangelink: I'll add a comment when I am sure this works but the idea is to say we want to have the special code only when we hit the 10K issue, the project we want to bind to is not in the 10K projects we have fetched AND this happens only when opening an already bound solution.

src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -243,7 +253,7 @@ private SonarQubeOrganization AskUserToSelectOrganizationOnUIThread(IEnumerable<
 
             notifications.ProgressChanged(Strings.DownloadingServerSettingsProgessMessage);
 
-            var properties = await this.host.SonarQubeService.GetAllPropertiesAsync(token);
+            var properties = await this.host.SonarQubeService.GetAllPropertiesAsync(this.host.VisualStateManager.BoundProjectKey, token);
```
Evangelink: After updating the client, I am now forced to pass a project key to this method.

src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -183,6 +183,21 @@ private void OnProjectsChanged(ConnectionInformation connection, IEnumerable<Son
                 var projects = await this.host.SonarQubeService.GetAllProjectsAsync(connection.Organization?.Key,
                     cancellationToken);
 
+                // The SonarQube client will limit the number of returned projects to 10K (hard limit on SonarQube side)
+                // but will no longer fail when trying to retrieve more. In the case where the project we want to bind to
+                // is not in the list AND the binding was already done (the .slconfig was manually created) then we are
+                // going to pretend we managed to retrieve this project from the list.
+                if (string.IsNullOrEmpty(connection.Password.ToUnsecureString()) &&
```
duncanp-sonar: I don't understand why the check for an empty password is needed. What happens if the user has supplied credentials?

src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -183,6 +183,21 @@ private void OnProjectsChanged(ConnectionInformation connection, IEnumerable<Son
...
+                    projects.Count == 10000 &&
+                    !projects.Any(p => p.Key == this.host.VisualStateManager.BoundProjectKey))
+                {
+                    projects = new List<SonarQubeProject>
+                    {
+                        new SonarQubeProject(this.host.VisualStateManager.BoundProjectKey,
+                            this.host.VisualStateManager.BoundProjectName ?? this.host.VisualStateManager.BoundProjectKey)
+                    };
```
duncanp-sonar: We could just add the required project to the returned list rather than replacing the list of returned items completely.  Not sure which approach is preferable.

src/Integration/Connection/ConnectionWorkflow.cs
Evangelink: From the different tests I have done, this is the best way I found to detect the case of a binding for a solution that was already bound. Basically when opening a solution which is already bound this password is empty for some reason (probably because we then use windows credential store).

src/Integration/Connection/ConnectionWorkflow.cs
Evangelink: I think it's better to add! At least we have a consistent UI when the user clicks the show all projects button.

src/Integration/Connection/ConnectionWorkflow.cs
duncanp-sonar: I've debugged what happens when binding for the first time. In that case, `this.host.VisualStateManager.BoundProjectKey` is `null`. However, this doesn't cause an exception since the web service call handles null in that case.
Technically, it's a bug, but it's not a new bug so we don't need worry about it now. I'll create a separate ticket for it.

FYI the impact isn't serious in any case. The code fetches the properties so it can get the regex to use for test projects - that's the only one it uses. The bug is that it will fetch the setting defined for the whole SQ instance, rather than for a specific SQ project. It just means that test projects in the solution might not be correctly identified if the user has set a different test regex for the SQ project being bound against.

duncanp-sonar: Update: the property is set correctly when re-opening a bound solution.

PR #860, opened 2019-04-02
Fixed #859

### LINKED ISSUE #859: Connected mode update incorrectly adds project-level rulesets
### Description

If a project references the generated ruleset (e.g. directly, via a custom ruleset, or via a a _Directory.Build.props_ file) but_ does not have a project-level ruleset, then updating a binding will incorrectly add a project-level ruleset.

Related to #790 _SLVS incorrectly complains that projects are not bound_

#790 is fixed so SLVS correctly detects that projects reference the generated ruleset when checking whether the bindings are up to date. However, it does not use this information when updating the bindings.

### Repro steps

1. Create a bound solution.

2. Manually edit the binding to remove the project-level rulsets, and instead set the _Ruleset_ property to directly to refer to the generated ruleset.

3. Close and re-open the solution

4. Update the binding from the SonarQube tab in the Team Explorer.

### Expected behavior

* Per-project rulesets are not added

### Actual behavior

* per-project rulesets are added

### Known workarounds

None

### Related information

* SonarLint for VS Version: v4.8
* Languages in your solution: C#, VB

duncanp-sonar:
* ruleset binding should not add project-level rulesets if the generated ruleset is already referenced

### DIFF
--- a/src/Integration/Binding/BindingController.cs
+++ b/src/Integration/Binding/BindingController.cs
@@ -116,23 +116,28 @@ namespace SonarLint.VisualStudio.Integration.Binding
 
         void IBindingWorkflowExecutor.BindProject(BindCommandArgs bindingArgs)
         {
-            var workflow = CreateBindingWorkflow(bindingArgs);
+            var workflow = CreateBindingWorkflow(host, bindingArgs);
 
             IProgressEvents progressEvents = workflow.Run();
             Debug.Assert(progressEvents != null, "BindingWorkflow.Run returned null");
             this.SetBindingInProgress(progressEvents, bindingArgs);
         }
 
-        internal /* for testing purposes */ IBindingWorkflow CreateBindingWorkflow(BindCommandArgs bindingArgs)
+        internal static /* for testing purposes */ IBindingWorkflow CreateBindingWorkflow(IHost host, BindCommandArgs bindingArgs)
         {
             //Choose the type of binding
-            var configProvider = this.host.GetService<IConfigurationProvider>();
+            var configProvider = host.GetService<IConfigurationProvider>();
             configProvider.AssertLocalServiceIsNotNull();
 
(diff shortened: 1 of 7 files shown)


src/Integration/Binding/BindingController.cs
```diff
@@ -116,23 +116,28 @@ private void OnBind(BindCommandArgs args)
...
-            var workflow = CreateBindingWorkflow(bindingArgs);
+            var workflow = CreateBindingWorkflow(host, bindingArgs);
 
             IProgressEvents progressEvents = workflow.Run();
             Debug.Assert(progressEvents != null, "BindingWorkflow.Run returned null");
             this.SetBindingInProgress(progressEvents, bindingArgs);
         }
 
-        internal /* for testing purposes */ IBindingWorkflow CreateBindingWorkflow(BindCommandArgs bindingArgs)
+        internal static /* for testing purposes */ IBindingWorkflow CreateBindingWorkflow(IHost host, BindCommandArgs bindingArgs)
```
duncanp-sonar: Making the method static makes writing unit tests easier - it avoids having to configure the set of services required by the constructor.

src/TestInfrastructure/Framework/TestLogger.cs
```diff
@@ -46,6 +46,17 @@ public void AssertPartialOutputStrings(params string[] orderedPartialOutputMessa
...
+        {
+            this.outputStrings.Should().Contain(expected + Environment.NewLine); // All messages are postfixed by a newline
+        }
+
+        public void AssertPartialOutputStringExists(params string[] expected)
+        {
+            this.outputStrings.Should()
+                .Contain(msg => expected.All(partial => msg.Contains(partial)));
+        }
+
```
duncanp-sonar: Additional methods to help check for the existence of a specific full or partial string. The existing helper methods check the complete set of logged messages, which is quite fragile.

Issue #818, opened 2019-02-28
[VS2019 RC/Preview 4] Analyzing additional languages crashes VS


duncanp-sonar:
### Description

VS crashes when analysis of additional languages (i.e. C++ and JavaScript) is enabled

### Repro steps

Please provide the steps required to reproduce the problem

1. Enable the analysis of additional languages

2. Restart VS

After a short time VS crashes.

### Workaround
1. Manually uninstall the extension by running the following command from a VS command prompt:
`vsixinstaller /u:SonarLint.36871a7b-4853-481f-bb52-1835a874e81b`

2. Re-install the extension and don't enable support for additional languages

### Error
```
$exception	{"Could not load file or assembly 'System.Interactive.Async, Version=3.0.1000.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263' or one of its dependencies. The system cannot find the file specified.":"System.Interactive.Async, Version=3.0.1000.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263"}	System.IO.FileNotFoundException
```

Looks to be related to this issue with Grpc: https://github.com/grpc/grpc/issues/16244

### Fusion log
```
=== Pre-bind state information ===
LOG: DisplayName = System.Interactive.Async, Version=3.0.1000.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263
 (Fully-specified)
LOG: Appbase = file:///C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/common7/ide/
LOG: Initial PrivatePath = NULL
Calling assembly : Grpc.Core, Version=1.0.0.0, Culture=neutral, PublicKeyToken=d754f35622e28bad.
===
LOG: This bind starts in LoadFrom load context.
WRN: Native image will not be probed in LoadFrom context. Native image will only be probed in default load context, like with Assembly.Load().
LOG: Using application configuration file: C:\Users\xxx\AppData\Local\Microsoft\VisualStudio\16.0_c1af37ef\devenv.exe.config
LOG: Using host configuration file: 
LOG: Using machine configuration file from C:\Windows\Microsoft.NET\Framework\v4.0.30319\config\machine.config.
LOG: Post-policy reference: System.Interactive.Async, Version=3.0.1000.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263
[...]

PR #861, opened 2019-04-02
Improve daemon error handling

duncanp-sonar:
Relates to #818

### DIFF
--- a/src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
@@ -34,18 +34,19 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         private string tempPath;
         private string storagePath;
         private VSIX.SonarLintDaemon daemon;
+        private TestLogger logger;
 
         [TestInitialize]
         public void SetUp()
         {
             ISonarLintSettings settings = new Mock<ISonarLintSettings>().Object;
-            ILogger output = new Mock<ILogger>().Object;
+            logger = new TestLogger();
 
             tempPath = Path.Combine(Path.GetRandomFileName());
             storagePath = Path.Combine(Path.GetRandomFileName());
             Directory.CreateDirectory(tempPath);
             Directory.CreateDirectory(storagePath);
-            daemon = new VSIX.SonarLintDaemon(settings, output, VSIX.SonarLintDaemon.daemonVersion, storagePath, tempPath);
+            daemon = new VSIX.SonarLintDaemon(settings, logger, VSIX.SonarLintDaemon.daemonVersion, storagePath, tempPath);
         }
 
         [TestMethod]
@@ -109,6 +110,30 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             var siaAsm = AssemblyHelper.GetVersionOfReferencedAssembly(typeof(VSIX.SonarLintDaemon), "System.Interactive.Async");
             siaAsm.Should().Be(new Version("3.0.1000.0"),
                 "SonarLint is not using the version of System.Interactive.Async expected by Grpc.Core. This will cause a runtime error.");
+        }
+
+        [TestMethod]
+        public void SafeOperation_NonCriticalException()
+        {
+            // Act
+            daemon.SafeOperation(() => throw new InvalidCastException("YYY"));
+
+            // Assert
+            logger.AssertPartialOutputStringExists("System.InvalidCastException", "YYY");
+        }
+
+        [TestMethod]
+        public void SafeOperation_CriticalExceptionsAreNotCaught()
+        {
+            // Arrange
+            Action op = () =>
+            {
+                daemon.SafeOperation(() => throw new StackOverflowException());
+            };
+
+            // Act and assert
+            op.Should().ThrowExactly<StackOverflowException>();
+            logger.AssertNoOutputMessages();
         }
 
         [TestCleanup]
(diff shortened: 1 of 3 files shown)


src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -131,23 +130,24 @@ public void Start()
             process.OutputDataReceived += (sender, args) =>
             {
                 string data = args.Data;
-                if (data != null)
+                if (data != null && data.Contains("Server started"))
                 {
-                    if (data.Contains("Server started"))
```
duncanp-sonar: Miscellaneous: combined `if` conditions.

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -131,23 +130,24 @@ public void Start()
...
                         {
                             CreateChannelAndStreamLogs();
                             Ready?.Invoke(this, EventArgs.Empty);
-                        }
-                        catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
-                        {
-                            logger.WriteLine($"Daemon error: {ex.ToString()}");
-                        }
-                    }
-                    if (IsVerbose())
```
duncanp-sonar: I moved this logging statement before the error handling block so that the logged output appears in time-order.

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -131,23 +130,24 @@ public void Start()
...
+                            succeeded = true;
+                        });
+
+                    if (!succeeded)
                     {
-                        WritelnToPane(data);
+                        SafeOperation(() => this.Stop());
```
duncanp-sonar: If we don't stop the daemon, our code will continue to run inside VS whenever a user opens a file, leading to further errors. So, we might as well explicitly stop the daemon.

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -170,10 +170,11 @@ public void Start()
                 process.BeginErrorReadLine();
                 logger.WriteLine(Strings.Daemon_Started);
             }
-            catch (Exception e)
+            catch (Exception e) when (!ErrorHandler.IsCriticalException(e))
```
duncanp-sonar: Miscellaneous: standard VS pattern - don't catch exceptions VS regards as critical.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -140,6 +140,23 @@ internal sealed class TaggerProvider : IViewTaggerProvider, ITableDataSource
         #endregion
 
         public void RequestAnalysis(string path, string charset, IEnumerable<SonarLanguage> detectedLanguages, IIssueConsumer issueConsumer, ProjectItem projectItem)
+        {
+            // Called on the UI thread -> unhandled exceptions will crash VS
```
duncanp-sonar: This was another public entry point that could crash VS in the event of errors.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -140,6 +140,23 @@ internal sealed class TaggerProvider : IViewTaggerProvider, ITableDataSource
...
+            try
+            {
+                UnsafeRequestAnalysis(path, charset, detectedLanguages, issueConsumer, projectItem);
+            }
+            catch (System.IO.FileNotFoundException fnf)
```
duncanp-sonar: Oops, meant to delete this code.

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
andrei-epure-sonarsource: this introduces a change of behavior: previously, `WritelnToPane(data);` would have happened even if `!data.Contains("Server started")`, but now it will happen only if it contains `Server started` - is this intended?

src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
```diff
@@ -34,18 +34,19 @@ public class SonarLintDaemonTests
         private string tempPath;
         private string storagePath;
         private VSIX.SonarLintDaemon daemon;
+        private TestLogger logger;
 
         [TestInitialize]
         public void SetUp()
```
andrei-epure-sonarsource: I suggest to add tests for the "Verbose" logging - with and without "Server started" inside the `OutputDataReceived`

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
andrei-epure-sonarsource: can it be tested?

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
duncanp-sonar: Good point; no, it wasn't. I'll separate the `if`s.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
duncanp-sonar: Not easily, no. It would require too much refactoring, and I don't want to do it at this point.

src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
duncanp-sonar: Unfortunately, that can't be done as a unit test without substantial refactoring either - too risky this close to shipping.

The daemon code is also on my list to clean up, but it's not at the top :-(.

PR #864, opened 2019-04-03
Added daemon unit tests

duncanp-sonar:


src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -113,59 +98,29 @@ public void Start()
...
-            {
-                string data = args.Data;
-                if (data != null)
-                {
-                    WritelnToPane(data);
-                }
-            };
+            process.OutputDataReceived += (sender, args) => HandleOutputDataReceived(args?.Data);
+            process.ErrorDataReceived += (sender, args) => HandleErrorDataReceived(args?.Data);
+
```
duncanp-sonar: I moved the event handlers to be `internal` methods so they can be tested.

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -441,5 +435,42 @@ internal void SafeOperation(Action op)
                 logger.WriteLine($"Daemon error: {ex.ToString()}");
             }
         }
+
+        #region IDisposable Support
```
duncanp-sonar: The Dispose pattern was implemented correctly before because the class was `sealed`.
However, it's no longer sealed to make it more testable, so we now need a `protected virtual Dispose` method.

This is the standard pattern implementation applied by VS.

src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
```diff
@@ -33,65 +32,81 @@ public class SonarLintDaemonTests
...
-            daemon.Start();
-            daemon.IsInstalled.Should().BeTrue();
-            daemon.IsRunning.Should().BeTrue();
-            daemon.Stop();
-            daemon.IsRunning.Should().BeFalse();
+            testableDaemon.Start();
+            testableDaemon.IsInstalled.Should().BeTrue();
+            testableDaemon.IsRunning.Should().BeTrue();
+            testableDaemon.Stop();
+            testableDaemon.IsRunning.Should().BeFalse();
```
andrei-epure-sonarsource: maybe also check that `testableDaemon.IsInstalled.Should().BeTrue();` is still the case after stopping?

src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
```diff
@@ -128,19 +143,146 @@ public void SafeOperation_CriticalExceptionsAreNotCaught()
...
+            // Assert
+            logger.AssertOutputStringExists("Server started");
+            testableDaemon.CreateChannelCallCount.Should().Be(1);
+            testableDaemon.WasReadyEventInvoked.Should().BeFalse();
+            logger.AssertPartialOutputStringExists("System.InvalidCastException");
+            testableDaemon.WasStopCalled.Should().BeTrue();
+        }
+
+        [TestMethod]
+        public void HandleOutputDataReceived_Data_ServerStarted_CriticalException_StopIsCalled()
```
andrei-epure-sonarsource:
```suggestion
        public void HandleOutputDataReceived_Data_ServerStarted_CriticalException_StopIsNotCalled()
```

src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
```diff
@@ -155,5 +297,34 @@ private static void ForceDeleteDirectory(string path)
                 Directory.Delete(path, true);
             }
         }
+
+        private class TestableSonarLintDaemon : VSIX.SonarLintDaemon
```
andrei-epure-sonarsource: nice! 👍

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -48,11 +48,9 @@ internal sealed class SonarLintDaemon : ISonarLintDaemon
         private readonly string version;
         private readonly string tmpPath;
         private readonly string storagePath;
-        private int port;
-
         private Process process;
 
-        private readonly string workingDirectory;
+        public string WorkingDirectory { get; }
```
andrei-epure-sonarsource: should this be `internal` instead? or is this immutable?

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -441,5 +435,42 @@ internal void SafeOperation(Action op)
...
         }
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                if (disposing)
```
andrei-epure-sonarsource: shouldn't `if (disposing)` be the outer if? because now, if `!disposedValue` and `!disposing`, `disposedValue` will be set to `true` - shouldn't it happen only if `disposing`?

src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
duncanp-sonar: This test is ignored and has been for a long time. I haven't tried to fix it as part of this PR.

andrei-epure-sonarsource: I missed that. And you added more test statements for documentation purposes? If the functionality is broken, do we need a ticket for fixing it?

duncanp-sonar: The only changes in this test are as a result of renaming `daemon` to `testableDaemon`.

I doubt this test has ever worked. It's also not a unit test, as it's actually attempting to download the daemon from the web and start running it. I've deleted it.

PR #865, opened 2019-04-03
Added more daemon unit tests

duncanp-sonar:


src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -242,28 +242,42 @@ private void WritelnToPane(string msg)
...
+                process = null;
+            });
+
             logger.WriteLine(Strings.Daemon_Stopped);
         }
 
         public int Port { get; private set; }
 
-        public bool IsInstalled => Directory.Exists(InstallationPath) && File.Exists(ExePath);
+        public bool IsInstalled => File.Exists(ExePath);
```
duncanp-sonar: Removed the redundant directory check (the exe is always under the installation path).

src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
```diff
@@ -58,24 +58,86 @@ public void SetUp()
         [TestCleanup]
         public void CleanUp()
         {
+            testableDaemon.Dispose();
```
duncanp-sonar: Additional test cleanup

src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
```diff
@@ -58,24 +58,86 @@ public void SetUp()
...
+
+        [TestMethod]
+        public void Run_With_Install_Succeeds()
+        {
+            testableDaemon.Port.Should().Be(0);
+
+            // Fake the installation i.e. create the directory and file to execute.
+            testableDaemon.SetUpDummyInstallation("test1.bat",
+@"@echo Hello world
+xxx yyy
```
duncanp-sonar: We can't launch the real daemon from the tests, so I've added some override points to allow the tests to substitute an alternative batch file to run instead. That allows us to check that the process is launched, and also that the standard and error outputs are captured correctly.

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -242,28 +242,42 @@ private void WritelnToPane(string msg)
...
-        public virtual /* for testing */ void Stop()
+        public void Stop()
         {
             if (!IsRunning)
             {
                 return;
                 // throw exception?
             }
 
+            this.SafeInternalStop();
```
duncanp-sonar: We need to be able to do cleanup internally in cases where the process failed to start correctly i.e. when `IsRunning` might be return false. I moved the internal cleanup into a separate method.

src/TestInfrastructure/Framework/TestLogger.cs
```diff
@@ -72,11 +87,15 @@ public void Reset()
...
+            if (logToConsole)
+            {
+                Console.WriteLine(message);
+            }
         }
 
         public void WriteLine(string messageFormat, params object[] args)
         {
-            outputStrings.Add(string.Format(System.Globalization.CultureInfo.CurrentCulture, messageFormat, args) + Environment.NewLine);
+            WriteLine(string.Format(System.Globalization.CultureInfo.CurrentCulture, messageFormat, args));
```
andrei-epure-sonarsource: shouldn't this check `logToConsole` value before?

if not, then why is it checked at line 90?

src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
```diff
@@ -58,24 +58,86 @@ public void SetUp()
...
+            // Fake the installation i.e. create the directory and file to execute.
+            testableDaemon.SetUpDummyInstallation("test1.bat",
+@"@echo Hello world
+@echo write to error stream... 1>&2
+");
+
+            // Act
+            testableDaemon.Start();
+            bool processFinished = testableDaemon.process.WaitForExit(1000);
+            processFinished.Should().BeTrue("Test execution error: timed out waiting for the dummy process to exit");
```
andrei-epure-sonarsource: strictly speaking, line 110 should be after `//Asert` comment

duncanp-sonar: For me, line 110 is a sanity check on the execution of the test, and I put those as close to the relevant bit of test code as possible.
I don't think of them in the same was as the checks in the `// Assert` section, which are checks on the product code.

src/TestInfrastructure/Framework/TestLogger.cs
duncanp-sonar: This overload with parameters just formats the message and then calls the other overload, which actually does the recording and writing to the console. `logToConsole` doesn't need to be checked in both places.

PR #895, opened 2019-06-18
Workaround for hang when binding .NET SDK projects in VS2019

### LINKED ISSUE #881: VS 2019 Freezes when connecting to Sonar and binding projects
### Description

Utilizing SonarLint repeatedly freezes VS 2019 during binding operations

### Repro steps

1. Go to the SonarQube window in Team Explorer
2. Find the list of projects 
3. Select the project to which you would like to bind your solution
4. This freezes VS 2019, and it has to be force-restarted. 
5. When VS opens again, most or all projects seem to have bindings associated with it. 
6. Right click on the connected project and click "update". This will freeze VS 2019 again.

### Related information

* SonarLint for VS Version (latest)
* Visual Studio Version (VS 2019 Enterprise, latest)
* SonarQube server version (not sure)
* Languages in your solution (C# and JavaScript)

duncanp-sonar:
Fix #881

### DIFF
--- a/src/Integration.UnitTests/Binding/ProjectBindingOperationTests.cs
+++ b/src/Integration.UnitTests/Binding/ProjectBindingOperationTests.cs
@@ -292,7 +292,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
         }
 
         [TestMethod]
-        public void ProjectBindingOperation_Commit()
+        public void ProjectBindingOperation_Commit_NewProjectSystem_AddsFile()
         {
             // Arrange
             ProjectBindingOperation testSubject = this.CreateTestSubject();
@@ -301,6 +301,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
             PropertyMock prop = CreateProperty(this.projectMock, "config1", ProjectBindingOperation.DefaultProjectRuleSet);
             testSubject.Initialize();
             testSubject.Prepare(CancellationToken.None);
+
+            this.projectSystemHelper.SetIsLegacyProjectSystem(false);
 
             // Act
             using (new AssertIgnoreScope()) // Ignore that the file is not on disk
@@ -311,7 +313,32 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
             // Assert
             string expectedFile = Path.Combine(Path.GetDirectoryName(this.projectMock.FilePath), Path.GetFileNameWithoutExtension(this.projectMock.FilePath) + ".ruleset");
             prop.Value.ToString().Should().Be(Path.GetFileName(expectedFile), "Should update the property value");
-            this.projectMock.Files.ContainsKey(expectedFile).Should().BeTrue("Should be added to the project");
+            this.projectMock.Files.ContainsKey(expectedFile).Should().BeFalse("Should not add the file to the project for the new project system");
+        }
+
+        [TestMethod]
+        public void ProjectBindingOperation_Commit_LegacyProjectSystem_DoesNotAddFile()
+        {
+            // Arrange
+            ProjectBindingOperation testSubject = this.CreateTestSubject();
+            this.projectMock.SetCSProjectKind();
+            this.ruleStore.RegisterRuleSetPath(Language.CSharp, @"c:\Solution\sln.ruleset");
+            PropertyMock prop = CreateProperty(this.projectMock, "config1", ProjectBindingOperation.DefaultProjectRuleSet);
+            testSubject.Initialize();
+            testSubject.Prepare(CancellationToken.None);
+
+            this.projectSystemHelper.SetIsLegacyProjectSystem(true);
+
+            // Act
+            using (new AssertIgnoreScope()) // Ignore that the file is not on disk
+            {
+                testSubject.Commit();
+            }
+
+            // Assert
+            string projectFile = Path.Combine(Path.GetDirectoryName(this.projectMock.FilePath), Path.GetFileNameWithoutExtension(this.projectMock.FilePath) + ".ruleset");
+            prop.Value.ToString().Should().Be(Path.GetFileName(projectFile), "Should update the property value");
+            this.projectMock.Files.ContainsKey(projectFile).Should().BeTrue("Should add the file to the project for the legacy project system");
         }
 
         #endregion Tests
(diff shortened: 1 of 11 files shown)


src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
```diff
@@ -618,6 +618,32 @@ public void ProjectSystemHelper_IsSolutionFullyLoaded_PropertyIsNotBoolean_Retur
...
+        public void ProjectSystemHelper_IsLegacyProject_Legacy_ReturnsTrue()
+        {
+            // Arrange
+            var mockProject  = this.solutionMock.AddOrGetProject("dummy.proj", isLegacy: true);
+
+            // Act
+            var result = this.testSubject.IsLegacyProjectSystem(mockProject);
+
+            // Assert
+            result.Should().BeTrue();
```
duncanp-sonar: Unit tests for the new behaviour - don't add the ruleset file to the project for the new .NET-SDK project system.

src/TestInfrastructure/VsObjectModelMocks/LegacyProjectMock.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Linq;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell.Flavor;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests
+{
+    /// <summary>
+    /// Mock for managed projects that use the legacy C#/VB project system
+    /// </summary>
+    public class LegacyProjectMock : ProjectMock, IVsAggregatableProjectCorrected
```
duncanp-sonar: I extracted this class from the existing `ProjectMock` class - no other changes.
The legacy mock implements the `IVsAggregatableProjectCorrected` interface that we use to detect which project system is being used.

src/TestInfrastructure/VsObjectModelMocks/SolutionMock.cs
```diff
@@ -515,12 +515,15 @@ public IEnumerable<ProjectMock> Projects
             }
         }
 
-        public ProjectMock AddOrGetProject(string projectFile, bool isLoaded = true)
+        public ProjectMock AddOrGetProject(string projectFile, bool isLoaded = true, bool isLegacy = true)
```
duncanp-sonar: By default assume we're using using the legacy project system (to minimise the amount of churn required to the existing unit tests).

duncanp-sonar:
I suggest reviewing the commit separately.
The second commit is mainly a refactoring of an existing test mock to make it possible to write tests to bump up the code coverage.

src/Integration/LocalServices/IProjectSystemHelper.cs
```diff
@@ -132,5 +132,12 @@ internal interface IProjectSystemHelper : ILocalService
         /// Returns a flag indicating whether there is fully-opened solution or not.
         /// </summary>
         bool IsSolutionFullyOpened();
+
+        /// <summary>
+        /// Returns true if the project is implemented by the old C#/VB project system, otherwise false
```
andrei-epure-sonarsource: nitpick: would be useful to say what 'old' means - 2 years from now it may mean something else

src/Integration/LocalServices/IProjectSystemHelper.cs
```diff
@@ -132,5 +132,12 @@ internal interface IProjectSystemHelper : ILocalService
...
+        /// </summary>
+        /// <returns>There are occasions when we need to code round differences between the old and new
```
andrei-epure-sonarsource:
```suggestion
        /// <returns>There are occasions when we need to code around differences between the old and new
```

src/Integration/Binding/ProjectBindingOperation.cs
```diff
@@ -172,7 +172,10 @@ private void AddFileToProject(Project project, string fullFilePath)
             var projectSystem = this.serviceProvider.GetService<IProjectSystemHelper>();
             projectSystem.AssertLocalServiceIsNotNull();
 
-            if (!projectSystem.IsFileInProject(project, fullFilePath))
+            // Workaround for bug https://github.com/SonarSource/sonarlint-visualstudio/issues/881
```
andrei-epure-sonarsource: it would be nice to mention that this file does not need to be added to the project in the 2019 system type

src/TestInfrastructure/VsObjectModelMocks/ProjectMock.cs
```diff
@@ -24,16 +24,14 @@
 using System.Linq;
 using EnvDTE;
 using Microsoft.VisualStudio;
-using Microsoft.VisualStudio.Shell.Flavor;
 using Microsoft.VisualStudio.Shell.Interop;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests
 {
-    public class ProjectMock : VsUIHierarchyMock, IVsProject, Project, IVsBuildPropertyStorage, IVsAggregatableProjectCorrected
+    public class ProjectMock : VsUIHierarchyMock, IVsProject, Project, IVsBuildPropertyStorage
```
andrei-epure-sonarsource: would be nice to say that this a 'new' project mock because it does not have `IVsAggregatableProjectCorrected` ?

src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
andrei-epure-sonarsource: but this behaviour (not adding the file) is not checked in this file, only the legacy project detection... so why this comment here?

src/Integration.UnitTests/Binding/ProjectBindingOperationTests.cs
```diff
@@ -311,7 +313,32 @@ public void ProjectBindingOperation_Commit()
             // Assert
             string expectedFile = Path.Combine(Path.GetDirectoryName(this.projectMock.FilePath), Path.GetFileNameWithoutExtension(this.projectMock.FilePath) + ".ruleset");
             prop.Value.ToString().Should().Be(Path.GetFileName(expectedFile), "Should update the property value");
-            this.projectMock.Files.ContainsKey(expectedFile).Should().BeTrue("Should be added to the project");
+            this.projectMock.Files.ContainsKey(expectedFile).Should().BeFalse("Should not add the file to the project for the new project system");
+        }
+
+        [TestMethod]
+        public void ProjectBindingOperation_Commit_LegacyProjectSystem_DoesNotAddFile()
```
andrei-epure-sonarsource:
```suggestion
        public void ProjectBindingOperation_Commit_LegacyProjectSystem_AddsFile()
```

src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
duncanp-sonar: My mistake - these tests are just for `IsLegacyProjectSystem`. The add/don't add tests are elsewhere.

PR #905, opened 2019-07-08
Added environment variable to specify daemon download URL

### LINKED ISSUE #904: Allow alternative download location for additional languages daemon to be specified
### Description
Currently the download location of daemon that provides support for additional languages is hard-coded to the public maven repository. This is problematic if the user does not have internet access - see #558.

### Changes
The download location can be specified in the environment variable `SONARLINT_DAEMON_DOWNLOAD_URL`.

The URL must be absolute. The daemon zip file must not be renamed i.e. it must in the format `sonarlint-daemon-{version}-windows.zip`.

If the supplied URL is invalid the default location will be used.

duncanp-sonar:
Fixes #904

### DIFF
--- a/src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
@@ -51,7 +51,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             storagePath = Path.Combine(TestContext.DeploymentDirectory, Path.GetRandomFileName());
             Directory.CreateDirectory(tempPath);
             Directory.CreateDirectory(storagePath);
-            testableDaemon = new TestableSonarLintDaemon(settings, logger, VSIX.SonarLintDaemon.daemonVersion, storagePath, tempPath);
+            testableDaemon = new TestableSonarLintDaemon(settings, logger, storagePath, tempPath);
+
+            logger.Reset(); // clear any messages logged during construction
         }
 
         [TestCleanup]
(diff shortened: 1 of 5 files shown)


src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -40,17 +41,20 @@ internal class SonarLintDaemon : ISonarLintDaemon
         private static readonly string DAEMON_HOST = "localhost";
         private static readonly int DEFAULT_DAEMON_PORT = 8050;
 
-        public const string daemonVersion = "4.0.2.2142";
-        private const string uriFormat = "http://repo1.maven.org/maven2/org/sonarsource/sonarlint/core/sonarlint-daemon/{0}/sonarlint-daemon-{0}-windows.zip";
+        public const string DefaultDaemonVersion = "4.0.2.2142";
+        public const string SonarLint_DownloadUrl_EnvVar = "SONARLINT_DAEMON_DOWNLOAD_URL";
+        public static readonly string DefaultMavenUrl = string.Format("http://repo1.maven.org/maven2/org/sonarsource/sonarlint/core/sonarlint-daemon/{0}/sonarlint-daemon-{0}-windows.zip",
```
andrei-epure-sonarsource: nitpick: this is the default version for `SonarLint_DownloadUrl_EnvVar`, so it might be better to have a name like `DefaultSonarLintDownloadUrl` ?

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -283,11 +293,67 @@ protected virtual void SafeInternalStop()
 
         public bool IsInstalled => File.Exists(ExePath);
 
+        private static string ExtractVersionStringFromUrl(string url)
+        {
+            // We're loosely match the "version" part of the zip file name
```
andrei-epure-sonarsource: grammar nitpick: `We're loosely matching` or `We'll loosely match`? (unless I'm missing something)

src/Integration.Vsix/Resources/Strings.resx
```diff
@@ -151,6 +155,14 @@
   <data name="Daemon_Installing" xml:space="preserve">
     <value>Installing the daemon...</value>
   </data>
+  <data name="Daemon_InvalidFileNameInDownloadEnvVar" xml:space="preserve">
+    <value>The daemon file name supplied in {0} environment variable is invalid ('{1}'). Using the default download URL.</value>
+    <comment>Output window message</comment>
+  </data>
+  <data name="Daemon_InvalidUrlInDownloadEnvVar" xml:space="preserve">
+    <value>Download URL supplied in {0} environment variable is invalid ('{1}'). Using default download URL.</value>
```
andrei-epure-sonarsource: is it worth logging the default download URL as well? (applies for above message, too)

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -40,17 +41,20 @@ internal class SonarLintDaemon : ISonarLintDaemon
         private static readonly string DAEMON_HOST = "localhost";
         private static readonly int DEFAULT_DAEMON_PORT = 8050;
 
-        public const string daemonVersion = "4.0.2.2142";
-        private const string uriFormat = "http://repo1.maven.org/maven2/org/sonarsource/sonarlint/core/sonarlint-daemon/{0}/sonarlint-daemon-{0}-windows.zip";
+        public const string DefaultDaemonVersion = "4.0.2.2142";
```
andrei-epure-sonarsource: naming nitpick: the constants above are UPPER_CASE, the 2 constants below have different casing - should the naming be more uniform?

also: do these need to be public or only internal for tests?

src/Integration.Vsix.UnitTests/SonarLintDaemon/SonarLintDaemonTests.cs
```diff
@@ -352,6 +354,73 @@ public void Dispose_WorkingDirectoryDeleted()
             testableDaemon.Dispose();
         }
 
+        [TestMethod]
+        public void DefaultVersionAndDownloadUrl()
```
andrei-epure-sonarsource: maybe mention in the name that the env var is not set (e.g. `DownloadUrlInEnvironmentVar_NotSet_UseDefault` )

src/Integration.Vsix/Resources/Strings.resx
duncanp-sonar: The actual version and url that end up being used are logged by the constructor, so that information will be in the output.

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
duncanp-sonar: The other constants in the project are mixed case. I'll change the ones in lines 41 and 42.

The class is internal so the effective scope is internal. However, I'll change them to internal since they are only exposed for testing.

PR #913, opened 2019-07-11
Updated organization UI to separate selection of own/public orgs

### LINKED ISSUE #446: Display the organizations the user is member of
When configuring a connection to SonarCloud, SonarLint has to retrieve the whole list of organizations to give the possibility to the user to select the one she's interested in.
As a result, a huge number of organizations are listed which don't make sense for the user. And this operation takes quite a long time.
We should first retrieve and propose the organizations the user is member of and keep the ability to connect to another organization as a fallback option.

Depends upon [SONAR-9000](https://jira.sonarsource.com/browse/SONAR-9000 ).

duncanp-sonar:
Fixed #446

### DIFF
--- a/src/Integration.UnitTests/WPF/IsValidOrganizationKeyConverterTests.cs
+++ b/src/Integration.UnitTests/WPF/IsValidOrganizationKeyConverterTests.cs
@@ -1,0 +1,102 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2018 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Globalization;
+using System.Windows;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.Integration.Connection.UI;
+using SonarLint.VisualStudio.Integration.WPF;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.WPF
+{
+    [TestClass]
+    public class IsValidOrganizationKeyConverterTests
+    {
+        [TestMethod]
+        public void Convert_NonBoolTargetType_ThrowsArgumentException()
+        {
+            var converter = new IsValidOrganisationKeyConverter();
+
+            Action act = () => converter.Convert("valid input", typeof(object), null, CultureInfo.CurrentCulture);
+
+            act.Should().ThrowExactly<ArgumentException>().And.ParamName.Should().Be("targetType");
+        }
+
+        [TestMethod]
+        public void Convert_NonTextInput_ThrowsArgumentException()
+        {
+            var converter = new IsValidOrganisationKeyConverter();
+
+            var notAString = new object();
+            Action act = () => converter.Convert(notAString, typeof(bool), null, CultureInfo.CurrentCulture);
+
+            act.Should().ThrowExactly<ArgumentException>().And.ParamName.Should().Be("value");
+        }
+
+        [TestMethod]
+        public void Convert_NullInputString_ReturnsFalse()
+        {
+            var converter = new IsValidOrganisationKeyConverter();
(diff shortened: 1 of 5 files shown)


src/Integration.UnitTests/WPF/IsValidOrganizationKeyConverterTests.cs
```diff
@@ -0,0 +1,102 @@
...
+using SonarLint.VisualStudio.Integration.Connection.UI;
+using SonarLint.VisualStudio.Integration.WPF;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.WPF
+{
+    [TestClass]
+    public class IsValidOrganizationKeyConverterTests
+    {
+        [TestMethod]
+        public void Convert_NonBoolTargetType_ThrowsArgumentException()
```
duncanp-sonar: Unit tests for the new class that takes a string and returns whether it is a valid organization key.

src/Integration/Connection/UI/OrganizationSelectionWindow.xaml
```diff
@@ -41,46 +42,98 @@
...
-                               IsEnabled="{Binding ElementName=OrganizationComboBox,
-                                                   Path=SelectedItem,
-                                                   Converter={StaticResource NotNullConverter}}" />
-            <vsui:DialogButton Content="Cancel" IsCancel="True" />
-        </StackPanel>
-    </StackPanel>
+    <!-- Note: we're using grids inside the TabItem with IsSharedSizeScope to make sure the tabs are the same size
+         (otherwise they will be different and the dialogue box will resize when the user switches tabs) -->
+    <TabControl Grid.IsSharedSizeScope="True" >
+        <TabItem Header="Your organizations">
```
duncanp-sonar: Previously, the UI consisted of a combobox bound to the list of organizations fetched from the server.
The changes to the webclient submodule (above) limited the list of organizations to just the orgs the user belongs to, rather than all orgs with public projects.
However, that meant there was no way using the old dialogue for the user to select an organization they don't belong to. 

To fix this, I've added another tab to the UI to allow the user to type the organization key into a textbox. The UI is isn't very polished but it's good enough for this bug fix, and we don't want to spend more time on this at present.

src/Integration/Connection/UI/OrganizationSelectionWindow.xaml
```diff
@@ -41,46 +42,98 @@
...
+            <Grid>
+                <Grid.RowDefinitions>
+                    <RowDefinition SharedSizeGroup="tabItemSize" />
+                </Grid.RowDefinitions>
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition SharedSizeGroup="tabItemHeight" />
+                </Grid.ColumnDefinitions>
+
+                <StackPanel
```
duncanp-sonar: This contents of this `StackPanel` are essentially the old UI with the combobox. I haven't changed it at all; all I've done is move it inside a tab labelled `Your organizations`.

src/Integration/Connection/UI/OrganizationSelectionWindow.xaml
```diff
@@ -41,46 +42,98 @@
...
+                                IsEnabled="{Binding ElementName=OrganizationComboBox,
+                                                    Path=SelectedItem,
+                                                    Converter={StaticResource NotNullConverter}}" />
+                        <vsui:DialogButton Content="Cancel" IsCancel="True" />
+                    </StackPanel>
+                </StackPanel>
+
+            </Grid>
+        </TabItem>
+        <TabItem Header="Other organizations">
```
duncanp-sonar: This is the new tab containing a textbox for the use to type in a organization key.

src/Integration/Connection/UI/OrganizationSelectionWindow.xaml
```diff
@@ -41,46 +42,98 @@
...
+                        </TextBox>
+                    </StackPanel>
+
+                    <Separator Style="{StaticResource DialogSeparator}" />
+
+                    <StackPanel VerticalAlignment="Center" Style="{StaticResource DialogButtonsPanel}">
+                        <vsui:DialogButton Click="OnOtherOrgOkButtonClick" Content="OK" IsDefault="True"
+                                IsEnabled="{Binding ElementName=OrganizationKeyTextBox,
+                                                    Path=Text,
+                                                    Converter={StaticResource IsValidOrgKeyConverter}}" />
```
duncanp-sonar: The `IsEnabled="{Binding... Converter={StaticResource IsValidOrgKeyConverter}}` is the WPF magic that takes care of enabling/disabling the `OK` button.

WPF will take care of calling the `IsValidOrgKeyConverter` each time the contents of the textbox containing the organization key change, and enable/disable the button based on the value returned by the converter.

src/Integration/Connection/UI/OrganizationSelectionWindow.xaml.cs
```diff
@@ -40,15 +43,29 @@ internal OrganizationSelectionWindow(IEnumerable<SonarQubeOrganization> organiza
             OrganizationComboBox.ItemsSource = sortedOrganizations;
         }
 
-        private void OnOkButtonClick(object sender, RoutedEventArgs e)
+        private void OnOwnOkButtonClick(object sender, RoutedEventArgs e)
```
duncanp-sonar: To keep things simple, I bound each `OK` button to a different handler method to fetch the data from the combobox/textbox as appropriate to work out which organization was selected.

PR #922, opened 2019-07-16
Initial version of the protocol

henryju:


src/Integration.Vsix/SonarLintDaemon/Protocol.cs
```diff
@@ -0,0 +1,231 @@
+﻿using System;
```
duncanp-sonar: All of the SLVS code files have copyright headers currently.

src/Integration.Vsix/SonarLintDaemon/Protocol.cs
```diff
@@ -0,0 +1,231 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.SonarLintDaemon
+{
+
+    class Request
```
duncanp-sonar: We're always explicit about the class scope. Here, `internal` would do i.e. only visible within this project, and to any projects we explicitly permit to see our internals.

src/Integration.Vsix/SonarLintDaemon/Protocol.cs
```diff
@@ -0,0 +1,231 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.SonarLintDaemon
```
duncanp-sonar: I was wondering whether to put the new C++ integration code in a separate folder/namespace to make it easy to see which bits are for the Java daemon and which bits are C++.
For now, I think it would be a good idea to put this class in a "cpp" subfolder.

src/Integration.Vsix/SonarLintDaemon/Protocol.cs
duncanp-sonar: If this class is straight port of an existing Java class it might be worth calling this out in a comment and include a link to the Java class in GitHub.

src/Integration.Vsix/SonarLintDaemon/Protocol.cs
```diff
@@ -0,0 +1,231 @@
...
+        public string ModuleName { get; set; } = "";
+        public string Predefines { get; set; } = "";
+        public string[] Macros { get; set; } = Array.Empty<string>();
+        public string TargetTriple { get; set; } = "x86_64-unknown-unknown";
+        public string File { get; set; } = "";
+    }
+
+    class Response
+    {
+        public Message[] Messages { get; set; }
```
duncanp-sonar: If this property is supposed to be read-only you can drop the `set; ` - the compiler is smart enough to work out that the property should only be set once from the constructor.

src/Integration.Vsix/SonarLintDaemon/Protocol.cs
```diff
@@ -0,0 +1,231 @@
...
+    {
+        public string Filename;
+        public int Line { get; }
+        public int Column { get; }
+        public int EndLine { get; }
+        public int EndColumn { get; }
+        public string Text { get; }
+
+        public MessagePart(
+      String filename,
```
duncanp-sonar: Weird indentation of the parameters.
If you delete then retype the final `}` in the document VS will reformat the indents (same as `Edit`, `Advanced`, `Format document` from the main menu).

src/Integration.Vsix/SonarLintDaemon/Protocol.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        /**
+           * This method does not close the provided stream.
+           */
+        public static Response read(BinaryReader reader)
+        {
+
+            if ("OUT" != readUTF(reader))
+            {
+                throw new IOException();
```
duncanp-sonar: There's a more specific exception type [`[System.IO.InvalidDataException]`](https://docs.microsoft.com/en-us/dotnet/api/system.io.invaliddataexception?view=netframework-4.6.1) for data streams that "are in an invalid format".

It would be also be worth adding a message in the constructor as this will end up in the output window in SLVS e.g. "stream does not contain output data"

src/Integration.Vsix/SonarLintDaemon/Protocol.cs
```diff
@@ -0,0 +1,231 @@
...
+                    int line = reader.ReadInt32();
+                    int column = reader.ReadInt32();
+                    int endLine = reader.ReadInt32();
+                    int endColumn = reader.ReadInt32();
+                }
+            }
+
+            if ("END" != readUTF(reader))
+            {
+                throw new IOException();
```
duncanp-sonar: Ditto System.IO.InvalidDataException("output stream terminator not found")

src/Integration.Vsix/SonarLintDaemon/Protocol.cs
```diff
@@ -0,0 +1,231 @@
...
+            write(writer, request.VfsOverlayFiles);
+            writeUTF(writer, request.ModuleName);
+            writeUTF(writer, request.Predefines);
+            write(writer, request.Macros);
+            writeUTF(writer, request.TargetTriple);
+            writeUTF(writer, request.File);
+            writeUTF(writer, "END");
+        }
+
+        private static void write(BinaryWriter writer, string[] strings)
```
duncanp-sonar: Method names are almost universally Pascal-cased in the C# world.

Issue #928, opened 2019-07-16
Implement the C++ analyzer protocol reader/writer


duncanp-sonar:
- [x] Implement the protocol binary/reader writer
- ~~Handle reading/writing Java modified-UTF8 strings~~
- [x] Unit tests

Out-of-scope:
- Integration tests to compare the generated files against the Java equivalent. To start with, we'll test this indirectly via higher level integration tests.

PR #942, opened 2019-07-19
Implement protocol to talk to the CFamily analyzer (#928)

henryju:
New PR limited to the protocol. This time I have implemented the (hopefully) correct encoding for numbers, and a good enough implementation for strings (we could fix NUL and supplementary characters if needed, assuming it is properly handled on C++ side).

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/ProtocolTest.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/ProtocolTest.cs
@@ -1,0 +1,347 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2018 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.IO;
+using EnvDTE;
+using FluentAssertions;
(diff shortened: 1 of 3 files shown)


src/Integration.Vsix.UnitTests/CFamily/ProtocolTest.cs
```diff
@@ -0,0 +1,197 @@
...
+            WriteUtf("0").Should().BeEquivalentTo(new byte[] { 0, 1, 48 });
+            WriteUtf("\n").Should().BeEquivalentTo(new byte[] { 0, 1, 10 });
+            // 3 bytes
+            WriteUtf("\u0800").Should().BeEquivalentTo(new byte[] { 0, 3, 224, 160, 128 });
+            // Special case of NUL
+            WriteUtf("\u0000").Should().BeEquivalentTo(new byte[] { 0, 2, 192, 128 });
+            // Supplementary characters as surrogate pair  (see CESU-8) are not supported for now
+            try
+            {
+                WriteUtf("\U00010400").Should().BeEquivalentTo(new byte[] { 0, 6, 237, 160, 129, 237, 176, 128 });
```
duncanp-sonar: Is the `WriteUtf` call expected to throw the `InvalidOperationException`? If so the `.Should().BeEquivalentTo(` is not needed and misleading.

src/Integration.Vsix.UnitTests/CFamily/ProtocolTest.cs
```diff
@@ -0,0 +1,197 @@
...
+                Assert.Fail();
+            }
+            catch (InvalidOperationException)
+            {
+                // Expected
+            }
```
duncanp-sonar: FYI the normal pattern used in SLVS tests for expected exceptions is this:

```
using FluentAssertions;
...

            Action act = () => WriteUtf("\U00010400");
            act.Should().ThrowExactly<InvalidOperationException>();
```

src/Integration.Vsix/CFamily/Protocol.cs
```diff
@@ -0,0 +1,328 @@
...
+            Write(writer, request.VfsOverlayFiles);
+            WriteUTF(writer, request.ModuleName);
+            WriteUTF(writer, request.Predefines);
+            Write(writer, request.Macros);
+            WriteUTF(writer, request.TargetTriple);
+            WriteUTF(writer, request.File);
+            WriteUTF(writer, "END");
+        }
+
+        public /*visible for testing*/ static void WriteLong(BinaryWriter writer, long l)
```
duncanp-sonar: The convention in this repo is `internal /* for testing */`.

As a rule of thumb, it's the only time a `internal` is used on a method.
If a class is only to be used internally then the class is marked `internal`, and any methods that are expected to be called by other types in the assembly are marked `public`.

PR #951, opened 2019-07-22
Custom build steps to fetch CFamily files and package in VSIX

### LINKED ISSUE #929: Modify build to acquire and package the correct C++ analyzer exe
#### Notes
We need to embed the following:
* the subprocess exe
* the json files specify which rules to run and the parameters for those rules
* ?a set of `.h` files? These files are only required for ObjectiveC analysis, so we should be able to avoid using them
* the following `.json` files: Sonar_way_profile.json, RulesList.json, and all of the XXX_params.json files

All of the above are embedded in the CFamily plugin jar, which is publicly available at https://binaries.sonarsource.com/CommercialDistribution/sonar-cfamily-plugin/.

The simplest solution will be to download the whole jar, unzip it, and extract the bits we need.
This would also mean that anyone forking this repo will still be able to build SLVS, even though the CFamily jar is commercially licensed. It also means that we don't need to worry about having to provide credentials in our internal builds to pull the bits from non-public locations.

To enable testing of intermediate builds, it would be useful if the download location and CFamily plugin version could be parameterised to allow developers to set them locally.

#### TODO
- [ ] modify the build to download the jar from https://binaries.sonarsource.com/CommercialDistribution/sonar-cfamily-plugin/
- [ ] package the jar and json files into the VSIX. Check the new VSIX size
- [ ] test on Azure Pipeline build
- [ ] check the increase in build time for PR/CI builds. If the build time increases too much then it might be necessary to skip the CFamily packaging step in PR builds.

duncanp-sonar:
Fixes #929

### DIFF
--- a/build/DownloadCFamilyPlugin/DownloadAndExtract.cs
+++ b/build/DownloadCFamilyPlugin/DownloadAndExtract.cs
@@ -49,15 +49,20 @@ namespace DownloadCFamilyPlugin
     {
         // The txz archive containing the subprocess.exe
         private const string WindowsTxzFilePattern = "clang-*-win.txz";
-        private const string CLangExeFilePattern = "SubProcess.exe";
 
-        private readonly string[] KnownJsonFiles = new string[]
+        // Sub-folder into which the tar file should be unzipped
+        private const string TarUnzipSubFolder = "tar_xz";
+
+        // List of patterns to match single files in the uncompressed output
+        private readonly string[] SingleFilePatterns = new string[]
         {
                 "Sonar_way_profile.json",
-                "RulesList.json"
+                "RulesList.json",
+                TarUnzipSubFolder + "\\SubProcess.exe"
         };
 
-        private readonly string[] AdditionalFilesToReturnPattern = new string[]
+        // List of patterns to match multiple files in the uncompressed output
+        private readonly string[] MultipleFilesPatterns = new string[]
         {
             "*_params.json"
         };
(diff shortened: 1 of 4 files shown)


build/DownloadCFamilyPlugin/DownloadAndExtract.cs
```diff
@@ -0,0 +1,266 @@
...
+    /// <remarks>
+    /// Assumptions:
+    /// * we are downloading the CFamily plugin jar e.g. https://binaries.sonarsource.com/CommercialDistribution/sonar-cfamily-plugin/sonar-cfamily-plugin-6.3.0.11371.jar
+    /// * the jar contains a .tarxz archive containing the Windows version of SubProcess.exe
+    /// * the jar contains various .json files, some with fixed names, some that we'll match using wildcards
+    /// 
+    /// We want the task to fail if it cannot locate all of the expected files (otherwise we'll build an invalid VSIX).
+    /// 
+    /// Downloading and extracting the files can be slow so we want to skip those steps if possible.
+    /// </remarks>
```
duncanp-sonar: There's quite a bit of processing required to handle fetching and embedding the files to I created a custom MSBuild task. This is in a new project under the existing `build` folder.

build/DownloadCFamilyPlugin/DownloadAndExtract.cs
```diff
@@ -0,0 +1,266 @@
...
+
+        private readonly string[] KnownJsonFiles = new string[]
+        {
+                "Sonar_way_profile.json",
+                "RulesList.json"
+        };
+
+        private readonly string[] AdditionalFilesToReturnPattern = new string[]
+        {
+            "S*_params.json"
```
henryju: Not all rules starts with 'S': https://github.com/SonarSource/sonar-cpp/tree/master/cpp-checks/src/main/resources/org/sonar/l10n/cpp/rules/params

build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net46</TargetFramework>
+    
+    <!-- Not a product project so we don't want it to be taken into account for code coverage -->
```
henryju: In the long run, we may even want to completely exclude it from SQ analysis.

henryju:
Will this way of packaging the files allow to use them in unit tests? I tried but seems the files are not in the bin folder.

build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.csproj
duncanp-sonar: Good point point. I've changed the property to `<SonarQubeExclude>true</SonarQubeExclude> so the entire project will be completely ignored.

duncanp-sonar:
Superceded.

PR #953, opened 2019-07-23
Load rules config from JSON files

henryju:
Based on #951 but unit tests are not passing since JSON files are not found.

### DIFF
--- a/src/Integration.Vsix/CFamily/RulesLoader.cs
+++ b/src/Integration.Vsix/CFamily/RulesLoader.cs
@@ -1,0 +1,89 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2018 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.IO;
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/CFamily/RulesLoader.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        public static List<string> ReadActiveRulesList()
+        {
+            var extensionFolder = Path.GetDirectoryName(typeof(RulesLoader).Assembly.Location);
+            string path = Path.Combine(extensionFolder, CFamilyFiles, "Sonar_way_profile.json");
+            using (StreamReader r = new StreamReader(path, Encoding.UTF8))
+            {
+                string json = r.ReadToEnd();
+                dynamic profile = JsonConvert.DeserializeObject(json);
```
duncanp-sonar: `dynamic` isn't commonly used because it's very simple to use typed classes with json e.g.

```
using Newtonsoft.Json;
using System.IO;
using System.Text;

namespace JsonSerializerExampleConsoleApp
{
    static class JsonExample
    {
        public static void Deserialize()
        {
            var rulesListFilePath = @"D:\repos\sq\slvs\src\Integration.Vsix\.CFamilyEmbedded\RulesList.json";
            var activeRulesFilePath = @"D:\repos\sq\slvs\src\Integration.Vsix\.CFamilyEmbedded\sonar_way_profile.json";
            var ruleParamsFilePath = @"D:\repos\sq\slvs\src\Integration.Vsix\.CFamilyEmbedded\S3222_params.json";

            // The rule list is just a list of strings so we don't need to define a custom type.
            // We could just as easily deserialize to a List<string>
            var rulesList = JsonConvert.DeserializeObject<string[]>(File.ReadAllText(rulesListFilePath, Encoding.UTF8));

            // For the more complicated types we can define custom data classes
            var rulesProfile = JsonConvert.DeserializeObject<RulesProfile>(File.ReadAllText(activeRulesFilePath, Encoding.UTF8));
            var ruleParams = JsonConvert.DeserializeObject<RuleParameter[]>(File.ReadAllText(ruleParamsFilePath, Encoding.UTF8));
        }
    }

    internal class RulesProfile
    {
        // Using attributes like this isn't strictly necessary but it allows us to
        // use the normal C# casing conventions for the properties.
        [JsonProperty("name")]
        public string Name { get; set; }

        [JsonProperty("ruleKeys")]
        public List<string> RuleKeys { get; set; }
    }

    internal class RuleParameter
    {
        [JsonProperty("key")]
        public string Key { get; set; }

        [JsonProperty("description")]
        public string Description { get; set; }

        [JsonProperty("defaultValue")]
        public string DefaultValue { get; set; }

        [JsonProperty("type")]
        public string Type { get; set; }
    }
}
```

PR #959, opened 2019-07-25
Call subprocess.exe to execute CFamily analysis

### LINKED ISSUE #958: Call subprocess.exe to execute CFamily analysis
Call `subprocess.exe` instead of the SonarLintDaemon to execute C++ analysis

duncanp-sonar:
Fixes #958

### DIFF
--- a/src/Integration.Vsix/CFamily/CFamilyHelper.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyHelper.cs
@@ -126,7 +126,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
 
         private static Sonarlint.Issue ToSonarLintIssue(Message cfamilyIssue, string sqLanguage)
         {
-            return new Sonarlint.Issue()
+            var issue = new Sonarlint.Issue()
             {
                 FilePath = cfamilyIssue.Filename,
                 Message = cfamilyIssue.Text,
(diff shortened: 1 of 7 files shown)


src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -138,6 +138,31 @@ private static Sonarlint.Issue ToSonarLintIssue(Message cfamilyIssue, string sqL
                 EndLine = cfamilyIssue.EndLine,
                 EndLineOffset = cfamilyIssue.EndColumn - 1
             };
+            return FixUpPositions(issue);
+        }
+
+        internal /* for testing */ static Sonarlint.Issue FixUpPositions(Sonarlint.Issue issue)
+        {
+            // SonarLint issue lines are 1-based, offsets are 0-based.
```
henryju: In fact the proper way to handle that would be to understand "no end line" as "select the entire line". IMO this is not doable at this location since we have no pointer on the document. That's why I think the code should be heavily reworked.

duncanp-sonar: Agreed - done in [this](https://github.com/SonarSource/sonarlint-visualstudio/commit/7867c93f712eb5d5cf1ff37bd0f184049597bc4a) new commit.
I'll close this PR and create another.

PR #960, opened 2019-07-25
Create a singleton to cache rules metadata

henryju:


src/Integration.Vsix/CFamily/RulesMetadataCache.cs
```diff
@@ -0,0 +1,53 @@
...
+        public Dictionary<string, Dictionary<string, string>> RulesParameters { get; }
+
+        public Dictionary<string, RuleMetadata> RulesMetadata { get; }
+
+        private RulesMetadataCache()
+        {
+            AllRulesList = RulesLoader.ReadRulesList();
+            ActiveRulesList = RulesLoader.ReadActiveRulesList();
+            RulesParameters = AllRulesList
+                .GroupBy(ruleKey => ruleKey)
```
duncanp-sonar: I don't think you need to group the elements first - the following should work:

```
            RulesParameters = AllRulesList.ToDictionary(key => key, key => RulesLoader.ReadRuleParams(key));
            RulesMetadata = AllRulesList.ToDictionary(key => key, key => RulesLoader.ReadRuleMetadata(key));
```

henryju: Indeed!

Issue #926, opened 2019-07-16
Delay starting the daemon until a file needs to be analyzed


duncanp-sonar:
The daemon starts automatically when VS is launched. This happens even before a solution is opened, and will be running even if there are never any files it can analyse.
Also, once the C++ analyzer is extracted from the daemon, the daemon will only analyse JavaScript files (unless and until we add analysis of more languages).

To reduce the load on the end-user's machine, we won't start the daemon until a file is opened that can be analysed by the daemon.

#### Implications
* the initial analysis for that first file will be slower because of the overhead in starting up the Java daemon. However, we've decided that this is acceptable as long as the initial analysis appears within a relatively short time e.g. five seconds.
* in many VS sessions the daemon will never start. TODO: consider impact on telemetry.

#### TODO
- [x] Review the initialization code in the SonarLintDaemonPackage - currently responsible for launching the daemon/UI on startup? Separate out the download code so newer daemon versions are downloaded but not started.
- [x] Review analysis trigger code - check whether this is async or not.
- [x] Check exception handling when daemon fails to start: (1) initial exception should be handled gracefully. (2) What do we do next time a file is opened - try to start the daemon again? UPDATE: exception during start is handled gracefully, but the current file won't be analyzed. An attempt is made to start the daemon when the file is saved again.

~~- [ ] Check the list of file types that need to be analysed (not just `.js`, there are other types like '.vue')~~

PR #966, opened 2019-07-26
Introduced IAnalyzerController to break direct dependency on ISonarLintDaemon

duncanp-sonar:
Fixed up existing unit tests

Note: C++ analysis no longer requires additional language support to be enabled

I'm trying to break the refactorings into small pieces so they can still be reviewed. I'm not worrying at the moment about the code coverage; there's no point in additional unit tests to cover code that is still changing.

### DIFF
--- a/src/Integration.Vsix/InternalsVisibleTo.cs
+++ b/src/Integration.Vsix/InternalsVisibleTo.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarLint for Visual Studio
  * Copyright (C) 2016-2018 SonarSource SA
  * mailto:info AT sonarsource DOT com
@@ -23,7 +23,14 @@ using System.Runtime.CompilerServices;
 #if SignAssembly
 [assembly: InternalsVisibleTo("SonarLint.VisualStudio.Integration.Vsix.UnitTests,PublicKey=002400000480000094000000060200000024000052534131000400000100010081b4345a022cc0f4b42bdc795a5a7a1623c1e58dc2246645d751ad41ba98f2749dc5c4e0da3a9e09febcb2cd5b088a0f041f8ac24b20e736d8ae523061733782f9c4cd75b44f17a63714aced0b29a59cd1ce58d8e10ccdb6012c7098c39871043b7241ac4ab9f6b34f183db716082cd57c1ff648135bece256357ba735e67dc6")]
 [assembly: InternalsVisibleTo("SonarLint.VisualStudio.Integration.TestInfrastructure,PublicKey=002400000480000094000000060200000024000052534131000400000100010081b4345a022cc0f4b42bdc795a5a7a1623c1e58dc2246645d751ad41ba98f2749dc5c4e0da3a9e09febcb2cd5b088a0f041f8ac24b20e736d8ae523061733782f9c4cd75b44f17a63714aced0b29a59cd1ce58d8e10ccdb6012c7098c39871043b7241ac4ab9f6b34f183db716082cd57c1ff648135bece256357ba735e67dc6")]
+
+// Moq -- see https://github.com/Moq/moq4/wiki/Quickstart
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
+
 #else
 [assembly: InternalsVisibleTo("SonarLint.VisualStudio.Integration.Vsix.UnitTests")]
 [assembly: InternalsVisibleTo("SonarLint.VisualStudio.Integration.TestInfrastructure")]
+
+// Moq
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
 #endif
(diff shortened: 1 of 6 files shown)


src/Integration.Vsix/InternalsVisibleTo.cs
```diff
@@ -23,7 +23,14 @@
...
 [assembly: InternalsVisibleTo("SonarLint.VisualStudio.Integration.TestInfrastructure,PublicKey=002400000480000094000000060200000024000052534131000400000100010081b4345a022cc0f4b42bdc795a5a7a1623c1e58dc2246645d751ad41ba98f2749dc5c4e0da3a9e09febcb2cd5b088a0f041f8ac24b20e736d8ae523061733782f9c4cd75b44f17a63714aced0b29a59cd1ce58d8e10ccdb6012c7098c39871043b7241ac4ab9f6b34f183db716082cd57c1ff648135bece256357ba735e67dc6")]
+
+// Moq -- see https://github.com/Moq/moq4/wiki/Quickstart
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```
duncanp-sonar: Some of the classes/interfaces we want to mock are internal, so they need to be made visible to the Moq library.

src/Integration.Vsix/SonarLintDaemon/IAnalyzerController.cs
```diff
@@ -0,0 +1,116 @@
...
+
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Linq;
+using EnvDTE;
+using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
+{
+    internal interface IAnalyzerController
```
duncanp-sonar: New interface introduced to break the link between the tagger classes and the daemon.
This means that we can use the same tagger and error list provider classes for both C++ and JavaScript.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -145,39 +137,11 @@ public void RequestAnalysis(string path, string charset, IEnumerable<SonarLangua
...
-                UnsafeRequestAnalysis(path, charset, detectedLanguages, issueConsumer, projectItem);
+                analyzerController.RequestAnalysis(path, charset, detectedLanguages, issueConsumer, projectItem);
             }
             catch (Exception ex) when (!Microsoft.VisualStudio.ErrorHandler.IsCriticalException(ex))
             {
-                logger.WriteLine($"Daemon error: {ex.ToString()}");
-            }
-        }
-
-        private void UnsafeRequestAnalysis(string path, string charset, IEnumerable<SonarLanguage> detectedLanguages, IIssueConsumer issueConsumer, ProjectItem projectItem)
```
duncanp-sonar: Code moved to the new `AnalyzerController`, with a few minor tweaks.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -94,11 +91,6 @@ internal sealed class TaggerProvider : IViewTaggerProvider, ITableDataSource
         /// </summary>
         public ITagger<T> CreateTagger<T>(ITextView textView, ITextBuffer buffer) where T : ITag
         {
-            if (!settings.IsActivateMoreEnabled)
-            {
-                return null;
-            }
```
duncanp-sonar: This check only applies to the daemon so we don't want to be doing it here.

src/Integration.Vsix/SonarLintDaemon/IAnalyzerController.cs
```diff
@@ -0,0 +1,116 @@
...
+            // languages and saves the file, we won't display any issues because there isn't a
+            // tagger. The user will have to close and re-open the file to see issues.
+            // Is this the behaviour we want, or should we return true here if the language
+            // is supported, and then check whether to analyze when RequestAnalysis is called?
+            bool isSupported = (languages.Contains(SonarLanguage.Javascript) &&
+                settings.IsActivateMoreEnabled);
+            return isSupported;
+        }
+
+        public void RequestAnalysis(string path, string charset, IEnumerable<SonarLanguage> detectedLanguages, IIssueConsumer consumer, ProjectItem projectItem)
```
duncanp-sonar: This method was moved from the `TaggerProvider`. Minor tweaks to handle cases where the daemon is not enabled/not started.

Ideally, this class wouldn't need to know which languages are supported by the daemon; that should be internal to the daemon. Fixing that would increase the scope of this PR too much.

src/Integration.Vsix/SonarLintDaemon/IAnalyzerController.cs
```diff
@@ -0,0 +1,116 @@
...
+                        handled = true;
+                        if (settings.IsActivateMoreEnabled)
+                        {
+                            // User might have disable additional languages in the meantime
+                            break;
+                        }
+
+                        if (!daemon.IsRunning) // daemon might not have finished starting / might have shutdown
+                        {
+                            logger.WriteLine("Daemon has not started yet. Analysis will not be performed");
```
duncanp-sonar: Starting the daemon will be handled as part of #926: Delay starting the daemon until a file needs to be analyzed

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -207,7 +171,7 @@ public void AddIssueTracker(TextBufferIssueTracker bufferHandler)
             lock (managers)
             {
                 issueTrackers.Add(bufferHandler);
-                daemon.Ready += bufferHandler.DaemonStarted;
+                //TODO duncanp: daemon.Ready += bufferHandler.DaemonStarted;
```
duncanp-sonar: Will be done as part of #926: Delay starting the daemon until a file needs to be analyzed

PR #1025, opened 2019-10-09
Changed C-family rules config to be language-specific

duncanp-sonar:
Not all C-Family rules apply to all C-Family languages (C, C++, ObjC). SLVS detects whether the code being analysed is C or C++, but previously it passed the same rules configuration to `subprocess.exe` regardless.

This change means that the configuration passed to `subprocess.exe` is now language-specific.

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/RulesMetadataCacheTest.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/RulesMetadataCacheTest.cs
@@ -18,7 +18,9 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using FluentAssertions;
 using FluentAssertions.Execution;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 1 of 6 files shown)


src/Integration.Vsix.UnitTests/CFamily/RulesMetadataCacheTest.cs
```diff
@@ -30,23 +32,46 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
...
+            RulesLoader.ReadActiveRulesList().Should().HaveCount(255); // unexpanded list of active rules
+
+            RulesMetadataCache.GetSettings("c").ActiveRuleKeys.Should().HaveCount(Active_C_Rules);
+            RulesMetadataCache.GetSettings("cpp").ActiveRuleKeys.Should().HaveCount(Active_CPP_Rules);
+
+            // We don't currently support ObjC rules in VS
+            RulesMetadataCache.GetSettings("objc").Should().BeNull();
         }
 
+        private static IEnumerable<string> GetRulesByLanguage(IEnumerable<string> ruleKeys, string language) =>
```
duncanp-sonar: Unused - I'll remove it

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -86,8 +86,9 @@ private async Task TriggerAnalysisAsync(Request request, IIssueConsumer consumer
             if (response != null)
             {
                 Debug.Assert(response.Messages.All(m => m.Filename == request.File), "Issue for unexpected file returned");
+                var rulesConfig = RulesMetadataCache.GetSettings(request.CFamilyLanguage);
```
duncanp-sonar: Now we look up the configuration for the specific language.

src/Integration.Vsix/CFamily/RulesLoader.cs
```diff
@@ -121,6 +121,8 @@ public class RuleMetadata
             [JsonProperty("defaultSeverity")]
             public Sonarlint.Issue.Types.Severity DefaultSeverity { get; set; }
 
+            [JsonProperty("compatibleLanguages")]
+            public string[] CompatibleLanguages { get; set; }
```
duncanp-sonar: Need to read more information from the metadata json file to know which language(s) the rule applies to.

src/Integration.Vsix/CFamily/RulesMetadataCache.cs
```diff
@@ -36,27 +36,68 @@ internal interface IRulesConfiguration
...
-
-        public IDictionary<string, RuleMetadata> RulesMetadata { get; }
+        public static IRulesConfiguration GetSettings(string cFamilyLanguage)
+        {
+            RulesByLanguage.TryGetValue(cFamilyLanguage, out var rulesConfiguration);
+            return rulesConfiguration;
+        }
 
-        private RulesMetadataCache()
+        static RulesMetadataCache()
```
duncanp-sonar: The rule files are loaded and processed in a static constructor, so it is thread-safe, and will be executed the first time the `GetSettings` method is called in code.

PR #1032, opened 2019-10-18
Trigger re-analysis when user settings file changes

### LINKED ISSUE #1017: Re-analyse C++ files when the user-level rules configuration changes
Relates to #1015 

When the user changes the configuration, either by disabling a rule using the UI or by manually editing and saving the file, the error list should be updated to reflect the new configuration.

- [x] implement file system watcher to check for file changes
- [ ] enumerate open files and trigger re-analysis

### Notes
Updating the list of displayed issues could be done either by re-executing the analyses for open files, or by always running all rules but filtering the list of displayed rules in the `IssuesSnapshot`.

1. Filtering IssuesSnapshot
Pros: showing/hiding issues is fast - no need to reanalyze. Simpler to implement.
Cons: no performance gain from disabling rules as they will still be executed.

2. Save config file and re-analyze immediately
Cons:
* more complex - need to find all of the appropriate files and re-trigger the analysis. This might require some refactoring (?currently analysis triggering is mixed up with the tagger implementation?)
* performance - every rule deactivation will trigger a re-analysis, potentially for a large number of open files.
*? what about open dirty files? i.e. those with unsaved changes?
Pros:
* this functionality will also mean that rules that manually activated by editing the file will be triggered immediately.

3. Hybrid
When the user deactivates a rule in the UI, filter the list.
When the config file is changed, re-analyse.

Decision: option (2), since this will cover the manual activation case too. It also means that no additional logic is required to trigger re-analysis when the "disable rule" UI command is implemented - the command will just need to update the file.

### Special cases
Scenario: user has multiple files open, user manually edits and saves the settings file.
Re-analysing all of the files could take multiple seconds (e.g. one or two per file) so there is window during which a number things that could happen:
* user closes a file -> changes the list of open docs
* user changes and saves the settings file again -> should re-start the analysis with the new configuration
* user closes VS

duncanp-sonar:
Fixes #1017

### DIFF
--- a/src/Integration.Vsix/CFamily/SingleFileMonitor.cs
+++ b/src/Integration.Vsix/CFamily/SingleFileMonitor.cs
@@ -30,6 +30,14 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
     /// Monitors a single file for all types of change (creation, modification, deletion, rename)
     /// and raises an event for any change.
     /// </summary>
+    internal interface ISingleFileMonitor : IDisposable
+    {
+        event EventHandler FileChanged;
+    }
+
+    /// <summary>
+    /// Concrete implementation of ISingleFileMonitor
+    /// </summary>
     /// <remarks>
     /// Duplicate events can be raised by the lower-level file system watcher class - see
     /// https://stackoverflow.com/questions/1764809/filesystemwatcher-changed-event-is-raised-twice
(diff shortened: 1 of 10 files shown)


src/Integration.Vsix/CFamily/SingleFileMonitor.cs
```diff
@@ -30,6 +30,14 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
     /// Monitors a single file for all types of change (creation, modification, deletion, rename)
     /// and raises an event for any change.
     /// </summary>
+    internal interface ISingleFileMonitor : IDisposable
```
duncanp-sonar: Added for testability.

PR #1065, opened 2019-11-01
Updated Integration.csproj to SDK-format

duncanp-sonar:
* VSIXes install and run (binding smoke test)
* the Microsoft.IdentityModel.Clients.ActiveDirectory.* assemblies are no longer included in the VSIX - don't appear to be required


duncanp-sonar:
Superceded by #1066

PR #1066, opened 2019-11-05
Updated Integration.csproj to SDK-format

### LINKED ISSUE #923: [Task] Migrate projects to new-style SDK format
### Description
The projects are currently in old-style MSBuild format, but use wildcard imports rather than listing the files individually. This confuses VS when adding new files and means you can't build in the IDE until you fix up the project by:
* reverting the changes to the project file
* closing and re-opening the solution

#### Notes
Changing to the new-style project format means that we will no longer be able to build the solution in VS2015. However, this is no longer necessary - the project files have been de-duplicated and cleaned up, so we can build in MSBuild 16 and set a parameter to specify which version of VS we want to target.

To debug in VS2015, we'd need to build a debug VSIX, install it in VS2015, and then attach to that instance of VS2015 from another VS instance.

#### TODO
- [x] migrate test projects
- [x] migrate product projects
- [x] check analysis results are the same post-migration i.e. LOC, issues, coverage percentage.
- [x] check the VSIX artefacts contents are the same, particularly with reference to third-party assemblies

#### Potential problems
Handling of interop types: see https://www.nuget.org/packages/Microsoft.VisualStudio.SDK.EmbedInteropTypes/

duncanp-sonar:
Fixes #923 

* remove some unnecessary system assemblies from the VSIX.


duncanp-sonar:
Manually tested the artefacts built by the pipeline on VS2015, 2017 and 2019
* analysis of C#, VB, JS (daemon), C++(sub-process.exe)
* menu items -> pkgdef correctly created and merged
* options pages -> WPF resources embedded correctly
* binding/updating -> version-specific Team Explorer MEF extensions handled correctly
* C++: disabling rules / editing _settings.json_ works as expected

Comparison of VSIX contents: the new VSIX does not contain about a dozen System.* assemblies. These do not seem to be necessary.

PR #1105, opened 2019-11-20
Apply CPP-2305 to SonarLint: Properly predefine language version macros

### LINKED ISSUE #1106: C++: MSVC Driver: Properly predefine language version macros
### Description

SLVS contains a port of the sonar-cpp Java `MSVCDriver` class.
Bug [CPP-2305](https://jira.sonarsource.com/browse/CPP-2305) was found and fixed on the Java side. [The fix](https://github.com/SonarSource/sonar-cpp/pull/670) needs to be ported to SLVS.

### Repro steps

1. Configure the project to use C++ Language Standard `ISO C++17 Standard (std:c++17)`

2. Analyze any code that uses C++17 standard library. Example:
```
#include<optional>
// std::optional is introduced in c++17.
std::optional<int> foo(int x) // Parsing error issue raised.
{
  if (x < 0)
    return std::nullopt; // Parsing error issue raised.
  return x;
};
```
This code will trigger parsing error issue and will impact the analysis results.

NOTE: to see the parsing errors reported as issues you need to enabled the hidden `cpp:ParsingError` rule in the `settings.json` file:

```
{
  "sonarlint.rules": {
    "cpp:ParsingError": {
      "Level": "On"
    }
  }
}
```

### Expected behavior

File is analysed successfully and all issues reported.

### Actual behavior

Not all issues are reported.

Any parsing errors are swallowed by the `subprocess.exe`. No errors or warnings are reported in the UI or the SonarLint output pane i.e. there is no indication that there were any problems during analysis or that there may be missing issues.

Other issues in the same file in code that can be parsed will still be reported, whether they occur before or after the code that cannot be parsed.

### Known workarounds

None

### Related information

* SonarLint for VS Version: v4.x+
* Visual Studio Version: v2019 - v2017
* Languages in your solution: C++17

abbas-sabra-sonarsource:
Fixes #1106

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTest.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTest.cs
@@ -294,6 +294,20 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
         }
 
         [TestMethod]
+        public void LanguageStandard()
+        {
+            CFamilyHelper.FileConfig.ConvertLanguageStandard("").Should().Be("");
+            CFamilyHelper.FileConfig.ConvertLanguageStandard("Default").Should().Be("");
+            CFamilyHelper.FileConfig.ConvertLanguageStandard(null).Should().Be("");
+            CFamilyHelper.FileConfig.ConvertLanguageStandard("stdcpplatest").Should().Be("/std:c++latest");
+            CFamilyHelper.FileConfig.ConvertLanguageStandard("stdcpp17").Should().Be("/std:c++17");
+            CFamilyHelper.FileConfig.ConvertLanguageStandard("stdcpp14").Should().Be("/std:c++14");
+
+            Action action = () => CFamilyHelper.FileConfig.ConvertLanguageStandard("foo");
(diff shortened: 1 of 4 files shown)


abbas-sabra-sonarsource:
Hello @duncanp-sonar,
I discovered a bug in supporting C++17 features in our analyzer. It was discovered using SonarLint :D 
[The fix](https://github.com/SonarSource/sonar-cpp/pull/670) was done and tested on our side. This is just the propagation of the fix to SonarLint. 
Can you please review the PR?

duncanp-sonar:
Hi @abbas-sabra-sonarsource.
Thanks for porting the fix.

I've created a bug #1106 to track the fix on the SLVS side. A link to the bug will appear in the SLVS release notes so users can see what the issue was, and that it's been fixed.

Unfortunately, there isn't much info in the original bug about how to repro it and what the impact was (e.g. incorrect issues reported/no issues reported for the file/VS crashes etc). Could you add some detail to #1106 please?

Thanks again.

abbas-sabra-sonarsource:
> how often do you think users will encounter this issue?

The user will have parsing error issues triggered on every use of the C++17 standard library. This is blocking for the users using C++17 heavily. On the other side, this issue was not reported, which means we don't have a lot of such users.

> Is this bug serious enough and common enough to require a bug-fix release just for this bug?
I believe so

PR #1132, opened 2019-12-08
Moved more helper classes to Core

duncanp-sonar:


duncanp-sonar:
Note: code coverage on async methods is not being reported -> coverage on new code is being incorrectly shown as too low.

Issue #1238, opened 2020-02-24
Fix flaky tests


duncanp-sonar:
### Description
Some of the multi-threaded unit tests are unreliable, causing build failures approximately 10% of the time:

https://sonarsource.visualstudio.com/DotNetTeam%20Project/_test/analytics?definitionId=47&contextType=build

UPDATE:
The flaky tests have been tweaked to improve their reliability. We'll monitor the builds and decide if they are now stable enough.

I considered tweaking the Azure Pipeline build to re-run the failing tests, but decided not to do so at this point. If the failure rate is still too high then this would be an easy way to address the issue.

PR #1244, opened 2020-02-26
Fixed SingleFileMonitor dispose error

### LINKED ISSUE #1248: File system watcher can be used after being disposed
### Description
`SingleFileMonitor.Dispose` disposes the wrapped `FileSystemWatcher`. However, this can happen while a file change event is being handled, leading to an exception when the disposed `FileSystemWatcher` is used.

#### Impact
In practice, this shouldn't currently happen in the released product as we only create one `SingleFileMonitor` which is not disposed until VS is shut down. Also, the exception is thrown on a background thread so it should be handled by VS i.e. it won't crash VS.

The `ObjectDisposedException` does occur frequently when debugging the `SingleFileMonitor` tests.

duncanp-sonar:
Fixed SingleFileMonitor dispose error
Fixes #1248

Increase timeout to reduce test flakiness
Relates to #1238

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
@@ -205,9 +205,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
             directoryMock.Setup(x => x.Exists(It.IsAny<string>())).Returns(true);
 
             var watcherFactoryMock = CreateFactoryAndWatcherMocks(out var watcherMock);
-            watcherMock.SetupProperty(x => x.EnableRaisingEvents);
-            watcherMock.Object.EnableRaisingEvents.Should().BeFalse();
-
             var fileMonitor = new SingleFileMonitor(watcherFactoryMock.Object, directoryMock.Object, "c:\\dummy\\file.txt", testLogger);
 
             var eventHandlerStartedEvent = new ManualResetEvent(false);
@@ -228,26 +225,26 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
                 disposedEventSignalled = disposeCalledEvent.WaitOne(timeout);
             };
 
-            watcherMock.Object.EnableRaisingEvents.Should().BeTrue();
-            watcherMock.VerifySet(x => x.EnableRaisingEvents = true, Times.Exactly(1));
-
 
             // Stage 2: raise the file system event then block until we are in the event handler
             var eventHandlerMethodTask = System.Threading.Tasks.Task.Run(() =>
                 watcherMock.Raise(x => x.Created += null, new FileSystemEventArgs(WatcherChangeTypes.Created, "", "")));
 
             eventHandlerStartedEvent.WaitOne(timeout).Should().BeTrue();
-            watcherMock.Object.EnableRaisingEvents.Should().BeFalse();
 
 
             // Stage 3: dispose the monitor, unblock the background thread and wait until it has finished
+            watcherMock.Reset(); // reset the counts of calls to the watcher
+
             fileMonitor.Dispose();
             disposeCalledEvent.Set();
-            eventHandlerMethodTask.Wait(timeout).Should().BeTrue();
 
+            eventHandlerMethodTask.Wait(timeout).Should().BeTrue();
             disposedEventSignalled.Should().BeTrue();
-            watcherMock.Object.EnableRaisingEvents.Should().BeFalse();
-            watcherMock.VerifySet(x => x.EnableRaisingEvents = true, Times.Exactly(1));
+
+            // Expect a single call to watcher.Dispose(), and no other calls
+            watcherMock.Verify(x => x.Dispose(), Times.Once);
+            watcherMock.VerifyNoOtherCalls();
         }
 
         #region Simple file operations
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
```diff
@@ -326,31 +427,33 @@ public void RealFile_MultipleOperations_NoDuplicates()
 
             void testBody()
             {
+                const int pauseBetweenOpsInMs = 125;
```
duncanp-sonar: Test flakiness:  increasing the delay slightly to improve the reliability of the test.

src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
```diff
@@ -149,8 +149,109 @@ public void OnlyRaisesEventsIfHasListeners()
             }
         }
 
+        [TestMethod]
+        public void AfterDispose_EventsAreIgnored()
```
duncanp-sonar: New test to check that the event handlers are unregistered when the watcher is disposed.

src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
```diff
@@ -149,8 +149,109 @@ public void OnlyRaisesEventsIfHasListeners()
...
+
+                // 2. Dispose then re-raise
+                fileMonitor.Dispose();
+                raiseEvent(watcherMock);
+                eventCount.Should().Be(1);
+            }
+        }
+    
+        [TestMethod]
+        public void DisposeWhileHandlingFileEvent_DisposedWatcherIsNotCalled()
```
duncanp-sonar: Regression test for an issue I found when debugging the flaky test. It's possible for the monitor to be dispose while it is in the middle of handling a file event, which can lead to an unhandled exception later in the event handler.

This isn't likely to happen in practice, and even if it does it shouldn't affect the user since the exception is on a background thread and will be silently swallowed by VS.

However, since the fix was trivial I decided to fix it (also, it was irritating when debugging some of the other tests).
The fix might have been trivial, but writing a regression test wasn't. In retrospect, this might have been a case in which wasn't worth writing a regression test.

src/Integration.Vsix/Analysis/SingleFileMonitor.cs
```diff
@@ -164,13 +171,17 @@ void Dispose(bool disposing)
         {
             if (!disposedValue)
             {
+                disposedValue = true;
```
duncanp-sonar: Need to set this flag before disposing the `fileWatcher` to avoid a race condition in the `OnFileChanged` handler.

src/Integration.Vsix/Analysis/SingleFileMonitor.cs
```diff
@@ -153,7 +156,11 @@ private void OnFileChanged(object sender, System.IO.FileSystemEventArgs args)
             }
             finally
             {
-                fileWatcher.EnableRaisingEvents = true;
+                // Re-check we haven't been disposed on another thread
+                if (!disposedValue)
+                {
+                    fileWatcher.EnableRaisingEvents = true;
```
duncanp-sonar: An `ObjectDisposedException` was being thrown here.

src/Integration.Vsix/Analysis/SingleFileMonitor.cs
```diff
@@ -73,7 +73,7 @@ public SingleFileMonitor(IFileSystemWatcherFactory factory, IDirectory directory
             fileWatcher.Changed += OnFileChanged;
             fileWatcher.Created += OnFileChanged;
             fileWatcher.Deleted += OnFileChanged;
-            fileWatcher.Renamed += (s, args) => OnFileChanged(s, args);
+            fileWatcher.Renamed += OnFileRenamed;
```
rita-gorokhod-sonarsource: Why not use OnFileChanged? (like the rest of the events)

duncanp-sonar: The event argument type is different.

duncanp-sonar: ...but we could still use the same handler method. Not sure why it was ever different...
I'll use `OnFileRenamed`

src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
rita-gorokhod-sonarsource: I think we should not use real IO

src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
```diff
@@ -149,8 +149,106 @@ public void OnlyRaisesEventsIfHasListeners()
...
+
+            fileMonitor.Dispose();
+            disposeCalledEvent.Set();
+
+            eventHandlerMethodTask.Wait(timeout).Should().BeTrue();
+            disposedEventSignalled.Should().BeTrue();
+
+            // Expect a single call to watcher.Dispose(), and no other calls
+            watcherMock.Verify(x => x.Dispose(), Times.Once);
+            watcherMock.VerifyNoOtherCalls();
```
duncanp-sonar: Changed to use `VerifyNoOtherCalls` as suggested by @rita-gorokhod-sonarsource.
Much neater - thanks!

src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
duncanp-sonar: This test is specifically about testing how the class handles duplicate events from the real file system.

PR #1245, opened 2020-02-26
Added pause to CancellableJobRunner test to reduce flakiness

duncanp-sonar:
The CancelAfterFirstOperation test fails intermittently because of timing issue.
It's not a product issues and it isn't worth the cost of refactoring the product code to make the test 100% reliable, so I've added a 200ms pause to see if that improves reliability sufficiently.

Relates to #1238

### DIFF
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/CancellableJobRunnerTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/CancellableJobRunnerTests.cs
@@ -141,6 +141,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintTagger
 
 
             WaitForRunnerToFinish(testSubject, testLogger);
+            // Pause for any final progress steps to be reported before checking the progressRecorder below
+            Thread.Sleep(200);
 
             // Other checks
             testSubject.State.Should().Be(CancellableJobRunner.RunnerState.Cancelled);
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix.UnitTests/SonarLintTagger/CancellableJobRunnerTests.cs
```diff
@@ -141,6 +141,8 @@ public void CancelAfterFirstOperation()
 
 
             WaitForRunnerToFinish(testSubject, testLogger);
+            // Pause for any final progress steps to be reported before checking the progressRecorder below
+            Thread.Sleep(200);
```
rita-gorokhod-sonarsource: @duncanp-sonar I'm not arguing with the fix, but can you explain why is this happening?
I see in the code that CancellableJobRunner.RunnerState is being changed before the token is cancelled, and the test progressRecorder is a sync operation. So what is it that we need to wait for?

duncanp-sonar: The test thread can restart running as soon as the `Cancel` operation is performed, so there are now two threads running:
a) the `CancellableJobRunner` thread that calls the `progressRecorder` to write the final "job cancelled" message, and
b) the test thread that is checking for that "job cancelled" message.

The test fails intermittently if (b) happens before (a). The delay is just to reduce/remove the number of times this happens.

src/Integration.Vsix/SonarLintTagger/CancellableJobRunner.cs
```diff
@@ -81,7 +81,7 @@ public static CancellableJobRunner Start(string jobDescription, IEnumerable<Acti
         {
             var runner = new CancellableJobRunner(jobDescription, operations, progress, logger);
 
-            runner.Execute()
+            runner.ExecuteAsync()
```
duncanp-sonar: Miscellaneous: fix code smell by changing method name to follow the convention.

PR #1246, opened 2020-02-26
1214 Integrate System.IO.Abstractions

rita-gorokhod-sonarsource:


src/Integration/Telemetry/TelemetryDataRepository.cs
```diff
@@ -86,53 +84,49 @@ private void OnStorageFileChanged(object sender, FileSystemEventArgs e)
...
             var success = RetryHelper.RetryOnException(3, TimeSpan.FromSeconds(2),
                 () =>
                 {
-                    writer = fileWrapper.CreateText(StorageFilePath);
-                    this.ignoreFileChange = true;
-                    telemetrySerializer.Serialize(writer, Data);
-                    writer?.Flush();
+                    ignoreFileChange = true;
+                    var stringBuilder = new StringBuilder();
+                    telemetrySerializer.Serialize(new StringWriter(stringBuilder), Data);
```
rita-gorokhod-sonarsource: I couldn't mock CreateText, since it returns StreamWriter - which is not convertable to StringWriter.
Our original implementation casted it to TextWriter, which allowed the tests to run.
In this issue https://github.com/System-IO-Abstractions/System.IO.Abstractions/issues/392 they suggest to just switch to WriteAllText and ReadAllText, and I agree that they're simpler.

src/Integration/ProfileConflicts/ConflictsManager.cs
```diff
@@ -133,9 +126,6 @@ private IReadOnlyList<ProjectRuleSetConflict> FindConflicts(RuleSetInformation[]
             var ruleSetInfoProvider = this.serviceProvider.GetService<ISolutionRuleSetsInformationProvider>();
             ruleSetInfoProvider.AssertLocalServiceIsNotNull();
 
-            var fileSystem = this.serviceProvider.GetService<IFileSystem>();
```
rita-gorokhod-sonarsource: So this would originally return SourceControlledFileSystem - but since the actual usage is only File.Exists we can use FileSystem.

src/Integration/LocalServices/RuleSetSerializer.cs
```diff
@@ -47,10 +43,7 @@ public RuleSet LoadRuleSet(string path)
                 throw new ArgumentNullException(nameof(path));
             }
 
-            var fileSystem = this.serviceProvider.GetService<IFileSystem>();
```
rita-gorokhod-sonarsource: Please see https://github.com/SonarSource/sonarlint-visualstudio/pull/1246/files#r385043691

src/Integration/LocalServices/SolutionRuleSetsInformationProvider.cs
```diff
@@ -160,10 +163,7 @@ public bool TryGetProjectRuleSetFilePath(Project project, RuleSetDeclaration dec
             // Note: currently we don't search in rule set directories since we expect the project rule set
             // to be relative to the project. We can add this in the future if it will be needed.
 
-            IFileSystem fileSystem = this.serviceProvider.GetService<IFileSystem>();
```
rita-gorokhod-sonarsource: Please see https://github.com/SonarSource/sonarlint-visualstudio/pull/1246/files#r385043691

src/Integration/Telemetry/TelemetryDataRepository.cs
```diff
@@ -86,53 +84,49 @@ private void OnStorageFileChanged(object sender, FileSystemEventArgs e)
...
         {
-            TextWriter writer = null;
             var success = RetryHelper.RetryOnException(3, TimeSpan.FromSeconds(2),
                 () =>
                 {
-                    writer = fileWrapper.CreateText(StorageFilePath);
-                    this.ignoreFileChange = true;
-                    telemetrySerializer.Serialize(writer, Data);
-                    writer?.Flush();
+                    ignoreFileChange = true;
```
rita-gorokhod-sonarsource: I had to re-write those methods, since File.CreateText returns StreamWriter (in system.io.abstraction's implementation), which is not convertible to StringWriter that we use in the tests.
It used to work in the past because in our FileWrapper we casted it to TextWriter.
I tried to mock StreamWriter using MemoryStream but it was too complicated for the test flow, and File.WriteAllText is easier in my opinion anyway.
I can instead re-write the tests and use System.IO.Abstractions.TestingHelpers.MockFileStream, whatever you prefer.

src/Integration/LocalServices/ISourceControlledFileSystem.cs
```diff
@@ -19,15 +19,18 @@
...
 namespace SonarLint.VisualStudio.Integration
 {
     /// <summary>
     /// Provides abstraction over the SCC file management. The files are queued up using <see cref="QueueFileWrite(string, Func{bool})"/>
     /// until <see cref="WriteQueuedFiles"/> which will call the SCC to check out the file and will run the queued up write operations.
     /// </summary>
-    internal interface ISourceControlledFileSystem : IFileSystem
+    internal interface ISourceControlledFileSystem : ILocalService
     {
+        IFileSystem FileSystem { get; }
```
duncanp-sonar: You've changed `ISourceControlledFileSystem` so it is no longer a `IFileSystem` (which is good). I think you should separate them completely i.e. remove this property. A quick look at the code suggests that there are only two product classes that would be affected by this (and it looks like one of them is only accessing the file system in assertions so it probably doesn't need an `IFileSystem` at all).

Let's treat `IFileSystem` as purely an abstraction for testing that can be injected into the constructor of classes that need it.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -408,4 +414,16 @@
...
+      <VSIXCopyLocalReferenceSourceItem Include="@(ReferenceCopyLocalPaths)" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="IncludeNuGetPackageReferences" AfterTargets="GetVsixSourceItems">
+    <ItemGroup>
+      <VSIXSourceItem Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' == 'System.IO.Abstractions'" />
+    </ItemGroup>
+  </Target>
+
```
duncanp-sonar: Are both of these targets required, and if so why?

src/Integration.Vsix/ThirdPartyLicenses/System.IO.Abstractions.txt
```diff
@@ -0,0 +1,23 @@
+﻿The MIT License (MIT)
```
duncanp-sonar: Maybe include a link to the source repo to make it easy to track back?

src/Integration/Integration.csproj
```diff
@@ -97,11 +98,7 @@
        NB the target is based on the one from Microsoft.VisualStudio.SDK.EmbedInteropTypes above. -->
   <Target Name="HackIncludeALMAuthInVsix" AfterTargets="ResolveReferences" BeforeTargets="FindReferenceAssembliesForReferences">
     <ItemGroup>
-      <ReferencePath Condition="
-              '%(FileName)' == 'Microsoft.Alm.Authentication'
-           or '%(FileName)' == 'Microsoft.Alm.Git'
-           or '%(FileName)' == 'Microsoft.Vsts.Authentication'
-              ">
+      <ReferencePath Condition="&#xD;&#xA;              '%(FileName)' == 'Microsoft.Alm.Authentication'&#xD;&#xA;           or '%(FileName)' == 'Microsoft.Alm.Git'&#xD;&#xA;           or '%(FileName)' == 'Microsoft.Vsts.Authentication'&#xD;&#xA;              ">
```
duncanp-sonar: VS doesn't seem to like line breaks in conditions so it keeps trying to change them to control characters.
It's probably better to remove the control characters and just have the conditions on a single (long) line.

src/Integration/LocalServices/RuleSetSerializer.cs
```diff
@@ -21,23 +21,19 @@
...
 
 namespace SonarLint.VisualStudio.Integration
 {
     internal sealed class RuleSetSerializer : IRuleSetSerializer
     {
-        private readonly IServiceProvider serviceProvider;
+        private readonly IFileSystem fileSystem;
 
-        public RuleSetSerializer(IServiceProvider serviceProvider)
+        public RuleSetSerializer(IFileSystem fileSystem)
```
duncanp-sonar: Add a test-only constructor to inject the `IFileSystem`

src/Integration/LocalServices/SolutionRuleSetsInformationProvider.cs
```diff
@@ -36,8 +37,9 @@ internal class SolutionRuleSetsInformationProvider : ISolutionRuleSetsInformatio
 
         private readonly IServiceProvider serviceProvider;
         private readonly ILogger logger;
+        private readonly IFileSystem fileSystem;
 
-        public SolutionRuleSetsInformationProvider(IServiceProvider serviceProvider, ILogger logger)
+        public SolutionRuleSetsInformationProvider(IServiceProvider serviceProvider, ILogger logger, IFileSystem fileSystem)
```
duncanp-sonar: Add a test-only constructor to inject the `IFileSystem`

src/Integration.Vsix.UnitTests/Analysis/UserSettingsProviderTests.cs
```diff
@@ -130,46 +130,55 @@ public void FileChanges_EventsRaised()
...
+            var testSubject = new UserSettingsProvider(new TestLogger(), fileSystemMock.Object, CreateMockFileMonitor(fileName).Object);
+            fileSystemMock.Reset();
 
             // Act
             testSubject.EnsureFileExists();
             
             // Assert
-            fileMock.Verify(x => x.Exists("c:\\missingFile.txt"), Times.Once);
-            fileMock.Verify(x => x.WriteAllText("c:\\missingFile.txt", It.IsAny<string>()), Times.Once);
+            fileSystemMock.Verify(x => x.File.Exists(fileName), Times.Exactly(2));
```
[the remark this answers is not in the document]
rita-gorokhod-sonarsource: Interesting, looks like a known issue - https://github.com/moq/moq4/issues/733
There are two calls - one in UserSettingsProvider's ctor and one in EnsureFileExists method.
None of the Moq methods Reset() or Invocations.Clear() actually stop the Verify from thinking there were 2 executions.
It used to work in the past because we used to do the setup _after_ UserSettingsProvider's ctor, so there really was only 1 call. But we can't do that anymore because we'll get Null exception if it's not configured.
I'll remove the call to Reset since it's misleading.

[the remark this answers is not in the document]
duncanp-sonar:
```
I've been expecting this issue to be raised eventually, thanks for reporting it. This is one of those not-so-clear-cut cases where it is perhaps up to personal interpretation what should happen.
``` 
:-)

PR #1250, opened 2020-02-27
Securely save token in the credentials store

### LINKED ISSUE #768: SonarQube token is visible in Credential Manager
### Description

When using a SonarQube token we store it as UserName in the Windows credential manager and it could be easily seen. Having in mind we show the tokens only once on SonarQube, it makes no sense for them to be visible in the Windows UI:

image

duncanp-sonar:
Fixes #768

This change only handles storing the token so it can't be seen in clear in the Windows Credential Manager.
It doesn't address the token being visible in the SL connection dialogue - that's covered by #1229

### DIFF
--- a/src/Integration.UnitTests/LocalServices/CredentialStoreTests.cs
+++ b/src/Integration.UnitTests/LocalServices/CredentialStoreTests.cs
@@ -29,39 +29,41 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.LocalServices
     [TestClass]
     public class CredentialStoreTests
     {
+        private Mock<ICredentialStore> mockCredentialStore;
+        private CredentialStore testSubject;
+        private TargetUri WellKnownTargetUri;
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            mockCredentialStore = new Mock<ICredentialStore>();
+            testSubject = new CredentialStore(mockCredentialStore.Object);
+            WellKnownTargetUri = new TargetUri("http://sonarcredtest/");
+        }
+
         [TestMethod]
         public void DeleteCreds()
         {
-            // Arrange
-            var mockCredentialStore = new Mock<ICredentialStore>();
-
-            var testSubject = new CredentialStore(mockCredentialStore.Object);
-            var targetUri = new TargetUri("http://sonarcredtest/");
-
             // Act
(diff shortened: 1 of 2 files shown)


src/Integration/LocalServices/CredentialStore.cs
```diff
@@ -28,9 +28,18 @@ namespace SonarLint.VisualStudio.Integration
     /// </summary>
     public class CredentialStore : ICredentialStoreService
     {
-        private readonly SecretStore store;
+        private readonly ICredentialStore store;
```
duncanp-sonar: Changed to an interface to make testing simpler.

duncanp-sonar:
There's a regression test to check credentials in the "old" format are handled correctly.
I've also manually tested the VSIX against saved old credentials, and they are correctly read and then saved in the new format.

src/Integration.UnitTests/LocalServices/CredentialStoreTests.cs
```diff
@@ -0,0 +1,157 @@
...
+namespace SonarLint.VisualStudio.Integration.UnitTests.LocalServices
+{
+    [TestClass]
+    public class CredentialStoreTests
+    {
+        [TestMethod]
+        public void DeleteCreds()
+        {
+            // Arrange
+            var mockCredentialStore = new Mock<ICredentialStore>();
```
rita-gorokhod-sonarsource: @duncanp-sonar how do you feel about TestInitialize? 
I think the Arrange blocks should move there, as they're a repetition in every test.

duncanp-sonar: My main goal is that unit tests should be completely independent of each other i.e. changing one test shouldn't impact another test. Both `TestInitialize` and sharing data between tests go against this, so in general I'm not in favour of `TestInitialize` as it tends to be misused. It's only really useful when all of the tests have exactly the same set of prerequisites which often isn't the case. There are lots of examples in the SLVS tests of `TestInitialize` being misused in this way. I tend to write tests as standalone first (i.e. without a `TestInitailize`), and then don't generally consider adding one later.

Unusually, this set of tests is currently simple enough to use TestInitialize, so I'll change them.

Constants - it depends. Here, we don't really care what the Uri is so making a shared const is ok. Generally, I'd want to vary the test data within each test so I wouldn't use constants.

PR #1253, opened 2020-02-28
1196 Fix utc profile timestamp

rita-gorokhod-sonarsource:


duncanp-sonar:
Is there a test that checks that a file with a date in the old format is deserialized correctly?

rita-gorokhod-sonarsource:
> Is there a test that checks that a file with a date in the old format is deserialized correctly?

Added now, thanks, good point.

Issue #787, opened 2019-01-09
[Test projects] test RegEx is not correctly applied when checking for unbound projects


duncanp-sonar:
### Description

SLVS downloads a regular expression that is used to detect test projects by name. It uses a default regex if the value on the server isn't set.

However, there is a timing issue: the check for unbound projects can happen before the regular expression has been set. This can lead to a warning gold bar being shown saying that there are unbound projects (i.e. test projects not recognized as such and since they don't have a SL ruleset they will be treated as unbound).

**Timing:**
_ProjectSystemFilter.SetTestRegex_ is called as part of the _ConnectionWorkflow_.
However, the _ErrorListInfoBarController_ can already have called the _ProjectSystemFilter_ indirectly via _bindingInformationProvider.GetUnboundProjects()_. There is a [TODO](https://github.com/SonarSource/sonarlint-visualstudio/blob/aadf9950c7d2eb3a6978738e071de752a4fcd1f5/src/Integration/LocalServices/ErrorListInfoBarController.cs#L88) in the code related to this issue.

### Repro steps

1. Create solution containing a test project that has "Test" in the project name but doesn't have the test _ProjectTypeGuid_ property (unfolding a new Analyzer + CodeFix template in VS2015 will do).

2. Bind the solution to a project in SonarQube.

3. Close the _Team Explorer_

4. Close VS

5. Re-open VS, and then re-open the solution

### Expected behavior

The test project is not reported as an unbound project

### Actual behavior

The gold bar appears warning of unbound projects and the _SonarLint_ output lists the test project as being unbound.

### Known workarounds

* Add the standard test ProjectTypeGuid to the project file (as suggested in the SO post referred to above)
or
* Add the SonarQubeTestProject property to the project file, set to true.

### Related information

* SonarLint for VS Version: 4.7
* Languages affected: C#, VB.Net

### Related bug(s)
#785

PR #1256, opened 2020-03-03
#787 - test RegEx is not correctly applied when checking for unbound projects

rita-gorokhod-sonarsource:


rita-gorokhod-sonarsource:
@duncanp-sonar I talked to @henryju - the property is no longer settable on the server side. So as we discussed, I initialized the default regex earlier.

duncanp-sonar:
Please fix the failing tests!

PR #1257, opened 2020-03-04
785 test projects

rita-gorokhod-sonarsource:
Related to #785


rita-gorokhod-sonarsource:
@duncanp-sonar I extracted the test project indication code out of ProjectSystemFilter and into TestProjectIndicator implementations.

src/Integration.UnitTests/LocalServices/ProjectCapabilityTestProjectIndicatorTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            actual.Should().BeTrue();
+        }
+
+        private static void SetCapability(ProjectMock projectMock, string capability)
+        {
+            var vsHierarchy = projectMock as IVsHierarchy;
+            vsHierarchy.SetProperty(VSConstants.VSITEMID_ROOT, -2124, capability);
+        }
+
+        private static void SetupCapabilityEvaluator(ConfigurableServiceProvider serviceProvider)
```
rita-gorokhod-sonarsource: @duncanp-sonar I want to cry

src/Integration/LocalServices/TestProjectIndicators/BuildPropertyTestProjectIndicator.cs
```diff
@@ -0,0 +1,67 @@
...
+
+using System;
+using System.Diagnostics;
+using EnvDTE;
+using Microsoft.VisualStudio.Shell.Interop;
+using SonarLint.VisualStudio.Integration.Resources;
+
+namespace SonarLint.VisualStudio.Integration
+{
+    internal class BuildPropertyTestProjectIndicator : ITestProjectIndicator
```
rita-gorokhod-sonarsource: @duncanp-sonar I'm thinking about ditching this class, and moving all of its logic into TestProjectIndicator.
1. This way the interface can be bool and not bool?
2. TestProjectIndicator is highly aware of this anyway due to precedence, so maybe we should remove this separation.

duncanp-sonar: I think it's ok to have both; this class encapsulates knowledge about accessing build properties, `TestProjectIndictor` encapsulates the fact that there is a 'primary' indicator that should take precedence over secondary indicators. 

`TestProjectIndicator' doesn't need to know which indicators are primary and which are secondary, just that there is a precedence rule and how to apply it.

src/Integration/LocalServices/TestProjectIndicators/TestProjectIndicator.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Linq;
+
+namespace SonarLint.VisualStudio.Integration
+{
+    internal class TestProjectIndicator : ITestProjectIndicator
+    {
+        private readonly ITestProjectIndicator buildPropertyIndicator;
+        private readonly IEnumerable<ITestProjectIndicator> testProjectIndicators;
+
+        public TestProjectIndicator(ITestProjectIndicator buildPropertyIndicator, IEnumerable<ITestProjectIndicator> testProjectIndicators)
```
duncanp-sonar: Thinking about the `bool?` problem; at the moment, this class knows that there is something special about the first `ITestProjectIndicator` and that it needs to be treated differently (although it doesn't know or care whether it is a `BuildPropertyIndicator`, so I don't think the first parameter should be called `buildPropertyIndicator`).

A couple of different approaches to consider:
1) leave the class as it is, but change the name of the first parameter to indicate its special importance e.g. `primaryTestProjectIndicator`.

2) change the meaning of the `ITestProjectIndicator` class so that all project indicators can be treated the same i.e. the indicator classes should return `null` to mean `don't know`, and only return `true` or `false` if they are sure.
The indicators should be evaluated by this class in order, and it should stop evaluating as soon as one of them returns true or false. If none of them return a value then it should return false.

Effectively, this means that `ProjectCapabilityTestProjectIndicator`, `ProjectNameTestProjectIndicator` and `ProjectKindTestProjectIndicator` should all return either `null` or `true`. They should never return `false` (because if the specific value they are looking for isn't present then they have to return "don't know" i.e. null).

rita-gorokhod-sonarsource: Sounds good to me, I'll make the changes.

rita-gorokhod-sonarsource: I made the changes but it looks a bit off to me that all the `ITestProjectIndicators` effectively return `true `or `null`.
Thinking out loud: `buildPropertyIndicator` is the only one that can definitively say "no, this is not a test project", while the others can only say "no, this is not a test project according to my definition". So maybe we should separate the interfaces, something like:
`ITestProjectMarkerChecker { bool HasTestMarker(Project project); }`
and 
`ITestProjectIndicator { bool? IsTestProject(Project project); }`

`buildPropertyIndicator` will inherit from `ITestProjectIndicator`, while the others will inherit from `ITestProjectMarkerChecker`.
And the main composite `TestProjectIndicator` can receive a list of primary `ITestProjectIndicator`, and a list of `ITestProjectMarkerChecker`. 
Wdyt?

duncanp-sonar: It seems a bit over-complicated: one interface for something that can return _yes/no/don't know_, and a different interface that can return only _yes/don't know_.

I don't have a strong preference. It's only a small amount of code and the impact is localized so it would be easy to change later.

src/Integration/LocalServices/TestProjectIndicators/ITestProjectRegexSetter.cs
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+namespace SonarLint.VisualStudio.Integration
+{
+    internal interface ITestProjectRegexSetter : ILocalService
+    {
+        /// <summary>
+        /// Sets the regex to that identifies project's name as test project
+        /// </summary>
+        /// <param name="pattern"></param>
```
duncanp-sonar: Redundant empty doc param.

src/Integration/LocalServices/TestProjectIndicators/ProjectKindTestProjectIndicator.cs
```diff
@@ -0,0 +1,51 @@
...
+            projectSystem.AssertLocalServiceIsNotNull();
+        }
+
+        public bool? IsTestProject(Project project)
+        {
+            var hierarchy = projectSystem.GetIVsHierarchy(project);
+            var aggregateProjectKinds = projectSystem.GetAggregateProjectKinds(hierarchy).ToList();
+
+            return aggregateProjectKinds.Contains(ProjectSystemHelper.TestProjectKindGuid) ||
+                   aggregateProjectKinds.Contains(ProjectSystemHelper.ExternalTestProjectKindGuid);
```
duncanp-sonar: _ExternalTestProjectKindGuid_ is the Service guid, not a project type guid.  At the MSBuild level they are represented differently:

```
<ItemGroup>
    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
</ItemGroup>
<PropertyGroup>
    <ProjectTypeGuids>{60dc8134-eba5-43b8-bcc9-bb4bc16c2548};{F184B08F-C81C-45F6-A57F-5ABD9991F28F
</ProjectTypeGuids>
 </PropertyGroup>
```

Have you manually tested that at runtime VS treats Service guids in the same way as project type guids?

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -301,6 +300,22 @@ private void RegisterLocalServices()
...
+            {
+                var buildPropertyTestProjectIndicator = new BuildPropertyTestProjectIndicator(this);
+                var testProjectIndicators = new List<ITestProjectIndicator>
+                {
+                    new ProjectKindTestProjectIndicator(this),
+                    projectNameTestProjectIndicator.Value as ITestProjectIndicator,
+                    new ProjectCapabilityTestProjectIndicator(this)
+                };
+
+                return new TestProjectIndicator(buildPropertyTestProjectIndicator, testProjectIndicators);
```
duncanp-sonar: Does this need to be a service? It looks like it is only used by the `ProjectSystemFilter`.

In general, let's try to avoid adding new services if they are not required - it just adds complexity and makes it harder to track dependencies.

src/Integration.UnitTests/LocalServices/BuildPropertyTestProjectIndicatorTests.cs
```diff
@@ -0,0 +1,106 @@
+﻿using System;
```
duncanp-sonar: Missing copyright notices on all of the new test files.

Issue #1243, opened 2020-02-26
Embed SonarCFamily 6.7


abbas-sabra-sonarsource:
sonarlint-visualstudio is currently on SonarCFamily 6.6.

[6.7 release notes](https://jira.sonarsource.com/secure/ReleaseNote.jspa?projectId=10090&version=15328)

This release shouldn't have any impact on SLVS.

PR #1258, opened 2020-03-04
Embed SonarCFamily 6.7.0.15300 #1243

rita-gorokhod-sonarsource:


rita-gorokhod-sonarsource:
@duncanp-sonar the tests fail because there were differences in the Lib folder - Sonary_way_profile.json was changed, and so were RulesList.json and additional rules (for example GlobalMainFunction.json is now only for cpp).
Do we need those tests? I think it's a problem that we need to know what was changed in the rules.

duncanp-sonar:
@rita-gorokhod-sonarsource those tests are intended to be a safety net to check a couple of things: (1) that nothing has changed on the CFamily side (e.g. location of files, file format), and (2) that the download, unzip, and import of the CFamily analyzer files has worked correctly,

I agree that we're shouldn't really need to know about the number of rules or the specifics of any particular rule. However, checking the rule count and configuration value for a few known rules is a cheap and easy way of doing a high-level sanity check; the downside is that we need to manually check and update the number of expected rules for each release. Feel free to suggest alternatives.

rita-gorokhod-sonarsource:
> @rita-gorokhod-sonarsource those tests are intended to be a safety net to check a couple of things: (1) that nothing has changed on the CFamily side (e.g. location of files, file format), and (2) that the download, unzip, and import of the CFamily analyzer files has worked correctly,
> 
> I agree that we're shouldn't really need to know about the number of rules or the specifics of any particular rule. However, checking the rule count and configuration value for a few known rules is a cheap and easy way of doing a high-level sanity check; the downside is that we need to manually check and update the number of expected rules for each release. Feel free to suggest alternatives.

Since this is a basic sanity check, I think checking the exact rule count is the problem - we shouldn't care if a rule or two were added as long as the general functionality works. How about we change the test to check for an estimate? e.g. at least 150~ rules?

src/Integration.Vsix.UnitTests/CFamily/CFamilyEmbeddedSonarWayRulesTests.cs
```diff
@@ -31,29 +33,34 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
...
+            Active_C_Rules = LoadRulesCount("c");
+            Active_CPP_Rules = LoadRulesCount("cpp");
+        }
 
-        private const int Active_CPP_Rules = 268;
-        private const int Inactive_CPP_Rules = 171;
+        private static int LoadRulesCount(string language)
+        {
+            using var client = new HttpClient();
+            var uri = $"https://sonarcloud.io/api/qualityprofiles/backup?qualityProfile=Sonar%20way&language={language}&organization=microsoft";
```
duncanp-sonar: Why `microsoft` rather than an org we own/control?

rita-gorokhod-sonarsource: Because then we'd have to be authenticated, as our org is Private. It doesn't matter much IMO as the profile is the same for everyone.

src/Integration.Vsix.UnitTests/CFamily/CFamilyEmbeddedSonarWayRulesTests.cs
```diff
@@ -31,29 +33,34 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
...
-        // 3. Filter by Repository = SonarAnalyzer C
-        // 4. Filter by Quality Profile = Sonar way C
-        // The QP filter has "active/inactive" tabs. The number of rules is shown in the top-right of the screen.
-        // 5. Repeat for C++.
+        private readonly CFamilySonarWayRulesConfigProvider rulesMetadataCache = new CFamilySonarWayRulesConfigProvider(CFamilyShared.CFamilyFilesDirectory);
 
-        // Rule data for C-Family plugin v6.6 (build 13759)
-        private const int Active_C_Rules = 165;
-        private const int Inactive_C_Rules = 107;
+        [ClassInitialize]
```
duncanp-sonar: This definitely needs some comments to explain what is going on.

> We want to sanity check that the number of rules in the embedded version is approximately correct.
However, it's expected that the number of active/inactive rules will change with every CFamily release.
Rather than updating these tests each time to check we have the actual number of tests, we're comparing
against the number of rules in SonarWay on SonarCloud.
We're assuming that the version of the CFamily plugin on SonarCloud will be the same as or newer than the
version we are embedding. Since it might not be exactly the same version, we won't check the exact rule
count; we'll allow +/- a few rules.

As discussed offline, I don't like this approach. Calling out to a web service from a unit test doesn't feel right. As stated in the assumptions, we can't assume the version of the plugin on SonarCloud is the same - or to put it another way, the test is designed to pass against a range of different plugin versions.
Checking against a specific number of rules makes me fairly confident that the VSIX contains the correct content. Checking against a fuzzy number doesn't give me same level of confidence.

rita-gorokhod-sonarsource: I don't think checking against this property is the proper indication of content-correctness. The analyzer can change without any rules being added or removed, so the test can pass even if the content has not been updated.
Moreover, I really don't think that we should be getting into specifics of rule counts between different analyzer versions. A) they don't have to change, and B) it's an implementation detail that is not our concern. It shouldn't matter to us if it's 159 rules or 158. 
I agree with you that our concern is content correctness - which is why we should focus on that. If anything, we can check the checksum of exe against what we know it should be, but even then - I think it should be part of the manual tests. I don't think we should unit test this.

duncanp-sonar: > I don't think checking against this property is the proper indication of content-correctness.

I agree. I'm saying that checking specific numbers is a more robust check than checking a range that is intended to match multiple versions.

> The analyzer can change without any rules being added or removed, so the test can pass even if the content has not been updated.

True, but this is an uncommon scenario. There's a big difference between a test that might still pass incorrectly in a corner case (two versions of the plugin with identical rule counts), and one that is designed to pass against multiple different analyzer versions.

We don't have any tests that cover the exe at all, which is a separate problem. Manual checking should be a last resort if we can think of an automated way to do it.

These test are about nearly five hundred json files that are extracted and copied as part of the build. There are too many of them to check individually.

src/Integration.Vsix.UnitTests/CFamily/CFamilyEmbeddedSonarWayRulesTests.cs
duncanp-sonar: Worth commenting, as it isn't obvious.

src/Integration.Vsix.UnitTests/CFamily/CFamilyEmbeddedSonarWayRulesTests.cs
rita-gorokhod-sonarsource: Ok, let's revisit this some other time.
I'll revert the sonarcloud code - can you please provide the rules count? 
I'm using sonarqube docker, which is the community edition, and it doesn't have cpp plugin.

PR #1260, opened 2020-03-04
1249 third party licenses

### LINKED ISSUE #1249: Check third-party license compliance
### Description
Check the third-party components we ship, and that we include appropriate license terms in the VSIX.

rita-gorokhod-sonarsource:
Fixes #1249


rita-gorokhod-sonarsource:
@duncanp-sonar I'm not sure if we should include all these licenses or not.
[System.IO.Abstractions](https://github.com/System-IO-Abstractions/System.IO.Abstractions) and [Newtonsoft.Json](https://github.com/JamesNK/Newtonsoft.Json/blob/master/LICENSE.md) both mention that "The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software." - so I do think we need to include them in the VSIX.
However protobuf and grpc do not state this, and we already have grpc here:
https://github.com/SonarSource/sonarlint-visualstudio/blob/master/THIRD-PARTY-NOTICES 

What do you think?

duncanp-sonar:
@rita-gorokhod-sonarsource the protobuf license does say `Redistributions in binary form must reproduce the above copyright notice`, so we should include that.

Apologies, I've never really noticed the `THIRD-PARY-NOTICES` file before, even though it's in the root folder of the project. 
I think we should just add the `System.IO.Abstractions`, `NewtonSoft` and `protobuf` licenses to the `THIRD-PARTY-NOTICES` file and include that in the VSIX, rather than shipping separate them in separate files.

How does that sound?

rita-gorokhod-sonarsource:
> @rita-gorokhod-sonarsource the protobuf license does say `Redistributions in binary form must reproduce the above copyright notice`, so we should include that.
> 
> Apologies, I've never really noticed the `THIRD-PARY-NOTICES` file before, even though it's in the root folder of the project.
> I think we should just add the `System.IO.Abstractions`, `NewtonSoft` and `protobuf` licenses to the `THIRD-PARTY-NOTICES` file and include that in the VSIX, rather than shipping separate them in separate files.
> 
> How does that sound?

No worries, the only reason I found it in the first place, was because I searched for other usages of Apache license to see what you're supposed to put instead of "Copyright [yyyy] [name of copyright owner]" (only to realize that we left it as yyyy :| ).
Ok, I'lll move everything there and include in the VSIX.

THIRD-PARTY-NOTICES
```diff
@@ -41,206 +41,18 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 License notice for Windows Azure Active Directory IdentityModel Extensions for .Net
 -----------------------------------------------------------------------------------
+Copyright (C) 2016-2020 SonarSource SA
```
duncanp-sonar: `Microsoft Corporation` rather than `SonarSource`?!

duncanp-sonar:
Unrelated, but in the same file: there's a referenced on line 10 to the old Jira project which is no longer used (`https://jira.sonarsource.com/browse/SVS`). Might be worth updating this to point to this repo instead.

THIRD-PARTY-NOTICES
rita-gorokhod-sonarsource: Nope, take a look at "APPENDIX: How to apply the Apache License to your work."

duncanp-sonar: `Windows Azure Active Directory IdentityModel Extensions for .Net` isn't our work, it's Microsoft's. 
The appendix describes what Microsoft should write in their license file to apply the license to their work.

Section 4c says that we should retain Microsoft's copyright notice in our work when redistributing it.

rita-gorokhod-sonarsource: Updated, took their license from https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/LICENSE.txt

PR #1264, opened 2020-03-09
1263 [Test projects] Projects tagged with test service GUID are not recognized as test projects

### LINKED ISSUE #1263: [Test projects] Projects tagged with test service GUID are not recognized as test projects
### Description

Test projects recognized by the _Test Explorer_ in VS will have the following property added to the project file:
```
<ItemGroup>
  <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
</ItemGroup>
```
SLVS doesn't handle this case.

See [this StackOverflow post](https://stackoverflow.com/questions/18614342/what-is-service-include-in-a-csproj-file-for) for more information.

### Notes
Both VS2015 and VS2017 add the service tag for old-style test projects (i.e. in MSBuild 14 format).

### Known workarounds

* Add the standard test _ProjectTypeGuid_ to the project file (as suggested in the SO post referred to above)
or
* Add the _SonarQubeTestProject_ property to the project file, set to _true_.

### Related information

* Languages affected: C#, VB.Net

### Background
For more information on how the Scanner for MSBuild detects test project, see [Analysis of product projects vs. test projects](https://github.com/SonarSource/sonar-scanner-msbuild/wiki/Analysis-of-product-projects-vs.-test-projects) in the S4MSB wiki

rita-gorokhod-sonarsource:
Fixes #1263

### DIFF
--- a/src/Integration.UnitTests/LocalServices/ServiceGuidTestProjectIndicatorTests.cs
+++ b/src/Integration.UnitTests/LocalServices/ServiceGuidTestProjectIndicatorTests.cs
@@ -1,4 +1,24 @@
-﻿using System;
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
 using System.IO.Abstractions;
 using FluentAssertions;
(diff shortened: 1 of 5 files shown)


src/Integration.UnitTests/LocalServices/ServiceGuidTestProjectIndicatorTests.cs
```diff
@@ -0,0 +1,161 @@
+﻿using System;
```
duncanp-sonar: Missing copyright notice

src/Integration/LocalServices/TestProjectIndicators/ServiceGuidTestProjectIndicator.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
```
duncanp-sonar: Missing copyright notice

src/Integration/LocalServices/TestProjectIndicators/ServiceGuidTestProjectIndicator.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        internal ServiceGuidTestProjectIndicator(IFileSystem fileSystem)
+        {
+            this.fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
+        }
+
+        public bool? IsTestProject(Project project)
+        {
+            var projectPath = project.FullName;
```
duncanp-sonar: It's probably worth guarding against file IO exceptions, especially if this code is ever called on the main UI thread. It might be simplest to catch and suppress all non-critical exceptions.

duncanp-sonar: FYI there a method `ErrorHandler.IsCriticalException` that's used to decide whether an exception should be suppressed or not. The usual pattern is like this:
```
```

I'll push a commit with this fix and then I'll merge the PR.

PR #1270, opened 2020-03-13
 CPP Suppressions Stubs 

rita-gorokhod-sonarsource:


src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -204,7 +196,7 @@ public void WhenNewIssuesAreFound_ButForIncorrectFile_ListenersAreNotUpdated()
             var errorListSinkMock2 = RegisterNewErrorListSink();
 
             var testSubject = new TextBufferIssueTracker(taggerProvider.dte, taggerProvider,
-                mockedJavascriptDocumentFooJs.Object, javascriptLanguage, new TestLogger());
+                mockedJavascriptDocumentFooJs.Object, javascriptLanguage, new TestLogger(), issuesFilter.Object);
```
duncanp-sonar: For me, creating a `testSubject` in `TestInitialize` then overriding it in a test method is a poor practice; it's inconsistent and clumsy. Personally, I'd create a separate `CreateDefaultTestSubject` method and explicitly call it from the tests that use it.

src/Integration/Suppression/SuppressedIssuesProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+    public class SuppressedIssuesProvider : ISonarQubeIssuesProvider
+    {
+        private readonly IActiveSolutionBoundTracker activeSolutionBoundTracker;
+        private readonly ISonarQubeIssuesProviderFactory sonarQubeIssuesProviderFactory;
+
+        private ISonarQubeIssuesProvider instance;
+        private bool disposed;
+
+        [ImportingConstructor]
+        public SuppressedIssuesProvider(ISonarQubeIssuesProviderFactory sonarQubeIssuesProviderFactory,
```
duncanp-sonar: This is too complicated and confusing: the ultimate consumer of the issues provider is the `IssuesFilter` class. However, it doesn't MEF-import the exported `ISonarQubeIssuesProviderFactory`.
Instead, it directly MEF-imports an implementation of `ISonarQubeIssueProvider`, which in turn MEF-imports the `ISonarQubeIssuesProviderFactory`.

duncanp-sonar:
As discussed offline, looks good apart from the problem around the factory class.

src/Integration/Suppression/SuppressedIssuesProvider.cs
duncanp-sonar: FYI for the avoidance of doubt, this code doesn't work in VS in its current form. For MEF to be able to use an `ImportingConstructor` it must be able to find matching `Export`s for all of the constructor parameters.

For this code to work, you would need to add an `[Export(typeof(ISonarQubeIssuesProviderFactory))]` attribute on the factory class and an `[ImportingConstructor]` on the factory class public constructor.

When making changes to how MEF components are created and wired up, it's worth debugging the extension in VS to make sure that VS can create all of the expected components (MEF composition errors happen silently in the background so they are easy to miss).

rita-gorokhod-sonarsource: Yep you're right, my bad. Extracted to class and didn't re-run the extension.

src/Integration/Suppression/SuppressedIssuesProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+        private readonly Func<BindingConfiguration, ISonarQubeIssuesProvider> createProviderFunc;
+
+        private ISonarQubeIssuesProvider instance;
+        private bool disposed;
+
+        [ImportingConstructor]
+        public SuppressedIssuesProvider(IActiveSolutionBoundTracker activeSolutionBoundTracker,
+            ISonarQubeService sonarQubeService,
+            ILogger logger)
+            : this(activeSolutionBoundTracker,
```
rita-gorokhod-sonarsource: @duncanp-sonar what do you think? 🙈

duncanp-sonar: As discussed offline, it will be slightly more readable with a delegate instead of a function.
Other than that, let's go with this for the time being so we can move on.

PR #1274, opened 2020-03-17
1269 Refresh suppression list

rita-gorokhod-sonarsource:


src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -38,24 +39,26 @@ internal interface IAnalysisConfigMonitor
     internal sealed class AnalysisConfigMonitor : IAnalysisConfigMonitor, IDisposable
     {
         private readonly IAnalysisRequester analysisRequester;
-        private readonly IUserSettingsProvider userSettingsProvider;
         private readonly IActiveSolutionBoundTracker activeSolutionBoundTracker;
```
rita-gorokhod-sonarsource: I haven't removed IActiveSolutionBoundTracker because unfortunately it is still needed in OnUserSettingsChanged.

duncanp-sonar: At least it isn't listening to the events.
The `IConfigurationProvider` provides a method to get the current binding info (and is what the active solution bound tracker uses internally). Unfortunately, it's a "local service" so it can't be directly imported.

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -72,24 +75,10 @@ private void OnUserSettingsChanged(object sender, EventArgs e)
...
-
-        private void OnSolutionBindingChanged(object sender, ActiveSolutionBindingEventArgs e)
-        {
-            // NB assumes exception handling is done by the AnalysisRequester
-            if (activeSolutionBoundTracker.CurrentConfiguration.Mode != NewConnectedMode.SonarLintMode.Standalone)
-            {
-                logger.WriteLine(AnalysisStrings.ConfigMonitor_SolutionBound);
-                analysisRequester.RequestAnalysis();
-            }
+            logger.WriteLine(AnalysisStrings.ConfigMonitor_BindingUpdated);
```
duncanp-sonar: Is this message still appropriate?

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -72,24 +75,10 @@ private void OnUserSettingsChanged(object sender, EventArgs e)
             }
         }
 
-        private void OnSolutionBindingUpdated(object sender, EventArgs e)
+        private void OnSuppressionsUpdated(object sender, EventArgs e)
         {
-            // NB assumes exception handling is done by the AnalysisRequester
```
duncanp-sonar: Error handling in event handlers in VS is important as events are often on the UI thread so an unhandled exception will crash VS. This comment captures the design decision that these event handlers deliberately are not handling errors.

If you don't like the comment, do something else to make this clear e.g. rename the method to `UnsafeHandleSuppresionsUpdated` (and in that case rename `OnUserSettingsChanged` for consistency).

src/Integration.UnitTests/Suppression/SuppressedIssuesProviderTests.cs
```diff
@@ -107,13 +107,43 @@ public void GetSuppressedIssues_HasSolutionBinding_ConnectedMode_ListFromBinding
...
 
             var actual = testSubject.GetSuppressedIssues("project guid", "file path");
 
             actual.Should().BeEquivalentTo(expectedIssues);
         }
 
+        [DataTestMethod]
+        [DataRow(SolutionBindingEventType.SolutionBindingUpdated)]
+        [DataRow(SolutionBindingEventType.SolutionBindingChanged)]
+        public void SolutionBindingEvent_Standalone_DoesNotRaiseSupressionsUpdatedEvent(SolutionBindingEventType eventType)
```
duncanp-sonar: FYI typo `...Supressions...` in this test and the next one. No need to fix it if you don't want to since it's in test code.

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
duncanp-sonar: For the avoidance of doubt, I'm not suggesting switching to use the `IConfigurationProvider` now.

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
rita-gorokhod-sonarsource: In a way - it _is_ caused due to updated binding...; but I was more concerned about changing our logging messages: do we do something with them? telemetry?

src/Integration.UnitTests/Suppression/SuppressedIssuesProviderTests.cs
rita-gorokhod-sonarsource: I'll fix, ty!

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
rita-gorokhod-sonarsource: Yea, I took a look at it, it won't be simple to MEF-it.

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
duncanp-sonar: Output messages are entirely separate from telemetry. They are supposed to be meaningful to the end user/us when doing support.

Reporting that the list of suppressions has been synchronized with the connected server might be better since that is all this class knows for sure.

PR #1275, opened 2020-03-17
Move some interfaces to Core

duncanp-sonar:
This PR contains almost no actual code changes. Instead, I've moved some of the files to other folders/projects:

* `IIssuesFilter` and `ISonarQubeIssuesProvider` don't depend on VS so I've moved them to the `Core` project. I haven't moved the implementation classes.

* Moved the `IssuesFilter` class to the `Suppressions` folder since that seems a better location than under the tagger folder. I moved the corresponding test classes.

* added the `IFilterableIssue` interface. There are currently multiple data classes that describe issues: the Protobuf one used by the daemon; the Protobuf one use to talk to the CFamily process; and the `LiveIssue` used when suppressing Roslyn issues. I'm not trying to standardise these at the moment.

### DIFF
--- a/src/Core/IIssuesFilter.cs
+++ b/src/Core/IIssuesFilter.cs
@@ -18,29 +18,26 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System;
 using System.Collections.Generic;
-using System.ComponentModel.Composition;
-using Sonarlint;
-using SonarLint.VisualStudio.Integration.Suppression;
 
-namespace SonarLint.VisualStudio.Integration.Vsix.SonarLintTagger
+namespace SonarLint.VisualStudio.Core
 {
-    [Export(typeof(IIssuesFilter))]
-    [PartCreationPolicy(CreationPolicy.Shared)]
-    public class IssuesFilter : IIssuesFilter
+    /// <summary>
+    /// Describes a single issue with the properties required for
+    /// it to be compared against server-side issues by the issues filter
+    /// </summary>
+    public interface IFilterableIssue
     {
-        private readonly ISonarQubeIssuesProvider sonarQubeIssuesProvider;
+        string RuleId { get; }
+        string FilePath { get; }
+        string LineHash { get; }
+        string ProjectGuid { get; }
+        int? StartLine { get; }
+        string WholeLineText { get; }
+    }
 
-        [ImportingConstructor]
-        public IssuesFilter(ISonarQubeIssuesProvider sonarQubeIssuesProvider)
-        {
-            this.sonarQubeIssuesProvider = sonarQubeIssuesProvider ?? throw new ArgumentNullException(nameof(sonarQubeIssuesProvider));
-        }
-
-        public IEnumerable<Issue> Filter(string path, IEnumerable<Issue> issues)
-        {
-            return issues;
-        }
+    public interface IIssuesFilter
+    {
+        IEnumerable<IFilterableIssue> Filter(string path, IEnumerable<IFilterableIssue> issues);
     }
 }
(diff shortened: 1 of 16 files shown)


src/Core/IIssuesFilter.cs
```diff
@@ -19,12 +19,25 @@
...
-using Sonarlint;
 
-namespace SonarLint.VisualStudio.Integration.Vsix.SonarLintTagger
+namespace SonarLint.VisualStudio.Core
 {
+    /// <summary>
+    /// Describes a single issue with the properties required for
+    /// it to be compared against server-side issues by the issues filter
+    /// </summary>
+    public interface IFilterableIssue
```
duncanp-sonar: The daemon Protobuf `Issue` class doesn't contain enough information to be able to compare against server-side issues. Specifically, it's missing the `WholeLineText` and `LineHash` properties.

I haven't added any implementations of this interface yet; I'll do that in the next commit.

src/Integration.Vsix/SonarLintIntegrationPackage.cs
```diff
@@ -88,7 +87,7 @@ private async System.Threading.Tasks.Task InitOnUIThreadAsync()
                 IServiceProvider serviceProvider = this;
 
                 var activeSolutionBoundTracker = await this.GetMefServiceAsync<IActiveSolutionBoundTracker>();
-                var sonarQubeIssuesProvider = await this.GetMefServiceAsync<ISonarQubeIssuesProvider>();
+                var sonarQubeIssuesProvider = await this.GetMefServiceAsync<Core.ISonarQubeIssuesProvider>();
```
duncanp-sonar: I'm qualifying the type name with `Core` here and in a few other places rather than adding a `using` statement because otherwise there is a naming conflict (both `Core` and one of the VS assemblies have an `ErrorHandler` class).

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -275,7 +274,9 @@ void IIssueConsumer.Accept(string path, IEnumerable<Issue> issues)
                 return;
             }
 
-            issues = issuesFilter.Filter(path, issues);
+            // TODO: collect the additional information required to compare be able to filter
+            // the issues (i.e. the text of the line and the line hash)
+            // issues = issuesFilter.Filter(path, issues);
```
duncanp-sonar: Next steps: the `TextBufferIssueTracker` class has access to the underlying text buffer so it can find the text for the line.

src/Core/IIssuesFilter.cs
rita-gorokhod-sonarsource: Maybe extract this to another file?

src/Integration.UnitTests/Suppression/IssuesFilterTests.cs
```diff
@@ -41,11 +42,28 @@ public void TestInitialize()
...
+        [TestMethod]
+        public void Filter_NullIssues_ThrowsArgumentNullException()
+        {
+
+            Action act = () => testSubject.Filter("c:\\path\\file1.txt", null);
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("issues");
+        }
+
+        [TestMethod]
+        public void Filter_NullFile_DoesNotThrow()
```
rita-gorokhod-sonarsource: Future comment: If the file path is irrelevant, then maybe it should be a DataRow input to the test that asserts that sonarQubeIssuesProvider was called with the correct arguments? (will be relevant when there is more implementation).

src/Core/IIssuesFilter.cs
duncanp-sonar: The interfaces are related and both very short; I didn't see any upside in having them in separate file.

src/Integration.UnitTests/Suppression/IssuesFilterTests.cs
duncanp-sonar: I'll write appropriate tests when there is an implementation. For now I just want enough coverage to pass the QG.

Issue #1102, opened 2019-11-19
Do not report issue if in suppression list


duncanp-sonar:
### Description

Plumbing: connect up the various components (issue loader, issue matching, analyzer) so that issues are checked against the suppression list and not reported if a match is found.

### TODO
- consider whether to log anything if an issue is suppressed (e.g. message in the output window)

Parent task: #1107

PR #1276, opened 2020-03-18
Apply issues filtering to the returned issues

duncanp-sonar:
Related #1102

### DIFF
--- a/src/Integration.Vsix.UnitTests/SonarLintDaemon/DaemonIssueAdapterTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintDaemon/DaemonIssueAdapterTests.cs
@@ -21,6 +21,7 @@
 using System;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration.Vsix;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintDaemon
@@ -37,16 +38,16 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintDaemon
         }
 
         [TestMethod]
-        public void Ctor_NullPathAndHash_NoException()
+        public void Ctor_NullPathAndHash_PathAndHashAreSetToNull()
         {
-            var result = new DaemonIssueAdapter(new Sonarlint.Issue(), null, null);
+            var result = (IFilterableIssue)new DaemonIssueAdapter(new Sonarlint.Issue(), null, null);
 
             result.WholeLineText.Should().BeNull();
             result.LineHash.Should().BeNull();
         }
 
         [TestMethod]
-        public void Ctor_ValidIssue()
+        public void Ctor_ValidIssue_CorrectlyInitialized()
         {
             // Arrange
             var issue = new Sonarlint.Issue
@@ -62,13 +63,14 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintDaemon
             // Assert
             result.SonarLintIssue.Should().BeSameAs(issue);
 
-            result.FilePath.Should().Be("path");
-            result.RuleId.Should().Be("rule id");
-            result.StartLine.Should().Be(111);
+            var filterableResult = (IFilterableIssue)result;
+            filterableResult.FilePath.Should().Be("path");
+            filterableResult.RuleId.Should().Be("rule id");
+            filterableResult.StartLine.Should().Be(111);
 
-            result.ProjectGuid.Should().BeNull();
-            result.WholeLineText.Should().Be("line text");
-            result.LineHash.Should().Be("line hash");
+            filterableResult.ProjectGuid.Should().BeNull();
+            filterableResult.WholeLineText.Should().Be("line text");
+            filterableResult.LineHash.Should().Be("line hash");
         }
     }
 }
--- a/src/Integration.Vsix.UnitTests/SonarLintDaemon/DaemonIssueAdapterTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintDaemon/DaemonIssueAdapterTests.cs
@@ -1,0 +1,75 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.Integration.Vsix;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintDaemon
+{
+    [TestClass]
+    public class DaemonIssueAdapterTests
+    {
+        [TestMethod]
+        public void Ctor_NullIssue_Throws()
+        {
+            Action act = () => new DaemonIssueAdapter(null, "line text", "hash");
+
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("sonarLintIssue");
+        }
+
+        [TestMethod]
+        public void Ctor_NullPathAndHash_NoException()
+        {
+            var result = new DaemonIssueAdapter(new Sonarlint.Issue(), null, null);
+
+            result.WholeLineText.Should().BeNull();
+            result.LineHash.Should().BeNull();
+        }
+
+        [TestMethod]
+        public void Ctor_ValidIssue()
+        {
+            // Arrange
+            var issue = new Sonarlint.Issue
+            {
+                FilePath = "path",
+                RuleKey = "rule id",
+                StartLine = 111
+            };
+
+
+            // Act
+            var result = new DaemonIssueAdapter(issue, "line text", "line hash");
+
+            // Assert
+            result.SonarLintIssue.Should().BeSameAs(issue);
+
+            result.FilePath.Should().Be("path");
+            result.RuleId.Should().Be("rule id");
+            result.StartLine.Should().Be(111);
+
+            result.ProjectGuid.Should().BeNull();
+            result.WholeLineText.Should().Be("line text");
+            result.LineHash.Should().Be("line hash");
+        }
+    }
+}
--- a/src/Integration.Vsix.UnitTests/SonarLintDaemon/DaemonIssueAdapterTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintDaemon/DaemonIssueAdapterTests.cs
@@ -56,7 +56,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintDaemon
                 StartLine = 111
             };
 
-
             // Act
             var result = new DaemonIssueAdapter(issue, "line text", "line hash");
 
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/IssueToFilterableIssueConverterTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/IssueToFilterableIssueConverterTests.cs
@@ -36,10 +36,10 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         [DataTestMethod]
         [DataRow(100, 1)]
         [DataRow(101, 100)]
-        public void CreateFilterableIssue_IssueLineOutsideSnapshot_ReturnsNull(int issueLine, int bufferLineCount)
+        public void CreateFilterableIssue_IssueLineOutsideSnapshot_ReturnsNull(int issueStartLine, int bufferLineCount)
         {
             // Arrange
-            var issue = new Sonarlint.Issue {StartLine = issueLine};
+            var issue = new Sonarlint.Issue {StartLine = issueStartLine};
             var mockSnapshot = CreateMockTextSnapshot(bufferLineCount, "unimportant");
 
             // Act and assert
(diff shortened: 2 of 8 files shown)


src/Integration.Vsix/SonarLintDaemon/DaemonIssueAdapter.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            SonarLintIssue = sonarLintIssue ?? throw new ArgumentNullException(nameof(sonarLintIssue));
+
+            WholeLineText = wholeLineText;
+            LineHash = lineHash;
+        }
+
+        public Sonarlint.Issue SonarLintIssue { get; }
+
+        #region ILiveIssue interface
```
rita-gorokhod-sonarsource: In general, I'm not a big fan of regions, I'm used to seeing it in badly written code.
I think it can be removed here, the class doesn't do much besides implementing the interface.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -294,6 +305,28 @@ private void UpdateIssues(IEnumerable<IssueMarker> issueMarkers)
             SnapToNewSnapshot(newSnapshot);
         }
 
+        internal /* for testing */ static Core.IFilterableIssue CreateFilterableIssue(Issue issue, ITextSnapshot textSnapshot)
```
rita-gorokhod-sonarsource: Extract to another class (like IssueToFilterableIssueConverter), this class already does too much.
It can also receive a list, instead of one Issue, and this way avoid the Where(x=> x!=null)

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -274,11 +275,21 @@ void IIssueConsumer.Accept(string path, IEnumerable<Issue> issues)
                 return;
             }
 
-            // TODO: collect the additional information required to compare be able to filter
-            // the issues (i.e. the text of the line and the line hash)
-            // issues = issuesFilter.Filter(path, issues);
+            // Collect the additional information required to compare be able to filter
+            // the issues, then apply the filter
+            var unfilteredIssues = issues.Select(x => CreateFilterableIssue(x, currentSnapshot))
```
rita-gorokhod-sonarsource: unfilteredIssues is a bit confusing, maybe "filterableIssues"?

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -274,11 +275,21 @@ void IIssueConsumer.Accept(string path, IEnumerable<Issue> issues)
...
+                .ToArray();
+
+            var filteredIssues = issuesFilter.Filter(path, unfilteredIssues);
+            Debug.Assert(filteredIssues.All(x => x is DaemonIssueAdapter), "Not expecting the issue filter to change the list item type");
 
-            var newMarkers = issues.Where(IsValidIssueTextRange).Select(CreateIssueMarker);
+            var suppressedCount = unfilteredIssues.Length - filteredIssues.Count();
+            logger.WriteLine(Strings.Daemon_SuppressedIssuesInfo, suppressedCount);
+
+            var newMarkers = filteredIssues.OfType<DaemonIssueAdapter>()
```
rita-gorokhod-sonarsource: I think this method is getting too big - I suggest extracting this to a method, something like:
IEnumerable<Issue> validIssues = GetNonSuppressedIssues(..)
var newMarkers = validIssues.Where(IsValidIssueTextRange).Select(CreateIssueMarker);

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -243,6 +244,36 @@ public void WhenNewIssuesAreFound_ListenersAreUpdated()
             actualMarkers[1].Issue.RuleKey.Should().Be("S333");
         }
 
+        [TestMethod]
+        public void WhenNewIssuesAreFound_IssueFilterIsApplied()
```
rita-gorokhod-sonarsource: Unrelated to this PR, but I think it's a good point that we decide on UTs naming conventions:
https://dzone.com/articles/7-popular-unit-test-naming

I do MethodName_StateUnderTest_ExpectedBehavior. 
I see that you do a variation of When_StateUnderTest_Expect_ExpectedBehavior.
I also like the popular JS Should_ExpectedBehavior_When_StateUnderTest.

Which one would you prefer? I'm cool with either, just so we will be consistent.

src/Integration.Vsix.UnitTests/SonarLintDaemon/DaemonIssueAdapterTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        [TestMethod]
+        public void Ctor_NullIssue_Throws()
+        {
+            Action act = () => new DaemonIssueAdapter(null, "line text", "hash");
+
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("sonarLintIssue");
+        }
+
+        [TestMethod]
+        public void Ctor_NullPathAndHash_NoException()
```
rita-gorokhod-sonarsource: Nitpicking: I think the behavior stated in the test's name is that one that should be asserted. So if it says "NoException" - then I'd expect to see act.Should().NotThrow();
Otherwise, I think it should be renamed to reflect the actual assertion, e.g. Ctor_NullPathAndHash_PathAndHashAreSetToNull, something like that.

src/Integration.Vsix.UnitTests/SonarLintDaemon/DaemonIssueAdapterTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            var result = new DaemonIssueAdapter(new Sonarlint.Issue(), null, null);
+
+            result.WholeLineText.Should().BeNull();
+            result.LineHash.Should().BeNull();
+        }
+
+        [TestMethod]
+        public void Ctor_ValidIssue()
```
rita-gorokhod-sonarsource: Following our naming convention discussion, I'd add the expected behavior, e.g. Ctor_ValidIssue_PropertiesInitializedWithIssueValues.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -243,6 +244,36 @@ public void WhenNewIssuesAreFound_ListenersAreUpdated()
...
+                new Sonarlint.Issue { RuleKey = "S222", StartLine = 2, EndLine = 2 },
+                new Sonarlint.Issue { RuleKey = "S333", StartLine = 10, EndLine = 20 },
+                new Sonarlint.Issue { RuleKey = "S444", StartLine = 100, EndLine = 101 }
+            };
+
+            // Set up the filter to return only one issue
+            var originalIssues = new List<IFilterableIssue>();
+            issuesFilter.Setup(x => x.Filter(It.IsAny<string>(), It.IsAny<IEnumerable<IFilterableIssue>>()))
+                .Callback((string path, IEnumerable<IFilterableIssue> issues) => originalIssues.AddRange(issues))
+                .Returns(originalIssues.Where(i => i.RuleId == "S222" || i.RuleId == "S444"));
```
rita-gorokhod-sonarsource: I think there are more conditions that we should check here, e.g.:
0/1/2 issues returned, null issue check, suppressions log, etc.
I think this will be much easier to test once it's in a separate class.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
duncanp-sonar: If I'm adding tests to an existing class (like here), I follow the naming pattern for the existing test.

I agree it would be worth standardising on naming for new test classes. I don't have a particular preference, as long as the names are not ridiculously long.

src/Integration.Vsix/SonarLintDaemon/DaemonIssueAdapter.cs
duncanp-sonar: It's here to make it clearer what isn't part of the interface. Explicitly implementing the interface would be an alternative, but I find the syntax for that ugly and harder to read.

rita-gorokhod-sonarsource: Oh, I didn't understand that at all. I think implementing the interface would be better.
Why do you think it's ugly? We can have a new interface for both this class and LiveIssue.

src/Integration.Vsix.UnitTests/SonarLintTagger/IssueToFilterableIssueConverterTests.cs
```diff
@@ -0,0 +1,162 @@
...
+
+namespace SonarLint.VisualStudio.Integration.UnitTests
+{
+    [TestClass]
+    public class IssueToFilterableIssueConverterTests
+    {
+        [DataTestMethod]
+        [DataRow(100, 1)]
+        [DataRow(101, 100)]
+        public void CreateFilterableIssue_IssueLineOutsideSnapshot_ReturnsNull(int issueLine, int bufferLineCount)
```
rita-gorokhod-sonarsource: really nitpicking: it took me some time to understand the params as opposed to the their names in the code; maybe rename them to "startLine" and "lineCount"?

src/Integration.Vsix/SonarLintTagger/IssueToFilterableIssueConverter.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.VisualStudio.Text;
+using Sonarlint;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Integration.Suppression;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
+{
+    internal static class IssueToFilterableIssueConverter
```
rita-gorokhod-sonarsource: I'm curious, why did you decide to go with a static class instead of interface?

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -274,14 +274,26 @@ void IIssueConsumer.Accept(string path, IEnumerable<Issue> issues)
...
-            var newMarkers = issues.Where(IsValidIssueTextRange).Select(CreateIssueMarker);
+            var newMarkers = filteredIssues.Where(IsValidIssueTextRange).Select(CreateIssueMarker);
             UpdateIssues(newMarkers);
         }
 
+        private IEnumerable<Issue> FilterIssues(string path, IEnumerable<Issue> issues)
+        {
+            var filterableIssues = IssueToFilterableIssueConverter.Convert(issues, currentSnapshot);
+
+            var filteredIssues = issuesFilter.Filter(path, filterableIssues);
```
rita-gorokhod-sonarsource: I realize now that we have a bit of naming problem:
1. The method is called FilterIssues, but IsValidIssueTextRange is not part of it
2. The class is called issuesFilter, so presumably we don't know what's the criteria, but then we write to log suppressedCount.

If we keep it "issuesFilter", then I think the  suppression log should be written inside the issuesFilter class. And maybe we should call IsValidIssueTextRange before filtering suppressions, since they are not valid anyway. Wdyt?

src/Integration.Vsix.UnitTests/SonarLintTagger/IssueToFilterableIssueConverterTests.cs
duncanp-sonar: If the naming is making is harder to understand then it's not nitpicking.
How about `issueStartLine` and `bufferLineCount`? For me, `startLine` and `lineCount` don't convey enough information - I have to read the test to see that they relate to different artefacts.

If something _really_ is just nitpicking then it's probably a waste of everyone's time mentioning it :-)

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -243,6 +248,98 @@ public void WhenNewIssuesAreFound_ListenersAreUpdated()
...
+            SetupIssuesFilter(issuesToReturnFromFilter, out var issuesPassedToFilter);
+
+            var errorListSinkMock1 = RegisterNewErrorListSink();
+
+            // Act
+            ((IIssueConsumer)testSubject).Accept(mockedJavascriptDocumentFooJs.Object.FilePath, inputIssues);
+
+            // Assert
+            // Check the expected issues were passed to the filter
+            issuesPassedToFilter.Count.Should().Be(2);
```
rita-gorokhod-sonarsource: Maybe setup the mock to the input, instead of It.IsAny? This way the assertion won't be needed.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -202,19 +198,28 @@ public void WhenNewIssuesAreFound_ButForIncorrectFile_ListenersAreNotUpdated()
...
+                new Issue { RuleKey = "S111", StartLine = 1, EndLine = 1 },
                 // The next issue is outside the range of the new snapshot and should be ignored
-                new Sonarlint.Issue { RuleKey = "S222", StartLine = 99999998, EndLine = 99999999 },
-                new Sonarlint.Issue { RuleKey = "S333", StartLine = 100, EndLine = 101 }
+                new Issue { RuleKey = "S222", StartLine = 99999998, EndLine = 99999999 },
+                new Issue { RuleKey = "S333", StartLine = 100, EndLine = 101 }
             };
 
+            // Setup up the filter to return whatever was supplied
+            var originalIssues = new List<IFilterableIssue>();
```
rita-gorokhod-sonarsource: Maybe call SetupIssuesFilter? It can receive optional output param, that if not specified, returns the original list.

src/Integration.Vsix.UnitTests/SonarLintTagger/IssueToFilterableIssueConverterTests.cs
rita-gorokhod-sonarsource: Yep, sounds good 👍

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
duncanp-sonar: I'll rename the method to `RemoveSuppressedIssues`.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
duncanp-sonar: Because the input isn't passed directly to the mock; it's converted to a list of `IFilteredIssue` first.

src/Integration.Vsix/SonarLintTagger/IssueToFilterableIssueConverter.cs
duncanp-sonar: It didn't seem worth the extra effort, although it would have made some of the tests slightly simpler.

Issue #1268, opened 2020-03-12
Implement suppression matching algorithm


duncanp-sonar:
### Description

- [x] decide on the issue matching algorithm. 
- [x] extract common issue matching code
- [x] implement and test for non-Roslyn issues
- [x] refactor Roslyn issue matching and tests

### Notes
The issue matching mechanism used by C#/VB is done one issue at a time; here, we have access to the whole file so we could potentially do better (e.g. removing server-side issues from the candidate list once they have been matched to stop the same server-side issue being matched to multiple client-side issues).

Update: start with simple algorithm first (line number/code/line hash), same as C#/VB.

PR #1278, opened 2020-03-18
Added issue filtering

duncanp-sonar:
Relates to #1268

### DIFF
--- a/src/Integration.UnitTests/Suppression/IssuesFilterTests.cs
+++ b/src/Integration.UnitTests/Suppression/IssuesFilterTests.cs
@@ -32,7 +32,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
     [TestClass]
     public class IssuesFilterTests
     {
-        private Mock<ISuppressionIssueMatcher> issueMatcher;
+        private Mock<ISuppressedIssueMatcher> issueMatcher;
         private IssuesFilter testSubject;
 
         private readonly IList<IFilterableIssue> matchedIssues = new List<IFilterableIssue>();
@@ -46,7 +46,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
         [TestInitialize]
         public void TestInitialize()
         {
-            issueMatcher = new Mock<ISuppressionIssueMatcher>();
+            issueMatcher = new Mock<ISuppressedIssueMatcher>();
             issueMatcher.Setup(x => x.SuppressionExists(It.IsAny<IFilterableIssue>()))
                 .Returns((IFilterableIssue i) => matchedIssues.Contains(i));
 
@@ -56,7 +56,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
         [TestMethod]
         public void Ctor_NullSonarQubeIssuesProvider_ThrowsArgumentNullException()
         {
-            Action act = () => new IssuesFilter((SuppressionIssueMatcher)null);
+            Action act = () => new IssuesFilter((SuppressedIssueMatcher)null);
 
             act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("issueMatcher");
         }
--- a/src/Integration.UnitTests/Suppression/IssuesFilterTests.cs
+++ b/src/Integration.UnitTests/Suppression/IssuesFilterTests.cs
@@ -23,6 +23,7 @@ using System.Collections.Generic;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
+using NuGet;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration.Suppression;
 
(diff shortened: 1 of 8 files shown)


duncanp-sonar:
This is the first of a couple of PRs to add suppression filtering.

This PR implements the issues filter so end-to-end suppressions for C++ issues now work (I haven't manually tested JavaScript yet).

The issue matching logic is currently the same as for Roslyn issues, so I added a new interface to encapsulate it. The next step will be to change the Roslyn code to use this interface, and then do any refactoring necessary to tidy things up.

src/Integration.UnitTests/Suppression/IssuesFilterTests.cs
```diff
@@ -31,39 +32,100 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
...
         {
-
             Action act = () => testSubject.Filter("c:\\path\\file1.txt", null);
             act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("issues");
         }
 
         [TestMethod]
         public void Filter_NullFile_DoesNotThrow()
         {
+            // TODO - remove. Path is not used so it can be removed from the interface
```
duncanp-sonar: I'll do this separately to keep the PRs simple.

src/Integration/Suppression/SuppressionIssueMatcher.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        public bool SuppressionExists(IFilterableIssue issue)
+        {
+            if (issue == null)
+            {
+                throw new ArgumentNullException(nameof(issue));
+            }
+
+            // Issues match if:
```
duncanp-sonar: This logic is the same as that in the `RoslynSuppressionHandler.cs`.

src/Integration.UnitTests/Suppression/SuppressionIssueMatcherTests.cs
```diff
@@ -0,0 +1,145 @@
...
+            Action act = () => testSubject.SuppressionExists(null);
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("issue");
+        }
+
+        [TestMethod]
+        public void MatchExists_NoServerIssues_ReturnsFalse()
+        {
+            // Arrange
+            var issueToMatch = CreateIssueToMatch("rule1", 1, "hash1");
+            ConfigureServerIssues(issueToMatch /* no server issues */);
```
rita-gorokhod-sonarsource: I think it'd be more readable to pass an empty list

src/Integration.UnitTests/Suppression/SuppressionIssueMatcherTests.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        [TestMethod]
+        public void MatchExists_NoMatches_ReturnsFalse()
+        {
+            // Arrange
+            var issueToMatch = CreateIssueToMatch("RightRuleId", 1, "RightHash");
+
+            ConfigureServerIssues(issueToMatch,
+                CreateServerIssue("WrongRuleId", 1, "RightHash"),  // wrong rule id
+                CreateServerIssue("RightRuleId", 2, "wrong hash"), // wrong line and hash
```
rita-gorokhod-sonarsource: Does the line make a difference? 
Maybe change it to be the same line, to emphasize that the hash is what's wrong.

src/Integration.UnitTests/Suppression/SuppressionIssueMatcherTests.cs
```diff
@@ -0,0 +1,145 @@
...
+                CreateServerIssue("RightRuleId", 3, "RIGHTHASH")); // wrong hash and wrong-case hash
```
rita-gorokhod-sonarsource: The comment is misleading, shouldn't it be just "wrong case hash"?

src/Integration.UnitTests/Suppression/SuppressionIssueMatcherTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        [TestMethod]
+        public void MatchExists_MatchingIdAndLine_ReturnsTrue()
+        {
+            // Arrange
+            var issueToMatch = CreateIssueToMatch("RightRuleId", 1, "RightHash");
+            ConfigureServerIssues(issueToMatch,
+                CreateServerIssue("wrong rule ID", 2, "RightHash"),
+                CreateServerIssue("RIGHTRULEID", 1, "WrongHash") // rule id comparison is case-insensitive -> match on line
```
rita-gorokhod-sonarsource: I'm not sure what is the test case here - maybe it deserves its own test?

src/Integration.UnitTests/Suppression/SuppressionIssueMatcherTests.cs
```diff
@@ -0,0 +1,145 @@
...
+                );
+
+            // Act and assert
+            testSubject.SuppressionExists(issueToMatch).Should().BeTrue();
+        }
+
+        [TestMethod]
+        public void MatchExists_MatchingIdAndLineHash_ReturnsTrue()
```
rita-gorokhod-sonarsource: I think it's worth separating the test cases into different test methods, for better readability. Every case here deserves its own test. It might also be worth doing DataTestMethod.
So something like 
MatchExists(string serverIssueRuleId, int serverIssueLineNumber, string serverIssueHash, bool shouldMatch)
[DataRow("right rule id", 1, "right hash", true)]
[DataRow("right rule id", 2, "right hash", true)]
[DataRow("right rule id", 1, "wrong hash", false)]

src/Integration.UnitTests/Suppression/SuppressionIssueMatcherTests.cs
duncanp-sonar: `wrong line and wrong-case hash`

src/Integration.UnitTests/Suppression/SuppressedIssueMatcherTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            // Act and assert
+            testSubject.SuppressionExists(issueToMatch).Should().BeFalse();
+        }
+
+        [DataTestMethod]
+        [DataRow("aaa", 222, "aaa hash", true)]
+        [DataRow("bbb", 333, "bbb hash", true)]
+        [DataRow("ccc", 444, "ccc hash", true)]
+        [DataRow("xxx", 111, "xxx hash", false)]
+        public void MatchExists_MultipleServerIssues_NoMatches_ReturnsFalse(string localRuleId, int localIssueLine, string localHash, bool expectedResult)
```
rita-gorokhod-sonarsource: Is the test name correct? it says "returns false" but most cases actually return true.
Maybe it should be MatchExists_MultipleServerIssues ?

duncanp-sonar: Yep, it should be `...MultipleServerIssues`

PR #1281, opened 2020-03-19
Trigger analysis on class initialization

### LINKED ISSUE #1279: Suppressions are not fetched until binding is triggered
### Description

Suppressions are not being fetched until the user navigates to Team Explorer -> SonarQube.

### Repro steps

1. Open a connected project that has suppressed issues
2. Don't navigate to SonarQube in Team explorer.
3. Open a file with suppressed issues.

### Expected behavior

Issue is suppressed.

### Actual behavior

Issue is not suppressed.

### Known workarounds

Navigate to the SonarQube binding pane.

### Related information

* SonarLint 4.20
* Visual Studio 2019

rita-gorokhod-sonarsource:
Fixes #1279 

@duncanp-sonar as we discussed offline, SuppressedIssuesProvider registers to IActiveSolutionBoundTracker's events too late.

However, there is actually another problem - ActiveSolutionBoundTracker in itself misses VS's event solutionTracker.ActiveSolutionChanged. Therefore OnActiveSolutionChanged is not being called, and events are not being fired. So even if SuppressedIssuesProvider would've been initialized on time, it still wouldn't have gotten the initial events since ActiveSolutionBoundTracker doesn't fire them. 
ActiveSolutionBoundTracker only executes OnImportsSatisfied, or raises events when a solution is re-opened or binding is updated. I suspect this is also what's contributing to the delays in C# that I've been experiencing.

In either case, right now by the time SuppressedIssuesProvider is called, ActiveSolutionBoundTracker is already initialized, so calling Refresh works.


src/Integration/Suppression/SuppressedIssuesProvider.cs
```diff
@@ -69,8 +69,15 @@ private static CreateProviderFunc GetCreateProviderFunc(ISonarQubeService sonarQ
             this.activeSolutionBoundTracker = activeSolutionBoundTracker ??
                                               throw new ArgumentNullException(nameof(activeSolutionBoundTracker));
 
+            if (activeSolutionBoundTracker.CurrentConfiguration == null)
+            {
+                throw new ArgumentNullException(nameof(activeSolutionBoundTracker.CurrentConfiguration));
```
rita-gorokhod-sonarsource: We get a code smell warning here, I'm not sure what to do with it.

duncanp-sonar: I don't think there is much reason to specifically check for a null configuration here - we don't check in line 88.

You could assert, but I'm not sure there's a lot of value in that either.

src/Integration.UnitTests/Suppression/SuppressedIssuesProviderTests.cs
```diff
@@ -118,26 +140,37 @@ public void GetSuppressedIssues_HasSolutionBinding_ConnectedMode_ListFromBinding
...
 
+        [DataTestMethod]
+        [DataRow(SolutionBindingEventType.SolutionBindingUpdated)]
+        [DataRow(SolutionBindingEventType.SolutionBindingChanged)]
+        public void SolutionBindingEvent_Standalone_SonarQubeIssuesProviderNotCreated(SolutionBindingEventType eventType)
+        {
+            var configuration = new BindingConfiguration(new BoundSonarQubeProject(), SonarLintMode.Standalone);
+            SimulateBindingEvent(eventType, configuration);
+
+            createProviderFunc.VerifyNoOtherCalls();
```
duncanp-sonar: Presumably we're not expecting the _SuppressionsEventRaised_ event to be raised in this case?

src/Integration.UnitTests/Suppression/SuppressedIssuesProviderTests.cs
```diff
@@ -65,17 +66,38 @@ public void Ctor_NullActiveSolutionBoundTracker_ArgumentNullException()
...
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("CurrentConfiguration");
+        }
+
+        [TestMethod]
+        public void Ctor_SolutionBindingIsStandalone_SonarQubeIssuesProviderNotCreated()
+        {
+            activeSolutionBoundTracker.CurrentConfiguration = BindingConfiguration.Standalone;
+            new SuppressedIssuesProvider(activeSolutionBoundTracker, createProviderFunc.Object);
+
+            createProviderFunc.VerifyNoOtherCalls();
```
duncanp-sonar: Presumably we're not expecting the _SuppressionsEventRaised_ event to be raised in this case?

rita-gorokhod-sonarsource: > The code looks good, although I'd probably just remove the check that is causing the code smell.
> 
> Is there a reason for not checking for the event being raised/not raised in the new tests?

Since it's part of the ctor, I was unable to test it -- since I can't register to the event before the class is created. Which means that AnalysisConfigMonitor can also miss it, similar to the race condition with ActiveSolutionBoundTracker -- it's possible that by the time we create SuppressedIssuesProvider, the binding event has already been raised, so we raise in CTOR, but AnalysisConfigMonitor hasn't registered to SuppressionsUpdateRequested event yet, and therefore the analysis is not triggered.
In reality this has not been an issue, since the blocking mechanism will save us (now that there is an initialization call in the CTOR).
So SuppressionsUpdateRequested event will only be raised when the binding is actively updated.

In hindsight, I probably shouldn't have gone with an event here, the whole flow is complicated enough as-is.

duncanp-sonar: Not to worry - more opportunities for refactoring later ;-)

PR #1282, opened 2020-03-19
Refactored Roslyn suppressions to use ISuppressedIssueMatcher

duncanp-sonar:
Relates to #1268

### DIFF
--- a/src/Integration.Vsix.UnitTests/Suppression/RoslynSuppressionHandlerTests.cs
+++ b/src/Integration.Vsix.UnitTests/Suppression/RoslynSuppressionHandlerTests.cs
@@ -19,7 +19,6 @@
  */
 
 using System;
-using System.Collections.Generic;
 using System.Linq.Expressions;
 using FluentAssertions;
 using Microsoft.CodeAnalysis;
@@ -29,7 +28,6 @@ using Moq;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration.Suppression;
 using SonarLint.VisualStudio.Integration.Vsix.Suppression;
-using SonarQube.Client.Models;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
 {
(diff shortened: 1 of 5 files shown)


src/Integration.Vsix/Suppression/RoslynSuppressionHandler.cs
```diff
@@ -60,19 +51,8 @@ public bool ShouldIssueBeReported(SyntaxTree syntaxTree, Diagnostic diagnostic)
...
-            // Retrieve all issues relating to this file (file level or precise location) or project (for module level issues)
-            var potentialMatchingIssues = serverIssuesProvider.GetSuppressedIssues(liveIssue.ProjectGuid, liveIssue.FilePath);
-
-            // Try to find an issue with the same ID and either the same line number or some line hash
-            bool matchFound = potentialMatchingIssues
-                .Where(i => StringComparer.OrdinalIgnoreCase.Equals(liveIssue.RuleId, i.RuleId))
-                .Any(i => liveIssue.StartLine == i.Line || StringComparer.Ordinal.Equals(liveIssue.LineHash, i.Hash));
-
-            return !matchFound;
+            var matchFound = !issueMatcher.SuppressionExists(liveIssue);
```
duncanp-sonar: The matching logic hasn't changed, it's just been moved to IssueMatcher.

src/Integration/Suppression/SuppressedIssueMatcher.cs
```diff
@@ -24,7 +24,7 @@
 
 namespace SonarLint.VisualStudio.Integration.Suppression
 {
-    internal class SuppressedIssueMatcher : ISuppressedIssueMatcher
+    public class SuppressedIssueMatcher : ISuppressedIssueMatcher
```
duncanp-sonar: This class is in the `Integration` project so it needs to be public to be visible to the `RoslynAnalyzerManager` in the `Integration.Vsix` project.

I'll move some of the non-VS dependent classes to the `Core` project in a separate commit.

src/Integration.Vsix.UnitTests/Suppression/RoslynSuppressionHandlerTests.cs
```diff
@@ -101,134 +99,25 @@ public void ShouldReport_CannotCreateLiveIssue_ReturnsTrue()
...
             bool result = handler.ShouldIssueBeReported(syntaxTreeMock.Object, diag);
 
             // Assert
             VerifyLiveIssueCreated(Times.Once());
             VerifyServerIssuesRequested(Times.Once());
-            result.Should().BeTrue();
-        }
-
-        [TestMethod]
-        public void ShouldReport_NoMatchingServerIssues_ReturnsTrue()
```
duncanp-sonar: The tests are much simpler since they are now only checking for a true/false from the IssueMatcher.

These tests were replaced the the IssueMatcherTests in the previous commit.

PR #1286, opened 2020-03-19
Fix file-level issue comparison bug

duncanp-sonar:
File-level issues should not match non-file-level issues

### DIFF
--- a/src/Integration/Suppression/SuppressedIssueMatcher.cs
+++ b/src/Integration/Suppression/SuppressedIssueMatcher.cs
@@ -59,20 +59,21 @@ namespace SonarLint.VisualStudio.Integration.Suppression
 
             return matchFound;
         }
+
         private static bool IsMatch(IFilterableIssue issue, SonarQubeIssue serverIssue)
         {
             if (!StringComparer.OrdinalIgnoreCase.Equals(issue.RuleId, serverIssue.RuleId))
             {
                 return false;
             }
-            if (issue.StartLine.HasValue)
-            {
-                return issue.StartLine == serverIssue.Line || StringComparer.Ordinal.Equals(issue.LineHash, serverIssue.Hash);
-            }
-            else
+
+            if (!issue.StartLine.HasValue) // i.e. file-level issue
             {
                 return !serverIssue.Line.HasValue;
             }
+
+            // Non-file level issue
+            return issue.StartLine == serverIssue.Line || StringComparer.Ordinal.Equals(issue.LineHash, serverIssue.Hash);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Integration/Suppression/SuppressedIssueMatcher.cs
```diff
@@ -40,19 +41,38 @@ public bool SuppressionExists(IFilterableIssue issue)
                 throw new ArgumentNullException(nameof(issue));
             }
 
-            // Issues match if:
+
+            // File-level issues (i.e. line = null) match if:
+            // 1. Same component, same file, same error code.
```
duncanp-sonar: I noticed this bug when looking at the Roslyn `LiveIssueFactory`. 

File-level issues have `null` for the `hash` and the line number, and they should not match any non-file-level issues.

This bug existed in the previous code, but no-one noticed/complained.

src/Integration/Suppression/SuppressedIssueMatcher.cs
```diff
@@ -40,19 +41,38 @@ public bool SuppressionExists(IFilterableIssue issue)
...
         }
+        private static bool IsMatch(IFilterableIssue issue, SonarQubeIssue serverIssue)
+        {
+            if (!StringComparer.OrdinalIgnoreCase.Equals(issue.RuleId, serverIssue.RuleId))
+            {
+                return false;
+            }
+            if (issue.StartLine.HasValue)
+            {
+                return issue.StartLine == serverIssue.Line || StringComparer.Ordinal.Equals(issue.LineHash, serverIssue.Hash);
```
rita-gorokhod-sonarsource: I'm really confused about the OR - I know it was in the code before, but if it's same line and different hash, shouldn't it be false?

duncanp-sonar: No. If the local analysis and the server both report the same issue on the same line then we assume they refer to the same.
If the hash is different it means the text of the line has changed, which could have multiple causes - whitespace, comments, variable renaming etc. However, in spite of whatever changed on the line, the local and server analysis are both reporting issue X on line Y.

src/Integration/Suppression/SuppressedIssueMatcher.cs
```diff
@@ -40,19 +41,38 @@ public bool SuppressionExists(IFilterableIssue issue)
...
 
             return matchFound;
         }
+        private static bool IsMatch(IFilterableIssue issue, SonarQubeIssue serverIssue)
+        {
+            if (!StringComparer.OrdinalIgnoreCase.Equals(issue.RuleId, serverIssue.RuleId))
+            {
+                return false;
+            }
+            if (issue.StartLine.HasValue)
```
rita-gorokhod-sonarsource: I think it's worth inverting the IF, it creates a bit of mental work.
It might also help to create a small helper method,
if (IsFileIssue(issue)) 
{
  return IsFileIssue(serverIssue);
}
return ....

duncanp-sonar: I think creating two helper methods (one for each issue type) is overkill.
I've inverted the `if`.

PR #1291, opened 2020-03-24
Renamed UserSettings to RulesSettings

duncanp-sonar:


src/Core/IUserSettingsProvider.cs
```diff
@@ -32,7 +32,7 @@ public interface IUserSettingsProvider
         /// <summary>
         /// The settings for the current user
         /// </summary>
-        UserSettings UserSettings { get; }
+        RulesSettings UserSettings { get; }
```
duncanp-sonar: This now looks a bit weird, since the settings provider now returns the `RuleSettings` directly.

I'll add a new `UserSettings` class in the next commit to make the separation cleaner. The new user settings will just be a wrapper that contains the RulesSettings:

```
    public class UserSettings
    {
        public UserSettings(RulesSettings rulesSettings)
        {
            this.RulesSettings = rulesSettings ?? throw new ArgumentNullException(nameof(rulesSettings));
        }

        public RulesSettings RulesSettings { get; }
    }
```

duncanp-sonar:
No product changes, just lots of renaming.

PR #1293, opened 2020-03-24
Added UserSettings class

duncanp-sonar:


src/Integration.Vsix/Analysis/UserSettingsProvider.cs
```diff
@@ -96,21 +96,21 @@ public void EnsureFileExists()
...
             {
-                serializer.SafeSave(SettingsFilePath, UserSettings);
+                serializer.SafeSave(SettingsFilePath, UserSettings.RulesSettings);
             }
         }
 
         #endregion
 
-        private RulesSettings SafeLoadUserSettings(string filePath, ILogger logger)
+        private UserSettings SafeLoadUserSettings(string filePath, ILogger logger)
```
rita-g-sonarsource: I'm unsure about the UserSettings class. Do you think it will evolve to contain something other than rules? If not, then maybe we should rename this class to RuleSettingsProvider and have it return RuleSettings directly (like before).

duncanp-sonar: I'm not sure. I can think of things it could contain (e.g. a list of the languages the user wants to analyse), but there's no guarantee that we will add anything else.
I considered renaming this class RulesSettingsProvider, but felt that it was confusing - we have too many classes with similar names to do with rules, settings, and configuration, and it's not clear how they relate to each other.
At least this way the provider name gives an indication of why it is logically different - it returns per-user settings.

I'll leave it for now and we can revisit it later if necessary.

Issue #1024, opened 2019-10-08
C++ rule enablement: do not allow project-level rules to be enabled


duncanp-sonar:
### Description

Currently some project-level C++ rules are disabled by default in SonarLint (i.e. rules that look at more than a single file). The rule-enablement UI should not list these rules.

List of disabled rules:
~~https://github.com/SonarSource/sonar-cpp/blob/00bf8bfa95c7eadbbf3aed2c092ba81b147b88e8/sonar-cfamily-plugin/src/main/java/com/sonar/cpp/plugin/AbstractRulesDefinition.java#L30~~
* S5536

abbas-sabra-sonarsource:
@duncanp-sonar @rita-gorokhod-sonarsource I'm updating the list of project-level rules:https://github.com/SonarSource/sonar-cpp/pull/839/files#diff-1c8c8ae2888c5848a89ff459b2843418R30

- We currently have only one: S5536
- The first three used to be project-level but they stop after we moved to clang front-end.
- I will create a "new issue" every time we add a new project-level rule to the list. There is no plan to add soon.
- S5536 is not clang-based:
> CLang-based rule S5536 should also be disabled.

duncanp-sonar:
The list of unavailable rules is shown in the SonarLint output window:

![image](https://user-images.githubusercontent.com/10757559/77671206-29245c80-6f7f-11ea-9bf7-3eb037e41521.png)

PR #1294, opened 2020-03-24
Refactor effective rules calculator

duncanp-sonar:
Relates to #1024

### DIFF
--- a/src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
+++ b/src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
@@ -18,11 +18,13 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.Collections.Generic;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Core.CFamily;
 using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.Integration.UnitTests.CFamily;
 
 namespace SonarLint.VisualStudio.Core.UnitTests.CFamily
 {
@@ -32,20 +34,23 @@ namespace SonarLint.VisualStudio.Core.UnitTests.CFamily
         [TestMethod]
         [DataRow(true /* custom settings are null */)]
         [DataRow(false /* custom settings are not null, but emtpy */)]
-        public void GetConfig_NullCustomSettings_DefaultConfigReturned(bool sourceCustomSettingsAreNull)
+        public void GetConfig_NoCustomSettings_NoError_DefaultsReturned(bool customSettingsAreNull)
         {
             // Arrange
             var testLogger = new TestLogger();
             var testSubject = new EffectiveRulesConfigCalculator(testLogger);
 
-            var defaultRulesConfig = CreateMockConfig("language1");
-            var sourcesSettings = sourceCustomSettingsAreNull ? null : new RulesSettings();
+            var defaultRulesConfig = CreateWellKnownRulesConfig("language1");
+            var sourcesSettings = customSettingsAreNull ? null : new RulesSettings();
 
             // Act
             var result = testSubject.GetEffectiveRulesConfig("language1", defaultRulesConfig, sourcesSettings);
 
-            // Assert - optimisation - expecting the same object instance
-            result.Should().BeSameAs(defaultRulesConfig);
+            // Assert
+            result.Should().NotBeSameAs(defaultRulesConfig);
+            result.LanguageKey.Should().Be("language1");
+            result.AllPartialRuleKeys.Should().BeEquivalentTo(defaultRulesConfig.AllPartialRuleKeys);
+
             testLogger.AssertOutputStringExists(CoreStrings.EffectiveRules_NoCustomRulesSettings);
         }
 
@@ -56,7 +61,33 @@ namespace SonarLint.VisualStudio.Core.UnitTests.CFamily
             var testLogger = new TestLogger();
             var testSubject = new EffectiveRulesConfigCalculator(testLogger);
 
-            var defaultRulesConfig = CreateMockConfig("key");
+            var defaultRulesConfig = CreateWellKnownRulesConfig("key");
+            var sourcesSettings = new RulesSettings
+            {
+                Rules = new Dictionary<string, RuleConfig>
+                {
+                    // Turn an active rule off...
+                    { "key:" + WellKnownPartialRuleKey1_Active, new RuleConfig { Level = RuleLevel.Off } },
+                    // ... and an inactive rule one
+                    { "key:" + WellKnownPartialRuleKey3_Inactive, new RuleConfig { Level = RuleLevel.On } }
+                }
+            };
+
+            var result = testSubject.GetEffectiveRulesConfig("language1", defaultRulesConfig, sourcesSettings);
+
+            result.LanguageKey.Should().Be("key");
+            result.AllPartialRuleKeys.Should().BeEquivalentTo(defaultRulesConfig.AllPartialRuleKeys);
+            result.ActivePartialRuleKeys.Should().BeEquivalentTo(WellKnownPartialRuleKey2_Active, WellKnownPartialRuleKey3_Inactive);
+        }
+
+        [TestMethod]
+        public void GetConfig_CachedResultsReturnedIfAvailable()
+        {
+            // Arrange
+            var testLogger = new TestLogger();
+            var testSubject = new EffectiveRulesConfigCalculator(testLogger);
+
+            var defaultRulesConfig = CreateWellKnownRulesConfig("key");
             var sourcesSettings = new RulesSettings
             {
                 Rules = new System.Collections.Generic.Dictionary<string, RuleConfig>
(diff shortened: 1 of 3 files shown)


src/Core/CFamily/EffectiveRulesConfigCalculator.cs
```diff
@@ -64,32 +64,34 @@ public EffectiveRulesConfigCalculator(ILogger logger)
...
             }
             else
             {
                 logger.WriteLine(CoreStrings.EffectiveRules_CacheMiss);
-                effectiveConfig = new DynamicCFamilyRulesConfig(defaultRulesConfig, customRulesConfiguration);
-                configCache.Add(languageKey, defaultRulesConfig, customRulesConfiguration, effectiveConfig);
             }
 
+            // TODO: apply filters to exclude which should not be run
+            effectiveConfig = new DynamicCFamilyRulesConfig(defaultRulesConfig, customSettings);
```
duncanp-sonar: The purpose of this change is so that all code paths that involve cache misses reach this point.
That gives a us single location in which to apply the project rule exclusions (either in this class or the `DynamicCFamilyRulesConfig` class.

src/Core/CFamily/EffectiveRulesConfigCalculator.cs
```diff
@@ -64,32 +64,34 @@ public EffectiveRulesConfigCalculator(ILogger logger)
...
-            if (effectiveConfig != null)
+            if ((customSettings?.Rules?.Count ?? 0) == 0)
             {
-                logger.WriteLine(CoreStrings.EffectiveRules_CacheHit);
+                logger.WriteLine(CoreStrings.EffectiveRules_NoCustomRulesSettings);
+                customSettings = new RulesSettings();
             }
             else
             {
                 logger.WriteLine(CoreStrings.EffectiveRules_CacheMiss);
```
rita-g-sonarsource: This should be logged regardless of custom rules, no?

src/Core/CFamily/EffectiveRulesConfigCalculator.cs
```diff
@@ -64,32 +64,34 @@ public EffectiveRulesConfigCalculator(ILogger logger)
...
             {
-                logger.WriteLine(CoreStrings.EffectiveRules_NoCustomRulesSettings);
-                return defaultRulesConfig;
+                logger.WriteLine(CoreStrings.EffectiveRules_CacheHit);
+                return effectiveConfig;
             }
 
-            var effectiveConfig = configCache.FindConfig(languageKey, defaultRulesConfig, customRulesConfiguration);
-            if (effectiveConfig != null)
+            if ((customSettings?.Rules?.Count ?? 0) == 0)
```
rita-g-sonarsource: I think this IF is redundant now.
We will always add to cache now, so I think it makes more sense to have this check inside DynamicCFamilyRulesConfig, or in the caller that passes customSettings.

src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
```diff
@@ -32,20 +34,23 @@ public class EffectiveRulesConfigCalculatorTests
...
+            var defaultRulesConfig = CreateWellKnownRulesConfig("language1");
+            var sourcesSettings = customSettingsAreNull ? null : new RulesSettings();
 
             // Act
             var result = testSubject.GetEffectiveRulesConfig("language1", defaultRulesConfig, sourcesSettings);
 
-            // Assert - optimisation - expecting the same object instance
-            result.Should().BeSameAs(defaultRulesConfig);
+            // Assert
+            result.Should().NotBeSameAs(defaultRulesConfig);
```
rita-g-sonarsource: Does it matter that it's not the same instance ref?
I think it's a bit confusing. The important assertion is the rule keys below, so I'd keep that one and remove this one.

src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
```diff
@@ -56,7 +61,33 @@ public void GetConfig_RulesInCustomSettings_MergedConfigReturned()
...
 
-            var defaultRulesConfig = CreateMockConfig("key");
+            var defaultRulesConfig = CreateWellKnownRulesConfig("key");
+            var sourcesSettings = new RulesSettings
+            {
+                Rules = new Dictionary<string, RuleConfig>
+                {
+                    // Turn an active rule off...
+                    { "key:" + WellKnownPartialRuleKey1_Active, new RuleConfig { Level = RuleLevel.Off } },
+                    // ... and an inactive rule one
```
rita-g-sonarsource: one = on?

src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
```diff
@@ -32,20 +34,23 @@ public class EffectiveRulesConfigCalculatorTests
         [TestMethod]
         [DataRow(true /* custom settings are null */)]
         [DataRow(false /* custom settings are not null, but emtpy */)]
-        public void GetConfig_NullCustomSettings_DefaultConfigReturned(bool sourceCustomSettingsAreNull)
+        public void GetConfig_NoCustomSettings_NoError_DefaultsReturned(bool customSettingsAreNull)
         {
             // Arrange
             var testLogger = new TestLogger();
```
rita-g-sonarsource: Please extract the repetitions to TestInitialize

src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
```diff
@@ -56,7 +61,33 @@ public void GetConfig_RulesInCustomSettings_MergedConfigReturned()
...
+
+            var result = testSubject.GetEffectiveRulesConfig("language1", defaultRulesConfig, sourcesSettings);
+
+            result.LanguageKey.Should().Be("key");
+            result.AllPartialRuleKeys.Should().BeEquivalentTo(defaultRulesConfig.AllPartialRuleKeys);
+            result.ActivePartialRuleKeys.Should().BeEquivalentTo(WellKnownPartialRuleKey2_Active, WellKnownPartialRuleKey3_Inactive);
+        }
+
+        [TestMethod]
+        public void GetConfig_CachedResultsReturnedIfAvailable()
```
rita-g-sonarsource: I don't think that we should rely on the logger as an indication of cache.
A better way IMO would be to change the sourceSettings on the second call, and to see that the changes are not being taken into effect.
I'd also remove assertions that have already been made in other tests, such as testing the results of call #1 - it was already tested in 2 tests above.

src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
```diff
@@ -165,11 +196,18 @@ public void Cache_MultipleEntries()
 
         #endregion Cache tests
 
-        private static ICFamilyRulesConfig CreateMockConfig(string languageKey)
+        internal const string WellKnownPartialRuleKey1_Active = "rule1";
+        internal const string WellKnownPartialRuleKey2_Active = "rule2";
+        internal const string WellKnownPartialRuleKey3_Inactive = "rule3";
+
+        private static ICFamilyRulesConfig CreateWellKnownRulesConfig(string languageKey)
```
rita-g-sonarsource: Maybe move it to TestInitialize and store it as a member, It's a bit easier IMO to follow it this way.

src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
```diff
@@ -32,20 +34,23 @@ public class EffectiveRulesConfigCalculatorTests
         [TestMethod]
         [DataRow(true /* custom settings are null */)]
         [DataRow(false /* custom settings are not null, but emtpy */)]
-        public void GetConfig_NullCustomSettings_DefaultConfigReturned(bool sourceCustomSettingsAreNull)
+        public void GetConfig_NoCustomSettings_NoError_DefaultsReturned(bool customSettingsAreNull)
```
rita-g-sonarsource: what do you mean by "no error"?

src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
duncanp-sonar: Well, it's part of checking that the code goes through a single point before returning. Previously the optimisation meant the original object would be returned, but that definitely isn't what we want now. Still, it's not a strong check so I'll remove it if it's confusing.

src/Core/CFamily/EffectiveRulesConfigCalculator.cs
duncanp-sonar: If we log it in the `CFamilyRulesConfigProvider` it will be logged for every analysis not just when a  new cache entry is being created, which is noisy. The `DynamicCFamilyRulesConfig` doesn't currently log anything, so moving the logging there would mean further changes that are pretty low value. 

This class doesn't know where it's being called from it needs to do some sort of check: an assert, an exception, or defensive coding. The same applies to `DynamicCFamilyRulesConfig`.

src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
duncanp-sonar: The test is checking the object identity to see if the cache is working.
The logger is being checked to see that the notifications are correct.

src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
duncanp-sonar: I've moved the two common lines of setup code into `TestInitialize`, and the test that relate to the inner cache class to a different file `EffectiveRulesConfigCalculatorCacheTests`.

On balance, though, I still agree with [Mr Newkirk's reasons](https://jamesnewkirk.typepad.com/posts/2007/09/why-you-should-.html) for avoiding `TestInitialize`.

src/Core.UnitTests/CFamily/EffectiveRulesConfigCalculatorTests.cs
duncanp-sonar: It doesn't throw. I'll remove it.

PR #1313, opened 2020-04-06
Detect cl.exe version correctly

### LINKED ISSUE #1289: C++ compiler "CL.exe" Minor and Macro version aren't detected correctly
### Description

`cl.exe` version has a Major, Minor, and Micro values. Currently, only the major part is detected correctly. The problem is that only the platform toolset is used(`ConvertPlatformToolset`) to detect the version while different compiler versions can be used with the same toolset. The compiler version should be detected directly instead of deducing it from the platform.

### Repro steps
Install 19.23.28107 c++ compiler (or any compiler other than the hardcoded ones in `ConvertPlatformToolset`)  on v142 and check that the detected version will always be 19.20.00.

### Expected behavior
The installed version of the compiler is detected. 
The two macros defined later on based on the compiler version should be set to the right value:
```
predefines.Append("#define _MSC_FULL_VER ").Append(probe.Major).Append(probe.Minor).Append(probe.Micro).Append("\n");
predefines.Append("#define _MSC_VER ").Append(probe.Major).Append(probe.Minor).Append("\n");
```

### Actual behavior

The compiler version is always set to a hardcoded value which leads to wrong values of the two mentioned macros.

### Known workarounds

No workaround

#### Questions
Can the compiler version change per project? Yes, but not straightforward. Corner case - maybe ignore?

abbas-sabra-sonarsource:
Fixes #1289

### DIFF
--- a/src/Integration.Vsix/CFamily/FileConfig.cs
+++ b/src/Integration.Vsix/CFamily/FileConfig.cs
@@ -44,13 +44,14 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
                 dynamic file = projectItem.Object; // Microsoft.VisualStudio.VCProjectEngine.VCFile
                 dynamic fileConfig = file.FileConfigurations.Item(configurationName); // Microsoft.VisualStudio.VCProjectEngine.VCFileConfiguration
                 dynamic fileTool = fileConfig.Tool; // Microsoft.VisualStudio.VCProjectEngine.VCCLCompilerTool
+                string platformToolset = config.Rules.Item("ConfigurationGeneral").GetEvaluatedPropertyValue("PlatformToolset");
 
                 return new FileConfig()
                 {
                     AbsoluteFilePath = absoluteFilePath,
 
                     PlatformName = platformName,
-                    PlatformToolset = config.Rules.Item("ConfigurationGeneral").GetEvaluatedPropertyValue("PlatformToolset"),
+                    PlatformToolset = platformToolset,
 
                     IncludeDirectories = config.Rules.Item("ConfigurationDirectories").GetEvaluatedPropertyValue("IncludePath"),
                     AdditionalIncludeDirectories = GetEvaluatedPropertyValue(fileTool, "AdditionalIncludeDirectories"),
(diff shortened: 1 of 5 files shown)


src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -247,32 +251,41 @@ public Request ToRequest(string path)
...
                     case "": // Bug https://github.com/SonarSource/sonarlint-visualstudio/issues/804
                         throw new ArgumentException(Strings.Daemon_PlatformToolsetNotSpecified);
+                    // VCToolsVersion is only avaialble from VS2017 onward
+                    // Before VS2017 the platform was enough to deduce the compiler version
                     case "v142":
-                        return "19.20.00";
                     case "v141":
                     case "v141_xp":
-                        return "19.10.00";
+                        int index = vcToolsVersion.IndexOf('.');
```
rita-g-sonarsource, line 280: @abbas-sabra-sonarsource I tested this on VS2015 -- the platform toolset is V140 and lower, so this code is not called. 
Would there be any way to have a toolset higher than that? To make sure that we're not getting a null exception.

src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -247,32 +251,41 @@ public Request ToRequest(string path)
...
-                        return "x86";
+                        return false;
                     case "x64":
-                        return "x64";
+                        return true;
                 }
             }
 
-            internal /* for testing */ static string ConvertPlatformToolset(string platformToolset)
+            internal /* for testing */ static string getCompilerVersion(string platformToolset, string vcToolsVersion)
```
rita-g-sonarsource, line 267: Please rename to PascalCase :) (and isPlatformX64 as well)

src/Integration.Vsix/CFamily/FileConfig.cs
abbas-sabra-sonarsource: @rita-g-sonarsource I confirm the new code will not be hit on VS2015. That is expected because there is only 19.00.00 for v140 and v140_xp.
No, it isn't possible to set a higher toolset.

PR #1319, opened 2020-04-09
CFamily - cancel in-process analysis if a new analysis is triggered for the same file

### LINKED ISSUE #1309: CFamily - cancel in-process analysis if a new analysis is triggered for the same file
### Description
It's possible to trigger a new analysis before the previous analysis has been finished. The previous analysis execution is not cancelled, and both analysis run in parallel.
This might lead to 2 problems:
1. Issue correctness: if the first analysis takes longer than the second analysis, we might display issues that are no longer relevant.
2. Machine resources: the first analysis is redundant and consumes resources.

### Proposed solution
Once a new request is made, cancel the previous one. We can queue the analysis requests in a thread-safe queue, based on file name. 

## Testing scenarios
- First analysis has not started yet
- First analysis has started, but not yet finished
- First analysis has finished, but issues not yet reported

## Notes
- Should be tested in VS 2015 / 2017 / 2019 due to different threading models.

rita-g-sonarsource:
Fixes #1309


src/Integration.Vsix.UnitTests/CFamily/ProcessRunnerTests.cs
```diff
@@ -409,11 +410,94 @@ public void ProcRunner_DoNotLogSensitiveData()
...
+            await runProcess;
+
+            result.Should().BeFalse("Expecting the process to have failed");
+            runner.ExitCode.Should().Be(-1, "Unexpected exit code");
+            logger.AssertOutputStringExists("Hello world");
+            logger.AssertPartialOutputStringDoesNotExist("Done!");
+        }
+
+        [TestMethod]
+        public void Execute_CancellationTokenCancelledAfterProcessAlreadyFinished_DoesNotThrow()
```
rita-g-sonarsource, line 476: @duncanp-sonar there are actually 2 scenarios here, and I only test the first:
1. Process finished completely, and then Cancel is called --> Cancel callback won't be executed at all, since the Register callback is disposed (which is what happens in this test).
2. Process finished, but Register is not yet disposed.

I don't know how to test for the second one. Got an idea? :/ 
3. Process

rita-g-sonarsource:
@duncanp-sonar I'm working on tests for TaggerProvider - if you got time, let me know please what you think about the overall approach.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -382,6 +382,7 @@
          This needs to be set after the SDK targets have been imported.
     -->
     <PrepareForBuildDependsOn>$(PrepareForBuildDependsOn);GetCFamilyFiles</PrepareForBuildDependsOn>
+    <LangVersion>latest</LangVersion>
```
duncanp-sonar, line 385: Could this cause the same sort of problem as we saw with the most recent version of sonar-dotnet i.e. using language features that aren't supported by the target runtime?

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -136,9 +141,22 @@ public bool Execute(ProcessRunnerArguments runnerArgs)
...
+                try
+                {
+                    if (process != null && !process.HasExited)
+                    {
+                        process.Kill();
+                    }
+                }
+                catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+                {
+                }
```
duncanp-sonar, line 157: Is there a reason not to log the error message?

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -92,6 +96,7 @@ internal sealed class TaggerProvider : IViewTaggerProvider, ITableDataSource, IS
             this.dte = serviceProvider.GetService<DTE>();
             this.languageRecognizer = languageRecognizer;
             this.logger = logger;
+            this.analysisJobs = new Dictionary<string, CancellationTokenSource>();
```
duncanp-sonar, line 99: It would be safer to explicitly specify the comparer as being `OrdinalIgnoreCase`.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -181,14 +186,34 @@ public void RequestAnalysis(string path, string charset, IEnumerable<AnalysisLan
...
+        private CancellationToken GetCancellationToken(string path)
+        {
+            CancellationToken token;
+            lock (analysisJobs)
+            {
+                if (analysisJobs.ContainsKey(path))
+                {
+                    logger.WriteLine($"Cancelled analysis for {path}");
+                    analysisJobs[path].Cancel(throwOnFirstException: false);
+                    analysisJobs[path].Dispose();
```
duncanp-sonar, line 207: What happens to the cancellation source if the analysis runs to completion? It just stays in the dictionary until the next analysis is triggered for that file?

rita-g-sonarsource: > What happens to the cancellation source if the analysis runs to completion? It just stays in the dictionary until the next analysis is triggered for that file?

Yep. The last analysis for each file will be kept. 
I didn't think it was a problem, but now that you mention it, it will blow up with prolonged work. I will fix it.

src/Integration.Vsix.UnitTests/CFamily/ProcessRunnerTests.cs
duncanp-sonar: Not really. You don't have a way of pausing the test code between those two points.
You could partially test some of the code by making the `KillProcess` methods `internal static` and testing it.

src/Integration.Vsix/CFamily/ProcessRunner.cs
rita-g-sonarsource: It's possible that the process has exited just between the IF check and before Kill() is called, in which case an exception is thrown. I think logging it will create un-necessary noise.

src/Integration.Vsix/SonarLintDaemon/DaemonAnalyzer.cs
```diff
@@ -140,7 +143,7 @@ private void MakeRequest()
             {
                 daemon.Ready -= HandleDaemonReady;
                 daemonInstaller.InstallationCompleted -= HandleInstallCompleted;
-                daemonAnalyzer.InvokeDaemon(path, charset, detectedLanguages, consumer, projectItem);
+                daemonAnalyzer.InvokeDaemon(path, charset, detectedLanguages, consumer, projectItem, CancellationToken.None);
```
duncanp-sonar, line 146: Why isn't the cancellation token passed through to this class?

src/Integration.Vsix/SonarLintTagger/AnalysisScheduler.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Threading;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
+{
+    [Export(typeof(IAnalysisScheduler))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class AnalysisScheduler : IAnalysisScheduler
+    {
+        private readonly ILogger logger;
+        private readonly IDictionary<string, CancellationTokenSource> analysisJobs;
```
duncanp-sonar, line 33: Out of interest, did you consider using a thread-safe `ConcurrentDictionary` instead a normal dictionary with locking?

src/Integration.Vsix/SonarLintTagger/AnalysisScheduler.cs
```diff
@@ -0,0 +1,84 @@
...
+
+                return newAnalysis;
+            }
+        }
+
+        private void CancelPreviousAnalysis(string filePath)
+        {
+            if (analysisJobs.ContainsKey(filePath))
+            {
+                logger.WriteLine($"Cancelled analysis for {filePath.ToLower()}");
```
duncanp-sonar, line 75: Why change the case when displaying the message? The original casing is probably what the user would expect to see.

src/Integration.Vsix/SonarLintTagger/AnalysisScheduler.cs
```diff
@@ -0,0 +1,84 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Threading;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
+{
+    [Export(typeof(IAnalysisScheduler))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class AnalysisScheduler : IAnalysisScheduler
```
duncanp-sonar, line 30: It isn't obvious what this class is for. A doc comment would be useful.

src/Integration.Vsix/SonarLintTagger/AnalysisScheduler.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            using (var newAnalysisToken = IssueToken(filePath))
+            {
+                analyzeAction(newAnalysisToken.Token);
+
+                lock (analysisJobs)
+                {
+                    if (analysisJobs[filePath] == newAnalysisToken)
+                    {
+                        analysisJobs[filePath] = null;
```
duncanp-sonar, line 52: Do entries ever get removed from the dictionary, or does it continually accumulate keys with null values?

src/Integration.Vsix/SonarLintTagger/IAnalysisScheduler.cs
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+using System;
+using System.Threading;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
+{
+    public interface IAnalysisScheduler
+    {
+        void Schedule(string filePath, Action<CancellationToken> analyzeAction);
```
duncanp-sonar, line 28: There's nothing specific to analysis in this interface; it's a generic job scheduler, where the `jobId` is a file path.

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -136,9 +141,23 @@ public bool Execute(ProcessRunnerArguments runnerArgs)
             return succeeded;
         }
 
-        #endregion Public methods
-
-        #region Private methods
+        private static Action KillProcess(Process process)
+        {
+            return () =>
```
duncanp-sonar, line 146: It looks a bit weird defining `KillProcess` to return an `Action`. 
Is there a reason not to have have `void KillProcess` and call it using `() => KillProcess(...)`?

src/Integration.Vsix/SonarLintDaemon/DaemonAnalyzer.cs
rita-g-sonarsource: I didn't want to make further changes to DelayedRequest, and the daemon is not supposed to be affected by this change anyway. Do you think I should change it?
Right now DaemonAnalyzer receives the cancellationToken but doesn't pass it along to DelayedRequest.

src/Integration.Vsix/SonarLintTagger/AnalysisScheduler.cs
rita-g-sonarsource: Since I needed two actions: Get (to dispose the previous) and Update (to set the new), and there is no concurrent option for this, so I'd have to use a lock anyway.

src/Integration.Vsix/SonarLintTagger/AnalysisScheduler.cs
rita-g-sonarsource: Keys with null, good point. I'll fix.

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
```diff
@@ -0,0 +1,83 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Threading;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
```
rita-g-sonarsource, line 26: @duncanp-sonar should I move this to another folder?
I wanted to move it to Core, but it needs the MEF decoration. I can move just the interface although that wouldn't do much.

duncanp-sonar: Leave it for now. We could make Core depend on MEF, but I've avoided doing that in the past because it didn't feel right.
One argument for moving the interface is that it makes it clear it's not dependent on VS (and also means we can't accidentally change the interface to depend on VS in a later refactoring). However, I don't think it matters too much in this case.

src/Integration.Vsix/SonarLintDaemon/DaemonAnalyzer.cs
duncanp-sonar: At the very least it's worth documenting that the daemon doesn't currently support cancellation.

src/Integration.Vsix.UnitTests/CFamily/ProcessRunnerTests.cs
```diff
@@ -409,7 +409,90 @@ public void ProcRunner_DoNotLogSensitiveData()
...
+            var args = new ProcessRunnerArguments(exeName, true);
+            var runner = CreateProcessRunner(logger);
+
+            args.CancellationToken = new CancellationToken(true);
+
+            // Act
+            var success = runner.Execute(args);
+
+            // Assert
+            success.Should().BeFalse("");
```
duncanp-sonar, line 431: Redundant ("")

src/Integration.Vsix.UnitTests/CFamily/ProcessRunnerTests.cs
```diff
@@ -409,7 +409,90 @@ public void ProcRunner_DoNotLogSensitiveData()
...
+                result = runner.Execute(args);
+            });
+
+            cancellationTokenSource.CancelAfter(100);
+
+            await runProcess;
+
+            result.Should().BeFalse("Expecting the process to have failed");
+            runner.ExitCode.Should().Be(-1, "Unexpected exit code");
+            logger.AssertOutputStringExists("Hello world");
```
duncanp-sonar, line 470: This looks like a potentially flaky check; it relies on the process having gotten far enough to write the first line before being cancelled.
It would be worth adding a comment calling it out as potentially flaky/timing dependent.

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            this.logger = logger;
+            this.jobs = new Dictionary<string, CancellationTokenSource>(StringComparer.OrdinalIgnoreCase);
+        }
+
+        public void Schedule(string jobId, Action<CancellationToken> action)
+        {
+            using (var newToken = IssueToken(jobId))
+            {
+                action(newToken.Token);
```
duncanp-sonar, line 46: As discussed offline, this won't work if the job includes asynchronous work.
I haven't reviewed the tests for this class yet since I guess they'll change fairly significantly.

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
```diff
@@ -0,0 +1,70 @@
...
+                var newTokenSource = new CancellationTokenSource();
+                jobs[jobId] = new WeakReference<CancellationTokenSource>(newTokenSource);
+                return newTokenSource;
+            }
+        }
+        private void CancelPreviousJob(string jobId)
+        {
+            if (jobs.ContainsKey(jobId))
+            {
+                logger.WriteLine($"Cancelled job for {jobId}");
```
duncanp-sonar, line 61: This message is going to be displayed even if the previous analysis for the file completed successfully hours previously, which is going to be confusing.

We either need to think of a form of words that won't confuse the user, or not write anything at all. Perhaps something like : "Any in-process job for {jobid} will be cancelled.", although I don't think that's great.

If a message is being displayed then the log statement should at least be inside the `if` below so at least it isn't shown if the source has been garbage collected.

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
```diff
@@ -0,0 +1,70 @@
...
+            this.logger = logger;
+            this.jobs = new Dictionary<string, WeakReference<CancellationTokenSource>>(StringComparer.OrdinalIgnoreCase);
+        }
+
+        public void Schedule(string jobId, Action<CancellationToken> action)
+        {
+            var newTokenSource = IssueToken(jobId);
+            action(newTokenSource.Token);
+        }
+        private CancellationTokenSource IssueToken(string jobId)
```
duncanp-sonar, line 47: Missing blank line between methods

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
```diff
@@ -0,0 +1,70 @@
...
+        public Scheduler(ILogger logger)
+        {
+            this.logger = logger;
+            this.jobs = new Dictionary<string, WeakReference<CancellationTokenSource>>(StringComparer.OrdinalIgnoreCase);
+        }
+
+        public void Schedule(string jobId, Action<CancellationToken> action)
+        {
+            var newTokenSource = IssueToken(jobId);
+            action(newTokenSource.Token);
```
duncanp-sonar, line 45: Needs a comment e.g. "We assume the job might be running asynchronously so we don't know when it will finish. This means we can't dispose the `CancellationTokenSources`, and have to rely on weak-refs and garbage collection to do it for us."

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Threading;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
+{
+    [Export(typeof(IScheduler))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class Scheduler : IScheduler
+    {
+        private readonly ILogger logger;
+        private readonly IDictionary<string, WeakReference<CancellationTokenSource>> jobs;
```
duncanp-sonar, line 33: Needs a comment about the known memory leak e.g. "Slow memory leak: each unique jobId will add a new entry to the dictionary. Entries for completed jobs are not removed."

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
rita-g-sonarsource: I can remove it entirely, and perhaps instead move it to ProcessRunner, where we know for sure that it's cancelled.

Issue #1295, opened 2020-03-25
Support reproducer for SLVS


abbas-sabra-sonarsource:
**Is your feature request related to a problem? Please describe.**
Currently, when subprocess fails on SLVS we have no way to debug why. 
When the `subprocess.exe` fails on SonarQube side, we automatically generate a reproducer file that can help us to debug the issue and fix it.
Also, the reproducer file can help if we want to regenerate what is happening on the user side. This would allow us to debug complex FP and performance bugs. 
**Describe the solution you'd like**
Add reproducer generation feature to SonarLint.

**Describe alternatives you've considered**
I don't think there is an alternative other than telling the user to try to regenerate the issue on SonarQube or SonarClound (if they already use them).

### Options
~~1) produce automatically in case of non-zero error code (i.e. crashes), and~~
2) user-triggered (to help diagnose FPs etc)
~3) always produce if logging level is detailed.~

### Implementation
Implement a new command in VS. The command won't be displayed on a menu. Instead, the user will invoke it from the Command Window with the code file open.

The command should:
* locate the active code window
* check it is a c/cpp file
* trigger an analysis with an extra flag to tell `subprocess.exe` to create the reproducer
* log any required information to the output window e.g. the location of the generated file.

The user will not be able to control the location to which the reproducer file is written.

TODO - decide whether to pop up a message box to tell the user where to find the reproducer as well as logging to the output window.

### TODO
- [x] change CFamily analyser so the caller can specify the reproducer should be run
- [x] add a menu command to trigger the analyser with reproducer
- [x] modify IAnalyzer to allow additional analyzer-specific properties to be passed
- [x] modify the analysis requestor to allow re-analysis of a single file
- [x] notify the user that the reproducer has been created and how to find it
- [x] unit tests
- [x] update wiki documentation

rita-g-sonarsource:
@abbas-sabra-sonarsource what is the name of the flag?

PR #1320, opened 2020-04-14
Add CFamily reproducer command

duncanp-sonar:
Relates to #1295.

Adds the menu command. The command is only enabled if document that can be analysed by the CFamily analyser is active.

TODO: trigger the reproducer.

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
@@ -437,26 +437,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
             #endregion ITableEntriesSnapshot methods
         }
 
-        private class DummyMenuCommandService : IMenuCommandService
-        {
-            public IList<MenuCommand> AddedMenuCommands { get; } = new List<MenuCommand>();
-
-            #region IMenuCommandService methods 
-
-            public DesignerVerbCollection Verbs => throw new NotImplementedException();
-
-            public void AddCommand(MenuCommand command) => AddedMenuCommands.Add(command);
-
-            public void AddVerb(DesignerVerb verb) => throw new NotImplementedException();
-            public MenuCommand FindCommand(CommandID commandID) => throw new NotImplementedException();
-            public bool GlobalInvoke(CommandID commandID) => throw new NotImplementedException();
-            public void RemoveCommand(MenuCommand command) => throw new NotImplementedException();
-            public void RemoveVerb(DesignerVerb verb) => throw new NotImplementedException();
-            public void ShowContextMenu(CommandID menuID, int x, int y) => throw new NotImplementedException();
-
-            #endregion IMenuCommandService methods 
-        }
-
         #endregion Helper classes
     }
 }
--- a/src/Integration.Vsix/CFamily/CFamilyReproducerCommand.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyReproducerCommand.cs
@@ -20,16 +20,17 @@
 
 using System;
 using System.ComponentModel.Design;
+using System.Diagnostics;
 using System.Linq;
+using Microsoft.VisualStudio.Editor;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
-using Microsoft.VisualStudio.Shell.TableControl;
-using Microsoft.VisualStudio.Shell.TableManager;
+
+
 using SonarLint.VisualStudio.Core;
-using SonarLint.VisualStudio.Core.CFamily;
 using Task = System.Threading.Tasks.Task;
 
-namespace SonarLint.VisualStudio.Integration.Vsix.Analysis
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
 {
     /// <summary>
     /// Command handler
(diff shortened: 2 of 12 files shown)


src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
```diff
@@ -437,26 +437,6 @@ public bool TryGetValue(int index, string keyName, out object content)
             #endregion ITableEntriesSnapshot methods
         }
 
-        private class DummyMenuCommandService : IMenuCommandService
```
duncanp-sonar, line 440: Moved to a separate class as it's used by the tests for the new command.

src/Integration.Vsix/CFamily/CFamilyReproducerCommand.cs
```diff
@@ -0,0 +1,166 @@
...
+using Microsoft.VisualStudio.Shell.Interop;
+using SonarLint.VisualStudio.Core;
+using Task = System.Threading.Tasks.Task;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
+{
+    /// <summary>
+    /// Command handler
+    /// </summary>
+    internal sealed class CFamilyReproducerCommand
```
duncanp-sonar, line 36: The structure of this class is very similar to the existing `DisableRuleCommand` class (basically cut and paste, with the implementations of `QueryStatus` and `Execute` changed).

src/Integration.Vsix/CFamily/CFamilyReproducerCommand.cs
```diff
@@ -0,0 +1,166 @@
...
+            menuCommandService.AddCommand(menuItem);
+        }
+
+        private void QueryStatus(object sender, EventArgs args)
+        {
+            // Settings enabled to false will stop the command being executed.
+            // VS will print "The command is not available" in the Command window.
+            try
+            {
+                menuItem.Visible = false;
```
duncanp-sonar, line 96: The command is only accessible from the command window so it never needs to be visible (although since it isn't attached to a menu it doesn't make a lot of difference).

src/Integration.Vsix/Helpers/ActiveDocumentLocator.cs
```diff
@@ -0,0 +1,78 @@
...
+    internal interface IActiveDocumentLocator
+    {
+        /// <summary>
+        /// Returns the current active document, or null if there is no active document
+        /// </summary>
+        /// <returns></returns>
+        ITextDocument FindActiveDocument();
+    }
+
+    internal class ActiveDocumentLocator : IActiveDocumentLocator
```
duncanp-sonar, line 39: New class to handle the ugly part of finding the active VS document and extracting the relevant text document. The required steps are:
* using the monitor selection service to the current doc frame, which should be an `IVsWindowFrame`
* get the backing DocData (essentially the data storage for the frame - multiple frames can have the same DocData). It should be an `IVsTextBuffer`.
* the `IVsTextBuffer` is a legacy VS interface. We need to use the adapter service to convert it to the new-style interface. See https://docs.microsoft.com/en-us/visualstudio/extensibility/legacy-interfaces-in-the-editor?view=vs-2015.
* the `ITextBuffer` should have an associated `ITextDocument`, which is what we want to return.

src/Integration.Vsix/SonarLintDaemon/DaemonCommands.vsct
```diff
@@ -2,11 +2,11 @@
 <CommandTable xmlns="http://schemas.microsoft.com/VisualStudio/2005-10-18/CommandTable" xmlns:xs="http://www.w3.org/2001/XMLSchema">
 
   <Commands package="guidDaemonPackagePkgString">
-    <Groups>
+    <!--<Groups>
```
rita-g-sonarsource, line 5: is that commented out on purpose?

src/Integration.Vsix/SonarLintDaemon/DaemonCommands.vsct
```diff
@@ -18,6 +18,20 @@
           <ButtonText>Disable SonarLint rule</ButtonText>
         </Strings>
       </Button>
+
+      <Button guid="guidDaemonCmdSet" id="cmdidCreateCFamilyReproducer" priority="0x0200" type="Button">
+        <!-- It looks like the command needs to belong to a parent menu group, even though it only
+             appears in the command window.
+             We're adding it to the top-level Analye menu, so the command the user types will be
```
rita-g-sonarsource, line 25: minor typo in Analyze

src/Integration.Vsix/SonarLintDaemon/DaemonCommands.vsct
```diff
@@ -18,6 +18,20 @@
...
+               "Analyze.SonarLint.CFamily.Reproducer" -->
+        <Parent guid="guidCodeAnalysisMenuGroup" id="IDG_CODEANALYSIS_TOPLEVEL"/>
+        
+        <CommandFlag>CommandWellOnly</CommandFlag>
+        <Strings>
+          <ButtonText>SonarLint.CFamily.Reproducer</ButtonText>
```
rita-g-sonarsource, line 31: Perhaps the name should actionable, i.e. RunReproducer?

src/Integration.Vsix/Helpers/ActiveDocumentLocator.cs
```diff
@@ -0,0 +1,78 @@
...
+using Microsoft.VisualStudio.TextManager.Interop;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
+{
+    internal interface IActiveDocumentLocator
+    {
+        /// <summary>
+        /// Returns the current active document, or null if there is no active document
+        /// </summary>
+        /// <returns></returns>
```
rita-g-sonarsource, line 35: Remove the empty <returns>?

src/Integration.Vsix/SonarLintDaemon/DaemonCommands.vsct
duncanp-sonar: Oops, no. It's probably stopped the "Disable SonarLint rule" command from appearing...

src/Integration.Vsix/SonarLintDaemon/DaemonCommands.vsct
duncanp-sonar: Ok.

PR #1323, opened 2020-04-15
Modified analysis requestor and repro commands

duncanp-sonar:
Relates to #1295 

TODO:
* modify the TaggerProvider to handle the new requester event args

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
@@ -19,6 +19,7 @@
  */
 
 using System;
+using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Core;
@@ -121,12 +122,12 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
 
             public void AssertAnalysisIsRequested()
             {
-                analysisRequesterMock.Verify(x => x.RequestAnalysis(null, null), Times.Once);
+                analysisRequesterMock.Verify(x => x.RequestAnalysis((IAnalyzerOptions)null, Array.Empty<string>()), Times.Once);
             }
 
             public void AssertAnalysisIsNotRequested()
             {
-                analysisRequesterMock.Verify(x => x.RequestAnalysis(null, null), Times.Never);
+                analysisRequesterMock.Invocations.Count.Should().Be(0);
             }
         }
     }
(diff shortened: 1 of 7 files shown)


duncanp-sonar:
The two commits are logically separable so it might be easier to look at them separately.

src/Integration.Vsix.UnitTests/CFamily/CFamilyReproducerCommandTests.cs
```diff
@@ -94,6 +103,7 @@ public void CheckStatus_NoActiveDocument_NotEnabled()
 
             VerifyDocumentLocatorCalled();
             VerifyLanguageRecognizerNotCalled();
+            VerifyAnalysisNotRequested();
```
rita-g-sonarsource, line 106: From these tests, I understand that Execute evaluated when "OleStatus" is accessed, is that correct?
I think it will be worth creating a separate test, to explicitly test for the analysis request: `CheckStatus_NoActiveDocument_AnalyzerNotCalled` and etc

src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
```diff
@@ -121,12 +121,12 @@ public void SimulateSuppressionsUpdated()
 
             public void AssertAnalysisIsRequested()
             {
-                analysisRequesterMock.Verify(x => x.RequestAnalysis(), Times.Once);
+                analysisRequesterMock.Verify(x => x.RequestAnalysis(null, null), Times.Once);
```
rita-g-sonarsource, line 124: Why would it be null?

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -67,7 +67,7 @@ private void OnUserSettingsChanged(object sender, EventArgs e)
             if (activeSolutionBoundTracker.CurrentConfiguration.Mode == NewConnectedMode.SonarLintMode.Standalone)
             {
                 logger.WriteLine(AnalysisStrings.ConfigMonitor_UserSettingsChanged);
-                analysisRequester.RequestAnalysis();
+                analysisRequester.RequestAnalysis(null, null);
```
rita-g-sonarsource, line 70: I think it's worth creating an overload that doesn't accept parameters. Both of them are optional, so passing nulls seems a bit messy.

duncanp-sonar: As discussed offline, I've added an extension method: 
```
public static void RequestAnalysis(this IAnalysisRequester analysisRequester, params string[] filePaths)
```

src/Integration.Vsix.UnitTests/CFamily/CFamilyReproducerCommandTests.cs
duncanp-sonar: As discussed offline, I've added the check to the other `CheckStatus` test.

PR #1324, opened 2020-04-16
Support setting analyzer options and reproducer flag 

rita-g-sonarsource:
#1295

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
@@ -20,10 +20,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using EnvDTE;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.VCProjectEngine;
 using Moq;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.CFamily;
@@ -37,6 +39,13 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
     public class CFamilyHelperTests
     {
         private const string FileName = @"C:\absolute\path\to\file.cpp";
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            CFamilyHelper.FileConfig.getEvaluatedPropertyValue = null;
+        }
+
         [TestMethod]
         public void FileConfig_Test1()
         {
@@ -387,7 +396,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
             var rulesConfigProviderMock = new Mock<ICFamilyRulesConfigProvider>();
 
             // Act
-            var request = CFamilyHelper.CreateRequest(loggerMock.Object, projectItemMock.Object, "c:\\dummy\\file.h", rulesConfigProviderMock.Object);
+            var request = CFamilyHelper.CreateRequest(loggerMock.Object, projectItemMock.Object, "c:\\dummy\\file.h", rulesConfigProviderMock.Object, null);
 
             // Assert
             AssertMessageLogged(loggerMock, "Cannot analyze header files. File: 'c:\\dummy\\file.h'");
@@ -404,7 +413,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
             var rulesConfigProviderMock = new Mock<ICFamilyRulesConfigProvider>();
 
             // Act
-            var request = CFamilyHelper.CreateRequest(loggerMock.Object, projectItemMock.Object, "c:\\dummy\\file.cpp", rulesConfigProviderMock.Object);
+            var request = CFamilyHelper.CreateRequest(loggerMock.Object, projectItemMock.Object, "c:\\dummy\\file.cpp", rulesConfigProviderMock.Object, null);
 
             // Assert
             AssertMessageLogged(loggerMock,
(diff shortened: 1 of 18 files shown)


src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -37,6 +39,13 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
     public class CFamilyHelperTests
     {
         private const string FileName = @"C:\absolute\path\to\file.cpp";
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            CFamilyHelper.FileConfig.getEvaluatedPropertyValue = null;
```
rita-g-sonarsource, line 46: @duncanp-sonar since it's a static property, my new tests (which depend on VCConfigurationMock) would fail. So I had to reset it between tests.

src/TestInfrastructure/VsObjectModelMocks/ProjectMock.cs
```diff
@@ -37,6 +37,8 @@ public class ProjectMock : VsUIHierarchyMock, IVsProject, Project, IVsBuildPrope
 
         public DTE DTE { get; set; }
 
+        public object Project { get; set; }
```
rita-g-sonarsource, line 40: @duncanp-sonar I'm using this to set it to VCProjectMock in the tests.
VCProjectMock is not a complete mock, so it didn't feel right to move it to Infrastructure.

rita-g-sonarsource:
@duncanp-sonar the relevant change is in `CLangAnalyzer`, the other classes are just passing the analyzer options.

src/TestInfrastructure/VsObjectModelMocks/ConfigurationMock.cs
```diff
@@ -38,124 +38,37 @@ public PropertiesMock Properties
...
+        object Configuration.Owner => throw new NotImplementedException();
 
-        string Configuration.PlatformName
-        {
-            get
-            {
-                throw new NotImplementedException();
-            }
-        }
+        string Configuration.PlatformName => "x64";
```
rita-g-sonarsource, line 67: @duncanp-sonar this is what I really changed. Sorry, I got carried away with the others...

src/Integration.Vsix/Analysis/IAnalyzerController.cs
```diff
@@ -30,7 +31,13 @@ public interface IAnalyzer
...
         bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages);
 
-        void ExecuteAnalysis(string path, string charset, IEnumerable<AnalysisLanguage> detectedLanguages, IIssueConsumer consumer, ProjectItem projectItem, CancellationToken cancellationToken);
+        void ExecuteAnalysis(string path,
+            string charset,
+            IEnumerable<AnalysisLanguage> detectedLanguages,
+            IIssueConsumer consumer,
+            ProjectItem projectItem,
+            CancellationToken cancellationToken,
+            IAnalyzerOptions analyzerOptions);
```
duncanp-sonar, line 40: The convention for APIs with a `CancellationToken` is to have it as the last parameter.

src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -85,7 +85,7 @@ public static FileConfig TryGet(ProjectItem projectItem, string absoluteFilePath
                 };
             }
 
-            private static MethodInfo getEvaluatedPropertyValue;
+            internal static MethodInfo getEvaluatedPropertyValue;
```
duncanp-sonar, line 88: I think it's worth including the "for testing" comment i.e. `internal /* for testing */ MethodInfo ...`
It captures the design intent that the property is expected to be called from test code but not from other product code. Without the comment, it's not clear whether the choice of scope was clearly thought about or not (and in some bits of the code base it doesn't seem to have been).

src/Integration.Vsix/CFamily/CFamilyAnalyzerOptions.cs
```diff
@@ -0,0 +1,9 @@
+﻿using SonarLint.VisualStudio.Core;
```
duncanp-sonar, line 1: This is now redundant - I moved this class to `Core` at the same time as `IAnalyzerOptions`.

src/Integration.Vsix/CFamily/FileConfig.cs
duncanp-sonar: ... or alternatively change the scope back to private and add an internal `SetEvaluatePropertyValueDelegateForTesting` method.

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -733,4 +799,46 @@ string IVCRulePropertyStorage.GetEvaluatedPropertyValue(string propertyName)
...
+        {
+            var propertyInfo = GetType().GetProperty(propertyName);
+            if (propertyInfo != null)
+            {
+                return (string)propertyInfo.GetValue(this);
+            }
+            return "dummy property value";
+        }
+
+        public dynamic Item(string name)
```
duncanp-sonar, line 839: As discussed offline, here a couple of things to check/do:
* move the classes out of this ugly hack namespace if they don't need to be there
* check whether `dynamic Item` is required, and if it needs to be dynamic
* replace the properties that don't need to be accessed directly as properties with a dictionary, and change `GetEvaluatedPropertyValue` to use the dictionary rather than reflection.

src/Integration.Vsix/CFamily/FileConfig.cs
duncanp-sonar: Better still, you never actually set the value from the tests, so you only need an internal method to clear the private static property.

PR #1328, opened 2020-04-16
Factored out IIssueTracker interface

duncanp-sonar:
Relates to #1295 

Two separate commits:
1) factored out IIssueTracker
2) added issue filtering to the TaggerProvider

### DIFF
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
@@ -265,6 +265,83 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             act.Should().NotThrow();
         }
 
+        [TestMethod]
+        [DataRow(null)]
+        [DataRow(new string[] { })]
+        public void FilterIssueTrackersByPath_NullOrEmptyPaths_AllTrackersReturned(string[] filePaths)
+        {
+            var trackers = CreateMockedIssueTrackers("any", "any2");
+
+            var actual = TaggerProvider.FilterIssuesTrackersByPath(trackers, filePaths);
+            
+            actual.Should().BeEquivalentTo(trackers);
+        }
+
+        [TestMethod]
+        public void FilterIssueTrackersByPath_WithPaths_NoMatches_EmptyListReturned()
+        {
+            var trackers = CreateMockedIssueTrackers("file1.txt", "c:\\aaa\\file2.cpp");
+
+            var actual = TaggerProvider.FilterIssuesTrackersByPath(trackers,
+                new string[] { "no matches", "file1.wrongextension" });
+
+            actual.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void FilterIssueTrackersByPath_WithPaths_SingleMatch_SingleTrackerReturned()
+        {
+            var trackers = CreateMockedIssueTrackers("file1.txt", "c:\\aaa\\file2.cpp", "d:\\bbb\\file3.xxx");
+
+            var actual = TaggerProvider.FilterIssuesTrackersByPath(trackers,
+                new string[] { "file1.txt" });
+
+            actual.Should().BeEquivalentTo(trackers[0]);
+        }
+
+        [TestMethod]
+        public void FilterIssueTrackersByPath_WithPaths_MultipleMatches_MultipleTrackersReturned()
+        {
+            var trackers = CreateMockedIssueTrackers("file1.txt", "c:\\aaa\\file2.cpp", "d:\\bbb\\file3.xxx");
+
+            var actual = TaggerProvider.FilterIssuesTrackersByPath(trackers,
+                new string[]
+                {
+                    "file1.txt",
+                    "D:\\BBB\\FILE3.xxx" // match should be case-insensitive
+                });
+
+            actual.Should().BeEquivalentTo(trackers[0], trackers[2]);
+        }
+
+        [TestMethod]
+        public void FilterIssueTrackersByPath_WithPaths_AllMatched_AllTrackersReturned()
+        {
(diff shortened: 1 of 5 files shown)


src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -154,7 +154,7 @@ private void OnAnalysisRequested(object sender, AnalysisRequestEventArgs args)
             if (detectedLanguages.Any() && analyzerController.IsAnalysisSupported(detectedLanguages))
             {
                 // Multiple views could have that buffer open simultaneously, so only create one instance of the tracker.
-                var issueTracker = buffer.Properties.GetOrCreateSingletonProperty(typeof(TextBufferIssueTracker),
+                var issueTracker = buffer.Properties.GetOrCreateSingletonProperty(typeof(IIssueTracker),
```
duncanp-sonar, line 157: The `Type` is being used as a key for the singleton property. I've checked that the behaviour is the same with this change.

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
```diff
@@ -265,6 +265,56 @@ public void RequestAnalysis_Should_NotThrow_When_AnalysisFails()
...
+        {
+            var tracker1 = CreateMockedIssueTracker("any");
+            var tracker2 = CreateMockedIssueTracker("any2");
+
+            var actual = TaggerProvider.FilterIssuesTrackersByPath(new IIssueTracker[] { tracker1, tracker2}, filePaths);
+            actual.Should().BeEquivalentTo(tracker1, tracker2);
+        }
+
+        [TestMethod]
+        public void FilterIssueTrackersByPath_WithPaths_ExpectedTrackersReturned()
```
duncanp-sonar, line 281: I tried using DataRows for this test too but it got too ugly and difficult to read.

src/Integration.Vsix/SonarLintTagger/IssueTagger.cs
```diff
@@ -39,9 +39,9 @@ internal sealed class IssueTagger : ITagger<IErrorTag>, IDisposable
     {
         private IssuesSnapshot issues;
         private bool isDisposed;
-        internal /* for testing */ TextBufferIssueTracker IssueTracker { get; }
+        internal /* for testing */ IIssueTracker IssueTracker { get; }
```
rita-g-sonarsource, line 42: hallelujah!

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
```diff
@@ -265,6 +265,56 @@ public void RequestAnalysis_Should_NotThrow_When_AnalysisFails()
...
+        {
+            var file1Tracker = CreateMockedIssueTracker("file1.txt");
+            var file2Tracker = CreateMockedIssueTracker("c:\\aaa\\file2.cpp");
+            var file3Tracker = CreateMockedIssueTracker("d:\\bbb\\file3.xxx");
+
+            // 1. No matches
+            filter("no matches", "file1.wrongextension")
```
rita-g-sonarsource, line 288: I believe that it's better for the test's readability and "health" to have each use-case tested in a separate test method. This way, when a tests fails, it's way easier to understand what happened. It's also a good way to document the intention of the test, rather than rely on comments.
So for example, I find these to be more self-explanatory: 

FilterIssueTrackersByPath_PathNotNull_NoMatchingTrackers_EmptyListReturned
FilterIssueTrackersByPath_PathNotNull_OneMatchingTracker_OneTrackerReturned
FilterIssueTrackersByPath_PathNotNull_TwoMatchingTrackers_TwoTrackersReturned

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -127,7 +129,17 @@ private void OnAnalysisRequested(object sender, AnalysisRequestEventArgs args)
...
+            IEnumerable<IIssueTracker> issueTrackers, IEnumerable<string> filePaths)
+        {
+            if (filePaths == null || filePaths.Count() == 0)
+            {
+                return issueTrackers;
+            }
+            return issueTrackers.Where(it => filePaths.Contains(it.FilePath, StringComparer.OrdinalIgnoreCase));
+        }
+
+        internal IEnumerable<IIssueTracker> ActiveTrackersForTesting { get { return this.issueTrackers; } }
```
rita-g-sonarsource, line 142: Did you end up using this property? I can't find the usages.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -214,7 +226,7 @@ public void RemoveSinkManager(SinkManager manager)
             }
         }
 
-        public void AddIssueTracker(TextBufferIssueTracker bufferHandler)
+        public void AddIssueTracker(IIssueTracker bufferHandler)
```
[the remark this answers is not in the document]
duncanp-sonar: It'll probably make the coverage worse, but I'll try it.

duncanp-sonar:
Unfortunately, extracting a `RequestAnalysis` method with that signature doesn't solve the problem as running the analysis also involves the `StatusBarReanalysisProgressHandler` and `CancellableJobRunner` classes.
I'm accepting the fact that there a couple of uncovered lines in this PR that can't easily be tested, and relying on the post-merge coverage numbers being high enough.

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
duncanp-sonar: I've changed the tests. As discussed offline, I find the new version less readable. I think part of the problem for me is that I find long test names very clumsy, and having individual tests for individual conditions makes it harder to tell whether the whole set of necessary conditions for a method have been tested or not.

PR #1334, opened 2020-04-17
Log the process exit code

rita-g-sonarsource:


src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -150,7 +150,7 @@ private bool WaitForProcessToFinish(Process process, ProcessRunnerArguments runn
             // true: we might still have timed out, but the process ended when we asked it to
             if (succeeded)
             {
-                LogDebug(CFamilyStrings.MSG_ExecutionExitCode, process.ExitCode);
+                LogMessage(CFamilyStrings.MSG_ExecutionExitCode, process.ExitCode);
```
abbas-sabra-sonarsource, line 153: This always prints the exit code(even in case of timeout). And the exit code isn't a helpful message.
Can we do something like:
```
if(process.ExitCode!=0){
LogMessage("Analysis of this file has failed. Please generate a reproducer file *maybe a link to reproducer doc* and share it on *link to our community*);
}
```

rita-g-sonarsource: This message will be a bit problematic, as it's a generic class that runs exe.
Will a message like this be ok? : "Process failed, error code: ..."

src/Integration.Vsix/CFamily/CFamilyStrings.resx
```diff
@@ -123,6 +123,9 @@
   <data name="ERROR_ProcessRunner_ExeNotFound" xml:space="preserve">
     <value>Execution failed. The specified executable does not exist: {0}</value>
   </data>
+  <data name="ERROR_ProcessRunner_Failed" xml:space="preserve">
+    <value>Process failed, error code: {0}</value>
```
duncanp-sonar, line 127: `Process failed. Exit code:`

... might be better (both grammatically, and because it's normal to talk about exit codes for processes)

PR #1343, opened 2020-04-21
Removed legacy telemetry implementation

### LINKED ISSUE #1341: Delete redundant telemetry
### Description
The original telemetry was implemented by the team at Microsoft and sends data to Microsoft. SonarSource have no access to this data (and in the event it was not used by Microsoft either).
It should be removed.

Note: there is a second, completely independent telemetry implementation added by SonarSource that should be preserved.

### TODO
- [x] remove the code under `Integration.Vsix\Telemetry`
- [x] remove calls to the delete telemetry commands
- [x] CHECK that this has no impact on the loading of other features
- [x] remove corresponding tests

duncanp-sonar:
Fixes #1341


duncanp-sonar:
The legacy telemetry feature worked by firing dummy VS commands, relying on the fact that VS automatically records any command invocation and sends telemetry about it to MS.
This required quite a lot of plumbing code, most of which was generated.

I've deleted it all, and checked that the other SonarLint features are not impacted e.g. tools options, binding, analysis of various languages, the menu commands.

src/Integration.Vsix/SonarLintIntegrationPackage.cs
```diff
@@ -36,6 +36,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix
     // Register this class as a VS package.
     [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid(CommonGuids.Package)]
+    [ProvideMenuResource("Menus.ctmenu", 1)]
```
duncanp-sonar, line 39: Previously this attribute was on the generated partial class. However, it's still required - it notifies VS (or more accurately the VSSDK build targets) that this package handles menu commands.

src/Integration.Vsix/BoundSolutionAnalyzer.cs
```diff
@@ -1,116 +0,0 @@
-﻿/*
```
duncanp-sonar, line 1: The class only existed to send telemetry data whenever a solution was opened. It's no longer required.

src/Integration/MefServices/ITelemetryLogger.cs
```diff
@@ -1,27 +0,0 @@
-/*
```
duncanp-sonar, line 1: There is a completely separate interface for the "new" SonarSource telemetry.

duncanp-sonar:
@rita-g-sonarsource this is more as an FYI. It just deletes a lot of boilerplate and generated code so I'd suggest not spending too long trying to understand the detail of everything that has been removed.

PR #1344, opened 2020-04-21
Changed the timestamp server used

duncanp-sonar:
The symantec server is been replaced following the takeover by digicert.

https://knowledge.digicert.com/alerts/migration-of-legacy-verisign-and-symantec-time-stamping-services.html


duncanp-sonar:
I've manually checked the artefacts from Azure Pipeline to check they are timestamped:

image

Issue #176, opened 2017-05-18
Additional files for parameterized SonarQube rules are retrieved from the SonarQube quality profile


fmallet:
Some analyzer (SonarQube analyzer for C#) requires additional files to be passed to the compiler to configure some rules (parametrized rules).
this is done in the continuous integration (See Scanner for MSBuild), but not yet for the connected experience.

Related [MMF-510](https://jira.sonarsource.com/browse/MMF-510)

Here are the parametered rules:

C# and VB.NET

- S2342 EnumNameShouldFollowRegexBase
- S1067 ExpressionComplexityBase
- S104 FileLinesBase
- S1541 FunctionComplexityBase (cyclomatic complexity)
- S134 FunctionNestingDepthBase
- S103 LineLengthBase

C#

- S110 AvoidExcessiveInheritance
- S1451 CheckFileLicense
- S3776 CognitiveComplexity
- S1541 FunctionComplexity
- S2436 TooManyGenericParameters
- S1479 TooManyLabelsInSwitch
- S107 TooManyParameters

VB.NET

- S101 ClassName
- S2343 EnumerationValueName
- S2347 EventHandlerName
- S2348 EventName
- S1542 FunctionName
- S114 InterfaceName
- S117 LocalVariableName
- S2304 NamespaceName
- S1654 ParameterName
- S2366 PropertyName
- S2373 TypeParameterName
- S2375 UseWithStatement 
- S2362 PrivateConstantFieldName
- S2364 PrivateFieldName
- S2363 PrivateSharedReadonlyFieldName
- S2367 PublicConstantFieldName
- S2369 PublicFieldName
- S2370 PublicSharedReadonlyFieldName

All of them are disabled by default, except S1479, which have been tuned lately to not report on enums and therefore we don't expect false positives with the default max value (30).

Comes from [SVS-5](https://jira.sonarsource.com/browse/SVS-5)

FarisFAhmed:
Hi,

you said:
_this is done in the continuous integration (See Scanner for MSBuild), but not yet for the connected experience._

does this mean that default rule parameters like _maximumLineLength_ in rule S103 can be changed when scanning using sonar-scanner-msbuild-2.3.2.573?
If this is possible in msbuild can you tell me how?

Regards,
Faris Ahmed

duncanp-sonar:
[...]

PR #1352, opened 2020-04-26
#176 - Remove code duplication + old migration code

rita-g-sonarsource:
Relates to #176


src/Integration/Persistence/ILegacySonarQubeFolderModifier.cs
```diff
@@ -20,8 +20,8 @@
 
 namespace SonarLint.VisualStudio.Integration.Persistence
 {
-    internal interface ISolutionBindingPostSaveOperation
+    internal interface ILegacySonarQubeFolderModifier
```
duncanp-sonar, line 23: It's not clear what the responsibilities of this interface are.
A doc comment would help.

duncanp-sonar:
Could you call out the duplication that's been removed please? It's not obvious.

src/Integration/Persistence/SolutionBindingSerializer.cs
```diff
@@ -64,7 +64,7 @@ public BoundSonarQubeProject Read(string configFilePath)
             return bound;
         }
 
-        public bool Write(string configFilePath, BoundSonarQubeProject binding, Predicate<string> onSuccessfulFileWrite)
```
duncanp-sonar, line 67: In future, please call in a comment why you are making non-obvious changes e.g. why this method is now returning an `Action` rather than a `Predicate`.

After wading through the changes, my guess is that it's because the only implementation always returned `true`, but it's just a guess.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -54,24 +54,27 @@ internal class SolutionBindingOperation : ISolutionBindingOperation
...
         private readonly IFileSystem fileSystem;
 
         public SolutionBindingOperation(IServiceProvider serviceProvider,
             ConnectionInformation connection,
             string projectKey,
             string projectName,
             SonarLintMode bindingMode,
-            ILogger logger)
-            : this(serviceProvider, connection, projectKey, projectName, bindingMode, logger, new FileSystem())
+            ILogger logger,
```
duncanp-sonar, line 65: The convention is to have the `logger` as the last parameter i.e. to have the "important"/task-specific parameters first.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -287,40 +291,7 @@ private void UpdateSolutionFile()
...
-            Debug.Assert(Path.IsPathRooted(fullFilePath) && fileSystem.File.Exists(fullFilePath), "Expecting a rooted path to existing file");
-
-            Project solutionItemsProject = this.projectSystem.GetSolutionFolderProject(Constants.LegacySonarQubeManagedFolderName, true);
-            if (solutionItemsProject == null)
-            {
-                Debug.Fail("Could not find the solution items project");
-            }
-            else
-            {
-                if (!this.projectSystem.IsFileInProject(solutionItemsProject, fullFilePath))
```
duncanp-sonar, line 306: Ok, so this is the duplicated code that has been removed.

There are slight differences between this version of the code and the version that replaces it i.e. this version calls `IsFileInProject`.
Going forwards, when sending out a review please
* call out the duplicated code that has been removed, and
* if there are differences between the old and new versions, call them out.
Otherwise, it's a real pain to find and review the de-duplicated code in the GitHub UI.

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -288,7 +288,7 @@ private void RegisterLocalServices()
                 var solution = this.GetService<SVsSolution, IVsSolution>();
                 var connectedModeConfigPathProvider = new ConnectedModeSolutionBindingPathProvider(solution);
                 var legacyConfigPathProvider = new LegacySolutionBindingPathProvider(this);
-                var legacyPostSaveOperation = new LegacySolutionBindingPostSaveOperation(this);
+                var legacyPostSaveOperation = new LegacySonarQubeFolderModifier(this);
```
duncanp-sonar, line 291: The field name hasn't been updated to match the new class name.

src/Integration/NewConnectedMode/ConfigurationProvider.cs
```diff
@@ -30,12 +30,12 @@ internal class ConfigurationProvider : IConfigurationProvider
...
         private readonly ISolutionBindingPathProvider connectedModePathProvider;
         private readonly ISolutionBindingSerializer solutionBindingSerializer;
-        private readonly ISolutionBindingPostSaveOperation legacyPostSaveOperation;
+        private readonly ILegacySonarQubeFolderModifier legacySonarQubeFolderModifier;
 
         public ConfigurationProvider(ISolutionBindingPathProvider legacyPathProvider,
             ISolutionBindingPathProvider connectedModePathProvider,
             ISolutionBindingSerializer solutionBindingSerializer,
-            ISolutionBindingPostSaveOperation legacyPostSaveOperation)
+            ILegacySonarQubeFolderModifier legacyPostSaveOperation)
```
duncanp-sonar, line 38: The parameter name hasn't changed to match the new interface.

src/Integration/Binding/BindingController.cs
```diff
@@ -156,13 +156,16 @@ void IBindingWorkflowExecutor.BindProject(BindCommandArgs bindingArgs)
                 nugetBindingOp = new NoOpNuGetBindingOperation(host.Logger);
             }
 
+            var folderModifier = new Persistence.LegacySonarQubeFolderModifier(host);
```
duncanp-sonar, line 159: This looks weird. Why is the binding controller directly new-ing up the `LegacySonarQubeFolderModifier`?

If we always expect the same implementation of `ILegacySonarQubeFolderModifier` to be used and it just being being injected in the constructor for testing, then it would be better to have an _internal /* for testing */_ constructor, and have the public constructor call it a new instance.

src/Integration.UnitTests/Persistence/LegacySonarQubeFolderModifierTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        [TestMethod]
+        public void Ctor_NullFileSystem_ArgumentNullException()
+        {
+            Action act = () => new LegacySonarQubeFolderModifier(Mock.Of<IServiceProvider>(), null);
+
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("fileSystem");
+        }
+
+        [TestMethod]
+        public void Add_AddsFileToSolutionFolder()
```
duncanp-sonar, line 50: Missing test case - the legacy solution folder doesn't exist.

src/Integration/Persistence/SolutionBindingSerializer.cs
rita-g-sonarsource: You're right, my bad. I'll reference it in the future.

src/Integration/Binding/SolutionBindingOperation.cs
rita-g-sonarsource: IsFileInProject is done internally inside projectSystemHelper, that's why I went with the version that didn't re-check it.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -63,17 +64,18 @@ internal class SolutionBindingOperation : ISolutionBindingOperation
             string projectName,
             SonarLintMode bindingMode,
             ILogger logger)
-            : this(serviceProvider, connection, projectKey, projectName, bindingMode, logger, new ProjectBinderFactory(serviceProvider), new FileSystem())
+            : this(serviceProvider, connection, projectKey, projectName, bindingMode,  new ProjectBinderFactory(serviceProvider), new LegacyConfigFolderItemAdder(serviceProvider), logger, new FileSystem())
```
rita-g-sonarsource, line 67: this ctor is a real mess, it should be better after the other refactoring bits.

PR #1353, opened 2020-04-26
Replace BindingRefactoringDumpingGround with factory

rita-g-sonarsource:
Relates to #176


src/Integration/Binding/RoslynConfigProjectBinder.cs
```diff
@@ -0,0 +1,26 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Integration.Binding
+{
+    internal class RoslynConfigProjectBinder : IConfigProjectBinder
```
duncanp-sonar, line 23: Is `Config` required in the name?

A doc comment explaining the responsibilities of the interface would be useful.

src/Integration/Binding/IConfigProjectBinderFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using EnvDTE;
+
+namespace SonarLint.VisualStudio.Integration.Binding
+{
+    internal interface IConfigProjectBinderFactory
```
duncanp-sonar, line 25: Same here - is `Config` required in the name?

src/Integration/Binding/ConfigProjectBinderFactory.cs
```diff
@@ -23,19 +23,14 @@
...
     {
-        internal /* for testing */ static bool IsProjectLevelBindingRequired(Project project)
+        public IConfigProjectBinder Get(Project project)
         {
-            var languages = ProjectToLanguageMapper.GetAllBindingLanguagesForProject(project);
-            return languages.Contains(Core.Language.VBNET) || languages.Contains(Core.Language.CSharp);
+            var languages = ProjectToLanguageMapper.GetAllBindingLanguagesForProject(project).ToList();
+            var isRoslynProject = languages.Contains(Core.Language.VBNET) || languages.Contains(Core.Language.CSharp);
+
+            return isRoslynProject ? new RoslynConfigProjectBinder() : null;
```
duncanp-sonar, line 33: As discussed offline, the convention is that `Get` won't return null. What's your plan for this code?

src/Integration/Binding/IConfigProjectBinder.cs
```diff
@@ -0,0 +1,26 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Integration.Binding
+{
+    internal interface IConfigProjectBinder
```
duncanp-sonar, line 23: Is `Config` required in the name?

A doc comment explaining the responsibilities of the interface would be useful.

src/Integration/Binding/RoslynConfigProjectBinder.cs
duncanp-sonar, line 23: I find the use of `Roslyn` in the name of this class and the associated binder class a bit confusing since they don't reference any of the `Roslyn` APIs.
Why `Roslyn`? Because the projects support Roslyn analyzers?

src/Integration/Binding/ProjectBindingOperation.cs
```diff
@@ -51,7 +51,10 @@ internal partial class ProjectBindingOperation : IBindingOperation
         {
         }
 
-        internal ProjectBindingOperation(IServiceProvider serviceProvider, Project project, ISolutionBindingConfigFileStore configFileStore, IFileSystem fileSystem)
+        internal ProjectBindingOperation(IServiceProvider serviceProvider,
```
duncanp-sonar, line 54: Most of the code in SLVS doesn't separate constructor arguments onto separate lines. Did you make this change just for stylistic reasons? If so, can we have an offline discussion about how we are going to handle unnecessary stylistic changes please?

src/Integration/Binding/RoslynConfigProjectBinder.cs
rita-g-sonarsource: Yep, to differentiate between roslyn / dotnet and non-roslyn projects. I'd expect roslyn ones to have a ruleset and parameters. How do you think it should be called?

src/Integration/Binding/ProjectBindingOperation.cs
rita-g-sonarsource: I saw a lot of code in SLVS that does separate, so I agree, we should decide on a convention.
I'll schedule a meeting? we can discuss other things as well, like tests naming, etc.

src/Integration/Binding/ConfigProjectBinderFactory.cs
rita-g-sonarsource: I will probably return a no-op, although I'm not a fan of that either.
I'll think about it. Maybe TryGet/TryCreate, not sure yet.

PR #1355, opened 2020-04-28
#176 Extract IsBound logic from UnboundProjectFinder into RoslynProjectBinder

rita-g-sonarsource:


src/TestInfrastructure/Framework/ConfigurableRuleSetSerializer.cs
```diff
@@ -124,11 +124,6 @@ public void AssertRuleSetLoaded(string ruleSet, int expectedNumberOfTimes)
             actual.Should().Be(expectedNumberOfTimes, "RuleSet {0} was loaded unexpected number of times", ruleSet);
         }
 
-        public void AssertAllRegisteredRuleSetsLoadedExactlyOnce()
```
rita-g-sonarsource, line 127: This is no longer true, as we don't cache the rulesets anymore.

src/Integration/ProfileConflicts/ConflictsManager.cs
```diff
@@ -150,7 +150,7 @@ private IReadOnlyList<ProjectRuleSetConflict> FindConflicts(RuleSetInformation[]
                     continue;
                 }
 
-                if (projectBinderFactory.Get(project) == null)
+                if (projectBinderFactory.Get(project) is CFamilyProjectBinder)
```
rita-g-sonarsource, line 153: I'm really not a fan of this, but this should be temporary:  the logic should eventually move to ProjectBinder.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -177,15 +177,15 @@ public void Initialize(IEnumerable<Project> projects, IDictionary<Language, Sona
 
             foreach (var project in projects)
             {
-                if (projectBinderFactory.Get(project) != null)
+                if (projectBinderFactory.Get(project) is CFamilyProjectBinder)
```
rita-g-sonarsource, line 180: This should eventually move to ProjectBinder.

src/Integration/Binding/RoslynProjectBinder.cs
```diff
@@ -18,9 +18,90 @@
...
+using System.Diagnostics;
+using System.IO.Abstractions;
+using System.Linq;
+using EnvDTE;
+using Microsoft.VisualStudio.CodeAnalysis.RuleSets;
+using SonarLint.VisualStudio.Integration.NewConnectedMode;
+
 namespace SonarLint.VisualStudio.Integration.Binding
 {
     internal class RoslynProjectBinder : IProjectBinder
```
rita-g-sonarsource, line 31: Cut-paste of the code that was in UnboundProjectFinder

src/Integration/Binding/CFamilyProjectBinder.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO.Abstractions;
+using System.Linq;
+using EnvDTE;
+using SonarLint.VisualStudio.Integration.NewConnectedMode;
+
+namespace SonarLint.VisualStudio.Integration.Binding
+{
+    internal class CFamilyProjectBinder : IProjectBinder
```
rita-g-sonarsource, line 10: As we discussed, solution-level language-specific test should still happen somewhere. I've put it here for now, but I'm unsure of its final destination; there is a logical duplication here, as the solution is being checked over and over again, so it should probably be checked only once, as part of the solution-level checks. In the meanwhile it's here, I'll revisit this once the other parts are integrated.

src/Integration.UnitTests/Binding/ProjectBinderFactoryTests.cs
```diff
@@ -0,0 +1,109 @@
...
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("fileSystem");
+        }
+
+        [TestMethod]
+        public void Get_CSharpProject_RoslynProjectBinderReturned()
+        {
+            var project = new ProjectMock("c:\\foo.proj");
+            project.SetCSProjectKind();
+
+            using (new AssertIgnoreScope())
```
rita-g-sonarsource, line 63: This is due to the fact that RoslynProjectBinder and CFamilyProjectBinder have inner dependencies that are being checked for not null in their CTORs. However, I don't want the factory class to know about them and I only need to verify that the correct type is being initialized.

src/Integration.UnitTests/ProfileConflicts/ConflictsManagerTests.cs
```diff
@@ -64,6 +64,8 @@ public void TestInit()
             };
             this.serviceProvider.RegisterService(typeof(ISolutionRuleSetsInformationProvider), this.ruleSetInfoProvider);
 
+            this.serviceProvider.RegisterService(typeof(IRuleSetSerializer), Mock.Of<IRuleSetSerializer>());
```
rita-g-sonarsource, line 67: Inner dependency of RoslynProjectBinder.

rita-g-sonarsource:
@duncanp-sonar this PR cuts-pastes the code from UnboundProjectFinder into RoslynProjectBinder.
It's far from perfect but I wanted to have something that you can already start to review and call out problems in advance.
The logic is currently still being checked by UnboundProjectFinder and the other usages (such as ConflictsManager, SolutionBindingOperation) -- I suggest to rewrite those tests in a different PR.

src/Integration/Binding/ProjectBinderFactory.cs
```diff
@@ -18,19 +18,37 @@
...
+        }
+
         public IProjectBinder Get(Project project)
         {
             var languages = ProjectToLanguageMapper.GetAllBindingLanguagesForProject(project).ToList();
             var isRoslynProject = languages.Contains(Core.Language.VBNET) || languages.Contains(Core.Language.CSharp);
 
-            return isRoslynProject ? new RoslynProjectBinder() : null;
+            return isRoslynProject
+                ? (IProjectBinder) new RoslynProjectBinder(serviceProvider, fileSystem)
```
duncanp-sonar, line 50: Is the cast required?

src/Integration/Binding/CFamilyProjectBinder.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
```
duncanp-sonar, line 1: Missing copyright

src/Integration/Binding/CFamilyProjectBinder.cs
duncanp-sonar: From a practical point of view, repeating the check for each project won't cause a problem in practice.

src/Integration/Binding/SolutionBindingOperation.cs
duncanp-sonar: How are you tracking the bits of code that should change? You've called out a number of them in review comments, but they will be effectively lost once the PR is merged.

If I'm working on a feature/refactoring that I know is going to be split across multiple sprints and have a something I want to revisit in a later commit, I normally put a comment in the code with `duncanp` in it.
I find it works better for me than a `TODO`, and it's easy to check that removed them all before committing.

src/Integration/LocalServices/UnboundProjectFinder.cs
```diff
@@ -33,166 +31,43 @@ namespace SonarLint.VisualStudio.Integration
...
             // Only applicable in connected mode (legacy or new)
             var bindingConfig = configProvider.GetConfiguration();
 
-            if (!bindingConfig.Mode.IsInAConnectedMode())
-            {
-                return Enumerable.Empty<Project>();
-            }
-
-            return this.GetUnboundProjects(bindingConfig);
+            return !bindingConfig.Mode.IsInAConnectedMode() ? Enumerable.Empty<Project>() : GetUnboundProjects(bindingConfig);
```
duncanp-sonar, line 50: This would be easier to read without the negation operator.

src/Integration/ProfileConflicts/ConflictsManager.cs
duncanp-sonar: The intent would have been clearer if the check have been against `RoslynProjectBinder` (i.e. "we only want to continue if the binder is of type X"), but it doesn't matter if it's temporary.

src/Integration/Binding/SolutionBindingOperation.cs
rita-g-sonarsource: I write in a notebook :) But I also periodically check the base branch and evaluate how everything looks.

src/Integration/Binding/CFamilyProjectBinder.cs
rita-g-sonarsource: Every time! :(
Perhaps we change the quality profile and add the copyright rule?

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -323,7 +323,7 @@ private void RegisterLocalServices()
 
                 return new ProjectSystemFilter(this, testProjectIndicator);
             }));
-            this.localServices.Add(typeof(IErrorListInfoBarController), new Lazy<ILocalService>(() => new ErrorListInfoBarController(this, new UnboundProjectFinder(this, new ProjectBinderFactory()))));
+            this.localServices.Add(typeof(IErrorListInfoBarController), new Lazy<ILocalService>(() => new ErrorListInfoBarController(this, new UnboundProjectFinder(this, new ProjectBinderFactory(this)))));
```
duncanp-sonar, line 326: Looks like the creation of the ProjectBinderFactory could be moved to an internal constructor on both `RuleSetConflictsController` and `UnboundProjectFinder`. Doesn't need to be done now though if the code is still changing.

src/Integration/Binding/SolutionBindingOperation.cs
duncanp-sonar: Ok, but writing in a notebook and having the information in your head doesn't help anyone else in the team if they have to pick up the work for any reason. It's better to have some kind of public/shared record of what still needs to be done.

src/Integration.UnitTests/ProfileConflicts/ConflictsManagerTests.cs
```diff
@@ -79,16 +81,16 @@ public void TestInit()
...
         [TestMethod]
         public void ConflictsManager_Ctor()
         {
-            Exceptions.Expect<ArgumentNullException>(() => new ConflictsManager(null, new Mock<ILogger>().Object, new ProjectBinderFactory(), fileSystem));
-            Exceptions.Expect<ArgumentNullException>(() => new ConflictsManager(serviceProvider, null, new ProjectBinderFactory(), fileSystem));
+            Exceptions.Expect<ArgumentNullException>(() => new ConflictsManager(null, new Mock<ILogger>().Object, new ProjectBinderFactory(serviceProvider), fileSystem));
+            Exceptions.Expect<ArgumentNullException>(() => new ConflictsManager(serviceProvider, null, new ProjectBinderFactory(serviceProvider), fileSystem));
             Exceptions.Expect<ArgumentNullException>(() => new ConflictsManager(serviceProvider, new Mock<ILogger>().Object,null, fileSystem));
-            Exceptions.Expect<ArgumentNullException>(() => new ConflictsManager(serviceProvider, new Mock<ILogger>().Object,new ProjectBinderFactory(), null));
+            Exceptions.Expect<ArgumentNullException>(() => new ConflictsManager(serviceProvider, new Mock<ILogger>().Object,new ProjectBinderFactory(serviceProvider), null));
```
duncanp-sonar, line 93: I know you mentioned re-writing the conflicts manager and some other tests later, but as you continue refactoring look out for tests that you could fix up immediately e.g. it would have been as easy to make this one use a mock of `IProjectBindingFactory` and then it might not need to change again.

src/Integration/Binding/ProjectBinderFactory.cs
rita-g-sonarsource: Yep, otherwise it will throw "conditional expression cannot be determined because there is no implicit conversion"

Issue #1129, opened 2019-12-06
Stop using api/qualityprofiles/export to fetch C#/VB rulesets


duncanp-sonar:
### Description

SLVS uses the legacy profile exporter mechanism to fetch rulesets for C# and VB.
It should use `api/rules/search` instead (same as S4MSB).

See #1099 for more background.

TODO:
- [x] check the implementation in the Sonar Scanner for MSBuild. DONE
- [x] decide whether to include security rules in the ruleset/SonarLint.XML. CHECKED - should not appear in the ruleset or SonarLint.xml.

PR #1356, opened 2020-04-28
Added RoslynProfileExporter replacement classes to Core

duncanp-sonar:
Relates to #1129 

Moved previously reviewed classes from the web client branch.
* no code changes
* changed copyright notices
* added classes to `Core` in the namespace `SonarLint.VisualStudio.Core.CSharpVB`.

Core already has a namespace called `SonarLint.VisualStudio.Core.CFamily`, so I went with `SonarLint.VisualStudio.Core.CSharpVB` for the sub-folder containing the  C#/VB-specific classes.
I'm trying to find an alternative to using `Roslyn` for code that doesn't actually use/reference any of the Roslyn assemblies. `DotNet` is an alternative, but it isn't strictly accurate either so I though I'd try `CSharpVB` to see how it feels.

### DIFF
--- a/src/Core.UnitTests/CSharpVB/NuGetPackageInfoGeneratorTests.cs
+++ b/src/Core.UnitTests/CSharpVB/NuGetPackageInfoGeneratorTests.cs
@@ -1,0 +1,183 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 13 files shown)


rita-g-sonarsource:
> I'm trying to find an alternative to using `Roslyn` for code that doesn't actually use/reference any of the Roslyn assemblies. `DotNet` is an alternative, but it isn't strictly accurate either so I though I'd try `CSharpVB` to see how it feels.

I think Roslyn doesn't necessarily have to mean "uses roslyn sdk", but perhaps it should be RoslynAnalyzer (i.e. `RoslynAnalyzerPluginExtensions", "RoslynAnalyzerProjectBinder", etc)? Food for thought, we can decide later.

PR #1358, opened 2020-04-28
Group and sort rules in generated ruleset

duncanp-sonar:


src/Core.UnitTests/CSharpVB/RuleSetGeneratorTests.cs
```diff
@@ -234,6 +234,61 @@ public void RoslynRuleSet_Sonar_Rules_Added()
...
+            };
+
+            // Act
+            var ruleSet = generator.Generate("cs", activeRules, inactiveRules);
+
+            // Assert
+            ruleSet.Rules.Should().HaveCount(3);
+
+            // Expecting groups to be sorted alphabetically by analyzer id (not namespace)...
+            ruleSet.Rules[0].AnalyzerId.Should().Be("AAA wintellect id");
```
rita-g-sonarsource, line 277: It was an effort for me to understand the test, mostly because it was difficult to follow the line from setup to assertions. 
I think would've helped to have more descriptive names, such as   
`{ "first.analyzerId", "first analyzer id" }`
`ruleSet.Rules[0].AnalyzerId.Should().Be("first analyzer id");`

duncanp-sonar: With like `{ "first.analyzerId", "first analyzer id" }`, the test would still pass if the groups were sorted by the property key ("first.analyzer.id").
I've tried to simplify the setup slightly.

duncanp-sonar: (BTW if you've finished a review, please mark it as "requires changes" or "approved" so I know you're not still looking at it).

rita-g-sonarsource: Sorry my bad, I thought I marked it as Approved :)

PR #1359, opened 2020-04-28
Added test for security rules in generated ruleset

duncanp-sonar:
Added unit test to check handling of SonarSecurity rules in the generated ruleset.

The security plugin is set up to work like a SonarQube Roslyn SDK plugin, so the rules are included and can be set to active/inactive.


duncanp-sonar:
Superceded.

PR #1361, opened 2020-04-29
Exclude taint-analysis rules from ruleset and SL config

duncanp-sonar:


src/Core.UnitTests/CSharpVB/RuleSetGeneratorTests.cs
```diff
@@ -234,6 +234,34 @@ public void RoslynRuleSet_Sonar_Rules_Added()
                 "Warning", "Warning", "None", "None");
         }
 
+        [TestMethod]
+        [DataRow("cs")]
+        [DataRow("vbnet")]
```
duncanp-sonar, line 239: There currently aren't any taint analysis rules for VB, but they would be excluded too (assuming a VB security analyzer followed the same naming pattern).

src/Core.UnitTests/CSharpVB/SonarLintConfigGeneratorTests.cs
```diff
@@ -165,6 +165,24 @@ public void Generate_ValidRules_OnlyRulesFromKnownRepositoryReturned(string know
             actual.Rules.Select(r => r.Key).Should().BeEquivalentTo(new string[] { "valid1", "valid2", "valid3" });
         }
 
+        [TestMethod]
+        [DataRow("cs")]
+        [DataRow("vbnet")]
+        public void Generate_SonarSecurityRules_AreNotReturned(string language)
```
duncanp-sonar, line 171: No changes are required to the SonarLint config generator since it specifically checks for the "normal" (i.e. non-security) C# and VB analyzer ids and ignores all others, including third-party plugins packaged using the SonarQube Roslyn SDK.

src/Core/CSharpVB/RuleSetGenerator.cs
```diff
@@ -93,9 +94,12 @@ public RuleSet Generate(string language, IEnumerable<SonarQubeRule> activeRules,
         private static bool IsSupportedRuleRepo(IGrouping<string, SonarQubeRule> analyzerRules)
         {
             var partialRepoKey = analyzerRules.Key;
-            return !string.IsNullOrEmpty(partialRepoKey);
+            return !string.IsNullOrEmpty(partialRepoKey) && !IsExcludedRuleRepo(partialRepoKey);
         }
 
+        private static bool IsExcludedRuleRepo(string partialRepoKey) =>
+            partialRepoKey.IndexOf("sonaranalyzer.security", StringComparison.OrdinalIgnoreCase) >= 0;
+
```
duncanp-sonar, line 102: There were lots of ways I could have written this check. I went for a fairly loose check that would match VB rules if they were added.

src/Core.UnitTests/CSharpVB/SonarLintConfigGeneratorTests.cs
rita-g-sonarsource: I think it's worth uniting with Generate_ValidRules_OnlyRulesFromKnownRepositoryReturned

since it's exactly the same use case.

duncanp-sonar: Sorry, I didn't notice this comment before merging.

I thought about it when writing the test and was in two minds about it. I decided in the end to have a security-specific test to match the corresponding security-specific test for the ruleset generator, but I didn't have a strong preference.

rita-g-sonarsource: It's ok, it's not that important. It just confused me a bit to see the tests change, even though the class didn't.

duncanp-sonar: That's why I added the review comment above...

PR #1362, opened 2020-04-29
Removed INuGetBindingOperation dependency on RoslynExportProfile

duncanp-sonar:
A small refactoring to remove one of the dependencies on the data class returned by the call to the soon-to-be-obsolete `GetRoslynExportProfileAsync` method.

I've changed the data types used in the interface from `RoslynProfileExportResponse` to `IEnumerable<NuGetPackageInfo>` and fixed up the tests.

There's a one-line temporary hack in the binding code to translate from the old to the new data type.

### DIFF
--- a/src/Core/CSharpVB/NuGetPackageInfo.cs
+++ b/src/Core/CSharpVB/NuGetPackageInfo.cs
@@ -1,0 +1,36 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarLint.VisualStudio.Core.CSharpVB
+{
+    public class NuGetPackageInfo
+    {
+        public NuGetPackageInfo(string id, string version)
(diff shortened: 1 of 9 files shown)


src/Core/CSharpVB/NuGetPackageInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarLint.VisualStudio.Core.CSharpVB
+{
+    public class NuGetPackageInfo
```
duncanp-sonar, line 25: A new data class that describes a single NuGet package.

It has the same data as the `SonarQube.Client.Messages.NuGetPackageInfoResponse` class. I've added it so the Core doesn't have to return the `NuGetPackageInfoResponse` in the public API.

src/Integration/Binding/INuGetBindingOperation.cs
```diff
@@ -43,6 +43,6 @@ internal interface INuGetBindingOperation
         /// Extracts any required information from the supplied Roslyn export profile
         /// </summary>
         /// <returns>True if processing was successful, otherwise false</returns>
-        bool ProcessExport(Language language, RoslynExportProfileResponse exportProfileResponse);
+        bool ProcessExport(Language language, IEnumerable<NuGetPackageInfo> nugetPackages);
```
duncanp-sonar, line 46: This is the main part of the change: we aren't going to use the `RoslynExportProfileResponse` any more, so the the NuGet binding interfaces need to use a different type.

src/Integration/Binding/DotNetBindingConfigProvider.cs
```diff
@@ -96,8 +96,11 @@ public async Task<IBindingConfigFile> GetConfigurationAsync(SonarQubeQualityProf
                 return null;
             }
 
+            // duncanp - WIP - next line will be removed when the RoslynProfileExporter is removed
+            var nugetPackages = roslynProfileExporter.Deployment.NuGetPackages.Select(x => new Core.CSharpVB.NuGetPackageInfo(x.Id, x.Version));
```
duncanp-sonar, line 100: Temporary workaround to limit the size of this PR.
After this I'll update the `DotNetBindingConfigProvider` to stop using the `RoslynProfileExporter` entirely.

src/Core/CSharpVB/NuGetPackageInfo.cs
rita-g-sonarsource: I agree

PR #1363, opened 2020-04-29
[Infra] Add custom local ruleset to enable additional rules

duncanp-sonar:


duncanp-sonar:
This will enable the rule in the IDE, but not in the CI build.

rita-g-sonarsource:
@duncanp-sonar let's enable it!

rita-g-sonarsource:
> This will enable the rule in the IDE, but not in the CI build.

Actually, now that I think about it, enabling it locally won't be enough.
1) I usually have the extension turned off, so I could test it.
2) I extract to files automatically, so I might not even see the file.

I think we should enable it in SonarCloud.

duncanp-sonar:
I did this as a test to see what happened. The scanner merges a new ruleset generated from the QP with the local ruleset. Unfortunately, the new generated ruleset contains entries for all of the SQ rules,  so it effectively replaces the local config.

PR #1364, opened 2020-04-29
Updated NuGet info generator to exclude taint-analysis packages

duncanp-sonar:
There are two small commits. It might be easier to look at them separate in the GitHub web UI.
The first just changes an existing test to be data-driven.
The second fixes the bug in the NuGet generator.

### DIFF
--- a/src/Core.UnitTests/CSharpVB/NuGetPackageInfoGeneratorTests.cs
+++ b/src/Core.UnitTests/CSharpVB/NuGetPackageInfoGeneratorTests.cs
@@ -109,6 +109,30 @@ namespace SonarQube.Client.Tests.RoslynExporterAdapter
         }
 
         [TestMethod]
+        [DataRow("cs")]
+        [DataRow("vbnet")]
+        public void Get_SonarSecurityAnalyzer_WithPropertiesAndMatchingRules_SecurityPackageAreNotReturned(string language)
+        {
+            string propertyPrefix = $"sonaranalyzer.security.{language}";
+            string repoKey = $"roslyn.{propertyPrefix}";
+
+            var properties = new Dictionary<string, string>
+            {
+                { $"{propertyPrefix}.analyzerId", "AAA" },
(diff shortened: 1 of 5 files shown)


src/Core.UnitTests/CSharpVB/NuGetPackageInfoGeneratorTests.cs
```diff
@@ -76,7 +76,7 @@ public void Get_MissingVersionProperty_NoPackageInfo()
         }
 
         [TestMethod]
-        public void Get_MissingPluginVerionProperty_NoPackageInfo()
+        public void Get_MissingPluginVersionProperty_NoPackageInfo()
```
duncanp-sonar, line 79: Miscellaneous - fix typo in "Version"

src/Core/CSharpVB/NuGetPackageInfoGenerator.cs
```diff
@@ -49,7 +49,7 @@ public static class NuGetPackageInfoGenerator
         private static IEnumerable<string> GetPluginPropertyPrefixes(IEnumerable<SonarQubeRule> rules) =>
             rules.Select(r => r.TryGetRoslynPluginPropertyPrefix())
                 .Distinct()
-                .Where(r => !string.IsNullOrEmpty(r))
+                .Where(r => !string.IsNullOrEmpty(r) && !RoslynPluginRuleKeyExtensions.IsExcludedRuleRepository(r))
```
rita-g-sonarsource, line 52: There is a duplication here.
Both NugetPackageInfoGenerator and RuleSetGenerator go through the same process: transforming active rules into repository keys and then filtering.
I think it should be done in one place, in a higher caller.
It would simplify the process and the tests, as we won't have to check everything twice.
WDYT?

duncanp-sonar: As discussed offline, I agree there's a duplication in the filtering logic but I don't think it's worth trying to change now.

PR #1365, opened 2020-04-30
Store FilePath in IBindingConfigFile + remove ConfigFileInformation

rita-g-sonarsource:


src/Core/Binding/IBindingConfigProvider.cs
```diff
@@ -44,9 +44,11 @@ public interface IBindingConfigProvider
     /// For C++ it will be in a json file in a Sonar-specific format</remarks>
     public interface IBindingConfigFile
     {
+        string FilePath { get; }
```
rita-g-sonarsource, line 47: This is main change: we will now store the filePath of the ruleset inside the BindingConfigFile.

src/Core/CFamily/CFamilyBindingConfigProvider.cs
```diff
@@ -68,7 +75,12 @@ public async Task<IBindingConfigFile> GetConfigurationAsync(SonarQubeQualityProf
             cancellationToken.ThrowIfCancellationRequested();
 
             var settings = CreateRulesSettingsFromQPRules(result);
-            var configFile = new CFamilyBindingConfigFile(settings);
+            var settingsFilePath = solutionBindingFilePathGenerator.Generate(
```
rita-g-sonarsource, line 78: This is where the magic happens -- calculating and assigning the path of the ruleset to IBindingConfig

src/Integration.UnitTests/Binding/SolutionBindingOperationTests.cs
```diff
@@ -424,13 +427,6 @@ private void ExecuteCommitSolutionBindingTest(SonarLintMode bindingMode)
             savedProject.Profiles[Language.CSharp].ProfileTimestamp.Should().Be(expectedTimeStamp);
         }
 
-        [TestMethod]
-        public void SolutionBindingOperation_ConfigFileInformation_Ctor_ArgChecks()
```
rita-g-sonarsource, line 428: Test no longer relevant as the class ConfigFileInformation was deleted.

src/Integration.UnitTests/Binding/SolutionBindingOperationTests.cs
```diff
@@ -456,28 +452,39 @@ public void SolutionBindingOperation_ConfigFileInformation_Ctor_ArgChecks()
...
 
-        private Mock<IBindingConfigFileWithRuleset> CreateMockRuleSetConfigFile(string displayName)
+        private Mock<IBindingConfigFileWithRuleset> CreateMockRuleSetConfigFile(SonarLintMode mode, Language language)
         {
             var rulesetConfig = new Mock<IBindingConfigFileWithRuleset>();
             rulesetConfig.Setup(x => x.RuleSet)
-                .Returns(new RuleSet(displayName));
+                .Returns(new RuleSet(language.Name));
+
+            var folder = mode == SonarLintMode.Connected
```
rita-g-sonarsource, line 461: The tests assert that the file was saved in the correct location, and it was tedious to re-write them, so I just calculated what they're expecting to assert.

src/Integration/Binding/DotNetBindingConfigProvider.cs
```diff
@@ -105,7 +112,12 @@ public async Task<IBindingConfigFile> GetConfigurationAsync(SonarQubeQualityProf
             // Remove/Move/Refactor code when XML ruleset file is no longer downloaded but the proper API is used to retrieve rules
             UpdateDownloadedSonarQubeQualityProfile(ruleSet, qualityProfile, bindingConfiguration.Project.ProjectName, bindingConfiguration.Project.ServerUri.ToString());
 
-            return new DotNetBindingConfigFile(ruleSet);
+            var ruleSetFilePath = solutionBindingFilePathGenerator.Generate(
```
rita-g-sonarsource, line 115: Calculation of ruleSetPath for IBindingConfigFile. Same as
https://github.com/SonarSource/sonarlint-visualstudio/pull/1365/files#r418649075

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -131,30 +133,17 @@ public void RegisterKnownConfigFiles(IDictionary<Language, IBindingConfigFile> l
...
-
-            foreach (var keyValue in languageToFileMap)
-            {
-                Debug.Assert(!this.bindingConfigInformationMap.ContainsKey(keyValue.Key), "Attempted to register an already registered rule set. Group:" + keyValue.Key);
-
-                string solutionRuleSet = ruleSetInfo.CalculateSolutionSonarQubeRuleSetFilePath(this.projectKey, keyValue.Key, this.bindingMode);
-                this.bindingConfigInformationMap[keyValue.Key] = new ConfigFileInformation(keyValue.Value) { NewFilePath = solutionRuleSet };
-            }
+            bindingConfigInformationMap.Clear();
+            bindingConfigInformationMap.AddRange(languageToFileMap);
```
rita-g-sonarsource, line 137: woohoo, no longer need to calculate the file path!

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -203,25 +192,25 @@ public void Prepare(CancellationToken token)
...
-                Debug.Assert(!string.IsNullOrWhiteSpace(info.NewFilePath), "Expected to be set during registration time");
-                
-                this.sourceControlledFileSystem.QueueFileWrite(info.NewFilePath, () =>
+                var info = keyValue.Value;
+                Debug.Assert(!string.IsNullOrWhiteSpace(info.FilePath), "Expected to be set during registration time");
+
+                sourceControlledFileSystem.QueueFileWrite(info.FilePath, () =>
                 {
-                    var ruleSetDirectoryPath = Path.GetDirectoryName(info.NewFilePath);
+                    var ruleSetDirectoryPath = Path.GetDirectoryName(info.FilePath);
```
rita-g-sonarsource, line 200: This is still here since SolutionBindingOperation has a Commit method so I couldn't easily to move it into IBindingConfigFile.Save

rita-g-sonarsource:
@duncanp-sonar I'm sorry that this PR has so many files -- the filePath was referenced in so many places. The changes are:
1. Calculating the filePath inside IBindingConfigProvider (and setting it to the new FilePath property in IBindingConfigFile.
2. Deletion of the class ConfigFileInformation as its only purpose was to wrap IBindingConfigFile and add the filePath to it.

src/Integration/Binding/SolutionBindingOperation.cs
duncanp-sonar: It looks like we only need the `FilePath` here so the solution binder can create the folder if required.

The solution binder should already have ensured the solution-level binding folder exists. If any other sub-folders are required then creating them should be the responsibility of the language-specific binder (as will be the case when we add the C#/VB additional files).

Then, we could remove `FilePath` from the interface and just call `info.Save()` here.

I wouldn't do it now, but we should definitely look at whether we can remove `FilePath` from the interface later.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -289,10 +278,10 @@ private void PendBindingInformation(ConnectionInformation connInfo)
 
         private void UpdateSolutionFile()
         {
-            foreach (ConfigFileInformation info in bindingConfigInformationMap.Values)
+            foreach (var info in bindingConfigInformationMap.Values)
             {
-                Debug.Assert(fileSystem.File.Exists(info.NewFilePath), "File not written " + info.NewFilePath);
-                legacyConfigFolderItemAdder.AddToFolder(info.NewFilePath);
+                Debug.Assert(fileSystem.File.Exists(info.FilePath), "File not written " + info.FilePath);
+                legacyConfigFolderItemAdder.AddToFolder(info.FilePath);
```
duncanp-sonar, line 284: Hhhmm. This will complicate removing `FilePath` from the interface.

src/Integration/Binding/ProjectBindingOperation.Writer.cs
```diff
@@ -45,22 +46,22 @@ internal partial class ProjectBindingOperation
         /// <param name="ruleSetFileName">The rule set file name</param>
         /// <param name="solutionRuleSet\">Full path of the parent solution-level SonarQube rule set</param>
         /// <returns>Full file path of the file that we expect to write to</returns>
-        internal /*for testing purposes*/ string QueueWriteProjectLevelRuleSet(string projectFullPath, string ruleSetFileName, ConfigFileInformation solutionRuleSet, string currentRuleSetPath)
+        internal /*for testing purposes*/ string QueueWriteProjectLevelRuleSet(string projectFullPath, string ruleSetFileName, IBindingConfigFile solutionRuleSet, string currentRuleSetPath)
```
duncanp-sonar, line 49: `solutionRuleSet` isn't a great name any more, although it could be changed when renaming `IBindingConfigFile`.

src/Integration/Binding/ProjectBindingOperation.Writer.cs
```diff
@@ -45,22 +46,22 @@ internal partial class ProjectBindingOperation
...
         {
             Debug.Assert(!string.IsNullOrWhiteSpace(projectFullPath));
             Debug.Assert(!string.IsNullOrWhiteSpace(ruleSetFileName));
             Debug.Assert(solutionRuleSet != null);
 
             // TODO: clean up to remove down-cast to check for RuleSet
-            solutionRuleSet.BindingConfigFile.TryGetRuleSet(out var dotNetRuleSet);
+            solutionRuleSet.TryGetRuleSet(out var dotNetRuleSet);
```
duncanp-sonar, line 56: Not part of this change, but the "TODO" above is now redundant and can be removed later - we are now deliberately expecting a language-specific interface at this point.

src/Integration.UnitTests/Binding/SolutionBindingOperationTests.cs
duncanp-sonar: Does that mean we now have tests that are just checking that the mocks return the values expected by the tests? 
Are the tests still testing anything useful? If not, just delete them.
If the tests do still other useful checks then why not just remove the checks for the specific values returned by the mocks? (it might enough just to check that the expected mock is being called).

src/Core/CFamily/CFamilyBindingConfigProvider.cs
duncanp-sonar: I'm not sure the `ISolutionBindingFilePathGenerator` interface will survive much longer; I'm not convinced it's still necessary or appropriate.

rita-g-sonarsource: What do you think we should do instead?

src/Integration/Binding/SolutionBindingOperation.cs
rita-g-sonarsource: Yep :(

src/Core/CFamily/CFamilyBindingConfigProvider.cs
duncanp-sonar: The interface centralises the file-naming logic for all languages.
In general, the refactoring is devolving language-specific handling to language-specific classes. I'm questioning whether the centralised interface is still appropriate, particularly since the additional file naming logic will be specific to C#/VB.

Either there should be a language-specific name calculator, or the binder should just handle the calculation itself. I don't think it need to be centralised, and I'm not sure it needs to be pluggable for testing (for C++, the logic is simple enough to be handled in a static method on the CFamily binder).

Have a look at `Language.FileSuffixAndExtension`. I had to add it to make the file name calculation work when I added support for CFamily connected mode. I don't think that property should be needed.

Either way, it doesn't affect this PR.

PR #1367, opened 2020-04-30
Move BindingConfiguration to Core

rita-g-sonarsource:


rita-g-sonarsource:
@duncanp-sonar Please take a look at the two commits separately: one is the file move, and the second is the namespace change.

src/Core/Binding/BindingConfiguration.cs
```diff
@@ -37,7 +37,7 @@ public static BindingConfiguration CreateBoundConfiguration(BoundSonarQubeProjec
             return new BindingConfiguration(project, sonarLintMode);
         }
 
-        internal /*for testing purposes*/ BindingConfiguration(BoundSonarQubeProject project, SonarLintMode mode)
+        public BindingConfiguration(BoundSonarQubeProject project, SonarLintMode mode)
```
duncanp-sonar, line 39: This was only being called from tests before - why does it need to be public now?

rita-g-sonarsource: It wasn't, it's also called from SolutionBindingOperation. I changed it to use the static method, and then a bunch of tests failed, since the tests pass null.

src/Core/ICredentials.cs
```diff
@@ -21,7 +21,7 @@
 using System;
 using SonarQube.Client.Models;
 
-namespace SonarLint.VisualStudio.Integration.Persistence
+namespace SonarLint.VisualStudio.Core
```
duncanp-sonar, line 24: Why `Core` rather than `Core.Binding`?

rita-g-sonarsource: I thought it was credentials in general. Moved to Binding.

PR #1368, opened 2020-04-30
Added interfaces for testing

duncanp-sonar:


src/Core/CSharpVB/RuleSetGenerator.cs
```diff
@@ -32,50 +32,40 @@
...
         private const string SONARANALYZER_PARTIAL_REPO_KEY = "sonaranalyzer-{0}";
         private const string ROSLYN_REPOSITORY_PREFIX = "roslyn.";
 
-        private readonly IDictionary<string, string> sonarProperties;
         private readonly string inactiveRuleActionText = GetActionText(RuleAction.None);
 
         private readonly string activeRuleActionText = GetActionText(RuleAction.Warning);
 
-        public RuleSetGenerator(IDictionary<string, string> sonarProperties)
+        public RuleSet Generate(string language, IEnumerable<SonarQubeRule> rules, IDictionary<string, string> sonarProperties)
```
duncanp-sonar, line 48: Changed the method signature so the properties don't need to be passed in the constructor.

PR #1369, opened 2020-04-30
Pass binding configuration to binding config providers

rita-g-sonarsource:
Relates to #176

### DIFF
--- a/src/Core/Binding/IBindingConfigProvider.cs
+++ b/src/Core/Binding/IBindingConfigProvider.cs
@@ -34,7 +34,7 @@ namespace SonarLint.VisualStudio.Core.Binding
         /// <summary>
         /// Returns a configuration file for the specified language
         /// </summary>
-        Task<IBindingConfigFile> GetConfigurationAsync(SonarQubeQualityProfile qualityProfile, string organizationKey, Language language, CancellationToken cancellationToken);
+        Task<IBindingConfigFile> GetConfigurationAsync(SonarQubeQualityProfile qualityProfile, Language language, CancellationToken cancellationToken);
     }
 
     /// <summary>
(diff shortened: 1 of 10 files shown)


src/Core/Binding/IBindingConfigProvider.cs
```diff
@@ -34,7 +34,7 @@ public interface IBindingConfigProvider
         /// <summary>
         /// Returns a configuration file for the specified language
         /// </summary>
-        Task<IBindingConfigFile> GetConfigurationAsync(SonarQubeQualityProfile qualityProfile, string organizationKey, Language language, CancellationToken cancellationToken);
+        Task<IBindingConfigFile> GetConfigurationAsync(SonarQubeQualityProfile qualityProfile, Language language, CancellationToken cancellationToken);
```
rita-g-sonarsource, line 37: This param is now taken from the binding configuration, that is passed in ctor

src/Core/CFamily/CFamilyBindingConfigProvider.cs
```diff
@@ -33,11 +33,13 @@ namespace SonarLint.VisualStudio.Core.CFamily
     public class CFamilyBindingConfigProvider : IBindingConfigProvider
     {
         private readonly ISonarQubeService sonarQubeService;
+        private readonly BindingConfiguration bindingConfiguration;
         private readonly ILogger logger;
 
-        public CFamilyBindingConfigProvider(ISonarQubeService sonarQubeService, ILogger logger)
+        public CFamilyBindingConfigProvider(ISonarQubeService sonarQubeService, BindingConfiguration bindingConfiguration, ILogger logger)
```
rita-g-sonarsource, line 39: It doesn't currently use it, but it will (to calculate the file path)

PR #1371, opened 2020-04-30
Move PathHelper to Core

rita-g-sonarsource:
Infra work for #176

### DIFF
--- a/src/Integration.UnitTests/Helpers/PathHelperTests.cs
+++ b/src/Integration.UnitTests/Helpers/PathHelperTests.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SonarLint for Visual Studio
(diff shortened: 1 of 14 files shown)


rita-g-sonarsource:
This is so I could use PathHelper.EscapeFileName, which is part of the logic inside GenerateSolutionRuleSetPath (which will be moved to Core as well).

src/Integration.UnitTests/Helpers/PathHelperTests.cs
```diff
@@ -21,6 +21,7 @@
 using System;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.Core.Helpers;
```
duncanp-sonar, line 24: This test file should be moved too.

PR #1372, opened 2020-04-30
Removed call to GetRoslynExportProfileAsync

duncanp-sonar:


src/Integration/Binding/DotNetBindingConfigProvider.cs
```diff
@@ -44,83 +45,98 @@ internal class DotNetBindingConfigProvider : IBindingConfigProvider
...
+            this.nuGetPackageInfoGenerator = nuGetPackageInfoGenerator;
         }
 
         public bool IsLanguageSupported(Language language)
         {
             return Language.CSharp.Equals(language) || Language.VBNET.Equals(language);
         }
 
-        public async Task<IBindingConfigFile> GetConfigurationAsync(SonarQubeQualityProfile qualityProfile, string organizationKey, Language language, CancellationToken cancellationToken)
+        public Task<IBindingConfigFile> GetConfigurationAsync(SonarQubeQualityProfile qualityProfile, string UNUSED_organizationKey_TODO_REMOVE, Language language, CancellationToken cancellationToken)
```
duncanp-sonar, line 75: I've split this method into two parts as recommended by S4457 - arg checks in the non-async part, async part in a separate method.

src/Integration/Binding/DotNetBindingConfigProvider.cs
```diff
@@ -44,83 +45,98 @@ internal class DotNetBindingConfigProvider : IBindingConfigProvider
...
-                    organizationKey, serverLanguage, cancellationToken),
-                this.logger);
-            if (roslynProfileExporter == null)
-            {
-                this.logger.WriteLine(string.Format(Strings.SubTextPaddingFormat,
-                    string.Format(Strings.QualityProfileDownloadFailedMessageFormat, qualityProfile.Name,
-                        qualityProfile.Key, language.Name)));
-                return null;
-            }
+            // First, fetch the active rules
```
duncanp-sonar, line 91: We now fetch the data in multiple calls. I've optimised the calls so we don't fetch data until it is needed, so if the function returns early some of the data won't be fetch.

src/Integration/Binding/DotNetBindingConfigProvider.cs
```diff
@@ -44,83 +45,98 @@ internal class DotNetBindingConfigProvider : IBindingConfigProvider
...
-            return new DotNetBindingConfigFile(ruleSet);
-        }
+            // Create the ruleset
+            var coreRuleset = ruleSetGenerator.Generate(language.ServerLanguage.Key, activeRules.Union(inactiveRules), sonarProperties);
 
-        private void UpdateDownloadedSonarQubeQualityProfile(RuleSet ruleSet, SonarQubeQualityProfile qualityProfile, string projectName, string serverUrl)
-        {
-            ruleSet.NonLocalizedDisplayName = string.Format(Strings.SonarQubeRuleSetNameFormat, projectName, qualityProfile.Name);
+            // Set the name and description
+            coreRuleset.Name = string.Format(Strings.SonarQubeRuleSetNameFormat, projectName, qualityProfile.Name);
```
duncanp-sonar, line 123: I'm not sure setting the name and description is useful (particularly having the API URL encoded in the description).
It isn't visible in the VS UI anywhere so the user would only see it if the open the file in a text editor.
Also, we need to pass in extra fields just to construct the messages.

However, I've left it as-if for now. Then only difference is that the separator in the description is now a space rather than a line-break. The line-break was being escaped in the file since XML attributes have to be single-line.

src/Integration.UnitTests/Binding/DotNetBindingConfigProviderTests.cs
```diff
@@ -24,15 +24,11 @@
...
 using Moq;
-using NuGet;
 using SonarLint.VisualStudio.Integration.Binding;
-using SonarLint.VisualStudio.Integration.NewConnectedMode;
 using SonarLint.VisualStudio.Integration.Resources;
 using SonarQube.Client;
-using SonarQube.Client.Messages;
 using SonarQube.Client.Models;
 using Language = SonarLint.VisualStudio.Core.Language;
 using NuGetPackageInfo = SonarLint.VisualStudio.Core.CSharpVB.NuGetPackageInfo;
```
duncanp-sonar, line 34: There's a naming clash between the `RuleSet` in Core and the VS `RuleSet`. I'm hoping it will occur less often after some of the later changes.
If it's still a problem we can consider renaming our version.

src/Integration/Binding/DotNetBindingConfigProvider.cs
```diff
@@ -44,83 +45,98 @@ internal class DotNetBindingConfigProvider : IBindingConfigProvider
...
 
-            var tempRuleSetFilePath = Path.GetTempFileName();
-            File.WriteAllText(tempRuleSetFilePath, roslynProfileExporter.Configuration.RuleSet.OuterXml);
-            RuleSet ruleSet = RuleSet.LoadFromFile(tempRuleSetFilePath);
-
-            // Give up if the quality profile is empty
-            if (ruleSet == null ||
-                ruleSet.Rules.Count == 0 ||
-                ruleSet.Rules.All(rule => rule.Action == RuleAction.None))
+            // Give up if the quality profile is empty - not point in fetching anything else
```
rita-g-sonarsource, line 95: This function is already getting quite long, maybe extract the 3 checks to 3 separate functions? (each with the logging in it)

src/Integration.UnitTests/Binding/DotNetBindingConfigProviderTests.cs
```diff
@@ -42,245 +38,302 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
...
-
-            var language = Language.CSharp;
-            var profile = this.ConfigureProfileExport(null, language, "");
+            var builder = new TestEnvironmentBuilder(ValidQualityProfile, Language.Cpp);
+            var testSubject = builder.CreateTestSubject();
 
-            // Act
-            var result = await testSubject.GetConfigurationAsync(profile, null, language, CancellationToken.None)
-                .ConfigureAwait(false);
+            // 1. Supported languages
```
rita-g-sonarsource, line 88: You can use DataTestMethod(language, shouldBeSupported).
It's a minor change but it makes the tests a lot more readable, and easier to understand when they fail. It's more descriptive than using comments.

duncanp-sonar: Unfortunately `Language` is a class so instances can't be passed in attributes.

src/Integration/Binding/DotNetBindingConfigProvider.cs
duncanp-sonar: I've extracted a couple of methods.

src/Integration.UnitTests/Binding/DotNetBindingConfigProviderTests.cs
duncanp-sonar: FYI for review comments like this it's enough write "Why not use DataRow?" - no need to include an explanation.

PR #1373, opened 2020-04-30
Persist SonarQubeProject instead of BindingConfiguration

rita-g-sonarsource:


src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -278,8 +278,7 @@ private void PendBindingInformation(ConnectionInformation connInfo)
                 credentials, connInfo.Organization);
             bound.Profiles = map;
 
-            var config = new BindingConfiguration(bound, this.bindingMode);
```
rita-g-sonarsource, line 281: I'm working on passing the config location as a ctor param in BindingConfiguration, and this bit was in the way.
We don't even need to create BindingConfiguration, as the inner classes save the SonarQubeProject, and they are the ones that know the path in the first place.

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -279,28 +280,15 @@ private BindingConfiguration SafeGetBindingConfig()
         #endregion
 
         #region IServiceProvider
+
         private void RegisterLocalServices()
         {
             this.localServices.Add(typeof(ISolutionRuleSetsInformationProvider), new Lazy<ILocalService>(() => new SolutionRuleSetsInformationProvider(this, Logger)));
             this.localServices.Add(typeof(IRuleSetSerializer), new Lazy<ILocalService>(() => new RuleSetSerializer()));
             this.localServices.Add(typeof(ICredentialStoreService), new Lazy<ILocalService>(() => new CredentialStore(new SecretStore("SonarLint.VisualStudio.Integration"))));
-            this.localServices.Add(typeof(IConfigurationProvider), new Lazy<ILocalService>(() =>
```
rita-g-sonarsource, line 287: Extracted to function as-is

PR #1374, opened 2020-05-01
Extract solution file path generation to another class

rita-g-sonarsource:
ISolutionBindingFilePathGenerator contains the logic from SolutionRuleSetsInformationProvider.GenerateSolutionRuleSetPath. Extracted as-is and added a few tests.

Please note the base: let me know if I should merge this to my base branch instead.


duncanp-sonar:
Base branch: I can't answer that without knowing more about the context. Why does this change need to go in the main feature branch?

PR #1389, opened 2020-05-06
Pass IBindingConfigFileWithRuleset to ProjectBindingOperation

rita-g-sonarsource:


src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -210,14 +192,25 @@ public void Prepare(CancellationToken token)
...
-            foreach (var binder in this.childBinder)
+            foreach (var project in projects)
             {
                 if (token.IsCancellationRequested)
                 {
                     return;
                 }
-
-                binder.Prepare(token);
+                if (projectBinderFactory.Get(project) is RoslynProjectBinder)
```
rita-g-sonarsource, line 201: This is the reason for the change:
I will change this code to be 
```
var projectBinder = projectBinderFactory.Get(project);
projectBinder.Bind(project, bindingConfigFile, token);
```
And the roslyn implementation will create ProjectBindingOperation and call it.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -210,14 +192,25 @@ public void Prepare(CancellationToken token)
...
+                {
+                    var languageForProject = ProjectToLanguageMapper.GetLanguageForProject(project);
+                    var bindingConfigFile = GetBindingConfig(languageForProject) as IBindingConfigFileWithRuleset;
+                    var binder = new ProjectBindingOperation(serviceProvider, project, bindingConfigFile);
+                    binder.Initialize();
```
rita-g-sonarsource, line 206: I need to unite now the calls to Initialize and Prepare, since the bindingConfigFile is not available in the solution Initialize phase, only on Prepare.

src/Integration.UnitTests/Binding/SolutionBindingOperationTests.cs
```diff
@@ -214,13 +212,6 @@ public void SolutionBindingOperation_Initialization()
 
             // Assert
             testSubject.SolutionFullPath.Should().Be(Path.Combine(SolutionRoot, "xxx.sln"));
-            testSubject.Binders.Should().HaveCount(3, "Should be one per managed project");
```
rita-g-sonarsource, line 217: No longer relevant, as the binders are now created during Prepare phase and not Initialize

src/Integration.UnitTests/Binding/SolutionBindingOperationTests.cs
```diff
@@ -259,12 +251,21 @@ public void SolutionBindingOperation_Prepare()
             testSubject.Prepare(CancellationToken.None);
 
             // Assert
-            prepareCalledForBinder.Should().BeTrue("Expected to propagate the prepare call to binders");
```
rita-g-sonarsource, line 262: I will be able to test that the relevant Binders are called only in future commits, after I change the code to use IProjectBinder instead of ProjectBindingOperation.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -210,14 +192,25 @@ public void Prepare(CancellationToken token)
...
             {
                 if (token.IsCancellationRequested)
                 {
                     return;
                 }
-
-                binder.Prepare(token);
+                if (projectBinderFactory.Get(project) is RoslynProjectBinder)
+                {
+                    var languageForProject = ProjectToLanguageMapper.GetLanguageForProject(project);
```
rita-g-sonarsource, line 203: This is how we calculate the language inside ProjectBindingOperation.Initialize
```
this.ProjectLanguage = ProjectToLanguageMapper.GetLanguageForProject(this.initializedProject);
```

PR #1390, opened 2020-05-06
Fix issue severity for CFamily

### LINKED ISSUE #1384: [CFamily] Issues should use the severity specified in the Quality Profile
### Description
See #808 [C#/VB] Issues should use the severity specified in the Quality Profile.

The CFamily analyzer has the same issue.

At present the severity is hard-coded to `Warning` [here](https://github.com/SonarSource/sonarlint-visualstudio/blob/558296b44d059b55830e22a0553c8613f038d4a0/src/Integration.Vsix/SonarLintTagger/IssuesSnapshot.cs#L90).

Parent issue: #884

### LINKED ISSUE #1386: [CFamily] Issue severity reported in standalone mode should not always be Warning
### Description
CFamily analyzer issues in standalone always return the severity "Warning" 

### Expected behaviour:
If the severity for the rule is specified in the user settings file then it should be respected.
Otherwise, the default severity for the rule should be used.

The same mapping from Sonar severity to VS severity described in #808 should be used, including the option to treat `Blocker` as `Error`.

### Notes
There are a number of related issues for connected mode, described in #884.

Part of the issue is the hard-coding [here](https://github.com/SonarSource/sonarlint-visualstudio/blob/558296b44d059b55830e22a0553c8613f038d4a0/src/Integration.Vsix/SonarLintTagger/IssuesSnapshot.cs#L90).
Removing this hard-coding is required to fix all of the related issues, but additional work is required in each case e.g. in standalone mode the severity is determined by user settings file and the rule defaults, whereas in the case of connected mode the settings from the quality profile should be used.

### Setting severity in the user settings file
See [this page](https://github.com/SonarSource/sonarlint-visualstudio/wiki/Choosing-which-C-or-Cpp-rules-to-run-in-Standalone-mode) in the wiki.

duncanp-sonar:
Fixes #1384 (connected mode)
Fixes #1386 (standalone mode)

### DIFF
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/IssuesSnapshotTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/IssuesSnapshotTests.cs
@@ -18,14 +18,18 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Generic;
 using FluentAssertions;
+using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Shell.TableManager;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.Text;
 using Moq;
 using Sonarlint;
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration.Vsix;
+using DaemonSeverity = Sonarlint.Issue.Types.Severity;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests
 {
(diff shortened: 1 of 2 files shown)


duncanp-sonar:
I was expecting to have to do more work to fix the severity correctly for standalone mode and connected mode, since in standalone mode the user settings should be taken in to account to override the defaults, and in connected mode the QP settings should be used.

However,  all of the necessary logic was already in place. The only issue was in the final translation to what appears in the Error List.

src/Integration.Vsix.UnitTests/SonarLintTagger/IssuesSnapshotTests.cs
```diff
@@ -158,6 +162,34 @@ public void ProjectName()
             GetValue(StandardTableKeyNames.ProjectName).Should().Be("MyProject");
         }
 
+        [TestMethod]
+        [DataRow(DaemonSeverity.Info, __VSERRORCATEGORY.EC_MESSAGE)]
+        [DataRow(DaemonSeverity.Minor, __VSERRORCATEGORY.EC_MESSAGE)]
+        [DataRow(DaemonSeverity.Major, __VSERRORCATEGORY.EC_WARNING)]
+        [DataRow(DaemonSeverity.Critical, __VSERRORCATEGORY.EC_WARNING)]
+        public void ToVsErrorCategory_NotBlocker_CorrectlyMapped(DaemonSeverity daemonSeverity, __VSERRORCATEGORY expectedVsErrorCategory)
```
duncanp-sonar, line 170: These tests are the same as those for the RuleSetGenerator, which the exception of the "invalid value" test.

src/Integration.Vsix.UnitTests/SonarLintTagger/IssuesSnapshotTests.cs
```diff
@@ -158,6 +162,34 @@ public void ProjectName()
...
+
+            var testSubject = new IssuesSnapshot("any", "any", 0, Array.Empty<IssueMarker>(), envSettingsMock.Object);
+
+            testSubject.ToVsErrorCategory(DaemonSeverity.Blocker).Should().Be(expectedVsErrorCategory);
+        }
+
+        [TestMethod]
+        public void ToVsErrorCategory_InvalidDaemonSeverity_DoesNotThrow()
+        {
+            snapshot.ToVsErrorCategory((DaemonSeverity)(-999)).Should().Be(__VSERRORCATEGORY.EC_MESSAGE);
```
duncanp-sonar, line 191: We don't want to throw on invalid values, for the reason given in the product code comment.

src/Integration.Vsix/SonarLintTagger/IssuesSnapshot.cs
```diff
@@ -125,6 +135,29 @@ public override bool TryGetValue(int index, string keyName, out object content)
             }
         }
 
+        internal /* for testing */ __VSERRORCATEGORY ToVsErrorCategory(Sonarlint.Issue.Types.Severity daemonSeverity)
```
duncanp-sonar, line 138: Logically the same as the code in `RuleSetGenerator`, except that source and target types are different.

duncanp-sonar:
I was expecting to have to do more work to fix the severity correctly for standalone mode and connected mode, since in standalone mode the user settings should be taken in to account to override the defaults, and in connected mode the QP settings should be used.

However,  all of the necessary logic was already in place. The only issue was in the final translation to what appears in the Error List.

PR #1391, opened 2020-05-06
Add bind functionality to ProjectBinder

rita-g-sonarsource:


src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -198,19 +194,14 @@ public void Prepare(CancellationToken token)
...
-                else
-                {
-                    logger.WriteLine(Strings.Bind_Project_NotRequired, project.FullName);
-                }
+
+                var languageForProject = ProjectToLanguageMapper.GetLanguageForProject(project);
+                var bindingConfigFile = GetBindingConfig(languageForProject);
+
+                var projectBinder = projectBinderFactory.Get(project);
+                var bindAction = projectBinder.GetBindAction(bindingConfigFile, project, token);
```
rita-g-sonarsource, line 202: GetBindAction will return the Commit() handle

src/Integration/Binding/IProjectBinder.cs
```diff
@@ -29,5 +31,6 @@ namespace SonarLint.VisualStudio.Integration.Binding
     internal interface IProjectBinder
     {
         bool IsBound(BindingConfiguration binding, Project project);
+        Action GetBindAction(IBindingConfigFile configFile, Project project, CancellationToken cancellationToken);
```
rita-g-sonarsource, line 34: I'm not a big fan of this method. It essentially returns a handle to ProjectBindingOperation.Commit.
At first I created PrepareBind and CommitBind methods, but that required to store the ProjectBindingOperation as a member of IProjectBinder and I was not a fan of that.
Let me know what you think.

src/Integration.UnitTests/Binding/SolutionBindingOperationTests.cs
```diff
@@ -268,85 +284,51 @@ public void SolutionBindingOperation_Prepare()
...
             testSubject.RuleSetsInformationMap[Language.CSharp].FilePath.Should().Be("c:\\csharp.txt");
             testSubject.RuleSetsInformationMap[Language.VBNET].FilePath.Should().Be("c:\\vb.txt");
-            prepareCalledForBinder.Should().BeFalse("Expected to be canceled as soon as possible i.e. after the first binder");
-
-            CheckSaveWasNotCalled(csConfigFile);
-            CheckSaveWasNotCalled(vbConfigFile);
-        }
-
-        [TestMethod]
-        public void SolutionBindingOperation_Prepare_Cancellation_BeforeBindersPrepare()
```
rita-g-sonarsource, line 316: It was difficult to test the two use-cases (beforeBindersPrepare and duringBindersPrepare), so I left only beforeBindersPrepare.

src/Integration/Binding/IProjectBinder.cs
duncanp-sonar: Yep, returning an `Action` feels wrong. It's like using `object` as a data type - it doesn't convey any semantic information. It would be better to have an interface/delegate to encapsulate the "commit binding" concept.

Probably the simplest thing to do at this point would be declare a delegate and use that as the return type instead e.g. `public delegate void BindProject();`
That would be simpler than adding a new interface with a single operation.

Essentially, at this point we want to get the top-down API right. We can hide existing implementation classes like `ProjectBindingOperation` behind then new API and refactor them later if necessary.

src/Integration.UnitTests/Binding/CFamilyProjectBinderTests.cs
```diff
@@ -0,0 +1,125 @@
+﻿using System;
```
duncanp-sonar, line 1: Missing copyright...

duncanp-sonar:
I haven't looked at the test yet. The product code looks ok, apart from returning an `Action`.

src/Integration.UnitTests/Binding/CFamilyProjectBinderTests.cs
rita-g-sonarsource: EVERY TIME. 😢 
We have to turn this on in sonarcloud...

src/Integration/Binding/IProjectBinder.cs
```diff
@@ -29,5 +30,8 @@ namespace SonarLint.VisualStudio.Integration.Binding
     internal interface IProjectBinder
     {
         bool IsBound(BindingConfiguration binding, Project project);
+        BindProject GetBindAction(IBindingConfigFile configFile, Project project, CancellationToken cancellationToken);
     }
+
+    public delegate void BindProject();
```
rita-g-sonarsource, line 36: Maybe `BindProjectAction`?

duncanp-sonar: I think BindProject is ok.

src/Integration.UnitTests/Binding/CFamilyProjectBinderTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        [DataRow(true)]
+        [DataRow(false)]
+        public void IsBound_ProjectHasOneLanguage(bool configFileExists)
+        {
+            var projectMock = new ProjectMock("c:\\test.csproj");
+            projectMock.SetCSProjectKind();
+
+            solutionRuleSetsInformationProviderMock
+                .Setup(x => 
+                    x.CalculateSolutionSonarQubeRuleSetFilePath("key", Language.CSharp, SonarLintMode.Connected))
```
duncanp-sonar, line 91: There's something wrong, either in the test or in the product code. The `CFamilyProjectBinder` should only be handling CFamily languages - it shouldn't be checking whether config files exist for non-CFamily languages.

It doesn't need to block this commit, but there's definitely more work to do on how "IsBound" is implemented i.e. a TODO that needs to be captured somewhere.

src/Integration.UnitTests/Binding/RoslynProjectBinderTests.cs
```diff
@@ -0,0 +1,213 @@
...
+
+        private string SetupRulesetPath()
+        {
+            var mockRulesetPath = "c:\\test.ruleset";
+
+            solutionRuleSetsInformationProviderMock
+                .Setup(x => x.CalculateSolutionSonarQubeRuleSetFilePath("key", Language.CSharp, SonarLintMode.Connected))
+                .Returns(mockRulesetPath);
+
+            return mockRulesetPath;
```
duncanp-sonar, line 194: There isn't a test for `IsBound` returning `true` (so you could simplifying the implementation to hard-code a return value of `false` and the tests would all still pass ;-) )

rita-g-sonarsource: I know, this is because the RuleSetIncludeChecker is a static class :( I wanted to minimize the scope of the commit

src/Integration.UnitTests/Binding/CFamilyProjectBinderTests.cs
rita-g-sonarsource: Ah, I see your point.
It returns back to the discussion we had this morning regarding the language per project handling: CFamilyProjectBinder doesn't know how many languages the project has, while in practice there would always be exactly 2 (cpp+c). So I wanted to write a test that mimics the fact that there is only one language per project.
But maybe I should remove this test entirely, 1) because it won't happen in reality, and 2) because it's confusing.

src/Integration.UnitTests/Binding/RoslynProjectBinderTests.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        private static BindingConfiguration GetBindingConfiguration()
+        {
+            var bindingConfiguration =
+                new BindingConfiguration(new BoundSonarQubeProject(new Uri("http://a.com"), "key", "name"),
+                    SonarLintMode.Connected, "dummy directory");
```
duncanp-sonar, line 201: It's not obvious which bits of data in the binding are meaningful and which are junk data that can be safely ignored when reading the tests. Using values like `dummy directory` helps, or `any` / `junk`.

As it happens, it looks like the value used for the project `key` does matter since the same value has to be used for in `SetupRulesPath`. The connection isn't obvious which makes the tests fragile.
The connection could be made more obvious by using a constant, or by passing the `BindingConfiguration` to `SetupRulesPath`.

src/Integration.UnitTests/Binding/RoslynProjectBinderTests.cs
duncanp-sonar: Ok; if it's deliberate then call it out in a review comment, and publicly record that it's still work to be done (e.g. any empty placeholder test that does nothing)

rita-g-sonarsource: Good idea 👍 will do

src/Integration.UnitTests/Binding/RoslynProjectBinderTests.cs
```diff
@@ -0,0 +1,213 @@
...
+            ruleSetSerializerMock.Setup(x => x.LoadRuleSet(mockRulesetPath)).Returns(new RuleSet("test"));
+
+            var ruleSetDeclaration = GetRuleSetDeclaration(projectMock);
+
+            solutionRuleSetsInformationProviderMock
+                .Setup(x => x.GetProjectRuleSetsDeclarations(projectMock))
+                .Returns(new List<RuleSetDeclaration> {ruleSetDeclaration});
+
+            var filePath = "";
+            solutionRuleSetsInformationProviderMock
```
duncanp-sonar, line 170: The `IsBoundXXX_False` tests are a linked set of tests that check a series of conditions in the order that they are called in the product code, which makes sense.
However, this means that there's a lot of duplication in the mock setup code, and also the mock setup is pretty ugly, which makes it hard to see the commonality and the differences. It would be easy to miss a copy-paste error.

The tests also mock a lot of things, then only check final return value. My immediate thought on reading such tests is "is the test returning `false` for the reason I think it is? Or is there a bug in the test setup, so I'm getting the right answer for the wrong reason?". I normally take one of two approaches. The first is to debug through each test when writing it to make sure it fails for the right reason. That tells me the tests are correct at the of writing, but doesn't help when the code changes in the future. The second is to make the test self-checking e.g. if the test expects to the first condition to fail, then assert that the second condition wasn't checked. It's more effort, but it gives more confidence that a test is passing for the right reason. The tests already implicitly depend on the order of operations in the product code; adding assertions just makes the dependency explicit.

src/Integration.UnitTests/Binding/SolutionBindingOperationTests.cs
duncanp-sonar: Couldn't you provide a mocked implementation of `projectBinder.GetBindAction` that cancels the operation? i.e. provide two binders, the first of which cancels the operation, then check the second wasn't called.

Or perhaps simpler - cancel the operation in  `sourceControlledFileSystem.QueueFileWrite`, then check the binders aren't called.

src/Integration.UnitTests/Binding/CFamilyProjectBinderTests.cs
duncanp-sonar: I think it's confusing because it's raising some design questions we haven't addressed e.g.
* what should a project binder do if it doesn't care about any of the languages in the project?
* should CFamily project binder know that _at present_ if a project supports `C` then it also supports `C++`?

Having a test is useful because it forces us to think about what the behaviour should be, and acts as record of those decisions.

It's up to you whether to remove this test for now, although I agree with your reason for adding it - I think the binder should handle the presence of only one language. The bit that is causing me a problem is that the fact that test is using a C# project!

We can add more tests later when we've decided what the behaviour should be.

PR #1392, opened 2020-05-06
Moved env var reads to IEnvironmentSettings

duncanp-sonar:


src/Core/EnvironmentSettings.cs
```diff
@@ -25,14 +25,36 @@ namespace SonarLint.VisualStudio.Core
     public class EnvironmentSettings : IEnvironmentSettings
     {
         internal const string TreatBlockerAsErrorEnvVar = "SONAR_INTERNAL_TREAT_BLOCKER_AS_ERROR";
+        internal const string CFamilyAnalysisTimeoutEnvVar = "SONAR_INTERNAL_CFAMILY_ANALYSIS_TIMEOUT_MS";
+        public const string SonarLintDownloadUrlEnvVar = "SONARLINT_DAEMON_DOWNLOAD_URL";
```
duncanp-sonar, line 29: The name of this environment variable is a in slightly different format to the INTERNAL ones, but it would be a breaking change to change it now.

src/Core/EnvironmentSettings.cs
```diff
@@ -25,14 +25,36 @@ namespace SonarLint.VisualStudio.Core
...
+            }
+
+            return 0;
+        }
+
+        public string SonarLintDaemonDownloadUrl()
+        {
+            // The URL validation and logging is being done by the daemon installer, so
+            // this is just a passthrough
+            return Environment.GetEnvironmentVariable(SonarLintDownloadUrlEnvVar);
```
duncanp-sonar, line 57: I wanted to move the URL validation here, but unfortunately that was complicated by the fact that the daemon installer is logging detailed messages about invalid URLs.
In the end I just left this method as a passthrough for now, so at least we have a level of abstraction that makes the daemon installer tests simpler.

src/Integration.Vsix.UnitTests/SonarLintDaemon/DaemonInstallerTests.cs
```diff
@@ -29,98 +31,77 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintDaemon
...
+        private Mock<IEnvironmentSettings> envSettingsMock = new Mock<IEnvironmentSettings>();
 
         public TestContext TestContext { get; set; }
 
         [TestInitialize]
         public void SetUp()
         {
             logger = new TestLogger(logToConsole: true);
-
-            tempPath = Path.Combine(TestContext.DeploymentDirectory, Path.GetRandomFileName());
```
duncanp-sonar, line 44: None of the deleted code was actually being used by any of the tests.

src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -266,17 +266,12 @@ private static bool ExecuteAnalysis(IProcessRunner runner, string fileName, stri
             return success;
         }
 
-        internal /* for testing*/ static int GetTimeoutInMs()
+        internal /* for testing*/ static int GetTimeoutInMs(IEnvironmentSettings environmentSettings = null)
```
duncanp-sonar, line 269: The `CFamilyHelper` is static which made injecting the mock `IEnvironmentSetting` tricky.
In the end I went for making the env setting optional, with a fall-back to the static singleton `SharedEnvSettings`.

The test `SubProcessTimeout_NoSettingsSupplied_NoError` is just that the code doesn't fail when the optional argument isn't supplied.

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -725,29 +725,22 @@ public void ConvertFromIssueType()
         }
 
         [TestMethod]
-        public void SubProcessTimeout()
+        [DataRow(-1, CFamilyHelper.DefaultAnalysisTimeoutMs)]
+        [DataRow(0, CFamilyHelper.DefaultAnalysisTimeoutMs)]
+        [DataRow(1, 1)]
+        [DataRow(999, 999)]
+        public void SubProcessTimeout(int envSettingsResponse, int expectedTimeout)
```
rita-g-sonarsource, line 732: 🚀

PR #1393, opened 2020-05-07
Record legacy connected mode telemetry

### LINKED ISSUE #1381: Collect telemetry about usage of legacy connected mode
### Description

The "new" connected mode experienced was introduced two years ago, although we still support solutions using the "legacy" connected mode. However, we have no idea how many users are still using legacy connected mode.

We would like to update the telemetry to include this information to allow us to make better decisions about how and when to deprecated/drop legacy connected mode features.

### Implementation
The parameter is: `legacy_connected_mode_used`.
The value is a boolean: `true`/`false`.

#### Note (1)
there is telemetry data item called `connected_mode_used` that records whether either type of connected mode is used. This means that for "new" connected mode we expect the following values:
```
connected_mode_used = true
legacy_connected_mode_used = false
```
... and for legacy connected mode we expect the values
```
connected_mode_used = true
legacy_connected_mode_used = true
```

#### Note (2)
The telemetry data about whether connected mode is being used is captured at the point the data is sent i.e. based entirely on the solution that is open at the time. So we could receive telemetry reports from the same user saying they are using standalone mode, connected mode or legacy connected mode, entirely dependent on which solution they have open at the time.
This doesn't seem very useful. but it is what is specified in the original MMF.

TODO - link to MMF

duncanp-sonar:
Fixes #1381


src/Integration/Telemetry/TelemetryPayload.cs
```diff
@@ -49,6 +49,9 @@ public sealed class TelemetryPayload
         [JsonProperty("connected_mode_used")]
         public bool IsUsingConnectedMode { get; set; }
 
+        [JsonProperty("legacy_connected_mode_used")]
+        public bool IsUsingLegacyConnectedMode { get; set; }
```
duncanp-sonar, line 53: I agree with the PM that this is what he wants for now. We won't be able to tell if someone is using both legacy connected mode and new connected mode, but that's acceptable for now.

PR #1395, opened 2020-05-10
Remove "FilePath" property from IBindingConfig

rita-g-sonarsource:
I suggest to look at the commits separately. This is additional cleanup as part of prep for additional files: the property `FilePath` was on the `IBindingConfig` interface so that `SolutionBindingOperation` could use it to:
1. Queue the file write using `ISourceControlledFileSystem`
2. Add the file to solution folder when in legacy mode

I moved the queuing functionality to `IBindingConfig` implementation.
I also removed the logic of adding it to solution folder when in legacy mode -- this was more due to limitation: I don't know the binding mode when in `IBindingConfig`, so I don't know if it's legacy or not. So the options were either to pass the binding mode, or to remove the functionality. Since we discussed that the functionality is mostly for cosmetics, and since we are considering dropping legacy mode, I decided not to support it -- so please let me know if I should.


src/Core/Binding/IBindingConfigProvider.cs
```diff
@@ -44,11 +44,9 @@ public interface IBindingConfigProvider
     /// For C++ it will be in a json file in a Sonar-specific format</remarks>
     public interface IBindingConfig
     {
-        string FilePath { get; }
-
         /// <summary>
         /// Saves the file, replacing any existing file
         /// </summary>
-        void Save();
+        void Save(ISourceControlledFileSystem sourceControlledFileSystem);
```
rita-g-sonarsource, line 50: I was debating if I should pass this as ctor parameter, and I think I should. It would require to pass it down from `BindingController`, but I think it would make for a cleaner API.

duncanp-sonar: I'm not convinced. Let's discuss it offline.

src/Integration/Binding/CSharpVBBindingConfig.cs
```diff
@@ -40,9 +52,17 @@ public CSharpVBBindingConfig(RuleSet ruleSet, string filePath)
             FilePath = filePath;
         }
 
-        public void Save()
+        public void Save(ISourceControlledFileSystem sourceControlledFileSystem)
         {
-            RuleSet.WriteToFile(FilePath);
+            sourceControlledFileSystem.QueueFileWrite(FilePath, () =>
```
duncanp-sonar, line 56: This now looks weird: the binder now has to know to queue the file operation, rather than just executed it directly. Also, it isn't clear what the relationship between the `sourceControlledFileSystem` and the `fileSystem` is, if any. I think there's a design issue here we need to clear up.

PR #1396, opened 2020-05-10
Change IsBound to IsBindingRequired

rita-g-sonarsource:


src/Integration.UnitTests/Binding/CFamilyProjectBinderTests.cs
```diff
@@ -79,31 +79,23 @@ public void Ctor_NullFileSystem_ArgumentNullException()
...
+            var result = testSubject.IsBindingRequired(bindingConfiguration, projectMock);
+            result.Should().Be(false);
+
+            solutionRuleSetsInformationProviderMock.VerifyNoOtherCalls();
+            fileSystemMock.VerifyNoOtherCalls();
         }
 
         [TestMethod]
-        public void IsBound_ProjectHasTwoLanguages_OneLanguageHasNoConfigFile_False()
+        public void IsBindingRequired_ProjectHasTwoLanguages_OneLanguageHasNoConfigFile_True()
```
duncanp-sonar, line 98: It should be possible to write the missing "ProjectHasTwoLanguages_AllConfigFilesExists_False" test now, shouldn't it?

src/Integration.UnitTests/Binding/CSharpVBProjectBinderTests.cs
```diff
@@ -175,14 +191,14 @@ public void IsBound_ProjectHasOneRuleset_CantLoadProjectRulesetFile_False()
                 .Setup(x => x.TryGetProjectRuleSetFilePath(projectMock, ruleSetDeclaration, out filePath))
                 .Returns(false);
 
-            var result = testSubject.IsBound(bindingConfiguration, projectMock);
-            result.Should().BeFalse();
+            var result = testSubject.IsBindingRequired(bindingConfiguration, projectMock);
+            result.Should().BeTrue();
```
duncanp-sonar, line 195: Perhaps verify the expected mocks were called, similar to what you'd done in the earlier tests?

src/Integration/Binding/IProjectBinder.cs
```diff
@@ -29,7 +29,7 @@ namespace SonarLint.VisualStudio.Integration.Binding
     /// </summary>
     internal interface IProjectBinder
     {
-        bool IsBound(BindingConfiguration binding, Project project);
+        bool IsBindingRequired(BindingConfiguration binding, Project project);
```
duncanp-sonar, line 32: This deserves a doc comment to make it clear that the method will return `false` if the project isn't supported by the binder or if the project is already bound correctly i.e. it will only return `true` for unbound/partially-bound projects.

PR #1398, opened 2020-05-11
Fix bug of first time binding

rita-g-sonarsource:
I found a nasty bug during QA: I used the existing binding information in order to fetch the configurations, which caused a bug when there was no existing binding information (first time bind).
Hence I had to move the code that calculates the new binding configuration, and use that new configuration.

### DIFF
--- a/src/Integration/Binding/BindingController.cs
+++ b/src/Integration/Binding/BindingController.cs
@@ -158,24 +158,20 @@ namespace SonarLint.VisualStudio.Integration.Binding
 
             var solutionBindingOp = new SolutionBindingOperation(
                 host,
-                bindingArgs.Connection,
-                bindingArgs.ProjectKey,
-                bindingArgs.ProjectName,
                 modeToBind,
                 host.Logger);
 
             var unboundProjectFinder = new UnboundProjectFinder(host, host.Logger);
 
-            var cSharpVBBindingConfigProvider = new CSharpVBBindingConfigProvider(host.SonarQubeService, nugetBindingOp, currentConfiguration, host.Logger);
-            
-            var cppConfigProvider = new CFamilyBindingConfigProvider(host.SonarQubeService, currentConfiguration, host.Logger);
+            var cSharpVBBindingConfigProvider = new CSharpVBBindingConfigProvider(host.SonarQubeService, nugetBindingOp, host.Logger);
+            var cppConfigProvider = new CFamilyBindingConfigProvider(host.SonarQubeService, host.Logger);
+           
             var ruleConfigProvider = new CompositeBindingConfigProvider(cSharpVBBindingConfigProvider, cppConfigProvider);
-
-            var bindingProcess = new BindingProcessImpl(host, bindingArgs, solutionBindingOp, nugetBindingOp, unboundProjectFinder, ruleConfigProvider, isFirstBinding);
+            var bindingProcess = new BindingProcessImpl(host, bindingArgs, solutionBindingOp, nugetBindingOp, unboundProjectFinder, ruleConfigProvider, modeToBind, isFirstBinding);
 
             return bindingProcess;
         }
-
+        
         internal /*for testing purposes*/ void SetBindingInProgress(IProgressEvents progressEvents, BindCommandArgs bindingArgs)
         {
             this.OnBindingStarted();
--- a/src/Integration/Binding/BindingProcessImpl.cs
+++ b/src/Integration/Binding/BindingProcessImpl.cs
@@ -27,6 +27,8 @@ using System.Threading;
 using System.Threading.Tasks;
 using EnvDTE;
 using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Integration.NewConnectedMode;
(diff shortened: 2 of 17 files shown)


src/Integration/Binding/BindingController.cs
```diff
@@ -158,24 +158,20 @@ void IBindingWorkflowExecutor.BindProject(BindCommandArgs bindingArgs)
...
                 host,
-                bindingArgs.Connection,
-                bindingArgs.ProjectKey,
-                bindingArgs.ProjectName,
                 modeToBind,
                 host.Logger);
 
             var unboundProjectFinder = new UnboundProjectFinder(host, host.Logger);
 
-            var cSharpVBBindingConfigProvider = new CSharpVBBindingConfigProvider(host.SonarQubeService, nugetBindingOp, currentConfiguration, host.Logger);
```
rita-g-sonarsource, line 169: This was an oversight on my part -- all the calculation should be done based on _new configuration_, and not on the current one.

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -122,8 +127,10 @@ public async Task<bool> DownloadQualityProfileAsync(IProgress<FixedStepsProgress
                 }
                 this.InternalState.QualityProfiles[language] = qualityProfileInfo;
 
+                var bindingConfiguration = WriteBindingInformation();
```
rita-g-sonarsource, line 130: The location of this calculation sucks. Logically it needs to be done after fetching the quality profile, and before calling GetConfigurationAsync.

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -143,6 +150,40 @@ public async Task<bool> DownloadQualityProfileAsync(IProgress<FixedStepsProgress
             return true;
         }
 
+        /// <summary>
+        /// Will add/edit the binding information for next time usage
+        /// </summary>
+        private BindingConfiguration WriteBindingInformation()
```
rita-g-sonarsource, line 156: Cut-paste of the logic that was inside SolutionBindingOperation.

src/Integration/Binding/ISolutionBindingOperation.cs
```diff
@@ -36,7 +34,7 @@ namespace SonarLint.VisualStudio.Integration.Binding
     /// </remarks>
     public interface ISolutionBindingOperation : ISolutionBindingConfigFileStore
     {
-        void Initialize(IEnumerable<Project> projects, IDictionary<Language, SonarQubeQualityProfile> profilesMap);
```
rita-g-sonarsource, line 39: No longer needed, as it was only for the calculation of the BindingConfiguration.

src/Integration/NewConnectedMode/ConfigurationProvider.cs
```diff
@@ -70,8 +70,10 @@ public bool Persist(BoundSonarQubeProject project, SonarLintMode bindingMode)
 
             var writeSettings = GetWriteSettings(bindingMode);
 
-            return writeSettings.HasValue && 
+            var success = writeSettings.HasValue && 
                    solutionBindingSerializer.Write(writeSettings?.ConfigPath, project, writeSettings?.OnSuccessfulFileWrite);
+
+            return success ? TryGetBindingConfiguration(writeSettings?.ConfigPath, project, bindingMode) : null;
```
rita-g-sonarsource, line 76: solutionBindingSerializer.Write only queues the writing operation, so I couldn't call GetConfiguration. I had to instead use the created `project` and initialize a new BindingConfiguration object.

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -143,6 +150,40 @@ public async Task<bool> DownloadQualityProfileAsync(IProgress<FixedStepsProgress
             return true;
         }
 
+        /// <summary>
```
duncanp-sonar, line 153: Please call out in a review comment when you've moved a chunk of code from one class to another and whether you've made any changes to it - it's hard to tell in the GitHub web UI.

src/Integration/NewConnectedMode/ConfigurationProvider.cs
```diff
@@ -82,7 +84,12 @@ private BindingConfiguration TryGetBindingConfiguration(string bindingPath, Sona
             }
 
             var boundProject = solutionBindingSerializer.Read(bindingPath);
+            
+            return TryGetBindingConfiguration(bindingPath, boundProject, sonarLintMode);
+        }
 
+        private BindingConfiguration TryGetBindingConfiguration(string bindingPath, BoundSonarQubeProject boundProject, SonarLintMode sonarLintMode)
```
duncanp-sonar, line 91: Rename to `Create`

src/Integration/NewConnectedMode/ConfigurationProvider.cs
duncanp-sonar: As discussed offline, I think that's ok. We need to create a new instance somewhere.

PR #1399, opened 2020-05-11
Support multiple files

rita-g-sonarsource:


src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -144,21 +144,25 @@ public void Prepare(CancellationToken token)
...
+                    {
+                        var ruleSetDirectoryPath = Path.GetDirectoryName(solutionItem);
 
-                    // Create or overwrite existing rule set
-                    info.Save();
+                        fileSystem.Directory.CreateDirectory(ruleSetDirectoryPath); // will no-op if exists
 
-                    return true;
-                });
+                        info.Save();
```
rita-g-sonarsource, line 158: This is really odd api :/
We will re-save all the files X number of files.

duncanp-sonar: Can the `sourceControlledFileSystem` be changed easily to support `QueueFileWrites(IEnumerable<string>)` ?

src/Core/Binding/IBindingConfigProvider.cs
```diff
@@ -44,7 +45,7 @@ public interface IBindingConfigProvider
     /// For C++ it will be in a json file in a Sonar-specific format</remarks>
     public interface IBindingConfig
     {
-        string FilePath { get; }
+        IList<string> SolutionItems { get; }
```
duncanp-sonar, line 48: `SolutionItems` is not a very descriptive name - it already has a specific meaning in VS, and it doesn't convey any information about the content of the strings. Something like `SolutionLevelFilePaths` would be better.

A doc comment would be useful: it's a list of all of the solution-level files that should exist if the solution is correctly bound. It doesn't say whether the files exist or not. It would also be useful to state whether the value can be null.

duncanp-sonar, line 48: Does it need to be a list i.e. is the order important?

src/Integration.UnitTests/LocalServices/SourceControlledFileSystemTests.cs
```diff
@@ -86,6 +88,45 @@ public void SourceControlledFileSystem_FileExistOrQueuedToBeWritten()
             testSubject.FileExistOrQueuedToBeWritten(file).Should().BeFalse();
         }
 
+        [TestMethod]
+        [DataRow(true, true, true)]
+        [DataRow(true, false, false)]
+        [DataRow(false, true, false)]
+        [DataRow(false, false, false)]
+        public void SourceControlledFileSystem_FilesExistOrQueuedToBeWritten_ReturnsIfAllFilesAreQueuedOrWritten(bool isFirstFileWritten, bool isSecondFileWritten, bool expectedResult)
```
duncanp-sonar, line 96: Naming suggestion: `firstFileExists`/`secondFileExists`

src/Integration/LocalServices/ISourceControlledFileSystem.cs
```diff
@@ -35,11 +36,21 @@ internal interface ISourceControlledFileSystem : ILocalService
         /// <param name="fileWriteOperation">The actual write operation, return false if failed</param>
         void QueueFileWrite(string filePath, Func<bool> fileWriteOperation);
 
+        /// <summary>
+        /// Queues a write operation for given file paths. New or edit is determined internally at the time of this method execution.
```
duncanp-sonar, line 40: Not clear what "determined internally" means. Does it just means "evaluated"?

src/Integration/LocalServices/ISourceControlledFileSystem.cs
```diff
@@ -35,11 +36,21 @@ internal interface ISourceControlledFileSystem : ILocalService
...
+        /// </summary>
+        void QueueFileWrites(IList<string> filePaths, Func<bool> fileWriteOperation);
```
duncanp-sonar, line 42: Miscellaneous note: we should rename this interface at some point - it's misleading to call it an abstraction over the file system. It would more accurate to call it a "SourceControlOperationQueue"

src/Integration/LocalServices/ISourceControlledFileSystem.cs
```diff
@@ -35,11 +36,21 @@ internal interface ISourceControlledFileSystem : ILocalService
         /// <param name="fileWriteOperation">The actual write operation, return false if failed</param>
         void QueueFileWrite(string filePath, Func<bool> fileWriteOperation);
 
+        /// <summary>
```
duncanp-sonar, line 39: Are both `QueueFileWrite` and `QueueFileWrite` required?

src/Integration/LocalServices/ISourceControlledFileSystem.cs
```diff
@@ -35,11 +36,21 @@ internal interface ISourceControlledFileSystem : ILocalService
...
         /// <summary>
         /// Returns whether the file exists on disk or in a queue to be written. <seealso cref="QueueFileWrite(string, Func{bool})"/>
         /// </summary>
         bool FileExistOrQueuedToBeWritten(string filePath);
 
+        /// <summary>
+        /// Returns whether all given files exist on disk or in a queue to be written. <seealso cref="QueueFileWrites(IList{string}, Func{bool})"/>
+        /// </summary>
+        bool FilesExistOrQueuedToBeWritten(IList<string> filePaths);
+
```
duncanp-sonar, line 53: Are both `FileExist...` and `FilesExist...` required? Same for `QueueFileWrite` and `QueueFileWrite`.

src/Core/Binding/IBindingConfigProvider.cs
rita-g-sonarsource: The order is not important, no. I'll change to IEnumerable.

src/Integration/LocalServices/ISourceControlledFileSystem.cs
rita-g-sonarsource: Ah, it was a copy-paste of QueueFileWrite doc comment. I changed both to "evaluated"

src/Core/Binding/IBindingConfigProvider.cs
```diff
@@ -45,7 +45,10 @@ public interface IBindingConfigProvider
     /// For C++ it will be in a json file in a Sonar-specific format</remarks>
     public interface IBindingConfig
     {
-        IList<string> SolutionItems { get; }
+        /// <summary>
+        /// Returns a list of all of the solution-level files that should exist if the solution is correctly bound. The files might not exist on disk. List cannot be null or empty.
+        /// </summary>
```
duncanp-sonar, line 50: I'd have thought the list would be empty if the provider doesn't handle the project...

rita-g-sonarsource: Not the same interface, this is the class that contains the actual data (ruleset + additional files).
We should rename it, maybe SolutionBindingData?

src/Integration/LocalServices/ISourceControlledFileSystem.cs
```diff
@@ -30,31 +30,40 @@ namespace SonarLint.VisualStudio.Integration
...
         /// <summary>
-        /// Queues a write operation for given file paths. New or edit is determined internally at the time of this method execution.
+        /// Returns whether all given files exist on disk or in a queue to be written. <seealso cref="QueueFileWrites(IEnumerable{string}, Func{bool})"/>
         /// </summary>
-        void QueueFileWrites(IList<string> filePaths, Func<bool> fileWriteOperation);
+        bool FilesExistOrQueuedToBeWritten(IEnumerable<string> filePaths);
 
         /// <summary>
-        /// Returns whether the file exists on disk or in a queue to be written. <seealso cref="QueueFileWrite(string, Func{bool})"/>
+        /// Checks out the file if the solution is under source control, and then executes all the queued file writes.  <seealso cref="QueueFileWrite(string, Func{bool})"/>
```
duncanp-sonar, line 43: The `cref` should be to `cref="QueueFileWrites(IEnumerable{string}, Func{bool})"` (although I'm not sure they are really much use in practice).

src/Core/Binding/IBindingConfigProvider.cs
duncanp-sonar: No, let's leave it for now. It still has a `Save` method, which you wouldn't expect on a `data` class.

PR #1402, opened 2020-05-12
Add AdditionalFile property

rita-g-sonarsource:


src/Integration/Binding/CSharpVBBindingConfigProvider.cs
```diff
@@ -119,7 +119,15 @@ private async Task<IBindingConfig> DoGetConfigurationAsync(SonarQubeQualityProfi
...
                 language.FileSuffixAndExtension);
 
-            return new CSharpVBBindingConfig(ToVsRuleset(coreRuleset), ruleSetFilePath);
+            var additionalFilePath = solutionBindingFilePathGenerator.Generate(
+                bindingConfiguration.BindingConfigDirectory,
+                bindingConfiguration.Project.ProjectKey,
+                "params_" + language.FileSuffixAndExtension);
+
+            var ruleset = new FilePathAndContent<VsRuleset>(ruleSetFilePath, ToVsRuleset(coreRuleset));
+            var additionalFile = new FilePathAndContent<string>(additionalFilePath, "todo");
```
rita-g-sonarsource, line 128: @duncanp-sonar which API should I call to get the additional file content?
It used to be GetRoslynExportProfileAsync, which returned ConfigurationResponse, that had AdditionalFileResponse.

duncanp-sonar:
```
var config = SonarLintConfigGenerator.Generate(activeRules, sonarProperties, language.ServerLanguage.Key);
```

... although the class is currently `internal` and `static` in `Core`, so it will need an interface for testing.

PR #1405, opened 2020-05-13
Support Additional Files

### LINKED ISSUE #176: Additional files for parameterized SonarQube rules are retrieved from the SonarQube quality profile

rita-g-sonarsource:
Fixes #176


src/Integration/Binding/CSharpVBBindingConfigProvider.cs
```diff
@@ -112,14 +114,40 @@ private async Task<IBindingConfig> DoGetConfigurationAsync(SonarQubeQualityProfi
...
+        }
+
+        private FilePathAndContent<SonarLintConfiguration> GetAdditionalFile(Language language, BindingConfiguration bindingConfiguration, IEnumerable<SonarQubeRule> activeRules, Dictionary<string, string> sonarProperties)
+        {
+            var additionalFilePathDirectory = solutionBindingFilePathGenerator.Generate(
+                bindingConfiguration.BindingConfigDirectory,
+                bindingConfiguration.Project.ProjectKey,
+                string.Empty);
+
+            var additionalFilePath = Path.Combine(additionalFilePathDirectory, language.Id, "SonarLint.xml");
```
rita-g-sonarsource, line 144: I'm not a fan of this split: the reason for this is that solutionBindingFilePathGenerator uses PathHelper.EscapeFileName, which replaces \ to _, and we need the file name to be precisely SonarLint.xml, and not csharp_sonarlint.xml.

src/Integration.UnitTests/Binding/CSharpVBBindingOperation.WriterTests.cs
```diff
@@ -196,16 +200,25 @@ public void ProjectBindingOperation_QueueWriteProjectLevelRuleSet_ProjectHasExis
...
             // Act
-            string actualPath = testSubject.QueueWriteProjectLevelRuleSet(projectFullPath, ruleSetName, dotNetConfig, existingProjectRuleSetPath);
+            string actualPath = testSubject.QueueWriteProjectLevelRuleSet(projectFullPath, ruleSetName, csharpVbConfig, existingProjectRuleSetPath);
 
             // Assert
             this.ruleSetFS.AssertRuleSetsAreEqual(actualPath, expectedRuleSet);
             actualPath.Should().Be(existingProjectRuleSetPath, "Expecting the rule set to be updated");
         }
 
+        private static CSharpVBBindingConfig CreateCSharpVbBindingConfig(string newSolutionRuleSetPath, RuleSet expectedRuleSet)
```
rita-g-sonarsource, line 214: extracted to function as it's now more complicated to do "new CSharpVBBindingConfig"

src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -87,25 +94,55 @@ public bool IsBindingRequired(BindingConfiguration binding, Project project)
...
+            {
+                return false;
+            }
+
+            solutionRuleset = GetSolutionRuleset(binding, language);
+
+            return solutionRuleset != null;
+        }
+
+        private string GetSolutionAdditionalFile(BindingConfiguration binding, Language language)
```
rita-g-sonarsource, line 131: Exact duplication of the logic inside `CSharpVBBindingConfigProvider`.
Same goes for the ruleset filepath, that is also being duplicated here.

The reason for the duplication at the moment is because "IsBindingRequired" is called by `IUnboundProjectFinder`, which is called by `ErrorListInfoBarController`, at a stage where we didn't call yet `CSharpVBBindingConfigProvider`, so we don't have the `CSharpVBBindingConfig` with the file paths.
It's worth taking a look at `ErrorListInfoBarController` and see what we can do there -- if we can pass `CSharpVBBindingConfig` as part of IsBindingRequired, it would remove the duplication.

src/Integration.UnitTests/LocalServices/UnboundProjectFinderTests.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                var project = new ProjectMock("any.proj");
+                project.ProjectKind = projectKind;
+                projects.Add(project);
+                return project;
+            }
+
+            public void SetupProjectBindingRequired(EnvDTE.Project project, bool isBindingRequired)
+            {
+                var projectBinderMock = new Mock<IProjectBinder>();
```
rita-g-sonarsource, line 111: I re-wrote the tests to use a mock ProjectBinder, so they are now much simpler.

src/Integration.UnitTests/LocalServices/UnboundProjectFinderTests_ProjectLevelConfig.cs
```diff
@@ -1,348 +0,0 @@
...
-            action.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("serviceProvider");
-
-            action = () => new UnboundProjectFinder(serviceProvider, projectBinderFactory: null);
-            action.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("projectBinderFactory");
-        }
-
-        [TestMethod]
-        [DataRow(SonarLintMode.Connected)]
-        [DataRow(SonarLintMode.LegacyConnected)]
-        public void GetUnboundProjects_ValidSolution_ProjectRuleSetIsMissing(SonarLintMode mode)
```
rita-g-sonarsource, line 88: Deleted, the tests now use mocks and all the relevant testing is part of CSharpVBProjectBinderTests

src/Integration.UnitTests/LocalServices/UnboundProjectFinderTests_SolutionConfigLevel.cs
```diff
@@ -1,218 +0,0 @@
...
-            var result = testSubject.GetUnboundProjects();
-
-            // Assert
-            AssertEmptyResult(result);
-        }
-
-        [TestMethod]
-        [DataRow(SonarLintMode.Connected)]
-        [DataRow(SonarLintMode.LegacyConnected)]
-        public void GetUnboundProjects_ValidSolution_SolutionRuleConfigIsMissing(SonarLintMode mode)
```
rita-g-sonarsource, line 62: Deleted, the tests now use mocks and all the relevant testing is part of CSharpVBProjectBinderTests and CFamilyProjectBinderTests.

src/Integration/Binding/CSharpVBBindingConfigProvider.cs
duncanp-sonar: As previously discussed, I think the language-specific config providers should be entirely responsible for calculating their own file paths. They should be handed the base directory to use, and left to work everything else out for themselves.
Here, the base directory is in `bindingConfiguration`. You don't actually need the `solutionBindingFilePathGenerator` at all.
Just build the whole path yourself. If you need call the PathHelper directly, do so. Or add an extension method to `BindingConfiguration` that returns the `EscapedProjectName` in a format that can be used in a  filename.

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -119,6 +125,21 @@ public void Commit()
...
+            try
+            {
+                var projectSystem = serviceProvider.GetService<IProjectSystemHelper>();
+                projectSystem.AssertLocalServiceIsNotNull();
+                projectSystem.RemoveFileFromProject(initializedProject, cSharpVBBindingConfig.AdditionalFile.Path);
+                projectSystem.AddFileToProject(initializedProject, cSharpVBBindingConfig.AdditionalFile.Path, "AdditionalFiles");
+            }
+            catch (Exception ex)
+            {
+                throw new Exception($"Failed to bind project {initializedProject.FullName}, exception details: {ex}");
```
duncanp-sonar, line 139: Throw Exception directly is a code smell, but I think we do it in a few places.

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -119,6 +125,21 @@ public void Commit()
             }
         }
 
+        private void AddAdditionalFile()
+        {
+            try
+            {
+                var projectSystem = serviceProvider.GetService<IProjectSystemHelper>();
+                projectSystem.AssertLocalServiceIsNotNull();
+                projectSystem.RemoveFileFromProject(initializedProject, cSharpVBBindingConfig.AdditionalFile.Path);
```
duncanp-sonar, line 134: This needs some explanation.

src/Integration/Binding/CSharpVBProjectBinder.cs
duncanp-sonar: i.e. the refactoring isn't complete yet.
Where/how is this infomration recorded? (needs to be recorded more permanently than in a review comment).

src/Core.UnitTests/Binding/FilePathAndContentTests.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("path");
+        }
+
+        [TestMethod]
+        public void Ctor_NullContent_ArgumentNullException()
+        {
+            Action act = () => new FilePathAndContent<string>("filepath", null);
+
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("content");
```
duncanp-sonar, line 46: I think these tests are generally pretty low value - they're practically junk tests, just to increase coverage.

Still, since we write them, there are a couple of other tests that would be useful:
* a test that checks that the content can be empty, and
* a test that checks the properties are actually set.

src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
```diff
@@ -152,7 +151,7 @@ public async Task GetConfig_NoSupportedActiveRules_ReturnsNull()
         }
 
         [TestMethod]
-        public async Task GetConfig_ReturnsCorrectFilePath()
+        public void GetConfig_ReturnsCorrectRuleset()
```
duncanp-sonar, line 154: Presumably these could be actually be `async`, if the issue before was a merge conflict?

src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
```diff
@@ -396,11 +424,23 @@ public CSharpVBBindingConfigProvider CreateTestSubject()
...
 
+                solutionBindingFilePathGeneratorMock
+                    .Setup(x => x.Generate(bindingRootFolder, ExpectedProjectKey, string.Empty))
+                    .Returns(AdditionalFilePathResponse);
+
+                var sonarProperties = PropertiesResponse.ToDictionary(p => p.Key, y => y.Value);
+
+                sonarLintConfigGeneratorMock = new Mock<ISonarLintConfigGenerator>();
+                sonarLintConfigGeneratorMock
+                    .Setup(x => x.Generate(It.IsAny<IEnumerable<SonarQubeRule>>(), sonarProperties, language))
```
duncanp-sonar, line 435: Is generating the `sonarProperties` dictionary here providing a stronger check than using `It.IsAny...`?

i.e. is Moq actually checking that the contents of the properties of the dictionaries are equivalent?

src/Integration/Binding/CSharpVBProjectBinder.cs
duncanp-sonar: As suggested above, the `solutionBindingFilePathGenerator` isn't required.

A less-ugly alternative to duplicating the code would be to make the path calculation methods `internal static` on the CSharpVB config provider and call them from here.
That way, at least the connection between the two bits of code is obvious.

src/Integration/Binding/CSharpVBBindingOperation.cs
rita-g-sonarsource: There are two uses cases here:
1. It's possible that the additional file is set property, but the ruleset is out of date, which will trigger the binding. So to avoid receiving a "file already exists" -- I remove the file first.
2. It's possible that the additional file exists but has a wrong type, e.g. Compile/Content, and not AdditionalFiles. In which case removing and re-adding it solves the problem.

I will add unit tests to cover those 2 cases.

src/Core.UnitTests/Binding/FilePathAndContentTests.cs
rita-g-sonarsource: what do you mean by empty? e.g. string empty? (since the class is generic)

src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
rita-g-sonarsource: yep, I can revert it back

src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
rita-g-sonarsource: Moq will fail unless you use It.IsAny or pass the exact dictionary it expects. So I think it makes more sense to pass sonarProperties that Any.

src/Core.UnitTests/Binding/FilePathAndContentTests.cs
duncanp-sonar: Ok, good point.

src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
duncanp-sonar: It isn't the "exact dictionary" - it's a different object with equivalent contents, hence the question
> is Moq actually checking that the contents of the properties of the dictionaries are equivalent?
And apparently it is.

src/Integration/Binding/CSharpVBBindingOperation.cs
duncanp-sonar: The explanation needs to be in the code.

src/Integration.UnitTests/Binding/CSharpVbBindingConfigTests.cs
```diff
@@ -36,41 +38,60 @@ public class CSharpVbBindingConfigTests
...
 
             // Assert
-            File.Exists(fullPath).Should().BeTrue();
+            File.Exists(rulesetFullPath).Should().BeTrue();
+            File.Exists(additionalFileFullPath).Should().BeTrue();
+
+            var savedAdditionalFile = File.ReadAllText(additionalFileFullPath);
+            savedAdditionalFile.Should().Be(@"{
+  ""Settings"": [],
+  ""Rules"": []
```
duncanp-sonar, line 93: I didn't notice this when looking at the product code, but the `SonarLint.xml` file should be saved as XML rather than json...

It makes me wonder if there's anything else I've missed.

src/Integration/Binding/CSharpVBBindingConfig.cs
```diff
@@ -20,31 +20,42 @@
...
+            AdditionalFile = additionalFile ?? throw new ArgumentNullException(nameof(additionalFile));
+            this.fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
         }
 
         public void Save()
         {
-            RuleSet.WriteToFile(FilePath);
+            RuleSet.Content.WriteToFile(RuleSet.Path);
+
+            var serializedContent = Serializer.ToString(AdditionalFile.Content);
```
rita-g-sonarsource, line 57: @duncanp-sonar fixed saving as xml

src/Integration.UnitTests/Binding/CSharpVbBindingConfigTests.cs
rita-g-sonarsource: I pushed a fix 👍

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -119,6 +125,23 @@ public void Commit()
...
 
+        private void AddAdditionalFile()
+        {
+            try
+            {
+                var projectSystem = serviceProvider.GetService<IProjectSystemHelper>();
+                projectSystem.AssertLocalServiceIsNotNull();
+                // It's possible that the additional file is set property, but the ruleset is out of date, which will trigger the binding. So we need to remove the file first, otherwise we get an error that the file already exists.
+                // It's possible that the additional file exists but has a wrong type, e.g. Compile/Content, and not AdditionalFiles. In which case removing and re-adding it solves the problem.
+                projectSystem.RemoveFileFromProject(initializedProject, cSharpVBBindingConfig.AdditionalFile.Path);
```
duncanp-sonar, line 136: Thinking about it again, this might causes a problem in practice. What will happen if the additional file is referenced via an imported targets file?
Even if it isn't, removing and re-adding the same reference might change where the line appears in the project file. This is less serious, but it's still irritating for end users to have unnecessary changes appearing when they diff the project file.

I'll see what the behaviour is when testing - it might be something we'll fix later.

PR #1406, opened 2020-05-13
Extract ItemGroup functionality to IProjectSystemHelper

rita-g-sonarsource:
The functionality in ServiceGuidTestProjectIndicator is needed in order to check if AdditionalFiles property has been set, so I extracted it to a generic function.

### DIFF
--- a/src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
+++ b/src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
@@ -19,12 +19,14 @@
  */
 
 using System;
+using System.IO.Abstractions;
 using System.Linq;
 using EnvDTE;
 using FluentAssertions;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests
 {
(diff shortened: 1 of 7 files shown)


src/Integration/LocalServices/TestProjectIndicators/ServiceGuidTestProjectIndicator.cs
```diff
@@ -64,17 +53,12 @@ internal ServiceGuidTestProjectIndicator(ILogger logger, IFileSystem fileSystem)
...
-        private bool HasTestGuid(string projectPath)
+        private bool HasTestGuid(Project project)
         {
-            var projectXml = fileSystem.File.ReadAllText(projectPath);
-            var xDocument = XDocument.Load(new StringReader(projectXml));
-            var xPathEvaluate = xDocument.XPathEvaluate("//Project//ItemGroup//Service/@Include") as IEnumerable;
-
-            var hasTestGuid = xPathEvaluate
-                .Cast<XAttribute>()
-                .Any(x => string.Equals(x.Value, TestServiceGuid, StringComparison.OrdinalIgnoreCase));
```
duncanp-sonar, line 75: I'm not very comfortable with rewriting this as a generic method and "promoting" it to the project system helper. It's a hacky solution that was just acceptable for the specific case of looking for a specific service tag we know was added in a precise form by an old version of VS.
Adding it as a general method to the project system helper implies it is good enough for a general purpose solution.

Taking as step back, do we need to check whether the file is already referenced before trying to add it via the DTE? What's the behaviour if the file is already referenced in the project?

src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
```diff
@@ -730,5 +736,177 @@ int IVsShell.UnadviseShellPropertyChanges(uint dwCookie)
...
+        {
+            var projectXml = @"<Project Sdk=""Microsoft.NET.Sdk"">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <MyGroup Include=""MyValue1"" />
+  </ItemGroup>
+  <ItemGroup>
+    <MyGroup Include=""MyValue"" />
```
duncanp-sonar, line 901: XML Attributes can be single-quoted or double-quoted, so a test with single-quotes would be good.

src/Integration/LocalServices/TestProjectIndicators/ServiceGuidTestProjectIndicator.cs
rita-g-sonarsource: The problem is that we need a way to re-bind existing projects, for users that already have a valid ruleset but don't have an additional file.
For that I need to mark the project as BindingOperationRequired, so I need some way to check that if project references the additional file or not. I will see if `IsFileInProject` gives me that functionality for new SDK projects, but I'm a bit afraid of corner-cases for file links and etc.
I'll check `IsFileInProject`.

duncanp-sonar: The corner cases are why we want to use a VS-provided API to do the check. VS will have processed file links, conditions, imports etc to have calculated what is logically included in the project.

rita-g-sonarsource:
@duncanp-sonar I closed this PR, won't merge

PR #1408, opened 2020-05-14
Exclude hotspots from C# and VB config

### LINKED ISSUE #1401: [C#/VB] Connected mode configuration should not contain hotspot rules config
### Description
The rulesets and SonarLint.xml files generated for C# and VB should not contain information about hotspot rules.

See parent ticket #1382 for more information.

##### Notes:
* currently the sonar-dotnet analyzers use the `SonarLint.xml` file to determine whether hotspots should be enabled or not. See [here](https://github.com/SonarSource/sonar-dotnet/blob/734df7968037387faade9353c7aedd1a77c010a0/sonaranalyzer-dotnet/src/SonarAnalyzer.Common/Common/AnalyzerConfiguration.cs#L67).
* the check is done once and stored statically, so it requires a VS restart to change whether cached hotspot configuration. However, this shouldn't be a problem immediately since at present we want hotspots to be disabled for now.

Related issue: [sonar-dotnet#2385](https://github.com/SonarSource/sonar-dotnet/issues/2385) _Remove Security Hotspot rules which are not targeting developers_

duncanp-sonar:
Fixes #1401

### DIFF
--- a/src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
+++ b/src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
@@ -48,6 +48,19 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         private IList<NuGetPackageInfo> validNugetPackages;
         private Core.CSharpVB.RuleSet validRuleSet;
 
+        private static readonly SonarQubeRule ActiveRuleWithUnsupportedSeverity = new SonarQubeRule("activeHotspot", "any1",
+            true, SonarQubeIssueSeverity.Blocker, null, SonarQubeIssueType.SecurityHotspot);
+
+        private static readonly SonarQubeRule InactiveRuleWithUnsupportedSeverity = new SonarQubeRule("inactiveHotspot", "any2",
+            false, SonarQubeIssueSeverity.Blocker, null, SonarQubeIssueType.SecurityHotspot);
+
+        private static readonly SonarQubeRule ActiveTaintAnalysisRule = new SonarQubeRule("activeTaint", "roslyn.sonaranalyzer.security.foo",
+            true, SonarQubeIssueSeverity.Blocker, null, SonarQubeIssueType.CodeSmell);
+
+        private static readonly SonarQubeRule InactiveTaintAnalysisRule = new SonarQubeRule("inactiveTaint", "roslyn.sonaranalyzer.security.bar",
+            false, SonarQubeIssueSeverity.Blocker, null, SonarQubeIssueType.CodeSmell);
+
+
         [TestInitialize]
         public void TestInitialize()
         {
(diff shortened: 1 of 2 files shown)


src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
```diff
@@ -48,14 +48,27 @@ public class CSharpVBBindingConfigProviderTests
...
+
+
         [TestInitialize]
         public void TestInitialize()
         {
             emptyRules = Array.Empty<SonarQubeRule>();
 
             validRules = new List<SonarQubeRule>
             {
-                new SonarQubeRule("key", "repoKey", true, SonarQubeIssueSeverity.Blocker, null, SonarQubeIssueType.Unknown)
```
duncanp-sonar, line 58: The filtering by issue type now means that `Unknown` means the rule isn't supported.

src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
```diff
@@ -110,12 +123,12 @@ public void IsLanguageSupported()
...
 
         [TestMethod]
-        public async Task GetConfig_NoActiveRules_ReturnsNull()
+        public async Task GetConfig_NoSupportedActiveRules_ReturnsNull()
         {
             // Arrange
             var builder = new TestEnvironmentBuilder(validQualityProfile, Language.VBNET)
             {
-                ActiveRulesResponse = emptyRules,
+                ActiveRulesResponse = new List<SonarQubeRule> { ActiveRuleWithUnsupportedSeverity },
```
duncanp-sonar, line 131: This checks we'll early-out if there are no active supported rules.

src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
```diff
@@ -196,13 +209,15 @@ public async Task GetConfig_HasActiveInactiveAndUnsupportedRules_ReturnsValidBin
             var activeRules = new SonarQubeRule[]
                 {
                     CreateRule("activeRuleKey", "repoKey1", true),
-                    CreateRule("activeUnsupported", "roslyn.sonaranalyzer.security.foo", true)
+                    ActiveTaintAnalysisRule,
+                    ActiveRuleWithUnsupportedSeverity
```
duncanp-sonar, line 213: ... and the changes to this test make sure we check unsupported issue types are filtered out of both active and inactive rules.

src/Integration/Binding/CSharpVBBindingConfigProvider.cs
```diff
@@ -166,9 +166,16 @@ private static VsRuleset ToVsRuleset(CoreRuleset coreRuleset)
...
         }
+
+
+        private static bool IsTaintAnalysisRule(SonarQubeRule rule) =>
+            rule.RepositoryKey.StartsWith(TaintAnalyisRepoPrefix, StringComparison.OrdinalIgnoreCase);
+
+        private static bool IsSupportedIssueType(SonarQubeIssueType issueType) =>
+            issueType == SonarQubeIssueType.CodeSmell ||
+            issueType == SonarQubeIssueType.Bug ||
+            issueType == SonarQubeIssueType.Vulnerability;
```
rita-g-sonarsource, line 179: I didn't know that we support vulnerabilities, we don't need the whole scope for that?

duncanp-sonar: No. All of the vulnerabilities detected by the taint analysis engine need more information, but there are a lots of simpler vulnerabilities that can be detected at project or file level.

Have a look at https://rules.sonarsource.com/csharp/type/Vulnerability/ for examples e.g. [5547](https://rules.sonarsource.com/csharp/type/Vulnerability/RSPEC-5547).

Issue #1377, opened 2020-05-01
Query for project properties should pass the project key


duncanp-sonar:
Reminder to-do: the ~~`DotNetBindingConfigurationProvider`~~ `CSharpVBBindingConfigProvider` needs the project key to correctly fetch the properties applicable to the project.

Once the binding refactoring code has been merged, the required data should be available to the class.

Depends on #1376

duncanp-sonar:
Fixed as part of the PR #1376

PR #1410, opened 2020-05-14
Fixed up unit test mocks to check for specific project key

duncanp-sonar:
Relates to #1377


src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
```diff
@@ -337,7 +338,7 @@ public CSharpVBBindingConfigProvider CreateTestSubject()
                     .ReturnsAsync(InactiveRulesResponse);
 
                 sonarQubeServiceMock
-                    .Setup(x => x.GetAllPropertiesAsync(It.IsAny<string>(), It.IsAny<CancellationToken>()))
```
duncanp-sonar, line 340: Updated the mock to only return the properties if the project key is correctly passed through.

PR #1411, opened 2020-05-15
Converted CSharpVB config to use Core RuleSet

duncanp-sonar:


rita-g-sonarsource:
@duncanp-sonar why did you decide to close it?

PR #1412, opened 2020-05-15
Remove SolutionBindingFilePathGenerator

rita-g-sonarsource:


src/Core/Binding/BindingConfiguration.cs
```diff
@@ -88,5 +89,12 @@ public override int GetHashCode()
         }
 
         #endregion
+
+        public string BuildEscapedPathUnderProjectDirectory(string fileName = "")
```
duncanp-sonar, line 93: 'fileName' is a bit misleading - isn't it always a file suffix, or empty?

`BuildPathUnderConfigDirectory` might be better for the method name; "ProjectDirectory" is ambiguous, and the method name probably doesn't need to specify that escaping is taking place.

PR #1413, opened 2020-05-15
[WIP] Fix additional file detection

rita-g-sonarsource:
1. Another sonarlint.xml exists in the project
2. SonarLint.xml is referenced but is not marked as AdditionalFiles

@duncanp-sonar this is WIP -- still missing tests (also, please note the branch base).


src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -127,19 +127,12 @@ private void AddRuleset()
...
-                // It's possible that the additional file is set property, but the ruleset is out of date, which will trigger the binding. So we need to remove the file first, otherwise we get an error that the file already exists.
-                // It's possible that the additional file exists but has a wrong type, e.g. Compile/Content, and not AdditionalFiles. In which case removing and re-adding it solves the problem.
-                projectSystem.RemoveFileFromProject(initializedProject, cSharpVBBindingConfig.AdditionalFile.Path);
-                projectSystem.AddFileToProject(initializedProject, cSharpVBBindingConfig.AdditionalFile.Path, "AdditionalFiles");
-            }
-            catch (Exception ex)
-            {
-                throw new Exception($"Failed to bind project {initializedProject.FullName}, exception details: {ex}");
-            }
+            var projectSystem = serviceProvider.GetService<IProjectSystemHelper>();
```
rita-g-sonarsource, line 130: try-catch removal

src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -156,10 +157,27 @@ private RuleSet FindDeclarationRuleSet(Project project, RuleSetDeclaration decla
             return ruleSetSerializer.LoadRuleSet(ruleSetFilePath);
         }
 
-        private bool ProjectHasAdditionalFile(Project project, string additionalFilePath)
+        public bool DoesProjectReferenceAdditionalFile(Project project, string solutionAdditionalFilePath)
         {
-            // ritag todo: 1. Check if there is another SonarLint.xml, 2. Check if the property is AdditionalFiles
-            return projectSystemHelper.IsFileInProject(project, additionalFilePath);
+            var additionalFile = projectSystemHelper.GetFileInProject(project, solutionAdditionalFilePath);
```
rita-g-sonarsource, line 162: I don't know if GetFileInProject behaves differently that IsFileInProject.
IsFileInProject uses underlying API `IsDocumentInProject`, while GetFileInProject simply iterates the ProjectItems.

src/Integration/LocalServices/ProjectSystemHelper.cs
```diff
@@ -218,7 +236,7 @@ public void RemoveFileFromProject(Project project, string fileName)
             }
 
             // If the project is an empty solution folder, then remove it
-            if (project.ProjectItems.Count == 0 && project.Kind == ProjectSystemHelper.VsProjectItemKindSolutionFolder)
+            if (project.ProjectItems.Count == 0 && project.Kind == VsProjectItemKindSolutionFolder)
```
rita-g-sonarsource, line 239: cosmetics

src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -156,10 +157,27 @@ private RuleSet FindDeclarationRuleSet(Project project, RuleSetDeclaration decla
...
+
+            if (additionalFile == null)
```
rita-g-sonarsource, line 164: WIP: missing tests:
1. Project does not reference solution SonarLint.xml and has no other SonarLint.xml --> IsBindingRequired=true.
2. Project does not reference solution SonarLint.xml and has another SonarLint.xml --> Exception.
3. Project references solution SonarLint.xml but it's not AdditionalFiles --> IsBindingRequired=true.
4. Project references solution SonarLint.xml and it's AdditionalFiles --> IsBindingRequired=false.

I think it's also worth extracting it to another class as it starts to get too much (and potentially to a generic IsBindingRequiredIndicator, that could be used for ruleset as well, and this class will just be a composite of all the indicators).

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -127,19 +127,12 @@ private void AddRuleset()
...
+            projectSystem.AssertLocalServiceIsNotNull();
+            // It's possible that the additional file is set property, but the ruleset is out of date, which will trigger the binding. So we need to remove the file first, otherwise we get an error that the file already exists.
+            // It's possible that the additional file exists but has a wrong type, e.g. Compile/Content, and not AdditionalFiles. In which case removing and re-adding it solves the problem.
+            projectSystem.RemoveFileFromProject(initializedProject, Path.GetFileName(cSharpVBBindingConfig.AdditionalFile.Path));
```
rita-g-sonarsource, line 134: WIP: missing tests for the two use-cases in the comment.

src/Integration/Binding/CSharpVBBindingOperation.cs
duncanp-sonar: Any file system operation either direct or indirect via the DTE could potentially throw.
Having said that, we should have a try-catch block higher up to catch exceptions during binding. I'll check.

src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -156,10 +157,27 @@ private RuleSet FindDeclarationRuleSet(Project project, RuleSetDeclaration decla
...
+        public bool DoesProjectReferenceAdditionalFile(Project project, string solutionAdditionalFilePath)
         {
-            // ritag todo: 1. Check if there is another SonarLint.xml, 2. Check if the property is AdditionalFiles
-            return projectSystemHelper.IsFileInProject(project, additionalFilePath);
+            var additionalFile = projectSystemHelper.FindFileInProject(project, solutionAdditionalFilePath);
+
+            if (additionalFile == null)
+            {
+                var additionalFileName = Path.GetFileName(solutionAdditionalFilePath);
+                var hasAnotherAdditionalFile = projectSystemHelper.IsFileInProject(project, additionalFileName);
```
duncanp-sonar, line 167: FYI I couldn't get VS to hit this code path when debugging - "IsFileInProject" never returned `true` when searching by just the file name.

src/Integration/Binding/CSharpVBBindingOperation.cs
duncanp-sonar, line 134: I encountered a few problems during testing:
1) the "remove" is removing based just on the file name, so it will remove any file called "SonarLint.xml". This seems odd, since we have a check elsewhere that is supposed to throw an exception and stop the binding if there is another "SonarLint.xml" file (although that check doesn't seem to be working);
2) attempting to remove file or change the item group value throws an exception if the reference to the file is in a `Directory.Build.props` file;
3) removing and re-adding the item can cause the entry to change location in the project file, which doesn't matter functionally but is irritating when diff-ing the file for changes.

It's simple enough to do a check to see if the correct file is already referenced using the full path since we already have a method for that, so we don't need to remove the file - if it doesn't exist we can add it, and if it does exist we can try to change the item group if necessary.

However, trying to change the item group will still fail if the entry is in an imported file, and I don't think we can easily do anything about that except make sure we catch the exception, fail the binding and warn the user.

PR #1416, opened 2020-05-19
Modified CS/VB binder to update existing additional file ref

duncanp-sonar:


src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -92,7 +91,7 @@ public void Prepare(CancellationToken token)
 
                 string targetRuleSetFileName = group.Key;
                 string currentRuleSetFilePath = group.First().CurrentRuleSetFilePath;
-                Debug.Assert(group.All(i => StringComparer.OrdinalIgnoreCase.Equals(currentRuleSetFilePath, currentRuleSetFilePath)), "Expected all the rulesets to be the same when the target rule set name is the same");
+                Debug.Assert(group.All(_ => StringComparer.OrdinalIgnoreCase.Equals(currentRuleSetFilePath, currentRuleSetFilePath)), "Expected all the rulesets to be the same when the target rule set name is the same");
```
duncanp-sonar, line 94: Unrelated: removed a code smell

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -125,20 +124,55 @@ private void AddRuleset()
...
         }
 
-        private void AddAdditionalFile()
+        private void EnsureAdditionalFileIsReferenced()
         {
-            try
+            var projectSystem = serviceProvider.GetService<IProjectSystemHelper>();
+            projectSystem.AssertLocalServiceIsNotNull();
+
+            // We assume that we've checked during the "Prepare" step that there are references to other
```
duncanp-sonar, line 132: There's enough complexity and weirdness around what does and doesn't worth that I though it worth the extensive documentation.

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -125,20 +124,55 @@ private void AddRuleset()
...
             }
-            catch (Exception ex)
+            else
             {
-                throw new Exception($"Failed to bind project {initializedProject.FullName}, exception details: {ex}");
+                var property = VsShellUtils.FindProperty(refdFile.Properties, Constants.ItemTypePropertyKey);
+                if (!Constants.AdditionalFilesItemTypeName.Equals(property.Value?.ToString(), StringComparison.OrdinalIgnoreCase))
+                {
+                    // This will fail if the existing item ref isn't in the main project file e.g. if it is in a Directory.Build.props file.
+                    // We can't do much about this, other than catch and log the error higher up.
```
duncanp-sonar, line 173: The binding will fail with the rather cryptic `Only included item inside the project can be moved.` mentioned above.
We can discuss whether it's worth logging more useful error here.
However, it should be very uncommon - the correct file needs to be referenced in a `Directory.Build.props`, with an item type that VS recognises.

Adding the exception handling and logging is straightforward, but it would require a re-write of the tests to inject a failure at the right point.

src/TestInfrastructure/Framework/ConfigurableVsProjectSystemHelper.cs
```diff
@@ -52,7 +52,8 @@ public Project GetSolutionFolderProject(string solutionFolderName, bool createOn
 
         public ProjectItem FindFileInProject(Project project, string fileName)
         {
-            throw new NotImplementedException();
+            var item = project.ProjectItems.OfType<ProjectItem>().FirstOrDefault(pi => StringComparer.OrdinalIgnoreCase.Equals(pi.Name, fileName));
+            return item;
```
duncanp-sonar, line 56: Fix the mock implementation to make the method work sufficiently for the new tests.

PR #1417, opened 2020-05-20
Detect conflicting SonarLint.xml

rita-g-sonarsource:
Part of #176


src/Integration/Binding/AdditionalFileConflictChecker.cs
```diff
@@ -0,0 +1,115 @@
...
+                   ExistsInProject(project, additionalFileName, out conflictingAdditionalFilePath);
+        }
+
+        private bool ExistsUnderRootFolder(Project project, string additionalFileName, out string conflictingAdditionalFilePath)
+        {
+            var projectRootDirectory = Path.GetDirectoryName(project.FullName);
+            conflictingAdditionalFilePath = Path.Combine(projectRootDirectory, additionalFileName);
+
+            // For old-style SDK projects, the file can exist on disk but not referenced in the project, so we check using the file system
+            return fileSystem.File.Exists(additionalFileName);
```
duncanp-sonar, line 72: This should be checking `conflictingAdditionalFilePath`.

src/Integration/Binding/AdditionalFileConflictChecker.cs
```diff
@@ -0,0 +1,115 @@
...
+            return ExistsUnderRootFolder(project, additionalFileName, out conflictingAdditionalFilePath) ||
+                   ExistsInProject(project, additionalFileName, out conflictingAdditionalFilePath);
+        }
+
+        private bool ExistsUnderRootFolder(Project project, string additionalFileName, out string conflictingAdditionalFilePath)
+        {
+            var projectRootDirectory = Path.GetDirectoryName(project.FullName);
+            conflictingAdditionalFilePath = Path.Combine(projectRootDirectory, additionalFileName);
+
+            // For old-style SDK projects, the file can exist on disk but not referenced in the project, so we check using the file system
```
duncanp-sonar, line 71: `Old-style MSBuild projects` would be more accurate - SDK projects are new-style projects.

src/Integration/Binding/AdditionalFileConflictChecker.cs
rita-g-sonarsource: Yep my bad :( I pushed a fix now

src/Integration.UnitTests/Binding/AdditionalFileConflictCheckerTests.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            result.Should().BeTrue();
+            conflictingPath.Should().Be(conflictingFilePath);
+        }
+
+        [TestMethod]
+        [DataRow("AdditionalFiles", true)]
+        [DataRow("Foo", false)]
+        [DataRow("Content", false)]
+        public void HasConflictingAdditionalFile_ConflictingFileFoundInProject_ReturnsIfFileIsAdditionalFile(string itemType, bool isConflicting)
```
rita-g-sonarsource, line 70: I'm having a tough time to test folders, ProjectItemsMock will need some adjusting, I'll do it in separate commit/PR

src/Integration/Binding/AdditionalFileConflictChecker.cs
```diff
@@ -0,0 +1,97 @@
...
+                }
+            }
+
+            conflictingAdditionalFilePath = string.Empty;
+            return false;
+        }
+
+        private bool HasAdditionalFile(ProjectItem projectItem, string additionalFileName, out string conflictingAdditionalFilePath)
+        {
+            if (projectItem.FileNames[0].EndsWith(additionalFileName))
```
duncanp-sonar, line 73: Two things:
* comparison needs to be `OrdinalIgnoreCase`
* should just check the file name part. We don't want to match "MySonarLint.xml`

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -99,6 +101,27 @@ public void Prepare(CancellationToken token)
...
+            // If the correct file is already in the project, it means we were successful in adding it and there is no clash
+            if (!projectSystem.IsFileInProject(initializedProject, cSharpVBBindingConfig.AdditionalFile.Path))
+            {
+                var additionalFileName = Path.GetFileName(cSharpVBBindingConfig.AdditionalFile.Path);
+
+                var hasConflict = additionalFileConflictChecker.HasConflictingAdditionalFile(initializedProject, additionalFileName, out var conflictingAdditionalFilePath);
+
+                if (hasConflict)
+                {
+                    throw new Exception($"Could not bind project {initializedProject.FullName}: {Path.GetFileName(cSharpVBBindingConfig.AdditionalFile.Path)} already exists. Please remove '{conflictingAdditionalFilePath}' and try again.");
```
duncanp-sonar, line 122: Throw `SonarLintException` and localize the error message.
Also, I'd put single quotes round the project name, and put a line break after the first sentence (more readable, and consistent with the new error messages I've just added).

FYI it looks like release builds of SLVS only log the exception message, not the full stack trace, so I don't think we need to do anything special to reduce the noise in the output window.

src/Integration/Binding/AdditionalFileConflictChecker.cs
```diff
@@ -0,0 +1,97 @@
...
+                   ExistsInProject(project, additionalFileName, out conflictingAdditionalFilePath);
+        }
+
+        private bool ExistsUnderRootFolder(Project project, string additionalFileName, out string conflictingAdditionalFilePath)
+        {
+            var projectRootDirectory = Path.GetDirectoryName(project.FullName);
+            conflictingAdditionalFilePath = Path.Combine(projectRootDirectory, additionalFileName);
+
+            // For old-style MSBuild projects, the file can exist on disk but not referenced in the project, so we check using the file system
+            return fileSystem.File.Exists(conflictingAdditionalFilePath);
```
duncanp-sonar, line 54: When manually testing this case, I observed the following:
1) SonarLint.xml exists in folder but is not referenced -> detected by this method and binding fails.
2) delete the file from the disc via Windows Explorer and try again. File is not detected and binding continues as expected. However, the commit then fails with the error that a file with the same name exists.

However, it works if I close and re-open the solution after deleting the file.
It looks like the DTE hasn't picked up the fact that the file no longer exists.

It appears to be only this case that requires the solution to be reloaded.

rita-g-sonarsource: Hmm, this might actually be a problem, since we ask the user to delete the file and retry.
So it's very likely that this scenario will be encountered.
Perhaps I should change the message, to "Please delete the file and re-open the solution".

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
```diff
@@ -390,64 +449,6 @@ public void ProjectBindingOperation_Commit_FileIsAlreadyReferenced(string existi
             projectMock.Files.Count().Should().Be(1); // existing file updated
         }
 
-        [TestMethod]
-        public void AddAdditional_NonCriticalException_IsCaughtAndWrapped()
```
duncanp-sonar, line 394: Did you mean to delete these tests? They're still required.

rita-g-sonarsource: Damn, a merge conflict! I'll return them of course

src/Integration/Binding/AdditionalFileConflictChecker.cs
duncanp-sonar: Ok. Ideally we wouldn't ask the user to re-open the solution if it isn't required but that would require more work, and most users won't have an existing SonarLint.xml so it won't matter.

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -100,6 +102,28 @@ public void Prepare(CancellationToken token)
...
                 }
             }
+
+            EnsureNoConflictingAdditionalFile();
+        }
+
+        private void EnsureNoConflictingAdditionalFile()
+        {
+            var projectSystem = serviceProvider.GetService<IProjectSystemHelper>();
+            projectSystem.AssertLocalServiceIsNotNull();
```
duncanp-sonar, line 112: Minor inconsistency - the other services are being fetched in the constructor.

I wouldn't change this now. I'd consider changing the constructor at a later stage so it doesn't use the service provider at all - the class that constructs it has almost of required services anyway.

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -100,6 +102,28 @@ public void Prepare(CancellationToken token)
...
+
+            // If the correct file is already in the project, it means we were successful in adding it and there is no clash
+            if (!projectSystem.IsFileInProject(initializedProject, cSharpVBBindingConfig.AdditionalFile.Path))
+            {
+                var additionalFileName = Path.GetFileName(cSharpVBBindingConfig.AdditionalFile.Path);
+
+                var hasConflict = additionalFileConflictChecker.HasConflictingAdditionalFile(initializedProject, additionalFileName, out var conflictingAdditionalFilePath);
```
duncanp-sonar, line 119: An alternative way to write the conflict checker would have been to make it responsible for throwing exception e.g. `ThrowIfConflictingSonarLintXmlExists`. That way the checker could have thrown specific error messages for different conditions.

Just an observation - don't change it.

src/Integration.UnitTests/Binding/AdditionalFileConflictCheckerTests.cs
duncanp-sonar: @rita-g-sonarsource do it in a separate PR i.e. let's get this committed as soon as possible, then Chris and I can start testing the artefacts created by the PR from the branch into master.

src/Integration/Binding/CSharpVBBindingOperation.cs
rita-g-sonarsource: Good idea, could've avoided the `out` param

PR #1419, opened 2020-05-20
Added more specific binding failure messages

duncanp-sonar:


src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -162,22 +166,47 @@ private void EnsureAdditionalFileIsReferenced()
...
+                AddAdditionalFileToProject(projectSystem, initializedProject, additionalFilePath);
             }
             else
             {
-                var property = VsShellUtils.FindProperty(refdFile.Properties, Constants.ItemTypePropertyKey);
+                EnsureItemTypeIsCorrect(refdFile);
+            }
+        }
+
+        internal /* for testing */ static void AddAdditionalFileToProject(IProjectSystemHelper projectSystem, Project project, string filePath)
```
duncanp-sonar, line 177: I've just extracted the existing calls to separate methods with try-catch blocks and customised error messages.
The new methods are static to make testing easier.

PR #1422, opened 2020-05-22
Fix ruleset duplication bug

### LINKED ISSUE #1423: Project RuleSet is duplicated when SonarLint.xml is not found
### Description

Bug caused by #176: when a project is already connected to the server but does not have `SonarLint.xml`, syncing the project will create duplicated rulesets.

### Steps to reproduce
1. Bind a project in a previous SonarLint version (before #176)
2. Re-bind the project in current SonarLint version (with #176)

Expected: Only `SonarLint.xml` is added.
Actual: Rulesets are re-added as well.

rita-g-sonarsource:
Fixes #1423 

I split the code inside `CSharpVBProjectBinder` into 2 classes for ease of testing: `RuleSetReferenceChecker` and `AdditionalFileReferenceChecker`.
* `AdditionalFileReferenceChecker` was unchanged, and the tests are cut-paste from `CSharpVBProjectBinderTests`.
* `RuleSetReferenceChecker` now contains the logic that was in `CSharpVBProjectBinder` and in the static `RuleSetIncludeChecker`.
* `RuleSetReferenceChecker` is now referenced in `CSharpVBBindingOperation`: in the Initialize phase, I check if the ruleset is already referenced. If so - no changes are made to the project's rulesets.


src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -239,30 +246,32 @@ private void CaptureProject()
...
-            var solutionRuleSetProvider = this.serviceProvider.GetService<ISolutionRuleSetsInformationProvider>();
-            RuleSetDeclaration[] ruleSetsInfo = solutionRuleSetProvider.GetProjectRuleSetsDeclarations(this.initializedProject).ToArray();
+            var solutionRuleSet = ruleSetSerializer.LoadRuleSet(cSharpVBBindingConfig.RuleSet.Path);
 
-            string sameRuleSetCandidate = ruleSetsInfo.FirstOrDefault()?.RuleSetPath;
-
-            // Special case: if all the values are the same use project name as the target ruleset name
-            bool useSameTargetName = false;
-            if (ruleSetsInfo.All(r => StringComparer.OrdinalIgnoreCase.Equals(sameRuleSetCandidate, r.RuleSetPath)))
+            if (ruleSetReferenceChecker.IsReferenced(initializedProject, solutionRuleSet))
```
rita-g-sonarsource, line 251: This is the bug fix

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -239,30 +246,32 @@ private void CaptureProject()
...
             {
-                useSameTargetName = true;
+                return;
             }
 
-            string projectBasedRuleSetName = Path.GetFileNameWithoutExtension(this.initializedProject.FullName);
-            foreach (RuleSetDeclaration singleRuleSetInfo in ruleSetsInfo)
+            var solutionRuleSetProvider = serviceProvider.GetService<ISolutionRuleSetsInformationProvider>();
```
rita-g-sonarsource, line 256: just cosmetics, "var" changes. Code remains the same.

src/Integration/Binding/RuleSetReferenceChecker.cs
```diff
@@ -22,31 +22,60 @@
...
 using System.IO;
 using System.Linq;
+using EnvDTE;
 using Microsoft.VisualStudio.CodeAnalysis.RuleSets;
 
-namespace SonarLint.VisualStudio.Integration
+namespace SonarLint.VisualStudio.Integration.Binding
 {
-    internal static class RuleSetIncludeChecker
+    internal class RuleSetReferenceChecker : IRuleSetReferenceChecker
```
rita-g-sonarsource, line 30: I wanted to re-use this logic in CSharpVBBindingOperation so I extracted it to class.

src/Integration/Binding/IRuleSetReferenceChecker.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using EnvDTE;
+using Microsoft.VisualStudio.CodeAnalysis.RuleSets;
+
+namespace SonarLint.VisualStudio.Integration.Binding
+{
+    internal interface IRuleSetReferenceChecker
+    {
+        /// <summary>
+        /// Return true if the the project's ruleSets directly include the target ruleSet, otherwise false
```
rita-g-sonarsource, line 29: Doc comment from `RuleSetIncludeChecker`

src/Integration/Binding/IAdditionalFileReferenceChecker.cs
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using EnvDTE;
+
+namespace SonarLint.VisualStudio.Integration.Binding
+{
+    internal interface IAdditionalFileReferenceChecker
```
rita-g-sonarsource, line 25: The tests for CSharpVBProjectBinder started to get a bit complicated, so it was much easier to test this way that the class behaves correctly when both conditions are true (the additional file exists + the ruleset is referenced). The code inside is not changed.

src/Integration.UnitTests/Binding/AdditionalFileReferenceCheckerTests.cs
```diff
@@ -0,0 +1,91 @@
...
+using System;
+using EnvDTE;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Integration.Binding;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
+{
+    [TestClass]
```
rita-g-sonarsource, line 30: Tests that were in CSharpVBProjectBinderTests

src/Integration.UnitTests/Binding/RuleSetReferenceCheckerTests.cs
```diff
@@ -0,0 +1,317 @@
...
+using EnvDTE;
+using FluentAssertions;
+using Microsoft.VisualStudio.CodeAnalysis.RuleSets;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Integration.Binding;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
+{
+    [TestClass]
```
rita-g-sonarsource, line 33: New tests + Tests that were in RuleSetIncludeCheckerTests

Directory.Build.props
```diff
@@ -8,12 +8,6 @@
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory).sonarlint\sonarlint-visualstudioCsharp.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 
-  <ItemGroup Label="Common items">
```
rita-g-sonarsource, line 11: This is not actually related to the PR, it was a way for me to test that the fix works. I can revert it and push it in another commit (that will also have the server project binding changes).

duncanp-sonar: Personally, I'd probably revert it since it isn't part of the fix, and the only thing that will change when we fix up the project binding is the path in line 12. However, it doesn't matter much either way.

src/Integration/Binding/IAdditionalFileReferenceChecker.cs
duncanp-sonar: Testability is important, but I think the increasing number of tiny interface and implementation classes is starting to obscure the overall structure of the system e.g. it's not easy to tell which interfaces as high-level abstractions and which ones are small-scale, language specific.

Here, there's nothing to suggests that it is either CSharpVB-specific and it appears at the same level as "larger" abstractions like `IProjectBinder`, both in the namespace and in the file directory. As a minimum, it should be possible from interface name/namespace to tell that this interface is a low-level language-specific interface.

The simplest thing to do would be rename the interface/namespace. An alternative would be to nest it inside a `partial class CSharpVBProjectBinder`. Nesting the class and interface would express more clearly that it is an implementation detail of that project binder and that it isn't expected to be used elsewhere (and if it turns out to needed elsewhere, then we'd refactor).

rita-g-sonarsource: I'll rename to ICSharpVBAdditionalFileReferenceChecker

src/Integration/Binding/IAdditionalFileReferenceChecker.cs
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using EnvDTE;
+
+namespace SonarLint.VisualStudio.Integration.Binding
+{
```
duncanp-sonar, line 24: A doc comment would be useful to make clear that `true` will only be returned if the file is referenced using the correct item type.

It's not obvious what the division of responsibilities between this interface and the similarly-named `IAdditionalFileConflictChecker` is, if any, but that's a separate problem.

rita-g-sonarsource:
> It was particularly difficult to review differences in the tests in the GitHub UI.

I agree, the UI is terrible for this. I just give up and view the file in directory, by selecting "View file" :/

src/Integration/Binding/IAdditionalFileReferenceChecker.cs
duncanp-sonar: The interface name should change too.

PR #1429, opened 2020-05-28
Removed VS RuleSet param from IRuleSetReferenceChecker interface

duncanp-sonar:
We only need the file path, not an actual ruleset object.

### DIFF
--- a/src/Integration/Binding/CSharpVBProjectBinder.cs
+++ b/src/Integration/Binding/CSharpVBProjectBinder.cs
@@ -87,7 +87,7 @@ namespace SonarLint.VisualStudio.Integration.Binding
 
         private bool IsFullyBoundProject(BindingConfiguration binding, Project project, Language language)
         {
-            if (!IsSolutionBound(binding, language, out var solutionRuleSet, out var additionalFilePath))
+            if (!IsSolutionBound(binding, language, out var solutionRuleSetFilePath, out var additionalFilePath))
             {
                 return false;
             }
@@ -99,14 +99,13 @@ namespace SonarLint.VisualStudio.Integration.Binding
                 return false;
             }
 
-            var isRuleSetBound = ruleSetReferenceChecker.IsReferenced(project, solutionRuleSet);
-
+            var isRuleSetBound = ruleSetReferenceChecker.IsReferenced(project, solutionRuleSetFilePath);
             return isRuleSetBound;
         }
 
-        private bool IsSolutionBound(BindingConfiguration binding, Language language, out RuleSet solutionRuleSet, out string additionalFilePath)
+        private bool IsSolutionBound(BindingConfiguration binding, Language language, out string solutionRuleSetFilePath, out string additionalFilePath)
         {
-            solutionRuleSet = null;
+            solutionRuleSetFilePath = null;
             additionalFilePath = CSharpVBBindingConfigProvider.GetSolutionAdditionalFilePath(language, binding);
 
             if (!fileSystem.File.Exists(additionalFilePath))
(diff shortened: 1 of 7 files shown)


src/Integration/Binding/IRuleSetReferenceChecker.cs
```diff
@@ -19,17 +19,16 @@
...
     internal interface IRuleSetReferenceChecker
     {
         /// <summary>
-        /// Return true if the the project's ruleSets directly include the target ruleSet, otherwise false
+        /// Return true if the project's ruleSets directly include the target ruleSet, otherwise false
         /// </summary>
         /// <remarks>We don't currently check nested includes in RuleSet
         /// i.e. if Project has RuleSet A which includes RuleSet B that includes RuleSet C, IsReferenced(Project, C) returns false.</remarks>
-        bool IsReferenced(Project project, RuleSet ruleSet);
+        bool IsReferenced(Project project, string targetRuleSetFilePath);
```
duncanp-sonar, line 32: This is the change - everything else is fix-up.

The implementation code is only interested in the path to the ruleset file - we don't actually need a ruleset object at this point.

src/Integration/Binding/RuleSetReferenceChecker.cs
```diff
@@ -74,15 +74,15 @@ private RuleSet FindDeclarationRuleSet(Project project, RuleSetDeclaration decla
...
+        private bool HasInclude(RuleSet source, string targetRuleSetFilePath)
         {
             Debug.Assert(Path.IsPathRooted(source.FilePath));
-            Debug.Assert(Path.IsPathRooted(target.FilePath));
+            Debug.Assert(Path.IsPathRooted(targetRuleSetFilePath));
 
             // The path in the RuleSetInclude could be relative or absolute.
             // If relative, we assume it's relative to the source ruleset file.
             var sourceDirectory = Path.GetDirectoryName(source.FilePath);
-            var canonicalTargetFilePath = Path.GetFullPath(target.FilePath);
```
duncanp-sonar, line 85: ... and this is the only place the RuleSet.FilePath was accessed.

duncanp-sonar:
I started out by trying to remove the VS `RuleSet` from the `ICSharpVBBindingConfig` interface but it was starting to get confusing with the fixing up the tests, some of which would now need both a VS ruleset object and an equivalent Core ruleset.

Then I noticed that the VS RuleSet isn't needed by the `IRuleSetReferenceChecker`, so I started by simplifying that.

src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -99,22 +99,22 @@ private bool IsFullyBoundProject(BindingConfiguration binding, Project project,
...
+            solutionRuleSetFilePath = null;
             additionalFilePath = CSharpVBBindingConfigProvider.GetSolutionAdditionalFilePath(language, binding);
 
             if (!fileSystem.File.Exists(additionalFilePath))
             {
                 return false;
             }
 
-            solutionRuleSet = GetSolutionRuleSet(binding, language);
+            var solutionRuleSet = GetSolutionRuleSet(binding, language);
```
rita-g-sonarsource, line 116: Hmm, in `GetSolutionRuleSet` we actually try to load the ruleset -- do you think that's really needed? the code can be further simplified and we can change it to `GetSolutionRuleSetFilePath`

src/Integration/Binding/RuleSetReferenceChecker.cs
rita-g-sonarsource: Wow.  🤦

src/Integration/Binding/CSharpVBProjectBinder.cs
duncanp-sonar: I did try removing it, but it's still required. `GetSolutionRuleSet` calls the ruleset serializer, which checks the file exists, tries to load it, and returns null if the file can't be loaded i.e. it's checking the presence and validity of the ruleset. If the ruleset is missing or invalid we want `IsSolutionBound` to return `false`.

I thought I'd put in a comment about this but I can't find it...

rita-g-sonarsource: I was thinking that maybe we can remove the validity check, since we can check that the file exists using the file system, so we only serialize to check that it's valid. But on second thought, it's useful to check, so ignore my comment :)

PR #1430, opened 2020-05-29
Don't generate project-level ruleset when project has no rulesets

### LINKED ISSUE #817: [Connected Mode] Stop generating project-level ruleset when project has no rulesets
### Description

Currently connected mode creates a per-project ruleset file that in turn references the generated solution-level ruleset. This was to provide developers with an easy way to customise the ruleset for each project (i.e. they could just customise the per-project ruleset).

However, this has led to complaints about the number of files being added by connected mode, and the ability to customise project-level rulesets does not seem to be widely used.

### Proposal

* when creating new bindings, stop creating project-level rulesets when the project has no defined rulesets.
* instead, directly reference the generated-solution level ruleset from the project

### Related information

* SonarLint for VS Version: all (current is 4.8)

rita-g-sonarsource:
Fixes #817

I think we can only point to the Solution's ruleset when the user has no ruleset at all.
If there is a ruleset, we might not be able to amend it -- for example, if it's a pre-defined MS ruleset (under C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\Team Tools\Static Analysis Tools\Rule Sets).
I think that if there is any ruleset in the project, we should continue with the previous behavior: create a new ruleset that references our ruleset + the project's. It also solves the problem if the ruleset is some global ruleset that is shared across multiple solutions.

### DIFF
--- a/src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
+++ b/src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
@@ -210,10 +210,10 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
             testSubject.Prepare(CancellationToken.None);
 
             // Assert
-            string expectedRuleSetFileForPropertiesWithDefaultRulSets = cSharpVBBindingConfig.RuleSet.Path;
-            fileSystem.GetFile(expectedRuleSetFileForPropertiesWithDefaultRulSets).Should().NotBe(null);
-            testSubject.PropertyInformationMap[defaultRuleSetProperty1].NewRuleSetFilePath.Should().Be(expectedRuleSetFileForPropertiesWithDefaultRulSets, "Expected all the properties with default ruleset to have the same new ruleset");
-            testSubject.PropertyInformationMap[defaultRuleSetProperty2].NewRuleSetFilePath.Should().Be(expectedRuleSetFileForPropertiesWithDefaultRulSets, "Expected all the properties with default ruleset to have the same new ruleset");
+            string expectedRuleSetFileForPropertiesWithDefaultRuleSets = cSharpVBBindingConfig.RuleSet.Path;
+            fileSystem.GetFile(expectedRuleSetFileForPropertiesWithDefaultRuleSets).Should().NotBe(null);
+            testSubject.PropertyInformationMap[defaultRuleSetProperty1].NewRuleSetFilePath.Should().Be(expectedRuleSetFileForPropertiesWithDefaultRuleSets, "Expected all the properties with default ruleset to have the same new ruleset");
+            testSubject.PropertyInformationMap[defaultRuleSetProperty2].NewRuleSetFilePath.Should().Be(expectedRuleSetFileForPropertiesWithDefaultRuleSets, "Expected all the properties with default ruleset to have the same new ruleset");
 
             string expectedRulesetFilePath = Path.Combine(Path.GetDirectoryName(this.projectMock.FilePath), Path.GetFileNameWithoutExtension(this.projectMock.FilePath) + ".ruleset");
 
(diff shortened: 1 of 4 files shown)


src/Integration/Binding/CSharpVBBindingOperation.Writer.cs
```diff
@@ -70,9 +70,13 @@ internal partial class CSharpVBBindingOperation
                 return existingRuleSetPath;
             }
 
+            if (ShouldIgnoreConfigureRuleSetValue(currentRuleSetPath))
```
duncanp-sonar, line 73: As discussed offline, the name of `ShouldIgnoreConfigureRuleSetValue` is confusing in this context as the method is being called for a different reason, namely "ShouldGenerateProjectRuleSet".
Either change the name of the method, or alias it in some way.

duncanp-sonar, line 73: Not changed in this commit, but the scope of `ShouldIgnoreConfigureRuleSetValue` doesn't need to be public.

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
```diff
@@ -416,13 +418,38 @@ public void ProjectBindingOperation_Commit_NewProjectSystem_DoesNotAddFile()
...
+        }
+
         [TestMethod]
         public void ProjectBindingOperation_Commit_LegacyProjectSystem_DoesAddFile()
         {
             // Arrange
             CSharpVBBindingOperation testSubject = this.CreateTestSubject();
             this.projectMock.SetCSProjectKind();
-            PropertyMock prop = CreateProperty(this.projectMock, "config1", CSharpVBBindingOperation.DefaultProjectRuleSet);
+            PropertyMock prop = CreateProperty(this.projectMock, "config1", "ConcurrencyRules.ruleset");
```
duncanp-sonar, line 452: A minor point, but a completely arbitrary name like `ConcurrencyRules.ruleset` doesn't give any clue about why the property is being set in the test.
A name like `MyCustomProjectLevelRules.ruleset` or `NonDefaultRules.ruleset` gives more of a hint as to what the logical purpose of the property is.

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
```diff
@@ -209,24 +210,25 @@ public void ProjectBindingOperation_Prepare_VariousRuleSetsInProjects()
             testSubject.Prepare(CancellationToken.None);
 
             // Assert
-            string expectedRuleSetFileForPropertiesWithDefaultRulSets = Path.Combine(Path.GetDirectoryName(this.projectMock.FilePath), Path.GetFileNameWithoutExtension(this.projectMock.FilePath) + ".ruleset");
-            fileSystem.GetFile(expectedRuleSetFileForPropertiesWithDefaultRulSets).Should().Be(null);
+            string expectedRuleSetFileForPropertiesWithDefaultRulSets = cSharpVBBindingConfig.RuleSet.Path;
```
duncanp-sonar, line 213: Not changed in this PR, but there's a typo in the variable name: `...RulSets`

src/Integration/Binding/CSharpVBBindingOperation.Writer.cs
rita-g-sonarsource: I think maybe it should be renamed to `IsDefaultRuleSet`. It's more descriptive and actually makes more sense when I look at the usages.

src/Integration/Binding/CSharpVBBindingOperation.Writer.cs
```diff
@@ -70,11 +75,6 @@ internal partial class CSharpVBBindingOperation
                 return existingRuleSetPath;
             }
 
-            if (ShouldIgnoreConfigureRuleSetValue(currentRuleSetPath))
```
rita-g-sonarsource: The flow before was:
1. Call TryUpdateExistingProjectRuleSet
  1.1 It checks if IsDefaultRuleSet -- returns false
2. We continue and check again IsDefaultRuleSet

Now it's simpler:
1. Check if IsDefaultRuleSet
2. If not - call TryUpdateExistingProjectRuleSet

PR #1432, opened 2020-05-29
Add unconditional ruleset when binding for the first time

### LINKED ISSUE #805: Add unconditional ruleset references to the project file when binding for the first time
### Description

Currently connected mode adds references to rulesets inside conditional blocks for all of the configurations defined in the solution e.g.

```
  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|netstandard2.0|AnyCPU'">
    <CodeAnalysisRuleSet>SonarScanner.MSBuild.PreProcessor.ruleset</CodeAnalysisRuleSet>
  </PropertyGroup>
  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Release|netstandard2.0|AnyCPU'">
    <CodeAnalysisRuleSet>SonarScanner.MSBuild.PreProcessor.ruleset</CodeAnalysisRuleSet>
  </PropertyGroup>
```

It would be better if a single unconditional property was added, particularly for MSBuild15-format projects which are much simpler and cleaner than projects in MSBuild14 format.

### Known workarounds

From SLVS v4.8 onwards, the project file can be manually edited to remove the conditions, and clicking "Update" in the SonarQube section of the _Team Explorer_ won't re-add the conditions.

### Related information

* SonarLint for VS Version: up to 4.7
* Languages in your solution: C#, VB

#### Related
Bug SLVS #1207 is a case where the conditional ruleset references do not work correctly. We are not adding ruleset references for all of the target frameworks referenced in the project file, so if the user makes one of the other target frameworks the active framework in VS then SLVS will think the project is not bound correctly. Removing the conditions would also fix #1207.

### Notes
* the code currently sets the ruleset property via the DTE. It might be better to use _IVsBuildPropertyStorage_ instead.
* need to decide how to handle existing projects that contain conditionals - clean them up, or leave them as they are?

rita-g-sonarsource:
Fixes #805

### DIFF
--- a/src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
+++ b/src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
@@ -107,8 +107,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
             // Arrange
             CSharpVBBindingOperation testSubject = this.CreateTestSubject();
             this.projectMock.SetVBProjectKind();
-            PropertyMock prop1 = CreateProperty(this.projectMock, "config1", CSharpVBBindingOperation.DefaultProjectRuleSet);
-            PropertyMock prop2 = CreateProperty(this.projectMock, "config2", CSharpVBBindingOperation.DefaultProjectRuleSet);
+            PropertyMock prop1 = CreateRuleSetProperty(this.projectMock, "config1", CSharpVBBindingOperation.DefaultProjectRuleSet);
+            PropertyMock prop2 = CreateRuleSetProperty(this.projectMock, "config2", CSharpVBBindingOperation.DefaultProjectRuleSet);
 
             // Act
             testSubject.Initialize();
@@ -131,8 +131,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
             // Arrange
             CSharpVBBindingOperation testSubject = this.CreateTestSubject();
             this.projectMock.SetVBProjectKind();
-            PropertyMock prop1 = CreateProperty(this.projectMock, "config1", null);
-            PropertyMock prop2 = CreateProperty(this.projectMock, "config2", string.Empty);
+            PropertyMock prop1 = CreateRuleSetProperty(this.projectMock, "config1", null);
+            PropertyMock prop2 = CreateRuleSetProperty(this.projectMock, "config2", string.Empty);
 
             // Act
             testSubject.Initialize();
@@ -155,8 +155,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
             // Arrange
             CSharpVBBindingOperation testSubject = this.CreateTestSubject();
             this.projectMock.SetVBProjectKind();
-            PropertyMock prop1 = CreateProperty(this.projectMock, "config1", "Custom1.ruleset");
-            PropertyMock prop2 = CreateProperty(this.projectMock, "config2", "Custom1.ruleset");
+            PropertyMock prop1 = CreateRuleSetProperty(this.projectMock, "config1", "Custom1.ruleset");
+            PropertyMock prop2 = CreateRuleSetProperty(this.projectMock, "config2", "Custom1.ruleset");
 
             // Act
             testSubject.Initialize();
@@ -179,8 +179,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
             // Arrange
             CSharpVBBindingOperation testSubject = this.CreateTestSubject();
             this.projectMock.SetCSProjectKind();
-            PropertyMock prop1 = CreateProperty(this.projectMock, "config1", CSharpVBBindingOperation.DefaultProjectRuleSet);
-            PropertyMock prop2 = CreateProperty(this.projectMock, "config2", "NonDefualtRuleSet.ruleset");
+            PropertyMock prop1 = CreateRuleSetProperty(this.projectMock, "config1", CSharpVBBindingOperation.DefaultProjectRuleSet);
+            PropertyMock prop2 = CreateRuleSetProperty(this.projectMock, "config2", "NonDefualtRuleSet.ruleset");
 
             // Act
             testSubject.Initialize();
@@ -202,10 +202,10 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
             // Arrange
             CSharpVBBindingOperation testSubject = this.CreateTestSubject();
             this.projectMock.SetVBProjectKind();
-            PropertyMock customRuleSetProperty1 = CreateProperty(this.projectMock, "config1", "Custom.ruleset");
-            PropertyMock customRuleSetProperty2 = CreateProperty(this.projectMock, "config2", "Custom.ruleset");
-            PropertyMock defaultRuleSetProperty1 = CreateProperty(this.projectMock, "config3", CSharpVBBindingOperation.DefaultProjectRuleSet);
-            PropertyMock defaultRuleSetProperty2 = CreateProperty(this.projectMock, "config4", CSharpVBBindingOperation.DefaultProjectRuleSet);
+            PropertyMock customRuleSetProperty1 = CreateRuleSetProperty(this.projectMock, "config1", "Custom.ruleset");
+            PropertyMock customRuleSetProperty2 = CreateRuleSetProperty(this.projectMock, "config2", "Custom.ruleset");
+            PropertyMock defaultRuleSetProperty1 = CreateRuleSetProperty(this.projectMock, "config3", CSharpVBBindingOperation.DefaultProjectRuleSet);
+            PropertyMock defaultRuleSetProperty2 = CreateRuleSetProperty(this.projectMock, "config4", CSharpVBBindingOperation.DefaultProjectRuleSet);
             testSubject.Initialize();
 
             // Act
@@ -241,8 +241,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
             // Arrange
             CSharpVBBindingOperation testSubject = this.CreateTestSubject();
             this.projectMock.SetVBProjectKind();
-            PropertyMock customRuleSetProperty1 = CreateProperty(this.projectMock, "config1", "Custom.ruleset");
-            PropertyMock customRuleSetProperty2 = CreateProperty(this.projectMock, "config2", "Custom.ruleset");
+            PropertyMock customRuleSetProperty1 = CreateRuleSetProperty(this.projectMock, "config1", "Custom.ruleset");
+            PropertyMock customRuleSetProperty2 = CreateRuleSetProperty(this.projectMock, "config2", "Custom.ruleset");
             testSubject.Initialize();
 
             // Act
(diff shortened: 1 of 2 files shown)


src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -142,6 +142,39 @@ public void Commit()
...
+
+            if (hasUserSpecifiedRuleSet)
+            {
+                // We could've done this for all projects that have configurations that point to the same ruleset but we don't want to create noise for existing users.
+                return;
+            }
+
+            var ruleSetFullFilePath = propertyInformationMap.Values.FirstOrDefault()?.NewRuleSetFilePath;
+
+            if (!string.IsNullOrEmpty(ruleSetFullFilePath))
```
rita-g-sonarsource, line 166: This IF will never happen in reality, as VS will return the default value even if the project does not specify it. 
But it caused some tests to fail when they didn't set it, and it caused other tests to fail when I did set it... So it was easier to just add the IF :(

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
```diff
@@ -572,7 +633,7 @@ public void EnsureItemTypeIsCorrect_CriticalException_IsNotCaught()
 
         #region Helpers
 
-        private static PropertyMock CreateProperty(ProjectMock project, string configurationName, object propertyValue)
+        private static PropertyMock CreateRuleSetProperty(ProjectMock project, string configurationName, object propertyValue)
```
rita-g-sonarsource, line 636: Cosmetics for readability.

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
```diff
@@ -396,13 +396,74 @@ public void ProjectBindingOperation_Prepare_NoAdditionalFileConflict_NoException
             additionalFileConflictChecker.VerifyAll();
         }
 
+        [TestMethod]
```
rita-g-sonarsource, line 399: Because we use mocks of VS project, I'm limited with the cases I can test.
For example, I couldn't test "when project has no property specified" -- since VS will return a default value anyway, and our mocks will not.
I also couldn't test that if the user didn't specify a property, then VS will unite Debug/Release into a single non-conditional property.

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -142,6 +142,39 @@ public void Commit()
...
+        /// Try to set a ruleset without specifying a configuration (e.g. Debug / Release).
+        /// No changes are made if the project already has a ruleset.
+        /// </summary>
+        private void TrySetNonConditionalRuleSet()
+        {
+            var hasUserSpecifiedRuleSet = propertyInformationMap.Keys.Any(x => !IsDefaultMicrosoftRuleSet(x.Value as string));
+
+            if (hasUserSpecifiedRuleSet)
+            {
+                // We could've done this for all projects that have configurations that point to the same ruleset but we don't want to create noise for existing users.
```
duncanp-sonar, line 160: I don't understand this comment. We could have done what?

rita-g-sonarsource: Set a single non-conditional ruleset

src/Integration/Binding/CSharpVBBindingOperation.cs
duncanp-sonar: Strictly speaking the default value is being returned because of the contents of the standard MS MSBuild targets, not VS. So far those targets always provide a default.

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -142,6 +142,39 @@ public void Commit()
...
+            if (!string.IsNullOrEmpty(ruleSetFullFilePath))
+            {
+                var relativeRuleSetValue = PathHelper.CalculateRelativePath(ProjectFullPath, ruleSetFullFilePath);
+                var projectSystem = serviceProvider.GetService<IProjectSystemHelper>();
+                projectSystem.SetProjectProperty(initializedProject, Constants.CodeAnalysisRuleSetPropertyKey, relativeRuleSetValue);
+            }
+        }
+
+        /// <summary>
+        /// Set a ruleset for each project configuration (e.g. Debug / Release).
```
duncanp-sonar, line 175: This needs an additional comment explaining that it will be a no-op for new-style projects that already have an unconditional reference i.e. it's mainly to handle legacy cases of already-bound projects that have conditional ruleset references.

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -142,6 +142,39 @@ public void Commit()
...
+        {
+            var hasUserSpecifiedRuleSet = propertyInformationMap.Keys.Any(x => !IsDefaultMicrosoftRuleSet(x.Value as string));
+
+            if (hasUserSpecifiedRuleSet)
+            {
+                // We could've done this for all projects that have configurations that point to the same ruleset but we don't want to create noise for existing users.
+                return;
+            }
+
+            var ruleSetFullFilePath = propertyInformationMap.Values.FirstOrDefault()?.NewRuleSetFilePath;
```
duncanp-sonar, line 164: Fetching the ruleset value from the property map doesn't look right. If we get this far we know that all of the `CodeAnalysisRuleSet` properties (if any) refer to the default MS ruleset. So we know that we point to the solution-level ruleset i.e. `cSharpVBBindingConfig.RuleSet.Path` (which is what `propertyInformationMap.Values...NewRuleSetFilePath` should be returning anyway).

src/Integration/Binding/CSharpVBBindingOperation.cs
duncanp-sonar: ...although as commented above, we don't need to be accessing the property information map to get the ruleset path anyway.

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
duncanp-sonar: I agree we don't need to test case where VS unites the Debug/Release properties - we're only interested in testing that our code correctly updates properties for all of the current configurations.

I'm not sure about your first point, though - it's our mock, so we can change its behaviour as necessary. Having said that, I'm not clear which cases the mock won't handle. What can't be tested - the case when the property information map is empty?

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
```diff
@@ -396,13 +396,74 @@ public void ProjectBindingOperation_Prepare_NoAdditionalFileConflict_NoException
...
+
+            // Assert
+
+            var nonConditionalProperty = projectMock.GetBuildProperty(Constants.CodeAnalysisRuleSetPropertyKey);
+            nonConditionalProperty.Should().BeNullOrEmpty();
+
+            var solutionLevelRuleSet = PathHelper.CalculateRelativePath(testSubject.ProjectFullPath, cSharpVBBindingConfig.RuleSet.Path);
+            defaultRuleSet.Value.ToString().Should().Be(solutionLevelRuleSet);
+
+            var projectLevelRuleSet = Path.GetFileNameWithoutExtension(projectMock.FilePath) + ".Release.ruleset";
```
duncanp-sonar, line 456: I'm getting confused between the various changes and bug-fixes.
The test sets a non-default project-level ruleset for `Release` - is that correct?
If so, why are we expecting a new project-level ruleset to be created, rather than updating the existing one?

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
```diff
@@ -396,13 +396,74 @@ public void ProjectBindingOperation_Prepare_NoAdditionalFileConflict_NoException
...
+            {
+                testSubject.Commit();
+            }
+
+            // Assert
+
+            var nonConditionalProperty = projectMock.GetBuildProperty(Constants.CodeAnalysisRuleSetPropertyKey);
+            nonConditionalProperty.Should().BeNullOrEmpty();
+
+            var solutionLevelRuleSet = PathHelper.CalculateRelativePath(testSubject.ProjectFullPath, cSharpVBBindingConfig.RuleSet.Path);
```
duncanp-sonar, line 453: Minor - `generatedRuleSet` is more accurate than `solutionLevel` - it's clearer that it means "our" ruleset.
It's possible to have a solution-level ruleset that is not the generated ruleset, and we probably have some tests that have such a configuration.

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
duncanp-sonar: Or is this fixed by your other PR (#1436)?

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
rita-g-sonarsource: The case where we set no project properties, yet the property information map is not empty. But it's a minor case.

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
rita-g-sonarsource: Hmm, no, this is unfortunately looks like another bug in `CSharpVBBindingOperation`: 
https://github.com/SonarSource/sonarlint-visualstudio/blob/branch-furtherCM/src/Integration/Binding/CSharpVBBindingOperation.cs#L270

If the project references different rulesets, then we seem to wrap each ruleset with a new one :/
I'll open an issue for it.

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
duncanp-sonar: But the property information map will only contain a value because VS has supplied a default property, no? Which we can simulate by having the test create a property that returns the default value.

From a test point of view, the cases we might want to cover are:
1) property information map is empty
2) property information map has default value(s)
3) property information map ha non-default value(s)
Can we test all of those?

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
rita-g-sonarsource: Opened an issue: https://github.com/SonarSource/sonarlint-visualstudio/issues/1437

duncanp-sonar: Is this a new issue? If it's a regression we should fix it before shipping, but if it's not new then it isn't essential.

rita-g-sonarsource: Not a regression, but I think it's a minor fix. I'll take a look at it and see if it's complicated/risky.

src/Integration.UnitTests/Binding/CSharpVBBindingOperationTests.cs
rita-g-sonarsource: I tested 2+3, the first one is not realistic, as it will never be empty. So I think it's ok.

PR #1433, opened 2020-05-29
Removed VS RuleSet from ICSharpVBBindingConfig

duncanp-sonar:


src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
```diff
@@ -170,12 +170,12 @@ public async Task GetConfig_ReturnsCorrectRuleset()
...
 
             response.Should().BeAssignableTo<ICSharpVBBindingConfig>();
-            
+
             var actualRuleset = ((ICSharpVBBindingConfig)response).RuleSet;
             actualRuleset.Path.Should().Be(expectedRulesetFilePath);
             actualRuleset.Content.Description.Should().Be(validRuleSet.Description);
-            actualRuleset.Content.Rules.Should().AllBeEquivalentTo(validRuleSet.Rules);
-            actualRuleset.Content.ToolsVersion.ToString().Should().Be(validRuleSet.ToolsVersion);
+            actualRuleset.Content.Rules.Should().BeEquivalentTo(validRuleSet.Rules);
```
duncanp-sonar, line 177: The VS `RuleSet.Rules` and Core `RuleSet.Rules` are different types of collection so I had to use a different FluentAssertion.

PR #1436, opened 2020-06-03
Don't update ruleset declarations that already reference solution ruleset

### LINKED ISSUE #863: Existing rulesets are ignored when binding a solution
### Description

If an unbound project has a custom existing ruleset, it will be ignored when the project is bound.

### TODO
Decide on the preferred behaviour e.g.
* include the generated shared ruleset in the existing custom rules, or
* create a new project-level ruleset that includes the existing custom ruleset and the generated shared ruleset, or
* ignore the existing custom ruleset and reference the generated shared ruleset directly.
* do nothing - current behaviour is good enough most cases.

What if the custom ruleset file is used by multiple projects (not all of which may be bound)?

### Related information

* SonarLint for VS Version: to current (v4.8)
* Languages in your solution: C#, VB.Net

### Other tickets in the same area
* #817: [Connected Mode] Stop generating project-level ruleset

rita-g-sonarsource:
Fixes #863

### DIFF
--- a/src/Integration/Binding/IRuleSetReferenceChecker.cs
+++ b/src/Integration/Binding/IRuleSetReferenceChecker.cs
@@ -25,10 +25,15 @@ namespace SonarLint.VisualStudio.Integration.Binding
     internal interface IRuleSetReferenceChecker
     {
         /// <summary>
-        /// Return true if the project's ruleSets directly include the target ruleSet, otherwise false
+        /// Return true if the RuleSet declaration directly includes the target ruleSet, otherwise false
         /// </summary>
         /// <remarks>We don't currently check nested includes in RuleSet
         /// i.e. if Project has RuleSet A which includes RuleSet B that includes RuleSet C, IsReferenced(Project, C) returns false.</remarks>
-        bool IsReferenced(Project project, string targetRuleSetFilePath);
+        bool IsReferenced(Project project, RuleSetDeclaration declaration, string targetRuleSetFilePath);
+
+        /// <summary>
+        /// Return true if all project's ruleSets directly include the target ruleSet, otherwise false
+        /// </summary>
+        bool IsReferencedByAllDeclarations(Project project, string targetRuleSetFilePath);
     }
 }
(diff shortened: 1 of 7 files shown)


src/Integration/Binding/IRuleSetReferenceChecker.cs
```diff
@@ -25,10 +25,15 @@ namespace SonarLint.VisualStudio.Integration.Binding
     internal interface IRuleSetReferenceChecker
     {
         /// <summary>
-        /// Return true if the project's ruleSets directly include the target ruleSet, otherwise false
+        /// Return true if the RuleSet declaration directly includes the target ruleSet, otherwise false
         /// </summary>
         /// <remarks>We don't currently check nested includes in RuleSet
         /// i.e. if Project has RuleSet A which includes RuleSet B that includes RuleSet C, IsReferenced(Project, C) returns false.</remarks>
-        bool IsReferenced(Project project, string targetRuleSetFilePath);
+        bool IsReferenced(Project project, RuleSetDeclaration declaration, string targetRuleSetFilePath);
```
duncanp-sonar, line 32: I realise you've just made a previously private method public and added it to the interface, but the method signature doesn't look right - it shouldn't be necessary to provide both the `project` and a `ruleSetDeclaration` since they are related objects.

However, I think this is just exposing the underlying problem
`ISolutionRuleSetsInformationProvider.TryGetProjectRuleSetFilePath`, which has a very similar signature.

The `ruleSetDeclaration` actually contains enough information to carry out the check without the supplying the `project` too.

I think we should fix that problem separately.

rita-g-sonarsource: Yep, I see now that ruleSetDeclaration has RuleSetProjectFullName.
I will do it in a separate PR.

duncanp-sonar: There are a number of possible fixes:
1) remove the `project` parameter, and just pass in the `rulesetDeclaration` and target ruleset path.
2) change the `rulesetDeclaration` parameter to `sourceRuleSetFilePath`,
3) change the signature completely to be more general
e.g. `bool TryResolveRuleSetFilePath(string sourceRuleSetPath, out string fullFilePath, params string[] searchPaths)`
I'm not sure which option is best - it depends what makes sense on the interface i.e. logically what operation do we want to expose? (and it also depends on the amount of work involved).

rita-g-sonarsource: I think removing the project parameter is simplest:
https://github.com/SonarSource/sonarlint-visualstudio/pull/1438

PR #1443, opened 2020-06-04
Fixed bug in IsSameDay

duncanp-sonar:
Now takes the local time-zone into account

### DIFF
--- a/src/Integration/Helpers/DateTimeOffsetExtensions.cs
+++ b/src/Integration/Helpers/DateTimeOffsetExtensions.cs
@@ -24,9 +24,28 @@ namespace SonarLint.VisualStudio.Integration
 {
     public static class DateTimeOffsetExtensions
     {
-
-        public static bool IsSameDay(this DateTimeOffset date, DateTimeOffset other) =>
+        public static bool IsSameDayOld(this DateTimeOffset date, DateTimeOffset other) =>
             Math.Abs(date.Date.Subtract(other.Date).TotalDays) < 1;
+
+        /// <summary>
+        /// Returns true if the two date-times fall in the same calendar day in
+        /// the specified time-zone, otherwise false
+        /// </summary>
+        /// <param name="timeZoneInfo">(Optional) time-zone to used to determine the calendar day.
+        /// Defaults to the local time-zone.</param>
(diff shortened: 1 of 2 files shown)


src/Integration/Helpers/DateTimeOffsetExtensions.cs
```diff
@@ -24,10 +24,29 @@ namespace SonarLint.VisualStudio.Integration
...
+        public static bool IsSameDayOld(this DateTimeOffset date, DateTimeOffset other) =>
             Math.Abs(date.Date.Subtract(other.Date).TotalDays) < 1;
 
+        /// <summary>
+        /// Returns true if the two date-times fall in the same calendar day in
+        /// the specified time-zone, otherwise false
+        /// </summary>
+        /// <param name="timeZoneInfo">(Optional) time-zone to used to determine the calendar day.
+        /// Defaults to the local time-zone.</param>
+        public static bool IsSameDay(this DateTimeOffset date, DateTimeOffset other, TimeZoneInfo timeZoneInfo = null)
```
rita-g-sonarsource, line 36: I'm debating about this.
A) DateTimeOffset should already contain the timezone, so it's a bit weird to specify the timeZoneInfo argument.
I'd expect a comparison of just DateTimeOffset parameters.

B) This stackoverflow thread was interesting to read: https://stackoverflow.com/questions/27093065/datetimeoffset-same-day-comparison

In particular, they highlight that we should specify we we really want: 
Do we want to know if it's the same calendar day, or do we really want to know if it's been over 24 hours? I think what we're really interested in is if it's been over 24H. In which case, we can simplify it to: 
```
TimeSpan difference = dateTime1 - dateTime2;
var days = difference.TotalDays;
return days > 0;
```

WDYT?

duncanp-sonar: We want to know if it's the same calendar day, not whether two times fall within the same 24-hour period. I specifically checked that with the team and PM.

Both DateTimeOffsets have timezone information, but they could be different (and in fact in the case of the bug they are different). We could designate one of the supplied datetimeoffsets as being the "target" timezone, which would then put the onus on the caller to make sure that that datetimeoffset is for the correct "target" timezone , converting it if necessary.

Or we could not worry about how the supplied datetimeoffsets are formatted, and simply supply the timezone that should be used to determine what the relevant twenty-hour period is, which is what this code does.

PR #1444, opened 2020-06-05
Added abstraction over DateTimeOffset.Now and TimeZoneInfo.Local for …

duncanp-sonar:
…testing

### DIFF
--- a/src/Core.UnitTests/SystemAbstractions/TimerFactoryTests.cs
+++ b/src/Core.UnitTests/SystemAbstractions/TimerFactoryTests.cs
@@ -22,7 +22,7 @@ using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarLint.VisualStudio.Core.SystemAbstractions;
 
-namespace SonarLint.VisualStudio.Integration.UnitTests
+namespace SonarLint.VisualStudio.Core.UnitTests
 {
     [TestClass]
     public class TimerFactoryTests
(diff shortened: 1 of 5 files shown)


src/Core.UnitTests/SystemAbstractions/TimerFactoryTests.cs
```diff
@@ -22,7 +22,7 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarLint.VisualStudio.Core.SystemAbstractions;
 
-namespace SonarLint.VisualStudio.Integration.UnitTests
+namespace SonarLint.VisualStudio.Core.UnitTests
```
duncanp-sonar, line 25: Unrelated - moved existing TimeFactoryTests.cs from the integration test project to the core test project (should have been done previously when the product code was moved).

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -43,6 +44,7 @@ public class TelemetryManagerTests
         private Mock<ILogger> loggerMock;
         private Mock<ITelemetryTimer> telemetryTimerMock;
         private Mock<IKnownUIContexts> knownUIContexts;
+        private ICurrentTimeProvider currentTimeProvider = DefaultCurrentTimeProvider.Instance;
```
duncanp-sonar, line 47: The current TelemetryManager tests work ok without us needing to inject specific date times so I didn't try to change them to use a mock instead.

However, I will the mock when adding new tests to fix the "increment days of use" bug.

src/Integration/Telemetry/TelemetryManager.cs
```diff
@@ -154,9 +158,10 @@ public void Update()
             try
             {
                 var lastAnalysisDate = telemetryRepository.Data.LastSavedAnalysisDate;
-                if (!DateTimeOffset.Now.IsSameDay(lastAnalysisDate))
+                var now = currentTimeProvider.Now;
```
duncanp-sonar, line 161: This is technically a change of behaviour since we are now only calling `Now` once, but it's not material (and it we did `Now` twice it would make the tests slightly more complicated).

PR #1447, opened 2020-06-05
Patch up invalid days_of_use telemetry

### LINKED ISSUE #1440: Patch incorrect days_of_use stored on local machine
### Description

See #1434 for the underlying problem about values being calculated incorrectly.

A secondary issue is that user machines will have an incorrect `days_of_use` value persisted locally in the telemetry data file so those machines will continue to send misleading data.

Incorrect data is where the `days_of_use` value is greater than the `days_since_installation`.
We can't know what the true `days_of_use` value is, but we could clean up the data by capping the `days_of_use` value to be at most the `days_since_installation` value (or by resetting it to zero).

We could either try clean the data when we receive it, or clean it up on the local machine.
If we clean it up on receipt then the data will continue to be meaningless in some cases.

TODO: decide what clean-up to do, and where:
* no cleanup
* set to zero on client
* set to `min(days_of_use, days_since_installation`) on client
* set to a value less than `days_since_installation` on client e.g. 50%, 80% of `days_since_installation`
* set to `min(days_of_use, days_since_installation`) on receipt
* ignore impossible data on receipt

Discussed with PM. Decision:
* set to `min(days_of_use, days_since_installation`) on client

UPDATE:
The `days_since_installation` value is zero on day one, and the expected `days_of_use` would be one, so the maximum allowed value for `days_of_use` is actually `min(days_of_use, days_since_installation` + 1)

? add flag that data has been patched? DECISION: no. If we really needed to know this we could recalculate it by looking at historical data.

duncanp-sonar:
Fixes #1440

### DIFF
--- a/src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
+++ b/src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
@@ -405,9 +405,9 @@ namespace SonarLint.VisualStudio.Integration.Tests
             telemetryRepositoryMock.Verify(x => x.Save(), Times.Never);
         }
 
-        private TelemetryManager CreateManager() => new TelemetryManager(activeSolutionTrackerMock.Object,
+        private TelemetryManager CreateManager(ICurrentTimeProvider mockTimeProvider = null) => new TelemetryManager(activeSolutionTrackerMock.Object,
             telemetryRepositoryMock.Object, loggerMock.Object, telemetryClientMock.Object,
-            telemetryTimerMock.Object, knownUIContexts.Object, currentTimeProvider);
+            telemetryTimerMock.Object, knownUIContexts.Object, mockTimeProvider ?? currentTimeProvider);
 
         #region Languages analyzed tests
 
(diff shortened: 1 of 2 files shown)


src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -452,5 +452,160 @@ private static void CheckExpectedLanguages(TelemetryData data, params string[] e
         }
 
         #endregion
+
+        [TestMethod]
+        public void Sequence_InstallFollowedByMultipleUpdates()
```
duncanp-sonar, line 457: I'm not particularly keen on this test, but it wasn't obvious what the expected behaviour on the day of installation/day 2 should be so I wrote this test to describe the expected behaviour.

The key takeaway is that `days_of_use` can be at most one greater than `days_since_installation`, since on the day of installation `days_since_installation` is zero and `day_of_use` is one.

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -452,5 +452,160 @@ private static void CheckExpectedLanguages(TelemetryData data, params string[] e
...
+                telemetryRepositoryMock.Setup(x => x.Data).Returns(telemetryData);
+            }
+        }
+
+        [TestMethod]
+        [DataRow("2020-05-31T20:30:00.000+00:00", "2020-06-01T06:30:00.000+08:00", 8, false)] // regression test for #1434
+        [DataRow("2020-06-01T04:30:00.000+00:00", "2020-06-01T06:30:00.000+00:00", 0, false)] // same day -> don't update
+        [DataRow("2020-05-31T20:30:00.000+00:00", "2020-06-01T06:30:00.000+00:00", 0, true)] // different day -> update
+        [DataRow("2020-06-01T04:30:00.000+00:00", "2020-08-23T02:30:00.000+08:00", 8, true)] // many days later -> update
+        public void Update_IfNotSameDayAsLastUpdate_DaysOfUseIsIncremented(string lastSavedDate, string now, int utcOffset, bool expectedIncrement)
```
duncanp-sonar, line 522: Again, this test isn't really related to the new fix. It's filling in a hole in the existing test coverage.

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -452,5 +452,160 @@ private static void CheckExpectedLanguages(TelemetryData data, params string[] e
...
+        [DataRow(0, 1, 1)] // second and subsequent check on day of installation -> can't be incremented above 1
+
+        [DataRow(4, 4, 5)] // used every day since installation, already checked and incremented on day 4 -> incremented
+        [DataRow(4, 5, 5)] // used every day since installation, second and subsequent checks on day 4 -> not capped or incremented
+
+        [DataRow(1001, 100, 101)] // not used every day -> incremented and saved
+
+        [DataRow(1, 100, 2)] // bad data - should be capped at (days since installation + 1)
+        [DataRow(999, 1001, 1000)] // bad data - should be capped at (days since installation + 1)
+        public void Update_DaysOfUseIsCorrectlyCalculated(int daysSinceInstallation, int daysOfUse,
```
duncanp-sonar, line 574: This is the new test for the bug fix. The `days_of_use` can be incremented at to at most `days_since_installation`.
If it's greater than that value then it's capped.

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -452,5 +452,157 @@ private static void CheckExpectedLanguages(TelemetryData data, params string[] e
...
+                telemetryRepositoryMock.Setup(x => x.Data).Returns(telemetryData);
+            }
+        }
+
+        [TestMethod]
+        [DataRow("2020-05-31T20:30:00.000+00:00", "2020-06-01T06:30:00.000+08:00", 8, false)] // regression test for #1434
+        [DataRow("2020-06-01T04:30:00.000+00:00", "2020-06-01T06:30:00.000+00:00", 0, false)] // same day -> don't update
+        [DataRow("2020-05-31T20:30:00.000+00:00", "2020-06-01T06:30:00.000+00:00", 0, true)] // different day -> update
+        [DataRow("2020-06-01T04:30:00.000+00:00", "2020-08-23T02:30:00.000+08:00", 8, true)] // many days later -> update
+        public void Update_IfNotSameDayAsLastUpdate_DaysOfUseIsIncremented(string lastSavedDate, string now, int utcOffset, bool expectedIncrement)
```
rita-g-sonarsource, line 522: As I mentioned in a different PR, I think this test really highlights the fact that "utcOffset" parameter is odd. The timezone should be taken from the `now` parameter, as it won't be possible for the `now` to be +08:00, but the utcOffset to be different than 8.
I think the whole calculation code could've been simplified without the additional time zone parameter. 
I wouldn't fix it now, but If we touch the code again I think it's worth considering to simplify it.

PR #1448, opened 2020-06-08
Bug fix: don't set non-conditional ruleset when existing ruleset already references the generated ruleset

rita-g-sonarsource:
Caused by #863 and #805

### DIFF
--- a/src/Integration/Binding/CSharpVBBindingOperation.cs
+++ b/src/Integration/Binding/CSharpVBBindingOperation.cs
@@ -153,7 +153,9 @@ namespace SonarLint.VisualStudio.Integration.Binding
         /// </summary>
         private void TrySetNonConditionalRuleSet()
         {
-            var hasUserSpecifiedRuleSet = propertyInformationMap.Keys.Any(x => !IsDefaultMicrosoftRuleSet(x.Value as string));
+            var solutionRuleSetProvider = serviceProvider.GetService<ISolutionRuleSetsInformationProvider>();
+            var ruleSetsInfo = solutionRuleSetProvider.GetProjectRuleSetsDeclarations(initializedProject).ToArray();
+            var hasUserSpecifiedRuleSet = ruleSetsInfo.Any(x => !IsDefaultMicrosoftRuleSet(x.RuleSetPath));
 
             if (hasUserSpecifiedRuleSet)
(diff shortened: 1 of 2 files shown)


src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -153,7 +153,9 @@ private void AddRuleset()
         /// </summary>
         private void TrySetNonConditionalRuleSet()
         {
-            var hasUserSpecifiedRuleSet = propertyInformationMap.Keys.Any(x => !IsDefaultMicrosoftRuleSet(x.Value as string));
```
rita-g-sonarsource, line 156: The bug was that in the `CalculateRuleSetInformation` method I ignore all rulesets that already reference the generated ruleset.
So propertyInformationMap was empty.

src/Integration/Binding/CSharpVBBindingOperation.cs
```diff
@@ -153,7 +153,9 @@ private void AddRuleset()
...
+            var solutionRuleSetProvider = serviceProvider.GetService<ISolutionRuleSetsInformationProvider>();
+            var ruleSetsInfo = solutionRuleSetProvider.GetProjectRuleSetsDeclarations(initializedProject).ToArray();
+            var hasUserSpecifiedRuleSet = ruleSetsInfo.Any(x => !IsDefaultMicrosoftRuleSet(x.RuleSetPath));
```
duncanp-sonar, line 158: This doesn't feel quite right: there's quite an overlap between this processing and what happens in `CalculateRuleSetInformation`.
`CalculateRuleSetInformation` which has already fetched the `rulesetsInfo` so calculating the `hasUserSpecifiedRuleSet` there would be a one-liner.Also, the naming of `CalculateRuleSetInformation` suggests it's intended to be the place were the calculations are done.

rita-g-sonarsource: So you're suggesting to extract `hasUserSpecifiedRuleSet` to a member?

duncanp-sonar: Yep. Logically it's the same kind of data as `propertyInformationMap`.

Issue #1450, opened 2020-06-08
Incorrect warning about "maximum API response limit reached" for all binding updates


duncanp-sonar:
### Description

The warning message `WARNING: Sonar web maximum API response limit reached.` is incorrectly displayed for each quality profile that is downloaded or updated during binding.

The binding/update works correctly, but the message is misleading and is likely to result in support questions.

The message is new in the as-yet-unreleased v4.22.

### Repro steps

1. Bind a solution to SQ/SC, or update an existing binding, 

### Expected behavior

No warning.

### Actual behavior

Misleading warnings about download limits reached when fetching individual quality profiles:

![image](https://user-images.githubusercontent.com/10757559/84057252-52199280-a9af-11ea-85a9-08d297184808.png)

### Known workarounds

None.

### Related information

* SonarLint for VS Version: unreleased v4.22
* Languages in your solution: C#, VB.Net. Probably occurs for C and C++ too.

PR #1451, opened 2020-06-08
Updated web client to get fix for #1450

duncanp-sonar:


duncanp-sonar:
Updating to the latest commit has pulled in a couple of other changes in the web client. However, they were only changes to tests - no product changes.

PR #1454, opened 2020-06-09
Update our binding and add SonarLint.xml

rita-g-sonarsource:


.sonarlint/sonarlint-visualstudio/CSharp/SonarLint.xml
```diff
@@ -0,0 +1,89 @@
+<?xml version="1.0" encoding="utf-8"?>
```
rita-g-sonarsource, line 1: The file is different from the previous SonarLint.xml, notably the license tag is not there anymore. However, I think that's ok, since I develop without our extension installed and thus don't see it anyway...

PR #1457, opened 2020-06-10
[Infra] Added build-time check to catch changes to assembly refs (v2)

duncanp-sonar:


src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -385,4 +385,53 @@
...
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
+
+  <!-- Target to help debugging the CI build - dumps report differences to the console. 
+       Note: there will always be differences in the reports since the timestamps will be
+       different for each run.
+  -->
+  <Target Name="DumpOutputAsmRef" AfterTargets="_GenerateAsmRefReportOnBuild" BeforeTargets="_PostGenerateAndRefProcessing">
```
duncanp-sonar, line 432: @rita-g-sonarsource I copied and tweak your debug task to dump the differences to the console so they appear in the Azure Pipeline log.

rita-g-sonarsource: Great! I was modifying the pipeline in Azure, but I think this is much cleaner (and we will see it when working locally as well)

PR #1458, opened 2020-06-11
Tweaked asmref check to use single baseline file

duncanp-sonar:


src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -410,10 +410,11 @@
...
     <AsmRefTargetVersionCompatibility>Strict</AsmRefTargetVersionCompatibility>
-    
-    <!-- Use the baseline file containing public key tokens if the assemblies are being strong-named -->
-    <AsmRefBaselineFilePath Condition=" $(SignAssembly) == 'true' ">$(MSBuildProjectDirectory)\AsmRef_$(MSBuildProjectName)_Baseline_Signed.txt</AsmRefBaselineFilePath>
-    <AsmRefBaselineFilePath Condition=" $(SignAssembly) == 'false' ">$(MSBuildProjectDirectory)\AsmRef_$(MSBuildProjectName)_Baseline.txt</AsmRefBaselineFilePath>
+
+    <!-- If we're strong-naming the binaries (i.e. on the CI machine) then do check strong-naming of our
+         assemblies. -->
+    <AsmRefIgnoreSourcePublicKeyToken Condition=" $(SignAssembly) == 'true' ">false</AsmRefIgnoreSourcePublicKeyToken>
+    <AsmRefIgnoreSourcePublicKeyToken Condition=" $(SignAssembly) == 'false' ">true</AsmRefIgnoreSourcePublicKeyToken>
```
rita-g-sonarsource, line 417: Do we really care about the strong naming of our own assemblies?
Perhaps we can do AsmRefIgnoreSourcePublicKeyToken=true in all cases, and just handle ALM

duncanp-sonar: As discussed offline, we do currently strong-name our assemblies as part of the CI process, and this checks that that has been done. 
If we want to stop strong-naming our assemblies I think we would tackle that as a separate topic.

PR #1460, opened 2020-06-11
Fixed local build failure caused by incorrect asmref comparison

duncanp-sonar:
Reverted to two asmref basline files, one with strong names, one without.

However, updating the baseline without strong names will auto-update the baseline with strong names.


duncanp-sonar:
I don't want to do any complex processing on the CI build - it's the safety net, so it should be as vanilla as possible.

I could have stuck with a single file and done more complex processing on the client e.g.
* on every build, patch the baseline file before the comparison is done. However, the baseline file is under source control, so we'd have to save the patched up version in e.g. the bin folder.
* on every build, patch up the latest report file after it has been created but before the comparison is done. That would involve hooking in to the internal processing of the asmrefs targets, which isn't ideal.

The approach I went for has two files, but neither is updated on every build i.e. there's less processing and less to go wrong. The only time the files change is when the baseline needs to be updated; the asmref package will update one, and we'll update the other automatically.

PR #68, opened 2016-04-25
[SVS-14] Edit the SonarQube exclusion MSBuild properties through a nicer UI in VS

mjcheetham:
Added context menu containing latched project options for `<SonarQubeTestProject/>` and `<SonarQubeExclude/>`. Will work on multiple project selections. Only works with `Language.IsSupported` languages.

nested_menu

PR #1463, opened 2020-06-11
Fixed project refs to Alm.Auth package

duncanp-sonar:
No code changes.

* moved reference to shared props file
* fixed package reference

### DIFF
--- a/src/Integration.TeamExplorer.UnitTests/Integration.TeamExplorer.UnitTests.csproj
+++ b/src/Integration.TeamExplorer.UnitTests/Integration.TeamExplorer.UnitTests.csproj
@@ -1,9 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup Label="Specify which VS-specific refs are required">
+  <PropertyGroup Label="Specify which external/VS-specific refs are required">
     <IsTestProject>true</IsTestProject>
     <RequiresTeamFoundation>true</RequiresTeamFoundation>
     <RequiresRuleset>true</RequiresRuleset>
+    <RequiresAlmAuthentication>true</RequiresAlmAuthentication>
   </PropertyGroup>
   <Import Project="..\SonarLint.VSSpecificAssemblies.props" />
 
@@ -36,18 +37,6 @@
     <PackageReference Include="NuGet.Core" Version="2.12.0" />
     <PackageReference Include="NuGet.VisualStudio" Version="3.3.0" />
   </ItemGroup>
-
-  
-  <ItemGroup Label="Alm authentication">
-    <!-- The package from MS isn't signed so we need to reference the StrongName package
-         and conditionally set the DisableStrongName properties -->
-    <PackageReference Include="Microsoft.Alm.Authentication" Version="4.0.0.1" />
-    <PackageReference Include="StrongNamer" Version="0.0.8" />
-  </ItemGroup>
-  <PropertyGroup>
-      <DisableStrongNamer Condition="'$(SignAssembly)' != 'true'">true</DisableStrongNamer>
-  </PropertyGroup>
-
   
   <ItemGroup Label="Test framework">
     <PackageReference Include="MSTest.TestAdapter" Version="1.4.0" />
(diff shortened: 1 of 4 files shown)


src/Integration/Integration.csproj
```diff
@@ -100,21 +88,6 @@
...
-  <Target Name="HackIncludeALMAuthInVsix" AfterTargets="ResolveReferences" BeforeTargets="FindReferenceAssembliesForReferences">
-    <ItemGroup>
-      <ReferencePath Condition="'%(FileName)' == 'Microsoft.Alm.Authentication' or '%(FileName)' == 'Microsoft.Alm.Git' or '%(FileName)' == 'Microsoft.Vsts.Authentication'">
-        <CopyLocal>true</CopyLocal>
-      </ReferencePath>
-    </ItemGroup>
-
-    <Message Importance="high" Text="In HackIncludeAlmAuthInVsix" />
-    <Message Importance="high" Text="Reference paths: @(ReferencePath)" />
-  </Target>
```
duncanp-sonar, line 117: This hack was only necessary because the package reference in line #68 in this file specified `IncludeAssets="compile"` i.e. the assets should only be available at compile time, and not copied to the output folder. The hack was then effectively making sure they did get copied to the output folder, and also included in the VSIX.

However, just removing the `InlcudeAssets` attribute would have been enough.
I've checked that the three Alm.* assemblies are still included in the VSIX.

rita-g-sonarsource:
@duncanp-sonar don't we still need parts of the hack? especially `<CopyLocal>true</CopyLocal>` so it would be included in the VSIX?

src/Integration.TeamExplorer.UnitTests/Integration.TeamExplorer.UnitTests.csproj
```diff
@@ -36,18 +37,6 @@
...
-  <ItemGroup Label="Alm authentication">
-    <!-- The package from MS isn't signed so we need to reference the StrongName package
-         and conditionally set the DisableStrongName properties -->
-    <PackageReference Include="Microsoft.Alm.Authentication" Version="4.0.0.1" />
-    <PackageReference Include="StrongNamer" Version="0.0.8" />
-  </ItemGroup>
-  <PropertyGroup>
-      <DisableStrongNamer Condition="'$(SignAssembly)' != 'true'">true</DisableStrongNamer>
-  </PropertyGroup>
-
```
duncanp-sonar, line 50: Moved this repeating block to a shared props file.

src/Integration/Integration.csproj
```diff
@@ -59,19 +60,6 @@
     </EmbeddedResource>
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="StrongNamer" Version="0.0.8" />
-  </ItemGroup>
-  <ItemGroup Label="Alm authentication">
-    <!-- The package from MS isn't signed so we need to reference the StrongName package
-         and conditionally set the DisableStrongName properties -->
-    <PackageReference Include="Microsoft.Alm.Authentication" Version="4.0.0.1" IncludeAssets="compile" />
```
duncanp-sonar, line 68: For some reason this reference was set the `IncludeAssets` property, whereas references in the test projects didn't.
However, it isn't required.

src/Integration/Integration.csproj
```diff
@@ -97,13 +85,13 @@
...
+         EmbedInteropAssemblies isn't set for PackagReferences. -->
     <PackageReference Include="Microsoft.VisualStudio.SDK.EmbedInteropTypes" Version="15.0.27" />
   </ItemGroup>
 
-  <!-- The Alm.Auth assemblies aren't being included in the VSIX for some reason.
-       This workaround adds metadata to three artifacts so that the standard VSSDK targets will
-       include them.
+  <!-- The Alm.Auth assemblies aren't being included in the VSIX for some reason.	
+       This workaround adds metadata to three artifacts so that the standard VSSDK targets will	
+       include them.
```
duncanp-sonar, line 94: I thought I could get rid of this hack, but unfortunately not.

PR #1468, opened 2020-06-15
Moved error list code to a separate subfolder

duncanp-sonar:


duncanp-sonar:
Minimal product code changes: moved the error-list related functionality out of the `TaggerProvider` and added some of the missing tests.
I fully expect that we will have to change the new `ISonarErrorListDataSource` interface in the current sprint. This PR is intended to make those changes easier to do.

src/Integration.Vsix/ErrorList/SinkManager.cs
```diff
@@ -25,10 +25,11 @@
 namespace SonarLint.VisualStudio.Integration.Vsix
 {
     // Interface introduced to simplify testing.
-    internal interface ISinkManagerRegister
```
duncanp-sonar, line 28: I moved the `ISinkManagerRegister` interface to the new `SonarErrorLIstDataSource` class and added `ISinkManager`.

src/Integration.Vsix/ErrorList/SinkManager.cs
```diff
@@ -67,12 +68,12 @@ public void Dispose()
             sinkRegister = null;
         }
 
-        public void AddFactory(SnapshotFactory factory)
+        public void AddFactory(ITableEntriesSnapshotFactory factory)
```
duncanp-sonar, line 71: Miscellaneous: changed the type to be the VS interface rather than our implementation of it.

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -0,0 +1,139 @@
...
+{
+    // Interface introduced to simplify testing.
+    internal interface ISinkManagerRegister
+    {
+        void AddSinkManager(ISinkManager manager);
+        void RemoveSinkManager(ISinkManager manager);
+    }
+
+    [Export(typeof(ISonarErrorListDataSource))]
+    internal class SonarErrorListDataSource : ITableDataSource, ISonarErrorListDataSource, ISinkManagerRegister
```
duncanp-sonar, line 37: New class containing code extracted from the `TaggerProvider`.

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                managers.Remove(manager);
+            }
+        }
+
+        #endregion
+
+        #region ISonarErrorListDataSource implementation
+
+        public void RefreshErrorList()
```
duncanp-sonar, line 100: This method tells the error list to update the issues for all of the files we've analysed. We should be able to tell the error list which factory (i.e. file with issues) has changed. But I didn't want to change that as part of this refactoring.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -207,62 +181,27 @@ public void RequestAnalysis(string path, string charset, IEnumerable<AnalysisLan
...
-            lock (managers)
-            {
-                managers.Remove(manager);
-            }
-        }
-
         public void AddIssueTracker(IIssueTracker issueTracker)
         {
-            lock (managers)
+            lock (issueTrackers)
```
duncanp-sonar, line 186: This class no longer manages the sink managers, so it now locks on the issue trackers list instead.

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        void AddSinkManager(ISinkManager manager);
+        void RemoveSinkManager(ISinkManager manager);
+    }
+
+    [Export(typeof(ISonarErrorListDataSource))]
+    internal class SonarErrorListDataSource : ITableDataSource, ISonarErrorListDataSource, ISinkManagerRegister
+    {
+        private readonly ISet<ISinkManager> managers = new HashSet<ISinkManager>();
+        private readonly ISet<ITableEntriesSnapshotFactory> factories = new HashSet<ITableEntriesSnapshotFactory>();
```
duncanp-sonar, line 40: Minor changes in lines 39 and 40 to the code extracted from the `TaggerProvider`: the sets use interfaces rather than classes.

duncanp-sonar:
Minimal product code changes: moved the error-list related functionality out of the `TaggerProvider` and added some of the missing tests.
I fully expect that we will have to change the new `ISonarErrorListDataSource` interface in the current sprint. This PR is intended to make those changes easier to do.

Relates to #1469

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
```diff
@@ -195,61 +192,13 @@ public void CreateTagger_tracker_should_be_distinct_per_file()
...
-            var trackers = provider.ActiveTrackersForTesting.ToArray();
-            trackers.Count().Should().Be(2); // sanity check
-
-            var mockTableDataSink = new Mock<ITableDataSink>();
-            provider.Subscribe(mockTableDataSink.Object);
 
-            // factories of existing trackers are propagated to new sink manager
-            mockTableDataSink.Verify(s => s.AddFactory(trackers[0].Factory, false));
-            mockTableDataSink.Verify(s => s.AddFactory(trackers[1].Factory, false));
+            mockSonarErrorDataSource.Verify(x => x.AddFactory(trackers[0].Factory), Times.Once);
```
rita-g-sonarsource, line 201: Nice, much cleaner

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            var mockFactory = new Mock<ITableEntriesSnapshotFactory>();
+
+            testSubject.AddFactory(mockFactory.Object);
+
+            CheckFactoryWasAdded(mockManager1, mockFactory.Object);
+            CheckFactoryWasAdded(mockManager2, mockFactory.Object);
+        }
+
+        [TestMethod]
+        public void RemoveSinkManager_ManagerIsNotLongerNotified()
```
rita-g-sonarsource, line 144: Minor: NoLonger

rita-g-sonarsource, line 144: General design question about RemoveSinkManager, and also about RemoveFactory: I see that they are all called only as part of Dispose of IssueTagger.
So I assume that in the end, when VS decided that the document is closed, it disposes ITagger, which triggers the Remove chain, right?

In this case, maybe in the future design changes, it would be useful to just change the classes to implement IDisposable, and test only that Dispose is clearing everything correctly.

duncanp-sonar: Not quite. There is one `IssueTracker` per file/document, but each `IssueTracker` will have multiple `IssueTagger` instances (one per view). The `IssueTracker` doesn't call `RemoveIssueTracker` until it has no more `IssueTagger` instances (i.e. the last view on the document is closed).

`RemoveSinkManager` is not called when a tagger is disposed. It's called when the sink manager is disposed, is done by VS to tell us that the `ITableDataSink` is no longer valid/available (which in the case of the error list probably means VS is being shut down).

We should be able to simplify the code, though.

Issue #1055, opened 2019-10-28
VS2015: open json files cause C++ analysis to fail


duncanp-sonar:
### Description

`Microsoft.VisualStudio.JSON.Package.ErrorSources.JSONTableDataSink` throws an exception when the change notification from the SonarLint implementation of `ITableDataSource` is handled by the MS `JSONTableDataSink`.

A class that implements `ITableDataSink` should handle notifications from any `ITableDataSource`. It seems that `JSONTableDataSink` doesn't. The bug seems to have been fixed in VS2017+.

### Repro steps

1. Open a project containing a C++ and a json file.
2. Open and view both documents (to make sure the json sink handler is initialized)
3. Switch to the C++ document
4. Start typing

### Expected behavior

No error and C/C++ squigglies are updated correctly.

### Actual behavior

Error dialogue appears, with more information in the SonarLint output window.
Squigglies are not displayed correctly for C/C++ issues.

![image](https://user-images.githubusercontent.com/10757559/67701977-d97a3e80-f9a8-11e9-923c-fd164e540503.png)

```
Daemon error occurred in editor integration: System.ArgumentNullException: Value cannot be null.
Parameter name: key
   at System.Collections.Concurrent.ConcurrentDictionary`2.TryRemove(TKey key, TValue& value)
   at Microsoft.VisualStudio.JSON.Package.ErrorSources.JSONTableDataSink.RemoveFactory(ITableEntriesSnapshotFactory oldFactory)
   at Microsoft.VisualStudio.JSON.Package.ErrorSources.JSONTableDataSink.FactorySnapshotChanged(ITableEntriesSnapshotFactory factory)
   at SonarLint.VisualStudio.Integration.Vsix.TaggerProvider.UpdateAllSinks()
   at SonarLint.VisualStudio.Integration.Vsix.TextBufferIssueTracker.SnapToNewSnapshot(IssuesSnapshot newIssues)
   at SonarLint.VisualStudio.Integration.Vsix.TextBufferIssueTracker.SafeOnBufferChange(Object sender, TextContentChangedEventArgs e)
```

### Known workarounds

None (other than closing all json files before editing any C/C++ files)

### Related information

* SonarLint for VS Version: v4.x
* Visual Studio Version: only VS2015. Not an issue in VS2017 or VS2019.
* Languages in your solution: C, C++

PR #1472, opened 2020-06-16
Removed SinkManager and associated interfaces

duncanp-sonar:
Simplified the error list integration code: removed `SinkManager` and the related interfaces.


src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -101,39 +92,55 @@ public void RefreshErrorList()
...
                 {
-                    manager.RemoveFactory(factory);
+                    SafeOperation(sink, "RemoveFactory", () => sink.RemoveFactory(factory));
                 }
             }
         }
 
         #endregion ISonarErrorListDataSource implementation
+
+        private void SafeOperation(ITableDataSink sink, string operationName, Action op)
```
duncanp-sonar, line 130: Copied from `SinkManager`.

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -21,22 +21,16 @@
...
-        void AddSinkManager(ISinkManager manager);
-        void RemoveSinkManager(ISinkManager manager);
-    }
-
     [Export(typeof(ISonarErrorListDataSource))]
-    internal class SonarErrorListDataSource : ITableDataSource, ISonarErrorListDataSource, ISinkManagerRegister
+    internal class SonarErrorListDataSource : ITableDataSource, ISonarErrorListDataSource
     {
-        private readonly ISet<ISinkManager> managers = new HashSet<ISinkManager>();
+        private readonly ISet<ITableDataSink> sinks = new HashSet<ITableDataSink>();
```
duncanp-sonar, line 33: Store the sinks directly, rather than storing wrappers around the sinks.

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -101,39 +92,55 @@ public void RefreshErrorList()
...
             // Mark all the sinks as dirty (so, as a side-effect, they will start an update pass that will get the new snapshot
             // from the snapshot factories).
-            lock (managers)
+            lock (sinks)
             {
-                foreach (var manager in managers)
+                foreach (var sink in sinks)
                 {
-                    manager.UpdateSink();
+                    SafeOperation(sink, "FactorySnapshotChanged", () => sink.FactorySnapshotChanged(null));
```
duncanp-sonar, line 99: We added this error handling in the past to work round a problem with a bad sink (see #1055).
The error handling wasn't previously in the `Subscribe` or `Dispose` methods so I haven't added it to them.

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -66,30 +60,27 @@ internal SonarErrorListDataSource(ITableManagerProvider tableManagerProvider)
...
+                sinks.Add(sink);
 
-                foreach (var factory in factories)
+                foreach(var factory in factories)
                 {
-                    manager.AddFactory(factory);
+                    sink.AddFactory(factory);
                 }
+
+                return new ExecuteOnDispose(() => Unsubscribe(sink));
```
duncanp-sonar, line 74: We need to hand VS a disposable object that it will dispose when the sink is no longer available.
We just need to remove our reference to the sink when this happens. That is essentially the only reason the `SinkManager` class was required.
I've replaced it with a general purpose disposable class that can be passed a delegate.

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
```diff
@@ -127,69 +129,169 @@ public void AddFactory_ExistingSinkManagersAreNotifiedOfNewFactory()
...
 
             testSubject.RemoveFactory(mockFactory.Object);
 
-            CheckFactoryWasRemoved(mockManager1, mockFactory.Object);
-            CheckFactoryWasRemoved(mockManager2, mockFactory.Object);
+            CheckFactoryWasRemoved(mockSink1, mockFactory.Object);
+            CheckFactoryWasRemoved(mockSink2, mockFactory.Object);
+        }
+
+        #region Exception handling tests
```
duncanp-sonar, line 185: The changes above this line are just fix-up so the tests work against `ITableDataSink` rather than `ISinkManager`.

The tests below this line are the exception handling tests copied from the `SinkManagerTests` and cleaned up slightly.  There are tests for each of `AddFactory`, `RemoveFactory` and `RefreshErrorList` to check that critical errors are not suppressed but non-critical errors are.

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
rita-g-sonarsource: That's a neat pattern, I like it!

PR #1475, opened 2020-06-17
Removed VS types from IAnalyzer interface

duncanp-sonar:
The `ProjectItem` parameter is only required by the CLangAnalyzer, and shouldn't be on the analyzer interface. I've refactored the `CLangAnalyzer` to fetch the `ProjectItem` instead.

The changes to all of the other files are just to remove the redundant `ProjectItem` parameter.

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
@@ -18,9 +18,13 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
+using System.Threading;
+using EnvDTE;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.CFamily;
 using SonarLint.VisualStudio.Integration.UnitTests;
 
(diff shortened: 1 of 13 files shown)


src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -58,28 +63,40 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
         public void ExecuteAnalysis(string path, string charset, IEnumerable<AnalysisLanguage> detectedLanguages,
-            IIssueConsumer consumer, ProjectItem projectItem, IAnalyzerOptions analyzerOptions,
+            IIssueConsumer consumer, IAnalyzerOptions analyzerOptions,
             CancellationToken cancellationToken)
         {
+            var projectItem = dte?.Solution?.FindProjectItem(path);
+            if (projectItem == null)
+            {
+                return;
+            }
```
duncanp-sonar, line 73: This is all that's required to fetch the `ProjectItem`, and this is the only place it's required.

The VS threading analyzers complain that this object should only be accessed on the UI thread. When manually testing I discovered that this is not always the case.
However, this isn't a new issue: this class was already accessing the object on the UI thread in the previous version of the code. I've created an issue: #1476.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -58,28 +63,40 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
                 return;
             }
 
+            TriggerAnalysis(request, consumer, cancellationToken);
+        }
+
+        protected /* for testing */ virtual Request CreateRequest(ILogger logger, ProjectItem projectItem, string absoluteFilePath, ICFamilyRulesConfigProvider cFamilyRulesConfigProvider, IAnalyzerOptions analyzerOptions) =>
+            CFamilyHelper.CreateRequest(logger, projectItem, absoluteFilePath, cFamilyRulesConfigProvider, analyzerOptions);
+
+        protected /* for testing */ virtual void TriggerAnalysis(Request request, IIssueConsumer consumer, CancellationToken cancellationToken) =>
```
duncanp-sonar, line 89: This class was virtually untested and was untestable in its previous state. I've added these two override points to allow tests to change the behaviour. I wanted to limit the scope of this change, so I didn't want to create any new interfaces or types at this point.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -260,7 +260,7 @@ private void SnapToNewSnapshot(IssuesSnapshot newIssues)
 
         public void RequestAnalysis(IAnalyzerOptions options)
         {
-            Provider.RequestAnalysis(FilePath, charset, detectedLanguages, this, ProjectItem, options);
```
duncanp-sonar, line 263: The `TextBufferIssueTracker` currently still needs the ProjectItem itself (although it should now be easier to refactor this class so that it doesn't).

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
rita-g-sonarsource: Is it possible that it also somehow caused the goldbar?

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -29,21 +33,110 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
...
+            mockDTE.Setup(d => d.Solution).Returns(solution);
+            var dte = mockDTE.Object;
+
+            var mockServiceProvider = new Mock<IServiceProvider>();
+            mockServiceProvider.Setup(s => s.GetService(typeof(DTE))).Returns(dte);
+
+            return mockServiceProvider;
+        }
+
+        private class TestableCLangAnalyzer : CLangAnalyzer
```
rita-g-sonarsource, line 119: I understand why you did it this way, and I saw the comment that you wanted to avoid creating new interfaces.

I think it's worth creating those interfaces.
In general, I'm not a fan of the internal methods that we expose only for testing. It disrupts code flow, and almost always ends with flow paths that are never tested, since we are laser-focused on testing just the internal method and not the whole class flow. It always means that there is a problem with the responsibility of the class, which should be resolved by breaking it down to smaller components.

In this particular case, an interface would be much simpler to read, test and maintain, and could easily be mocked, instead of inheriting from it.

duncanp-sonar: Let's discuss this offline. I don't think it's worth spending the time on now; it wasn't the purpose of this PR. This class needs more substantial refactoring to be properly testable, and probably to fix the threading issue too.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
duncanp-sonar: It might, but I think it's more likely to cause a crash/deadlock.

duncanp-sonar:
As agreed offline, the `CLangAnalyzer` class needs further refactoring but it's not in the scope of this change.

PR #1477, opened 2020-06-17
CLangAnalyzer integration tests

rita-g-sonarsource:


src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -404,7 +404,7 @@
   -->
   <PropertyGroup>
     <!-- Only check the VS2019 build -->
-    <AsmRefDisableBaselining Condition=" $(VersionSpecificSuffix) != '2019' ">true</AsmRefDisableBaselining>
+    <AsmRefDisableBaselining>true</AsmRefDisableBaselining>
```
duncanp-sonar, line 407: This doesn't look like it belongs as part of this PR.

src/Integration.Vsix/CFamily/PortedFromJava/Analyzer.cs
```diff
@@ -97,8 +97,8 @@ internal class Message : MessagePart
         public bool PartsMakeFlow { get; }
         public MessagePart[] Parts { get; }
 
-        public Message(string ruleKey, string filename, int line, int column, int endLine, int endColumn, string message, bool partsMakeFlow, MessagePart[] parts)
-            : base(filename, line, column, endLine, endColumn, message)
+        public Message(string ruleKey, string filename, int line, int column, int endLine, int endColumn, string text, bool partsMakeFlow, MessagePart[] parts)
+            : base(filename, line, column, endLine, endColumn, text)
```
duncanp-sonar, line 101: FYI in general, we try to avoid making gratuitous changes to the files that have been ported directly from Java to make it easier to compare our version against the Java version. See the ReadMe.txt in the `PortedfromJava` folder: https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/CFamily/PortedFromJava/ReadMe.txt

As it is, it looks like the Java version has been updated more recently, although a quick look at their changes suggests it's only in the formatting.

src/Integration.Vsix.UnitTests/CFamily/IntegrationTests/CFamily_CLangAnalyzer_IntegrationTests.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
```
duncanp-sonar, line 1: Do we have a rule that could remind us to add copyright notices? ;-)

src/Integration.Vsix.UnitTests/Integration.Vsix.UnitTests.csproj
```diff
@@ -28,6 +28,43 @@
...
+    <None Remove="CFamily\IntegrationTests\CLangAnalyzerTestFile_NoIssues_EmptyFile.txt" />
+    <None Remove="CFamily\IntegrationTests\CLangAnalyzerTestFile_NoIssues_EmptyFile_response.json" />
+    <None Remove="CFamily\IntegrationTests\CLangAnalyzerTestFile_OneIssue.txt" />
+    <None Remove="CFamily\IntegrationTests\CLangAnalyzerTestFile_OneIssue_response.json" />
+    <None Remove="CFamily\IntegrationTests\CLangAnalyzerTestFile_TwoIssues.txt" />
+    <None Remove="CFamily\IntegrationTests\CLangAnalyzerTestFile_TwoIssues_response.json" />
+  </ItemGroup>
+  
+  <ItemGroup>
+    <Content Include="CFamily\IntegrationTests\CLangAnalyzerTestFile_NoIssues_EmptyFile_response.json">
```
duncanp-sonar, line 42: Do you need to change the ItemGroup type from `None` to `Content`, or do you just need to add the metadata property `CopyToOutputDirectory`?
If you just need to update the metadata you can use `Update` instead `Remove` + `Include` in SDK-style projects: https://docs.microsoft.com/en-us/visualstudio/msbuild/item-element-msbuild?view=vs-2019#attributes-and-elements

src/Integration.Vsix.UnitTests/CFamily/IntegrationTests/CFamily_CLangAnalyzer_IntegrationTests.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    [TestClass]
+    public class CFamily_CLangAnalyzer_IntegrationTests
+    {
+        private string testsDataDirectory;
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            testsDataDirectory = Path.Combine(Environment.CurrentDirectory, "CFamily\\IntegrationTests\\");
```
duncanp-sonar, line 19: The other tests that create directories create them under `TestContext.DeploymentDirectory`. I thought there was a helper function that did this, but I can't find it. Most of the tests use a method like this:

```
        private string CreateTestSpecificDirectory()
        {
            var dir = Path.Combine(TestContext.DeploymentDirectory, TestContext.TestName);
            Directory.CreateDirectory(dir);
            return dir;
        }
```

src/Integration.Vsix.UnitTests/CFamily/IntegrationTests/CFamily_CLangAnalyzer_IntegrationTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            var request = GetRequest(testedFile);
+            var expectedResponse = GetExpectedResponse(testCaseFileName, testedFile);
+
+            var response = InvokeAnalyzer(request);
+            response.Should().BeEquivalentTo(expectedResponse);
+        }
+
+        private Request GetRequest(string testedFile)
+        {
+            var requestJson = File.ReadAllText(testsDataDirectory + "CLangAnalyzerRequestTemplate.json");
```
duncanp-sonar, line 39: Minor: `Path.Combine`

src/Integration.Vsix.UnitTests/Integration.Vsix.UnitTests.csproj
```diff
@@ -28,6 +28,43 @@
     </Content>
   </ItemGroup>
   
+  <ItemGroup>
+    <None Remove="CFamily\CLangAnalyzerRequest.json" />
```
duncanp-sonar, line 32: Was there are reason for the request file not to be in the `IntegrationTests` folder with the the other resource files?

src/Integration.Vsix.UnitTests/CFamily/IntegrationTests/CFamily_CLangAnalyzer_IntegrationTests.cs
duncanp-sonar, line 19: Relying on `Environment.CurrentDirectory` is quite fragile and makes a lot of assumptions about the execution environment.

A less fragile option (i.e. one that makes fewer assumptions) is to look for the files relative to the current assembly  e.g.
```
            var resourcesPath = Path.Combine(
                Path.GetDirectoryName(typeof(MyTestClass).Assembly.Location), "TestResources");
``` 
(adapted from [here](https://github.com/SonarSource/sonarlint-visualstudio/blob/558296b44d059b55830e22a0553c8613f038d4a0/src/Core.UnitTests/CFamily/CFamilySonarWayRulesConfigProviderTests.cs#L92))

The most robust option is to embed the resources in the assembly and unpack them when needed, but that's more work so I'd only normally do that for production code.

src/Integration.Vsix/CFamily/PortedFromJava/Analyzer.cs
rita-g-sonarsource: I made this change because the (response object)[https://github.com/SonarSource/sonarlint-visualstudio/blob/8e01992209d7e136191f64fece181aaff65ee948/src/Integration.Vsix/CFamily/PortedFromJava/Analyzer.cs#L80] contains a "Text" field, but it's not being filled when deserializing because the ctor's argument is named "message".
I can undo this change and instead change the response jsons to have a "Message" field, but it feels awkward. 
Another option is to remove it completely and ignore it in the comparison. Wdyt?

src/Integration.Vsix.UnitTests/CFamily/IntegrationTests/CFamily_CLangAnalyzer_IntegrationTests.cs
rita-g-sonarsource: If only!

src/Integration.Vsix.UnitTests/Integration.Vsix.UnitTests.csproj
rita-g-sonarsource: Hmm that's a mistake, the file was renamed to CLangAnalyzerRequestTemplate and moved to the correct folder, must be a left-over.

src/Integration.Vsix/CFamily/PortedFromJava/Analyzer.cs
duncanp-sonar: Ah. That's an interesting one. The class wasn't designed to be json-serializable and doesn't need to be from a product point of view, but we're now using json-serialization for convenience in our tests.  Hhhmm.

We shouldn't change the public API of the class, since we want it to match the CFamily version.

The risk is we update the class in the future to match any changes made on the CFamily side, and start getting hard-to-diagnose test failures.

As a minimum, I'd add comment to class saying we want it to be json-serializable to simplify testing.
I'd also add a comment to that constructor saying the parameter names have to match the property names for the default json serializer to work.

The next step beyond that would be to write a test that checks it is actually json-serialization-round-trippable, but that might overkill.

Is there anything else we could do to make such failures easy to diagnose?

(for future reference, this is definitely the sort of thing that's worth calling out in a review comment)

rita-g-sonarsource: If the parameter names don't match, the deserialization would not work and the fields would be null. That's the only indicator unfortunately. Perhaps if we encounter this issue in the future then we could invest more into this: e.g., maybe add a TestResponse converts to Response. Or, maybe not use jsons and instead add code functions that return the expected object.

I added a comment for now, let's revise it in the future if it breaks, if it's ok with you.

PR #1481, opened 2020-06-19
Remove double quotations from additional options

rita-g-sonarsource:
@duncanp-sonar recently a user complained about this issue https://community.sonarsource.com/t/failed-to-execute-analysis-sonarlint-visual-studio-2017/26386

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
@@ -229,16 +229,65 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         }
 
         [TestMethod]
-        public void AdditionalOptions()
+        [DynamicData(nameof(AdditionalOptionsTestCases))]
+        public void AdditionalOptions(dynamic testCase)
         {
-            CFamilyHelper.FileConfig.GetAdditionalOptions("/D test").Should().BeEquivalentTo("/D", "test");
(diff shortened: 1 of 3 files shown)


src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -264,6 +264,15 @@ public Request ToRequest(string path)
                 }
             }
 
+            internal /* for testing */ static string[] GetAdditionalOptions(string options)
+            {
+                var additionalOptions = Regex.Split(options, "(?<=^[^\"]*(?:\"[^\"]*\"[^\"]*)*) (?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)");
```
rita-g-sonarsource, line 269: Courtesy of https://stackoverflow.com/a/4780801/339720

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -228,6 +228,19 @@ public void PlatformName()
                 .StartWith("Unsupported PlatformName: foo");
         }
 
+        [TestMethod]
+        public void AdditionalOptions()
```
duncanp-sonar, line 232: Why not use a data-driven test?

src/Integration.Vsix/CFamily/FileConfig.cs
duncanp-sonar: Can we use a compiled regex to avoid the code of creating it each time?

PR #1482, opened 2020-06-19
Fix comment

duncanp-sonar:


src/Integration.Vsix/CFamily/PortedFromJava/Analyzer.cs
```diff
@@ -60,7 +60,6 @@ internal partial class Request
         public string File { get; set; } = "";
     }
 
-    [Serializable]
```
duncanp-sonar, line 63: This attribute relates to the framework binary serializer, not the Newtonsoft serializer. It suggests we use binary serialization, which we don't.

src/Integration.Vsix/CFamily/PortedFromJava/Analyzer.cs
```diff
@@ -91,14 +90,14 @@ public MessagePart(string filename, int line, int column, int endLine, int endCo
...
     internal class Message : MessagePart
     {
         public string RuleKey { get; }
         public bool PartsMakeFlow { get; }
         public MessagePart[] Parts { get; }
 
-        // we expect the class to be serializable for the ease of testing, and therefore the ctor parameter names must match the property names
+        // SLVS: we expect the class to be JSON-serializable for the ease of testing, and therefore the ctor parameter names must match the property names
```
duncanp-sonar, line 99: Changed to mention the type of particular type of serialization we use.

src/Integration.Vsix/CFamily/PortedFromJava/Analyzer.cs
rita-g-sonarsource: Thanks for the fix 👍

PR #1484, opened 2020-06-19
Moved and renamed tagger-specific types

duncanp-sonar:
No code changes

* Moved tagger-specific types to Tagger subfolder
* Renamed types

### DIFF
--- a/src/Integration.Vsix/SonarLintTagger/IIssueMarkerFactory.cs
+++ b/src/Integration.Vsix/SonarLintTagger/IIssueMarkerFactory.cs
@@ -21,10 +21,10 @@
 using Microsoft.VisualStudio.Text;
 using SonarLint.VisualStudio.Core.Analysis;
 
-namespace SonarLint.VisualStudio.Integration.Vsix.Helpers
+namespace SonarLint.VisualStudio.Integration.Vsix
 {
-    internal interface IIssueConverter
+    internal interface IIssueMarkerFactory
     {
-        IssueMarker ToMarker(IAnalysisIssue issue, ITextSnapshot currentSnapshot);
+        IssueMarker Create(IAnalysisIssue issue, ITextSnapshot currentSnapshot);
     }
 }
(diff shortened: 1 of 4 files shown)


src/Integration.Vsix/SonarLintTagger/IssueMarkerFactory.cs
```diff
@@ -22,11 +22,11 @@
 using Microsoft.VisualStudio.Text;
 using SonarLint.VisualStudio.Core.Analysis;
 
-namespace SonarLint.VisualStudio.Integration.Vsix.Helpers
+namespace SonarLint.VisualStudio.Integration.Vsix
 {
-    internal class IssueConverter : IIssueConverter
+    internal class IssueMarkerFactory : IIssueMarkerFactory
```
duncanp-sonar, line 27: Feel free to suggest an alternative name. I'm not particularly happy with the new name, but it's less misleading than the old one.

src/Integration.Vsix/SonarLintTagger/IIssueMarkerFactory.cs
```diff
@@ -21,10 +21,10 @@
 using Microsoft.VisualStudio.Text;
 using SonarLint.VisualStudio.Core.Analysis;
 
-namespace SonarLint.VisualStudio.Integration.Vsix.Helpers
+namespace SonarLint.VisualStudio.Integration.Vsix
```
duncanp-sonar, line 24: These files were in the `Helper` subfolder, but they're only used by the tagger.

src/Integration.Vsix/SonarLintTagger/IssueMarkerFactory.cs
rita-g-sonarsource: Hm, that's tough. The class is responsible for calculating the span, and then returning an object containing the issue+span, right?
Maybe it should do just the calculation, and become `IIssueSpanCalculator`, and have `TextBufferIssueTracker` call it and create IssueMarker. But I'm not sure it's really worth the change.

PR #1488, opened 2020-06-22
Added comments and design notes about doc editing

duncanp-sonar:
Only comment changes.

Added comments related to tagging, the error list, and handling document updates about things that are not obvious from the current state of the code.

### DIFF
--- a/src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
+++ b/src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
@@ -84,10 +84,14 @@ namespace SonarLint.VisualStudio.Integration.Vsix
                     return true;
 
                 case StandardTableKeyNames.Line:
+                    // Note: the line and column numbers are taken from the SnapshotSpan, not the Issue.
+                    // The SnapshotSpan represents the live document, so the text positions could have
+                    // changed from those reported from the Issue.
                     content = this.issueMarkers[index].Span.Start.GetContainingLine().LineNumber;
                     return true;
 
                 case StandardTableKeyNames.Column:
+                    // Use the span, not the issue. See comment immediately above.
                     var position = this.issueMarkers[index].Span.Start;
                     var line = position.GetContainingLine();
                     content = position.Position - line.Start.Position;
--- a/src/Integration.Vsix/SonarLintTagger/README.md
+++ b/src/Integration.Vsix/SonarLintTagger/README.md
@@ -82,3 +82,47 @@ Error list integration:
 - `SinkManager` gets updates through factories.
 
 
+
+Tracking document edits:
+------------------------
+See MS docs "Inside the editor": https://docs.microsoft.com/en-us/visualstudio/extensibility/inside-the-editor?view=vs-2019
+
+A text file opened in the VS Editor is represented as an ITextBuffer. The buffer can be thought of as series of
+immutable ITextSnapshots.  A region of text in a ITextSnapshot is represented by a SnapshotSpan.
+Each time the buffer is edited, a new ITextSnapshot is created. The buffer tracks the changes between snapshots,
+making it possible to translate from a SnapshotSpan in the previous snapshot to the corresponding SnapshotSpan
+in the new ITextSnapshot e.g.
+* lines are added at the beginning of the buffer -> all old spans can be translate to spans in the new snapshot.
+* lines are added at the end of the buffer -> all old spans can be translated to snaps in the new snapshot.
+   The old spans still need to be translated, even though the positions of the span haven't changed.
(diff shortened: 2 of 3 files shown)


src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -274,6 +278,8 @@ void IIssueConsumer.Accept(string path, IEnumerable<IAnalysisIssue> issues)
 
             var filteredIssues = RemoveSuppressedIssues(issues);
 
+            // See bug #1487: we should be creating the issue markers using the text snapshot
```
duncanp-sonar, line 281: This is a newly-created bug for a pre-existing issue. I've added it to the current sprint since (1) we're going to have to change this code anyway and (2) the new UX will increase the interval during which the bug could appear since the background analysis will run for longer.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -84,10 +84,14 @@ public override bool TryGetValue(int index, string keyName, out object content)
...
 
                 case StandardTableKeyNames.Line:
+                    // Note: the line and column numbers are taken from the SnapshotSpan, not the Issue.
+                    // The SnapshotSpan represents the live document, so the text positions could have
+                    // changed from those reported from the Issue.
                     content = this.issueMarkers[index].Span.Start.GetContainingLine().LineNumber;
                     return true;
 
                 case StandardTableKeyNames.Column:
+                    // Use the span, not the issue. See comment immediately above.
```
rita-g-sonarsource, line 94: It might be an overkill, but maybe it's worth creating a function for `this.issueMarkers[index].Span.Start.GetContainingLine()` and add a proper doc comment on it.

src/Integration.Vsix/SonarLintTagger/README.md
```diff
@@ -82,3 +82,47 @@ Error list integration:
 - `SinkManager` gets updates through factories.
 
 
+
+Tracking document edits:
+------------------------
+See MS docs "Inside the editor": https://docs.microsoft.com/en-us/visualstudio/extensibility/inside-the-editor?view=vs-2019
+
+A text file opened in the VS Editor is represented as an ITextBuffer. The buffer can be thought of as series of
+immutable ITextSnapshots.  A region of text in a ITextSnapshot is represented by a SnapshotSpan.
```
rita-g-sonarsource, line 91: So SnapshotSpan is also immutable, right? (since the whole ITextSnapshot is thrown away when the file is edited)

src/Integration.Vsix/SonarLintTagger/README.md
```diff
@@ -82,3 +82,47 @@ Error list integration:
...
+in the new ITextSnapshot e.g.
+* lines are added at the beginning of the buffer -> all old spans can be translate to spans in the new snapshot.
+* lines are added at the end of the buffer -> all old spans can be translated to snaps in the new snapshot.
+   The old spans still need to be translated, even though the positions of the span haven't changed.
+* a line is deleted from the buffer -> any spans that included positions in the deleted line cannot be translated
+   to spans in the new snapshot. All other old spans can be translated to new spans.
+ 
+A SnapshotSpan can be translated across multiple snapshot versions e.g. from snapshot v1 to snapshot v3.
+
+Implementation note: the VS Editor also supports "tracking spans" that work across snapshots. The Roslyn tagger
```
rita-g-sonarsource, line 103: Hm I'm not sure what you mean by tracking spans. Do you think it's worth elaborating, or is it irrelevant?

src/Integration.Vsix/SonarLintTagger/README.md
duncanp-sonar: I'm not sure if the span is immutable but it is tied to a specific snapshot. The snapshot isn't thrown away when the buffer is edited. It's still valid, and represents the state of the document at a point in time (it will presumably be garbage collected if it's not being referenced).

A span represents a range of characters in the buffer. There can be multiple spans associated with a snapshot.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
duncanp-sonar: It's a thought. I'm not going to make any code changes in this PR though.

src/Integration.Vsix/SonarLintTagger/README.md
duncanp-sonar: I've updated the comment. Tracking spans automatically adjust as the snapshot changes so there is no need to manually translate SnapshotSpans between snapshots.

PR #1490, opened 2020-06-23
Fix double quoted compilation flags

### LINKED ISSUE #1426: Double quoted compilation flag passed through additional options are not parsed correctly
SLVS crashes due to the fact that double quoted additional options are not parsed correctly. "Additional options" are used to define compilation options. often /D option is used with double quotation: https://docs.microsoft.com/en-us/cpp/build/reference/d-preprocessor-definitions?view=vs-2019

**Workaround**
Remove the quotations from "additional options".

**Describe the solution you'd like**
Support all sort of additional options quotation patterns.

The crash impact all sort of projects and it is very easy to reproduce by adding valid option in quotation mark:
image

image

Interesting "Q" in FileConfig.cs:

```
                // TODO Q: what if it contains space in double quotes?
                AddRange(c.Cmd, AdditionalOptions.Split(' '));
```

rita-g-sonarsource:
Fixes #1426

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
@@ -230,6 +230,73 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         }
 
         [TestMethod]
+        [DynamicData(nameof(AdditionalOptionsTestCases))]
+        public void AdditionalOptions(dynamic testCase)
+        {
+            string optionsString = testCase.optionsString;
+            string[] expectedOptions = testCase.expectedOptions;
+            CFamilyHelper.FileConfig.GetAdditionalOptions(optionsString).Should().BeEquivalentTo(expectedOptions);
+        }
+
+        public static IEnumerable<object[]> AdditionalOptionsTestCases
+        {
+            get
+            {
+                return new[]
+                {
+                    new object[] {new {optionsString = "/arch:\"IA32\"", expectedOptions = new[] { "/arch:\"IA32\"" } }},
+                    new object[] {new {optionsString = "/D A", expectedOptions = new[] {"/D", "A"}}},
+                    new object[] {new {optionsString = "/D \"A\"", expectedOptions = new[] {"/D", "\"A\"" } }},
+                    new object[] {new {optionsString = "/D \"A= str\"", expectedOptions = new[] {"/D", "\"A= str\"" } }},
+                    new object[] {new {optionsString = "/U \"A\"", expectedOptions = new[] {"/U", "\"A\"" } }},
+                    new object[] {new {optionsString = "/U \" A\"", expectedOptions = new[] {"/U", "\" A\"" } }},
+                    new object[] {new {optionsString = "/D \"A# str\"", expectedOptions = new[] {"/D", "\"A# str\"" } }},
+                    new object[] {new {optionsString = "/FI \"C:\\Repos\\a.h\"", expectedOptions = new[] {"/FI", "\"C:\\Repos\\a.h\"" } }},
+                    new object[]
+                    {
+                        new
+                        {
+                            optionsString = "/D \"my test\" /D test",
+                            expectedOptions = new[] {"/D", "\"my test\"", "/D", "test"}
+                        }
+                    },
+                    new object[]
+                    {
+                        new
+                        {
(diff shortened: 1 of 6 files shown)


src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs
```diff
@@ -260,7 +273,8 @@ public static Request ToRequest(CFamilyHelper.Capture[] captures)
                     }
                     else
                     {
-                        string file = Absolute(capture.Cwd, arg);
+                        var relativeOrAbsolutePath = arg.Replace("\"", "");
+                        string file = Absolute(capture.Cwd, relativeOrAbsolutePath);
```
abbas-sabra-sonarsource, line 277: If you look at the definition of absolute, there are asserts and sometimes it throws. I think it would be better if we catch the exception and log and the exact user option that lead to this exception( it will help us in debugging) and then we simply skip the option instead of letting the exception propagate(this what MSVC does when it faces invalid additional options it just ignores it)

src/Integration.Vsix/CFamily/PortedFromJava/Protocol.cs
```diff
@@ -177,6 +178,12 @@ public static Response Read(BinaryReader reader, string issueFilePath = null)
                   throw new InvalidDataException("Communication issue with the C/C++ analyzer");
               case "message":
                   var message = readMessage(reader);
+
+                  if ("ParsingError".Equals(message.RuleKey))
```
abbas-sabra-sonarsource, line 182: What is the reason for this change? "ParsingError" is a valid issue and should be displayed to the user when enabled in addition to the other issues.

rita-g-sonarsource: My bad, I didn't see them and I thought that it wasn't handled. But it's due to the fact that they had no file location, so they just failed like all issues without location.
I'll revert this change.

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -229,6 +229,73 @@ public void PlatformName()
...
+                        new
+                        {
+                            optionsString = "/D \"my test\" /D \"my test\" /D test",
+                            expectedOptions = new[] {"/D", "\"my test\"", "/D", "\"my test\"", "/D", "test"}
+                        }
+                    }
+                };
+            }
+        }
+
```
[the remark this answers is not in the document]
rita-g-sonarsource: :( 
What do you think would make this better?

src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs
rita-g-sonarsource: @abbas-sabra-sonarsource Absolute is called in other places as well, did you mean to add the try-catch only in case of this "else" block, or in all the cases?

abbas-sabra-sonarsource: It's better to do it for all 3 calls since all of them can suffer from the same problem. 
It might be a better idea to do it inside the function `Absolute` and then print the "file path" that leads to this exception and return null(that is ignored in the three cases)

src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs
```diff
@@ -273,7 +269,7 @@ public static Request ToRequest(CFamilyHelper.Capture[] captures)
                     }
                     else
                     {
-                        var relativeOrAbsolutePath = arg.Replace("\"", "");
+                        var relativeOrAbsolutePath = RemoveDoubleQuotes(arg);
```
abbas-sabra-sonarsource, line 272: you should do the same `RemoveDoubleQuotes ` on  `dirStr` before calling absolute in case of the `/I` option

src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs
rita-g-sonarsource: Great, thanks 👍

abbas-sabra-sonarsource: Just double-checked the 3rd call should be safer since it cannot be added from additional options(it is coming from env).

PR #1493, opened 2020-06-23
Use IO streams to communicate with subprocess.exe #1466

### LINKED ISSUE #1466: Use IO streams to communicate with subprocess.exe
### Description
Callers can now communicate with `subprocess.exe` either using the existing file-based mechanism, or via standard IO streams.

SLVS should change to using the IO streams. 

As a first step, only the communication mechanism should be changed i.e. SLVS should not report any issues until the all issues have been received.

rita-g-sonarsource:
Fixes #1466

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/ProcessRunnerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/ProcessRunnerTests.cs
@@ -174,6 +174,7 @@ $@"waitfor /t 2 {Guid.NewGuid():N}
             // Arrange
             var logger = new TestLogger();
             var runner = CreateProcessRunner(logger);
+            var output = "";
 
             try
             {
@@ -196,7 +197,11 @@ $@"waitfor /t 2 {Guid.NewGuid():N}
 
                 var args = new ProcessRunnerArguments(exeName, true)
                 {
-                    EnvironmentVariables = envVariables
+                    EnvironmentVariables = envVariables,
+                    HandleOutputStream = reader =>
+                    {
+                        output = reader.ReadToEnd();
+                    }
                 };
 
                 // Act
@@ -213,9 +218,9 @@ $@"waitfor /t 2 {Guid.NewGuid():N}
             }
 
             // Check the child process used expected values
-            logger.AssertMessageLogged("file: machine override");
-            logger.AssertMessageLogged("file: process override");
-            logger.AssertMessageLogged("file: user override");
+            output.Should().Contain("file: machine override");
+            output.Should().Contain("file: process override");
+            output.Should().Contain("file: user override");
 
             // Check the runner reported it was overwriting existing variables
             // Note: the existing non-process values won't be visible to the child process
--- a/src/Integration.Vsix.UnitTests/CFamily/ProcessRunnerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/ProcessRunnerTests.cs
@@ -1,631 +1,631 @@
-﻿/*
- * SonarLint for Visual Studio
- * Copyright (C) 2016-2020 SonarSource SA
- * mailto:info AT sonarsource DOT com
- *
- * This program is free software; you can redistribute it and/or
- * modify it under the terms of the GNU Lesser General Public
- * License as published by the Free Software Foundation; either
- * version 3 of the License, or (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
- * Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public License
- * along with this program; if not, write to the Free Software Foundation,
- * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
- */
-
-// Note: copied from the S4MSB
-// https://github.com/SonarSource/sonar-scanner-msbuild/blob/5c23a7da9171e90a1970a31507dce3da3e8ee094/Tests/SonarScanner.MSBuild.Common.UnitTests/ProcessRunnerTests.cs#L32
-
-using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.IO;
-using System.Linq;
-using System.Text;
-using System.Threading;
-using System.Threading.Tasks;
-using FluentAssertions;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
-using SonarLint.VisualStudio.Integration.UnitTests;
-
-namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
-{
-    [TestClass]
-    public class ProcessRunnerTests
-    {
-        public TestContext TestContext { get; set; }
-
-        [TestMethod]
-        public void Execute_WhenRunnerArgsIsNull_ThrowsArgumentNullException()
-        {
-            // Arrange
-            Action action = () => new ProcessRunner(new ConfigurableSonarLintSettings(), new TestLogger()).Execute(null);
-
-            // Act & Assert
-            action.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("runnerArgs");
-        }
-
-        [TestMethod]
-        public void ProcRunner_ExecutionFailed()
-        {
-            // Arrange
-            var exeName = WriteBatchFileForTest(TestContext, "exit -2");
-
-            var logger = new TestLogger();
-            var args = new ProcessRunnerArguments(exeName, true);
-            var runner = CreateProcessRunner(logger);
-
-            // Act
-            runner.Execute(args);
-
-            // Assert
-            runner.ExitCode.Should().Be(-2, "Unexpected exit code");
-        }
-
-        [TestMethod]
-        public void ProcRunner_ExecutionSucceeded()
-        {
-            // Arrange
-            var exeName = WriteBatchFileForTest(TestContext,
-@"@echo Hello world
-@echo xxx yyy
-@echo Testing 1,2,3...>&2
-");
-
-            var logger = new TestLogger();
-            var args = new ProcessRunnerArguments(exeName, true);
-            var output = "";
-            args.HandleOutputStream = reader => output = reader.ReadToEnd();
-            var runner = CreateProcessRunner(logger);
-
-            // Act
-            runner.Execute(args);
-
-            // Assert
-            runner.ExitCode.Should().Be(0, "Unexpected exit code");
-
-            output.Should().Contain("Hello world");
-            output.Should().NotContain("Testing 1,2,3...");
-        }
-
-        [TestMethod]
-        public void ProcRunner_FailsOnTimeout()
-        {
-            // Arrange
-
-            // Calling TIMEOUT can fail on some OSes (e.g. Windows 7) with the error
-            // "Input redirection is not supported, exiting the process immediately."
-            // Alternatives such as
-            // pinging a non-existent address with a timeout were not reliable.
-            var exeName = WriteBatchFileForTest(TestContext,
-$@"waitfor /t 2 {Guid.NewGuid():N}
-@echo Hello world
-");
-
-            var logger = new TestLogger();
-            var args = new ProcessRunnerArguments(exeName, true)
-            {
-                TimeoutInMilliseconds = 100
-            };
-            var output = "";
-            args.HandleOutputStream = reader => output = reader.ReadToEnd();
-            var runner = CreateProcessRunner(logger);
-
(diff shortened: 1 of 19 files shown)


src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -117,60 +117,56 @@ private static Request CreateRequest(FileConfig fileConfig, string absoluteFileP
...
-                // Create a FileInfo object to set the file's attributes
-                FileInfo fileInfo = new FileInfo(tempFileName);
-
-                // Set the Attribute property of this file to Temporary. 
-                // Although this is not completely necessary, the .NET Framework is able 
-                // to optimize the use of Temporary files by keeping them cached in memory.
-                fileInfo.Attributes = FileAttributes.Temporary;
-
-                using (var writeStream = new FileStream(tempFileName, FileMode.Open))
+                if (analyzerExeFilePath == null)
```
rita-g-sonarsource, line 124: This used to be below, in ExecuteAnalysis method which no longer exists

src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -117,60 +117,56 @@ private static Request CreateRequest(FileConfig fileConfig, string absoluteFileP
...
+                        }
+                    },
+                    HandleOutputStream = reader =>
                     {
-                        var response = Protocol.Read(new BinaryReader(readStream), request.File);
-                        return response;
+                        if ((request.Flags & Request.CreateReproducer) != 0)
+                        {
+                            // When running with reproducer flag, we don't want to show analysis results.
+                            // todo: no need to actually wait for results, just need to know if the reproducer file has been created
```
rita-g-sonarsource, line 150: Need to ask the CFamily guys if it's possible not to run the analysis at all

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -52,7 +54,7 @@ public ProcessRunner(ISonarLintSettings settings, ILogger logger)
         /// Runs the specified executable and returns a boolean indicating success or failure
         /// </summary>
         /// <remarks>The standard and error output will be streamed to the logger. Child processes do not inherit the env variables from the parent automatically</remarks>
-        public bool Execute(ProcessRunnerArguments runnerArgs)
+        public void Execute(ProcessRunnerArguments runnerArgs)
```
rita-g-sonarsource, line 57: I'm not sure about changing it to void.
We would no longer check for it anyway (except in tests), since we would handle the issues before the process ends.

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -116,66 +120,87 @@ public bool Execute(ProcessRunnerArguments runnerArgs)
...
+                        runnerArgs.WorkingDirectory,
+                        runnerArgs.TimeoutInMilliseconds,
+                        process.Id);
+
+                    try
+                    {
+                        runnerArgs.HandleInputStream(process.StandardInput);
+                        runnerArgs.HandleOutputStream(process.StandardOutput);
+                    }
+                    catch (Exception) when (isRunningProcessCancelled)
```
rita-g-sonarsource, line 145: Without the catch, when an analysis was cancelled it would show an ugly stack trace to the user coming from Protocol.Read.

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -116,66 +120,87 @@ public bool Execute(ProcessRunnerArguments runnerArgs)
...
-
-            if (succeeded)
-            {
-                process.WaitForExit(); // Give any asynchronous events the chance to complete
-            }
-
-            // false means we asked the process to stop but it didn't.
-            // true: we might still have timed out, but the process ended when we asked it to
-            if (succeeded)
+            try
```
rita-g-sonarsource, line 159: Method unchanged, except BeginOutputReadLine and LogDebug were extracted.

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -116,66 +120,87 @@ public bool Execute(ProcessRunnerArguments runnerArgs)
...
+                        LogWarning(CFamilyStrings.WARN_ExecutionTimedOutNotKilled, runnerArgs.TimeoutInMilliseconds,
+                            runnerArgs.ExeName);
+                    }
                 }
-            }
 
-            return succeeded && ExitCode == 0;
+                return succeeded && ExitCode == 0;
+            }
+            catch (InvalidOperationException ex) when (ex.Message == "No process is associated with this object.")
```
rita-g-sonarsource, line 199: Since it runs on a different task, when we cancel the process, an exception is thrown since the process is disposed (but it's not ObjectDisposedException... it's InvalidOperationException with that message).

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -116,66 +120,87 @@ public bool Execute(ProcessRunnerArguments runnerArgs)
...
                     {
                         LogMessage(CFamilyStrings.MSG_ExecutionCancelled);
-                        return false;
+                        return;
                     }
                 }
 
-                var result = WaitForProcessToFinish(process, runnerArgs);
-                return result;
+                using (var timeoutTask = Task.Factory.StartNew(() => WaitForProcessToFinish(process, runnerArgs)))
```
rita-g-sonarsource, line 127: What should I use instead of StartNew? :/ 
I can't use `await TaskScheduler.Default` since we're already on a background thread :/

src/Integration.Vsix.UnitTests/CFamily/ProcessRunnerTests.cs
```diff
@@ -423,60 +424,47 @@ xxx yyy
...
-            success.Should().BeFalse();
             runner.ExitCode.Should().Be(0, "Unexpected exit code");
 
-            logger.AssertOutputStringDoesNotExist("Hello world"); 
+            output.Should().NotContain("Hello world");
         }
 
-        [Ignore] // Flaky https://github.com/SonarSource/sonarlint-visualstudio/issues/1330
         [TestMethod]
         public void Execute_CancellationTokenCancelledMidway_ProcessKilled()
```
rita-g-sonarsource, line 439: It doesn't rely on files anymore, so hopefully it won't flake now :)

src/Integration.Vsix/CFamily/IProcessRunner.cs
```diff
@@ -25,6 +25,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
 {
     public interface IProcessRunner
     {
-        bool Execute(ProcessRunnerArguments runnerArgs);
+        void Execute(ProcessRunnerArguments runnerArgs);
```
duncanp-sonar, line 28: There's a comment in line 21 saying this was copied from the Scanner for MSBuild, which suggests that the implementation is the same. Since this isn't the case, the comment needs to be updated or removed.
Same for the implementation class.

src/Integration.Vsix/CFamily/ProcessRunner.cs
duncanp-sonar: So we're tying up two background threads for each analysis? That's not great.
Did you look at the threading docs for VS? https://github.com/microsoft/vs-threading/blob/master/doc/index.md

src/Integration.Vsix/CFamily/ProcessRunner.cs
duncanp-sonar: Exception messages are localized so checking against the en-us version isn't going to work reliably.

src/Integration.Vsix/CFamily/ProcessRunner.cs
duncanp-sonar: There's no point in having it if we don't use it, but I agree it seems odd. Maybe it's because it shows we don't provide any information about the success/failure of the execution.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -99,34 +99,33 @@ private async Task TriggerAnalysisAsync(Request request, IIssueConsumer consumer
             await TaskScheduler.Default;
 
             logger.WriteLine($"Analyzing {request.File}");
+            var messages = new List<Message>();
+            Action<Message> handleMessage = message => messages.Add(message); // todo: change to consumer.Accept
 
             // We're tying up a background thread waiting for out-of-process analysis. We could
             // change the process runner so it works asynchronously. Alternatively, we could change the
             // RequestAnalysis method to be synchronous, rather than fire-and-forget.
```
duncanp-sonar, line 107: Unrelated: I think I meant to say "asynchronous" in the comment on line 107, not "synchronous"...

src/Integration.Vsix/CFamily/ProcessRunner.cs
rita-g-sonarsource: Yes and I'm still not sure how to handle it instead of new Task :/

src/Integration.Vsix/CFamily/CFamilyHelper.cs
duncanp-sonar, line 124: Minor: no need for this check being inside the try-catch block.

It seems odd that we're checking for null here, but presumably there's a reason the code was like that before.

src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -117,60 +117,56 @@ private static Request CreateRequest(FileConfig fileConfig, string absoluteFileP
...
-                        Path.Combine(workingDirectory, "sonar-cfamily.reproducer"));
-
-                    // When running with reproducer flag, we don't want to show analysis results.
-                    return null;
-                }
-
-                if (success)
-                {
-                    using (var readStream = new FileStream(tempFileName, FileMode.Open))
+                    CmdLineArgs = new[] { "-" },
```
duncanp-sonar, line 134: Worth commenting that "-" tells the subprocess we want to communicate via standard IO streams.

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -116,66 +119,87 @@ public bool Execute(ProcessRunnerArguments runnerArgs)
...
+                    }
+                    finally
+                    {
+                        timeoutTask.Wait();
+                    }
+                }
             }
         }
 
         private bool WaitForProcessToFinish(Process process, ProcessRunnerArguments runnerArgs)
```
duncanp-sonar, line 156: Is the return value used?

src/Integration.Vsix/CFamily/ProcessRunner.cs
duncanp-sonar: Also, does this mean we aren't reliably writing the success/failure/timeout status to the output window?

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -116,66 +119,87 @@ public bool Execute(ProcessRunnerArguments runnerArgs)
...
+                        runnerArgs.HandleInputStream(process.StandardInput);
+                        runnerArgs.HandleOutputStream(process.StandardOutput);
+                    }
+                    catch (Exception) when (isRunningProcessCancelled)
+                    {
+                        // If a process is cancelled mid-stream, an exception will be thrown.
+                    }
+                    finally
+                    {
+                        timeoutTask.Wait();
```
duncanp-sonar, line 150: Why do we need to wait for the timeout task here? If I understand correctly, we'll already have read to the end of the output stream so we won't have any other processing to do.

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -52,7 +53,7 @@ public ProcessRunner(ISonarLintSettings settings, ILogger logger)
         /// Runs the specified executable and returns a boolean indicating success or failure
```
duncanp-sonar, line 53: Most of this comment is no longer accurate. More specifically, the threading implications of how the process is executed are not at all obvious so it needs some documentation to help us maintain it in future (not necessarily in a doc comment for this method - I think it needs more detail than that).

Some of the important points to capture (you can probably think of more):
* calling Execute is a blocking call. The method will return after the timeout or when the handler has read to the end of the output stream.
* the stream reader callbacks are executed on the original calling thread (?)
* we're using two threads: one to read the data from the stream, one to handle the timeout.
* errors and timeouts are logged to the output window. The caller won't see them and has no way of checking the outcome.

src/Integration.Vsix/CFamily/ProcessRunner.cs
duncanp-sonar: `runnerArgs.HandleInputStream/HandleOutputStream` could be null. Using null-coalesce operators should be enough (unless we're actually relying on the streams really being processed).

PR #1494, opened 2020-06-23
Deleted redundant code from tagger

duncanp-sonar:


src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -174,7 +171,7 @@ private void SafeOnBufferChange(object sender, TextContentChangedEventArgs e)
                 // The spans we have stored for issues relate to the previous text buffer and
                 // are no longer valid, so we need to translate them to the equivalent spans
                 // in the new text buffer.
-                UpdateDirtySpans(e);
+                currentSnapshot = e.After;
```
duncanp-sonar, line 174: This is the only line of code we need to keep from the `UpdateDirtySpans` method.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -113,8 +112,6 @@ public void AddTagger(IssueTagger tagger)
 
                 textBuffer.ChangedLowPriority += SafeOnBufferChange;
 
-                this.dirtySpans = new NormalizedSnapshotSpanCollection(new SnapshotSpan(currentSnapshot, 0, currentSnapshot.Length));
```
duncanp-sonar, line 116: This variable was been set and updated, but wasn't actually being used.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -186,20 +183,6 @@ private void SafeOnBufferChange(object sender, TextContentChangedEventArgs e)
...
-
-            var newDirtySpans = dirtySpans.CloneAndTrackTo(e.After, SpanTrackingMode.EdgeInclusive);
-
-            foreach (var change in e.Changes)
-            {
-                newDirtySpans = NormalizedSnapshotSpanCollection.Union(newDirtySpans, new NormalizedSnapshotSpanCollection(e.After, change.NewSpan));
-            }
-
-            dirtySpans = newDirtySpans;
-        }
```
duncanp-sonar, line 201: The original version of this code was copied from the VSSDK [Error List](https://github.com/microsoft/VSSDK-Extensibility-Samples/tree/master/ErrorList) sample.

The sample is for a spelling checker that displays typos in the Error List. It did use the set of dirty spans, as it used them to work out which bits of the document had changed and needed to be spell-checked again (i.e. it effectively re-ran the "analysis" on the just the bits of the document that had changed).

PR #1495, opened 2020-06-23
Simplified IIssueTracker and IssueTagger

duncanp-sonar:
* moved the responsibility for managing the taggers to the tracker
* removed dependency between the tagger and the tracker (dependency still exists the other way)
* added some tests for the tagger

### DIFF
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/IssueTaggerTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/IssueTaggerTests.cs
@@ -32,9 +32,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintTagger
     [TestClass]
     public class IssueTaggerTests
     {
-        private readonly IEnumerable<IssueMarker> ValidMarkerList = new[] { CreateValidMarker() };
-
-        private readonly SnapshotSpan ValidSnapshotSpan = new SnapshotSpan();
+        private static readonly IEnumerable<IssueMarker> ValidMarkerList = new[] { CreateValidMarker() };
+        private static readonly ITextSnapshot ValidTextSnapshot = Mock.Of<ITextSnapshot>();
+        private static readonly SnapshotSpan ValidSnapshotSpan = new SnapshotSpan(ValidTextSnapshot, 0, 0);
 
         [TestMethod]
         public void UpdateMarkers_NoSpan_EventNotRaised()
@@ -76,6 +76,17 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintTagger
         }
 
         [TestMethod]
+        public void GetTags_NullIssues_ReturnEmptyList()
+        {
+            var testSubject = new IssueTagger(null, null);
+            var normalizedSpans = new NormalizedSnapshotSpanCollection();
+
+            var result = testSubject.GetTags(normalizedSpans);
+
+            result.Should().BeEmpty();
+        }
+
+        [TestMethod]
         public void Dispose_CallsDelegateOnlyOnce()
         {
             var delegateCallCount = 0;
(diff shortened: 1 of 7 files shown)


src/Integration.Vsix/SonarLintTagger/IIssueTracker.cs
```diff
@@ -24,8 +24,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix
 {
     internal interface IIssueTracker
     {
-        void AddTagger(IssueTagger tagger);
-        void RemoveTagger(IssueTagger tagger);
```
duncanp-sonar, line 28: I've moved the responsibility for keeping track of the taggers-per-tracker to inside the tracker
i.e. the tracker will call add/remove itself, so it no longer requires publicly visible add/remove methods.

src/Integration.Vsix/SonarLintTagger/IssueTagger.cs
```diff
@@ -37,21 +37,21 @@ namespace SonarLint.VisualStudio.Integration.Vsix
...
-        private IssuesSnapshot issues;
+        public delegate void OnTaggerDisposed(IssueTagger issueTagger);
+
+        private IEnumerable<IssueMarker> issueMarkers;
+        private readonly OnTaggerDisposed onTaggerDisposed;
         private bool isDisposed;
-        internal /* for testing */ IIssueTracker IssueTracker { get; }
 
-        public IssueTagger(IIssueTracker issueTracker)
+        public IssueTagger(IEnumerable<IssueMarker> issueMarkers, OnTaggerDisposed onTaggerDisposed)
```
duncanp-sonar, line 46: Simplification: the tagger no longer needs to know about the tracker. It only needs to know the list of markers to display, and a callback to notify when it is disposed.
I could have used a `Func<>` instead of a delegate, but a delegate is slightly more self-descriptive.

src/Integration.Vsix/SonarLintTagger/IssueTagger.cs
```diff
@@ -37,21 +37,21 @@ namespace SonarLint.VisualStudio.Integration.Vsix
...
-            this.IssueTracker = issueTracker;
-            this.issues = issueTracker.LastIssues;
-
-            issueTracker.AddTagger(this);
+            this.issueMarkers = issueMarkers;
+            this.onTaggerDisposed = onTaggerDisposed;
         }
 
-        public void UpdateMarkers(IssuesSnapshot newIssues, SnapshotSpan? affectedSpan)
+        public void UpdateMarkers(IEnumerable<IssueMarker> newMarkers, SnapshotSpan? affectedSpan)
```
duncanp-sonar, line 52: Simplification: the tagger doesn't need to know about the `IssuesSnapshot` either. It only ever accessed the list of `IssueMarkers`.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -101,7 +101,15 @@ internal sealed class TextBufferIssueTracker : IIssueTracker, IIssueConsumer
...
         }
 
-        public void AddTagger(IssueTagger tagger)
+        public IssueTagger CreateTagger()
+        {
+            var tagger = new IssueTagger(this.Factory.CurrentSnapshot.IssueMarkers, RemoveTagger);
+            this.AddTagger(tagger);
+            
+            return tagger;
+        }
```
duncanp-sonar, line 110: The tracker is now responsible for managing the list of active trackers it has.

src/Integration.Vsix.UnitTests/SonarLintTagger/IssueTaggerTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            var testSubject = new IssueTagger(null, null);
+            var normalizedSpans = new NormalizedSnapshotSpanCollection();
+
+            var result = testSubject.GetTags(normalizedSpans);
+
+            result.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void GetTags_HasIntersectingIssues_ReturnsIntersectingIssues()
```
rita-g-sonarsource, line 90: Nice, it's very clear 👍

PR #1497, opened 2020-06-24
Use IO streams to communicate with subprocess.exe #1466

### LINKED ISSUE #1466: Use IO streams to communicate with subprocess.exe

rita-g-sonarsource:
Fixes #1466

### DIFF
--- a/src/Core.UnitTests/CFamily/CFamilyAnalyzerOptionsTests.cs
+++ b/src/Core.UnitTests/CFamily/CFamilyAnalyzerOptionsTests.cs
@@ -1,4 +1,24 @@
-﻿using FluentAssertions;
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Core.CFamily;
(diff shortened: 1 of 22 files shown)


rita-g-sonarsource:
@duncanp-sonar you can take a look at the commits separately, even though they're both large... First commit puts the Scheduler in charge of the cancellation, and the second commit is the streaming.

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -116,68 +120,33 @@ public bool Execute(ProcessRunnerArguments runnerArgs)
...
+                    process.Id);
 
                 try
                 {
-                    process.Kill();
-                    LogWarning(CFamilyStrings.WARN_ExecutionTimedOutKilled, runnerArgs.TimeoutInMilliseconds,
-                        runnerArgs.ExeName);
+                    runnerArgs.HandleInputStream(process.StandardInput);
+                    runnerArgs.HandleOutputStream(process.StandardOutput);
+                    process.WaitForExit();
```
rita-g-sonarsource, line 141: WaitForExit() is just so we could have the ExitCode. It was used in tests, so I left it.

src/Integration.Vsix/CFamily/ProcessRunnerArguments.cs
```diff
@@ -71,15 +67,16 @@ public ProcessRunnerArguments(string exeName, bool isBatchScript)
...
-        public int TimeoutInMilliseconds { get; set; }
-
         private bool IsBatchScript { get; set; }
 
         /// <summary>
         /// Additional environments variables that should be set/overridden for the process. Can be null.
         /// </summary>
         public IDictionary<string, string> EnvironmentVariables { get; set; }
 
+        public Action<StreamWriter> HandleInputStream { get; set; } = writer => { };
```
rita-g-sonarsource, line 77: @duncanp-sonar assigning to no-op so the actions won't be null.

src/Core/Analysis/IAnalyzerOptions.cs
```diff
@@ -25,5 +25,6 @@ namespace SonarLint.VisualStudio.Core.Analysis
     /// </summary>
     public interface IAnalyzerOptions
     {
+        int AnalysisTimeoutInMiliseconds { get; }
```
duncanp-sonar, line 28: Minor: typo in "Miliseconds"

src/Core.UnitTests/CFamily/CFamilyAnalyzerOptionsTests.cs
```diff
@@ -0,0 +1,30 @@
+﻿using FluentAssertions;
```
duncanp-sonar, line 1: (c)

src/Integration.Vsix/SonarLintTagger/IScheduler.cs
```diff
@@ -30,6 +30,6 @@ public interface IScheduler
         /// </summary>
         /// <param name="jobId">Unique key based on which previous job is cancelled</param>
         /// <param name="action">Job action</param>
-        void Schedule(string jobId, Action<CancellationToken> action);
+        void Schedule(string jobId, Action<CancellationToken> action, int timeoutInMiliseconds);
```
duncanp-sonar, line 33: Same typo, different context

src/Integration.Vsix/CFamily/ProcessRunner.cs
duncanp-sonar, line 141: Does the caller still have to start a blocking read operation in the callback on line 141 to stop the method from exiting immediately? If so, it's worth a comment.

src/Integration.Vsix/CFamily/ProcessRunnerArguments.cs
duncanp-sonar: A null coalesce at the point of invocation would make it clearer that nulls are ok; validating the arguments and throwing on null would make it clear that nulls are not ok. Setting a non-null default but allowing it to be changed to null isn't as clear.

src/Integration.Vsix/CFamily/ProcessRunner.cs
```diff
@@ -116,68 +120,33 @@ public bool Execute(ProcessRunnerArguments runnerArgs)
...
+                    ExitCode = process.ExitCode;
                 }
-                catch
+                catch (Exception) when (isRunningProcessCancelled)
```
duncanp-sonar, line 144: We shouldn't be catching critical exceptions.

duncanp-sonar: It's interesting that the previous code used `catch` without an exception type so it would catch non-CLS compliant exception. However, that was probably just a mistake.

src/Integration.Vsix/CFamily/ProcessRunner.cs
duncanp-sonar: Assuming the comment in the old code is correct, then it also gives the asynchronous events a chance to complete i.e. the handling of the error stream.

rita-g-sonarsource: Good point, I'll return the comment

src/Core.UnitTests/CFamily/CFamilyAnalyzerOptionsTests.cs
rita-g-sonarsource: 😢

PR #1501, opened 2020-06-25
Give user feedback during CFamily analysis #1498

### LINKED ISSUE #1498: Give user feedback during CFamily analysis
### Description

Users should be given some indication that the analysis is running in the background:
* a visual progress indicator that the analysis is running, and
* logged output indicating the result of the analysis (succeeded, error, timeout, cancelled/re-triggered)

https://docs.microsoft.com/en-us/visualstudio/extensibility/ux-guidelines/notifications-and-progress-for-visual-studio?view=vs-2019#BKMK_ProgressIndicators

rita-g-sonarsource:
Fixes #1498


src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        private void ShowStatus(string text, params object[] args)
+        {
+            vsStatusBar.SetText(string.Format(text, args));
+        }
+
+        private void ShowSpinner()
+        {
+            vsStatusBar.Animation(1, Microsoft.VisualStudio.Shell.Interop.Constants.SBAI_General);
```
rita-g-sonarsource, line 72: I can't get the damn spinner to show!
It returns an error code `-2147024809` which I can't find anywhere.
I've been following the [documentation](https://docs.microsoft.com/en-us/visualstudio/extensibility/extending-the-status-bar?view=vs-2015&redirectedfrom=MSDN), I don't know what I'm missing. I turned off "Automatically adjust visual experience based on client performance" like they suggested, and I still get an error code :( 
Do you have an idea why?

src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
```diff
@@ -0,0 +1,80 @@
...
+    [Export(typeof(IAnalysisStatusNotifier))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class AnalysisStatusNotifier : IAnalysisStatusNotifier
+    {
+        private readonly IVsStatusbar vsStatusBar;
+
+        [ImportingConstructor]
+        public AnalysisStatusNotifier([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider)
+        {
+            vsStatusBar = serviceProvider.GetService(typeof(IVsStatusbar)) as IVsStatusbar;
```
rita-g-sonarsource, line 38: Do I need to do something to make sure I'm on the UI thread?
I see that in the StatusBarDownloadProgressHandler we don't do anything special.

src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
duncanp-sonar: Try using the sample code code from the VS docs: https://docs.microsoft.com/en-us/visualstudio/extensibility/extending-the-status-bar?view=vs-2019#display-an-animation

The fact that the sample code is creating an object first does make a difference in this case.

Then, you'll find it throws an exception because you aren't on the UI thread. ;-)

src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
duncanp-sonar: I think the MEF composition is being done on the UI thread. You can check using `ThreadHelper.CheckAccess()` or `ThreadHelper.ThrowIfNotOnUIThread()`.

src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
rita-g-sonarsource: Hooray, it works! Thanks!

src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            Notify(string.Format(AnalysisStrings.Notifier_AnalysisCancelled, filePath), false);
+        }
+
+        public void AnalysisFailed(string filePath)
+        {
+            Notify(string.Format(AnalysisStrings.Notifier_AnalysisFailed, filePath), false);
+        }
+
+        private void Notify(string message, bool showSpinner)
```
rita-g-sonarsource, line 61: @duncanp-sonar made changes like you suggested

duncanp-sonar:
I wonder if the status bar message would be more readable/useful if it just showed the file name rather than the full file path. I haven’t tried just the file name, but I did find the full path difficult to follow/read when testing. It’s unlikely the user will have two files with the same name open at the same time so full path probably isn’t necessary.

src/Integration.Vsix/Analysis/AnalysisStrings.resx
```diff
@@ -163,4 +163,16 @@
...
     <value>Default settings will be used</value>
   </data>
+  <data name="Notifier_AnalysisCancelled" xml:space="preserve">
+    <value>SonarLint: Cancelled analysis of {0}</value>
+  </data>
+  <data name="Notifier_AnalysisEnded" xml:space="preserve">
+    <value>SonarLint: Finished analyzing {0}</value>
+  </data>
+  <data name="Notifier_AnalysisFailed" xml:space="preserve">
+    <value>SonarLint: Failed to analyze {0}, please see the Output Window for more information.</value>
```
duncanp-sonar, line 173: I wouldn't expect really to see "please" in a status bar message, and I don't think the use a comma here is correct - it should be a new sentence. Just "See the Output Window for more information." would do.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -41,18 +41,18 @@ internal class CLangAnalyzer : IAnalyzer
...
         private readonly ISonarLintSettings settings;
         private readonly ICFamilyRulesConfigProvider cFamilyRulesConfigProvider;
+        private readonly IAnalysisStatusNotifier analysisStatusNotifier;
         private readonly ILogger logger;
         private readonly DTE dte;
 
         [ImportingConstructor]
-        public CLangAnalyzer(ITelemetryManager telemetryManager, ISonarLintSettings settings, ICFamilyRulesConfigProvider cFamilyRulesConfigProvider,
-            [Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider, ILogger logger)
+        public CLangAnalyzer(ITelemetryManager telemetryManager, ISonarLintSettings settings, ICFamilyRulesConfigProvider cFamilyRulesConfigProvider, [Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider, IAnalysisStatusNotifier analysisStatusNotifier, ILogger logger)
```
duncanp-sonar, line 49: You don't need to import the service provider here.

duncanp-sonar:
How does this notification interact with the status bar notification when the settings file is changed?

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
rita-g-sonarsource: It was already there, it's used to get the DTE

rita-g-sonarsource:
> How does this notification interact with the status bar notification when the settings file is changed?

It's hard to say. In general, it's a race condition issue. And since we display an icon and a message, it's possible even to get unintended combinations, such as `cancelled` yet the spinner is on.
image

src/Integration.Vsix/SonarLintTagger/IScheduler.cs
```diff
@@ -31,6 +31,6 @@ public interface IScheduler
         /// <param name="jobId">Unique key based on which previous job is cancelled</param>
         /// <param name="action">Job action</param>
         /// <param name="timeoutInMilliseconds">number of milliseconds after which the job should be cancelled</param>
-        void Schedule(string jobId, Action<CancellationToken> action, int timeoutInMilliseconds);
+        void Schedule(string jobId, Action<CancellationToken> action, int? timeoutInMilliseconds);
```
rita-g-sonarsource, line 34: @duncanp-sonar I changed the parameter to nullable, since the tests started to flake. I suspect that this is because CancelAfter triggers the timer anyway, even if you pass Timeout.Inifinite. Since that's what we pass for JS, I decided to change that to null, so it will be explicit that a timer is triggered only when a value is deliberately passed.

src/Integration.Vsix/SonarLintTagger/ExtendedCancellationTokenSource.cs
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Reflection;
+using System.Threading;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
```
rita-g-sonarsource, line 24: This class and the Scheduler should move to Core, I'll do that in a different PR

src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            Notify(AnalysisStrings.Notifier_AnalysisStarted, filePath, true);
+        }
+
+        public void AnalysisFinished(string filePath)
+        {
+            Notify(AnalysisStrings.Notifier_AnalysisFinished, filePath, false);
+        }
+
+        public void AnalysisCancelled(string filePath)
```
rita-g-sonarsource, line 52: Currently I just clear the status bar. 
The output window will have the retriggered/cancelled message.

src/Integration.Vsix/CFamily/CFamilyStrings.resx
```diff
@@ -118,7 +118,10 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="ERROR_Analysis_Failed" xml:space="preserve">
-    <value>Failed to execute analysis, exception details: {0}</value>
+    <value>Failed to analyze {0}, exception details: {1}</value>
```
rita-g-sonarsource, line 121: I noticed that the "Failed" message didn't have the file name. I thought it was useful information to have.

src/Integration.Vsix/Analysis/AnalysisStrings.resx
```diff
@@ -163,4 +163,13 @@
...
     <value>Default settings will be used</value>
   </data>
+  <data name="Notifier_AnalysisFinished" xml:space="preserve">
+    <value>SonarLint: Finished analyzing {0}</value>
+  </data>
+  <data name="Notifier_AnalysisFailed" xml:space="preserve">
+    <value>SonarLint: Failed to analyze {0}. See the Output Window for more information.</value>
+  </data>
+  <data name="Notifier_AnalysisStarted" xml:space="preserve">
+    <value>SonarLint: Analyzing {0}</value>
```
duncanp-sonar, line 173: Minor: should this message end with an ellipsis ("...") to indicate it is an ongoing action?
I'm trying to remember if this is a general convention.

rita-g-sonarsource: I don't think so, the MS's one doesn't have it. 
(But then again, they have a spinner right next to it 🤷 we also have the progress icon but it's not as noticeable)

src/Integration.Vsix/SonarLintTagger/ExtendedCancellationTokenSource.cs
```diff
@@ -0,0 +1,29 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Threading;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
+{
+    public class ExtendedCancellationTokenSource : CancellationTokenSource
+    {
+        public bool IsCancelledExplicitly { get; set; }
```
duncanp-sonar, line 27: Is this now used for anything except testing? If so, it seems a bit heavy-weight.

rita-g-sonarsource: Just for testing now. It was useful in the tests to know why it was cancelled, to make sure that the tests are correct.

PR #1505, opened 2020-06-29
Refactor IIssueMarkerFactory

duncanp-sonar:
* changed to IIssueSpanCalculator


duncanp-sonar:
@rita-g-sonarsource I think this is a refactoring you suggested earlier in the sprint - changing the IIssueMarkerFactory to just return the span instead of calculating the span and returning it as part of a new object.

PR #1508, opened 2020-06-29
Bug Fix C++ Analysis Timeout is ignored

### LINKED ISSUE #1507: [BUG] C++ Analysis Timeout is ignored
### Description

C++ subprocess is not terminated after the allocated timeout.
Caused by #1466.

rita-g-sonarsource:
Fixes #1507


src/Integration.Vsix/SonarLintTagger/Scheduler.cs
```diff
@@ -38,29 +39,26 @@ public Scheduler()
...
-            var newTokenSource = IssueToken(jobId);
+            var newCancellationToken = IssueToken(jobId, timeoutInMilliseconds);
 
-            if (timeoutInMilliseconds.HasValue)
-            {
-                newTokenSource.CancelAfter(timeoutInMilliseconds.Value);
-            }
-
-            action(newTokenSource.Token);
+            action(newCancellationToken);
```
rita-g-sonarsource, line 46: There was a race condition here all along, which got discovered due to CancelAfter: 
1. Thread A gets to IssueToken, acquires a lock and issues a new CancellationTokenSource.
2. Thread B gets to IssueToken, is blocked by thread A.
3. Thread A releases the lock, and proceeds to CancelAfter.
4. Thread B acquires a lock and issues a new CancellationTokenSource, cancelling the token issued by thread A.
5. Thread A proceeds to `action(newTokenSource.Token)` and receives an Object disposed exception.

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
```diff
@@ -38,29 +38,26 @@ public Scheduler()
...
 
-        private CancellationTokenSource IssueToken(string jobId)
+        private CancellationToken IssueToken(string jobId, int timeoutInMilliseconds)
         {
             lock (jobs)
             {
                 CancelPreviousJob(jobId);
 
                 var newTokenSource = new CancellationTokenSource();
+                newTokenSource.CancelAfter(timeoutInMilliseconds);
```
duncanp-sonar, line 56: I can see how the race condition existed before since the `newTokenSource` was being accessed outside the lock.
However, I don't understand why the timeout needs to be mandatory to fix the bug.

rita-g-sonarsource: Yep, existed before but was much less likely to happen.
We will always pass a timeout now, so we might as well make it non-nullable.

src/Core/IEnvironmentSettings.cs
```diff
@@ -35,7 +35,7 @@ public interface IEnvironmentSettings
         /// <summary>
         /// Returns the user-specific timeout, or zero if one has not been specified
         /// </summary>
-        int CFamilyAnalysisTimeoutInMs();
+        int AnalysisTimeoutInMs();
```
rita-g-sonarsource, line 38: Should I rename the env variable as well?
It's currently SONAR_INTERNAL_CFAMILY_ANALYSIS_TIMEOUT_MS

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
duncanp-sonar: Is there any point in calling `CancelAfter` with an infinite timeout?

src/Core/IEnvironmentSettings.cs
duncanp-sonar: Yes, we should. The only problem is that some users are using it is a workaround for analysis timeout issues: see https://community.sonarsource.com/t/crash-with-sonarlint-in-with-extended-timeout-for-large-c-projects/21861.

So we should mention it in the release notes and post a note to those community threads.

duncanp-sonar: FYI I've added a documentation ticket to the sprint: #1509

rita-g-sonarsource: Alternatively, we could support both for backwards compatibility. WDYT?

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
rita-g-sonarsource: Are you suggesting to do `if timeoutInMilliseconds !=Timeout.Infinite`? 
I think it can be a large timeout anyway.

src/Core/IEnvironmentSettings.cs
duncanp-sonar: I'd say not - that's why it has "INTERNAL" in the name, because it's not a publicly supported feature and we're not guaranteeing we'll continue to support it.
But that doesn't mean we shouldn't give a heads-up to those users we know were using the old variable.

src/Integration.Vsix/SonarLintTagger/Scheduler.cs
duncanp-sonar: I was thinking about skipping the call to `CancelAfter`. However, since we're setting a non-infinite timeout in the product code, the only time this could be infinite would be if the user set it via the environment variable. And the validation logic for the environment variable ignores values < 1, so the user couldn't set it to infinite either.
So there's no point in changing anything here.

Issue #1487, opened 2020-06-22
Analyzer issues highlight the wrong text


duncanp-sonar:
### Description

Issues reported an analyzer may not be mapped to the correct text in the document if the document is edited in the interval between the analysis being triggered and the issues being reported.

#### Impact
The wrong text is underlined in the editor and the line and column displayed in the Error List can be incorrect.

![image](https://user-images.githubusercontent.com/10757559/85310821-6334ca00-b4ac-11ea-8ecc-2d15b33bc963.png)

Error markers can also be reported for issues in code that has been deleted.

### Repro steps

1. Trigger an analysis e.g. save a C++ file

2. Add empty lines at the start of the document before the analysis returns.

Note: this repro is timing-dependent. You can increase the likelihood of the issue appearing by analysing a large/complex C++ file for which symbolic execution will take many seconds to analyse.
Alternatively, artificially introduce a delay in the analysis in a debug version of the code.

### Expected behavior

The correct text is underlined in the editor.
The line and column numbers reported in the Error List are correct.
Issues in text that has been edited or deleted are not reported.

### Actual behavior

The wrong text is underlined, the line and column numbers are wrong, and "deleted" issues are still reported.

### Known workarounds

Don't edit the document until the analysis has completed.

### Related information

* SonarLint for VS Version: current and previous (v4.22)
* Languages in your solution: non-Roslyn languages i.e. JavaScript, C/C++

PR #1512, opened 2020-06-30
Refactoring - make IssueMarker filterable

duncanp-sonar:
* remove FilterableIssueAdapter

The filterable issue adapter was used to temporarily add a couple of properties to the "raw" analysis issue to allow it to be filtered. This seemed a bit over-complicated, and also makes it harder to re-filter issues when text buffer changes.
I've added the two properties to the IssueMarker class and make it implement `IFilterableIssue` so the adapter is no longer required.


src/Integration.Vsix/SonarLintTagger/IssueMarker.cs
```diff
@@ -30,20 +31,36 @@ namespace SonarLint.VisualStudio.Integration.Vsix
     /// <remarks>
     /// See the README.md in this folder for more information
     /// </remarks>
-    internal class IssueMarker
+    internal class IssueMarker : IFilterableIssue
     {
         public IAnalysisIssue Issue { get; }
         public SnapshotSpan Span { get; }
+        public string WholeLineText { get; set; }
+        public string LineHash { get; set; }
```
duncanp-sonar, line 39: I've added these two properties which are the ones required by `IFilterableIssue`.

src/Integration.Vsix/SonarLintTagger/IssueToIssueMarkerConverter.cs
```diff
@@ -19,47 +19,50 @@
...
-using System.Collections.Generic;
-using System.Linq;
 using Microsoft.VisualStudio.Text;
 using SonarLint.VisualStudio.Core.Analysis;
-using SonarLint.VisualStudio.Core.Suppression;
 using SonarLint.VisualStudio.Integration.Suppression;
 
 namespace SonarLint.VisualStudio.Integration.Vsix
 {
-    internal static class IssueToFilterableIssueConverter
```
duncanp-sonar, line 31: I renamed the static class, changed it to be non-static and added an interface.

src/Integration.Vsix/SonarLintTagger/IssueToIssueMarkerConverter.cs
```diff
@@ -19,47 +19,50 @@
...
+    internal interface IIssueToIssueMarkerConverter
     {
-        public static IEnumerable<IFilterableIssue> Convert(IEnumerable<IAnalysisIssue> issues, ITextSnapshot textSnapshot)
```
duncanp-sonar, line 33: This method converted a list. It's no longer needed.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -66,13 +66,13 @@ internal sealed class TextBufferIssueTracker : IIssueTracker, IIssueConsumer
...
             ISonarErrorListDataSource sonarErrorDataSource, ILogger logger)
-            : this(dte, provider, document, detectedLanguages, issuesFilter, sonarErrorDataSource, logger, new IssueSpanCalculator())
+            : this(dte, provider, document, detectedLanguages, issuesFilter, sonarErrorDataSource, logger, new IssueToIssueMarkerConverter())
         {
         }
 
         internal TextBufferIssueTracker(DTE dte, TaggerProvider provider, ITextDocument document,
             IEnumerable<AnalysisLanguage> detectedLanguages, IIssuesFilter issuesFilter,
-            ISonarErrorListDataSource sonarErrorDataSource, ILogger logger, IIssueSpanCalculator issueSpanCalculator)
+            ISonarErrorListDataSource sonarErrorDataSource, ILogger logger, IIssueToIssueMarkerConverter issueMarkerConverter)
```
duncanp-sonar, line 75: The span calculator is used internally by the new type, so the tracker doesn't need to know about it.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -207,12 +207,6 @@ private IssuesSnapshot TranslateMarkerSpans()
         private bool IsValidIssueTextRange(IAnalysisIssue issue) =>
             1 <= issue.StartLine && issue.EndLine <= currentSnapshot.LineCount;
 
-        private IssueMarker CreateIssueMarker(IAnalysisIssue issue)
```
duncanp-sonar, line 210: Handled by the new class.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -273,26 +267,29 @@ void IIssueConsumer.Accept(string path, IEnumerable<IAnalysisIssue> issues)
...
             }
 
-            var filteredIssues = RemoveSuppressedIssues(issues);
-
             // See bug #1487: we should be creating the issue markers using the text snapshot
             // from when the analysis was triggered, not the current one.
-            var newMarkers = filteredIssues.Where(IsValidIssueTextRange).Select(CreateIssueMarker);
-            UpdateIssues(newMarkers);
+            var newMarkers = issues
+                .Where(IsValidIssueTextRange)
```
duncanp-sonar, line 273: I re-ordered the processing slightly so that out-of-range issues are still removed before the new `IssueMarker` instances are created.
However, I don't think this matters too much; the ordering will change again when we fix #1487, and we might be able to drop `IsValidIssueTextRange`.

PR #1514, opened 2020-06-30
Add analysis time

rita-g-sonarsource:


src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
```diff
@@ -59,15 +59,15 @@ public void AnalysisFailed(string filePath)
             Notify(AnalysisStrings.Notifier_AnalysisFailed, filePath, false);
         }
 
-        private void Notify(string messageFormat, string filePath, bool showSpinner)
+        private void Notify(string messageFormat, string filePath, bool showSpinner, double time = 0)
```
rita-g-sonarsource, line 62: Optional param is throw-away code, not gonna do it like that in the feature branch.

duncanp-sonar: That's fine. The prototype branch can have throw-away and hacking implementations.
By extension, it doesn't matter if there are test failures, as long as we are confident that there aren't any blocking bugs that would prevent our UX consultant from evaluating the new behaviour(s).

PR #1517, opened 2020-07-01
WIP file-level issue

rita-g-sonarsource:


src/Integration.Vsix/CFamily/PortedFromJava/Protocol.cs
```diff
@@ -178,7 +178,7 @@ public static void Read(BinaryReader reader, Action<Message> handleIssue, string
               case "message":
                   var message = readMessage(reader);
 
-                  if (!doFilterResults || string.Equals(issueFilePath, Path.GetFullPath(message.Filename), StringComparison.InvariantCultureIgnoreCase))
+                  if (!doFilterResults || string.IsNullOrEmpty(message.Filename) || string.Equals(issueFilePath, Path.GetFullPath(message.Filename), StringComparison.InvariantCultureIgnoreCase))
```
rita-g-sonarsource, line 181: Path.GetFullPath is what's originally causing the exception in the ticket.

src/Integration.Vsix/SonarLintTagger/IssueMarker.cs
```diff
@@ -33,9 +33,9 @@ namespace SonarLint.VisualStudio.Integration.Vsix
     internal class IssueMarker
     {
         public IAnalysisIssue Issue { get; }
-        public SnapshotSpan Span { get; }
+        public SnapshotSpan? Span { get; }
```
rita-g-sonarsource, line 36: I made the span nullable. I suspect the better alternative would be to create IIssueMarker with just IAnalysisIssue, and ITextBufferIssueMarker  that also has a SnapshotSpan.

src/Integration.Vsix/SonarLintTagger/IssueToFilterableIssueConverter.cs
```diff
@@ -56,8 +56,8 @@ public static IEnumerable<IFilterableIssue> Convert(IEnumerable<IAnalysisIssue>
                 return null;
             }
 
-            // A start line of zero means the issue is file-level i.e. not associated with a particular line
-            if (issue.StartLine == 0)
+            // A start line of zero or empty file path means the issue is file-level i.e. not associated with a particular line
+            if (issue.StartLine == 0  || string.IsNullOrEmpty(issue.FilePath))
```
rita-g-sonarsource, line 60: I don't know why, but the StartLine returned is not 0. Maybe it's a bug in the subprocess. In either case, I can also change the Protocol to set it to 0.

src/Integration.Vsix/SonarLintTagger/IssueTagger.cs
```diff
@@ -82,8 +82,8 @@ public IEnumerable<ITagSpan<IErrorTag>> GetTags(NormalizedSnapshotSpanCollection
             }
 
             return issueMarkers
-                .Where(marker => spans.IntersectsWith(marker.Span))
-                .Select(marker => new TagSpan<IErrorTag>(marker.Span, new ErrorTag(PredefinedErrorTypeNames.Warning, marker.Issue.Message)));
+                .Where(marker => marker.Span != null && spans.IntersectsWith(marker.Span.Value))
```
rita-g-sonarsource, line 85: 🤮

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -239,14 +244,14 @@ private void SnapToNewSnapshot(IssuesSnapshot newIssues)
 
             if (oldIssues != null && oldIssues.Count > 0)
             {
-                start = oldIssues.IssueMarkers.Select(i => i.Span.Start.TranslateTo(currentSnapshot, PointTrackingMode.Negative)).Min();
-                end = oldIssues.IssueMarkers.Select(i => i.Span.End.TranslateTo(currentSnapshot, PointTrackingMode.Positive)).Max();
+                start = oldIssues.IssueMarkers.Where(x=> !string.IsNullOrEmpty(x.Issue.FilePath)).Select(i => i.Span?.Start.TranslateTo(currentSnapshot, PointTrackingMode.Negative)).WhereNotNull().Min();
```
rita-g-sonarsource, line 247: 🤮 🤮 🤮

src/Integration.Vsix/SonarLintTagger/IssueToFilterableIssueConverter.cs
duncanp-sonar: I'd check with the CFamily folk and ask what the line numbers represent in this case.

src/Integration.Vsix/SonarLintTagger/IssueTagger.cs
duncanp-sonar: There's no point in passing non-taggable markers to the issue tagger.
The code in the tracker that calculates the affected span has does have to care about which issues are taggable, so it could be responsible for just passing taggable issues to the issue taggers.

src/Integration.Vsix/SonarLintTagger/IssueMarker.cs
duncanp-sonar: I think making the span nullable is ok; I don't think switching on type in multiple places is any better.

Adding an extension method that filters a list of issue markers that are "taggable" would be effectively the same as doing the `OfType` check.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -87,30 +89,30 @@ public override bool TryGetValue(int index, string keyName, out object content)
...
-                    content = this.issueMarkers[index].Span.Start.GetContainingLine().LineNumber;
+                    content = issueMarker.Span?.Start.GetContainingLine().LineNumber;
                     return true;
 
                 case StandardTableKeyNames.Column:
-                    // Use the span, not the issue. See comment immediately above.
-                    var position = this.issueMarkers[index].Span.Start;
-                    var line = position.GetContainingLine();
-                    content = position.Position - line.Start.Position;
+                        // Use the span, not the issue. See comment immediately above.
```
duncanp-sonar, line 96: Added whitespace

Issue #1471, opened 2020-06-16
Display CFamily analysis issues incrementally


duncanp-sonar:
### Description

Streaming CFamily issues from `subprocess.exe` allows us to display new issues as they are reported, rather than waiting for all of the issues to reported before updating the UI.

There are many different user experiences we could build on top of the new streaming mechanism. 
The proposed workflow described below is the simplest from an implementation point of view. We will implement this solution and evaluate whether the resulting UX is acceptable.

#### Status reporting
There should be a visible indication that analysis has started, is in progress, and has completed (either successfully or unsuccessfully).
In this sprint we are not planning on any custom UI feedback mechanisms. We will use only those provided out of the box by Visual Studio. The VS guidance on progress indicators is [here](https://docs.microsoft.com/en-us/visualstudio/extensibility/ux-guidelines/notifications-and-progress-for-visual-studio?view=vs-2019#BKMK_ProgressIndicators).

#### Proposed initial workflow
1. Trigger analysis for a file, streaming the results from `subprocess.exe`
2. When the first issue arrives, clear the existing issues for the file
3. Create a new snapshot containing the new issue
4. Update the error list
5. Repeat steps (3) and (4) as further issues are reported; each time, the new snapshot will contain one more issue than the previous snapshot.

TODO: decide when to update the tags in the editor. This could be done at the same time the error list is updated, or at the end.

#### Special cases to consider
* a new analysis is triggered before the previous analysis has completed: check the in-process analysis is cancelled
* ?time limit for a single file analysis? Infinite? One hour? One minute?
* highlighting the correct line of text in the editor: e.g. if an issue is reported after twenty seconds, there is a risk that the user will have edited the file in the meantime so the issue location will have changed. This is not a new issue, although it is more likely to occur if we increase the time an analysis can run before it is cancelled. CHECK: do we currently handle this scenario correctly?

#### UX Risks
Removing all of the old issues at once might be visually distracting.
Pushing an update to the error list/editor as each new issue arrives might negatively impact the responsiveness of the UI. It might also be visually distracting to have the UI update repeatedly as new issues arrive (e.g. if the error list re-ordered each time).

PR #1518, opened 2020-07-01
Refactor issue consumer to separate class

### LINKED ISSUE #1487: Analyzer issues highlight the wrong text

duncanp-sonar:
Relates to #1471
Fixes #1487

* separate issue consumer into separate class
* translate spans from the analysis snapshot to the current snapshot
* minor refactoring of buffer handling code
* added unit test for buffer handling code

TODO: change the CLangAnalyzer to actually stream the issues...

### DIFF
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
@@ -69,22 +69,24 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintTagger
         }
 
         [TestMethod]
-        [DataRow(-1, 10, false)] // start line < 1
+        [DataRow(-1, 1, false)] // start line < 1
         [DataRow(0, 0, false)] // file-level issue, can't be mapped to snapshot
+        [DataRow(0, 1, false)] // illegal i.e. shouldn't happen, but should be ignored if it does
         [DataRow(1, 1, true)] // starts in first line of snapshot
         [DataRow(9, 10, true)] // in snapshot
         [DataRow(10, 10, true)] // end is in last line of snapshot
         [DataRow(10, 11, false)] // end is outside snapshot
-        [DataRow(11, 11, false)] // ditto
-        [DataRow(99, 99, false)] // ditto
-        public void Accept_IssuesNotInSnapshotAreIgnored_CallbackIsCalledWithExpectedIssues(int oneBasedIssueStartLine, int oneBasedIssueEndLine, bool isMappableToSnapshot)
+        [DataRow(11, 11, false)] // end is outside snapshot
+        public void Accept_IssuesNotInSnapshotAreIgnored_CallbackIsCalledWithExpectedIssues(int issueStartLine, int issueEndLine, bool isMappableToSnapshot)
         {
+            // Issues are 1-based.
+            // Snapshots are 0-based so last line = index 9
             const int LinesInSnapshot = 10;
             var snapshot = CreateSnapshot(LinesInSnapshot);
-            var issues = new[] { CreateIssue(oneBasedIssueStartLine, oneBasedIssueEndLine) };
+            var issues = new[] { CreateIssue(issueStartLine, issueEndLine) };
 
             var callbackSpy = new OnIssuesChangedCallbackSpy();
-            var converter = CreateWrapAllConverter();
+            var converter = CreatePassthroughConverter();
 
             var testSubject = new AccumulatingIssueConsumer(snapshot, ValidFilePath, callbackSpy.Callback, converter);
 
@@ -119,7 +121,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintTagger
             };
 
             var snapshot = CreateSnapshot(lineCount: 10);
-            var converter = CreateWrapAllConverter();
+            var converter = CreatePassthroughConverter();
 
             var testSubject = new AccumulatingIssueConsumer(snapshot, ValidFilePath, callbackSpy.Callback, converter);
 
@@ -170,7 +172,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintTagger
                 Message = "any message"
             };
 
-        private static IIssueToIssueMarkerConverter CreateWrapAllConverter()
+        private static IIssueToIssueMarkerConverter CreatePassthroughConverter()
         {
             // Set up an issue converter that just wraps and returns the supplied issues as IssueMarkers
             var mockIssueConverter = new Mock<IIssueToIssueMarkerConverter>();
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
@@ -1,0 +1,182 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.Text;
+using Moq;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Integration.Vsix;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintTagger
+{
+    [TestClass]
+    public class AccumulatingIssueConsumerTests
+    {
+        private static readonly IAnalysisIssue ValidIssue = CreateIssue(startLine: 1, endline: 1);
+        private static readonly ITextSnapshot ValidTextSnapshot = CreateSnapshot(lineCount: 10);
+        private const string ValidFilePath = "c:\\myfile.txt";
+
+        [TestMethod]
+        public void Ctor_InvalidArgs_Throws()
+        {
+            AccumulatingIssueConsumer.OnIssuesChanged validCallback = _ => { };
+
+            Action act = () => new AccumulatingIssueConsumer(null, ValidFilePath, validCallback);
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("analysisSnapshot");
+
+            act = () => new AccumulatingIssueConsumer(ValidTextSnapshot, null, validCallback);
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("analysisFilePath");
+
+            act = () => new AccumulatingIssueConsumer(ValidTextSnapshot, ValidFilePath, null);
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("onIssuesChangedCallback");
+        }
+
+        [TestMethod]
+        public void Accept_WrongFile_CallbackIsNotCalled()
+        {
+            var callbackSpy = new OnIssuesChangedCallbackSpy();
+            var issues = new IAnalysisIssue[] { ValidIssue };
+
+            var testSubject = new AccumulatingIssueConsumer(ValidTextSnapshot, "c:\\file1.txt", callbackSpy.Callback);
+
+            using (new AssertIgnoreScope())
+            {
+                testSubject.Accept("wrong file", issues);
+            }
+
+            callbackSpy.CallCount.Should().Be(0);
+        }
+
+        [TestMethod]
+        [DataRow(-1, 10, false)] // start line < 1
+        [DataRow(0, 0, false)] // file-level issue, can't be mapped to snapshot
+        [DataRow(1, 1, true)] // starts in first line of snapshot
+        [DataRow(9, 10, true)] // in snapshot
+        [DataRow(10, 10, true)] // end is in last line of snapshot
+        [DataRow(10, 11, false)] // end is outside snapshot
+        [DataRow(11, 11, false)] // ditto
+        [DataRow(99, 99, false)] // ditto
+        public void Accept_IssuesNotInSnapshotAreIgnored_CallbackIsCalledWithExpectedIssues(int oneBasedIssueStartLine, int oneBasedIssueEndLine, bool isMappableToSnapshot)
+        {
+            const int LinesInSnapshot = 10;
+            var snapshot = CreateSnapshot(LinesInSnapshot);
+            var issues = new[] { CreateIssue(oneBasedIssueStartLine, oneBasedIssueEndLine) };
+
+            var callbackSpy = new OnIssuesChangedCallbackSpy();
+            var converter = CreateWrapAllConverter();
+
+            var testSubject = new AccumulatingIssueConsumer(snapshot, ValidFilePath, callbackSpy.Callback, converter);
+
+            using (new AssertIgnoreScope())
+            {
+                testSubject.Accept(ValidFilePath, issues);
+            }
+
+            callbackSpy.CallCount.Should().Be(1);
+            if (isMappableToSnapshot)
+            {
+                callbackSpy.LastSuppliedIssues.Should().BeEquivalentTo(issues);
+            }
+            else
+            {
+                callbackSpy.LastSuppliedMarkers.Should().BeEmpty();
+            }
+        }
+
+        [TestMethod]
+        public void Accept_MultipleCallsToAccept_IssuesAreAccumulated()
+        {
+            var callbackSpy = new OnIssuesChangedCallbackSpy();
+            var firstSetOfIssues = new[]
+            {
+                CreateIssue(1, 1), CreateIssue(2, 2)
+            };
+
+            var secondSetOfIssues = new[]
+            {
+                CreateIssue(3,3), CreateIssue(4,4)
+            };
+
+            var snapshot = CreateSnapshot(lineCount: 10);
+            var converter = CreateWrapAllConverter();
+
+            var testSubject = new AccumulatingIssueConsumer(snapshot, ValidFilePath, callbackSpy.Callback, converter);
+
+            // 1. First call
+            testSubject.Accept(ValidFilePath, firstSetOfIssues);
+
+            callbackSpy.CallCount.Should().Be(1);
+            callbackSpy.LastSuppliedIssues.Should().BeEquivalentTo(firstSetOfIssues);
+
+            // 2. Second call
+            testSubject.Accept(ValidFilePath, secondSetOfIssues);
+
+            callbackSpy.CallCount.Should().Be(2);
+            callbackSpy.LastSuppliedIssues.Should().BeEquivalentTo(firstSetOfIssues.Union(secondSetOfIssues));
+        }
+
+        private class OnIssuesChangedCallbackSpy
+        {
+            public int CallCount { get; private set; }
+            public IList<IssueMarker> LastSuppliedMarkers { get; private set; }
+            public IList<IAnalysisIssue> LastSuppliedIssues
+            {
+                get
+                {
+                    return LastSuppliedMarkers?.Select(x => x.Issue).ToList();
+                }
+            }
+
+            public void Callback(IEnumerable<IssueMarker> issueMarkers)
+            {
+                CallCount++;
+                LastSuppliedMarkers = issueMarkers?.ToList();
+            }
+        }
+
+        private static ITextSnapshot CreateSnapshot(int lineCount)
+        {
+            var mockSnapshot = new Mock<ITextSnapshot>();
+            mockSnapshot.Setup(x => x.LineCount).Returns(lineCount);
+            return mockSnapshot.Object;
+        }
+
+        private static IAnalysisIssue CreateIssue(int startLine, int endline) =>
+            new DummyAnalysisIssue
+            {
+                StartLine = startLine,
+                EndLine = endline,
+                Message = "any message"
+            };
+
+        private static IIssueToIssueMarkerConverter CreateWrapAllConverter()
+        {
+            // Set up an issue converter that just wraps and returns the supplied issues as IssueMarkers
+            var mockIssueConverter = new Mock<IIssueToIssueMarkerConverter>();
+            mockIssueConverter.Setup(x => x.Convert(It.IsAny<IAnalysisIssue>(), It.IsAny<ITextSnapshot>()))
+                .Returns<IAnalysisIssue, ITextSnapshot>((issue, snapshot) => new IssueMarker(issue, new SnapshotSpan(), null, null));
+            return mockIssueConverter.Object;
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -41,7 +41,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix
     /// See the README.md in this folder for more information
     /// </para>
     /// </remarks>
-    internal sealed class TextBufferIssueTracker : IIssueTracker, IIssueConsumer
+    internal class TextBufferIssueTracker : IIssueTracker
```
duncanp-sonar, line 44: `IIssueConsumer` is now implemented by a separate class.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -66,13 +65,6 @@ internal sealed class TextBufferIssueTracker : IIssueTracker, IIssueConsumer
         public TextBufferIssueTracker(DTE dte, TaggerProvider provider, ITextDocument document,
             IEnumerable<AnalysisLanguage> detectedLanguages, IIssuesFilter issuesFilter,
             ISonarErrorListDataSource sonarErrorDataSource, ILogger logger)
-            : this(dte, provider, document, detectedLanguages, issuesFilter, sonarErrorDataSource, logger, new IssueToIssueMarkerConverter())
```
duncanp-sonar, line 69: Now the issue conversion/decoration is handled by the new class, -> the converter parameter is not required -> the internal constructor is not required.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -184,29 +176,25 @@ private void SafeOnBufferChange(object sender, TextContentChangedEventArgs e)
                 // in the new text buffer.
                 currentSnapshot = e.After;
 
-                var newMarkers = TranslateMarkerSpans();
-
-                SnapToNewSnapshot(newMarkers);
+                var newMarkers = TranslateSpans(Factory.CurrentSnapshot.IssueMarkers, currentSnapshot);
+                UpdateIssues(newMarkers);
```
duncanp-sonar, line 180: Minor refactoring so there is only one "translate" method and only one method that creates a new snapshot.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -101,6 +104,31 @@ private void CheckFactoryWasUnregisteredFromDataSource(SnapshotFactory factory,
             mockSonarErrorDataSource.Verify(x => x.RemoveFactory(factory), times);
         }
 
+        [TestMethod]
+        public void OnBufferChange_ErrorListIsUpdated()
```
duncanp-sonar, line 108: Previously there weren't any tests for the buffer change handling.
This at least tests that an analysis is not triggered and the error list is updated.

We can't easily check whether the taggers are called since to do that we need some valid spans.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -187,7 +213,7 @@ public void WhenFileIsLoaded_AnalysisIsNotRequested()
 
                 // Sanity check (that the test setup is correct and that events are actually being handled)
                 RaiseFileSavedEvent(mockedJavascriptDocumentFooJs);
-                mockAnalyzerController.Verify(x => x.ExecuteAnalysis("foo.js", "utf-8", new AnalysisLanguage[] { AnalysisLanguage.Javascript }, testSubject, It.IsAny<IAnalyzerOptions>(), It.IsAny<CancellationToken>()), Times.Once);
+                CheckAnalysisWasRequested();
```
duncanp-sonar, line 216: Minor cleanup to avoid repeated code.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -61,7 +63,8 @@ public void SetUp()
             mockAnalyzerController = new Mock<IAnalyzerController>();
             issuesFilter = new Mock<IIssuesFilter>();
             taggerProvider = CreateTaggerProvider();
-            mockedJavascriptDocumentFooJs = CreateDocumentMock("foo.js");
+            mockDocumentTextBuffer = CreateTextBufferMock();
+            mockedJavascriptDocumentFooJs = CreateDocumentMock("foo.js", mockDocumentTextBuffer.Object);
```
duncanp-sonar, line 67: Minor change to the test setup: we need access to the mock text buffer so we can simulate it raising an event.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -209,89 +235,51 @@ private static void RaiseFileLoadedEvent(Mock<ITextDocument> mockDocument)
...
             mockAnalyzerController.Verify(x => x.ExecuteAnalysis(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<IEnumerable<AnalysisLanguage>>(),
                 It.IsAny<IIssueConsumer>(), It.IsAny<IAnalyzerOptions>(), It.IsAny<CancellationToken>()), Times.Never);
         }
-
-        #endregion
-
-        #region Processing analysis results tests
-
-        [TestMethod]
-        public void WhenNewIssuesAreFound_ButForIncorrectFile_FilterNotCalledAndListenersAreNotUpdated()
```
duncanp-sonar, line 223: Removed two tests: now the responsibility of the accumulating issue consumer and tested there.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -428,28 +424,59 @@ private static Mock<ITextDocument> CreateDocumentMock(string fileName)
...
 
-        private static Mock<ITextSnapshot> CreateMockTextSnapshot(int lineCount, string textToReturn)
+        private static Mock<ITextSnapshot> CreateMockTextSnapshot(int lineCount, string textToReturn, int versionNumber = 0)
         {
+            const int length = 9999;
+
+            var mockVersion = new Mock<ITextVersion>();
+            versionNumber = Math.Max(versionNumber, 1);
+            mockVersion.Setup(x => x.VersionNumber).Returns(versionNumber);
+            mockVersion.Setup(x => x.Length).Returns(length);
```
duncanp-sonar, line 449: I'll undo this change - we don't need a valid text version (left over from when I was trying to mock the data required by `SnapshotSpan.TranslateTo`).

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
```diff
@@ -0,0 +1,117 @@
...
+            Debug.Assert(issues.All(IsIssueInAnalysisSnapshot), "Not all reported issues could be mapped to the analysis snapshot");
+
+            var newMarkers = issues
+                .Where(IsIssueInAnalysisSnapshot)
+                .Select(x => issueMarkerConverter.Convert(x, analysisSnapshot))
+                .ToArray();
+
+            allIssueMarkers.AddRange(newMarkers);
+
+            onIssuesChanged.Invoke(allIssueMarkers);
```
rita-g-sonarsource, line 106: There is a problem here: we invoke this callback time a new issue arrives, and we hand it all accumulated issues so far. Which means that we keep repeating the same logic over and over again: checking for suppressions and calculating spans for issues that have already been calculated. As long as the text buffer has not been changed, there is no reason to keep re-calculating this data.
This logic should move out of the TextBufferIssueTracker as well:
1. IIssuesFilter should be called from AccumulatingIssueConsumer
2. TextBufferIssueTracker.TranslateSpans method should be extracted to a service, and called from both AccumulatingIssueConsumer (when issues arrive) and from TextBufferIssueTracker (when the buffer changes). It would also simplify the tests this way.

duncanp-sonar: 1. I deliberately didn't do the filtering here precisely so the original list of issues was preserved. The issue filtering should be "live" i.e. if we have fetched new suppressions from the server in the meantime then what is filtered out could change. The TextBufferIssueTracker is also filtering too early, but that's a separate issue.

2. This class only maps the "raw" issues to spans in the original analysis snapshot, and it only maps spans for the newly-received issues. The issues have to be mapped against the analysis snapshot (to fix #1487).
This class has no concept of the notion of _translating_ spans between different snapshots. This is deliberate - only the text buffer needs to know that multiple snapshots exist, and which one is the current snapshot.

The list of accumulated markers is then sent to the TextBufferIssueTracker, which may or may not have a changed in the meantime. If it has changed, then all of the spans need to be _translated_.

This implementation is functionally correct, and I chose it as it's the simplest one that could work. I agree that at present it may be translating spans repeatedly in some cases. However, translating spans has to be very fast, as it happens every time to press a key in the editor. Since it doesn't seem to be causing a performance problem, do we need to do anything more complicated?

rita-g-sonarsource: To the outside reviewer, I disagree that this is the simplest implementation. 
The snapshot logic is complicated, and it being split between the two classes (AccumulatingIssueConsumer and TextBufferIssueTracker). 
The fact that AccumulatingIssueConsumer calculates the spans of the issues for _a snapshot_ and not _the current snapshot_ is extremely confusing. It's doing half a job. The calculation is incomplete.

Either AccumulatingIssueConsumer should know about the translation, or the TextBufferIssueTracker should only know about markers and not issues -- which means that the suppression logic shouldn't be there.
If we say that TextBufferIssueTracker is responsible for translating markers to the latest snapshot, then that's all it should do.

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            this.issueMarkerConverter = issueMarkerConverter ?? throw new ArgumentNullException(nameof(analysisSnapshot));
+
+            allIssueMarkers = new List<IssueMarker>();
+        }
+
+        public void Accept(string path, IEnumerable<IAnalysisIssue> issues)
+        {
+            // Callback from the daemon when new results are available
+            if (path != analysisFilePath)
```
rita-g-sonarsource, line 91: Unrelated: We should change the signature of IIssueConsumer. The 'path' parameter is redundant: each issue has a filePath, so if anything, we should be testing that all issues are applicable analysisFilePath.

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
duncanp-sonar: As discussed offline (in case we ever need to dig out this PR again in the future)
* the issue consumer knows about how to map from file-level issues to the in-memory snapshot representation. It doesn't need to know that multiple snapshots exist.
* the text buffer already knows that multiple snapshots exists, and mapping between them is part of its job.
So I think the division of responsibilities with regard to the spans makes sense.

Where the filtering is done is an orthogonal issue. It could be done in either, and at this point it probably doesn't matter much which.
For me, issue consumer is a transient object; it's lifetime is limited to the execution of a single analysis run for a single file, and it should produce a list of all issues for that file.
What should be filtered/suppressed changes independently of the analysis results (e.g. based on suppressions on the server). Filtering/suppression is a view-level operation that should be done every time the view is refreshed i.e. we should have a "model" which contains all the issues for the file which we filter before rendering. It shouldn't be a one-off operation that discards data, since if the filter changes then when we re-apply it we might get the wrong results.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
```diff
@@ -0,0 +1,182 @@
...
+            callbackSpy.LastSuppliedIssues.Should().BeEquivalentTo(firstSetOfIssues);
+
+            // 2. Second call
+            testSubject.Accept(ValidFilePath, secondSetOfIssues);
+
+            callbackSpy.CallCount.Should().Be(2);
+            callbackSpy.LastSuppliedIssues.Should().BeEquivalentTo(firstSetOfIssues.Union(secondSetOfIssues));
+        }
+
+        private class OnIssuesChangedCallbackSpy
```
rita-g-sonarsource, line 139: We can use Mock< AccumulatingIssueConsumer.OnIssuesChanged> and verify what it was called with and the count.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        private static IAnalysisIssue CreateIssue(int startLine, int endline) =>
+            new DummyAnalysisIssue
+            {
+                StartLine = startLine,
+                EndLine = endline,
+                Message = "any message"
+            };
+
+        private static IIssueToIssueMarkerConverter CreateWrapAllConverter()
```
rita-g-sonarsource, line 173: Maybe CreateCallThroughConverter? Or CreateDummyConverter/ just CreateIssueConverter. I didn't understand what `WrapAll` meant and had to look at the code anyway.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
```diff
@@ -0,0 +1,182 @@
...
+        [TestMethod]
+        [DataRow(-1, 10, false)] // start line < 1
+        [DataRow(0, 0, false)] // file-level issue, can't be mapped to snapshot
+        [DataRow(1, 1, true)] // starts in first line of snapshot
+        [DataRow(9, 10, true)] // in snapshot
+        [DataRow(10, 10, true)] // end is in last line of snapshot
+        [DataRow(10, 11, false)] // end is outside snapshot
+        [DataRow(11, 11, false)] // ditto
+        [DataRow(99, 99, false)] // ditto
+        public void Accept_IssuesNotInSnapshotAreIgnored_CallbackIsCalledWithExpectedIssues(int oneBasedIssueStartLine, int oneBasedIssueEndLine, bool isMappableToSnapshot)
```
rita-g-sonarsource, line 80: Took me time to understand what you meant by 'oneBased'. I read it like "one issue based on...", and thought "based on what"? 😄 
Is it relevant that it's one based? I think it's clear enough enough without it.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        [TestMethod]
+        [DataRow(-1, 10, false)] // start line < 1
+        [DataRow(0, 0, false)] // file-level issue, can't be mapped to snapshot
+        [DataRow(1, 1, true)] // starts in first line of snapshot
+        [DataRow(9, 10, true)] // in snapshot
+        [DataRow(10, 10, true)] // end is in last line of snapshot
+        [DataRow(10, 11, false)] // end is outside snapshot
+        [DataRow(11, 11, false)] // ditto
```
rita-g-sonarsource, line 78: It's a bit dangerous IMO to leave the "ditto" comment, if someone were to change the order of the cases, or add something in between, it would change the meaning. Maybe just copy-paste the comment.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
```diff
@@ -0,0 +1,182 @@
...
+        [DataRow(99, 99, false)] // ditto
```
rita-g-sonarsource, line 79: Do we care about the 99 case?

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
```diff
@@ -0,0 +1,182 @@
...
+            using (new AssertIgnoreScope())
+            {
+                testSubject.Accept("wrong file", issues);
+            }
+
+            callbackSpy.CallCount.Should().Be(0);
+        }
+
+        [TestMethod]
+        [DataRow(-1, 10, false)] // start line < 1
```
rita-g-sonarsource, line 72: I think it's worth giving a different number than 10 for the last line, just so it would be more readable in terms of zero-based/one-based. (-1, 1) for example.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
```diff
@@ -0,0 +1,182 @@
...
+        [DataRow(0, 0, false)] // file-level issue, can't be mapped to snapshot
```
rita-g-sonarsource, line 73: Is it worth checking (0,1)?

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -101,6 +104,31 @@ private void CheckFactoryWasUnregisteredFromDataSource(SnapshotFactory factory,
...
+            // Need to register a tagger for the buffer to start listening to buffer change events
+            using (testSubject.CreateTagger())
+            {
+                mockAnalyzerController.Invocations.Clear();
+                mockSonarErrorDataSource.Invocations.Clear();
+
+                RaiseBufferChangedEvent(mockDocumentTextBuffer, beforeSnapshot.Object, afterSnapshot.Object);
+
+                CheckAnalysisWasNotRequested();
+                CheckErrorListRefreshWasRequested(1);
```
rita-g-sonarsource, line 128: This is unrelated and was here before, but maybe it's worth changing anyway if you're doing other changes: CheckErrorListRefreshWasRequested is always accepting 1 as an argument, so it's redundant. I had to go and look what it's doing, as I thought it was number of expected issues or something.

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
```diff
@@ -0,0 +1,117 @@
...
+using Microsoft.VisualStudio.Text;
+using SonarLint.VisualStudio.Core.Analysis;
+
+/* 
+ * Instancing: a new instance of this class should be created for each analysis request.
+ * 
+ * The class is initialized with the text snapshot representing the state of the text buffer
+ * at the point the analysis was triggered ("the analysis snaphost").
+ * 
+ * Each time IIssueConsumer.Accept is called, the new issues will be mapped back to the
```
rita-g-sonarsource, line 34: Maybe "mapped back to the _given_ analysis snapshot"?

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
```diff
@@ -0,0 +1,117 @@
...
+ * analysis snapshot and decorated with the additional data required for filtering and tagging.
+ * Then, all of the issues that have been received that habe been received so far will be
```
rita-g-sonarsource, line 36: habe = typo

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
```diff
@@ -0,0 +1,117 @@
...
+ * passed to the OnIssuesChanged delegate.
+ * 
+ * However, it's possible that the text buffer could have been edited since the analysis
+ * was triggered. It is the responsibity of the callback to translate to supplied IssueMarkers
```
rita-g-sonarsource, line 40: "to translate _the_ supplied"?

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
```diff
@@ -0,0 +1,117 @@
...
+ * to the current snapshot, if necessary.
+ * 
+ * 
+ * Mapping from reported issue line/char positions to the analysis snapshot
+ * ------------------------------------------------------------------------
+ * Currently, the JS/CFamily analyzers run against the file on disc, not the content of 
+ * the snapshot. We're making a couple of assumptions:
+ * 1) we're triggering analysis on save, so the state of the text buffer matches the file on disc, and
+ * 2) the file won't have had time to change again before the analyzer reads it so we expect always to
```
rita-g-sonarsource, line 49: "the file won't have had time to change again before the analyzer reads it" -- not necessarily true, but we would've cancelled the analysis and started a new one. 
Hm, maybe it's better to remove this line. And just say that we trigger on save, so the text buffer matches the file on disk at the time the analysis is triggered.

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
duncanp-sonar: To me that's more confusing - what "the analysis snapshot" is is defined in line 32. Saying the "given analysis snapshot" suggests there could be more than one "analysis snapshot".

I'll drop the reference to "analysis snapshot" and change it to "supplied snapshot"

rita-g-sonarsource: "supplied" sounds good.

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
duncanp-sonar: Yep, it is odd. I considered changing it, but as you say it's unrelated so I left it for another day.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
duncanp-sonar: I don't understand, but ok.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
duncanp-sonar: That doesn't represent a meaningful range i.e. an invalid response from the analyzer. I did have it before and took it out, but it's probably worth checking.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
duncanp-sonar: I didn't think it was clear enough when writing the test - I wanted it to be clear that we were comparing a 1-based thing against a 0-based thing.
I've change the names and added a comment instead.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
duncanp-sonar: Here, we don't know in advance the expected IssueMarker instances since they are created by the product code. We only know the underlying IAnalysisIssues that we expect the IssueMarkers to wrap.

Checking the call count here with mocks would be easy, but doing the more complicated check on the expected issues isn't.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
duncanp-sonar: I've changed it to `CreatePassthroughConverter`

PR #1523, opened 2020-07-02
Ignore issues without filename

### LINKED ISSUE #1489: Issues that don't report a filename are not handled correctly
### Description

Some of the C and C++ rules can be triggered without a specific filename. For example, S3744 will be triggered if when a macro is defined multiples times with different values as a compiler option. This rule will not have an associated filename since the issue doesn't exist in a specific location of the file.

When it happens, the subprocess.exe finishes successfully, however it returns an issue without filename, which causes SLVS to crash.

### Repro steps

Create a C++ project and define a macro to two different values using additional options:

image
When analyzing the file you will get a crash while reading the analysis results:
image

### Expected behavior

Handle rules without an associated file. Maybe display them a warning without mapping them to a specific location
### Actual behavior

No issues are reported for the file (it's also possible no issues will be reported for the entire project, depending on which rule was triggered. See explanation from @abbas-sabra-sonarsource below.).
A message is logged in the SonarLint output window.

### Known workarounds

Disable the rule causing the issue from the setting file:
image

rita-g-sonarsource:
Fixes #1489

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/ProtocolTest.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/ProtocolTest.cs
@@ -72,6 +72,28 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         }
 
         [TestMethod]
+        public void Read_RequestHasFileName_ReturnedMessageHasNoFileName_MessageIgnored()
+        {
+            const string returnedIssueFileName = "";
+            const string testedFileName = "myfile.cpp";
+
+            var response = CallProtocolRead(MockResponse(returnedIssueFileName), testedFileName);
+
+            response.Messages.Length.Should().Be(0);
+        }
+
+        [TestMethod]
+        public void Read_RequestHasNoFileName_ReturnedMessageHasNoFileName_MessageReturned()
+        {
+            const string returnedIssueFileName = "";
+            const string testedFileName = "";
+
+            var response = CallProtocolRead(MockResponse(returnedIssueFileName), testedFileName);
+
+            response.Messages.Length.Should().Be(1);
+        }
+
+        [TestMethod]
         public void Read_Response()
         {
             var response = CallProtocolRead(MockResponse());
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/CFamily/PortedFromJava/Protocol.cs
```diff
@@ -177,11 +176,14 @@ public static void Read(BinaryReader reader, Action<Message> handleIssue, string
                   throw new InvalidDataException("Communication issue with the C/C++ analyzer");
               case "message":
                   var message = readMessage(reader);
-
-                  if (!doFilterResults || string.Equals(issueFilePath, Path.GetFullPath(message.Filename), StringComparison.InvariantCultureIgnoreCase))
+                  
+                  if (!doFilterResults || 
+                      !string.IsNullOrEmpty(message.Filename) &&
```
duncanp-sonar, line 181: When combining || and && expressions it's safer to add brackets to make the grouping explicit, rather than relying on operator precendence.

src/Integration.Vsix.UnitTests/CFamily/ProtocolTest.cs
```diff
@@ -71,6 +71,28 @@ public void Read_Empty_Response()
...
+        public void Read_RequestHasNoFileName_ReturnedMessageHasNoFileName_MessageReturned()
+        {
+            const string returnedIssueFileName = "";
+            const string testedFileName = "";
+
+            var response = CallProtocolRead(MockResponse(returnedIssueFileName), testedFileName);
+
+            response.Messages.Length.Should().Be(1);
+        }
+
```
duncanp-sonar, line 95: Are there already tests for the other two cases? i.e. both have same file name, both have different file names

rita-g-sonarsource: Yep. Same name is "Response_Filtering_WithFiltering", and different name is "Response_Filtering_NoMatches"

PR #1525, opened 2020-07-03
Change CLangAnalyzer to stream issues

### LINKED ISSUE #1471: Display CFamily analysis issues incrementally

duncanp-sonar:
Change CLangAnalyzer to stream issues

* removed message about suppressed issues

Fixes #1471

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
@@ -165,6 +165,62 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         }
 
         [TestMethod]
+        public void TriggerAnalysisAsync_IssuesForInactiveRulesAreNotStreamed()
+        {
+            const string fileName = "c:\\data\\aaa\\bbb\file.txt";
+            var rulesConfig = new DummyCFamilyRulesConfig("c")
+                .AddRule("inactiveRule", isActive: false)
+                .AddRule("activeRule", isActive: true);
+
+            var request = new Request
+            {
+                File = fileName,
+                RulesConfiguration = rulesConfig
+            };
+
+            var inactiveRuleMessage = new Message("inactiveRule", fileName, 1, 1, 1, 1, "inactive message", false, Array.Empty<MessagePart>());
+            var activeRuleMessage = new Message("activeRule", fileName, 2, 2, 2, 2, "active message", false, Array.Empty<MessagePart>());
+
+            var mockConsumer = new Mock<IIssueConsumer>();
+            var subProcess = new SubProcessSimulator();
+
+            var testSubject = new TestableCLangAnalyzer(telemetryManagerMock.Object, new ConfigurableSonarLintSettings(),
+                rulesConfigProviderMock.Object, serviceProviderWithValidProjectItem.Object, testLogger);
+            testSubject.SetCallSubProcessBehaviour(subProcess.CallSubProcess);
+
+            try
+            {
+                // Call the CLangAnalyzer on another thread (that thread is blocked by subprocess wrapper)
+                var analysisTask = Task.Run(() => testSubject.TriggerAnalysisAsync(request, mockConsumer.Object, CancellationToken.None));
+                subProcess.WaitUntilSubProcessCalledByAnalyzer();
+
+                // Stream the inactive rule message to the analyzer
+                subProcess.PassMessageToCLangAnalyzer(inactiveRuleMessage);
+                mockConsumer.Verify(x => x.Accept(fileName, It.IsAny<IEnumerable<IAnalysisIssue>>()), Times.Never);
+
+                // Now stream an active rule message
+                subProcess.PassMessageToCLangAnalyzer(activeRuleMessage);
+
+                mockConsumer.Verify(x => x.Accept(fileName, It.IsAny<IEnumerable<IAnalysisIssue>>()), Times.Once);
+                var suppliedIssues = (IEnumerable<IAnalysisIssue>)mockConsumer.Invocations[0].Arguments[1];
+                suppliedIssues.Count().Should().Be(1);
+                suppliedIssues.First().Message.Should().Be("active message");
+
+                // Tell the subprocess mock there are no more messages and wait for the analyzer method to complete
+                subProcess.SignalNoMoreIssues();
+                bool succeeded = analysisTask.Wait(10000);
+                succeeded.Should().BeTrue();
+
+                testLogger.AssertOutputStringExists($"Found 1 issue(s) for {fileName}");
+            }
+            finally
+            {
+                // Unblock the subprocess wrapper in case of errors so it can finish
+                subProcess.SignalNoMoreIssues();
+            }
+        }
+
+        [TestMethod]
         public void TriggerAnalysisAsync_AnalysisIsCancelled_NoIssuesFoundMessage()
         {
             var mockConsumer = new Mock<IIssueConsumer>();
--- a/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
@@ -19,7 +19,10 @@
  */
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using EnvDTE;
 using FluentAssertions;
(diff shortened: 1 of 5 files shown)


src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -99,33 +99,35 @@ private async Task TriggerAnalysisAsync(Request request, IIssueConsumer consumer
...
-
-            var issues = messages
-                .Where(m => IsIssueForActiveRule(m, request.RulesConfiguration))
-                .Select(m => CFamilyHelper.ToSonarLintIssue(m, request.CFamilyLanguage, request.RulesConfiguration))
-                .ToList();
-
             telemetryManager.LanguageAnalyzed(request.CFamilyLanguage); // different keys for C and C++
 
-            logger.WriteLine($"Found {issues.Count} issue(s)");
+            logger.WriteLine($"Found {issueCount} issue(s)");
```
rita-g-sonarsource, line 112: I think it's worth adding the fileName into the message. There could be multiple analysis running at the same time.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -271,9 +271,6 @@ private IEnumerable<IssueMarker> RemoveSuppressedIssues(IEnumerable<IssueMarker>
             var filteredIssues = issuesFilter.Filter(filterableIssues);
             Debug.Assert(filteredIssues.All(x => x is IssueMarker), "Not expecting the issue filter to change the list item type");
 
-            var suppressedCount = filterableIssues.Count() - filteredIssues.Count();
```
rita-g-sonarsource, line 274: If we have an easy/elegant way to identify the issue, we could output a generic "Issue X has been suppressed"

duncanp-sonar: Even if there was, we'd see the message multiple times, in the same way we say the suppressed issue count message multiple times.

rita-g-sonarsource: Ah right.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -89,7 +89,13 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
             TriggerAnalysisAsync(request, consumer, cancellationToken)
                 .Forget(); // fire and forget
 
-        private async Task TriggerAnalysisAsync(Request request, IIssueConsumer consumer, CancellationToken cancellationToken)
+        protected /* for testing */ virtual void CallSubProcess(Action<Message> handleMessage, Request request, IAnalysisStatusNotifier analysisStatusNotifier, ISonarLintSettings settings,
```
duncanp-sonar, line 92: I originally define a delegate, but it looked weird to have a mix of delegates and `virtual override` points in the class (the existing two virtual methods are defined immediately above this one).
I also looked at creating an interface, but that involved more change. This was the minimum change I could make to the product code to make the behaviour testable.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -137,6 +241,51 @@ protected override void TriggerAnalysis(Request request, IIssueConsumer consumer
...
+                    base.CallSubProcess(handleMessage, request, analysisStatusNotifier, settings, logger, cancellationToken);
+                }
+                else
+                {
+                    onCallSubProcess(handleMessage, request, analysisStatusNotifier, settings, logger, cancellationToken);
+                }
+            }
+        }
+
+        private class SubProcessSimulator
```
duncanp-sonar, line 258: This class is ugly but necessary of the way the CLangAnalyzer expects the stream handler to not return until all of the issues have been received.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -100,6 +104,99 @@ public void ExecuteAnalysis_ValidProjectItem_RequestCanBeCreated_AnalysisIsTrigg
             testSubject.TriggerAnalysisCallCount.Should().Be(1);
         }
 
+        [TestMethod]
+        public void TriggerAnalysisAsync_StreamsIssuesFromSubProcessToConsumer()
```
duncanp-sonar, line 108: The new tests have to call the analyzer on a different thread from the test thread, and then selectively block/wait as the messages are processed.
As far as I can tell, the tests are reliable (i.e. repeatably run correctly without flakiness) and also clean up properly in the event of errors.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -99,33 +105,38 @@ private async Task TriggerAnalysisAsync(Request request, IIssueConsumer consumer
...
+                logger.WriteLine($"Found {issueCount} issue(s) for {request.File}");
+            }
+        }
+
+        private void HandleMessage(Message message, Request request, IIssueConsumer consumer, ref int issueCount)
+        {
+            Debug.Assert(message.Filename == request.File, $"Issue for unexpected file returned: {message.Filename}");
+            if (!IsIssueForActiveRule(message, request.RulesConfiguration))
+            {
+                return;
```
duncanp-sonar, line 129: According to SonarCloud there are 3uncovered lines of code out of 7, giving 57% coverage. This is one of the uncovered lines, which I can add a test for quite easily.

The other two are for the new virtual method declaration and the one-line passthough call to CFamilyHelper. Testing that would require an integration tests, which I'm not proposing to do at this point.

rita-g-sonarsource: I agree. It's worth testing the non-active rule logic.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -137,6 +241,51 @@ protected override void TriggerAnalysis(Request request, IIssueConsumer consumer
...
+                // Store the callback passed to us from the CLangAnalyzer
+                handleMessageCallback = handleMessage;
+
+                // Tell the calling test we're ready and the test can continue
+                callbackFromCLangReceived.Set();
+
+                // Block until the test tells us we can finish
+                noMoreIssues.WaitOne();
+            }
+
```
rita-g-sonarsource, line 280: I feel stupid. Why do we need this?
Why can't we just use `noMoreIssues`?

rita-g-sonarsource:
LGTM

PR #1526, opened 2020-07-03
Change message to debug

rita-g-sonarsource:
This message provides no value to the user, and might even be scary.


src/Integration.Vsix/CFamily/CFamilyStrings.resx
```diff
@@ -130,7 +130,7 @@
     <value>&lt;sensitive data removed&gt;</value>
   </data>
   <data name="MSG_ExectutionCancelledKilled" xml:space="preserve">
-    <value>Process {0} was cancelled and killed.</value>
+    <value>Process {0} was killed.</value>
```
duncanp-sonar, line 133: You could fix the name of resource (typo and still refers to "cancelled"): `MSG_ExectutionCancelledKilled`

PR #1535, opened 2020-07-15
Fix MD5 FIPS compliance

rita-g-sonarsource:


src/Integration/Suppression/ManagedMD5.cs
```diff
@@ -0,0 +1,461 @@
+﻿/* Copyright (C) David Anson
+ * https://dlaa.me/blog/post/9380245
```
duncanp-sonar, line 2: What, if anything, needs to go into the `THIRD-PARTY-NOTICES` file?

src/Integration/Suppression/ChecksumCalculator.cs
```diff
@@ -41,7 +40,7 @@ public static string Calculate(string text)
             }
 
             string content = Regex.Replace(text, "\\s", ""); // strip whitespace
-            using (MD5 md5Digest = MD5.Create())
+            using (var md5Digest = new Md5Managed())
```
duncanp-sonar, line 43: As previously discussed, what validation do you think we need to do to satisfy ourselves that the algorithm is is correctly implemented?

rita-g-sonarsource: We are using it for suppression matching so I think we should focus on that. I tried C++/C# suppressions, long lines, short lines, variables, comments. I'll try unicode characters. Anything else do you think we should test?

src/Integration/Suppression/ManagedMD5.cs
rita-g-sonarsource: I don't see a copyright in his project/site that I can point to. 
How about this:
```
License notice for MD5 Algorithm
-----------------------------------------------------
Copyright (C) David Anson 
https://dlaa.me/blog/post/9380245
```

src/Integration/Suppression/ChecksumCalculator.cs
duncanp-sonar: Those sound like the right sort of things to test. I was expecting to see some unit tests comparing the results against the .NET implementation.
Did you consider using an existing NuGet package like https://www.nuget.org/packages/BouncyCastle/ or https://www.nuget.org/packages/CryptSharpOfficial/ that have seem to be popular?

src/Integration/Suppression/ManagedMD5.cs
duncanp-sonar: Seems ok. What about the RSA license notice from the comments in lines 10-44?

PR #1543, opened 2020-07-29
Update readme with VS2019

andrei-epure-sonarsource:


rita-g-sonarsource:
Thanks @andrei-epure-sonarsource !

Issue #1209, opened 2020-01-29
Binding fails with "SonarQube request failed: Failed to write credentials"


duncanp-sonar:
### Description

See the following two community forum threads: [here](https://community.sonarsource.com/t/error-binding-vs-2017-solution-failed-to-write-credentials/3500) and [here](https://community.sonarsource.com/t/not-able-to-connect-to-sonarqube-7-7-from-visual-studio-2017-using-sonarlint-4-10/11468).

### Expected behavior

Credentials should be saved successfully and the binding process completed.

### Actual behavior

The binding process failed.

### Known workarounds

None

### Related information

* SonarLint for VS Version: 4.6.
* Visual Studio Version: ? 
* Languages in your solution; C#

### Notes
We've been unable to repro the issue locally.

The `Failed to write credentials` error message is thrown by the `Microsoft.Alm.Authentication` component as the result of a failed Win32 API call. The API is throwing `Win32Exception` which should have the HRESULT error code. Unfortunately, we aren't logging that part so we're losing the Windows error code.

In the absence of a repro, we could amend the logging to capture the Windows error code to give us a better idea of what the underlying issue might be.

RogerW-VZC:
Same error, VS2019  16.4.6.

I get one additional line before the "Failed to write credentials" error.
"Ruleset binding: all projects need to be updated reference the generated ruleset."

Every time the binding tries to run, an additional ruleset file is added to every project.

I have just deleted the .sonarlint and .sonarqube folders, etc  as recommended https://groups.google.com/forum/#!topic/sonarlint/_hgsvV6WHEU because I needed to switch from a localhost sonarqube to our corporate one.

PR #1544, opened 2020-07-29
Log Win32 error code when Win32Exception is thrown

rita-g-sonarsource:
Relates to #1209, provides more information to the user when an exception occurs.

### DIFF
--- a/src/Core/CoreStrings.Designer.cs
+++ b/src/Core/CoreStrings.Designer.cs
@@ -88,6 +88,15 @@ namespace SonarLint.VisualStudio.Core {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Failed to {0} credentials, Win32ErrorCode: {1}. See https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes for more information..
+        /// </summary>
+        public static string CredentialStore_Win32Error {
+            get {
+                return ResourceManager.GetString("CredentialStore_Win32Error", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to C#.
         /// </summary>
         public static string CSharpLanguageName {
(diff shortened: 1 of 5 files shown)


rita-g-sonarsource:
@duncanp-sonar I finally managed to reproduce this issue:
image

I got a hint from [this](https://github.com/microsoft/Git-Credential-Manager-for-Windows/issues/502#issuecomment-339673118) comment on a similar issue, and I did manage to reproduce this by impersonating. 
It's probably not the root cause that was reported by the users, but it was still useful as a way to test the exception logging.
Here's my sample code if you want to play with it, just create another user on your machine:

[CredentialStoreFailedToWriteCredentials.zip](https://github.com/SonarSource/sonarlint-visualstudio/files/4996658/CredentialStoreFailedToWriteCredentials.zip)

rita-g-sonarsource:
Small update: 
1. It looks like the latest version of Microsoft.Alm.Authentication now swallows this exception. 
[This](https://github.com/microsoft/Git-Credential-Manager-for-Windows/blob/16e6052fba0aadbd51c8eacef6c3a9e09db65cfc/Microsoft.Alm.Authentication/BaseSecureStore.cs#L207) is the version we're using, which throws "Failed to write credentials", and [this](https://github.com/microsoft/Git-Credential-Manager-for-Windows/blob/master/Microsoft.Alm.Authentication/Src/BaseSecureStore.cs#L170) is the latest version, which swallows it. For future reference: we should  stick to the old version and not upgrade, otherwise we won't know at all what's the problem.
2. The [code](https://github.com/microsoft/Git-Credential-Manager-for-Windows/blob/16e6052fba0aadbd51c8eacef6c3a9e09db65cfc/Microsoft.Alm.Authentication/NativeMethods.cs#L50) uses [CredWriteW](https://docs.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-credwritew). The documentation lists a few possible codes that it can return, and in my impersonation example I indeed get ERROR_NO_SUCH_LOGON_SESSION (1312), similarly to [this stack overflow question](https://stackoverflow.com/questions/18501250/credwrite-returning-1312-when-called-under-impersonation). 
3. I think all we can do for now is just log the exception, and try to use the error code to help the users next time it happens.

Issue #1312, opened 2020-04-06
[CFamily] Investigate removal of reflection/dynamic access when fetching config


duncanp-sonar:
### Description
The CFamily code uses a mixture of Reflection and `dynamic` code when fetching the compiler settings from the VC engine. This makes writing and debugging code more complicated, and means the compiler isn't helping to catch errors. It also means that when the code is changed it needs to be manually tested against multiple versions of VS.

It is not clear why this is being done, instead of referencing the required assemblies directly.
It's possible that the VC assemblies do not following the same versioning/publishing pattern as other VS binaries (e.g. might not be backward-compatible / not available via NuGet)

### TODO
* investigate whether the code can be changed to use static references to the required assemblies.

duncanp-sonar:
### Notes
The VC binaries are not available via NuGet (there is one unofficial package for VS2015, but that's all)

We use types defined in the `Microsoft.VisualStudio.VCProjectEngine.dll` assembly. The assembly name is the same across VS versions, but the file version changes.

For each VS version, the interfaces are defined in `Common7\IDE\PublicAssemblies\Microsoft.VisualStudio.VCProjectEngine.dll`, and implemented in `Common7\IDE\CommonExtensions\Microsoft\VC\Project\Microsoft.VisualStudio.Project.VisualC.VCProjectEngine.dll`.
The interfaces assembly is marked as a Primary Interop Assembly (i.e. contains COM type definitions). However, all of the implementation assemblies also export the implementation classes via MEF (to hook into bits of the project system),

However, there are no binding redirects for the assembly in VS2017. There is a binding redirect in VS2019, but only from v15 to v16 (i.e. from VS2107 to VS2019, not from VS2015).

### Summary
There is no type-compatibility across all three versions of VS, so we can't just reference a single version of the assembly/NuGet package and have it work across all three versions of VS as we do the VS SDK assemblies. It looks like the VS2017 and VS2019 versions might be type-compatible, so once we've dropped support for VS2015 we might be able reference the VS2017 version. However, the assembly isn't available as a NuGet package so we'd need to check in a reference copy to build against.

In the meantime, our only way to have strongly-typed code would be to use the same approach that we do for the Team Explorer assemblies i.e.  use VS-version-specific references and build a different version of our assembly for each version of VS.
In fact, we reference the VCProject assembly from our `Integration.Vsix` project, which already produces a different assembly for each version of VS.

The alternative is to so stick with using `dynamic`.

PR #1545, opened 2020-07-29
Use statically-typed VC interfaces when fetching CFamily configuration

### LINKED ISSUE #1537: Platform is not considered when detecting compiler options
### Description

In visual studio, you can compile for different platforms: x64, x86. Compiler options can be different for each platform. In that case, the wrong options will be detected.
### Repro steps

1.  For Release x86 set language standard to c++14

2. For Release x64 set language standard to c++17

3. When analyzing in x64 mode, SLVS is detecting c++14 instead of c++17

### Expected behavior

detecting the right option for each platform

### Related information
In `FileConfig.cs`, Replacing `dynamic fileConfig = file.FileConfigurations.Item(configurationName);` by `dynamic fileConfig = file.FileConfigurations.Item(pattern);`  should fix the problem.

duncanp-sonar:
Fixes #1537 
Relates to #1312


src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -267,11 +267,14 @@ internal static FileConfig TryGetConfig(ILogger logger, ProjectItem projectItem,
...
             {
+                // Note: if the C++ tools are not installed then it's likely an exception will be thrown when
+                // the framework tries to JIT-compile the TryGet method (since it won't be able to find the MS.VS.VCProjectEngine
+                // types).
                 return FileConfig.TryGet(projectItem, absoluteFilePath);
             }
             catch (Exception e)
             {
-                logger.WriteLine($"Unable to collect C/C++ configuration for {absoluteFilePath}: {e.ToString()}");
+                logger.WriteLine($"Unable to collect C/C++ configuration for {absoluteFilePath}: {e}");
```
duncanp-sonar, line 277: Miscellaneous: remove code smell (complaining about unnecessary `ToString`)

src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -267,11 +267,14 @@ internal static FileConfig TryGetConfig(ILogger logger, ProjectItem projectItem,
 
             try
             {
+                // Note: if the C++ tools are not installed then it's likely an exception will be thrown when
+                // the framework tries to JIT-compile the TryGet method (since it won't be able to find the MS.VS.VCProjectEngine
+                // types).
```
duncanp-sonar, line 272: The previous version of the code would have failed too if the C++ tools weren't installed, but the error type would have been different.

src/SonarLint.VSSpecificAssemblies.props
```diff
@@ -143,4 +143,16 @@
...
 
+
+  <!-- ******************************************************************************* -->
+  <!-- VCProject assembly -->
+  <!-- ******************************************************************************* -->
+  <ItemGroup Label="VCProjectEngine" Condition="$(RequiresVCProjectEngine)=='true'">
+    <Reference Include="Microsoft.VisualStudio.VCProjectEngine">
+      <HintPath>$(ThirdPartyPath)\Microsoft.VisualStudio.VCProjectEngine.dll</HintPath>
+      <Private>$(CopyVSAssembliesToOutput)</Private>
+    </Reference>
```
duncanp-sonar, line 154: Same pattern as for the TeamExplorer assemblies - three versions of the assembly are checked in, and we pick the one to build against depending on the version of VS being targetted.

src/Integration.Vsix.UnitTests/CFamily/FileConfigTests.cs
```diff
@@ -0,0 +1,62 @@
...
+using Moq;
+using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
+{
+    [TestClass]
+    public class FileConfigTests
+    {
+        [TestMethod]
+        public void TryGet_NoVCProject_ReturnsNull()
```
duncanp-sonar, line 34: I've added this new file for FileConfig tests, and added two tests for the new conditions of "no VCFile" and "no VCProject".

Quite a few of the tests in `CFamilyHelperTests.cs` actually just testing the nested `FileConfig` class. I'll move them in a separate PR and do some refactoring/clean-up of the tests.

src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -35,19 +35,31 @@ internal class FileConfig
...
+                var vcProject = dteProjectItem.ContainingProject.Object as VCProject;
+                var file = dteProjectItem.Object as VCFile;
+                if (vcProject == null || file == null)
+                {
+                    return null;
+                }
 
-                dynamic project = projectItem.ContainingProject.Object; // Microsoft.VisualStudio.VCProjectEngine.VCProject
-                dynamic config = project.Configurations.Item(pattern); // Microsoft.VisualStudio.VCProjectEngine.VCConfiguration
+                var vcConfig = vcProject.ActiveConfiguration;
```
duncanp-sonar, line 47: As discussed offline, this fixes #1537 - we're using the API to get the active configuration rather than trying to fetch it based on a name that we calculate from other properties.

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -853,86 +871,53 @@ private Request GetSuccessfulRequest(IAnalyzerOptions analyzerOptions)
...
 
-        private static void AssertPartialMessageLogged(Mock<ILogger> loggerMock, string message)
-        {
-            loggerMock.Verify(x => x.WriteLine(It.Is<string>(
-                s => s.Contains(message))), Times.Once);
-        }
+            var vcConfigMock = new Mock<VCConfiguration>();
+            vcConfigMock.SetupGet(x => x.Platform).Returns(vcPlatformMock.Object);
 
-        private class DummyProjectEngine : IVCRulePropertyStorage
```
duncanp-sonar, line 868: We're no longer using reflection/dynamic code to fetch properties, so we can just use normal Mocks of the real product interfaces.

src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -475,11 +463,6 @@ public Request ToRequest(string path)
                         return "/std:c++14";
                 }
             }
-
-            internal /* for testing */ static void ClearGetEvaluatedPropertyValueMethod()
```
rita-g-sonarsource, line 479: 🎉

src/SonarLint.VSSpecificAssemblies.props
```diff
@@ -143,4 +143,16 @@
     <StrongNamerKeyFile>$(AssemblyOriginatorKeyFile)</StrongNamerKeyFile>
   </PropertyGroup>
 
+
+  <!-- ******************************************************************************* -->
+  <!-- VCProject assembly -->
+  <!-- ******************************************************************************* -->
+  <ItemGroup Label="VCProjectEngine" Condition="$(RequiresVCProjectEngine)=='true'">
```
rita-g-sonarsource, line 150: Out of curiosity, why not always add the assembly? Why have the "RequiresVCProjectEngine" condition?

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -39,11 +39,25 @@ public class CFamilyHelperTests
     {
         private const string FileName = @"C:\absolute\path\to\file.cpp";
 
-        [TestInitialize]
-        public void TestInitialize()
+        private const string ValidPlatformName = "Win32";
+
+        private static readonly IDictionary<string, string> MandatoryConfigProperties = new Dictionary<string, string>
```
rita-g-sonarsource, line 44: Maybe MandatoryProjectConfigProperties ?

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -159,63 +173,53 @@ public void FileConfig_Test2()
         public void GetEvalutedPropertyValue_NoException_ReturnsValue()
```
rita-g-sonarsource, line 173: Test name is incorrect, should be something like 'GetPotentiallyUnsupportedPropertyValue_PropertySupported_ReturnsValue'

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -159,63 +173,53 @@ public void FileConfig_Test2()
...
+                CFamilyHelper.FileConfig.GetPotentiallyUnsupportedPropertyValue(settingsMock.Object, "propertyName1",
                     "default xxx");
 
             // Assert
             result.Should().Be("propertyValue");
-            suppliedPropertyName.Should().Be("propertyName1");
         }
 
         [TestMethod]
         public void GetEvalutedPropertyValue_Exception_ReturnsDefault()
```
rita-g-sonarsource, line 190: Test name is incorrect, should be something like 'GetPotentiallyUnsupportedPropertyValue_PropertyUnsupported_ReturnsDefaultValue'

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -523,8 +527,8 @@ public void CreateRequest_ErrorGetting_IsHandled()
             var loggerMock = new Mock<ILogger>();
 
             var projectItemMock = CreateProjectItemWithProject("c:\\foo\\xxx.vcxproj");
-            var vcProjectMock = projectItemMock.Object.Object as VCProjectMock;
-            vcProjectMock.ActiveConfiguration = null;
+            // NB we want the exception to be thrown from inside the FileConfig::TryGet
```
rita-g-sonarsource, line 530: NB ?

rita-g-sonarsource, line 530: I think it's better to rename this test instead of the comment, e.g. 
CreateRequest_ExceptionInFileConfig_IsHandled

rita-g-sonarsource:
@duncanp-sonar LGTM, left some minor comments

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
duncanp-sonar: https://en.wikipedia.org/wiki/Nota_bene

src/SonarLint.VSSpecificAssemblies.props
duncanp-sonar: So we know which projects require the assembly and which ones don't. It's a bad practice to reference assemblies that aren't required, and this is a common settings file that is imported in multiple projects.

PR #1550, opened 2020-07-30
Handle analysis for files that have no associated project

rita-g-sonarsource:


rita-g-sonarsource:
@duncanp-sonar FYI, we do start analysis when file is opened, due to the fact that we start analysis each time a tagger is created, and VS creates one when a file is opened.

I'm not sure what is exactly the root cause of the issue reported by the user, as they claim it happens only after a few hours. 
Which means that it's unlikely that they were opening the file, as opening the file would've caused the exception immediately (due to the triggered analysis).
So perhaps the file was opened, left open, and then renamed / moved / somehow become disassociated with the project, which then caused the exception.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -79,20 +78,27 @@ internal class TextBufferIssueTracker : IIssueTracker
...
+            this.Factory = new SnapshotFactory(new IssuesSnapshot(ProjectName, FilePath, 0, new List<IssueMarker>()));
 
             document.FileActionOccurred += SafeOnFileActionOccurred;
         }
 
+        private string ProjectName
+        {
+            get
+            {
+                // Bug #676: https://github.com/SonarSource/sonarlint-visualstudio/issues/676
```
rita-g-sonarsource, line 92: This issue was actually happening again. This PR re-fixes it as well.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -79,20 +78,27 @@ internal class TextBufferIssueTracker : IIssueTracker
...
-            // e.g. files under the "External Dependencies" project folder in the Solution Explorer
-            var projectName = this.ProjectItem?.ContainingProject.Name ?? "{none}";
-            this.Factory = new SnapshotFactory(new IssuesSnapshot(projectName, this.FilePath, 0,
-                new List<IssueMarker>()));
+            this.Factory = new SnapshotFactory(new IssuesSnapshot(ProjectName, FilePath, 0, new List<IssueMarker>()));
 
             document.FileActionOccurred += SafeOnFileActionOccurred;
         }
 
+        private string ProjectName
```
duncanp-sonar, line 88: I think this should be a `GetProjectName` method rather than a property. I wouldn't expect a property to do this much processing when it's called. Also, this code is actually thread-sensitive i.e. we should be calling the `dte` on the UI thread., which I wouldn't expect to be done in property.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -276,13 +281,13 @@ private IEnumerable<IssueMarker> RemoveSuppressedIssues(IEnumerable<IssueMarker>
...
-            UpdateIssues(this.Factory.CurrentSnapshot.IssueMarkers ?? Enumerable.Empty<IssueMarker>());
+            UpdateIssues(Factory.CurrentSnapshot.IssueMarkers ?? Enumerable.Empty<IssueMarker>());
         }
 
         private void UpdateIssues(IEnumerable<IssueMarker> issueMarkers)
         {
-            var oldSnapshot = this.Factory.CurrentSnapshot;
-            var newSnapshot = new IssuesSnapshot(this.ProjectItem.ContainingProject.Name, this.FilePath, oldSnapshot.VersionNumber + 1, issueMarkers);
+            var oldSnapshot = Factory.CurrentSnapshot;
+            var newSnapshot = new IssuesSnapshot(ProjectName, FilePath, oldSnapshot.VersionNumber + 1, issueMarkers);
```
duncanp-sonar, line 290: The `ProjectItem` instance variable should now be redundant and can be removed.

rita-g-sonarsource: Already did

PR #1569, opened 2020-08-05
Raise document focused event

### LINKED ISSUE #1559: Implement file selection tracker that detects when a code file receives focus
### Description

Listen to VS selection events to determine when a selection change has occurred, filtering out changes we are not interested in.

We are only interested known when an editor containing a file has been activated. This includes switching focus between open documents, or from a a tool window to a file editor.

We are not interested in knowing when documents are closed or lose focus.

TODO:
* check what happens when a file is opened for the first time - presumably the selection changed event will also fire?
* check what happens when multiple files are opened in a batch operation e.g. search and replace.

Relates to #1557

rita-g-sonarsource:
Fixes #1559

### DIFF
--- a/src/Integration.Vsix.UnitTests/Helpers/DocumentFocusedEventRaiserTests.cs
+++ b/src/Integration.Vsix.UnitTests/Helpers/DocumentFocusedEventRaiserTests.cs
@@ -19,8 +19,6 @@
  */
 
 using System;
-using System.Collections.Generic;
-using System.Linq;
 using FluentAssertions;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell.Interop;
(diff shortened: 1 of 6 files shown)


rita-g-sonarsource:
@duncanp-sonar I suggest taking a look at the two commits separately:
1. The first commit is the Listener implementation, which uses `AdviseSelectionEvents`. Coincidentally, my implementation ended up very similar to [MS' VS-Macros](https://github.com/microsoft/VS-Macros/blob/7e9484341d7f61ad3ce8cd71c87568e3562f9194/VSMacros/RecorderListeners/WindowActivationWatcher.cs#L53), which seems to be doing a similar thing.
2. The second commit is UI threading handling, as we discussed. I took a look at how [roslyn do it](https://github.com/dotnet/roslyn/blob/83a6d3ae12a54b3c9ddd830f1318bc9912eb06b2/src/VisualStudio/Core/Def/Implementation/Workspace/VisualStudioActiveDocumentTracker.cs#L55). Interestingly, they don't call `UnadviseSelectionEvents`, don't know why.

General notes:
* I originally extracted RunOnUIThread in a separate branch, but thought it was small enough to fit here. But I can do it in a different PR for cleanliness, let me know what you think.
* The OnDocumentFocused event is popping up (naturally) quite a lot, i.e. each time that you click on Solution Explorer and then go back to the file. I guess that since we will cancel the previous PCH it would be ok in terms of machine resources.

Issue #1560, opened 2020-08-03
Trigger pre-analysis build when a code file is activated


duncanp-sonar:
### Description

Depends on #1559. 

- [x] listen to "file activated" events
- [x] trigger a pre-analysis request

### TODO
* do we need to do anything special to filter out files we don't support analysis for (e.g. C# files), or should we just make the normal "IsAnalysisSupported" call to the analyzer controller?
* directly call the CLang analyzer?
* decide PCH file name to pass to subprocess.exe - specific to the current instance of VS? (#1574)

Relates to #1557

PR #1570, opened 2020-08-05
Pass PCH parameters to ClangAnalyzer

rita-g-sonarsource:
Part of #1560

### DIFF
--- a/src/Integration.Vsix/CFamily/CFamilyHelper.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyHelper.cs
@@ -82,6 +82,8 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
 
             if (analyzerOptions is CFamilyAnalyzerOptions cFamilyAnalyzerOptions)
             {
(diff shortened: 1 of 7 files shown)


src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -80,9 +80,18 @@ private static Request CreateRequest(FileConfig fileConfig, string absoluteFileP
...
-                request.Flags |= Request.CreateReproducer;
+                if (cFamilyAnalyzerOptions.CreateReproducer)
+                {
+                    request.Flags |= Request.CreateReproducer;
+                }
+
+                if (cFamilyAnalyzerOptions.CreatePreCompiledHeaders)
+                {
+                    request.Flags |= Request.BuildPreamble;
+                    request.PchFile = cFamilyAnalyzerOptions.PreCompiledHeadersFilePath;
```
duncanp-sonar, line 93: It's worth adding at least some assertions to validate the arguments i.e. that the `CreateReproducer` and `CreatePreCompilerHeaders` are not both passed, and that the path isn't null.

PR #1573, opened 2020-08-05
Upgrade moq to 4.14.5

rita-g-sonarsource:
Upgraded package to take advantage of [event verification](https://github.com/moq/moq4/pull/857)


src/Integration.UnitTests/Persistence/SolutionBindingCredentialsLoaderTests.cs
```diff
@@ -87,15 +87,15 @@ public void Save_ServerUriIsNull_CredentialsNotSaved()
 
             testSubject.Save(credentials, null);
```
rita-g-sonarsource, line 89: We had a bug in the tests, wrong assertion so they always passed. This was caught by the new Moq version.

PR #1575, opened 2020-08-05
Extract text document retrival logic from ActiveDocumentLocator

rita-g-sonarsource:
Part of #1559

### DIFF
--- a/src/Integration.Vsix/CFamily/CFamilyReproducerCommand.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyReproducerCommand.cs
@@ -56,10 +56,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
             // the UI thread.
             await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(package.DisposalToken);
 
-            var monitorSelection = await package.GetServiceAsync(typeof(SVsShellMonitorSelection)) as IVsMonitorSelection;
-            var textDocumentProvider = await package.GetMefServiceAsync<ITextDocumentProvider>();
-            var docLocator = new ActiveDocumentLocator(monitorSelection, textDocumentProvider);
-
+            var docLocator = await package.GetMefServiceAsync<IActiveDocumentLocator>();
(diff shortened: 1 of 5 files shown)


rita-g-sonarsource:
@duncanp-sonar this is the exact logic that I need to extract the text document from the active frame. 
I made no product changes, just extracted as-is to a new class that I could re-use.

src/Integration.Vsix/CFamily/CFamilyReproducerCommand.cs
```diff
@@ -58,8 +57,8 @@ public static async Task InitializeAsync(AsyncPackage package, ILogger logger)
             await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(package.DisposalToken);
 
             var monitorSelection = await package.GetServiceAsync(typeof(SVsShellMonitorSelection)) as IVsMonitorSelection;
-            var adapterService = await package.GetMefServiceAsync<IVsEditorAdaptersFactoryService>();
-            var docLocator = new ActiveDocumentLocator(monitorSelection, adapterService);
+            var textDocumentProvider = await package.GetMefServiceAsync<ITextDocumentProvider>();
+            var docLocator = new ActiveDocumentLocator(monitorSelection, textDocumentProvider);
```
duncanp-sonar, line 61: This way of construction doesn't seem quite right. If we are going to make one of our classes MEF-constructable, it should be `ActiveDocumentLocator`, since `ITextDocumentProvider` is an implementation detail as far the `CFamilyReproducerCommand` is concerned.

It might make more sense when `ITextDocumentProvider` is being used elsewhere but it looks odd here.

rita-g-sonarsource: Take a look at [WIP ](https://github.com/SonarSource/sonarlint-visualstudio/blob/4749d102e3ff5f20c10b61aeabb414503caeac2a/src/Integration.Vsix/Helpers/DocumentEvents/DocumentFocusedEventRaiser.cs#L66) of how it will be used

PR #1577, opened 2020-08-05
Extract RunOnUIThread

rita-g-sonarsource:
Part of #1559


rita-g-sonarsource:
@duncanp-sonar following our conversation about threading, I took a look how [roslyn do it](https://github.com/dotnet/roslyn/blob/83a6d3ae12a54b3c9ddd830f1318bc9912eb06b2/src/VisualStudio/Core/Def/Implementation/Workspace/VisualStudioActiveDocumentTracker.cs#L55) and it seems that I would need to wrap my ctor with switching to UI thread. We already had this code so I extracted it to class, no product changes.
The usage will be like [this](https://github.com/SonarSource/sonarlint-visualstudio/blob/4749d102e3ff5f20c10b61aeabb414503caeac2a/src/Integration.Vsix/Helpers/DocumentEvents/DocumentFocusedEventRaiser.cs#L45).

PR #1578, opened 2020-08-05
Raise document focused event

### LINKED ISSUE #1559: Implement file selection tracker that detects when a code file receives focus

rita-g-sonarsource:
Fixes #1559


src/Integration.Vsix/Helpers/DocumentEvents/ActiveDocumentTracker.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            this.textDocumentProvider = textDocumentProvider;
+
+            RunOnUIThread.Run(() =>
+            {
+                monitorSelection = serviceProvider.GetService<SVsShellMonitorSelection, IVsMonitorSelection>();
+
+                if (monitorSelection == null)
+                {
+                    throw new ArgumentNullException(nameof(monitorSelection));
```
duncanp-sonar, line 56: This looks a bit odd; unhandled exceptions on the UI thread normally crash VS. In practice, it won't make any difference; if that service is missing then there's something seriously wrong with VS anyway.

src/Integration.Vsix/Helpers/DocumentEvents/ActiveDocumentTracker.cs
```diff
@@ -0,0 +1,110 @@
...
+                }
+
+                monitorSelection.AdviseSelectionEvents(this, out cookie);
+            });
+        }
+
+        public int OnElementValueChanged(uint elementId, object oldValue, object newValue)
```
duncanp-sonar, line 63: Looks odd having these methods as public; may implement the interface explicitly?

src/Integration.Vsix/Helpers/DocumentEvents/ActiveDocumentTracker.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            if (newValue is IVsWindowFrame frame &&
+                IsFrameElement() && 
+                IsDocumentFrame())
+            {
+                var textDocument = textDocumentProvider.GetFromFrame(frame);
```
duncanp-sonar, line 71: `GetFromFrame` can return null, can't it?
Whose responsibility is it to handle this - this class (in which case it shouldn't raise the event in that case), or the event listeners (who would all then have to handle the document being null)?

If this class is making a guarantee that it will never raise with a null document then it's worth documenting in a doc comment on the event declaration.

src/Integration.Vsix/Helpers/DocumentEvents/ActiveDocumentTracker.cs
rita-g-sonarsource: I'll remove the exception

src/Integration.Vsix/Helpers/DocumentEvents/ActiveDocumentTracker.cs
rita-g-sonarsource: You're right; I'll add a null check + document that it shouldn't raise an event with a null document.

src/Integration.Vsix.UnitTests/Helpers/ActiveDocumentTrackerTests.cs
```diff
@@ -0,0 +1,209 @@
...
+
+            testSubject = new ActiveDocumentTracker(serviceProviderMock.Object, textDocumentProviderMock.Object);
+            testSubject.OnDocumentFocused += (sender, e) => mockEventHandler.Object(e);
+        }
+
+        [TestMethod]
+        public void Ctor_ShouldRegisterToSelectionEvents()
+        {
+            uint cookie = 1234;
+            monitorSelectionMock.Reset();
```
duncanp-sonar, line 69: Is this `Reset` required? Nothing has happened since initialization.

rita-g-sonarsource: Sort of: I init monitorSelectionMock in the TestInitialize, which is then passed to another instance of ActiveDocumentTracker, which calls AdviseSelectionEvents on it.
So then when I call `monitorSelectionMock.VerifyNoOtherCalls` - it throws that there was an additional invocation.

src/Integration.Vsix/Helpers/DocumentEvents/ActiveDocumentTracker.cs
duncanp-sonar: We'll just crash a few lines later in line 59 unless you code defensively there too. It does seem odd that the Roslyn code you linked to throws explicitly in this case.

src/Integration.Vsix.UnitTests/Helpers/ActiveDocumentTrackerTests.cs
duncanp-sonar: Ah. That illustrates what's wrong with [TestInitialize] unless it applies to every test, which it almost never does.

It's not obvious why the reset is there, so at least it deserves a comment.

PR #1579, opened 2020-08-06
Generate PCH on document focus

### LINKED ISSUE #1560: Trigger pre-analysis build when a code file is activated

rita-g-sonarsource:
Fixes #1560

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyPreCompiledHeadersEventListenerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyPreCompiledHeadersEventListenerTests.cs
@@ -1,0 +1,175 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Utilities;
+using Moq;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Integration.Vsix;
+using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+using SonarLint.VisualStudio.Integration.Vsix.Helpers.DocumentEvents;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
+{
+    [TestClass]
+    public class CFamilyPreCompiledHeadersEventListenerTests
+    {
+        private const string FocusedDocumentFilePath = "c:\\myfile.cpp";
+        private IContentType contentTypeMock;
+
+        private Mock<ICLangAnalyzer> clangAnalyzerMock;
+        private Mock<IActiveDocumentTracker> activeDocumentTrackerMock;
+        private Mock<IScheduler> schedulerMock;
+        private Mock<ISonarLanguageRecognizer> languageRecognizerMock;
+
+        private CFamilyPreCompiledHeadersEventListener testSubject;
+
+        [TestInitialize]
+        public void TestInitialize()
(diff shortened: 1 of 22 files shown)


src/Integration.Vsix/CFamily/CFamilyPreCompiledHeadersEventListener.cs
```diff
@@ -0,0 +1,103 @@
...
+using System;
+using System.ComponentModel.Composition;
+using System.IO;
+using System.Linq;
+using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Integration.Vsix.Helpers.DocumentEvents;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
+{
+    interface ICFamilyPreCompiledHeadersEventListener : IDisposable
```
duncanp-sonar, line 30: I'm not sure we need `CFamily` in the name since it's in the `CFamily` namespace.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -34,9 +34,14 @@
 
 namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
 {
+    internal interface ICLangAnalyzer : IAnalyzer
```
duncanp-sonar, line 37: To avoid confusion I think we shouldn't use `CLang` unless we are actually referring to something that is CLang-specific (and at some point we should rename `CLangAnalyzer`).
It might be better to call this `ICFamilyAnalyzer` (even though it means repeating the `CFamily` from the namespace).

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -136,7 +141,7 @@ private void HandleMessage(Message message, Request request, IIssueConsumer cons
             // returned. This doesn't cause a crash; all active taggers will have been detached from the
             // TextBufferIssueTracker when the file was closed, but the TextBufferIssueTracker will
             // still exist and handle the call.
-            consumer.Accept(request.File, new[] { issue });
+            consumer?.Accept(request.File, new[] { issue });
```
duncanp-sonar, line 144: If the consumer is null then there's no point in doing any of the work in this method (or even in calling the method).

src/Integration.Vsix/CFamily/CFamilyPreCompiledHeadersEventListener.cs
```diff
@@ -0,0 +1,103 @@
...
+            IScheduler scheduler,
+            ISonarLanguageRecognizer sonarLanguageRecognizer)
+        {
+            this.cLangAnalyzer = cLangAnalyzer;
+            this.activeDocumentTracker = activeDocumentTracker;
+            this.scheduler = scheduler;
+            this.sonarLanguageRecognizer = sonarLanguageRecognizer;
+        }
+
+        public void Listen()
```
duncanp-sonar, line 62: Is there an advantage in having 'Listen' as a separate method rather than just starting listening in the constructor?

src/Integration.Vsix.UnitTests/CFamily/CFamilyPreCompiledHeadersEventListenerTests.cs
```diff
@@ -0,0 +1,176 @@
...
+using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+using SonarLint.VisualStudio.Integration.Vsix.Helpers.DocumentEvents;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
+{
+    [TestClass]
+    public class CFamilyPreCompiledHeadersEventListenerTests
+    {
+        private const string FocusedDocumentFilePath = "c:\\myfile.cpp";
+        private IContentType contentTypeMock;
```
duncanp-sonar, line 41: Could be `readonly`. It could also be named to make it's relationship to `FocusedDocumentFilePath` clearer e.g. `FocusedDocumentContentType`.

src/Integration.Vsix.UnitTests/CFamily/CFamilyPreCompiledHeadersEventListenerTests.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            clangAnalyzerMock.VerifyNoOtherCalls();
+            schedulerMock.VerifyNoOtherCalls();
+            languageRecognizerMock.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public void OnDocumentFocused_NoLanguagesDetected_PchGenerationNotScheduled()
+        {
+            languageRecognizerMock
```
duncanp-sonar, line 98: This setup step is repeated in three places and only the return value varies - a helper method would be make the tests easier to read.

src/Integration.Vsix/CFamily/CFamilyPreCompiledHeadersEventListener.cs
rita-g-sonarsource: I think it's just odd to have an interface without any functions in it.

duncanp-sonar: Yep, but maybe that says something about the design or usage. What happens if `Listen` is called multiple times? Or `Listen` is called after `Dispose`?
It's not that uncommon to have a marker interface that doesn't do anything.

PR #1580, opened 2020-08-06
Unify output window and status bar analysis messages

rita-g-sonarsource:
Part of #1560

### DIFF
--- a/src/Core/Analysis/IAnalysisStatusNotifier.cs
+++ b/src/Core/Analysis/IAnalysisStatusNotifier.cs
@@ -18,13 +18,15 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
+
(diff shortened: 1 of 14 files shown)


rita-g-sonarsource:
@duncanp-sonar as you suggested before, I changed AnalysisStatusNotifier to also write messages to the output window. This will allow me to override the status notifications in the PCH branch (otherwise they would still be written to the output window). 
The major change is in `CFamilyHelper.cs` -- it no longer catches the exception from the subprocess. Instead, it propogates it, and lets the caller (`CLangAnalyzer.cs`) to catch it and deal with all the analysis status notifications.
So all the notifications (started/finished/cancelled/failed) are being called from `CLangAnalyzer.cs`.

It might be helpful to take a look at the commits separately, with the [second commit](https://github.com/SonarSource/sonarlint-visualstudio/pull/1580/commits/e3d67f5586e7f16b39ffe92fd8a5d2173c4a1db9) being the actual change, and the others are prep/cleanup.

PR #1583, opened 2020-08-06
Delete PCH files on VS shutdown

### LINKED ISSUE #1562: Implement PCH cache clean up strategy
### Description

The PCH files can be many MB in size. We don't to fill up the user's disc so we need to make sure that (a) the PCH cache doesn't grow in an unlimited fashion and (b) unnecessary files are removed.

Summary:
* cache size of 1.
* single preamble file per-user/machine combination, stored in %temp%
* preamble file will be deleted when VS is closed

### Background - questions about scenarios/implementation trade-offs
* what should the cache size be - 1 to start with?
* single machine-wide cache, or a separate cache per executing VS instance? (note: a machine-wide cache means the cache file format must be the same for all installed versions of SLVS)
* where to save the file(s)?
* when to clean up - never (rely on %temp% cleanup), when the solution is closed, when VS shuts down?

#### Cache size
? what should the cache size be - 1 to start with?

Feedback from the CFamily team is that they noticed a qualitative performance improvement with a cache size of 1. This causes the PCH header file to be recreated every time focus changes to a different document, or when focus changes from a non-document window to a document window. However, the CFamily also reported that there was no observable performance impact in the UI and no observable CPU spike.
The CFamily team confirmed that the `subprocess.exe` is optimised so that the PCH cache file will not be recreated if the existing cache file can be used.

Decision: the cache size will be one for the initial implementation. Implementing a more complicated cache significantly increases the implementation complexity. We assess whether to add this later based on dogfood and user feedback.

#### Handling multiple VS instances
? single machine-wide cache, or a separate cache per executing VS instance? 

It's possible that a developer could be running multiple instances of VS simultaneously e.g. either switching between two separate products (and editing in both instances), or working on a product in one VS instance while looking at a code in another (i.e. editing in one instance, browsing in the other). However, these usage patterns could also occur within a single instance of VS with multiple document windows. The user will only have one keyboard, so they can only be working actively on one document at a time, irrespective of whether they are using one instance of VS or two.

Sharing a single-file cache between multiple instances of VS potentially causes a problem if the format of the cache file changes between versions of SLVS. However, it's unlikely that this would cause a problem in practice: the `subprocess.exe` will handle corrupt/apparently corrupt preamble files correctly. The only circumstance in which this would cause a problem would be if the user was working on the same codebase in two instances of VS that have different versions of SLVS installed. In that case, every switch between VS instances would cause a cache miss because the preamble file format has changed (rather than causing a cache miss because the user is working on a different set of files).

Decision: do not make the cache file specific to the instance of using it.

#### File location
? where to save the file(s)?
Decision: in a location under %temp% to benefit from automated Windows cleanup.

#### When to clean up
 when to clean up - never (rely on %temp% cleanup), when the solution is closed, when VS shuts down?
Decision: on VS shutdown.
There's no obvious advantage in cleaning up when the solution is closed when compared to waiting until VS is closed.

Relates to #1557

rita-g-sonarsource:
Fixes #1562

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
@@ -1,0 +1,91 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO.Abstractions.TestingHelpers;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily.PreCompiledHeaders
+{
+    [TestClass]
+    public class PchFilesDeleterTests
+    {
+        private MockFileSystem fileSystemMock;
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            fileSystemMock = new MockFileSystem();
+            fileSystemMock.Directory.CreateDirectory("c:\\test\\pch");
+        }
+
+        [TestMethod]
+        public void DeletePchFiles_NoFilesInDirectory_NoException()
+        {
+            var testSubject = new PchFilesDeleter(fileSystemMock, "c:\\test\\pch\\myPch.abc");
+
+            Action act = () => testSubject.DeletePchFiles();
+            act.Should().NotThrow();
+        }
+
+        [TestMethod]
+        public void DeletePchFiles_NoMatchingFilesInDirectory_NonMatchingFilesAreNotDeleted()
+        {
+            var nonMatchingFilePaths = new List<string>
+            {
+                "c:\\test\\pch\\test.abc",
+                "c:\\test\\pch\\myPch.ab",
+                "c:\\test\\pch\\myPch.abcd",
+                "c:\\test\\pch\\amyPch.abc",
+                "c:\\test\\pch\\sub\\myPch.abc",
+                "c:\\test\\myPch.abc"
+            };
+
+            foreach (var filePath in nonMatchingFilePaths)
+            {
+                fileSystemMock.AddFile(filePath, new MockFileData(""));
+            }
+
+            var testSubject = new PchFilesDeleter(fileSystemMock, "c:\\test\\pch\\myPch.abc");
+            testSubject.DeletePchFiles();
+
+            fileSystemMock.AllFiles.Should().BeEquivalentTo(nonMatchingFilePaths);
+        }
+
+        [TestMethod]
+        public void DeletePchFiles_HasMatchingFilesInDirectory_MatchingFilesAreDeleted()
+        {
+            var nonMatchingFilePaths = new List<string>
+            {
+                "c:\\test\\pch\\myPch.abc",
+                "c:\\test\\pch\\myPch.abc.a",
+                "c:\\test\\pch\\myPch.abc.a.b"
+            };
+
+            foreach (var filePath in nonMatchingFilePaths)
+            {
+                fileSystemMock.AddFile(filePath, new MockFileData(""));
+            }
+
+            var testSubject = new PchFilesDeleter(fileSystemMock, "c:\\test\\pch\\myPch.abc");
+            testSubject.DeletePchFiles();
+
+            fileSystemMock.AllFiles.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void DeletePchFiles_HasMatchingAndNonMatchingFilesInDirectory_OnlyMatchingFilesAreDeleted()
+        {
+            var matchingFile = "c:\\test\\pch\\myPch.abc.d";
+            var nonMatchingFile = "c:\\test\\pch\\myPch.abcd";
+
+            fileSystemMock.AddFile(matchingFile, new MockFileData(""));
+            fileSystemMock.AddFile(nonMatchingFile, new MockFileData(""));
+
+            var testSubject = new PchFilesDeleter(fileSystemMock, "c:\\test\\pch\\myPch.abc");
+            testSubject.DeletePchFiles();
+
+            fileSystemMock.AllFiles.Should().BeEquivalentTo(new List<string>{nonMatchingFile});
+        }
+    }
+}
--- a/src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
@@ -20,16 +20,16 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily.PreCompiledHeader
         }
 
         [TestMethod]
-        public void DeletePchFiles_NoFilesInDirectory_NoException()
+        public void Cleanup_NoFilesInDirectory_NoException()
         {
-            var testSubject = new PchFilesDeleter(fileSystemMock, "c:\\test\\pch\\myPch.abc");
+            var testSubject = new PchCacheCleaner(fileSystemMock, "c:\\test\\pch\\myPch.abc");
 
-            Action act = () => testSubject.DeletePchFiles();
+            Action act = () => testSubject.Cleanup();
             act.Should().NotThrow();
         }
 
         [TestMethod]
-        public void DeletePchFiles_NoMatchingFilesInDirectory_NonMatchingFilesAreNotDeleted()
+        public void Cleanup_NoMatchingFilesInDirectory_NonMatchingFilesAreNotDeleted()
         {
             var nonMatchingFilePaths = new List<string>
             {
@@ -46,14 +46,14 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily.PreCompiledHeader
                 fileSystemMock.AddFile(filePath, new MockFileData(""));
             }
 
-            var testSubject = new PchFilesDeleter(fileSystemMock, "c:\\test\\pch\\myPch.abc");
-            testSubject.DeletePchFiles();
+            var testSubject = new PchCacheCleaner(fileSystemMock, "c:\\test\\pch\\myPch.abc");
+            testSubject.Cleanup();
 
             fileSystemMock.AllFiles.Should().BeEquivalentTo(nonMatchingFilePaths);
         }
 
         [TestMethod]
-        public void DeletePchFiles_HasMatchingFilesInDirectory_MatchingFilesAreDeleted()
+        public void Cleanup_HasMatchingFilesInDirectory_MatchingFilesAreDeleted()
         {
             var nonMatchingFilePaths = new List<string>
             {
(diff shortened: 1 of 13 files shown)


src/Integration.Vsix/CFamily/PreCompiledHeaders/PchFilesDeleter.cs
```diff
@@ -0,0 +1,55 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.IO;
+using System.IO.Abstractions;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
+{
+    internal interface IPchFilesDeleter
```
rita-g-sonarsource, line 26: Couldn't think of a better name... Got an idea?

src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    [TestClass]
+    public class PchFilesDeleterTests
+    {
+        private MockFileSystem fileSystemMock;
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            fileSystemMock = new MockFileSystem();
```
rita-g-sonarsource, line 18: I'm using a mock file system since I want to check the searchFilter, and it's supported by the mock.

src/Integration.Vsix/CFamily/PreCompiledHeaders/PreCompiledHeadersEventListener.cs
```diff
@@ -100,6 +104,16 @@ public void Dispose()
             {
                 activeDocumentTracker.OnDocumentFocused -= OnActiveDocumentFocused;
                 activeDocumentTracker?.Dispose();
+
+                try
```
rita-g-sonarsource, line 108: Dispose() does gets called when VS shuts down so it's a good place to delete the files.

rita-g-sonarsource:
@duncanp-sonar please note the base branch: this PR is blocked #1582, so once that is merged, I'll change the base of this PR.

src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
```
duncanp-sonar, line 1: Why didn't the "missing copyright" rule complain?

src/Integration.Vsix/CFamily/PreCompiledHeaders/PchFilesDeleter.cs
duncanp-sonar: `IPchCacheCleaner`?

src/Integration.Vsix/CFamily/PreCompiledHeaders/PchFilesDeleter.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        void DeletePchFiles();
```
duncanp-sonar, line 28: It's better to use interface names that describe logical operations that are independent of the implementation e.g. `Cleanup` rather than `DeletePchFiles`.

src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
rita-g-sonarsource: Hm, that's very odd.
I see that we're still referencing the [correct QP](https://sonarcloud.io/organizations/sonarsource/quality_profiles/show?language=cs&name=C%23+QP+for+SonarLint), and the copyright rule is enabled... who should we talk to about this? SC folks? Or dotnet guild?

src/Integration.Vsix/CFamily/PreCompiledHeaders/PchFilesDeleter.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public void DeletePchFiles()
+        {
+            var pchDirectory = Path.GetDirectoryName(pchFilePath);
+            var pchFileName = Path.GetFileName(pchFilePath);
+            var pchFiles = fileSystem.Directory.GetFiles(pchDirectory, $"{pchFileName}.*");
+
+            foreach (var pchFile in pchFiles)
+            {
+                fileSystem.File.Delete(pchFile);
```
duncanp-sonar, line 50: I would put the error handling in this class round each delete. That way, if one file is locked then it won't stop the others from being deleted.

src/Integration.Vsix/CFamily/PreCompiledHeaders/PchFilesDeleter.cs
```diff
@@ -0,0 +1,55 @@
...
+            }
+            fileSystem.File.Delete(pchFilePath);
```
duncanp-sonar, line 52: I'm missing something - won't the wildcard search always pick up this file?

src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
duncanp-sonar: ... because it's a test file.

rita-g-sonarsource: 🤦

src/Integration.Vsix/CFamily/PreCompiledHeaders/PchFilesDeleter.cs
rita-g-sonarsource: No, the wildcard is for files that start with "X." (with the dot), and this one is for the exact pch file. 
If I were to remove the dot, then we would also delete "XY", instead of "X.Y".

src/Integration.Vsix/CFamily/PreCompiledHeaders/PchFilesDeleter.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            this.fileSystem = fileSystem;
+            this.pchFilePath = pchFilePath;
+        }
+
+        public void DeletePchFiles()
+        {
+            var pchDirectory = Path.GetDirectoryName(pchFilePath);
+            var pchFileName = Path.GetFileName(pchFilePath);
+            var pchFiles = fileSystem.Directory.GetFiles(pchDirectory, $"{pchFileName}.*");
```
duncanp-sonar, line 46: Have the CFamily guaranteed that any files they create will always append a "." to our file name, or that just what they do at the moment?
A safer option would be to match "[ourfilename]*" (or even `[ourfilename without extension].*`. 
Our file naming should be unique enough that it won't accidentally pick up any external files.

src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
```diff
@@ -0,0 +1,91 @@
...
+            testSubject.DeletePchFiles();
+
+            fileSystemMock.AllFiles.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void DeletePchFiles_HasMatchingAndNonMatchingFilesInDirectory_OnlyMatchingFilesAreDeleted()
+        {
+            var matchingFile = "c:\\test\\pch\\myPch.abc.d";
+            var nonMatchingFile = "c:\\test\\pch\\myPch.abcd";
```
duncanp-sonar, line 80: See my comment below - I'd expect our cleanup to delete "myPch.abcd" because it starts with the name of the file we know about.

src/Integration.Vsix/CFamily/PreCompiledHeaders/PchFilesDeleter.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public PchFilesDeleter(IFileSystem fileSystem, string pchFilePath)
+        {
+            this.fileSystem = fileSystem;
+            this.pchFilePath = pchFilePath;
+        }
+
+        public void DeletePchFiles()
+        {
+            var pchDirectory = Path.GetDirectoryName(pchFilePath);
```
duncanp-sonar, line 44: I think this deserves a comment to document the agreement with the CFamily team i.e. we give them a file name to use, and they might create one or more files with the same root.

src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        [TestMethod]
+        public void DeletePchFiles_HasMatchingFilesInDirectory_MatchingFilesAreDeleted()
+        {
+            var nonMatchingFilePaths = new List<string>
+            {
+                "c:\\test\\pch\\myPch.abc",
+                "c:\\test\\pch\\myPch.abc.a",
+                "c:\\test\\pch\\myPch.abc.a.b"
```
duncanp-sonar, line 62: Do any of the tests cover case-sensitivity?

src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -35,9 +35,8 @@ internal static partial class CFamilyHelper
     {
         public const string CPP_LANGUAGE_KEY = "cpp";
         public const string C_LANGUAGE_KEY = "c";
-        internal const string PchFilePathSuffix = "SonarLintPCH.preamble";
 
-        internal static readonly string PchFilePath = Path.Combine(Path.GetTempPath(), PchFilePathSuffix);
+        internal static readonly string PchFilePath = Path.Combine(Path.GetTempPath(), "SonarLintPCH.preamble");
```
duncanp-sonar, line 39: I think the name should make it clear that it's SonarLint for VS (SL for VS Code and the other SL flavours also create files, and it would help if we could tell them apart easily).

Maybe:
`SonarLintForVS.PCH.preamble`
`SonarLintForVisualStudio.PCH.preamble`
... just as long as it's clearly VS not VSCode / IntelliJ / Eclipse

src/Integration.Vsix/CFamily/PreCompiledHeaders/PreCompiledHeadersEventListener.cs
```diff
@@ -105,6 +104,16 @@ public void Dispose()
...
+                {
+                    pchFilesDeleter.DeletePchFiles();
+                }
+                catch
```
duncanp-sonar, line 112: It's a good practice to follow the convention of only catching non-critical exceptions.

src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PchFilesDeleterTests.cs
rita-g-sonarsource: Good point, I'll add

src/Integration.Vsix/CFamily/PreCompiledHeaders/PchFilesDeleter.cs
rita-g-sonarsource: I changed to `[ourfilename]*` as that's our agreement with them so far and it's simpler

PR #1585, opened 2020-08-07
Fix parameter type assertion

rita-g-sonarsource:


src/Integration.UnitTests/Persistence/SolutionBindingCredentialsLoaderTests.cs
```diff
@@ -87,15 +87,15 @@ public void Save_ServerUriIsNull_CredentialsNotSaved()
 
             testSubject.Save(credentials, null);
 
-            store.Verify(x=> x.WriteCredentials(It.IsAny<Uri>(), It.IsAny<Credential>()), Times.Never);
+            store.Verify(x=> x.WriteCredentials(It.IsAny<TargetUri>(), It.IsAny<Credential>()), Times.Never);
```
rita-g-sonarsource, line 90: A real error that was caught when upgrading to the latest Moq version: `WriteCredentials` receives `TargetUri`, not `Uri`, so the assertion always succeeds.

rita-g-sonarsource:
@duncanp-sonar found this when I was trying to upgrade the moq version.

PR #1587, opened 2020-08-07
Add PreAnalysis option that doesn't specify RulesConfiguration

rita-g-sonarsource:


src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -37,11 +37,9 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
 {
     internal interface ICFamilyAnalyzer : IAnalyzer
     {
```
rita-g-sonarsource, line 39: @duncanp-sonar WDYT? Instead of passing more and more null arguments, I thought about just creating a PreAnalysis option, that would run without rulesConfig and won't call an issue consumer.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -73,66 +71,85 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
-        public void ExecuteAnalysis(string path, string charset, IEnumerable<AnalysisLanguage> detectedLanguages,
-            IIssueConsumer consumer, IAnalyzerOptions analyzerOptions,
-            CancellationToken cancellationToken)
+        public void ExecutePreAnalysis(string path, IEnumerable<AnalysisLanguage> detectedLanguages, CFamilyAnalyzerOptions analyzerOptions, CancellationToken cancellationToken)
         {
-            ExecuteAnalysis(path, detectedLanguages, consumer, analyzerOptions, analysisStatusNotifier, cancellationToken);
-        }
+            var request = CreateRequest(path, detectedLanguages, analyzerOptions, null);
 
-        public void ExecuteAnalysis(string path,
```
rita-g-sonarsource, line 83: Deleted the previous overload that was used to pass a null notifier

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -73,66 +71,85 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
-            var request = CreateRequest(logger, projectItem, path, cFamilyRulesConfigProvider, analyzerOptions);
             if (request == null)
             {
                 return;
             }
+            
+            TriggerAnalysis(request, consumer, cancellationToken);
+        }
+
+        private Request CreateRequest(string path, IEnumerable<AnalysisLanguage> detectedLanguages, IAnalyzerOptions analyzerOptions, ICFamilyRulesConfigProvider rulesConfigProvider)
```
rita-g-sonarsource, line 98: Helper method for repeating code. It accepts rulesConfigProvider that can be null now when called from PreAnalysis

PR #1588, opened 2020-08-07
Don't pass rules config when creating pch files

### LINKED ISSUE #1586: Don't pass rules configuration when updating the PCH cache
### Description

The rules configuration is not required when creating a PCH preamble file so we shouldn't calculate or pass the rules configuration in that case. All of the configuration is still required.

Note: not calculating the rules configuration will give a minor perf gain, but will also significantly reduce the number of irrelevant messages logged to the output window.

Relates to #1557

duncanp-sonar:
Fixes #1586

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
@@ -37,8 +37,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
     [TestClass]
     public class CFamilyHelperTests
     {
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -37,8 +37,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
     [TestClass]
     public class CFamilyHelperTests
     {
-        private const string FileName = @"C:\absolute\path\to\file.cpp";
```
duncanp-sonar, line 40: Miscellaneous: unused

Issue #1592, opened 2020-08-10
Create assembly for secondary locations code


duncanp-sonar:
### Description

Relates to https://jira.sonarsource.com/browse/MMF-2098

The secondary locations feature is quite large, but also quite self-contained. Rather than adding more code to the existing `Integration` and `Integration.VSIX` projects, we can create a new assembly for the new feature. This will help make the code more manageable, and should also makes it easier to check that the code in the new assembly is well-structured and does not introduce any new VS threading issues.

The risk is that something in the new assembly turns out to require a reference to a type in the `Integration.VSIX` project (in fact this is certain to happen since the new code will need to reference our existing table data source implementation).
There are various ways this could be handled: abandoning the new assembly, moving code from `Integration.VSIX` to the new assembly, introducing a new interface in the new assembly or `Core` that is implemented by the appropriate type in `Integration.VSIX`.

Proposed name: `SonarLint.VisualStudio.IssueVisualization`

TODO
- [x] add a new product assembly importing appropriate settings/props files
- [x] reference the new assembly from the VSIX assembly and update the manifest
- [x] add a new test assembly that can see the product assembly internals
- [x] test that the new assembly is included in the VSIX.

PR #1608, opened 2020-08-11
Added IssueVisualization project

duncanp-sonar:
Relates to #1592


src/Integration.Vsix/Manifests/VS2017/source.extension.vsixmanifest
```diff
@@ -33,6 +33,7 @@
         <Asset Type="Microsoft.VisualStudio.Analyzer" d:Source="File" Path="Google.Protobuf.dll" />
         <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="File" Path="Google.Protobuf.dll" />
         <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="Integration.TeamExplorer" Path="|Integration.TeamExplorer|" />
+        <Asset Type="Microsoft.VisualStudio.MefComponent" Path="SonarLint.VisualStudio.IssueVisualization.dll" />
```
rita-g-sonarsource, line 36: Out of curiosity, why is it different from the way we declare other project assets? (e.g. Integration.TeamExplorer)

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -133,6 +133,9 @@
       <IncludeOutputGroupsInVSIX>BuiltProjectOutputGroup%3bBuiltProjectOutputGroupDependencies%3bGetCopyToOutputDirectoryItems%3bSatelliteDllsProjectOutputGroup%3b</IncludeOutputGroupsInVSIX>
       <IncludeOutputGroupsInVSIXLocalOnly>DebugSymbolsProjectOutputGroup%3b</IncludeOutputGroupsInVSIXLocalOnly>
     </ProjectReference>
+    <ProjectReference Include="..\IssueViz\IssueViz.csproj">
```
rita-g-sonarsource, line 136: What about IncludeOutputGroupsInVSIX and IncludeOutputGroupsInVSIXLocalOnly?

src/IssueViz/InternalsVisibleTo.cs
```diff
@@ -0,0 +1,28 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Runtime.CompilerServices;
+
+#if SignAssembly
```
rita-g-sonarsource, line 23: Wouldn't we need Integration.Vsix to register the commands?
Or, better yet: who will be responsible for registering the commands and tool windows?

src/Integration.Vsix/Integration.Vsix.csproj
duncanp-sonar: I'm not sure what added those properties or whether they are required.
The reference to `Core` doesn't include them, so I decided to start simple. It may be that we'll need to add them later as we add dependencies to IssueViz to make sure those dependencies end up in the vsix.

src/Integration.Vsix/Manifests/VS2017/source.extension.vsixmanifest
duncanp-sonar: That's weird: I thought I'd cut and paste the setting from the VS2019 manifest. I'll see what happens if I change it.

src/IssueViz/InternalsVisibleTo.cs
duncanp-sonar: I'd expect we'll add a new package to this assembly and update the manifest at that point.

src/Integration.Vsix/Manifests/VS2017/source.extension.vsixmanifest
duncanp-sonar: I cut and paste the entry from the built vsix, which has translated the "Project" etc settings into a "Path" to a file in the vsix. I'll change the other manifests to be the same as the VS2019 one.

PR #1610, opened 2020-08-11
Add secondary locations property

### LINKED ISSUE #1593: Enrich IAnalysisIssue to describes flows and multiple locations
### Description

The issue data classes need to be richer to support the flow and secondary location data returned by SonarQube/SonarCloud.

Note: there is a separate ticket (#1607) regarding how different classes of issues are rendered in the UI that might have an impact on this ticket.

rita-g-sonarsource:
Fixes #1593

### DIFF
--- a/src/Core/Analysis/AnalysisIssue.cs
+++ b/src/Core/Analysis/AnalysisIssue.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
@@ -25,6 +26,8 @@ namespace SonarLint.VisualStudio.Core.Analysis
 {
     public class AnalysisIssue : IAnalysisIssue
     {
+        private static readonly IReadOnlyList<IAnalysisIssueLocation> EmptyLocations = Array.Empty<IAnalysisIssueLocation>();
+
         public AnalysisIssue(
             string ruleKey, AnalysisIssueSeverity severity, AnalysisIssueType type,
             string message, string filePath,
@@ -42,7 +45,10 @@ namespace SonarLint.VisualStudio.Core.Analysis
             EndLineOffset = endLineOffset;
             FilePath = filePath;
             Message = message;
-            Locations = locations.Reverse().ToList().AsReadOnly();
+
+            Locations = locations == null || !locations.Any()
+                ? EmptyLocations
+                : locations.ToList().AsReadOnly();
         }
 
         public string RuleKey { get; }
--- a/src/Core/Analysis/AnalysisIssue.cs
+++ b/src/Core/Analysis/AnalysisIssue.cs
@@ -20,20 +20,19 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 
 namespace SonarLint.VisualStudio.Core.Analysis
 {
     public class AnalysisIssue : IAnalysisIssue
     {
-        private static readonly IReadOnlyList<IAnalysisIssueLocation> EmptyLocations = Array.Empty<IAnalysisIssueLocation>();
+        private static readonly IReadOnlyList<IAnalysisIssueFlow> EmptyFlows = Array.Empty<IAnalysisIssueFlow>();
 
         public AnalysisIssue(
             string ruleKey, AnalysisIssueSeverity severity, AnalysisIssueType type,
             string message, string filePath,
             int startLine, int endLine,
             int startLineOffset, int endLineOffset,
-            IEnumerable<IAnalysisIssueLocation> locations
+            IReadOnlyList<IAnalysisIssueFlow> flows
             )
         {
             RuleKey = ruleKey;
@@ -45,10 +44,7 @@ namespace SonarLint.VisualStudio.Core.Analysis
             EndLineOffset = endLineOffset;
             FilePath = filePath;
             Message = message;
-
-            Locations = locations == null || !locations.Any()
-                ? EmptyLocations
-                : locations.ToList().AsReadOnly();
+            Flows = flows ?? EmptyFlows;
         }
 
         public string RuleKey { get; }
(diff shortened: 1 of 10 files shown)


src/Core/Analysis/IAnalysisIssue.cs
```diff
@@ -39,6 +41,23 @@ public interface IAnalysisIssue
         string Message { get; }
 
         string FilePath { get; }
+
+        IReadOnlyList<IAnalysisIssueLocation> Locations { get; }
+    }
+
+    public interface IAnalysisIssueLocation
```
rita-g-sonarsource, line 48: I'm debating if IAnalysisIssue should inherit from IAnalysisIssueLocation, as the "main" location is a location as well. WDYT?

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -359,25 +384,18 @@ public void ToSonarLintIssue_EndLineIsZero()
...
 
         [TestMethod]
-        public void ToSonarLintIssue_SeverityAndTypeLookup()
+        [DataRow("rule2", AnalysisIssueSeverity.Info, AnalysisIssueType.CodeSmell)]
+        [DataRow("rule3", AnalysisIssueSeverity.Critical, AnalysisIssueType.Vulnerability)]
+        public void ToSonarLintIssue_SeverityAndTypeLookup(string ruleKey, AnalysisIssueSeverity expectedSeverity, AnalysisIssueType expectedType)
         {
             var ruleConfig = GetDummyRulesConfiguration();
 
-            // 1. Check rule2
```
rita-g-sonarsource, line 366: Changed to data tests

src/Core/Analysis/AnalysisIssue.cs
```diff
@@ -38,6 +42,7 @@ public class AnalysisIssue : IAnalysisIssue
             EndLineOffset = endLineOffset;
             FilePath = filePath;
             Message = message;
+            Locations = locations.Reverse().ToList().AsReadOnly();
```
duncanp-sonar, line 45: There should be a doc comment on the interface describing the ordering of the locations.
It also seems weird that this class is the one that's responsible for reversing them, rather than the converter class that is mapping from whatever is the underlying source of the issues.

src/Integration.Vsix.UnitTests/CFamily/IntegrationTests/CLangAnalyzerTestFile_OneIssue_HasSecondaryLocations_response.json
```diff
@@ -0,0 +1,48 @@
+{
+  "Messages": [
+    {
+      "Filename": "",
+      "RuleKey": "S2259",
+      "PartsMakeFlow": true,
```
rita-g-sonarsource, line 6: FYI - We could use the property `PartsMakeFlow` to display the steps differently in the UI, should we want to.
For example, in cognitive complexity rule the property is False. In null reference it's True.

src/Core/Analysis/AnalysisIssue.cs
rita-g-sonarsource: Yea I was debating about that too. Originally wanted to re-use with the deamon, but the deamon has no support for secondary locations in the protobuf, so it's not relevant. 
I'll move it to the cfamilyHelper.

src/Core/Analysis/IAnalysisIssue.cs
duncanp-sonar: No, I don't think so - doesn't feel right, somehow. "Favour composition over inheritance" - if it needs to have a "primary location", give it a location property.

src/Integration.Vsix.UnitTests/CFamily/IntegrationTests/CLangAnalyzerTestFile_OneIssue_HasSecondaryLocations_response.json
duncanp-sonar: Is this field passed to SonarQube/SonarCloud? (i.e. is it returned by the get issues API?)
If so then I think we should include it in the data model.

src/Core/Analysis/AnalysisIssue.cs
duncanp-sonar, line 45: Most issues don't have secondary locations so most of the time line 45 won't be doing anything except doing lots of object allocations to represent "no secondary locations".

In this case, I'd probably add an optimisation for the "normal" case to use a "empty" singleton if the supplied list of locations is null or empty.

There's a precedent for this in the `SonarQubeRule` class that defines a singleton `Empty` dictionary to avoid creating unnecessary objects.

src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs
```diff
@@ -379,7 +378,8 @@ private async System.Threading.Tasks.Task ProcessIssues(AsyncServerStreamingCall
                 startLine: daemonIssue.StartLine,
                 endLine: daemonIssue.EndLine,
                 startLineOffset: daemonIssue.StartLineOffset,
-                endLineOffset: daemonIssue.EndLineOffset
+                endLineOffset: daemonIssue.EndLineOffset,
+                locations: new List<IAnalysisIssueLocation>()
```
duncanp-sonar, line 382: I think passing `Array.Empty<IAnalysisIssueLocation>` would avoid creating an object each time. Or change the `AnalysisIssue` to support `locations` being `null`.

src/Core/Analysis/IAnalysisIssue.cs
```diff
@@ -39,6 +41,23 @@ public interface IAnalysisIssue
         string Message { get; }
 
         string FilePath { get; }
+
+        IReadOnlyList<IAnalysisIssueLocation> Locations { get; }
```
duncanp-sonar, line 45: This is the key data structure for secondary locations and the security vulnerabilities sprint to follow shortly.
We don't need flows for CFamily issues, but we'll have to rework practically every component we write this sprint when we move on to the security work. I think we could avoid some of that rework by modeling flows in the data model now, and treating the CFamily case as a degenerate case that only ever has zero or one flows.

src/Integration.Vsix.UnitTests/CFamily/IntegrationTests/CLangAnalyzerTestFile_OneIssue_HasSecondaryLocations_response.json
rita-g-sonarsource: No, it's only in CFamily objects. The API has "flows" property instead (that we used in the prototype).

PR #1612, opened 2020-08-12
Automatically update the asmref baseline when building in the IDE

duncanp-sonar:
I'm in two minds about whether auto-updating the baseline during development defeats the point of having the AsmRef checking. On balance I think it's ok since we should notice that the baseline has been updated and decide whether or not to commit the changes.


rita-g-sonarsource:
@duncanp-sonar I agree with the change, a) it hinders development, and b) we should notice the changes when committing or during the PR.

Issue #1599, opened 2020-08-10
Implement the issue navigation panel


duncanp-sonar:
### Description

Implement a tool window that display the selected issue.

The UI should support:
* displaying multiple flows
* display the steps for the current flow, grouped by file.
* theming

TODO:
? always render every step, or give the user a way of expanding the list?
? track the navigability of an issue? dynamically?
? initial size and visibility?
? auto-hide when not needed? (perhaps based on UserContext)

Relates to https://jira.sonarsource.com/browse/MMF-2098

PR #1613, opened 2020-08-12
Add AnalysisIssueSelectionService

### LINKED ISSUE #1594: Implement service to return currently selected issue
### Description

Implement a stateful service to track the currently selected issue, if any.

The service should track the selected issue, selected flow and selected step, and allow the callers to change the current selection. 
The selected issue can be null.
The service should raise events when any of the selected properties changes.

TODO:
? VS service or MEF export?

rita-g-sonarsource:
Fixes #1594


src/IssueViz/AnalysisIssueSelectionEvents.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        private IAnalysisIssue selectedIssue;
+        private IAnalysisIssueFlow selectedFlow;
+        private IAnalysisIssueLocation selectedLocation;
+
+        public event EventHandler<IssueChangedEventArgs> SelectedIssueChanged;
+        public event EventHandler<FlowChangedEventArgs> SelectedFlowChanged;
+        public event EventHandler<LocationChangedEventArgs> SelectedLocationChanged;
+
+        public IAnalysisIssue SelectedIssue
```
rita-g-sonarsource, line 39: I'm debating if this service should also be responsible for fixing the relationship between the selected items, e.g.:
1. When SelectedIssue changes, then the SelectedFlow should change (either to null or to first flow of the issue)
2. When SelectedFlow changes, then the SelectedLocation should change (either to null or to the first location of the flow). 

WDYT?

src/IssueViz/InternalsVisibleTo.cs
```diff
@@ -22,7 +22,12 @@
 
 #if SignAssembly
 [assembly: InternalsVisibleTo("SonarLint.VisualStudio.IssueVisualization.UnitTests,PublicKey=002400000480000094000000060200000024000052534131000400000100010081b4345a022cc0f4b42bdc795a5a7a1623c1e58dc2246645d751ad41ba98f2749dc5c4e0da3a9e09febcb2cd5b088a0f041f8ac24b20e736d8ae523061733782f9c4cd75b44f17a63714aced0b29a59cd1ce58d8e10ccdb6012c7098c39871043b7241ac4ab9f6b34f183db716082cd57c1ff648135bece256357ba735e67dc6")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```
rita-g-sonarsource, line 25: Needed for moq, `new Mock<EventHandler<IssueChangedEventArgs>>();`

src/IssueViz/Models/AnalysisIssueLocationVisualization.cs
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.Core.Analysis;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Models
+{
+    internal interface IAnalysisIssueLocationVisualization
```
rita-g-sonarsource, line 25: In the prototype we also included LineHash and SourceText but it doesn't feel right to have them here. 
IsNavigable is calculated by a service, `ILocationNavigationChecker`, so I think that should be the place to calculate the hash and compare.

src/IssueViz/LocationNavigationChecker.cs
```diff
@@ -0,0 +1,40 @@
...
+        bool IsNavigable(IAnalysisIssueLocation location);
+    }
+
+    [Export(typeof(ILocationNavigationChecker))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class LocationNavigationChecker : ILocationNavigationChecker
+    {
+        public bool IsNavigable(IAnalysisIssueLocation location)
+        {
+            return true;
```
rita-g-sonarsource, line 37: Will be part of #1599

src/IssueViz/Models/AnalysisIssueVisualizationConverter.cs
```diff
@@ -0,0 +1,89 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Linq;
+using SonarLint.VisualStudio.Core.Analysis;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Models
+{
+    internal interface IAnalysisIssueVisualizationConverter
```
rita-g-sonarsource, line 29: This is what we could use to translate between the object in the error list and the consumers (e.g. the panel).

rita-g-sonarsource:

src/IssueViz/Models/AnalysisIssueLocationVisualization.cs
duncanp-sonar: Does that mean the line text data and the hashing calculation would be repeated each time navigability is checked?
How is the navigation checker going to fetch the text (which should be the text in the document at the point the issue was analysed)?

src/IssueViz/Models/AnalysisIssueVisualizationConverter.cs
duncanp-sonar: I suspect the object fetching the data from the table control might need to do more of the work, but we'll see how it goes.

src/IssueViz/SelectionEvents/IAnalysisIssueSelectionEvents.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace SonarLint.VisualStudio.IssueVisualization.SelectionEvents
+{
+    internal interface IAnalysisIssueSelectionEvents : IDisposable
+    {
+        event EventHandler<IssueChangedEventArgs> SelectedIssueChanged;
+        event EventHandler<FlowChangedEventArgs> SelectedFlowChanged;
+        event EventHandler<LocationChangedEventArgs> SelectedLocationChanged;
+
+        IAnalysisIssueVisualization SelectedIssue { get; set; }
```
duncanp-sonar, line 32: Does `Visualization`mean `ViewModel` in this context?

src/IssueViz/Models/AnalysisIssueFlowVisualization.cs
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+using System.Collections.Generic;
+using SonarLint.VisualStudio.Core.Analysis;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Models
+{
+    internal interface IAnalysisIssueFlowVisualization
+    {
+        int FlowNumber { get; }
```
duncanp-sonar, line 28: As discussed offline, this and the `StepNumber` should perhaps be a text `Label` instead.

src/IssueViz/Models/AnalysisIssueLocationVisualization.cs
rita-g-sonarsource: As discussed offline, LineHash might be part of IAnalysisIssue and IAnalysisIssueLocation. In which case, no further changes would be required to the Visualization objects (as they hold a reference to the original data objects).

src/IssueViz/SelectionEvents/IAnalysisIssueSelectionEvents.cs
rita-g-sonarsource: Not really; `Visualization` are decorator objects, similar to `IFilterableIssue`. They will be the objects that get passed to the editor as well, and it seems odd to have the editor reference view models.

duncanp-sonar: The editor UI is tied to WPF so using the term view model doesn't seem odd to me.

PR #1614, opened 2020-08-12
Put each project and test project in a dedicated folder

rita-g-sonarsource:
image


duncanp-sonar:
I'm not sure whether the `src` folder is adding much value, other than adding an extra level to navigate.
The `SonarQube` folder might be better named something like `BindingConfig`.

rita-g-sonarsource:
> I'm not sure whether the `src` folder is adding much value, other than adding an extra level to navigate.
> The `SonarQube` folder might be better named something like `BindingConfig`.

maybe `.sonarlint`?

PR #1615, opened 2020-08-12
Add Issue Visualization Panel

### LINKED ISSUE #1599: Implement the issue navigation panel

rita-g-sonarsource:
Fixes #1599

image

image


src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml.cs
```diff
@@ -18,17 +18,39 @@
...
+
             InitializeComponent();
         }
+
+        public void Dispose()
+        {
+            ViewModel?.Dispose();
+        }
+
+        private void Hyperlink_OnRequestNavigate(object sender, RequestNavigateEventArgs e)
```
rita-g-sonarsource, line 45: Same as [roslyn](https://github.com/dotnet/roslyn/blob/9625a14e7405d7a3ea1dc77957d5293850b69cc8/src/VisualStudio/Core/Impl/Options/CodeStyleNoticeTextBlock.xaml.cs#L23) (also using [VsShellUtilities](https://github.com/dotnet/roslyn/blob/e704ca635bd6de70a0250e34c4567c7a28fa9f6d/src/VisualStudio/Core/Def/Implementation/Utilities/BrowserHelper.cs#L103))

src/IssueViz/IssueVisualizationControl/SeverityToMonikerConverter.cs
```diff
@@ -0,0 +1,68 @@
...
+using Microsoft.VisualStudio.Imaging;
+using Microsoft.VisualStudio.Imaging.Interop;
+using Microsoft.VisualStudio.Shell.Interop;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.IssueVisualization.Helpers;
+
+namespace SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl
+{
+    [ValueConversion(typeof(AnalysisIssueSeverity), typeof(ImageMoniker))]
+    public class SeverityToMonikerConverter : IValueConverter
```
rita-g-sonarsource, line 33: UX stuff: displaying the issue's severity as an icon next to the description

src/IssueViz/IssueVisualizationControl/ViewModels/FileNameLocationListItem.cs
```diff
@@ -0,0 +1,37 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+
+namespace SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl.ViewModels
+{
+    internal sealed class FileNameLocationListItem : ILocationListItem
+    {
+        public string FullPath { get; }
```
rita-g-sonarsource, line 26: UX stuff: using FullPath as a tooltip when you hover over the file node in the list

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -0,0 +1,207 @@
...
+                    listItems.AddRange(sequentialLocations.Select(x => (ILocationListItem)new LocationListItem(x)));
+
+                    i += sequentialLocations.Count - 1;
+                }
+            }
+
+            return listItems;
+        }
+
+        private ImageMoniker GetImageMonikerForFile(string filePath)
```
rita-g-sonarsource, line 178: UX stuff: displaying an icon with file's type in the list

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -0,0 +1,207 @@
...
+        private readonly IAnalysisIssueSelectionService selectionEvents;
+        private readonly IVsImageService2 vsImageService;
+        private readonly IRuleHelpLinkProvider ruleHelpLinkProvider;
+        private readonly ILogger logger;
+
+        private IAnalysisIssueVisualization currentIssue;
+        private IAnalysisIssueFlowVisualization currentFlow;
+        private LocationListItem currentLocation;
+
+        private bool isBindingUpdatedOutsideOfControl;
```
rita-g-sonarsource, line 48: There is a two-way binding between the panel and the selectionService - this is a way to avoid the infinite recursion. 

Value changed in UI --> the view model updates the property in selectionService --> selectionService raises an event of flow/location Changed --> view model listens to it and calls NotifyPropertyChanged

src/IssueViz.UnitTests/IssueVisualizationControl/IssueVisualizationViewModelTests.cs
```diff
@@ -0,0 +1,544 @@
...
+            var locationViz = new Mock<IAnalysisIssueLocationVisualization>();
+            locationViz.Setup(x => x.Location).Returns(location.Object);
+
+            return locationViz.Object;
+        }
+
+        private void VerifyLocationList(IEnumerable<ILocationListItem> expectedLocationsList)
+        {
+            testSubject.LocationListItems.Should().BeEquivalentTo(expectedLocationsList, assertionOptions =>
+                assertionOptions
```
rita-g-sonarsource, line 538: 1. RespectingRuntimeTypes = Since the list is ILocationListItem, this allows to compare each item in the list based on the expected actual implementation. Otherwise, it will attempt to compare only the properties that exist on ILocationListItem.
2. ComparingByMembers<ImageMoniker> = ImageMoniker is a struct, so this is a way to compare by fields and properties of the struct.

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml.cs
```diff
@@ -18,17 +18,39 @@
...
 
 namespace SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl
 {
-    internal sealed partial class IssueVisualizationControl : UserControl
+    internal sealed partial class IssueVisualizationControl : UserControl, IDisposable
     {
-        public object ViewModel { get; }
-        
-        public IssueVisualizationControl()
+        public IssueVisualizationViewModel ViewModel { get; }
```
duncanp-sonar, line 31: Does this need to be public?

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml.cs
```diff
@@ -18,17 +18,39 @@
...
+        public IssueVisualizationControl(IssueVisualizationViewModel viewModel)
         {
+            ViewModel = viewModel;
+
             InitializeComponent();
         }
+
+        public void Dispose()
+        {
+            ViewModel?.Dispose();
```
duncanp-sonar, line 42: It seems odd that the control is disposing something it didn't create. Both the view model and the control are created by the `IssueVisualizationToolWindow`, which already implements `IDisposable`. That would seem like a more natural place to dispose the view model.
In practical terms it doesn't look like it will make much difference since the lifetimes of the window pane and the control appear to be linked.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
duncanp-sonar: That feels pretty hacky, but I guess it works and I don't have any better suggestions.
It would be worth adding that explanation as a comment in the code.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -0,0 +1,207 @@
...
+                    .OfType<LocationListItem>()
+                    .FirstOrDefault(x => x.Location == e.Location);
+
+                CurrentLocationListItem = selectedLocationListItem;
+            }
+
+            isBindingUpdatedOutsideOfControl = false;
+        }
+
+        private IReadOnlyList<ILocationListItem> BuildLocationListItems(IAnalysisIssueFlowVisualization flow)
```
duncanp-sonar, line 155: I think this method deserves a comment explaining how we're flattening and grouping a hierarchy into a flat list of items of different types, as it's not obvious why the code is doing what it's doing.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -0,0 +1,207 @@
...
+        {
+            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+        }
+
+        public void Dispose()
+        {
+            selectionEvents.SelectedIssueChanged -= SelectionEvents_SelectedIssueChanged;
+            selectionEvents.SelectedFlowChanged -= SelectionEventsOnSelectedFlowChanged;
+            selectionEvents.SelectedLocationChanged -= SelectionEvents_SelectedLocationChanged;
+            selectionEvents.Dispose();
```
duncanp-sonar, line 204: The view model doesn't own the service so it shouldn't dispose it.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -0,0 +1,207 @@
...
+            {
+                logger.WriteLine(Resources.ERR_FailedToGetFileImageMoniker, filePath, e);
+
+                return KnownMonikers.Blank;
+            }
+        }
+
+        public event PropertyChangedEventHandler PropertyChanged;
+
+        private void OnPropertyChanged(string propertyName)
```
duncanp-sonar, line 194: It's a common pattern to use `[CallerMemberName]` here to avoid specifying the property name in the default case.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            if (e.Location == null)
+            {
+                CurrentLocationListItem = null;
+            }
+            else
+            {
+                var selectedLocationListItem = LocationListItems?
+                    .OfType<LocationListItem>()
+                    .FirstOrDefault(x => x.Location == e.Location);
```
duncanp-sonar, line 147: This looks questionable: it seems odd that we have to walk the list comparing the items to work out which is the current location.
We might need to revisit this if we want also want to render the current location adornment differently.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -0,0 +1,207 @@
...
+
+        public AnalysisIssueSeverity Severity => currentIssue?.Issue?.Severity ?? AnalysisIssueSeverity.Info;
+
+        public IAnalysisIssueVisualization CurrentIssue
+        {
+            get => currentIssue;
+            private set
+            {
+                currentIssue = value;
+                OnPropertyChanged(null);
```
duncanp-sonar, line 76: This raises a notification that all properties in the view model have changed. Have the current flow and current locations been updated at this point? It doesn't look like they have.
It looks as if changing the selected issue will cause the service to raise multiple events, which in turn will trigger multiple unnecessary refresh passes of the UI.

I think the root of the problem is with the service: it starts raising events before the model is stable.

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -2,8 +2,258 @@
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:shell="clr-namespace:Microsoft.VisualStudio.Shell;assembly=Microsoft.VisualStudio.Shell.14.0"
+             xmlns:viewModels="clr-namespace:SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl.ViewModels"
              xmlns:visualization="clr-namespace:SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl"
+             xmlns:imaging="clr-namespace:Microsoft.VisualStudio.Imaging;assembly=Microsoft.VisualStudio.Imaging"
+             xmlns:catalog="clr-namespace:Microsoft.VisualStudio.Imaging;assembly=Microsoft.VisualStudio.ImageCatalog"
+             xmlns:theming="clr-namespace:Microsoft.VisualStudio.PlatformUI;assembly=Microsoft.VisualStudio.Imaging"
+             xmlns:utilities="clr-namespace:Microsoft.VisualStudio.PlatformUI;assembly=Microsoft.VisualStudio.Utilities"
+             xmlns:wpf="clr-namespace:SonarLint.VisualStudio.Core.WPF;assembly=SonarLint.VisualStudio.Core"
```
duncanp-sonar, line 11: `wpf` is quite a misleading prefix - it sounds like it refers to a framework component, whereas it refers to our code.
More generally, it would make the XAML a bit easier to follow if it were clear which aliases are for VS assemblies and which ones are for our code e.g. by starting the VS aliases with `vs`.

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -2,8 +2,258 @@
...
+                            <TextBlock Grid.Column="1" 
+                                       Style="{StaticResource FileNameLocationListItemTextBlockStyle}" 
+                                       ToolTip="{Binding FullPath}"
+                                       Text="{Binding FileName}" />
+                        </Grid>
+                    </DataTemplate>
+                    <DataTemplate DataType="{x:Type viewModels:LocationListItem}">
+                        <Grid Margin="10,0,0,0">
+                            <Grid.ColumnDefinitions>
+                                <ColumnDefinition Name="IndentIcon" Width="Auto"/>
```
duncanp-sonar, line 232: Visually this icon looks like an expander in a hierarchy, but it isn't.  What's it purpose? To make the relationship to the file more obvious>

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -2,8 +2,258 @@
...
+                                SnapsToDevicePixels="{TemplateBinding SnapsToDevicePixels}" 
+                                VerticalAlignment="{TemplateBinding VerticalContentAlignment}"/>
+                        </Border>
+                    </ControlTemplate>
+                </Setter.Value>
+            </Setter>
+            <Style.Triggers>
+                <DataTrigger x:Name="FileNameListItemStyleTrigger" Binding="{Binding Path=Location}" Value="{x:Null}">
+                    <Setter Property="IsEnabled" Value="False"/>
+                    <Setter Property="IsSelected" Value="False"/>
```
duncanp-sonar, line 122: How does this interact with the `IsSelected` trigger below? Does it means the user can't select those items in the list?

src/IssueViz.UnitTests/IssueVisualizationControl/IssueVisualizationViewModelTests.cs
duncanp-sonar: You could add that info to relevant lines:
```
  .RespectingRuntimeTypes() // check underlying types rather than `ILocationListItem`
  .ComparingByMembers<ImageMoniker>()); // check struct fields and properties
```

If it's sufficiently obscure to call out in a review comment then it's probably worth capturing in the code to help whoever is maintaining the code in the future.

duncanp-sonar:
The screenshots look good - better than the prototype.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
rita-g-sonarsource: Hm, perhaps we can change the service to raise each time a single event with all the changed data, so for example:
1. When issue changes -> the service will raise IssueChanged event that will contain Issue, Flow and Location (instead of raising IssueChanged, FlowChanged and LocationChanged).
2. When flow changes -> the service will raise FlowChanged event that will contain Flow and Location (instead of raising FlowChanged and LocationChanged).

WDYT?

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
rita-g-sonarsource: Yep, it felt a bit blank without it. I tried to put the line like in SonarQube, but it looked like crap.
If you got time, maybe you can install Mads' [KnownMonikers explorer](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.KnownMonikersExplorer) and see if there's anything that looks better?
Maybe something like `OverlayTimelineMarkGray` or `InitialNode` could work better, I don't know.

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
rita-g-sonarsource: Yep, exactly. The "file" list items are non-selectable.

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml.cs
rita-g-sonarsource: Yep, doesn't seem to work otherwise

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
rita-g-sonarsource: Hm, it's not what I want though - I want to pass null to completely render the UI (see the [remarks](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.inotifypropertychanged.propertychanged?redirectedfrom=MSDN&view=netcore-3.1))

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
duncanp-sonar: I think it would be better if the service didn't start raising events until it had finished making state changes. That's confusing to work with - if you handle the `IssueChanged` event, you shouldn't ask the service what the currently selected flow is, since it will belong to a different issue.

The data is hierarchical, so it isn't possible to change the flow without changing the location, or the issue without changing the flow. One option would be to make all of the state changes, then raise a single event that specifies which level of the hierarchy changed - All, FlowAndLocation, LocationOnly.

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
rita-g-sonarsource: OverlayTimelineMarkGray 
image

InitialNode
image

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
rita-g-sonarsource: Maybe dictionary then? <int stepNumber, LocationListItem>

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
rita-g-sonarsource: Good point about the flow/location being wrong when listening to IssueChanged event.
I'll make the changes in a separate PR.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
duncanp-sonar: I'd leave it for now. The problem for me is that it's the control that is working it out, rather than it being a property on the IAnalysisIssueLocationVisualization.

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
duncanp-sonar: Maybe ask Chris. I think it looks better without anything.
For me, having a image that looks like an expander but isn't is too confusing (and irritating if you try clicking on it when nothing happens).

rita-g-sonarsource: FYI, Chris also said without anything, so I removed it

PR #1618, opened 2020-08-13
Add IssueViz VS package

rita-g-sonarsource:


src/IssueViz/IssueVizPackage.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell;
+using Task = System.Threading.Tasks.Task;
+
+namespace SonarLint.VisualStudio.IssueVisualization
+{
+    [ExcludeFromCodeCoverage]
+    [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
+    [Guid(PackageGuidString)]
+    [ProvideAutoLoad(VSConstants.UICONTEXT.ShellInitialized_string, PackageAutoLoadFlags.BackgroundLoad)]
```
duncanp-sonar, line 34: Is the `PackageAutoLoad` attribute just for testing?
If we don't definitely need it, please put a TODO/reminder of some description to prompt us to remove it later.

Issue #1595, opened 2020-08-10
Implement Error List selection monitor


duncanp-sonar:
### Description

Implement a monitor class that tracks the current selection in the Error List and updates the SonarLint current issue service when the selection changes.

Notes:
* the events for the standard Error List can be handled by implementing a `ITableControlEventProcessorProvider`.

* there is a separate ticket for implementing any special translation from the data in the Error List to the UI representation for rendering - see #1607.

Relates to https://jira.sonarsource.com/browse/MMF-2098

PR #1619, opened 2020-08-13
Added table control selection monitor

duncanp-sonar:
Relates to #1595


src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests.cs
```diff
@@ -67,61 +68,61 @@ public void SetUp()
         }
 
         [TestMethod]
-        public void Line()
+        public void GetValue_Line()
```
duncanp-sonar, line 71: Miscellaneous - renamed existing tests.

src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests.cs
```diff
@@ -157,11 +158,17 @@ public void HelpLink()
...
         {
             GetValue(StandardTableKeyNames.ProjectName).Should().Be("MyProject");
         }
 
+        [TestMethod]
+        public void GetValue_Issue()
+        {
+            GetValue(SonarLintTableControlConstants.IssueColumnName).Should().BeSameAs(issue);
+        }
+
```
duncanp-sonar, line 171: New test for the new column.

src/IssueViz/TableControls/IssueTablesSelectionMonitor.cs
```diff
@@ -0,0 +1,65 @@
...
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.IssueVisualization.TableControls
+{
+    /// <summary>
+    /// Handles issue selection changes from table controls
+    /// </summary>
+    internal interface IIssueTablesSelectionMonitor
+    {
+        void SelectionChanged(IAnalysisIssue analysisIssue);
```
duncanp-sonar, line 32: The parameter should probably be something other than `IAnalysisIssue`, but I'll look at that in a separate PR.

src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
```diff
@@ -0,0 +1,142 @@
...
+                    selectionMonitor.SelectionChanged(selectedIssue);
+                }
+                catch(Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+                {
+                    // No-op
+                }
+            }
+
+            #region Event callbacks we are not interested in
+            void ITableControlEventProcessor.KeyDown(System.Windows.Input.KeyEventArgs args) { /* no-op */ }
```
duncanp-sonar, line 98: The code coverage is low because none of these unused methods are tested. I don't think it's worth adding tests for them, though.

src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
```diff
@@ -0,0 +1,142 @@
...
+                        issueObject is IAnalysisIssue issueFromTable)
+                    {
+                        selectedIssue = issueFromTable;
+                    }
+
+                    selectionMonitor.SelectionChanged(selectedIssue);
+                }
+                catch(Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+                {
+                    // No-op
```
duncanp-sonar, line 93: There aren't tests for the error handling. I'll add them.

src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
```diff
@@ -0,0 +1,142 @@
...
+using SonarLint.VisualStudio.Core.Analysis;
+
+namespace SonarLint.VisualStudio.IssueVisualization.TableControls
+{
+    [Export(typeof(ITableControlEventProcessorProvider))]
+    [Name(nameof(VSTableEventProcessorProvider))]
+    [Order(Before = Priority.Default)]
+    // We can specify multiple data source types and data source i.e. the same provider can be used for
+    // multiple table controls/sources
+    [DataSourceType(StandardTableDataSources.ErrorTableDataSource)]
```
rita-g-sonarsource, line 38: Do we have to specify DataSourceType? (e.g. would it work / make a difference in perf with just DataSource?)
If not, then will be a bit "cleaner" not to specify the type, so we could just say that we support source A and source B, without necessarily knowing their type.

src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+                this.wpfTableControl = wpfTableControl;
+                this.selectionMonitor = errorListSelectionMonitor;
+            }
+
+            void ITableControlEventProcessor.PostprocessSelectionChanged(TableSelectionChangedEventArgs e)
+            {
+                try
+                {
+                    IAnalysisIssue selectedIssue = null;
```
rita-g-sonarsource, line 79: Just for my understanding: this will be called on every selection change in the error list, right? So if we go from non-SL issue to non-SL issue, that will be called too?

src/IssueViz/TableControls/IssueTablesSelectionMonitor.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        private readonly ILogger logger;
+
+        [ImportingConstructor]
+        public IssueTablesSelectionMonitor(ILogger logger)
+        {
+            this.logger = logger;
+        }
+
+        void IIssueTablesSelectionMonitor.SelectionChanged(IAnalysisIssue selectedIssue)
```
rita-g-sonarsource, line 51: Will this be the place that calls the new SelectionService?

src/IssueViz.UnitTests/TableControls/IssueTablesSelectionMonitorTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.TableControls;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.TableControls
+{
+    [TestClass]
+    public class IssueTablesSelectionMonitorTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
```
rita-g-sonarsource, line 33: Interesting. I'll add that to my future classes as well.

src/IssueViz.UnitTests/TableControls/VSTableEventProcessorTests.cs
```diff
@@ -0,0 +1,194 @@
...
+
+            SimulatePostProcessEvent();
+
+            CheckMonitorCalledOnce(null);
+
+            SanityCheckSnapshotAccessedOnce();
+        }
+
+        [TestMethod]
+        public void SelectionChanged_SingleItem_TryGetValueReturnsTrueButObjectIsNull_PassesNullToMonitor()
```
rita-g-sonarsource, line 105: This means that we will never "clear" the selected issue in the SelectionService. So nothing will clear the panel/adornments. 
I wonder when we should clear it - maybe on analysis start?
I'm not sure what would be a more intuitive behavior.

src/IssueViz.UnitTests/TableControls/IssueTablesSelectionMonitorTests.cs
duncanp-sonar: If I remember correctly, the test helper is pretty smart.
You give it two lists: optional imports and required imports.
It will check that composition fails if any required imports are missing, and that composition succeeds if any optional import is missing.

src/IssueViz/TableControls/IssueTablesSelectionMonitor.cs
duncanp-sonar: Yes, I think so.

src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
duncanp-sonar: I think "type" in this context refers to the table control to which the data source is attached.
The attribute is required. If we want to specify that we are interested in every table control then we'd specify `StandardTableDataSources.AnyDataSource`.

src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
duncanp-sonar: Yes, it will be called with a `null` issue. We want to know when a flow issue isn't highlighted so we can clear the flow pane and adornments.

PR #1620, opened 2020-08-13
Add initial IssueVisualizationToolWindow + command to open it 

rita-g-sonarsource:
I couldn't find any other way to open the tool window, so in the end I just created a command. It's currently under 'Tools' like we did in the prototype. We could hide/move it elsewhere.
This branch is prep for #1615.


src/IssueViz/IssueVisualizationControl/IssueVisualizationToolWindow.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        public IssueVisualizationToolWindow(IServiceProvider serviceProvider) : base(null)
+        {
+            Caption = "SonarLint Issue Visualization";
+
+            var componentModel = serviceProvider.GetService(typeof(SComponentModel)) as IComponentModel;
+            var selectionService = componentModel.GetService<IAnalysisIssueSelectionService>();
+
+            //var viewModel = new IssueVisualizationViewModel(selectionService);
+            Content = new IssueVisualizationControl();
```
rita-g-sonarsource, line 42: Just a blank tool window for now. Is implemented in #1615.

src/IssueViz/Resources.resx
```diff
@@ -0,0 +1,126 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="ERR_VisualizationToolWindow_Exception" xml:space="preserve">
+    <value>Error in displaying Visualization Tool Window: {0}</value>
```
duncanp-sonar, line 121: "Error in displyaing..." doesn't sound right (and isn't consistent with our other error messages).
Just "Error displaying..." is enough. Same for the other error message.

rita-g-sonarsource: I took that from the CFamily command, which says "Error in calculating" 😄

src/IssueViz/IssueViz.csproj
```diff
@@ -11,7 +11,7 @@
   <PropertyGroup>
     <TargetFramework>net46</TargetFramework>
     <ProjectGuid>{94130D79-3F30-47B1-9C3F-0769E57D1F4D}</ProjectGuid>
-  <RootNamespace>SonarLint.VisualStudio.IssueVisualization</RootNamespace>
+    <RootNamespace>SonarLint.VisualStudio.IssueVisualization</RootNamespace>
```
duncanp-sonar, line 14: The root namespace is set twice, presumably as the result of an automatic merge.

src/IssueViz/IssueViz.csproj
```diff
@@ -30,6 +30,25 @@
...
+    <PackageReference Include="Microsoft.VisualStudio.ComponentModelHost" Version="14.0.25424" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.14.0" Version="14.3.25407" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime" Version="16.7.30328.74" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Page Include="IssueVisualizationControl\IssueVisualizationControl.xaml">
+      <Generator>XamlIntelliSenseFileGenerator</Generator>
+      <CopyToOutputDirectory>Never</CopyToOutputDirectory>
+    </Page>
```
duncanp-sonar, line 51: Grouping of related elements: VS seems to make additional/deletions to the project file at arbitrary locations.
The project file is easier to read related entries are grouped e.g.
* all PackageReferences together
* items that are removed and re-added with a different item type.

src/IssueViz.UnitTests/Commands/IssueVisualizationToolWindowCommandTests.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            SetupWindowCreation(new ToolWindowPane {Frame = Mock.Of<IVsWindowFrame>()}, new NotImplementedException("this is a test"));
+
+            VerifyExecutionDoesNotThrow();
+
+            VerifyMessageLogged("this is a test");
+        }
+
+        [TestMethod]
+        public void Execute_WindowCreatedWithIVsWindowFrame_WindowIsShown()
```
rita-g-sonarsource, line 109: It's actually not possible to create a window with a non-IVsWindowFrame, an exception is thrown by VS `new ToolWindowPane { Frame = new object() }` that it's not an IVsWindowFrame. So that's why there's no test for that.

src/IssueViz/IssueVisualizationControl/IssueVisualizationToolWindow.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl
+{
+    [ExcludeFromCodeCoverage]
+    [Guid("bb3677d1-3b5c-45a3-8a3a-897108c3ba28")]
+    public class IssueVisualizationToolWindow : ToolWindowPane
+    {
+        public IssueVisualizationToolWindow(IServiceProvider serviceProvider) : base(null)
+        {
+            Caption = "SonarLint Issue Visualization";
```
duncanp-sonar, line 36: This string should be localised too.

src/IssueViz/IssueVisualizationControl/IssueVisualizationToolWindow.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            var componentModel = serviceProvider.GetService(typeof(SComponentModel)) as IComponentModel;
+            var selectionService = componentModel.GetService<IAnalysisIssueSelectionService>();
```
duncanp-sonar, line 39: Doesn't one of the SDK analysis rules complain about these calls not being on the main thread?

src/IssueViz.UnitTests/Commands/IssueVisualizationToolWindowCommandTests.cs
```diff
@@ -0,0 +1,165 @@
...
+            VerifyExecutionDoesNotThrow();
+
+            frame.Verify(x => x.Show(), Times.Once);
+
+            VerifyMessageLogged("this is a test");
+        }
+
+        private void SetupWindowCreation(WindowPane windowPane, Exception exceptionToThrow = null)
+        {
+            var setup = package.Protected().Setup<WindowPane>("CreateToolWindow", typeof(IssueVisualizationToolWindow), 0);
```
rita-g-sonarsource, line 138: In the code we're calling FindToolWindow which is not virtual. I'm relaying on the fact that FindToolWindow calls CreateToolWindow, which _is_ virtual.

src/IssueViz/Commands/IssueVizPackage.vsct
```diff
@@ -0,0 +1,48 @@
...
+<CommandTable xmlns="http://schemas.microsoft.com/VisualStudio/2005-10-18/CommandTable" xmlns:xs="http://www.w3.org/2001/XMLSchema">
+
+  <Extern href="stdidcmd.h"/>
+  <Extern href="vsshlids.h"/>
+
+  <Commands package="guidIssueVizPackage">
+
+    <Groups>
+      <Group guid="guidIssueVizPackageCmdSet" id="SonarLintMenuGroup" priority="0x0100">
+        <Parent guid="guidSHLMainMenu" id="IDM_VS_MENU_TOOLS" />
```
duncanp-sonar, line 11: The normal place to put "view tool window" commands is on the Tools, Other Window menu.

src/IssueViz/IssueVisualizationControl/IssueVisualizationToolWindow.cs
rita-g-sonarsource: Hm, nope. I guess because it's called from IssueVizPackage.InstantiateToolWindow, which will always run on UI thread?

src/IssueViz/Commands/IssueVizPackage.vsct
rita-g-sonarsource: I moved it there. I had to take it out of the "SonarLintMenuGroup" though, since "Other Windows" doesn't seem to allow adding groups.

duncanp-sonar: `IDG_VS_WNDO_OTRWNDWS1` is the location used when  you create a new VS template and use the "Async Tool  Window" item template,

The nesting of menu items is a bit weird. If I remember correctly, Groups can contain Menus, and Menus can contain Groups, and Buttons and placed in Groups.

rita-g-sonarsource: I couldn't get the button to show when it's in a group. Based on github everyone is placing the command directly, but it could be because it's the default template as you said. I can place it in a group when a group is under View menu, but not under View -> Other Windows.

duncanp-sonar: That's probably because you were trying add a group to a group.
`IDG_VS_WNDO_OTRWNDWS1` is a group - in this case the prefix `IDG-` means "ID of a Group".
You should have been able to place your group on the parent `IDM_VS_MNUCTRL_OTHRWNDS` (where `IDM_` means "ID of a Menu").
Install the [Command Explorer](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.CommandExplorer) extension from MadsK. You can use it to see how menus and groups are organised in VS.

rita-g-sonarsource: Ah! thanks! I'll take a look at that and fix accordingly.

duncanp-sonar: I don't think there's anything fix - the VSCT is fine as it is.
My comment was just for info to explain why you couldn't add a group to a group.
If you were to add your own group to the menu then the command would appear at a different location, which isn't what we want in this case.

PR #1626, opened 2020-08-14
Move BoolToVisibilityConverter to Core

rita-g-sonarsource:
No product changes.


src/Core.UnitTests/WPF/BoolToVisibilityConverterTests.cs
```diff
@@ -44,10 +44,9 @@ public void BoolToVisibilityConverter_Convert_NonBoolInput_ThrowsArgumentExcepti
         {
             var converter = new BoolToVisibilityConverter();
 
-            Exceptions.Expect<ArgumentException>(() =>
-            {
-                converter.Convert("NotABoolean", typeof(Visibility), null, CultureInfo.InvariantCulture);
-            });
+            Action act = () => converter.Convert("NotABoolean", typeof(Visibility), null, CultureInfo.InvariantCulture);
```
rita-g-sonarsource, line 47: `Exceptions` is part of Integration and essentially can be replaced with Should.Throw.

PR #1628, opened 2020-08-16
Implement keyboard navigation 

### LINKED ISSUE #1603: Implement keyboard navigation
### Description

Implement key navigation: next/previous step, next/previous flow.
Keyboard navigation should work regardless of whether the focus is in the flow panel or the editor pane.

No need to handle navigation between issues - that is done via the Error List.

rita-g-sonarsource:
Fixes #1603

### DIFF
--- a/src/IssueViz.UnitTests/Selection/AnalysisIssueNavigationTests.cs
+++ b/src/IssueViz.UnitTests/Selection/AnalysisIssueNavigationTests.cs
@@ -1,0 +1,254 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Selection;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Selection
+{
+    [TestClass]
+    public class AnalysisIssueNavigationTests
+    {
+        private Mock<IAnalysisIssueSelectionService> selectionServiceMock;
+
+        private AnalysisIssueNavigation testSubject;
+
+        [TestInitialize]
+        public void TesInitialize()
+        {
+            selectionServiceMock = new Mock<IAnalysisIssueSelectionService>();
+
+            testSubject = new AnalysisIssueNavigation(selectionServiceMock.Object);
+        }
+
+        public enum GotoDirection
+        {
+            Next,
+            Previous
+        }
+
+        [TestMethod]
+        [DataRow(GotoDirection.Next)]
+        [DataRow(GotoDirection.Previous)]
+        public void GotoLocation_NoCurrentFlow_NoNavigation(GotoDirection direction)
+        {
+            SetupCurrentFlow((IAnalysisIssueFlowVisualization)null);
+            SetupCurrentLocation(CreateLocation(1, true));
+
+            if (direction == GotoDirection.Next)
+            {
+                testSubject.GotoNextLocation();
+            }
+            else
+            {
+                testSubject.GotoPreviousLocation();
+            }
+
+            VerifyNoNavigation();
+        }
+
+        [TestMethod]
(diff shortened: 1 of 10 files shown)


src/IssueViz/Commands/IssueVizPackage.vsct
```diff
@@ -30,6 +52,15 @@
 
   </Commands>
 
+  <!-- Short cut keybindings: see https://docs.microsoft.com/en-us/visualstudio/extensibility/binding-keyboard-shortcuts-to-menu-items?view=vs-2019 -->
+  <KeyBindings>
+    <KeyBinding guid="guidIssueVizPackageCmdSet" id="GotoNextLocationCommandId"
```
rita-g-sonarsource, line 57: It was hard to find free keys... ctrl+alt+arrows are used for resharper commands, ctrl+shift+arrows are used by VS.

I thought we could bind it to the num pad, with ctrl+4/6 for locations, and ctrl+5/8 for flows. 

In either case, the commands are in the command well, and will appear as "Tools.SonarLintGoToPreviousLocation" and "Tools.SonarLintGoToNextLocation".

src/IssueViz/Commands/IssueVizPackage.vsct
```diff
@@ -38,6 +69,8 @@
     <GuidSymbol name="guidIssueVizPackageCmdSet" value="{FDEF405A-28C2-4AFD-A37B-49EF2B0D142E}">
       <IDSymbol name="SonarLintMenuGroup" value="0x1020" />
       <IDSymbol name="IssueVisualizationToolWindowCommandId" value="0x0100" />
+      <IDSymbol name="GotoNextLocationCommandId" value="4129" />
```
duncanp-sonar, line 72: It would be better if all of the `value` entries used the same base, rather than mixing decimal and hex.

src/IssueViz/Commands/NavigationCommands.cs
```diff
@@ -0,0 +1,94 @@
...
+using Task = System.Threading.Tasks.Task;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Commands
+{
+    internal sealed class NavigationCommands
+    {
+        internal const int NextLocationCommandId = 4129;
+        internal const int PreviousLocationCommandId = 4130;
+
+        public static readonly Guid CommandSet = new Guid("FDEF405A-28C2-4AFD-A37B-49EF2B0D142E");
```
duncanp-sonar, line 37: How many places in code is this GUID used?

src/IssueViz/Resources.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="ERR_NavigationException" xml:space="preserve">
+    <value>Error navigating Analysis Issue location: {0}</value>
```
duncanp-sonar, line 121: "Error navigating to... " would read more naturally.

src/IssueViz/Commands/IssueVizPackage.vsct
duncanp-sonar: As discussed offline, lots of laptops/keyboards don't have numeric keypads.
There might be a suitable key chord combination we could use.

src/IssueViz/Commands/IssueVizPackage.vsct
```diff
@@ -22,6 +22,28 @@
...
+        <CommandFlag>CommandWellOnly</CommandFlag>
+        <Strings>
+          <ButtonText>Go to next issue location</ButtonText>
+          <CanonicalName>SonarLintGoToNextLocation</CanonicalName>
+          <LocCanonicalName>SonarLintGoToNextLocation</LocCanonicalName>
+          <CommandName>SonarLint Go to next issue location</CommandName>
+        </Strings>
+      </Button>
+
+      <Button guid="guidIssueVizPackageCmdSet" id="GotoPreviousLocationCommandId" priority="0x0100" type="Button">
```
duncanp-sonar, line 36: Specifying two commands in the same group with the same priority doesn't make sense, but it probably doesn't matter since the commands are never displayed in the UI.

src/IssueViz.UnitTests/Selection/AnalysisIssueNavigationTests.cs
```diff
@@ -0,0 +1,265 @@
...
+
+        [TestMethod]
+        [DataRow(GotoDirection.Next)]
+        [DataRow(GotoDirection.Previous)]
+        public void GotoLocation_NoCurrentFlow_NoNavigation(GotoDirection direction)
+        {
+            SetupCurrentFlow((IAnalysisIssueFlowVisualization)null);
+            SetupCurrentLocation(CreateLocation(1, true));
+
+            if (direction == GotoDirection.Next)
```
duncanp-sonar, line 68: Data-driven tests that have conditional behaviour based on the input always feel wrong to me - seems to go against the spirit of data-driven testing. Still, it doesn't make much difference in practice in this case.

src/IssueViz/Commands/NavigationCommands.cs
rita-g-sonarsource: Two. I extracted it to a Constants class.

src/IssueViz/Commands/IssueVizPackage.vsct
rita-g-sonarsource: Yep, I read that if two commands have the same priority, then they are just rendered in different order (which wouldn't matter, as they're not shown anyway).

src/IssueViz.UnitTests/Selection/AnalysisIssueNavigationTests.cs
rita-g-sonarsource: Yea, the IFs was more lines than the test itself. I split them to separate tests.

PR #1638, opened 2020-08-19
Add LineHash to IAnalysisIssueLocation

rita-g-sonarsource:


src/IssueViz/Editor/LineHashCalculator.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System;
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core.Helpers;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Editor
+{
+    public interface ILineHashCalculator
+    {
+        string Calculate(string wholeText, int line);
```
duncanp-sonar, line 29: `documentText` might be clearer than `wholeText`

Is `line` 0-based or 1-based? I think we should start being more explicit about this to avoid confusion or having to check in the code.
Maybe a doc comment, or we could make it explicit in the parameter name e.g. `oneBasedLineNumber`.

src/Core/Helpers/ChecksumCalculator.cs
```diff
@@ -26,10 +26,15 @@
...
+        string Calculate(string text);
+    }
+
     /// <summary>
     /// Calculates the checksum for an issue using the same method as the SonarQube server
     /// </summary>
-    public static class ChecksumCalculator
+    public class ChecksumCalculator : IChecksumCalculator
     {
         public static string Calculate(string text)
```
duncanp-sonar, line 39: Does this still need to be `public static`?

rita-g-sonarsource: Yep, it's still used statically by IssueToIssueMarkerConverter and LiveIssue

src/IssueViz.UnitTests/Editor/LineHashCalculatorTests.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        [TestMethod]
+        [DataRow("")]
+        [DataRow(null)]
+        public void Calculate_NoText_Null(string text)
+        {
+            var result = testSubject.Calculate(text, 10);
+
+            result.Should().BeNullOrEmpty();
```
duncanp-sonar, line 60: Can the assertion be more specific? Presumably we're in control of whether we're returning null/empty in this case?
Same for the other "null or empty" assertion in this file.

src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueToAnalysisIssueConverterTests.cs
```diff
@@ -119,14 +127,136 @@ public void Convert_EndLineIsZero()
...
             issue.StartLine.Should().Be(101);
+            issue.StartLineOffset.Should().Be(0);
 
             issue.EndLine.Should().Be(0);
-            issue.StartLineOffset.Should().Be(0);
             issue.EndLineOffset.Should().Be(0);
+        }
+
+        [TestMethod]
+        public void Convert_LocationEndLineIsNotZero_OffsetsAreSubstructed()
```
duncanp-sonar, line 137: `Substructed`? Is this a typo of "Subtracted"?
Either way, `OffsetsAreCalculatedCorrectly` might be better.

duncanp-sonar (approved):
Just few suggestions inline.

src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueToAnalysisIssueConverterTests.cs
rita-g-sonarsource: Ah, I knew something was off!

PR #1642, opened 2020-08-19
Open relevant document when selecting secondary location

### LINKED ISSUE #1641: Implement navigation to secondary location
### Description

When an analysis issue contains secondary locations that are in a different file (e.g. header file), clicking on the issue in the error list will open only the analyzed file.

There should be a way to navigate to open the document of the secondary location (e.g. when selecting a location in the navigation panel or via the keyboard shortcuts).

### TODO

Automatically open the document of the selected location.

rita-g-sonarsource:
Fixes #1641.

There is an undesired behavior when the analysis issue has secondary locations in different files **and** VS is set to display issues only for the current document / current project. When that happens and a user selects a secondary location, we navigate to a new file, which clears the error list, and with it the navigation panel. So the user has no way to navigate back.
I'm not sure how to handle that besides never clearing the navigation panel. But that would result in other undesired behavior, so it's a lose-lose.


src/IssueViz/Selection/IAnalysisIssueSelectionService.cs
```diff
@@ -32,7 +32,7 @@ internal interface IAnalysisIssueSelectionService : IDisposable
         IAnalysisIssueLocationVisualization SelectedLocation { get; set; }
     }
 
-    internal enum SelectionChangeLevel
+    public enum SelectionChangeLevel
```
rita-g-sonarsource, line 35: Has to be public, or else I can't use it in tests' DataRow

src/IssueViz/Selection/LocationChangedEventListener.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        private void SelectionService_SelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            var locationFilePath = e.SelectedLocation?.Location?.FilePath;
+
+            if (!string.IsNullOrEmpty(locationFilePath))
+            {
+                try
+                {
+                    documentOpener.Open(locationFilePath);
```
rita-g-sonarsource, line 59: We should also navigate to the relevant span, but I'm not sure how to do that.
In the prototype I've used the IVsTextView returned from `VsShellUtilities.OpenDocument` to get the buffer and create a span -- should I do that as well? How would that interact with the new tagger and the snapshot span?

src/IssueViz/Selection/IAnalysisIssueSelectionService.cs
duncanp-sonar: Ok. FYI I passed it as a string in my data tests and used `Enum.Parse`.

src/IssueViz/Selection/LocationChangedEventListener.cs
duncanp-sonar, line 59: It might be better if the document opener handled the exception, rather than making it the responsibility of every caller.

src/IssueViz/Selection/LocationChangedEventListener.cs
```diff
@@ -0,0 +1,73 @@
...
+                }
+                catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+                {
+                    logger.WriteLine(Resources.ERR_OpenDocumentException, locationFilePath, ex);
```
duncanp-sonar, line 63: Passing "ex" here will dump the call stack. Do we need it? Most of the time `ex.Message` is enough.

src/IssueViz/Selection/LocationChangedEventListener.cs
duncanp-sonar: As discussed offline, this won't affect the tagging.
Ping me when you've added the span navigation and I'll finish the review.

src/IssueViz/Selection/LocationChangedEventListener.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        private void SelectionService_SelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            var issueLocation = e.SelectedLocation?.Location;
+            var locationFilePath = issueLocation?.FilePath;
+
+            if (!string.IsNullOrEmpty(locationFilePath))
+            {
+                try
```
rita-g-sonarsource, line 63: I left the try-catch here since now it wraps multiple actions

src/IssueViz/Selection/LocationChangedEventListener.cs
rita-g-sonarsource: Done, please re-review

duncanp-sonar (approved):

PR #1648, opened 2020-08-20
IssueSpanCalculator will return null if LineHash is different

### LINKED ISSUE #1647: Change IssueSpanCalculator to take LineHash into consideration
### Description

Change IssueSpanCalculator so it would try to find a matching span also based on the issue's line hash.

rita-g-sonarsource:
Fixes #1647


rita-g-sonarsource:
@duncanp-sonar I suggest taking a look at the commits separately.
Commit 1: `WholeText` field was never used except to calculate the line hash inside `LiveIssue`'s ctor. So I just pass the line hash directly.
Commit 2: `IssueSpanCalculator` now has the responsibility to check if the line hash is the same. `IssueToIssueMarkerConverter` now just wraps the issue with the span, and all the calculations are inside `IssueSpanCalculator`.

src/Integration.Vsix/SonarLintTagger/IssueToIssueMarkerConverter.cs
```diff
@@ -51,30 +50,15 @@ public IssueMarker Convert(IAnalysisIssue issue, ITextSnapshot textSnapshot)
...
             {
                 throw new ArgumentNullException(nameof(textSnapshot));
             }
 
-            // SonarLint issues line numbers are 1-based, spans lines are 0-based
-
             var span = issueSpanCalculator.CalculateSpan(issue, textSnapshot);
 
-            // A start line of zero means the issue is file-level i.e. not associated with a particular line
-            if (issue.StartLine == 0)
```
rita-g-sonarsource, line 64: Logic moved to `IssueSpanCalculator`, so it could also be re-used in [document navigation PR](https://github.com/SonarSource/sonarlint-visualstudio/blob/ccc322205d6caa045705b33432b726397260915c/src/IssueViz/Selection/LocationChangedEventListener.cs#L66)

src/Integration.Vsix/Suppression/RoslynLiveIssueFactory.cs
```diff
@@ -112,7 +113,8 @@ public LiveIssue Create(SyntaxTree syntaxTree, Diagnostic diagnostic)
             {
                 var lineText = syntaxTree.GetText().Lines[lineSpan.EndLinePosition.Line].ToString();
                 var sonarQubeLineNumber = lineSpan.StartLinePosition.Line + 1; // Roslyn lines are 0-based, SonarQube lines are 1-based
-                return new LiveIssue(diagnostic.Id, projectGuid, lineSpan.Path, sonarQubeLineNumber, lineText); // Line-level issue
+                var lineHash = ChecksumCalculator.Calculate(lineText);
```
rita-g-sonarsource, line 116: Passing lineHash instead of lineText to remove `WholeText` property.

src/IssueViz.UnitTests/Editor/IssueSpanCalculatorTests.cs
```diff
@@ -0,0 +1,302 @@
...
+using Microsoft.VisualStudio.Text;
+using Moq;
+using SonarLint.VisualStudio.Core.Helpers;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Editor;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor
+{
+    [TestClass]
+    public class IssueSpanCalculatorTests
```
rita-g-sonarsource, line 32: Class was renamed and a few tests added, not sure why the diff is not shown correctly.

src/IssueViz/Editor/IIssueSpanCalculator.cs
```diff
@@ -21,27 +21,57 @@
...
+                return null;
+            }
+
+            // A start line of zero means the issue is file-level i.e. not associated with a particular line
+            var isFileLevelIssue = location.StartLine == 0;
+
             // SonarLint issues line numbers are 1-based, spans lines are 0-based
+            var startLine = currentSnapshot.GetLineFromLineNumber(location.StartLine - 1);
+
+            if (!isFileLevelIssue && IsLineHashDifferent(location, startLine))
```
rita-g-sonarsource, line 66: This is the new check

src/Integration.Vsix/SonarLintTagger/IssueMarker.cs
```diff
@@ -35,32 +35,26 @@ internal class IssueMarker : IFilterableIssue
...
         public SnapshotSpan Span { get; }
-        public string WholeLineText { get; set; }
-        public string LineHash { get; set; }
 
         string IFilterableIssue.RuleId => Issue.RuleKey;
 
         string IFilterableIssue.FilePath => Issue.FilePath;
 
-        string IFilterableIssue.LineHash => LineHash;
+        string IFilterableIssue.LineHash => Issue.LineHash;
```
rita-g-sonarsource, line 43: IssueMarker doesn't need a dedicated LineHash property anymore: if we created an IssueMarker, it means that we were successful in creating a SnapshotSpan -- which means that the line hash is the same as the issue, and we can just use `Issue.LineHash`

src/IssueViz/Editor/IIssueSpanCalculator.cs
```diff
@@ -62,5 +92,13 @@ public SnapshotSpan CalculateSpan(IAnalysisIssueLocation location, ITextSnapshot
 
             return new SnapshotSpan(start, end);
         }
+
+        private bool IsLineHashDifferent(IAnalysisIssueLocation location, ITextSnapshotLine snapshotLine)
```
rita-g-sonarsource, line 96: New code

src/IssueViz/Editor/IIssueSpanCalculator.cs
```diff
@@ -21,27 +21,57 @@
...
+        }
+        
+        internal IssueSpanCalculator(IChecksumCalculator checksumCalculator)
+        {
+            this.checksumCalculator = checksumCalculator;
+        }
+
+        public SnapshotSpan? CalculateSpan(IAnalysisIssueLocation location, ITextSnapshot currentSnapshot)
         {
+            if (location.StartLine > currentSnapshot.LineCount)
```
rita-g-sonarsource, line 53: Logic extracted from IssueToIssueMarkerConverter

src/Integration.Vsix/SonarLintTagger/IssueMarker.cs
```diff
@@ -35,32 +35,26 @@ internal class IssueMarker : IFilterableIssue
...
             Issue = issue;
             Span = span;
-            WholeLineText = wholeLineText;
-            LineHash = lineHash;
         }
 
         public IssueMarker Clone()
         {
-            return new IssueMarker(Issue, Span, WholeLineText, LineHash);
+            return new IssueMarker(Issue, Span);
```
rita-g-sonarsource, line 57: See [comment](https://github.com/SonarSource/sonarlint-visualstudio/pull/1648#discussion_r474244596)

src/Integration.Vsix.UnitTests/SonarLintTagger/IssueToIssueMarkerConverterTests.cs
```diff
@@ -24,69 +24,16 @@
...
 
 namespace SonarLint.VisualStudio.Integration.UnitTests
 {
     [TestClass]
     public class IssueToIssueMarkerConverterTests
     {
-        [DataTestMethod]
-        [DataRow(100, 1)]
-        [DataRow(101, 100)]
-        public void Convert_IssueLineOutsideSnapshot_ReturnsNull(int issueStartLine, int bufferLineCount)
```
rita-g-sonarsource, line 38: Moved to IssueSpanCalculatorTests

src/IssueViz.UnitTests/Editor/IssueSpanCalculatorTests.cs
duncanp-sonar: The poor diff experience makes it very difficult to tell which tests are the new tests, particularly since it looks like you've renamed the existing tests. 
In future PRs it would help if you called out the new tests so reviewers could focus on them, rather than having to review the whole file.

duncanp-sonar: I'm guessing the diff doesn't show correctly because you've made so many changes to the existing tests that git doesn't recognise it as a renamed file. The volume of change effectively means the whole file has to be reviewed.

src/IssueViz.UnitTests/Editor/IssueSpanCalculatorTests.cs
```diff
@@ -0,0 +1,302 @@
...
+            var mockSnapshot = CreateSnapshotMock(bufferLineCount);
+
+            // Act and assert
+            var result = testSubject.CalculateSpan(issue, mockSnapshot.Object);
+            result.Should().BeNull();
+
+            checksumCalculatorMock.VerifyNoOtherCalls();
+        }
+
+        [DataTestMethod]
```
duncanp-sonar, line 62: This isn't a `DataTestMethod`.

src/IssueViz/Editor/IIssueSpanCalculator.cs
```diff
@@ -21,27 +21,57 @@
...
+                // Race condition: the line reported in the diagnostic is beyond the end of the file, so presumably
+                // the file has been edited while the analysis was being executed
+                return null;
+            }
+
+            // A start line of zero means the issue is file-level i.e. not associated with a particular line
+            var isFileLevelIssue = location.StartLine == 0;
+
             // SonarLint issues line numbers are 1-based, spans lines are 0-based
+            var startLine = currentSnapshot.GetLineFromLineNumber(location.StartLine - 1);
```
duncanp-sonar, line 64: Calling `currentSnapshot.GetLineFromLineNumber(-1)` in VS throws an exception, so I expect this line to throw for file-level issues. Does it actually work in VS?

Related to this, are there tests for file level issues? You said that tests had been moved from `IIssueToIssueMarkerConverterTests`, but I can't work out which one is the equivalent of `Convert_FileLevelIssue_ReturnsFilterableIssue`, and I can't see any tests where the issue StartLine is zero.

duncanp-sonar (changes requested):
See comments - main question is about the effect of calling `currentSnapshot.GetLineFromLineNumber` with a value of -1.

src/IssueViz/Editor/IIssueSpanCalculator.cs
rita-g-sonarsource: We always [calculated](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/SonarLintTagger/IssueToIssueMarkerConverter.cs#L60) the span, even for a zero start line. It's possible that it doesn't work, I've asked Abbas for a rule that can raise a file level issue and I will check it.

Regarding the test -- I forgot to commit one (CalculateSpan_FileLevelIssue_HashNotChecked) and I'll write another one that compares the expected positions (once I verify that it does work).

PR #1649, opened 2020-08-20
Preserve Error List item selection when snapshot changes

### LINKED ISSUE #1645: Preserve ErrorList selection when data source content has not changed
### Description

The ErrorList has a mechanism for preserving the selection when the content provided by an `ITableEntriesSnapshot` has changed but the previously selected issue is still present in the new snapshot. Our implementation does not implement this mechanism, so editing a document can cause the Error List selection to be lost if the selected item relates to document being edited.

This bug has existed for a long time without really being noticeable. However, it will make a big difference when visualising secondary locations, as the act of editing the document causes the Error List selection to be lost, which in turn triggers a "selection changed" event. This has unwanted side effects when visualising secondary locations (it sets the tracked issue to null, causing the flow pane to be cleared).

### Repro steps

1. Analyse a C++ file
2. Select an issue in the file in the Error List
3. Edit the file

### Expected behavior

The Error List selection is preserved.

### Actual behavior

The ErrorList  selection is not preserved.

### Known workarounds

None

### Related information

* SonarLint for VS Version: all
* Visual Studio Version: all
* SonarQube server version: n/a
* Languages in your solution: reproduces for Javascript, C, C++

duncanp-sonar:
Fixes #1645


src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -48,24 +50,70 @@ internal class IssuesSnapshot : WpfTableEntriesSnapshotBase
...
+        /// Every snapshot that has the same analysis id describes the same set of
+        /// issues in the same order. The implementation of "IndexOf" depends on this.
+        /// </summary>
+        public Guid AnalysisRunId { get; }
+
         public IEnumerable<IssueMarker> IssueMarkers => readonlyIssueMarkers;
 
-        internal IssuesSnapshot(string projectName, string filePath, int versionNumber, IEnumerable<IssueMarker> issueMarkers)
-            : this(projectName, filePath, versionNumber, issueMarkers, new AnalysisSeverityToVsSeverityConverter(), new RuleHelpLinkProvider())
+        // Every snapshot has a unique version number. It doesn't matter what it is, as
```
duncanp-sonar, line 62: The handling of version numbers doesn't need to be visible outside the class. We only need to ensure it's an incrementing number.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -48,24 +50,70 @@ internal class IssuesSnapshot : WpfTableEntriesSnapshotBase
...
+        // will assume the data hasn't changed and won't update the rows).
+        private static int nextVersionNumber = 0;
+        private static int GetNextVersionNumber() => ++nextVersionNumber;
+
+        #region Construction methods
+
+        /// <summary>
+        /// Create a snapshot with new set of issues from a new analysis run
+        /// </summary>
+        public static IssuesSnapshot CreateNew(string projectName, string filePath, IEnumerable<IssueMarker> issueMarkers) =>
```
duncanp-sonar, line 73: I've added three methods for the higher-level operations we want to perform. Each of those methods set the `AnalysisRunId` appropriately, depending on whether the snapshot relates to a different set of issues or the same set of locations with modified location spans.

If we could reliably (and cheaply) compare two lists of issue markers then we could automatically work out which operation to perform. However, this solution is much simpler and less risky.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -186,21 +189,21 @@ private void SafeOnBufferChange(object sender, TextContentChangedEventArgs e)
             return newMarkers;
         }
 
-        private void SnapToNewSnapshot(IssuesSnapshot newIssues)
+        private void SnapToNewSnapshot(IssuesSnapshot newSnapshot)
```
duncanp-sonar, line 192: No code changes in this method, just renaming the variable.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -265,18 +269,6 @@ private IEnumerable<IssueMarker> RemoveSuppressedIssues(IEnumerable<IssueMarker>
             return filteredIssues.OfType<IssueMarker>().ToArray();
         }
 
-        private void RefreshIssues()
-        {
-            UpdateIssues(Factory.CurrentSnapshot.IssueMarkers ?? Enumerable.Empty<IssueMarker>());
```
duncanp-sonar, line 270: `IssueMarkers` is always set to a value so the new code doesn't have an equivalent of the null-coalesce condition (and in fact it wasn't necessary in the old version of the code either).

Issue #1639, opened 2020-08-19
Check location navigability when buffer changes


rita-g-sonarsource:
### Description

Add location navigability check that will set AnalysisIssueLocationVisualization.IsNavigable when the buffer changes.

PR #1653, opened 2020-08-24
Check location navigability when selecting a location

rita-g-sonarsource:


rita-g-sonarsource:
@duncanp-sonar I suggest taking a look at the commits separately.
I don't feel that the work is done - mainly, I've been trying hard _not_ to make `IAnalysisIssueLocationVisualization` implement `INotifyPropertyChanged` (see the last commit), but I don't think that my solution will hold in the future. There are 3 additional scenarios that I can think of, where it might not fit:

1. In keyboard navigation, I think we'd want to skip to the next/previous location if the "navigable" location turned out to be non-navigable. For that, I'd have to change `[NavigateToMatchingLocation](https://github.com/SonarSource/sonarlint-visualstudio/blob/branch-dataflow/src/IssueViz/Selection/AnalysisIssueNavigation.cs#L77)` to something like this:
```
var navigableLocations = order(currentFlow.Locations)
    .Where(x =>
        x.IsNavigable &&
        match(x, currentLocation));

foreach (var locationViz in navigableLocations)
{
    selectionService.Select(locationViz);

    // very ugly: relaying on the knowledge that `Select` will also navigate and change IsNavigable.
    if (selectionService.SelectedLocation.IsNavigable)
    {
        break;
    }
}
```

2. I'm not sure how the Buffer change will fit with this design (#1639). I don't know how we would raise the "recalculate navigability" event, and what would it do. The current implementation relies on the fact that you try to select a location, whereas with the buffer, we don't want to select all the locations just to trigger the navigability. So I think I would be forced to extract the IsNavigable=<value> out of the SelectionService, and then probably implement INotifyPropertyChanged to propagate it to the UI.

3. Minor: when we select a flow, we automatically select the first location. It depends on how we implement the buffer tagging, but there might be a corner case where the location can be non-navigable by the time that you click on the item in the error list. If anything, the user can always edit the file outside of VS. So I think that should be doing the "Go to next navigable" logic instead, which means re-using the navigation logic of the keyboard shortcuts (which have a problem currently as pointed out above).

I'd love to hear your ideas, I'm at wits end trying to solve it without implementing INotifyPropertyChanged.

src/IssueViz/Editor/LocationNavigator.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using System;
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Editor
+{
+    internal interface ILocationNavigator
```
rita-g-sonarsource, line 29: Logic from the deleted `LocationChangedEventListener`

src/IssueViz/Selection/AnalysisIssueSelectionService.cs
```diff
@@ -29,46 +29,34 @@ namespace SonarLint.VisualStudio.IssueVisualization.Selection
...
+            SelectedFlow = flowVisualization;
+            SelectedLocation = GetFirstLocationOrDefault();
 
-                RaiseSelectionChanged(SelectionChangeLevel.Flow);
-            }
+            RaiseSelectionChanged(SelectionChangeLevel.Flow);
         }
 
-        public IAnalysisIssueLocationVisualization SelectedLocation
+        public void Select(IAnalysisIssueLocationVisualization locationVisualization)
```
duncanp-sonar: One problem with both the old and new versions is that there's no checking of the logical consistency of the data i.e. the caller could set the flow/location to objects that aren't related to the current issue.
It's not necessarily a blocker to using this approach, but ideally it wouldn't be possible.

src/IssueViz/Models/AnalysisIssueLocationVisualization.cs
```diff
@@ -25,23 +25,21 @@ namespace SonarLint.VisualStudio.IssueVisualization.Models
     internal interface IAnalysisIssueLocationVisualization
     {
         int StepNumber { get; }
-
-        bool IsNavigable { get; }
-
         IAnalysisIssueLocation Location { get; }
+        bool IsNavigable { get; set; }
```
duncanp-sonar, line 29: FYI I'm just writing down ideas as a I look at the code at the moment. Don't treat them as review comments that need to be actioned.

I agree this property needs to be dynamically settable. The buffer tagger should play a part in updating it - if it can't translate a span, it should update the navigability.
Maybe it should be a calculated property: if we have span then it implies that the location is navigable. No span, not navigable.

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -129,7 +129,7 @@
                     <Setter Property="Background" Value="{DynamicResource {x:Static vsShell:VsBrushes.StartPageSelectedItemBackgroundKey}}"/>
                     <Setter Property="BorderBrush" Value="{DynamicResource {x:Static vsShell:VsBrushes.StartPageSelectedItemBackgroundKey}}" />
                 </Trigger>
-                <DataTrigger Binding="{Binding Location.IsNavigable}" Value="False">
+                <DataTrigger Binding="{Binding IsNavigable}" Value="False">
```
duncanp-sonar, line 132: This is definitely what we want from a XAML point of view.

duncanp-sonar:
Let's think about it from a different angle. A file is either open or closed.
If the file is closed, we can't know if any locations in it are still navigable (and we don't need to, until the point the user opens the file or navigates to the location).

At the point the user opens the document, we need to translate from issue location data to a span in the document. We might find at that point that a location is un-navigable, so we'd need to update the navigability and have it reflected in the flow panel.

If the file open, we should be able to track span changes as the user edits the file. If a location becomes un-navigable (e.g. because the code has been deleted), we'd want the flow pane to update immediately and show the location as un-navigable.

The analysis issue results are being stored at the moment in the Error List (or more accurately our custom table data source, which has one factory/snapshot per open file). Our `TextBufferIssueTracker` stores data in `IssueMarkers`, which a `SnapshotSpan` for each primary issue location. At present, we don't store any `SnapshotSpan` information in our issue visualization classes. We will need to store span information somewhere. There's some overlap between the `IssueMarker` and location viz classes, but neither stores all the data we need. I'm not sure we need both; we can probably enrich one of them and drop the other.

I don't have a particular problem with the issue service data classes implementing `INotifyPropertyChanged`. It does seem slightly odd, but we already have change events and we will need more to handle navigability etc.

One difference between the current tagger implementation and our new requirements is that the old tagger only had to handle primary issue locations, which by definition meant there was an open file and a tagger that could track changes.
Secondary locations means we can issue locations in a file that is not opened, so we might not find out that a secondary location is not navigable until we open the file or try to navigate to it. I think it makes sense to update the navigability of the location at that point like this new code is doing. I'll if I can make that requirement fit with the new "issue location service" I'm trying to create to support the buffer tracking.

src/IssueViz/Models/AnalysisIssueLocationVisualization.cs
rita-g-sonarsource: It's [already](https://github.com/SonarSource/sonarlint-visualstudio/blob/c8d0305aee7ea570d4f03284fd8310bb7389791f/src/IssueViz/Editor/LocationNavigator.cs#L64) somewhat calculated based on if there is a span or not. The navigability is currently:
1. Open the document
2. Get the document text view
2. Calculate span
3. [IOutliningManager.ExpandAll](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.text.outlining.ioutliningmanager.expandall?view=visualstudiosdk-2019)
4. [IEditorOperations.SelectAndMoveCaret](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.text.operations.ieditoroperations.selectandmovecaret?view=visualstudiosdk-2019#Microsoft_VisualStudio_Text_Operations_IEditorOperations_SelectAndMoveCaret_Microsoft_VisualStudio_Text_VirtualSnapshotPoint_Microsoft_VisualStudio_Text_VirtualSnapshotPoint_Microsoft_VisualStudio_Text_Editor_TextSelectionMode_System_Nullable_Microsoft_VisualStudio_Text_Editor_EnsureSpanVisibleOptions__)

So the span is currently only one of element of the equation. Although, perhaps we can assume that if we managed to get the text view AND calculate the span, then we will most likely succeed in the other operations as well.

rita-g-sonarsource:
> Let's think about it from a different angle. A file is either open or closed.
> If the file is closed, we can't know if any locations in it are still navigable (and we don't need to, until the point the user opens the file or navigates to the location).
> 
> At the point the user opens the document, we need to translate from issue location data to a span in the document. We might find at that point that a location is un-navigable, so we'd need to update the navigability and have it reflected in the flow panel.
> 
> If the file open, we should be able to track span changes as the user edits the file. If a location becomes un-navigable (e.g. because the code has been deleted), we'd want the flow pane to update immediately and show the location as un-navigable.
> 
> The analysis issue results are being stored at the moment in the Error List (or more accurately our custom table data source, which has one factory/snapshot per open file). Our `TextBufferIssueTracker` stores data in `IssueMarkers`, which a `SnapshotSpan` for each primary issue location. At present, we don't store any `SnapshotSpan` information in our issue visualization classes. We will need to store span information somewhere. There's some overlap between the `IssueMarker` and location viz classes, but neither stores all the data we need. I'm not sure we need both; we can probably enrich one of them and drop the other.
> 
> I don't have a particular problem with the issue service data classes implementing `INotifyPropertyChanged`. It does seem slightly odd, but we already have change events and we will need more to handle navigability etc.
> 
> One difference between the current tagger implementation and our new requirements is that the old tagger only had to handle primary issue locations, which by definition meant there was an open file and a tagger that could track changes.
> Secondary locations means we can issue locations in a file that is not opened, so we might not find out that a secondary location is not navigable until we open the file or try to navigate to it. I think it makes sense to update the navigability of the location at that point like this new code is doing. I'll if I can make that requirement fit with the new "issue location service" I'm trying to create to support the buffer tracking.

I opened a new PR with INotifyPropertyChanged implementation. It made things easier and less hacky. 
I'm currently updating IsNavigable directly, it's not a calculated property yet. I agree with your comment that maybe we should hold the Span instead. I think once the rest of the pieces fall into place we could revisit it and modify accordingly.

PR #1654, opened 2020-08-25
Check location navigability when selecting a location

### LINKED ISSUE #1650: Check location navigability when selecting a location
### Description

Add location navigability check that will set AnalysisIssueLocationVisualization.IsNavigable when the user tries to navigable to a location.

rita-g-sonarsource:
Fixes #1650


src/IssueViz.UnitTests/Editor/LocationNavigatorTests.cs
```diff
@@ -28,94 +28,79 @@
...
 using SonarLint.VisualStudio.IssueVisualization.Editor;
-using SonarLint.VisualStudio.IssueVisualization.Models;
-using SonarLint.VisualStudio.IssueVisualization.Selection;
 
-namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Selection
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor
 {
     [TestClass]
-    public class LocationChangedEventListenerTests
+    public class LocationNavigatorTests
```
rita-g-sonarsource, line 35: Replaced the EventListener with a simple service and modified the tests accordingly.

src/IssueViz/Editor/LocationNavigator.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        [ImportingConstructor]
+        internal LocationNavigator(IDocumentNavigator documentNavigator, IIssueSpanCalculator spanCalculator, ILogger logger)
+        {
+            this.documentNavigator = documentNavigator;
+            this.spanCalculator = spanCalculator;
+            this.logger = logger;
+        }
+
+        public bool TryNavigate(IAnalysisIssueLocation location)
```
rita-g-sonarsource, line 50: Extracted code from the deleted EventListener

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -125,10 +125,16 @@
                     <Setter Property="Background" Value="{DynamicResource {x:Static vsShell:VsBrushes.CommandBarHoverOverSelectedKey}}"/>
                     <Setter Property="BorderBrush" Value="{DynamicResource {x:Static vsShell:VsBrushes.CommandBarHoverOverSelectedIconBorderKey}}" />
                 </Trigger>
-                <Trigger Property="IsSelected" Value="True">
-                    <Setter Property="Background" Value="{DynamicResource {x:Static vsShell:VsBrushes.StartPageSelectedItemBackgroundKey}}"/>
-                    <Setter Property="BorderBrush" Value="{DynamicResource {x:Static vsShell:VsBrushes.StartPageSelectedItemBackgroundKey}}" />
-                </Trigger>
+                <MultiDataTrigger>
```
rita-g-sonarsource, line 128: This is a small UI hack.
Reminder of the binding cycle: the user click on an item in the list --> selection service is updated --> selection services raises an event --> UI is updated.
Unfortunately, if the item is not actually navigable, there is no way for me to break this cycle. I can't undo the user's selection, it's already too late. WPF does not have a BeforeSelectionChanged event and any trick that I tried (e.g. not setting the view model property to the new `value`, or not raising the selection event, or reverting it back to the previous selection, etc) has resulted in side effects.

So instead, I just make the item appear as not selected. 🤷‍♀️

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -39,11 +40,12 @@ internal sealed class IssueVisualizationViewModel : INotifyPropertyChanged, IDis
         private readonly IAnalysisIssueSelectionService selectionEvents;
         private readonly IVsImageService2 vsImageService;
         private readonly IRuleHelpLinkProvider ruleHelpLinkProvider;
+        private readonly ILocationNavigator locationNavigator;
         private readonly ILogger logger;
 
         private IAnalysisIssueVisualization currentIssue;
         private IAnalysisIssueFlowVisualization currentFlow;
-        private LocationListItem currentLocation;
+        private LocationListItem currentLocationListItem;
```
rita-g-sonarsource, line 48: Renamed for better readability

src/IssueViz/Models/AnalysisIssueVisualizationConverter.cs
```diff
@@ -74,12 +66,7 @@ private IReadOnlyList<IAnalysisIssueLocationVisualization> Convert(IEnumerable<I
             var locationNumber = 1;
 
             var convertedLocations = locations
-                .Select(location =>
-                {
-                    var isNavigable = navigationChecker.IsNavigable(location);
-
-                    return new AnalysisIssueLocationVisualization(locationNumber++, isNavigable, location);
-                })
+                .Select(location => new AnalysisIssueLocationVisualization(locationNumber++, location))
```
rita-g-sonarsource, line 69: The navigationChecker was never used so I just removed it.
The navigability will be handled now by the buffer and the explicit check on selection.

src/IssueViz/Selection/AnalysisIssueNavigation.cs
```diff
@@ -22,36 +22,39 @@
...
+using SonarLint.VisualStudio.IssueVisualization.Editor;
 using SonarLint.VisualStudio.IssueVisualization.Models;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Selection
 {
     internal interface IAnalysisIssueNavigation
     {
-        void GotoNextLocation();
-        void GotoPreviousLocation();
+        void GotoNextNavigableLocation();
```
rita-g-sonarsource, line 32: Renamed for better accuracy

src/IssueViz/Selection/AnalysisIssueNavigation.cs
```diff
@@ -22,36 +22,39 @@
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Linq;
+using SonarLint.VisualStudio.IssueVisualization.Editor;
 using SonarLint.VisualStudio.IssueVisualization.Models;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Selection
 {
     internal interface IAnalysisIssueNavigation
```
rita-g-sonarsource, line 30: Speaking of naming, I think the name of this service is confusing when we also have ILocationNavigator.
Maybe it should be something like ISelectedLocationCycler? (and in the future, ISelectedFlowCycler, with one service SelectionCycler?)

src/IssueViz/Selection/AnalysisIssueNavigation.cs
```diff
@@ -74,14 +77,20 @@ private void NavigateToMatchingLocation(Func<IReadOnlyList<IAnalysisIssueLocatio
...
+            var navigableLocations = order(currentFlow.Locations)
+                .Where(x =>
                     x.IsNavigable &&
                     match(x, currentLocation));
 
-            if (navigateToLocation != null)
+            foreach (var locationViz in navigableLocations)
             {
-                selectionService.SelectedLocation = navigateToLocation;
+                locationViz.IsNavigable = locationNavigator.TryNavigate(locationViz.Location);
```
rita-g-sonarsource, line 87: New code: navigability update

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -141,12 +148,24 @@ private void SelectionEvents_SelectionChanged(object sender, SelectionChangedEve
...
         }
 
+        private void NavigateToLocation(IAnalysisIssueLocationVisualization locationVisualization)
+        {
+            if (locationVisualization == null)
+            {
+                return;
+            }
+
+            locationVisualization.IsNavigable = locationNavigator.TryNavigate(locationVisualization.Location);
```
rita-g-sonarsource, line 166: New logic: navigability update

src/IssueViz/Selection/AnalysisIssueNavigation.cs
```diff
@@ -74,14 +77,20 @@ private void NavigateToMatchingLocation(Func<IReadOnlyList<IAnalysisIssueLocatio
                 return;
             }
 
-            var navigateToLocation = order(currentFlow.Locations)
-                .FirstOrDefault(x =>
+            var navigableLocations = order(currentFlow.Locations)
```
rita-g-sonarsource, line 80: This is why I needed to replace the EventListener with a service that I can call: I want to keep checking until I find a navigable issue and update the selection service only at the end.

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
duncanp-sonar: A simpler UX would just to allow all list items to be selected, and if the location is navigable then we open the file in the editor. The keyboard navigation commands could do the same. The behaviour would be consistent and the code simpler.

I'm not convinced that the UX of "only navigable items can be selected" is clearly more intuitive that "all items are selectable - navigate to the file if possible". What do we do in the other IDEs?

src/IssueViz/Selection/AnalysisIssueNavigation.cs
duncanp-sonar: Could be `GotoNextNavigableStep` / `GotoNextNavigableFlowStep` to make it clearer that this is navigation with a single issue flow.

src/IssueViz/Selection/AnalysisIssueNavigation.cs
duncanp-sonar: `ILocationNavigationCommands`, `IIssueStepNavigator`, `IIssueFlowStepNavigator`...

duncanp-sonar (approved):

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
rita-g-sonarsource: As we discussed offline with the PM, lets keep the non-navigable items not selectable.

PR #1658, opened 2020-08-27
Add SnapshotSpan? to issue vizualization...

duncanp-sonar:
... and change the `IssueSnapshot` to store issue visualizations.

Preparatory work for the new tagging code.

For the new tagging code to work, we need to be able to store spans for primary and secondary locations in our Error List table data source. The simplest way to do this is to add span to `IAnalysisIssueLocationVisualization`, and change the `IssueSnapshot` to store `IAnalysisIssueLocationVisualization` instances instead of `IssueMarker` instances.

Ultimately, we won't actually need the `IssueMarker` class at all. I started trying to remove it completely as part of this change, but it required many more changes. Instead, I made a temporary change to the `IssueMarker` so it is created using an issue viz rather than an issue.

Now that the issue viz is stored in the error list snapshot, we need to create the issue viz instances when we process the results of an analysis. Previously, we had two separate converters - one that converted an `IAnalysisIssue` into an `IssueMarker` (which calculated the span), and one that converted an `IAnalysisIssue` into an `IAnalysisIssueVisualization`.  We need to combine the functionality of these two converters.

For now, to limit the volume of the changes, I've modified issue marker converter so it calls the issue viz converter. I'll consolidate them later in a separate PR.


src/IssueViz/Models/AnalysisIssueLocationVisualization.cs
```diff
@@ -18,19 +18,31 @@
...
+        /// <summary>
+        /// Editor span associated with the issue location
+        /// </summary>
+        /// <remarks>
+        /// The span gives the current region in an open editor document for the issue. This can be different
+        /// from the Location of the issue, which is fixed at time of analysis.
+        /// It will be null if the document has not been opened in the editor.
+        /// It will be an empty span if the issue location does not exist in the document due to editing.
+        /// </remarks>
+        SnapshotSpan? Span { get; set; }
```
duncanp-sonar, line 45: Key part of the change - add a span to the location.

src/IssueViz/Models/AnalysisIssueFlowVisualization.cs
```diff
@@ -23,7 +23,7 @@
 
 namespace SonarLint.VisualStudio.IssueVisualization.Models
 {
-    internal interface IAnalysisIssueFlowVisualization
+    public interface IAnalysisIssueFlowVisualization
```
duncanp-sonar, line 26: The error list/snapshot/analysis processing classes are all in the `Integration.VSIX` project, so they now need to be able to refer to the issue visualization types and converter. I've made the interfaces public but not the classes.

src/IssueViz/Models/AnalysisIssueVisualization.cs
```diff
@@ -40,5 +41,9 @@ public AnalysisIssueVisualization(IReadOnlyList<IAnalysisIssueFlowVisualization>
 
         public IReadOnlyList<IAnalysisIssueFlowVisualization> Flows { get; }
         public IAnalysisIssue Issue { get; }
+        public int StepNumber => 0;
+        public bool IsNavigable { get; set; }
+        public IAnalysisIssueLocation Location => Issue;
+        public SnapshotSpan? Span { get; set; }
```
duncanp-sonar, line 47: The issue viz will also need a span and the location properties for tagging, so I've made it implement the location viz interface.

src/IssueViz/Models/AnalysisIssueVisualization.cs
```diff
@@ -40,5 +41,9 @@ public AnalysisIssueVisualization(IReadOnlyList<IAnalysisIssueFlowVisualization>
 
         public IReadOnlyList<IAnalysisIssueFlowVisualization> Flows { get; }
         public IAnalysisIssue Issue { get; }
+        public int StepNumber => 0;
+        public bool IsNavigable { get; set; }
```
duncanp-sonar, line 45: The location viz class now implements `INotifyPropertyChanged`. We don't currently need it for this class but that might change.

src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
```diff
@@ -76,17 +76,17 @@ void ITableControlEventProcessor.PostprocessSelectionChanged(TableSelectionChang
             {
                 try
                 {
-                    IAnalysisIssue selectedIssue = null;
+                    IAnalysisIssueVisualization selected = null;
```
duncanp-sonar, line 79: We're now storing the issue viz in the Error List so we can fetch it directly, rather than fetching an issue and converting it separately.

src/IssueViz/TableControls/IssueTablesSelectionMonitor.cs
```diff
@@ -42,21 +41,18 @@ internal interface IIssueTablesSelectionMonitor
     internal class IssueTablesSelectionMonitor : IIssueTablesSelectionMonitor
     {
         private readonly IAnalysisIssueSelectionService selectionService;
-        private readonly IAnalysisIssueVisualizationConverter converter;
 
         [ImportingConstructor]
-        public IssueTablesSelectionMonitor(IAnalysisIssueSelectionService selectionService, IAnalysisIssueVisualizationConverter converter)
+        public IssueTablesSelectionMonitor(IAnalysisIssueSelectionService selectionService)
```
duncanp-sonar, line 46: This class handles selection change notifications from the error list. Previously it was being handed an issue which it converted to a issue viz. Now it's being handed an issue viz, so it doesn't need to do the conversion.

src/Integration.Vsix/SonarLintTagger/IssueMarker.cs
```diff
@@ -33,8 +36,11 @@ namespace SonarLint.VisualStudio.Integration.Vsix
     /// </remarks>
     internal class IssueMarker : IFilterableIssue
     {
-        public IAnalysisIssue Issue { get; }
-        public SnapshotSpan Span { get; }
+        public IAnalysisIssueVisualization IssueViz { get; }
```
duncanp-sonar, line 39: The changes to this class are hacks to minimise the churn in this PR.
This class will be removed later.

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
```diff
@@ -70,19 +70,13 @@ internal class AccumulatingIssueConsumer : IIssueConsumer
...
         public delegate void OnIssuesChanged(IEnumerable<IssueMarker> issueMarkers);
 
-        public AccumulatingIssueConsumer(ITextSnapshot analysisSnapshot, string analysisFilePath, OnIssuesChanged onIssuesChangedCallback)
-            : this(analysisSnapshot, analysisFilePath, onIssuesChangedCallback, new IssueToIssueMarkerConverter())
-        {
-        }
-
-        internal /* for testing */ AccumulatingIssueConsumer(ITextSnapshot analysisSnapshot, string analysisFilePath, OnIssuesChanged onIssuesChangedCallback,
-            IIssueToIssueMarkerConverter issueMarkerConverter)
+        public AccumulatingIssueConsumer(ITextSnapshot analysisSnapshot, string analysisFilePath, OnIssuesChanged onIssuesChangedCallback, IIssueToIssueMarkerConverter issueMarkerConverter)
```
duncanp-sonar, line 73: The `IIssueToIssueMarkerConverter` is now more complicated than it was before so unfortunately it can't be created directly anymore.

src/Integration.Vsix/SonarLintTagger/IssueToIssueMarkerConverter.cs
```diff
@@ -58,7 +62,8 @@ public IssueMarker Convert(IAnalysisIssue issue, ITextSnapshot textSnapshot)
 
             var span = issueSpanCalculator.CalculateSpan(issue, textSnapshot);
 
-            return span == null ? null : new IssueMarker(issue, span.Value);
+            var issueViz = issueVizConverter.Convert(issue);
```
duncanp-sonar, line 65: This converter is called when the issue consumer processes analysis results.
Previously, it was calculating the span for each issue at this point. Now it's calculating the span and converting the issue into an issue viz.
I'll combine the converters into a single type later.

src/Integration.Vsix.UnitTests/SonarLintTagger/IssueMarkerTests.cs
```diff
@@ -25,20 +25,22 @@
 using SonarLint.VisualStudio.Core.Analysis;
 using SonarLint.VisualStudio.Core.Suppression;
 using SonarLint.VisualStudio.Integration.Vsix;
+using SonarLint.VisualStudio.IssueVisualization.Models;
```
duncanp-sonar, line 28: I fixed up all of the existing tests so they pass, but in fact quite a few of the affected tests in the older text buffer/tagger classes will be deleted later.

Perhaps I should have deleted all of the old tagger behaviour and tests first to avoid having to fix them up. However, it's been useful to be able run the code and manually validate that I haven't broken anything.

src/Integration.Vsix/SonarLintTagger/IssueMarker.cs
```diff
@@ -33,8 +36,11 @@ namespace SonarLint.VisualStudio.Integration.Vsix
...
+
+        public IAnalysisIssue Issue => IssueViz.Issue;
+
+        public SnapshotSpan Span => IssueViz.Span.Value;
```
rita-g-sonarsource, line 43: Check if it has value?

duncanp-sonar: There didn't seem to be much point doing any special here since the whole class is going to be removed.

src/Integration.Vsix/SonarLintTagger/IssueToIssueMarkerConverter.cs
```diff
@@ -58,7 +62,8 @@ public IssueMarker Convert(IAnalysisIssue issue, ITextSnapshot textSnapshot)
...
+            return span == null ? null : new IssueMarker(issueViz, span.Value);
```
rita-g-sonarsource, line 66: Just for my understanding: 
The issue viz will always be created only if there is a span.
That span can later be changed to a null if the code changes, but it will never be null at the beginning, correct?

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -73,6 +74,7 @@ internal sealed class TaggerProvider : IViewTaggerProvider
             [Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider,
             ISonarLanguageRecognizer languageRecognizer,
             IAnalysisRequester analysisRequester,
+            IIssueToIssueMarkerConverter converter,
```
rita-g-sonarsource, line 77: Future suggestion: I think it's worth having a MEF-able TextBufferIssueTrackerFactory, instead of receiving dependencies just to pass them through. Same for the AccumulatingIssueConsumer.

src/Integration.Vsix/SonarLintTagger/IssueToIssueMarkerConverter.cs
duncanp-sonar: Currently yes, since that's what the previous code did. That will probably change in the future when I merge the issue marker and an issue viz converters.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
```diff
@@ -35,21 +36,26 @@ public class AccumulatingIssueConsumerTests
     {
         private static readonly IAnalysisIssue ValidIssue = CreateIssue(startLine: 1, endline: 1);
         private static readonly ITextSnapshot ValidTextSnapshot = CreateSnapshot(lineCount: 10);
+        private static readonly IIssueToIssueMarkerConverter ValidConverter = Mock.Of<IIssueToIssueMarkerConverter>();
```
rita-g-sonarsource, line 39: Is there a reason to call it Valid? Seems a bit confusing, as if it has a setup logic. Can be just `ConverterMock`.

rita-g-sonarsource (approved):
LGTM.
When you delete the IssueMarker class, I think it's worth creating a Factory class for the AccumulatingIssueConsumer and TextBufferIssueTracker. The ctors are getting bloated and the tests keep changing with passing dummy mocks. It would be much simpler to have a MEF factory for them.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
duncanp-sonar: It's not typed as a mock so it should have "Mock" in the name. I use "Valid" as a prefix for reusable instances where the value isn't important and we don't need to do any checks against the instance.

Issue #1661, opened 2020-08-28
Add IIssueLocationStore service


duncanp-sonar:
### Description

The new service will provide callers with a list of all primary and secondary locations for a particular file.

We have a table data source factory for each file we analyse. However, the analysis results could contain secondary locations that are in other files so to find all of the issue locations for a specific file we will need to search across all of our factories. This service hides the details of finding the locations from callers (of which there will be only one for now - the new base issue location tagger).

- [x] Define interface
- [x] Modify ISonarErrorLIstDataSource to implement the new interface
- [x] Modify IssueSnapshot to implement "get locations for file" methods
- [x] Tests

Relates to https://jira.sonarsource.com/browse/MMF-2098

PR #1666, opened 2020-08-28
(1) Added file-retrieval methods to IIssuesSnapshot

duncanp-sonar:
Relates to #1661

### DIFF
--- a/src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests.cs
+++ b/src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests.cs
@@ -37,7 +37,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         private const int IndexOf_NotFoundResult = -1;
         private const string ValidProjectName = "aproject";
         private const string ValidFilePath = "c:\\file.txt";
-        private readonly IEnumerable<IssueMarker> ValidMarkerList = new IssueMarker[] { CreateMarker() };
+        private readonly IEnumerable<IssueMarker> ValidMarkerList = new IssueMarker[] { CreateMarkerWithSpan() };
 
         [TestMethod]
         public void Construction_CreateNew_SetsProperties()
@@ -82,8 +82,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         public void Construction_UpdateIssues_PreservesIdAndUpdatesVersion()
         {
             // Both snapshots must have same number of issues for this test case
-            var originalIssues = new IssueMarker[] { CreateMarker("hash1") };
-            var newIssues = new IssueMarker[] { CreateMarker("hash2") };
+            var originalIssues = new IssueMarker[] { CreateMarkerWithSpan("hash1") };
+            var newIssues = new IssueMarker[] { CreateMarkerWithSpan("hash2") };
 
             // Sanity check
             originalIssues.Should().NotBeEquivalentTo(newIssues);
@@ -103,8 +103,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         [TestMethod]
         public void Construction_ReviseIssues_NumberOfIssuesChanged_Throws()
         {
-            var originalIssues = new IssueMarker[] { CreateMarker(), CreateMarker() };
-            var newIssues = new IssueMarker[] { CreateMarker() };
+            var originalIssues = new IssueMarker[] { CreateMarkerWithSpan(), CreateMarkerWithSpan() };
+            var newIssues = new IssueMarker[] { CreateMarkerWithSpan() };
 
             var original = IssuesSnapshot.CreateNew(ValidProjectName, ValidFilePath, originalIssues);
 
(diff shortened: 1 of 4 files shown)


src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -27,14 +27,35 @@
...
+        /// <summary>
+        /// Returns the set of files for which this snapshot contains location information (primary and secondary)
+        /// </summary>
+        IEnumerable<string> FilesInSnapshot { get; }
+
+        /// <summary>
+        /// Returns all of the location visualizations (primary and secondary) that are in the specified file
+        /// </summary>
+        /// <returns></returns>
+        IEnumerable<IAnalysisIssueLocationVisualization> GetLocationsVizsForFile(string filePath);
```
duncanp-sonar, line 58: Added two new members that will be consumed by the `IIssueLocationStore` implementation (to be added in the next PR).
The tagger will ask the `IIssueLocationStore` for all issue locations relating to a particular file; the `IIssueLocationStore` will in turn loop round all of the snapshots, calling these methods to get the relevant files.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -56,15 +77,6 @@ internal class IssuesSnapshot : WpfTableEntriesSnapshotBase, IIssuesSnapshot
...
 
-        /// <summary>
-        /// Logical identifier for a set of issues produced by an analysis run.
-        /// Every snapshot that has the same analysis id describes the same set of
-        /// issues in the same order. The implementation of "IndexOf" depends on this.
-        /// </summary>
-        public Guid AnalysisRunId { get; }
-
-        public IEnumerable<IssueMarker> IssueMarkers => readonlyIssueMarkers;
-
```
duncanp-sonar, line 67: Moved to be with the other interface members.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -116,10 +128,14 @@ private IssuesSnapshot(Guid snapshotId, string projectName, string filePath, IEn
             this.ruleHelpLinkProvider = ruleHelpLinkProvider;
             this.issueMarkers = new List<IssueMarker>(issueMarkers);
             this.readonlyIssueMarkers = new ReadOnlyCollection<IssueMarker>(this.issueMarkers);
+
+            this.FilesInSnapshot = CalculateFilesInSnapshot(issueMarkers);
```
duncanp-sonar, line 132: Optimisation to avoid having to walk all of the issues/flows/locations to determine if there are any relevant files.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -240,5 +256,46 @@ public override int IndexOf(int currentIndex, ITableEntriesSnapshot newSnapshot)
...
+
+        private static IEnumerable<string> CalculateFilesInSnapshot(IEnumerable<IssueMarker> markers)
+        {
+            var allFilePaths = EnumerateAllLocationVisualizations(markers)
+                .Select(locViz => locViz.Location.FilePath);
+
+            return new HashSet<string>(allFilePaths, StringComparer.OrdinalIgnoreCase);
+        }
+
+        private static IEnumerable<IAnalysisIssueLocationVisualization> EnumerateAllLocationVisualizations(IEnumerable<IssueMarker> markers) =>
```
rita-g-sonarsource, line 291: Maybe 'GetAllLocationVisualizations' ? `Enumerate` implies to me that you receive a callback to do something with the enumeration. `Get` is more accurate IMO.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
rita-g-sonarsource: Perhaps we should calculate a `Dictionary<string,IAnalysisIssueLocationVisualization` instead. `FilesInSnapshot` could then be Dictionary.Keys, and GetLocationsVizsForFile will be O(1).

src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests.cs
```diff
@@ -161,16 +223,42 @@ private static T GetValue<T>(ITableEntriesSnapshot snapshot, string columnName)
             return (T)content;
         }
 
-        private static IssueMarker CreateMarker(string ruleKey = "rule key") =>
+        private static IssueMarker CreateMarkerWithSpan(string ruleKey = "rule key") =>
             new IssueMarker(CreateIssueViz(new DummyAnalysisIssue { RuleKey = ruleKey }), new SnapshotSpan());
 
-        private static IAnalysisIssueVisualization CreateIssueViz(IAnalysisIssue issue)
+        private static IssueMarker CreateMarkerWithPath(string primaryFilePath, params IAnalysisIssueFlowVisualization[] flowVizs) =>
```
rita-g-sonarsource, line 229: `CreateMarkerWithSpan` implies that `CreateMarkerWithPath` doesn't have a span.
Maybe a single `CreateMarker` with default parameters and use named parameters to initialize them in usage?

 `private static IssueMarker CreateMarker(string ruleKey = "rule key", string primaryFilePath = null, params IAnalysisIssueFlowVisualization[] flowVizs)`

rita-g-sonarsource (approved):

src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests.cs
duncanp-sonar: I knew there was a reason I didn't have a single method; if you writing something like `CreateMarker(primaryFilePath: "foo", CreateFlowViz("bar"));`
... the compiler complains that `primaryFilePath` is used out of position and is followed by an unnamed argument.

I'll rename the methods to `CreateMarker` and `CreateMarkerWIthSpecifcPaths`.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
duncanp-sonar: The dictionary would have to be string->list of locations.

As discussed offline, I'm only trying to optimise the most common case of taggers being able quickly work out that they don't need to do any work because the file they manage isn't affected.
If it turns out we need to optimise the lookup too then we can do that separately.

src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests.cs
rita-g-sonarsource: You can remove the `params`, and do `CreateMarker(primaryFilePath:"foo", flows: new []{ CreateFlowViz("bar") }`

duncanp-sonar: Yes, but that's more cumbersome, particularly since we don't have any tests that need set both the rule key and the paths.

PR #1667, opened 2020-08-28
(2) Added factory parameter to RefreshErrorList(...)

duncanp-sonar:
Means we can tell the error list which factory has updated, rather than it checking them all

### DIFF
--- a/src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
+++ b/src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
@@ -80,12 +80,35 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.ErrorList
         }
 
         [TestMethod]
-        public void Refresh_NoSubscribedSinksNoError()
+        public void Refresh_NullArg_Throws()
         {
             var testSubject = new SonarErrorListDataSource(mockTableManagerProvider.Object);
 
-            Action act = () => testSubject.RefreshErrorList();
+            Action act = () => testSubject.RefreshErrorList(null);
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("factory");
+        }
+
+        [TestMethod]
+        public void Refresh_NoSubscribedSinks_NoError()
+        {
+            var testSubject = CreateTestSubjectWithFactory(ValidFactory);
+
+            Action act = () => testSubject.RefreshErrorList(ValidFactory);
             act.Should().NotThrow();
+        }
+
+        [TestMethod]
+        public void Refresh_FactoryIsNotRegistered_SinksAreNotNotified()
+        {
+            var mockSink = new Mock<ITableDataSink>();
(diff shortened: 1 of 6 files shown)


src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
```diff
@@ -80,39 +80,62 @@ public void Subscribe_ReturnsNewDisposableToken()
         }
 
         [TestMethod]
-        public void Refresh_NoSubscribedSinksNoError()
```
duncanp-sonar, line 83: This test hasn't changed - now stars on line 91.

src/Integration.Vsix/ErrorList/ISonarErrorListDataSource.cs
```diff
@@ -25,7 +25,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix
     /// </summary>
     internal interface ISonarErrorListDataSource
     {
-        void RefreshErrorList();
+        void RefreshErrorList(SnapshotFactory factory);
```
duncanp-sonar, line 28: Tell the Error List that it needs to refresh a single factory, rather than have it check all of the factories.

Actually, the main reason for this change is so that when it comes to notifying the taggers that something has changed (in a future PR), we won't need to notify the taggers of all open documents. We'll be able to notify only the taggers that deal with files that have issues in the factory.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -369,7 +369,7 @@ public void WhenNoIssuesAreFound_ListenersAreUpdated()
 
         private void CheckErrorListRefreshWasRequestedOnce()
         {
-            mockSonarErrorDataSource.Verify(x => x.RefreshErrorList(), Times.Once);
+            mockSonarErrorDataSource.Verify(x => x.RefreshErrorList(It.IsAny<SnapshotFactory>()), Times.Once);
```
rita-g-sonarsource, line 372: Does it have to be Any? We should be able to verify against the relevant factory.

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
```diff
@@ -80,39 +80,62 @@ public void Subscribe_ReturnsNewDisposableToken()
...
         {
             var testSubject = new SonarErrorListDataSource(mockTableManagerProvider.Object);
 
-            Action act = () => testSubject.RefreshErrorList();
+            Action act = () => testSubject.RefreshErrorList(null);
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("factory");
+        }
+
+        [TestMethod]
+        public void Refresh_NoSubscribedSinksNoError()
```
rita-g-sonarsource, line 92: Refresh_NoSubscribedSinks_NoError

rita-g-sonarsource (approved):

PR #1668, opened 2020-08-29
(3) Added IIssueLocationStore interface

duncanp-sonar:
* implemented GetLocations and refresh of error list

Relates to #1661

### DIFF
--- a/src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_IssueLocationStoreTests.cs
+++ b/src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_IssueLocationStoreTests.cs
@@ -1,0 +1,165 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using FluentAssertions;
+using Microsoft.VisualStudio.Shell.TableManager;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Integration.Vsix;
+using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.ErrorList
+{
+    [TestClass]
+    public class SonarErrorListDataSource_IssueLocationStoreTests
+    {
+        private const string ValidPath = "valid.txt";
+
+        [TestMethod]
+        public void GetLocations_NullArg_Throws()
+        {
+            var testSubject = CreateTestSubject();
+
+            Action act = () => testSubject.GetLocations(null);
+            act.Should().ThrowExactly<ArgumentNullException>("filePath");
+        }
+
+        [TestMethod]
+        public void GetLocations_NoFactories_EmptyListReturned()
+        {
+            var testSubject = CreateTestSubject();
+
+            var actual = testSubject.GetLocations(ValidPath);
+            actual.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void GetLocations_FactoryButNoMatches_EmptyListReturned()
+        {
+            var testSubject = CreateTestSubject();
+
+            var factory1 = CreateFactoryWithLocationVizs(ValidPath);
+            testSubject.AddFactory(factory1);
+
+            var actual = testSubject.GetLocations(ValidPath);
+
+            actual.Should().BeEmpty();
+            CheckSnapshotGetLocationsCalled(factory1);
+        }
+
+        [TestMethod]
+        public void GetLocations_FactoriesWithMatches_ExpectedIssuesReturned()
+        {
+            var testSubject = CreateTestSubject();
+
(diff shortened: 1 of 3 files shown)


src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -141,7 +160,38 @@ public void RemoveFactory(SnapshotFactory factory)
...
+            foreach (var factory in currentFactories)
+            {
+                locVizs.AddRange(factory.CurrentSnapshot.GetLocationsVizsForFile(filePath));
+            }
+            return locVizs;
+        }
+
+        public void LocationsUpdated(IEnumerable<string> affectedFilePaths)
+        {
+            // TODO
```
duncanp-sonar, line 189: Done in the next commit (which also uses the IssuesChanged event).

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -21,17 +21,23 @@
...
 using Microsoft.VisualStudio.Shell.TableManager;
+using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+using SonarLint.VisualStudio.IssueVisualization.Models;
 using SonarLint.VisualStudio.IssueVisualization.TableControls;
 
 namespace SonarLint.VisualStudio.Integration.Vsix
 {
     [Export(typeof(ISonarErrorListDataSource))]
+    [Export(typeof(IIssueLocationStore))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
```
duncanp-sonar, line 36: I've added a MEF test in the next commit.

src/IssueViz/Editor/LocationTagging/IIssueLocationStore.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <summary>
+        /// Returns all primary and secondary locations for the specified file
+        /// </summary>
+        IEnumerable<IAnalysisIssueLocationVisualization> GetLocations(string filePath);
+
+        /// <summary>
+        /// Used by callers to notify the service that the locations in the specified
+        /// files have been updated
+        /// </summary>
+        void LocationsUpdated(IEnumerable<string> affectedFilePaths);
```
rita-g-sonarsource, line 43: It should be `UpdateLocations`, `LocationsUpdated` sounds like an event.

src/IssueViz/Editor/LocationTagging/IIssueLocationStore.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        public IssuesChangedEventArgs(IEnumerable<string> affectedFiles)
+        {
+            AffectedFiles = affectedFiles ?? Array.Empty<string>();
+        }
+
+        /// <summary>
+        /// The set of files impacted by the new issues
+        /// </summary>
+        public IEnumerable<string> AffectedFiles { get; }
```
rita-g-sonarsource, line 56: Maybe `AnalyzedFiles` would be more accurate, WDYT? `Affected` has a change-vibe to it, whereas this event would be raised only on analysis.

src/IssueViz/Editor/LocationTagging/IIssueLocationStore.cs
```diff
@@ -0,0 +1,58 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging
+{
+    public interface IIssueLocationStore
```
rita-g-sonarsource, line 27: Does it have to be public? Will the Vsix use it?

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_IssueLocationStoreTests.cs
```diff
@@ -0,0 +1,165 @@
...
+            testSubject.AddFactory(factory1);
+
+            var actual = testSubject.GetLocations(ValidPath);
+
+            actual.Should().BeEmpty();
+            CheckSnapshotGetLocationsCalled(factory1);
+        }
+
+        [TestMethod]
+        public void GetLocations_FactoriesWithMatches_ExpectedIssuesReturned()
```
rita-g-sonarsource, line 70: Maybe add a check for a factory with a different file path?

rita-g-sonarsource (approved):

src/IssueViz/Editor/LocationTagging/IIssueLocationStore.cs
duncanp-sonar: Yes. It's implemented by `SonarErrorListDataSource` which is in the Integration assembly currently.

src/IssueViz/Editor/LocationTagging/IIssueLocationStore.cs
duncanp-sonar: The doc comment says what `affectedFilePaths` means. 

`INotifyPropertyChanged` is implemented by the type that is doing the notification i.e. the event source. Here, the service is the sink i.e. it's being notified that something has changed. For me `NotifyLocationsChanged` is more confusing.
I'm happy to rename the method to something else.  I'll change it to `LocationsModified` for now.

We could pass the set of changed locations, but it's more work and not necessary.
The tagger doesn't currently track which locations have changed - it just asks the TextSnapshot to update all of the affected spans.
The service doesn't care which individual locations have changed. It's only interested in knowing the files that have changed so it can tell the appropriate factories to change.

rita-g-sonarsource: `LocationsModified` is a bit problematic, it's an event name, not a function name (same for `LocationsUpdated` for that matter).
Also, the function is supposed to notify the error list, no? In which case, it is the one doing the notification. `NotifyX` sounds right to me, unless I'm missing something.

duncanp-sonar: The callers are notifying the locations service that they have updated some of its data.
The interface is the contract between the implementing type and it's callers. What the implementing type does in response to the call is an implementation detail.

rita-g-sonarsource: The API is still off. The callers are notifying the service with seemingly arbitrary data: why just file names? why not the locations list? -- it's because you know that's what you need for the error list, and you know that all the locations service is gonna do is notify the error list with this exact information. Everything relies on the implementation details.
Maybe I'll change my mind when I see how it all comes together, but for now it feels too specific.

duncanp-sonar: I'll change the method name to `Refresh` for now.

Issue #1662, opened 2020-08-28
Issue tagging: handle file renames


duncanp-sonar:
### Description

Handle file renames for open files and update the error list/flow UI appropriately.

- [x] Detect file renames and fix up the issue visualization data
- [x] Propagate the changes to the Error List
- [x] Propagate the changes the flow panel.

Relates to https://jira.sonarsource.com/browse/MMF-2098

PR #1672, opened 2020-09-01
Handle file renames in Issue panel

rita-g-sonarsource:
Part of #1662

As we discussed offline:
The panel's locations list contains fake "file" nodes. In order to update those nodes, I'm implementing NotifyPropertyChanged, receiving an instance of the first `IAnalysisIssueVisualizationLocation` that comes after it in the list and binding to its PropertyChanged.


src/IssueViz/Models/AnalysisIssueVisualization.cs
```diff
@@ -42,8 +43,12 @@ public AnalysisIssueVisualization(IReadOnlyList<IAnalysisIssueFlowVisualization>
...
         public IAnalysisIssue Issue { get; }
         public int StepNumber => 0;
-        public bool IsNavigable { get; set; }
         public IAnalysisIssueLocation Location => Issue;
+
+        public bool IsNavigable { get; set; }
+        public string FilePath { get; set; }
         public SnapshotSpan? Span { get; set; }
+
+        public event PropertyChangedEventHandler PropertyChanged;
```
rita-g-sonarsource, line 52: This is temporary.
I will introduce a new interface instead, in a different PR.

```
public interface IAnalysisIssueSecondaryLocationVisualization : IAnalysisIssueLocationVisualization, INotifyPropertyChanged
    {
        int StepNumber { get; }
    }
```

src/IssueViz/Models/AnalysisIssueLocationVisualization.cs
```diff
@@ -45,21 +47,33 @@ public interface IAnalysisIssueLocationVisualization
...
             Location = location;
             IsNavigable = true;
+            FilePath = location.FilePath;
         }
 
         public int StepNumber { get; }
 
         public IAnalysisIssueLocation Location { get; }
 
+        public string FilePath
```
rita-g-sonarsource, line 67: New property: settable FilePath that the tagger could update.

src/IssueViz/IssueVisualizationControl/ViewModels/FileNameLocationListItemCreator.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using Microsoft.VisualStudio.Shell.Interop;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl.ViewModels
+{
+    internal interface IFileNameLocationListItemCreator
+    {
+        IFileNameLocationListItem Create(IAnalysisIssueLocationVisualization location);
```
rita-g-sonarsource, line 29: FileNameLocationListItem is now doing a lot of stuff, and I didn't want to further complicate the tests of the IssueVisualizationViewModel.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -179,6 +171,21 @@ private void NavigateToLocation(IAnalysisIssueLocationVisualization locationVisu
             locationVisualization.IsNavigable = locationNavigator.TryNavigate(locationVisualization.Location);
         }
 
+        private void UpdateLocationsList()
+        {
+            if (LocationListItems != null)
+            {
+                foreach (var disposable in LocationListItems.OfType<IDisposable>())
```
rita-g-sonarsource, line 178: Avoiding memory leak when switching between issues/flows.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -207,20 +217,6 @@ private IReadOnlyList<ILocationListItem> BuildLocationListItems(IAnalysisIssueFl
             return listItems;
         }
 
-        private object GetImageMonikerForFile(string filePath)
```
rita-g-sonarsource, line 210: Moved to FileNameLocationListItem

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -192,12 +199,15 @@ private IReadOnlyList<ILocationListItem> BuildLocationListItems(IAnalysisIssueFl
             {
                 for (var i = 0; i < flowLocations.Count; i++)
                 {
-                    var filePath = flowLocations[i].Location.FilePath;
```
rita-g-sonarsource, line 195: Moved to FileNameLocationListItem

src/IssueViz/Models/AnalysisIssueLocationVisualization.cs
```diff
@@ -25,12 +25,14 @@
 
 namespace SonarLint.VisualStudio.IssueVisualization.Models
 {
-    public interface IAnalysisIssueLocationVisualization
+    public interface IAnalysisIssueLocationVisualization : INotifyPropertyChanged
```
rita-g-sonarsource, line 28: I had to implement it on the interface so that FileNameLocationListItem could register to it.

src/IssueViz/Models/AnalysisIssueLocationVisualization.cs
duncanp-sonar: It's not obvious why the viz type has a property of the same name as the `Location` it wraps.
It would help if the property was called something different e.g. `CurrentFilePath`, with a doc comment on the interface explaining why it can be different from the underlying `Location.FilePath`.

src/IssueViz/Models/AnalysisIssueVisualization.cs
duncanp-sonar: As discussed offline, let's defer this until we decide if/how we are going to show the primary location in the UI.

src/IssueViz/IssueVisualizationControl/ViewModels/FileNameLocationListItem.cs
```diff
@@ -18,20 +18,86 @@
...
+        public event PropertyChangedEventHandler PropertyChanged;
+
+        private void NotifyPropertyChanged([CallerMemberName] string propertyName = null)
+        {
+            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+        }
+
+        public void Dispose()
+        {
+            location.PropertyChanged -= Location_PropertyChanged;
```
duncanp-sonar, line 86: Out of interest, doesn't our "implement IDisposable correctly" complain about this implementation?

duncanp-sonar (approved):

src/IssueViz/IssueVisualizationControl/ViewModels/FileNameLocationListItem.cs
rita-g-sonarsource: Nope, since the class is sealed.

PR #1673, opened 2020-09-02
Fixed IssuesSnapshot.FilesInSnapshot to include the analyzed file path

duncanp-sonar:


src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -300,12 +300,17 @@ public IEnumerable<IAnalysisIssueLocationVisualization> GetLocationsVizsForFile(
...
+            var allLocationFilePaths = GetAllLocationVisualizations(markers)
                 .Select(locViz => locViz.Location.FilePath);
 
-            return new HashSet<string>(allFilePaths, StringComparer.OrdinalIgnoreCase);
+            var files = new HashSet<string>(allLocationFilePaths, StringComparer.OrdinalIgnoreCase);
+
+            // The list of files should always contain the name of the file being analyzed. This is to handle the case
+            // where the file has been analyzed but doesn't contain any issues.
+            files.Add(analyzedFilePath);
+            return files;
```
duncanp-sonar, line 313: Scenario: 
1) analyze file `X.cpp` -> 1 issues. Tagger updates correctly.
2) edit the file to remove the issue -> save -> re-analyze -> 0 issues.
Currently the tagger won't update correctly because when the `IssuesChanged` event args won't include `X.cpp` in the list of impacted files so the tagger won't clear its existing tags.

In fact, this will happen for every re-analysis because of the way we report issues incrementally - the first thing we do it is to create a new snapshot with no issues.

rita-g-sonarsource (approved):

PR #1675, opened 2020-09-02
Change IsNavigable to rely on SnapshotSpan

rita-g-sonarsource:
The main changes are in 
1. `IAnalysisIssueLocationVisualization` - removing IsNavigable property and adding extension methods that rely on Span.
2. `LocationNavigator` - now uses the existing `IAnalysisIssueLocationVisualization.Span`, and is responsible to update it if it's null.
The rest of the files were changed as a result.

### DIFF
--- a/src/Core/WPF/AggregatorValueConverter.cs
+++ b/src/Core/WPF/AggregatorValueConverter.cs
@@ -1,0 +1,43 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Windows.Data;
+
+namespace SonarLint.VisualStudio.Core.WPF
+{
+    /// <summary>
+    /// https://riptutorial.com/wpf/example/16146/group-multiple-converters--ivalueconverter-
+    /// </summary>
+    public class AggregatorValueConverter : List<IValueConverter>, IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
+            return this.Aggregate(value, (current, converter) => converter.Convert(current, targetType, parameter, culture));
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/IssueToIssueMarkerConverterTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/IssueToIssueMarkerConverterTests.cs
@@ -57,13 +57,13 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         }
 
         [TestMethod]
-        public void Convert_NullSpan_Null()
+        public void Convert_EmptySpan_Null()
         {
             var mockIssue = Mock.Of<IAnalysisIssue>();
             var mockSnapshot = Mock.Of<ITextSnapshot>();
 
             var spanCalculator = new Mock<IIssueSpanCalculator>();
-            spanCalculator.Setup(x => x.CalculateSpan(mockIssue, mockSnapshot)).Returns((SnapshotSpan?) null);
+            spanCalculator.Setup(x => x.CalculateSpan(mockIssue, mockSnapshot)).Returns(new SnapshotSpan());
             
             var testSubject = new IssueToIssueMarkerConverter(Mock.Of<IAnalysisIssueVisualizationConverter>(), spanCalculator.Object);
             var issueMarker = testSubject.Convert(mockIssue, mockSnapshot);
@@ -72,18 +72,19 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         }
 
         [TestMethod]
-        public void Convert_SpanIsNotNull_ReturnsMarkerWithSpan()
+        public void Convert_SpanIsNotEmpty_ReturnsMarkerWithSpan()
         {
             var mockIssue = Mock.Of<IAnalysisIssue>();
-            var mockSnapshot = Mock.Of<ITextSnapshot>();
-            var mockSpan = new SnapshotSpan();
+            var mockSnapshot = new Mock<ITextSnapshot>();
+            mockSnapshot.SetupGet(x => x.Length).Returns(20);
+            var mockSpan = new SnapshotSpan(mockSnapshot.Object, 0, 10);
 
             var spanCalculator = new Mock<IIssueSpanCalculator>();
-            spanCalculator.Setup(x => x.CalculateSpan(mockIssue, mockSnapshot)).Returns(mockSpan);
+            spanCalculator.Setup(x => x.CalculateSpan(mockIssue, mockSnapshot.Object)).Returns(mockSpan);
             var passThroughConverter = CreatePassthroughConverter();
 
             var testSubject = new IssueToIssueMarkerConverter(passThroughConverter, spanCalculator.Object);
-            var issueMarker = testSubject.Convert(mockIssue, mockSnapshot);
+            var issueMarker = testSubject.Convert(mockIssue, mockSnapshot.Object);
 
             issueMarker.Should().NotBeNull();
             issueMarker.Issue.Should().Be(mockIssue);
--- a/src/Integration.Vsix/SonarLintTagger/IssueToIssueMarkerConverter.cs
+++ b/src/Integration.Vsix/SonarLintTagger/IssueToIssueMarkerConverter.cs
@@ -62,8 +62,14 @@ namespace SonarLint.VisualStudio.Integration.Vsix
 
             var span = issueSpanCalculator.CalculateSpan(issue, textSnapshot);
 
+            if (span.IsEmpty)
+            {
+                return null;
+            }
+
             var issueViz = issueVizConverter.Convert(issue);
-            return span == null ? null : new IssueMarker(issueViz, span.Value);
+
+            return new IssueMarker(issueViz, span);
         }
     }
 }
--- a/src/IssueViz.UnitTests/Editor/LocationNavigatorTests.cs
+++ b/src/IssueViz.UnitTests/Editor/LocationNavigatorTests.cs
@@ -28,6 +28,7 @@ using SonarLint.VisualStudio.Core.Analysis;
 using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.Integration.UnitTests;
 using SonarLint.VisualStudio.IssueVisualization.Editor;
+using SonarLint.VisualStudio.IssueVisualization.Models;
 
 namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor
 {
@@ -37,6 +38,9 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor
         private Mock<IDocumentNavigator> documentOpenerMock;
         private Mock<IIssueSpanCalculator> spanCalculatorMock;
         private TestLogger logger;
+
+        private Mock<ITextView> textViewMock;
+        private ITextSnapshot textViewCurrentSnapshotMock;
 
         private LocationNavigator testSubject;
 
@@ -48,6 +52,10 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor
             logger = new TestLogger();
 
             testSubject = new LocationNavigator(documentOpenerMock.Object, spanCalculatorMock.Object, logger);
+
+            textViewCurrentSnapshotMock = Mock.Of<ITextSnapshot>();
+            textViewMock = new Mock<ITextView>();
+            textViewMock.SetupGet(x => x.TextBuffer.CurrentSnapshot).Returns(textViewCurrentSnapshotMock);
         }
 
         [TestMethod]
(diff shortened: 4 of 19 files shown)


src/Integration.Vsix/SonarLintTagger/IssueToIssueMarkerConverter.cs
```diff
@@ -62,8 +62,14 @@ public IssueMarker Convert(IAnalysisIssue issue, ITextSnapshot textSnapshot)
 
             var span = issueSpanCalculator.CalculateSpan(issue, textSnapshot);
 
+            if (span.IsEmpty)
```
rita-g-sonarsource, line 65: issueSpanCalculator no longer returns null. It will return an empty span.

src/Integration.Vsix.UnitTests/SonarLintTagger/IssueToIssueMarkerConverterTests.cs
```diff
@@ -57,13 +57,13 @@ public void Convert_NullTextSnapshot_Throws()
         }
 
         [TestMethod]
-        public void Convert_NullSpan_Null()
+        public void Convert_EmptySpan_Null()
```
rita-g-sonarsource, line 60: issueSpanCalculator no longer returns null. It will return an empty span.

src/Integration.Vsix.UnitTests/SonarLintTagger/IssueToIssueMarkerConverterTests.cs
```diff
@@ -75,15 +75,16 @@ public void Convert_NullSpan_Null()
         public void Convert_SpanIsNotNull_ReturnsMarkerWithSpan()
         {
             var mockIssue = Mock.Of<IAnalysisIssue>();
-            var mockSnapshot = Mock.Of<ITextSnapshot>();
-            var mockSpan = new SnapshotSpan();
+            var mockSnapshot = new Mock<ITextSnapshot>();
+            mockSnapshot.SetupGet(x => x.Length).Returns(20);
```
rita-g-sonarsource, line 79: Need to create now a valid span, since `new SnapshotSpan()` is an empty one and will be considered invalid.

src/IssueViz.UnitTests/Editor/LocationNavigatorTests.cs
```diff
@@ -28,6 +28,7 @@
 using SonarLint.VisualStudio.Integration;
```
rita-g-sonarsource, line 28: Please re-review all the tests. The implementation was changed, so I pretty much re-wrote them.

src/IssueViz.UnitTests/IssueVisualizationControl/IssueVisualizationViewModelTests.cs
```diff
@@ -574,21 +574,16 @@ public void SelectionService_OnLocationChanged_NewLocationIsNull_NoNavigation()
         }
 
         [TestMethod]
-        [DataRow(true)]
-        [DataRow(false)]
-        public void SelectionService_OnLocationChanged_NewLocationIsNotNull_NavigabilityUpdated(bool navigationSucceeded)
```
rita-g-sonarsource, line 579: No longer relevant, there is no property to set anymore, and the span is set in LocationNavigator

src/IssueViz.UnitTests/Models/AnalysisIssueVisualizationTests.cs
```diff
@@ -0,0 +1,137 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.Text;
+using Moq;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Models
+{
+    [TestClass]
+    public class AnalysisIssueVisualizationTests
```
rita-g-sonarsource, line 33: This is a blunt copy-paste of AnalysisIssueLocationVisualizationTests, so I think your original suggestion of creating a PrimaryLocation property on AnalysisIssueVisualization was better than implementing the interface.
That, or maybe we should have a separate interface IAnalysisIssueSecondaryLocationVisualization

src/IssueViz.UnitTests/Selection/IssueFlowStepNavigatorTests.cs
```diff
@@ -247,50 +247,6 @@ public void GotoPreviousNavigableFlowStep_NavigatesToPreviousNavigableLocation()
             VerifyNavigation(expectedNavigation);
         }
 
-        [TestMethod]
```
rita-g-sonarsource, line 250: No longer relevant, there is no property to set anymore, and the span is set in LocationNavigator

src/IssueViz.UnitTests/Selection/IssueFlowStepNavigatorTests.cs
```diff
@@ -314,10 +270,17 @@ private IAnalysisIssueLocationVisualization CreateLocation(int stepNumber = 1, b
             var locationViz = new Mock<IAnalysisIssueLocationVisualization>();
             locationViz.Setup(x => x.Location).Returns(Mock.Of<IAnalysisIssueLocation>());
             locationViz.Setup(x => x.StepNumber).Returns(stepNumber);
-            locationViz.SetupProperty(x => x.IsNavigable);
-            locationViz.Object.IsNavigable = isMarkedAsNavigable;
 
-            locationNavigatorMock.Setup(x => x.TryNavigate(locationViz.Object.Location)).Returns(isTrulyNavigable);
+            if (isMarkedAsNavigable)
```
rita-g-sonarsource, line 274: This name is no longer correct but I can't think of a better one. `isPotentiallyNavigable`?

src/IssueViz/Editor/LocationNavigator.cs
```diff
@@ -47,33 +49,43 @@ internal LocationNavigator(IDocumentNavigator documentNavigator, IIssueSpanCalcu
             this.logger = logger;
         }
 
-        public bool TryNavigate(IAnalysisIssueLocation location)
+        public bool TryNavigate(IAnalysisIssueLocationVisualization locationVisualization)
```
rita-g-sonarsource, line 52: It now accepts LocationViz for 2 reasons:
1. To use `locationVisualization.CurrentFilePath` in-case of renames.
2. To update `locationVisualization.Span` with the calculated value.

src/IssueViz/Editor/LocationNavigator.cs
```diff
@@ -47,33 +49,43 @@ internal LocationNavigator(IDocumentNavigator documentNavigator, IIssueSpanCalcu
...
             }
 
             return false;
         }
+
+        private SnapshotSpan GetLocationSpan(IAnalysisIssueLocationVisualization locationVisualization, ITextView textView)
+        {
+            var shouldCalculateSpan = !locationVisualization.Span.HasValue || locationVisualization.Span.Value.IsEmpty;
+
+            if (shouldCalculateSpan)
```
rita-g-sonarsource, line 83: I'm debating about `shouldCalculateSpan`.
If the user navigates to the location once (thus setting Span to a valid value), and then modifies the code without triggering the tagger (e.g. outside of VS) -- we will never reset the wrong span and will keep trying navigating to it.

src/IssueViz/Editor/LocationNavigator.cs
```diff
@@ -47,33 +49,43 @@ internal LocationNavigator(IDocumentNavigator documentNavigator, IIssueSpanCalcu
...
-                    documentNavigator.Navigate(textView, locationSpan.Value);
+                    documentNavigator.Navigate(textView, locationSpan);
 
                     return true;
                 }
             }
             catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
             {
-                logger.WriteLine(Resources.ERR_OpenDocumentException, locationFilePath, ex.Message);
+                logger.WriteLine(Resources.ERR_OpenDocumentException, locationVisualization.CurrentFilePath, ex.Message);
```
rita-g-sonarsource, line 73: Perhaps I should reset the LocationViz.Span to null if there was an exception, WDYT? See my comment below https://github.com/SonarSource/sonarlint-visualstudio/pull/1675/files#r482050233
So we won't get stuck with an irrelevant span.

src/IssueViz/IssueVisualizationControl/AggregatorValueConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Windows.Data;
+
+namespace SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl
+{
+    public class AggregatorValueConverter : List<IValueConverter>, IValueConverter
```
rita-g-sonarsource, line 28: Used for the warning icon in the xaml. 
I need a double conversion -> From span to IsNavigable, and from IsNavigable to Visibility.

src/IssueViz/IssueVisualizationControl/SpanToNavigabilityConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl
+{
+    [ValueConversion(typeof(SnapshotSpan?), typeof(bool))]
+    public class SpanToNavigabilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            var span = (SnapshotSpan?)value;
+
+            return span.IsNavigable();
```
rita-g-sonarsource, line 36: Extension method

src/Core/WPF/AggregatorValueConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Windows.Data;
+
+namespace SonarLint.VisualStudio.Core.WPF
+{
+    public class AggregatorValueConverter : List<IValueConverter>, IValueConverter
```
rita-g-sonarsource, line 28: Used for the warning icon in the xaml.
I need a double conversion -> From span to IsNavigable, and from IsNavigable to Visibility.

src/IssueViz/Editor/IIssueSpanCalculator.cs
```diff
@@ -32,13 +32,15 @@ public interface IIssueSpanCalculator
...
         /// </summary>
-        SnapshotSpan? CalculateSpan(IAnalysisIssueLocation location, ITextSnapshot currentSnapshot);
+        SnapshotSpan CalculateSpan(IAnalysisIssueLocation location, ITextSnapshot currentSnapshot);
     }
 
     [Export(typeof(IIssueSpanCalculator))]
     [PartCreationPolicy(CreationPolicy.Shared)]
     public class IssueSpanCalculator : IIssueSpanCalculator
     {
+        private static SnapshotSpan EmptySpan { get; } = new SnapshotSpan();
```
rita-g-sonarsource, line 42: Just for readability

src/IssueViz/Editor/LocationNavigator.cs
duncanp-sonar: What's the alternative? Always calculating the span? That would be the wrong thing to do in any case where the user has edited the document inside VS so the tagger will have updated `locationVisualization.Span`.

src/IssueViz/Editor/LocationNavigator.cs
```diff
@@ -47,33 +49,43 @@ internal LocationNavigator(IDocumentNavigator documentNavigator, IIssueSpanCalcu
...
             }
 
             return false;
         }
+
+        private SnapshotSpan GetLocationSpan(IAnalysisIssueLocationVisualization locationVisualization, ITextView textView)
```
duncanp-sonar, line 79: Weird naming - I wouldn't expect a `GetX` method to set X in some circumstances. I'm not sure what I'd call it, though - `FetchOrCalculateSpan` ?

src/IssueViz/Editor/LocationNavigator.cs
duncanp-sonar: Making the location non-navigable seems reasonable.
`null` spans are treated as navigable, aren't they?

src/IssueViz/Editor/LocationNavigator.cs
```diff
@@ -47,33 +49,43 @@ internal LocationNavigator(IDocumentNavigator documentNavigator, IIssueSpanCalcu
...
+        {
+            var shouldCalculateSpan = !locationVisualization.Span.HasValue || locationVisualization.Span.Value.IsEmpty;
```
duncanp-sonar, line 81: An empty span means we've previously been unable to calculate the span correctly e.g. because the file has been edited. Why would we try to recalculate the span from the original location in this case?

src/IssueViz.UnitTests/Editor/LocationNavigatorTests.cs
```diff
@@ -68,116 +76,163 @@ public void MefCtor_CheckIsExported()
...
 
             documentOpenerMock.VerifyNoOtherCalls();
+            spanCalculatorMock.VerifyNoOtherCalls();
         }
 
         [TestMethod]
-        [DataRow(null)]
-        [DataRow("")]
-        public void TryNavigate_LocationHasNoFilePath_DocumentNotOpened(string filePath)
+        public void TryNavigate_FailedToOpenDocument_NoException()
```
duncanp-sonar, line 90: For me, it isn't clear from the name `FailedToX` that the failure was caused by an exception. `ErrorOpeningDocument_` is less ambiguous.

src/IssueViz.UnitTests/Models/AnalysisIssueVisualizationTests.cs
duncanp-sonar: Not sure; this is test code, though, so it probably doesn't check.

src/IssueViz.UnitTests/Selection/IssueFlowStepNavigatorTests.cs
duncanp-sonar: I think either is ok.

duncanp-sonar (commented):
Some questions inline about `LocationNavigator.cs`

src/IssueViz.UnitTests/Models/AnalysisIssueVisualizationTests.cs
rita-g-sonarsource: Ah, again with this limitation :/

src/IssueViz/Editor/LocationNavigator.cs
rita-g-sonarsource: I thought about setting it to null so we would attempt to recalculate it next time. But I guess it depends on the reason we failed in the first place.

src/IssueViz/Editor/LocationNavigator.cs
rita-g-sonarsource: You're right; I'll change it.

duncanp-sonar (approved):

src/IssueViz/Editor/LocationNavigator.cs
rita-g-sonarsource: I extracted it outside the method, so now it's
```
var locationSpan = GetLocationSpan(locationVisualization, textView);
locationVisualization.Span = locationSpan;

if (!locationSpan.IsEmpty)
...
```
But I guess it's the side-effect in itself which is problematic -- the fact that TryNavigate is really TryNavigateAndUpdateSpan

rita-g-sonarsource: `RecalculateLocationSpanIfNeeded` ? 🙈

PR #1676, opened 2020-09-02
Added location tagger

duncanp-sonar:


src/IssueViz.UnitTests/Editor/LocationTagging/LocationTaggerTests.cs
```diff
@@ -0,0 +1,431 @@
...
+
+            var testSubject = new LocationTagger(ValidBuffer, storeMock, ValidSpanCalculator, ValidLogger);
+            testSubject.TagSpans.Count().Should().Be(3); // sanity check
+
+            var matchingTags = testSubject.GetTags(inputSpans);
+            matchingTags.Select(x => x.Span.Span).Should().BeEquivalentTo(locSpan2, locSpan3);
+        }
+
+        [TestMethod]
+        public void GetTags_DifferentSnapshots_SnapshotsAreTranslated()
```
duncanp-sonar, line 209: I worked out how to mock enough of the snapshot infrastructure to be able to write tests for "TranslateTo".
Essentially, each snapshot has an `ITextVersion` object which points to the `Next` `version object, and contains a list of changes between the version. I created versions with an empty changes collection which seemed to do the trick.

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -0,0 +1,188 @@
...
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging
+{
+    // TODO: track document renames: https://github.com/SonarSource/sonarlint-visualstudio/issues/1662
+
+    // Buffer tagger for Sonar issue locations
+    // * adds location tags for primary and secondary locations for a file
+    // * tracks buffer edits, re-calculate spans, and notifies the issues location store about the changes
+    internal class LocationTagger : ITagger<IIssueLocationTag>, IDisposable
```
rita-g-sonarsource, line 37: SC complains about the dispose pattern; make the class sealed?

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -0,0 +1,188 @@
...
+                {
+                    locViz.Span = spanCalculator.CalculateSpan(locViz.Location, currentSnapshot);
+                }
+            }
+        }
+
+        private static ITagSpan<IIssueLocationTag> CreateTagSpan(IAnalysisIssueLocationVisualization locViz) =>
+            new TagSpan<IIssueLocationTag>(locViz.Span.Value, new IssueLocationTag(locViz));
+
+        private string GetFileName(ITextBuffer textBuffer)
```
rita-g-sonarsource, line 101: How do you think you will handle file renames?
We used to subscribe to `Document.FileActionOccurred`, so it might be easier to just receive ITextDocument in the ctor instead of ITextBuffer, and then we can use it to get `Document.FilePath`.

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        public LocationTagger(ITextBuffer buffer, IIssueLocationStore locationService, IIssueSpanCalculator spanCalculator, ILogger logger)
+        {
+            this.buffer = buffer;
+            this.locationService = locationService;
+            this.spanCalculator = spanCalculator;
+            this.logger = logger;
+
+            FilePath = GetFileName(buffer);
+            UpdateTags();
```
rita-g-sonarsource, line 54: Why do we need to call UpdateTags in the ctor?

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -0,0 +1,188 @@
...
+            TagSpans = locations.Select(CreateTagSpan).ToList();
+
+            // Notify the editor that our set of tags has changed
+            var wholeSpan = new SnapshotSpan(textSnapshot, 0, textSnapshot.Length);
+            TagsChanged?.Invoke(this, new SnapshotSpanEventArgs(wholeSpan));
+        }
+
+        private void EnsureSpansExist(IEnumerable<IAnalysisIssueLocationVisualization> locVizs, ITextSnapshot currentSnapshot)
+        {
+            // All spans for issues in the primary file should have been calculated already (whether primary or secondary).
```
rita-g-sonarsource, line 87: Instead of " in the primary file" - maybe "in the current file"?

Also, what would've calculate it? The converter? (perhaps add a `<cref>` to it)

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -0,0 +1,188 @@
...
+            // However, there could be secondary locations relating to issues in other files that have not been calculated.
```
rita-g-sonarsource, line 88: Something like "However, there could be secondary locations in this file that relate to a primary location in another file, and that file has not been opened yet" ?

rita-g-sonarsource, line 88: "However, there could be secondary locations in this file that relate to a primary location in another file, so their spans have not been calculated yet" ?

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -0,0 +1,188 @@
...
+            var textSnapshot = buffer.CurrentSnapshot;
+
+            // Get the new locations and calculate the spans if necessary
+            var locations = locationService.GetLocations(FilePath);
+            EnsureSpansExist(locations, textSnapshot);
+
+            TagSpans = locations.Select(CreateTagSpan).ToList();
+
+            // Notify the editor that our set of tags has changed
+            var wholeSpan = new SnapshotSpan(textSnapshot, 0, textSnapshot.Length);
```
rita-g-sonarsource, line 81: Maybe extract to a function and reuse in SafeOnBufferChange

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -0,0 +1,188 @@
...
+
+                TranslateTagSpans(newTextSnapshot);
+                locationService.Refresh(new string[] { FilePath });
+
+                var wholeSpan = new SnapshotSpan(newTextSnapshot, 0, newTextSnapshot.Length);
+                TagsChanged?.Invoke(this, new SnapshotSpanEventArgs(wholeSpan));
+            }
+            catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+            {
+                logger.WriteLine(Resources.ERR_HandlingBufferChange, ex.Message);
```
rita-g-sonarsource, line 176: If there was an exception, what does this mean regarding the existing `TagSpans` property, and the spans stored in each locationViz? Is it worth clearing those?

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTaggerTests.cs
```diff
@@ -0,0 +1,431 @@
...
+            var storeMock = new Mock<IIssueLocationStore>();
+            var testSubject = new LocationTagger(ValidBuffer, storeMock.Object, ValidSpanCalculator, ValidLogger);
+            var initialTags = testSubject.TagSpans;
+
+            SnapshotSpanEventArgs actualTagsChangedArgs = null;
+            testSubject.TagsChanged += (senders, args) => actualTagsChangedArgs = args;
+
+            RaiseIssuesChangedEvent(storeMock, "not the file being tracked.txt");
+
+            actualTagsChangedArgs.Should().BeNull();
```
rita-g-sonarsource, line 122: This is not the same as testing that the event was not raised, as it could've been raised with a null... why not create a mock and verify that it was not called?

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTaggerTests.cs
```diff
@@ -0,0 +1,431 @@
...
+            testSubject.TagSpans.Should().BeSameAs(initialTags);
```
rita-g-sonarsource, line 123: Without proper setup, won't `initialTags` always be null? (in which case, the test will pass even if we change the code to set it to null)

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
duncanp-sonar: There's a separate ticket for that; I'll worry about it then.

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
duncanp-sonar: It's hard to say - depends what the exception was and where it occurred.
This part of the implementation is the same as the old one from the `TextBufferIssueTracker` and is purely defensive to stop us crashing VS. We didn't do anything with the existing spans in this case previously, and we didn't have any reports of tags becoming corrupted.

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTaggerTests.cs
duncanp-sonar: No, it's an empty list.

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTaggerTests.cs
```diff
@@ -0,0 +1,431 @@
...
+            var existingSpan = CreateUniqueSpan();
+            var locVizWithSpan = CreateLocViz(existingSpan);
+            var locVizWithoutSpan = CreateLocViz(null);
+
+            var storeMock = new Mock<IIssueLocationStore>();
+            storeMock.Setup(x => x.GetLocations(ValidBufferDocName)).Returns(new[] { locVizWithSpan, locVizWithoutSpan });
+
+            var newSpan = CreateUniqueSpan();
+            var calculatorMock = new Mock<IIssueSpanCalculator>();
+            calculatorMock.Setup(x => x.CalculateSpan(It.IsAny<IAnalysisIssueLocation>(), It.IsAny<ITextSnapshot>())).Returns(newSpan);
```
rita-g-sonarsource, line 96: Setup the calculator only for locVizWithoutSpan ?

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
duncanp-sonar: I think it's ok to document a general assumption (e.g. most spans will have been set when the analysis was done") as part of explaining why we're calculating some paths here, but I don't think we should have implementation-specific information (particularly since the converter classes are going to be changed).

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        private void TranslateTagSpans(ITextSnapshot newSnapshot)
+        {
+            var translatedTagSpans = new List<ITagSpan<IIssueLocationTag>>();
+
+            foreach (var old in TagSpans)
+            {
+                var newSpan = old.Span.TranslateTo(newSnapshot, SpanTrackingMode.EdgeExclusive);
+                old.Tag.Location.Span = newSpan; // update the span stored in the location visualization
```
rita-g-sonarsource, line 139: Shouldn't we also call Store.Refresh?
Right now it's been called outside this method, in SafeOnBufferChange, but not on GetTags.
I think it the call to Refresh should be inside this method.

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTaggerTests.cs
```diff
@@ -0,0 +1,431 @@
...
+        }
+
+        private static int spanCounter;
+
+        private static SnapshotSpan CreateUniqueSpan()
+        {
+            // NB SnapshotSpans are value types, so any "new SpanshotSpan() == new SpanshotSpan()"
+            // If we want to differentiate between spans we can do so by creating
+            // them with different lengths.
+            var length = ++spanCounter;
```
rita-g-sonarsource, line 372: Cosmetics: if the length doesn't matter, we can just do Guid.NewGuid().GetHashCode() for a random int.

rita-g-sonarsource (approved):
See some questions inside the Tagger

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
duncanp-sonar: I expect that we'd get a buffer changed event in that case before the tags are fetched, so the call to `TranslateTo` in `GetTags` is probably redundant but I'm not sure.

rita-g-sonarsource: We can experiment with removing one of the calls, or we should call Refresh in both cases.

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTaggerTests.cs
duncanp-sonar: To me, that seems like a pretty non-intuitive way of generating a unique int, and `GetHashCode` isn't guaranteed to be unique.

PR #1680, opened 2020-09-04
Added filtering tagger and error tagger

duncanp-sonar:


src/IssueViz/Editor/FilteringTaggerBase.cs
```diff
@@ -0,0 +1,175 @@
...
+
+namespace SonarLint.VisualStudio.IssueVisualization.Editor
+{
+    /// <summary>
+    /// Base class for a tagger that produces tags that are based on another tag type
+    /// e.g. a tagger that produces IErrorTags based on our IIssueLocationTags
+    /// </summary>
+    /// <typeparam name="TTrackedTagType">The type of tag being tracked</typeparam>
+    /// <typeparam name="TTagType">The type of tag being produced</typeparam>
+    internal abstract class FilteringTaggerBase<TTrackedTagType, TTagType> : ITagger<TTagType>, IDisposable
```
duncanp-sonar, line 37: We will have multiple taggers inheriting from this base class:
* a buffer tagger that produces `IErrorTags` by filtering `IIssueLocationTags` (i.e. the `ErrorTagger` in this PR)
* a buffer tagger that tracks the active selected issue and produces `ISecondaryLocationTags`
* a view tagger that produces highlights for the `ISecondaryLocationTags` and
* a view tagger that produces adornments for the `ISecondaryLocationTags`

src/IssueViz/Editor/FilteringTaggerBase.cs
```diff
@@ -0,0 +1,175 @@
...
+                }
+
+                disposedValue = true;
+            }
+        }
+
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+            Dispose(disposing: true);
```
duncanp-sonar, line 169: TODO: I'll check whether we should dispose the aggregator here

src/IssueViz.UnitTests/Editor/Common/CommonTaggerProviderTestsBase.cs
```diff
@@ -0,0 +1,95 @@
...
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Tagging;
+using Moq;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.ErrorTagging
+{
+    /// <summary>
+    /// Common tests that apply to our TaggerProviders
+    /// </summary>
+    public abstract class CommonTaggerProviderTestsBase
```
duncanp-sonar, line 33: There will be enough variation between the tagger providers that it's not worth having a common base class for them. However, we do expect some common behaviours so we can have this set of common tests.

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTaggerProviderTests.cs
```diff
@@ -18,21 +18,19 @@
...
 using SonarLint.VisualStudio.Integration.UnitTests;
 using SonarLint.VisualStudio.IssueVisualization.Editor;
 using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+using SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.ErrorTagging;
 
 namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.LocationTagging
 {
     [TestClass]
-    public class LocationTaggerProviderTests
+    public class LocationTaggerProviderTests : CommonTaggerProviderTestsBase
```
duncanp-sonar, line 33: Moved the common tests to a separate class

src/IssueViz/Editor/FilteringTaggerBase.cs
rita-g-sonarsource, line 37: Maybe IssueLocationTaggerBase? Or just TaggerBase?

rita-g-sonarsource: I don't think it has to be inheritance. If you take out `GetTags` method, there isn't a lot left to to inherit. 
Why not extract `GetTags` to a service and use it in each tagger as needed? It can accept the relevant methods as a callback (CreateTagSpan, Filter).

src/IssueViz.UnitTests/Editor/Common/TaggerTestHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Tagging;
+using Moq;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.Common
+{
+    internal static class TaggerTestHelper
+    {
+        public static readonly ITextBuffer ValidBuffer = CreateBufferWithSnapshot();
```
rita-g-sonarsource, line 30: So helpful! I got tired of copy-pasting that code everywhere.

src/IssueViz.UnitTests/Editor/ErrorTagging/ErrorTaggerTests.cs
```diff
@@ -0,0 +1,105 @@
...
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using static SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.Common.TaggerTestHelper;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.ErrorTagging
+{
+    [TestClass]
+    public class ErrorTaggerTests
+    {
+        [TestMethod]
+        public void GetTags_FilterIsApplied_ExpectedTagsCreated()
```
rita-g-sonarsource, line 39: See my other comment regarding GetTags -- I think it should be a service.

src/IssueViz.UnitTests/Editor/FilteringTaggerBaseTests.cs
```diff
@@ -0,0 +1,298 @@
...
+        [TestMethod]
+        public void OnAggregatorTagsChanged_NotifiesEditorOfChange()
+        {
+            var aggregatorMock = new Mock<ITagAggregator<TrackedTag>>();
+
+            var testSubject = new TestableFilteringTagger(aggregatorMock.Object, ValidBuffer);
+
+            SnapshotSpanEventArgs suppliedArgs = null;
+            int eventCount = 0;
+            Span expectedSpan = new Span(0, ValidBuffer.CurrentSnapshot.Length);
```
rita-g-sonarsource, line 87: expectedSpan is not used

src/IssueViz.UnitTests/Editor/FilteringTaggerBaseTests.cs
```diff
@@ -0,0 +1,298 @@
...
+            {
+                SuppliedTrackedTag = trackedTag;
+                SuppliedSpans = spans;
+            }
+
+            public TrackedTag SuppliedTrackedTag { get; }
+            public NormalizedSnapshotSpanCollection SuppliedSpans { get; }
+        }
+
+        private class TestableFilteringTagger : FilteringTaggerBase<TrackedTag, ProducedTag>
```
rita-g-sonarsource, line 251: This class exists so it would be easier to test the filtering functionality of GetTags. If GetTags were a service that accepts Filter callback, it would've been much easier to test and maintain.

src/IssueViz/Editor/ErrorTagging/ErrorTagger.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+        }
+
+        protected override TagSpan<IErrorTag> CreateTagSpan(IIssueLocationTag trackedTag, NormalizedSnapshotSpanCollection spans)
+        {
+            return new TagSpan<IErrorTag>(trackedTag.Location.Span.Value, new ErrorTag(PredefinedErrorTypeNames.Warning, trackedTag.Location.Location.Message));
+        }
+
+        protected override IEnumerable<IMappingTagSpan<IIssueLocationTag>> Filter(IEnumerable<IMappingTagSpan<IIssueLocationTag>> trackedTagSpans) =>
+            trackedTagSpans.Where(x => IsValidPrimaryLocation(x.Tag.Location));
```
rita-g-sonarsource, line 44: Is this where we would also put the suppressions?

rita-g-sonarsource (commented):
@duncanp-sonar please see my comments. Other than that, looks good.

src/IssueViz/Editor/ErrorTagging/ErrorTagger.cs
duncanp-sonar: The suppression handling hasn't changed; it's still done as part of the analysis post-processing, so suppressed issues have already been removed by this point.

src/IssueViz/Editor/FilteringTaggerBase.cs
duncanp-sonar: It does need to be disposed.

src/IssueViz.UnitTests/Editor/FilteringTaggerBaseTests.cs
duncanp-sonar: This class exists so all of the functionality of the abstract class can be tested i.e. event registration and cleanup, event handling, editor notification, filtering and new tag creation,

src/IssueViz/Editor/FilteringTaggerBase.cs
duncanp-sonar: It's a tricky one. I try to avoid inheritance, but it's useful here since there is a lot of common behaviour that would be more painful to due using composition i.e. event registration, handling, propagation and cleanup.
For me, the problem is how to isolate the variable bits of the sub-classes to make them easier to test, which means the new tag creation and the filtering, rather than `GetTags` itself. I considered a couple of alternatives, including separating out `GetTags` from the base class, but I didn't think that really helped as it didn't make the tag creation or filtering methods easier to test. I also considered a separate interface with `Filter` and `CreateTagSpan` methods, similar to your suggestion about using separate callbacks.
I'll look at it when adding the tagger classes to see if it makes things easier.

PR #1683, opened 2020-09-07
Added secondary location tagger

duncanp-sonar:


src/IssueViz/Editor/SelectedIssueTagging/Buffer/SelectedIssueLocationTagger.cs
```diff
@@ -0,0 +1,86 @@
...
+        private readonly IAnalysisIssueSelectionService issueSelectionService;
+
+        private bool disposedValue;
+
+        public SelectedIssueLocationTagger(ITagAggregator<IIssueLocationTag> tagAggregator, ITextBuffer textBuffer, IAnalysisIssueSelectionService issueSelectionService)
+            : base(tagAggregator, textBuffer)
+        {
+            this.issueSelectionService = issueSelectionService;
+
+            // Changing any of the selected issue/flow/location will always result in the
```
rita-g-sonarsource, line 43: Obsolete comment

src/IssueViz/Editor/SelectedIssueTagging/Buffer/SelectedIssueLocationTagger.cs
```diff
@@ -0,0 +1,86 @@
...
+            return new TagSpan<ISelectedIssueLocationTag>(trackedTag.Location.Span.Value, new SelectedIssueLocationTag(trackedTag.Location));
+        }
+
+        protected override IEnumerable<IMappingTagSpan<IIssueLocationTag>> Filter(IEnumerable<IMappingTagSpan<IIssueLocationTag>> trackedTagSpans)
+        {
+            var selectedSecondaryLocVizs = issueSelectionService.SelectedFlow?.Locations;
+            if (selectedSecondaryLocVizs == null || selectedSecondaryLocVizs.Count == 0) { return Array.Empty<IMappingTagSpan<IIssueLocationTag>>(); }
+
+            // Filter to any selected locations.
+            // There could be multiple secondary issues locations in this file for different issues.
```
rita-g-sonarsource, line 69: Misc: the last comment sounds misleading, since it sounds like "filter the locations and _show_ secondary locations for different issues" -- might be worth just switching them, e.g.
"There could be multiple secondary issues locations in this file for different issues so we display only the secondary locations of currently selected flow".
(or maybe remove the comment, since that's what line 65 means)

src/IssueViz/Editor/SelectedIssueTagging/Buffer/SelectedIssueLocationTagger.cs
```diff
@@ -0,0 +1,86 @@
...
+            // Changing any of the selected issue/flow/location will always result in the
+            // "SelectedLocationChanged" event being raised
+            issueSelectionService.SelectionChanged += OnSelectionChanged;
+        }
+
+        private void OnSelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            if (e.SelectionChangeLevel == SelectionChangeLevel.Flow || e.SelectionChangeLevel == SelectionChangeLevel.Issue)
+            {
+                UpdateWholeSpan();
```
rita-g-sonarsource, line 52: I think this should be renamed to NotifyTagsChanged or something of the sort. 
It occurred to me when I was looking at the tests, where we raise an event with `null` issue/flow, and I was confused for a moment how did that not fail, since we check in Filter that it's not null. And I had to re-remind myself when looking at the code that  UpdateWholeSpan just raises an event, and it doesn't execute the GetTags implementation. Hence why I think it should be more explicit that it's just an event raise.

src/IssueViz/Editor/SelectedIssueTagging/Buffer/SelectedIssueLocationTagger.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        #region ITagger methods
+
+        protected override TagSpan<ISelectedIssueLocationTag> CreateTagSpan(IIssueLocationTag trackedTag, NormalizedSnapshotSpanCollection spans)
+        {
+            return new TagSpan<ISelectedIssueLocationTag>(trackedTag.Location.Span.Value, new SelectedIssueLocationTag(trackedTag.Location));
+        }
+
+        protected override IEnumerable<IMappingTagSpan<IIssueLocationTag>> Filter(IEnumerable<IMappingTagSpan<IIssueLocationTag>> trackedTagSpans)
```
rita-g-sonarsource, line 63: I think the Filter is misused here. When we receive SelectionChanged event, we already know if it's null or not. And if it's null, there is no reason to notify the editor to begin with.

rita-g-sonarsource (approved):

src/IssueViz/Editor/SelectedIssueTagging/Buffer/SelectedIssueLocationTagger.cs
duncanp-sonar: I'll remove the comment.

PR #1689, opened 2020-09-08
Create TaggerProviders only for supported file types

### LINKED ISSUE #1685: Performance: TaggerProviders should ignore unsupported file types
### Description

The legacy `TaggerProvider` only creates tagger instances for supported document types (i.e. it checks whether the doc type is analysable). The new tagger providers should make make the same check, otherwise we'll be creating multiple unnecessary taggers for every file that is opened (e.g. for C# files).

The `TaggerProvider` implementation requires a couple of different imports to handle the check. It might be worth encapsulating this is a separate `supported document` service.

rita-g-sonarsource:
Fixes #1685


rita-g-sonarsource:
@duncanp-sonar feel free to look at the commits separately:
1. I moved SonarLanguageRecognizer to IssueViz so I could use it.
2. Just a merge commit from master.
3. I added TaggableBufferIndicator -- which I thought about making TaggableViewIndicator, since if the view taggers won't be created, then neither will the buffer taggers. However, ErrorTaggerProvider is not a view (I'm guessing it's being called by VS because it exports IErrorTag?). Do you think it will be worth adding the check to all the Buffer tagger providers (e.g. LocationTaggerProvider, SelectedIssueLocationTaggerProvider) ? It will be redundant if there are no view taggers, but it might help with readability of why ErrorTaggerProvider has the check and the other Buffer tagger providers don't.

src/IssueViz.UnitTests/Editor/Common/TaggerTestHelper.cs
```diff
@@ -47,6 +47,13 @@ public static Mock<ITextBuffer> CreateBufferMock(int length = 999)
             snapshotMock.Setup(x => x.TextBuffer).Returns(bufferMock.Object);
             snapshotMock.Setup(x => x.Length).Returns(length);
 
+            var properties = new Microsoft.VisualStudio.Utilities.PropertyCollection();
+            bufferMock.Setup(x => x.Properties).Returns(properties);
+
+            var docMock = new Mock<ITextDocument>();
```
rita-g-sonarsource, line 53: I use it in TaggableBufferIndicatorTests

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -50,7 +50,7 @@ public LocationTagger(ITextBuffer buffer, IIssueLocationStore locationService, I
             this.spanCalculator = spanCalculator;
             this.logger = logger;
 
-            FilePath = GetFileName(buffer);
+            FilePath = buffer.GetFilePath();
```
rita-g-sonarsource, line 53: I need this also in TaggableBuffferIndicator

src/IssueViz/Editor/LanguageDetection/SonarLanguageRecognizer.cs
```diff
@@ -26,32 +26,26 @@
...
+namespace SonarLint.VisualStudio.IssueVisualization.Editor.LanguageDetection
 {
+    public interface ISonarLanguageRecognizer
+    {
+        IEnumerable<AnalysisLanguage> Detect(string filePath, IContentType bufferContentType);
+    }
+
     [Export(typeof(ISonarLanguageRecognizer))]
-    internal class SonarLanguageRecognizer : ISonarLanguageRecognizer
+    public class SonarLanguageRecognizer : ISonarLanguageRecognizer
```
duncanp-sonar, line 37: Does this class need to be public? MEF will create internal classes, and the any imports will be using the public interface.

src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Highlighting/IssueHighlightViewTaggerProviderTests.cs
```diff
@@ -48,7 +63,22 @@ protected override IViewTaggerProvider CreateTestSubject()
...
+        }
+
+        [TestMethod]
+        public void CreateTagger_BufferIsNotTaggable_Null()
+        {
+            var textBuffer = TaggerTestHelper.ValidBuffer;
+            var textView = CreateValidTextView(textBuffer);
+
+            taggableBufferIndicator.Reset();
+            taggableBufferIndicator.Setup(x => x.IsTaggable(textBuffer)).Returns(false);
```
duncanp-sonar, line 76: This is not a good use of `TestInitialize`. The `taggableBufferIndicator` mock is only used in two tests, and in this one you are completely resetting it so the setup carried out in `TestInitialize` isn't used at all.

It would be better to have a `CreateTaggableIndicatorMock(bool isTaggable)` method that created and returned an appropriate object that each test could call.

src/IssueViz.UnitTests/Editor/Common/CommonTaggerProviderTestsBase.cs
```diff
@@ -46,7 +45,7 @@ public void CreateTagger_BufferIsNull_Throws()
         [TestMethod]
         public void CreateTagger_BufferIsNotNull_ReturnsSingletonTagger()
         {
-            var buffer = CreateValidTextBuffer();
+            var buffer = TaggerTestHelper.CreateBuffer();
```
duncanp-sonar, line 48: Is there a reason not have a `using static` like the other tests that use the `TaggerTestHelper`?

src/IssueViz.UnitTests/Editor/ErrorTagging/ErrorTaggerProviderTests.cs
```diff
@@ -45,7 +63,21 @@ protected override ITaggerProvider CreateTestSubject()
             aggregatorMock.Setup(x => x.CreateTagAggregator<IIssueLocationTag>(It.IsAny<ITextBuffer>()))
                 .Returns(Mock.Of<ITagAggregator<IIssueLocationTag>>());
 
-            return new ErrorTaggerProvider(aggregatorMock.Object);
+            return new ErrorTaggerProvider(aggregatorMock.Object, taggableBufferIndicator.Object);
+        }
+
+        [TestMethod]
+        public void CreateTagger_BufferIsNotTaggable_Null()
```
duncanp-sonar, line 70: Could this test be added to the common test base rather than being repeated?

src/IssueViz.UnitTests/Editor/Common/CommonViewTaggerProviderTestsBase.cs
```diff
@@ -128,7 +128,7 @@ public void CreateTagger_DifferentBuffers_ReturnsTaggerPerViewBuffer()
             tagger1.Should().NotBeSameAs(tagger2);
         }
 
-        private static ITextView CreateValidTextView(ITextBuffer buffer = null)
+        protected static ITextView CreateValidTextView(ITextBuffer buffer = null)
```
duncanp-sonar, line 131: If the "IsTaggable" test can be added to this class then this can remain private.

duncanp-sonar (approved):
Looks ok: some questions about the tests, and the `TestInitialize`s should be fixed or removed.

duncanp-sonar:
> Do you think it will be worth adding the check to all the Buffer tagger providers (e.g. LocationTaggerProvider, SelectedIssueLocationTaggerProvider) ? It will be redundant if there are no view taggers, but it might help with standardization of why ErrorTaggerProvider has the check and the other Buffer tagger providers don't.

Hhmm, tricky. I was going to say I'd just document what's happening in `LocationTaggerProvider` and `SelectedIssueLocationTaggerProvider`, but on second thoughts it might be better to include the check in case we change the code incorrectly in one of the other tagger providers in the future.
I'd add the check, and also an assertion in the case that the provider is called for an untaggable document to highlight that one of the downstream tagger providers isn't behaving correctly.

src/IssueViz.UnitTests/Editor/Common/CommonTaggerProviderTestsBase.cs
rita-g-sonarsource: I feel dumb, I didn't know this feature exists!

PR #1694, opened 2020-09-09
[1] Update CurrentFilePath when document is renamed

### LINKED ISSUE #1662: Issue tagging: handle file renames

rita-g-sonarsource:
Fixes #1662


src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -280,11 +280,13 @@ public override int IndexOf(int currentIndex, ITableEntriesSnapshot newSnapshot)
 
         public IEnumerable<IAnalysisIssueVisualization> Issues => readonlyIssues;
 
-        public IEnumerable<string> FilesInSnapshot { get; }
+        public IEnumerable<string> FilesInSnapshot { get; private set; }
 
         public void IncrementVersion()
```
rita-g-sonarsource, line 285: My listener calls LocationStore.Refresh, which in turn calls IncrementVersion

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -280,11 +280,13 @@ public override int IndexOf(int currentIndex, ITableEntriesSnapshot newSnapshot)
...
         {
             versionNumber = GetNextVersionNumber();
+
+            FilesInSnapshot = CalculateFilesInSnapshot(filePath, issues);
```
rita-g-sonarsource, line 289: I need to refresh the cache to pickup the renames

src/IssueViz/IssueVizPackage.cs
```diff
@@ -37,12 +39,17 @@ namespace SonarLint.VisualStudio.IssueVisualization
     [ProvideToolWindowVisibility(typeof(IssueVisualizationToolWindow), Constants.UIContextGuid)]
     public sealed class IssueVizPackage : AsyncPackage
     {
+        private IFileRenamesEventHandler fileRenamesEventHandler;
```
duncanp-sonar, line 42: We won't start tracking file renames until the package is loaded, which won't happen until either a command is called or the tool window is opened.

We either need to change when the package loads, or instantiate the rename tracker in a different way. We need the tracker from the point that a file is analysed/issues pushed to the error list, so we could create it in part of the analysis infrastructure.

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -29,8 +29,6 @@
 
 namespace SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging
 {
-    // TODO: track document renames: https://github.com/SonarSource/sonarlint-visualstudio/issues/1662
```
duncanp-sonar, line 32: This tagger is storing the file path at the point it is created, and using that value when calling the location service. That works partially at the moment because the snapshot file name is not being updated currently. However, the tagger won't work properly in the following scenario:
* open a file -> analysed ok, tagger created
* rename the file while it is open -> snapshot is refreshed; it knows about both old and new names
* edit the doc -> the tagger will still update correctly at this point. It's using the old file name, but it works because the snapshot still has the old name in its list of affected files
* save the file -> re-analysis -> new snapshot -> tagger told about the new issues.
At this point, the tagger only knows about the old name, and the new snapshot only knows about the new name.

The tagger is no longer tracking the latest issues, and any notifications it sends to the location store are ignored because they are using the old file name.

duncanp-sonar: FYI I pushed another commit in a branch called `RG/1662-file-renames_DP1` that changes the tagger so it doesn't ever store the file path (it fetches it from the document each time). However, it's only part of the solution, and I'm not sure it's the right approach.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -63,6 +65,19 @@ internal interface IIssuesSnapshot : ITableEntriesSnapshot
...
+        /// issues are the same but the source file has been renamed
+        /// </summary>
+        IIssuesSnapshot CreateUpdatedSnapshot(string newFilePath);
+
+        /// <summary>
+        /// Create and return an updated version of an existing snapshot, where the
+        /// set of issues are the same but their locations have changed (by the user editing the file)
+        /// </summary>
+        /// <remarks>The number and ordering of issues must be the same</remarks>
+        IIssuesSnapshot CreateUpdatedSnapshot(IEnumerable<IAnalysisIssueVisualization> updatedIssues);
```
rita-g-sonarsource, line 80: Do we use this somewhere? I think we used to, but there are no usages of this anymore.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -278,13 +283,17 @@ public override int IndexOf(int currentIndex, ITableEntriesSnapshot newSnapshot)
...
         public IEnumerable<IAnalysisIssueVisualization> Issues => readonlyIssues;
 
-        public IEnumerable<string> FilesInSnapshot { get; }
+        public IEnumerable<string> FilesInSnapshot { get; private set; }
 
         public void IncrementVersion()
         {
             versionNumber = GetNextVersionNumber();
+
+            FilesInSnapshot = CalculateFilesInSnapshot(AnalyzedFilePath, issues);
```
rita-g-sonarsource, line 296: I'm not sure about this. It feels a bit out of place, but then again, "IncrementVersion" means that _something_ has changed, so it can be appropriate to refresh the cache. I wouldn't want to expose a Rename method as that would give away that there is a cache.

rita-g-sonarsource:

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -223,5 +226,61 @@ private static void SafeOperation(ITableDataSink sink, string operationName, Act
...
+                        if (factoryChanged)
+                        {
+                            InternalRefreshErrorList(factory);
+                        }
+                    }
+                }
+            }
+        }
+
+        private static bool HandleFileRenames(SnapshotFactory factory, string oldFilePath, string newFilePath)
```
rita-g-sonarsource, line 253: Maybe this should be part of SnapshotFactory, WDYT?

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
duncanp-sonar: No. There are three static constructor methods for the snapshot that were all used by the TBIT. Only one of them should be used now.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
duncanp-sonar: It would be more efficient to do this as part of the file renaming code since that's the only time we'd need to recalculate the set of files. Otherwise, we'll end up recalculating unnecessarily every time the text buffer is changed.

duncanp-sonar (approved):

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
rita-g-sonarsource: So are you suggesting to expose a Rename method on IssuesSnapshot?

duncanp-sonar: I'd leave it as it is for now since it works.
There's something that doesn't feel right, even after your second PR.  If the analyzed file has changed we create a new snapshot but preserve the existing issue instances; if the analyzed file has not changed we just call `IncrementVersion`.
There doesn't seem to be a good reason to always update the issue location file paths in-place (by updating the issue visualizations), but to "update" the analyzed file path by creating a new snapshot instance.

rita-g-sonarsource: Fair point. Maybe we should just always create a new snapshot then. Renames won't happen very often, so it's not gonna be a performance hit (and it would simplify the code). WDYT?

duncanp-sonar: Yep, why not. Then `IncrementVersion` won't need to change since the constructor will calculate the affected files anyway.

PR #1697, opened 2020-09-09
Calculate affected span

### LINKED ISSUE #1686: Performance: Taggers should report affected span correctly
### Description

The removed legacy tagger implementation calculated the affected span before raising the `TagsChanged` event (the affected span being the smallest span that enclosed the previous set of tags and the tags - see `TextBufferIssueTracker.CalculateAffectedSpan`).

Our new taggers are reporting the entire span as having changed each time, which will cause the editor to do unnecessary work.

As a start, the `LocationTagger` should correctly calculate the affected span.
We should then check how the change notification is propagated to the consuming taggers to work out what they need to do, if anything.

rita-g-sonarsource:
Fixes #1686

### DIFF
--- a/src/IssueViz/Editor/LocationTagging/LocationTagger.cs
+++ b/src/IssueViz/Editor/LocationTagging/LocationTagger.cs
@@ -75,9 +75,11 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging
             var locations = locationService.GetLocations(FilePath);
             EnsureSpansExist(locations, textSnapshot);
 
+            var oldTags = TagSpans;
             TagSpans = locations.Select(CreateTagSpan).ToList();
 
-            NotifyTagsChanged(textSnapshot);
+            var affectedSpan = CalculateSpanOfAllTags(oldTags, textSnapshot);
+            NotifyTagsChanged(affectedSpan);
         }
 
         private void EnsureSpansExist(IEnumerable<IAnalysisIssueLocationVisualization> locVizs, ITextSnapshot currentSnapshot)
@@ -157,9 +159,10 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging
                 var newTextSnapshot = e.After;
 
                 TranslateTagSpans(newTextSnapshot);
-                locationService.Refresh(new string[] { FilePath });
+                locationService.Refresh(new[] { FilePath });
 
-                NotifyTagsChanged(newTextSnapshot);
+                var affectedSpan = CalculateSpanOfChanges(e.Changes, newTextSnapshot);
+                NotifyTagsChanged(affectedSpan);
             }
             catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
             {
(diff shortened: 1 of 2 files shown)


src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -157,21 +159,55 @@ private void SafeOnBufferChange(object sender, TextContentChangedEventArgs e)
...
+        /// </summary>
+        private SnapshotSpan CalculateSpanOfChanges(INormalizedTextChangeCollection changeCollection, ITextSnapshot newTextSnapshot)
+        {
+            var startSpans = changeCollection.Select(x => x.NewSpan).ToArray();
+            var endSpans = TagSpans.Select(x => x.Span.Span);
+            
+            return CalculateAffectedSpan(newTextSnapshot, startSpans, endSpans);
+        }
+
+        private static SnapshotSpan CalculateAffectedSpan(ITextSnapshot textSnapshot, IEnumerable<Span> startSpans, IEnumerable<Span> endSpans)
```
rita-g-sonarsource, line 203: In the TBIT [we used to call Span.TranslateTo](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs#L215) with PointTrackingMode.Negative / PointTrackingMode.Positive and I don't do that. I'm not sure if I need to, the editor seems to behave ok without it.

duncanp-sonar: All of our other taggers still tell the editor to refresh everything so I think the editor would still look ok even if this code isn't correct.
Let's discuss this offline. I think the translation is necessary when the set of issues changes. When the buffer changes, it might be enough to just use the changed span.

duncanp-sonar:
I'm getting a weird behaviour with secondary location tags not being updated correctly some of the time. I'm using the nested-switch code sample so there are two issues with secondary locations: (A) cognitive complexity with 8, and (B) don't nest more than 3 statements. If I select (A) then (B) then (A), then the secondary locations are not highlighted correctly. Only three are show instead of the expected 8. I've tried tweaking the code in various ways, but I can't get it to work :-(

rita-g-sonarsource:
> I'm getting a weird behaviour with secondary location tags not being updated correctly some of the time. I'm using the nested-switch code sample so there are two issues with secondary locations: (A) cognitive complexity with 8, and (B) don't nest more than 3 statements. If I select (A) then (B) then (A), then the secondary locations are not highlighted correctly. Only three are show instead of the expected 8. I've tried tweaking the code in various ways, but I can't get it to work :-(

Which branch are you working on? It works for me on this branch but I'm guessing it's because the view taggers are unchanged.

duncanp-sonar:
I'll merge this as it is because at least we're notifying smaller spans than before and the tagging still seems to work correctly.

The issue location notifications should be translating the old tag locations to the new snapshot before calculating the affected span, but in practice it doesn't seem to make much difference. The user would have to type quite quickly while issues are still being reported, and in any event the buffer edit tracking would kick in to fix up the modified span locations. There might be a particular set of issues/edits that would cause the tags to be redrawn incorrectly, but I haven't managed to find one.

I think we could reduce the span we notify for buffer tracking changes further, but it's difficult to be sure and any change requires quite a lot of testing.

Issue #1598, opened 2020-08-10
Implement issue flow adornment tagger


duncanp-sonar:
### Description

Implement an adornment tagger that displays a numbered tag for each step in the selected flow.

Look and feel:
* theming: the colourisation used should work appropriately in different themes
* font settings: the appropriate editor font settings should be used and the adornment updated as the settings change (e.g. text font and height)

TODO:
- [x]  tooltip
- [x] optimistation? the MS example has optimisations to reuse WPF items if possible. DONE.
- [ ] ~~? potential bug? the MS example refers to a workaround for a bug (delays creating the tags until the WPF layout has finished?)~~
- [ ] ~~? different highlight for selected step?~~
- [ ] ~~? clickable?~~

Relates to https://jira.sonarsource.com/browse/MMF-2098

PR #1699, opened 2020-09-10
Added adornment caching

duncanp-sonar:
Relates to #1598

### DIFF
--- a/src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTaggerTests.cs
+++ b/src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTaggerTests.cs
@@ -18,23 +18,91 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Linq;
-using System.Windows.Media;
-using System.Windows.Media.TextFormatting;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Text.Editor;
-using Microsoft.VisualStudio.Text.Formatting;
+using Microsoft.VisualStudio.Text.Tagging;
 using Moq;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Editor.SelectedIssueTagging;
 using SonarLint.VisualStudio.IssueVisualization.Editor.SelectedIssueTagging.Adornment;
+using static SonarLint.VisualStudio.IssueVisualization.Editor.SelectedIssueTagging.Adornment.IssueLocationAdornmentTagger;
 using static SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.Common.TaggerTestHelper;
 
-namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.SelectedIssueTagging.Highlighting
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.SelectedIssueTagging.Adornment
 {
     [TestClass]
     public class IssueLocationAdornmentTaggerTests
     {
+        private static readonly IWpfTextView ValidTextView = CreateWpfTextView();
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            ThreadHelper.SetCurrentThreadAsUIThread();
+        }
+
+        [TestMethod]
+        public void Ctor_SubscribesToEvents()
+        {
+            var aggregatorMock = new Mock<ITagAggregator<ISelectedIssueLocationTag>>();
+            aggregatorMock.SetupAdd(x => x.BatchedTagsChanged += (sender, args) => { });
+
+            // Act
+            var testSubject = new IssueLocationAdornmentTagger(aggregatorMock.Object, ValidTextView);
+
+            var expectedCallCount = Times.Exactly(2); // once for our class, once for the base class
+            aggregatorMock.VerifyAdd(x => x.BatchedTagsChanged += It.IsAny<EventHandler<BatchedTagsChangedEventArgs>>(), expectedCallCount);
+        }
+
+        [TestMethod]
+        public void Dispose_UnsubscribesFromEvents()
+        {
+            var aggregatorMock = new Mock<ITagAggregator<ISelectedIssueLocationTag>>();
+            aggregatorMock.SetupRemove(x => x.BatchedTagsChanged -= (sender, args) => { });
+
+            var testSubject = new IssueLocationAdornmentTagger(aggregatorMock.Object, ValidTextView);
+
+            // Act
+            testSubject.Dispose();
+
+            var expectedCallCount = Times.Exactly(2); // once for our class, once for the base class
+            aggregatorMock.VerifyRemove(x => x.BatchedTagsChanged -= It.IsAny<EventHandler<BatchedTagsChangedEventArgs>>(), expectedCallCount);
+        }
+
+        [TestMethod]
+        public void OnBatchedTagsChanged_CacheCleanupIsCalled()
+        {
+            var cacheMock = new Mock<ICachingAdornmentFactory>();
+
+            var snapshot = CreateSnapshot(length: 123);
+            var textView = CreateWpfTextView(snapshot);
+            var locViz1 = CreateLocationViz();
+            var locViz2 = CreateLocationViz();
+
+            // The setup for the aggregator is different for this test since we're using a different
+            // overload of "GetTags" from all of the other tests.
+            // Here, we're expecting a SnapshotSpan covering the whole snapshot to be passed in
+            var wholeSnapshotSpan = new SnapshotSpan(snapshot, 0, snapshot.Length);
+
+            var mappingSpan1 = CreateMappingTagSpan(snapshot, new SelectedIssueLocationTag(locViz1));
+            var mappingSpan2 = CreateMappingTagSpan(snapshot, new SelectedIssueLocationTag(locViz2));
+            var aggregatorMock = new Mock<ITagAggregator<ISelectedIssueLocationTag>>();
+            aggregatorMock.Setup(x => x.GetTags(wholeSnapshotSpan))
+                .Returns(new[] { mappingSpan1, mappingSpan2 });
+
+            var testSubject = new IssueLocationAdornmentTagger(aggregatorMock.Object, textView, cacheMock.Object);
+            cacheMock.Invocations.Count.Should().Be(0);
+
+            // Act
+            RaiseBatchedTagsChanged(aggregatorMock);
+            aggregatorMock.Verify(x => x.GetTags(wholeSnapshotSpan), Times.Once);
+            cacheMock.Verify(x => x.RemoveUnused(new[] { locViz1, locViz2 }), Times.Once);
+        }
+
         [TestMethod]
         public void GetTags_NoSelectedIssueLocationTags_ReturnsEmpty()
         {
@@ -42,7 +110,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.SelectedIss
             var inputSpans = CreateSpanCollectionSpanningWholeSnapshot(snapshot);
 
             var aggregator = CreateSelectedIssueAggregator();
-            var viewMock = CreateValidTextView(snapshot);
+            var viewMock = CreateWpfTextView(snapshot);
 
             var testSubject = new IssueLocationAdornmentTagger(aggregator, viewMock);
 
@@ -61,7 +129,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.SelectedIss
             var selectedLoc2 = CreateLocationViz(snapshot, new Span(20, 25), "selection 2");
             var aggregator = CreateSelectedIssueAggregator(selectedLoc1, selectedLoc2);
 
-            var viewMock = CreateValidTextView(snapshot);
+            var viewMock = CreateWpfTextView(snapshot);
 
             var testSubject = new IssueLocationAdornmentTagger(aggregator, viewMock);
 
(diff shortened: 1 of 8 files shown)


src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornment.cs
```diff
@@ -28,37 +28,42 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor.SelectedIssueTagging.
...
             // e.g. if the user changes the font size
-            Location = issueTag.Location;
+            LocationViz = locationViz;
 
             Margin = new System.Windows.Thickness(1, 0, 1, 0); // Space between this UI element and the editor text
             Padding = new System.Windows.Thickness(0);  // Space between the side of the control and its content    
 
             Background = Brushes.Pink;
-            BorderBrush = Brushes.Red;
+            BorderBrush = Brushes.LightCoral;
```
duncanp-sonar, line 43: Unrelated: minor visual tweak so the contrast between the border and background is less harsh.

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornment.cs
```diff
@@ -28,37 +28,42 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor.SelectedIssueTagging.
     internal class IssueLocationAdornment : Button
     {
         private readonly TextBlock textBlock;
-        public IAnalysisIssueLocationVisualization Location { get; }
+        public IAnalysisIssueLocationVisualization LocationViz { get; }
 
-        public IssueLocationAdornment(ISelectedIssueLocationTag issueTag, IFormattedLineSource formattedLineSource)
+        public IssueLocationAdornment(IAnalysisIssueLocationVisualization locationViz, IFormattedLineSource formattedLineSource)
```
duncanp-sonar, line 33: Miscellaneous: it's simpler just to pass the location visualisation - we weren't using the tag.

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornment.cs
```diff
@@ -28,37 +28,42 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor.SelectedIssueTagging.
...
+                Text = LocationViz.StepNumber.ToString(),
+                Padding = new System.Windows.Thickness(0.2, 0, 0.2, 0),
+                Margin = new System.Windows.Thickness(0),
+            };
+            Update(formattedLineSource);
+
+            ToolTip = new TextBlock
+            {
+                Text = LocationViz.Location.Message
+            };
```
duncanp-sonar, line 58: Miscellaneous: added the tooltip.

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornment.cs
```diff
@@ -28,37 +28,42 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor.SelectedIssueTagging.
...
+
             Content = textBlock;
         }
 
-        private void FormatText(IFormattedLineSource formattedLineSource)
+        public void Update(IFormattedLineSource formattedLineSource)
```
duncanp-sonar, line 63: Now called whenever an existing tag is updated.

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger.cs
```diff
@@ -40,11 +64,89 @@ protected override TagSpan<IntraTextAdornmentTag> CreateTagSpan(ISelectedIssueLo
             // should have zero length spans. Overriding this method allows control
             // over the tag spans.
             var adornmentSpan = new SnapshotSpan(trackedTag.Location.Span.Value.Start, 0);
-            var adornment = new IssueLocationAdornment(trackedTag, wpfView.FormattedLineSource);
+            var adornment = adornmentCache.CreateOrUpdate(trackedTag.Location);
 
-            // If we don't call Measure here the tag is positioned incorrectly
-            adornment.Measure(new System.Windows.Size(double.PositiveInfinity, double.PositiveInfinity));
```
duncanp-sonar, line 46: Moved to the cache/factory method.

src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTaggerTests.cs
```diff
@@ -77,31 +145,11 @@ public void GetTags_HasSelectedIssueLocationTags_ReturnsExpectedAdornmentTags()
...
             adornment1.Should().NotBeNull();
-            adornment1.Location.Should().Be(selectedLoc1);
+            adornment1.LocationViz.Should().Be(selectedLoc1);
 
             var adornment2 = actual[1].Tag.Adornment as IssueLocationAdornment;
             adornment2.Should().NotBeNull();
-            adornment2.Location.Should().Be(selectedLoc2);
-        }
-
-        private static IWpfTextView CreateValidTextView(ITextSnapshot snapshot)
```
duncanp-sonar, line 87: Moved to the helper class

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger.cs
```diff
@@ -40,11 +66,85 @@ protected override TagSpan<IntraTextAdornmentTag> CreateTagSpan(ISelectedIssueLo
...
+            if (disposing)
+            {
+                tagAggregator.BatchedTagsChanged -= OnTagsChanged;
+            }
+            base.Dispose(disposing);
+        }
+
+        #region Adornment cache
+
+        internal interface ICachingAdornmentFactory
```
rita-g-sonarsource, line 85: Maybe extract this to another file? I think it has enough in it that it deserves its own file.

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger.cs
```diff
@@ -40,11 +66,85 @@ protected override TagSpan<IntraTextAdornmentTag> CreateTagSpan(ISelectedIssueLo
...
+        {
+            /// <summary>
+            /// Clears unnecessary entries from the cache
+            /// </summary>
+            /// <param name="currentLocations">The list of locations that are still in use</param>
+            void RemoveUnused(IEnumerable<IAnalysisIssueLocationVisualization> currentLocations);
```
rita-g-sonarsource, line 91: Maybe just "Refresh" or "RefreshCache"?
The convention is that "Remove" accepts the items to remove, so it's confusing that it's actually the items to keep.

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger.cs
```diff
@@ -18,20 +18,46 @@
...
+        private readonly ITagAggregator<ISelectedIssueLocationTag> tagAggregator;
 
         public IssueLocationAdornmentTagger(ITagAggregator<ISelectedIssueLocationTag> tagAggregator, IWpfTextView textView)
-            : base(tagAggregator, textView)
+            : this(tagAggregator, textView, new CachingAdornmentFactory(textView))
         {
-            wpfView = textView;
+        }
+
+        internal /* for testing */ IssueLocationAdornmentTagger(ITagAggregator<ISelectedIssueLocationTag> tagAggregator,
```
rita-g-sonarsource, line 43: Why not have the TaggerProvider create the Factory and pass it down? It's not a problem to expose that the AdornmentTagger uses a cache.
In many places of the code we end up with the main class receiving parameters that it doesn't need, just so it could pass it down to the classes that it creates. I think it creates unnecessary complexity and results with the main class knowing too much about the inner class' dependencies.
I think it would be better to have just a public ctor and have the TaggerProvider do the orchestrating.

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger.cs
```diff
@@ -18,20 +18,46 @@
...
+            IWpfTextView wpfView, ICachingAdornmentFactory adornmentCache)
+            : base(tagAggregator, wpfView)
+        {
+            this.wpfView = wpfView;
+            this.adornmentCache = adornmentCache;
+
+            this.tagAggregator = tagAggregator;
+            tagAggregator.BatchedTagsChanged += OnTagsChanged;
```
rita-g-sonarsource, line 51: It feels wrong having to register to the event twice in the same class instance. Perhaps we can make the base `NotifyTagsChanged` a virtual, but I don't know if it's better.

src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger_CacheTests.cs
```diff
@@ -0,0 +1,177 @@
...
+            testSubject.CreateOrUpdate(CreateLocationViz());
+            testSubject.CreateOrUpdate(CreateLocationViz());
+            testSubject.CachedAdornments.Count.Should().Be(3); // sanity check
+
+            testSubject.RemoveUnused(Array.Empty<IAnalysisIssueLocationVisualization>());
+            testSubject.CachedAdornments.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void RemoveUnused_AllLocationsAreCurrent_NoAdormentsRemoved()
```
rita-g-sonarsource, line 131: Maybe
`Refresh_LocationsAreUnchanged_NoAdormentsRemoved` ?
(I think it should be Refresh)

src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger_CacheTests.cs
```diff
@@ -0,0 +1,177 @@
...
+        public void RemoveUnused_EmptyCache_NoError()
+        {
+            var testSubject = new CachingAdornmentFactory(ValidTextView);
+
+            Action act = () => testSubject.RemoveUnused(new[] { ValidLocViz });
+            act.Should().NotThrow();
+        }
+
+        [TestMethod]
+        public void RemoveUnused_NoCurrentLocations_AllRemoved()
```
rita-g-sonarsource, line 118: Maybe
`Refresh_NoNewLocations_CacheCleared` ?

src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger_CacheTests.cs
```diff
@@ -0,0 +1,177 @@
...
+            var adornment2 = testSubject.CreateOrUpdate(locViz2);
+            var expectedAdornments = new[] { adornment1, adornment2 };
+
+            // Act
+            testSubject.RemoveUnused(new[] { locViz1, locViz2 });
+            testSubject.CachedAdornments.Should().BeEquivalentTo(expectedAdornments);
+        }
+
+        [TestMethod]
+        public void RemoveUnused_NotAllLocationsAreCurrent_ExpectedEntriesRemoved()
```
rita-g-sonarsource, line 148: Maybe `Refresh_SomeLocationsChanged_ExpectedAdormentsRemoved`?

src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger_CacheTests.cs
```diff
@@ -0,0 +1,177 @@
...
+            var locViz2 = CreateLocationViz();
+            var locViz3 = CreateLocationViz();
+
+            var adornment1 = testSubject.CreateOrUpdate(locViz1);
+            var adornment2 = testSubject.CreateOrUpdate(locViz2);
+            var adornment3 = testSubject.CreateOrUpdate(locViz3);
+            testSubject.CachedAdornments.Should().BeEquivalentTo(new[] { adornment1, adornment2, adornment3 }); // sanity check
+
+            // Remove multiple items
+            testSubject.RemoveUnused(new[] { locViz3 });
```
rita-g-sonarsource, line 162: Btw, the usage really implies that the unused is `locViz3`.

rita-g-sonarsource (approved):
See comments please

src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger_CacheTests.cs
duncanp-sonar: "NoNewLocations" isn't very clear; the cache only needs to know the  entire set of current locations, not whether they are new or not.

I'll change it to `Refresh_NoCurrentLocations_CacheCleared`

src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger_CacheTests.cs
duncanp-sonar: I think the important point is whether the cached locations are current or not, not whether the supplied set of locations has changed or not. For me
`Refresh_AllCachedLocationsAreCurrent_NoAdormentsRemoved` is clearer.

src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger_CacheTests.cs
duncanp-sonar: `Refresh_NotAllCachedLocationsAreCurrent_ExpectedAdornmentsRemoved`

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger.cs
duncanp-sonar: There are arguments both ways.  We could just make `OnAggregatorTagsChanged` virtual, but we don't want subclasses to be able to override the base class behaviour (or rely on them calling the base class method). We could add another empty virtual method as an extension point that is called by the base class to all subclasses to extend if necessary, but at this point that would be adding an extension point specifically to handle the adornment tagger case. Or we have the adornment tagger adds its own functionality separately by registering to the aggregator event.

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger.cs
duncanp-sonar: Adding the parameter to the public constructor implies that the caller has a choice about the caching strategy to use, which isn't what is intended. This is pattern we've used in other places where an interface has been solely to enable testing of an internal implementation, which is the case here.
Normally when a caller is responsible for passing in arguments it's because they are shared instances/external dependencies, not internal internal implementation types.

src/IssueViz.UnitTests/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTagger_CacheTests.cs
duncanp-sonar: Calling it `Refresh` helps, but it's still odd to pass in the list of things you want to keep.

rita-g-sonarsource: Hm, maybe `Set` then?

duncanp-sonar: Seems odd to call `Initialize` multiple times, and `Set` and `Initialize` both imply that the cache is being populated i.e. new items created if necessary, which isn't the case. I think `Refresh` is good enough.

PR #1701, opened 2020-09-10
Fix object disposed exception thrown intermittently by tagger

### LINKED ISSUE #1693: Object disposed exception thrown by tagger
### Description

The preview pane of the "Map" view in the editor scroll bar can cause an "object disposed" error to be thrown:

```
// inside FilteringBaseTagger.GetTags when calling aggregator.GetTags:

    at Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregator`1.GetTags(NormalizedSnapshotSpanCollection snapshotSpans)
   at SonarLint.VisualStudio.IssueVisualization.Editor.FilteringTaggerBase`2.<GetTags>d__15.MoveNext() in D:\repos\sq\slvs\src\IssueViz\Editor\FilteringTaggerBase.cs:line 125
   at Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregator`1.<GetTagsForBuffer>d__47.MoveNext()
```

#### Repro
1. Open a file with analyzed issues
2. Turn on map view
image
3. Hover over the scroll bar so the code preview appears, then move the mouse so the preview disappears

#### Expected:
The taggers and error reporting continue to behave as normal.

#### Actual:
The error tagger is disposed when the preview pane disappears, causing an exception when it is later accessed.
Tags no longer appear for the file and issues do not appear in the error list until the file is closed and re-opened.

image

### Notes
The preview View creates aggregators to access e.g. our buffer error tagger. We create only one error tagger per buffer.
When the preview View disappears, it is disposed, which in turn disposes the aggregator and our singleton-per-buffer tagger.

This problem didn't occur with the old tagging code because it the TBIT was a view tagger that reference-counted the number of views that were created, and was only disposed when all of the views had closed.

duncanp-sonar:
Fixes #1693

### DIFF
--- a/src/IssueViz/Editor/FilteringTaggerBase.cs
+++ b/src/IssueViz/Editor/FilteringTaggerBase.cs
@@ -120,6 +120,17 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor
 
         public IEnumerable<ITagSpan<TTagType>> GetTags(NormalizedSnapshotSpanCollection spans)
         {
+            if (disposedValue)
+            {
+                // See https://github.com/SonarSource/sonarlint-visualstudio/issues/1693
+                // Intermittently GetTags gets called even after the tagger instance has been
+                // disposed. We don't want an exception to be thrown in this case as it causes
+                // a gold bar to appear. To the user, it would look like the bug is in our code,
+                // whereas in fact VS has disposed the tagger but is continuing to call it.
+                Debug.Fail($"GetTags called on disposed tagger. Tagger type: {this.GetType().FullName}. File: { GetSnapshot().TextBuffer.GetFilePath()}");
+                yield break;
+            }
+
             if (spans.Count == 0) { yield break; }
 
             var trackedTagSpans = tagAggregator.GetTags(spans).ToArray();
--- a/src/IssueViz/Editor/FilteringTaggerBase.cs
+++ b/src/IssueViz/Editor/FilteringTaggerBase.cs
@@ -127,7 +127,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor
                 // disposed. We don't want an exception to be thrown in this case as it causes
                 // a gold bar to appear. To the user, it would look like the bug is in our code,
                 // whereas in fact VS has disposed the tagger but is continuing to call it.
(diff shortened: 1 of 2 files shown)


src/IssueViz/Editor/FilteringTaggerBase.cs
```diff
@@ -120,6 +120,17 @@ private void OnAggregatorTagsChanged(object sender, BatchedTagsChangedEventArgs
...
         public IEnumerable<ITagSpan<TTagType>> GetTags(NormalizedSnapshotSpanCollection spans)
         {
+            if (disposedValue)
+            {
+                // See https://github.com/SonarSource/sonarlint-visualstudio/issues/1693
+                // Intermittently GetTags gets called even after the tagger instance has been
+                // disposed. We don't want an exception to be thrown in this case as it causes
+                // a gold bar to appear. To the user, it would look like the bug is in our code,
+                // whereas in fact VS has disposed the tagger but is continuing to call it.
+                Debug.Fail($"GetTags called on disposed tagger. Tagger type: {this.GetType().FullName}. File: { GetSnapshot().TextBuffer.GetFilePath()}");
```
rita-g-sonarsource, line 130: I think the Debug.Fail will be noisy. We can't do anything with it, that's VS's behavior. I think we should just do `yield break`.

rita-g-sonarsource (approved):

src/IssueViz/Editor/FilteringTaggerBase.cs
duncanp-sonar: Yep, it will be noisy when it happens. I was undecided between `Debug.Fail` and `Debug.WriteLine` and decided to start with `Fail` because I want to assess how often this actually occurs and what the impact is.
I'll change the committed version to `WriteLine` and keep the `Fail` on my machine for now.

rita-g-sonarsource: But why do you need the debug at all? What will you do with the information?

duncanp-sonar: I'm interested in two things: 1) does it happen in real world usage, and 2) if so does it have a noticeable impact on the user experience.

PR #1703, opened 2020-09-11
Move HandleFileRenames to SnapshotFactory

rita-g-sonarsource:


src/Integration.Vsix/ErrorList/SnapshotFactory.cs
```diff
@@ -18,17 +18,29 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.Linq;
 using Microsoft.VisualStudio.Shell.TableManager;
+using SonarLint.VisualStudio.Core.Helpers;
 
 namespace SonarLint.VisualStudio.Integration.Vsix
 {
+    internal interface ISnapshotFactory : ITableEntriesSnapshotFactory
```
duncanp-sonar, line 27: `Snapshot` is used both the the error list and text span code. It would be better to be more explicit about which one we are talking about here e.g. `IIssuesSnaphotFactory`.

It might be worth considering making the namespace more specific too.

src/Integration.Vsix/ErrorList/SnapshotFactory.cs
```diff
@@ -18,17 +18,29 @@
...
+    {
+        IIssuesSnapshot CurrentSnapshot { get; }
+
+        /// <summary>
+        /// Returns true/false if the factory was updated due to the rename
+        /// </summary>
+        bool HandleFileRename(string oldFilePath, string newFilePath);
```
duncanp-sonar, line 34: The snapshot construction creation/updating code needs some tidying up, but we can do it in a different sprint.
This change adds a "snapshot factory interface", but there are still a couple of static construction methods on `IssuesSnapshot` itself. Also, "factory" interface doesn`t actually ever create anything; it just does an in-place update of a snapshot. Still, it doesn't need to be fixed now.

duncanp-sonar (approved):

PR #1704, opened 2020-09-11
Visualization command default visibility

### LINKED ISSUE #1700: Show toolwindow command should be visible and enabled
### Description

Currently the "show tool window" command (`Views`, `Other windows`, `SonarLint Issue Visualization`) is not visible unless an issue with a secondary location is selected. This makes the command hard to find. It should at least be always visible.

TODO:
* ?should it also be always enabled? Yes.
* add the command to the SonarLint group in the Error List context menu
* change the flow panel overlay message to refer to SonarLint issues with secondary locations, and include a link to the wiki home page

rita-g-sonarsource:
Fixes #1700

### DIFF
--- a/src/IssueViz/Commands/IssueVizPackage.vsct
+++ b/src/IssueViz/Commands/IssueVizPackage.vsct
@@ -56,6 +56,12 @@
 
   </Commands>
 
+  <CommandPlacements>
+    <CommandPlacement guid="guidIssueVizPackageCmdSet" id="IssueVisualizationToolWindowCommandId" priority="0x200">
+      <Parent guid="guidDaemonCmdSet" id="grpDaemonErrorList" />
+    </CommandPlacement>
+  </CommandPlacements>
+
   <!-- Short cut keybindings: see https://docs.microsoft.com/en-us/visualstudio/extensibility/binding-keyboard-shortcuts-to-menu-items?view=vs-2019 -->
   <KeyBindings>
 
@@ -99,6 +105,10 @@
 
     <GuidSymbol name="guidImages" value="{DA8715CC-ADBA-45C7-BBB3-DBAA608EF40A}" >
       <IDSymbol name="bmpSonarQube" value="1"/>
+    </GuidSymbol>
+
+    <GuidSymbol name="guidDaemonCmdSet" value="{1F83EA11-3B07-45B3-BF39-307FD4F42194}">
+      <IDSymbol name="grpDaemonErrorList" value="0x100" />
     </GuidSymbol>
 
   </Symbols>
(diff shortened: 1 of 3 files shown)


src/IssueViz/IssueVizPackage.cs
```diff
@@ -33,7 +33,7 @@ namespace SonarLint.VisualStudio.IssueVisualization
     [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid("4F3D7D24-648B-4F3B-ACB0-B83AFE239210")]
     [ProvideMenuResource("Menus.ctmenu", 1)]
-    [ProvideToolWindow(typeof(IssueVisualizationToolWindow), MultiInstances = false, Style = VsDockStyle.Float, Width = 325, Height = 400)]
+    [ProvideToolWindow(typeof(IssueVisualizationToolWindow), MultiInstances = false, Transient = true, Style = VsDockStyle.Float, Width = 325, Height = 400)]
```
rita-g-sonarsource, line 36: This is important: the window has to be Transient (= always closed when VS is opened). Otherwise, since the command is always visible ( = not attached to a context), it means that the user can manually open it and leave it opened when VS shuts down. It will then cause a goldbar to appear next time VS is opened, complaining about our window.

src/IssueViz/Commands/IssueVizPackage.vsct
```diff
@@ -56,6 +56,12 @@
 
   </Commands>
 
+  <CommandPlacements>
+    <CommandPlacement guid="guidIssueVizPackageCmdSet" id="IssueVisualizationToolWindowCommandId" priority="0x200">
```
duncanp-sonar, line 60: Is this going to be visible for every type of issue, or just SonarLint issues? I don't think it should appear for all issues - that would just be confusing for the user.

rita-g-sonarsource: Hm, I don't see an option to specify a context for a command placement. There are [conditions](https://docs.microsoft.com/en-us/visualstudio/extensibility/vsct-xml-schema-conditional-attributes?view=vs-2019) but I don't think it's the same. I'll check it.

duncanp-sonar: Have a look at the existing "disable SonarLint rule" command. It checks whether the selected error is a Sonar error or not.
In this case, we might also want to set the enabled status based on whether issue has secondary locations or not.

It would be simpler to separate the changes: the first two commits are simple and are related, so commit them now, and work on the "add to context menu" separately since it's logically separate, needs more work, and it's not obvious what the UX should be.

PR #1711, opened 2020-09-15
Display Visualization command in the error list context menu

rita-g-sonarsource:


src/IssueViz/IssueVizPackage.cs
```diff
@@ -33,7 +33,7 @@ namespace SonarLint.VisualStudio.IssueVisualization
     [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid("4F3D7D24-648B-4F3B-ACB0-B83AFE239210")]
     [ProvideMenuResource("Menus.ctmenu", 1)]
-    [ProvideToolWindow(typeof(IssueVisualizationToolWindow), MultiInstances = false, Transient = true, Style = VsDockStyle.Float, Width = 325, Height = 400)]
+    [ProvideToolWindow(typeof(IssueVisualizationToolWindow), MultiInstances = false, Style = VsDockStyle.Float, Width = 325, Height = 400)]
```
rita-g-sonarsource, line 36: I had to remove the Transient.
I don't know why, but it caused the window to stop automatically appearing and disappearing.
I reset the hive, I did a fresh installation on 2017 to make sure that it wasn't caching that the window is closed, but it still didn't work.
I think we need to pick whether we want to document that a goldbar can appear, or remove the auto show/hide. :/

src/IssueViz/Commands/IssueVisualizationToolWindowCommand.cs
```diff
@@ -34,21 +35,38 @@ internal sealed class IssueVisualizationToolWindowCommand
...
-            commandService = commandService ?? throw new ArgumentNullException(nameof(commandService));
 
-            var menuCommandID = new CommandID(CommandSet, CommandId);
-            var menuItem = new MenuCommand(Execute, menuCommandID);
+            if (commandService == null)
+            {
+                throw new ArgumentNullException(nameof(commandService));
+            }
+
+            var menuCommandId = new CommandID(CommandSet, CommandId);
```
duncanp-sonar, line 60: Now this class is handling two menu commands it would help if some of the names were less generic e.g.
`CommandId`-> `ViewToolWindowCommandId`
`QueryStatus`-> `ErrorListQueryStatus`

src/IssueViz/Commands/IssueVisualizationToolWindowCommand.cs
```diff
@@ -34,21 +35,38 @@ internal sealed class IssueVisualizationToolWindowCommand
...
+            var menuItem = new MenuCommand(Execute, menuCommandId);
             commandService.AddCommand(menuItem);
+
+            menuCommandId = new CommandID(CommandSet, ErrorListCommandId);
+            ErrorListMenuItem = new OleMenuCommand(Execute, null, QueryStatus, menuCommandId);
```
duncanp-sonar, line 65: It also looks like a mistake that we are using the same `Execute` method for two seemly-different commands, even though it's not. A short comment would help e.g.
`We're showing the command in two places in the UI, but we only do a status check when it's called from the Error List context menu`.

src/IssueViz/Commands/IssueVizPackage.vsct
```diff
@@ -22,6 +22,15 @@
         </Strings>
       </Button>
 
+      <Button guid="guidIssueVizPackageCmdSet" id="IssueVisualizationToolWindowErrorListCommandId" priority="0x0200" type="Button">
+        <Parent guid="guidDaemonCmdSet" id="grpDaemonErrorList" />
+        <CommandFlag>DynamicVisibility</CommandFlag>
+        <CommandFlag>DefaultInvisible</CommandFlag>
+        <Strings>
+          <ButtonText>SonarLint Issue Visualization</ButtonText>
```
duncanp-sonar, line 30: I'm not sure about the text label for the menu item. It's ok for the tool window menu since it's similar to the other commands on the same menu, but it's not as meaningful in this context. I'm not sure would be better - `Show SonarLint Issue Visualization`?

src/IssueViz.UnitTests/Commands/IssueVisualizationToolWindowCommandTests.cs
```diff
@@ -38,41 +39,91 @@ public class IssueVisualizationToolWindowCommandTests
...
+                    x.AddCommand(It.Is((OleMenuCommand c) =>
+                        c.CommandID.Guid == IssueVisualizationToolWindowCommand.CommandSet &&
+                        c.CommandID.ID == IssueVisualizationToolWindowCommand.ErrorListCommandId)),
+                Times.Once);
+
+            commandService.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public void QueryStatus_FailsToQuery_NoException()
```
duncanp-sonar, line 97: `Fails` doesn't necessarily suggest an exception was thrown.
There's an existing test in this class called `Execute_ExceptionInCreatingWindow_NoException` which is less ambiguous, although I'd be more explicit and say `QueryStatus_NonCriticalException_IsSuppressed`.

duncanp-sonar (approved):
Just a few minor comments.

PR #1712, opened 2020-09-15
Changed dependant taggers to be per-request rather than per-buffer/view

duncanp-sonar:
Relates to #1693

### DIFF
--- a/src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
+++ b/src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
@@ -36,6 +36,7 @@ using SonarLint.VisualStudio.Core.Suppression;
 using SonarLint.VisualStudio.Integration.Vsix.Analysis;
 using SonarLint.VisualStudio.Integration.Vsix.ErrorList;
 using SonarLint.VisualStudio.Integration.Vsix.Resources;
+using SonarLint.VisualStudio.IssueVisualization.Editor;
 using SonarLint.VisualStudio.IssueVisualization.Editor.LanguageDetection;
 using SonarLint.VisualStudio.IssueVisualization.Models;
 
@@ -160,14 +161,20 @@ namespace SonarLint.VisualStudio.Integration.Vsix
 
             if (detectedLanguages.Any() && analyzerController.IsAnalysisSupported(detectedLanguages))
             {
-                var issueTracker = buffer.Properties.GetOrCreateSingletonProperty(typeof(IIssueTracker),
-                    () => new TextBufferIssueTracker(dte, this, textDocument, detectedLanguages, issuesFilter, sonarErrorDataSource, converter, logger));
+                // We only want one TBIT per buffer and we don't want it be disposed until
+                // it is not being used by any tag aggregators, so we're wrapping it in a SingletonDisposableTaggerManager
+                var singletonTaggerManager = buffer.Properties.GetOrCreateSingletonProperty(typeof(SingletonDisposableTaggerManager<IErrorTag>),
+                    () => new SingletonDisposableTaggerManager<IErrorTag>(_ => InternalCreateTextBufferIssueTracker(textDocument, detectedLanguages)));
 
-                return issueTracker as ITagger<T>;
+                var tagger = singletonTaggerManager.CreateTagger(buffer);
+                return tagger as ITagger<T>;
             }
 
             return null;
         }
+
+        private TextBufferIssueTracker InternalCreateTextBufferIssueTracker(ITextDocument textDocument, IEnumerable<AnalysisLanguage> analysisLanguages) =>
+            new TextBufferIssueTracker(dte, this, textDocument, analysisLanguages, issuesFilter, sonarErrorDataSource, converter, logger);
 
         #endregion IViewTaggerProvider members
 
--- a/src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
+++ b/src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
@@ -36,7 +36,6 @@ using SonarLint.VisualStudio.Core.Suppression;
 using SonarLint.VisualStudio.Integration.Vsix.Analysis;
 using SonarLint.VisualStudio.Integration.Vsix.ErrorList;
 using SonarLint.VisualStudio.Integration.Vsix.Resources;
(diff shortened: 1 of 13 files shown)


src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornmentTaggerProvider.cs
```diff
@@ -66,7 +66,7 @@ internal class IssueLocationAdornmentTaggerProvider : IViewTaggerProvider
                 return null;
             }
 
-            var tagger = textView.Properties.GetOrCreateSingletonProperty(typeof(IssueLocationAdornmentTagger), () => Create(textView as IWpfTextView));
+            var tagger = Create(textView as IWpfTextView);
```
duncanp-sonar, line 69: Same change for all of the non-singleton taggers - always return a new tagger instance when asked.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -160,15 +161,21 @@ private void OnAnalysisRequested(object sender, AnalysisRequestEventArgs args)
 
             if (detectedLanguages.Any() && analyzerController.IsAnalysisSupported(detectedLanguages))
             {
-                var issueTracker = buffer.Properties.GetOrCreateSingletonProperty(typeof(IIssueTracker),
-                    () => new TextBufferIssueTracker(dte, this, textDocument, detectedLanguages, issuesFilter, sonarErrorDataSource, converter, logger));
+                // We only want one TBIT per buffer and we don't want it be disposed until
```
duncanp-sonar: As discussed, I've reverted this change to the TaggerProvider/TBIT as it TaggerProvider has additional behaviours that will need more fixing up.

src/IssueViz.UnitTests/Editor/SingletonDisposableTaggerManagerTests.cs
```diff
@@ -0,0 +1,156 @@
...
+
+using System;
+using System.Linq;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Tagging;
+using Moq;
+using static SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.Common.TaggerTestHelper;
+using TypedSingleton = SonarLint.VisualStudio.IssueVisualization.Editor.SingletonDisposableTaggerManager<SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.SingletonDisposableTaggerManagerTests.IDummyTagType>;
```
duncanp-sonar, line 29: This is the longest 'using' alias I've ever seen. The singleton manager has generics and nested delegates so I added this alias to make the test code easier to read. Unfortunately, all of the paths in the alias have to be fully qualified, which is why it is so long.

src/IssueViz/Editor/LocationTagging/LocationTaggerProvider.cs
```diff
@@ -61,10 +61,16 @@ internal class LocationTaggerProvider : ITaggerProvider
...
 
-            // Need to make sure we only create one per buffer
-            var tagger = buffer.Properties.GetOrCreateSingletonProperty(typeof(LocationTagger),
-                () => new LocationTagger(buffer, locationStore, spanCalculator, logger));
+            // We only want one location tagger per buffer and we don't want it be disposed until
+            // it is not being used by any tag aggregators, so we're wrapping it in a SingletonDisposableTaggerManager
+            var taggerManager = buffer.Properties.GetOrCreateSingletonProperty(typeof(SingletonDisposableTaggerManager<IIssueLocationTag>),
+                () => new SingletonDisposableTaggerManager<IIssueLocationTag>(InternalCreateTagger));
+
+            var tagger = taggerManager.CreateTagger(buffer);
```
duncanp-sonar, line 69: This is the only change to the existing product code - switching the location tagger provider to create the singleton manager rather than the tagger directly.

src/IssueViz.UnitTests/Editor/Common/CommonTaggerProviderTestsBase.cs
```diff
@@ -46,7 +46,7 @@ public void CreateTagger_BufferIsNull_Throws()
         }
 
         [TestMethod]
-        public void CreateTagger_BufferIsNotNull_ReturnsSingletonTagger()
+        public void CreateTagger_BufferIsNotNull_ReturnsUniqueTaggerPerRequest()
```
rita-g-sonarsource, line 49: Maybe `new` instead of `unique`? It will be more evident that it's a new instance each time.

src/IssueViz.UnitTests/Editor/SingletonDisposableTaggerManagerTests.cs
```diff
@@ -0,0 +1,156 @@
...
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor
+{
+    [TestClass]
+    public class SingletonDisposableTaggerManagerTests
```
rita-g-sonarsource, line 34: I'm missing a test that checks there is indeed only one singleton when 2 taggers are created.

src/IssueViz.UnitTests/Editor/SingletonDisposableTaggerManagerTests.cs
rita-g-sonarsource: 🤯

src/IssueViz.UnitTests/Editor/SingletonDisposableTaggerManager_FlyweightTests.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        [TestMethod]
+        public void GetTags_ForwardsCall()
+        {
+            var singletonMock = CreateSingletonTaggerMock();
+            var inputSpans = new NormalizedSnapshotSpanCollection();
+
+            var testSubject = CreateTestSubject(singletonMock.Object, ValidOnDisposedDelegate);
+
+            testSubject.GetTags(inputSpans);
```
rita-g-sonarsource, line 64: It's worth also asserting that the result is forwarded as-is.

rita-g-sonarsource (approved):
Looks good, some minor test comments

PR #1713, opened 2020-09-15
Invalidate tag span when span length changes

rita-g-sonarsource:


rita-g-sonarsource:
@duncanp-sonar please take a look at the commits separately.
1. The first commit is the actual bug fix.
2. The second commit is test "refactoring" - I moved the buffer changed tests to another test suite, and as a result had to extract some methods to common helper classes.
3. The third commit is a test for the span invalidation, `BufferChanged_BufferChangedInsideTagSpan_TagSpanInvalidated`.

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTagger_BufferChangeTests.cs
```diff
@@ -0,0 +1,279 @@
...
+
+            var beforeSnapshot = CreateSnapshot(buffer, bufferLength, versionMock1);
+            var afterSnapshot = CreateSnapshot(buffer, bufferLength, versionMock2);
+
+            return new Tuple<ITextSnapshot, ITextSnapshot>(beforeSnapshot, afterSnapshot);
+        }
+
+        private static ITextChange CreateTextChange(Span oldSpan)
+        {
+            var newSpan = new Span(oldSpan.Start, 0); // simulate that the text in span was deleted
```
rita-g-sonarsource, line 237: This way I'm simulating that the span length was changed.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -147,7 +147,7 @@ private IssuesSnapshot(Guid snapshotId, string projectName, string filePath, IEn
 
         public override bool TryGetValue(int index, string keyName, out object content)
         {
-            if (index < 0 || this.issues.Count <= index)
+            if (index < 0 || index >= issues.Count || !issues[index].Span.HasValue || issues[index].Span.Value.IsEmpty)
```
duncanp-sonar, line 150: I think the new condition deserves a comment as it's not obvious why we're doing it e.g.
`If the requested issue doesn't have a valid span we return false so the Error List won't display it`

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -132,9 +132,16 @@ private void TranslateTagSpans(ITextSnapshot newSnapshot)
             foreach (var old in TagSpans)
             {
                 var newSpan = old.Span.TranslateTo(newSnapshot, SpanTrackingMode.EdgeExclusive);
-                old.Tag.Location.Span = newSpan; // update the span stored in the location visualization 
 
-                translatedTagSpans.Add(new TagSpan<IIssueLocationTag>(newSpan, old.Tag));
+                if (newSpan.Length != old.Span.Length)
+                {
+                    old.Tag.Location.InvalidateSpan();
```
duncanp-sonar, line 138: It wouldn't hurt to explain our decision here either e.g. `If the user has typed inside the tagged region we'll stop showing a squiggly for that item`

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTagger_BufferChangeTests.cs
```diff
@@ -0,0 +1,256 @@
...
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using static SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.Common.TaggerTestHelper;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.LocationTagging
+{
+    [TestClass]
+    public class LocationTagger_BufferChangeTests
+    {
+        [TestMethod]
+        public void BufferChanged_FileIsRenamed_CurrentFileNameIsUsed()
```
duncanp-sonar, line 40: I'm assuming there are no changes to these tests, other than them being moved.

rita-g-sonarsource: No logic changes, just cosmetics stemming from different helper methods.

duncanp-sonar (approved):
Looks good, just some comments about comments.

PR #1715, opened 2020-09-16
Fix Visualization tool window goldbar and auto show/hide

rita-g-sonarsource:


src/IssueViz/IssueVisualizationControl/IssueVisualizationToolWindow.cs
```diff
@@ -57,6 +57,9 @@ protected override void Dispose(bool disposing)
             if (disposing)
             {
                 viewModel.Dispose();
+
+                var vsWindowFrame = Frame as IVsWindowFrame;
+                vsWindowFrame.CloseFrame((uint)__FRAMECLOSE.FRAMECLOSE_NoSave);
```
duncanp-sonar, line 62: Presumably this is testable?

duncanp-sonar (approved):
The change looks ok, but why no unit test?

rita-g-sonarsource:
> The change looks ok, but why no unit test?

I don't know what you're talking about, of course there are unit tests 😛

duncanp-sonar:
My mistake, I wasn't wearing my "seeing into the future" glasses ;-)

Issue #1644, opened 2020-08-19
Adornments and highlight UX improvements


rita-g-sonarsource:
### Description 

~~- [ ] display the primary location as well?~~
- [x] highlight colors that work in different VS themes
- [x] tooltips with location message
- [x] prominent step number adornment?
~~- [ ] additional UX in other SL flavors? (clickable adornments?)~~

PR #1718, opened 2020-09-16
Use different highlight and adornment colors

rita-g-sonarsource:
Part of #1644

image


src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornment.cs
```diff
@@ -18,36 +18,34 @@
...
             // e.g. if the user changes the font size
             LocationViz = locationViz;
 
-            Margin = new System.Windows.Thickness(1, 0, 1, 0); // Space between this UI element and the editor text
-            Padding = new System.Windows.Thickness(0);  // Space between the side of the control and its content    
-
-            Background = Brushes.Pink;
-            BorderBrush = Brushes.LightCoral;
-            BorderThickness = new System.Windows.Thickness(1);
+            Margin = new Thickness(3, 0, 3, 0); // Space between this UI element and the editor text
```
rita-g-sonarsource, line 38: Similar to other SL flavors https://github.com/SonarSource/sonarlint-intellij/blob/master/src/main/java/org/sonarlint/intellij/editor/SecondaryLocationIndexRenderer.java#L40

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornment.cs
```diff
@@ -18,36 +18,34 @@
...
 
             // Visible content of the adornment
-            Content = new TextBlock
+            Child = new TextBlock
             {
                 Text = LocationViz.StepNumber.ToString(),
-                Padding = new System.Windows.Thickness(0.2, 0, 0.2, 0),
-                Margin = new System.Windows.Thickness(0),
+                FontWeight = FontWeights.SemiBold,
+                Padding = new Thickness(4, 0, 4, 0)
```
rita-g-sonarsource, line 48: Similar to other SL flavors https://github.com/SonarSource/sonarlint-intellij/blob/master/src/main/java/org/sonarlint/intellij/editor/SecondaryLocationIndexRenderer.java#L41
(5 seemed a bit too much, so I went with 4)

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornment.cs
```diff
@@ -18,36 +18,34 @@
...
+using System.Windows;
 using System.Windows.Controls;
-using System.Windows.Media;
 using Microsoft.VisualStudio.Text.Formatting;
 using SonarLint.VisualStudio.IssueVisualization.Models;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Editor.SelectedIssueTagging.Adornment
 {
-    internal class IssueLocationAdornment : Button
+    internal class IssueLocationAdornment : Border
```
rita-g-sonarsource, line 28: Changed to Border so I could set CornerRadius, similar to the other SL flavors

src/IssueViz/Editor/SelectedIssueTagging/Adornment/IssueLocationAdornment.cs
```diff
@@ -60,8 +58,15 @@ public IssueLocationAdornment(IAnalysisIssueLocationVisualization locationViz, I
 
         public void Update(IFormattedLineSource formattedLineSource)
         {
-            FontSize = formattedLineSource.DefaultTextProperties.FontRenderingEmSize;
-            FontFamily = formattedLineSource.DefaultTextProperties.Typeface.FontFamily;
+            var textBlock = Child as TextBlock;
+
+            textBlock.Foreground = formattedLineSource.DefaultTextProperties.ForegroundBrush;
+            textBlock.FontSize = formattedLineSource.DefaultTextProperties.FontRenderingEmSize - 2;
```
rita-g-sonarsource, line 64: I wanted to make the number stand out without making the adornment too big, so I made the font Bold (line 47) and the font a bit smaller as a compromise. I do the same thing in the Viz panel.

src/IssueViz/Editor/SelectedIssueTagging/ThemeColors.cs
```diff
@@ -0,0 +1,70 @@
...
+        internal static readonly IThemeColors LightTheme = new LightThemeColors();
+        internal static readonly IThemeColors DarkTheme = new DarkThemeColors();
+
+        public static bool IsLightTheme(Brush textBrush)
+        {
+            var isLightTheme = true;
+
+            if (textBrush is SolidColorBrush solidColorBrush)
+            {
+                var isDarkText = solidColorBrush.Color.IsDark();
```
rita-g-sonarsource, line 37: IsDark/IsLight is a built-in functionality.
Basically, we need to set a contrast between the text and the highlight/adornment. So if the text font is dark, then we should display a light adornment and highlight. Otherwise, if the text is light, then we should display a dark adornment.

src/IssueViz/Editor/SelectedIssueTagging/ThemeColors.cs
```diff
@@ -0,0 +1,70 @@
...
+            return isLightTheme;
+        }
+
+        public static IThemeColors BasedOnText(Brush textBrush)
+        {
+            return IsLightTheme(textBrush) ? LightTheme : DarkTheme;
+        }
+    }
+
+    internal class LightThemeColors : IThemeColors
```
rita-g-sonarsource, line 50: I went with colors that are brighter than the ones we use in other SL flavors.
Which ones do you think look better?
Top is the ones in this PR, bottom is the ones in other SL flavors.

image

rita-g-sonarsource:
@duncanp-sonar I tried various colors and opacity levels; I couldn't find a color that looked good in both dark and light theme. In the end I opted for different theme colors, like we do in other SL flavors.

src/IssueViz/Editor/SelectedIssueTagging/ThemeColors.cs
duncanp-sonar: My preferences would be top-right and bottom-left, but it's no a strong preference.
The light border around the highlight in dark mode looks a bit odd, though.

duncanp-sonar (approved):

PR #1724, opened 2020-09-17
Add light bulb to show/hide issue viz

### LINKED ISSUE #1624: Investigate using light bulb suggestion to hide/show adornments
### Description

Investigate using quick info to hide/show adornments:
- [ ] technical investigation - https://docs.microsoft.com/en-us/visualstudio/extensibility/walkthrough-displaying-light-bulb-suggestions?view=vs-2019
- [ ] UX evaluation

rita-g-sonarsource:
Fixes #1624


src/IssueViz/Editor/QuickActions/BaseSuggestedAction.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public abstract void Invoke(CancellationToken cancellationToken);
+
+        public ImageMoniker IconMoniker => new ImageMoniker();
+        public string IconAutomationText => null;
+        public string InputGestureText => null;
+        public bool HasPreview => false;
+        public bool HasActionSets => false;
+
+        public Task<IEnumerable<SuggestedActionSet>> GetActionSetsAsync(CancellationToken cancellationToken)
```
duncanp-sonar, line 43: Cosmetic - is there any reason not to use `=>` for this method and the next one?

src/IssueViz/Editor/QuickActions/DeselectIssueVisualizationAction.cs
```diff
@@ -0,0 +1,49 @@
...
+using SonarLint.VisualStudio.IssueVisualization.Selection;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Editor.QuickActions
+{
+    internal class DeselectIssueVisualizationAction : BaseSuggestedAction
+    {
+        private readonly IAnalysisIssueSelectionService selectionService;
+
+        /// <summary>
+        /// <see cref="IssueLocationActionsSource.GetSuggestedActions"/> can return a cached version of the actions, which will show a deselect action even though the issue has already been deselected.
```
duncanp-sonar, line 31: It's not clear what this means. Does it mean a lightbulb for the deselect action will be shown when it shouldn't be?

src/IssueViz/Editor/QuickActions/IssueLocationActionsSource.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                return new[] {new SuggestedActionSet(allActions)};
+            }
+
+            return Enumerable.Empty<SuggestedActionSet>();
+        }
+
+        public Task<bool> HasSuggestedActionsAsync(ISuggestedActionCategorySet requestedActionCategories, SnapshotSpan range, CancellationToken cancellationToken)
+        {
+            return Task.Factory.StartNew(() => IsOnIssueVisualization(range, out _) || IsOnSelectedVisualization(range),
```
duncanp-sonar, line 80: Is this the recommended way to start new tasks in the VS threading cookbook?

rita-g-sonarsource: It is how they recommend it in the walkthrough
https://docs.microsoft.com/en-us/visualstudio/extensibility/walkthrough-displaying-light-bulb-suggestions?view=vs-2019#implement-the-isuggestedactionsource

src/IssueViz/Editor/QuickActions/IssueLocationActionsSource.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        public event EventHandler<EventArgs> SuggestedActionsChanged;
+
+        public IEnumerable<SuggestedActionSet> GetSuggestedActions(ISuggestedActionCategorySet requestedActionCategories, SnapshotSpan range, CancellationToken cancellationToken)
+        {
+            var allActions = new List<ISuggestedAction>();
+
+            if (IsOnIssueVisualization(range, out var issueVisualizations))
+            {
+                var actions = issueVisualizations.Select(x => new SelectIssueVisualizationAction(selectionService, x));
```
duncanp-sonar, line 61: Will we show the light bulb for an issue that is already selected?

src/IssueViz/Editor/QuickActions/DeselectIssueVisualizationAction.cs
rita-g-sonarsource: Yes. In general, GetSuggestedActions can return cached actions. For example:
1. Place the caret on span that should have a lightbulb
2. Open the lightbulb with the keyboard shortcut
3. Select an action ==> action works
4. Don't move the caret, and open the lightbulb again with the keyboard shortcut ==> GetSuggestedActions won't be called. The list will be populated with previous items. If move the caret, GetSuggestedActions is called.
Registering to TagsChanged should've fixed that, as it raises the SuggestedActionsChanged event. But it doesn't. SuggestedActionsChanged is indeed raised, but GetSuggestedActions is still not being called.

src/IssueViz/Editor/QuickActions/IssueLocationActionsSource.cs
```diff
@@ -0,0 +1,123 @@
...
+            return Enumerable.Empty<SuggestedActionSet>();
+        }
+
+        public Task<bool> HasSuggestedActionsAsync(ISuggestedActionCategorySet requestedActionCategories, SnapshotSpan range, CancellationToken cancellationToken)
+        {
+            return Task.Factory.StartNew(() => IsOnIssueVisualization(range, out _) || IsOnSelectedVisualization(range),
+                cancellationToken);
+        }
+
+        private bool IsOnIssueVisualization(SnapshotSpan range, out IEnumerable<IAnalysisIssueVisualization> issueVisualizationsWithSecondaryLocations)
```
duncanp-sonar, line 84: The method name is bit misleading.
`IsOnIssueWithSecondaryLocations` / `ContainsIssuesWithSecondaryLocations` would be clearer.

src/IssueViz/Editor/QuickActions/IssueLocationActionsSource.cs
rita-g-sonarsource: Yes, I was debating about this. If there are multiple issues on the same line, one of which is selected and another isn't, I thought it might be confusing to the user -- as he might think that the "Show" action is for the issue that he already sees, and not a different one. This way it's more clear that there are 2+ issues on the same row.

src/IssueViz/Editor/QuickActions/IssueLocationActionsSource.cs
```diff
@@ -0,0 +1,123 @@
...
+            selectedIssueLocationsTagAggregator.TagsChanged += TagAggregator_TagsChanged;
+        }
+
+        public event EventHandler<EventArgs> SuggestedActionsChanged;
+
+        public IEnumerable<SuggestedActionSet> GetSuggestedActions(ISuggestedActionCategorySet requestedActionCategories, SnapshotSpan range, CancellationToken cancellationToken)
+        {
+            var allActions = new List<ISuggestedAction>();
+
+            if (IsOnIssueVisualization(range, out var issueVisualizations))
```
duncanp-sonar, line 59: It's not immediately obvious what the desired behaviour is; a comment would help. I think the expected behaviour is:

> * Add a "show" lightbulb for each issue in the range that has secondary locations, and
> * Add a single "hide" lightbulb if the range contains any selected locations

src/IssueViz/Editor/QuickActions/DeselectIssueVisualizationAction.cs
rita-g-sonarsource: Btw, unrelated: do you think it's worth uncapitalizing the text? e.g. `cpp:S3776: hide SonarLint issue visualization` ?

duncanp-sonar (commented):
See inline; some questions about the behaviour.

src/IssueViz/Editor/QuickActions/DeselectIssueVisualizationAction.cs
duncanp-sonar: Caps - look at what other suggestions do and go with whatever is most common.

src/IssueViz/Editor/QuickActions/IssueLocationActionsSource.cs
duncanp-sonar: So currently we'll always suggest a pointless action in what is probably the most common case of a single issue already being selected? That could be annoying.

Two issues on the same line for the same rule seems pretty uncommon.

src/IssueViz/Editor/QuickActions/IssueLocationActionsSource.cs
duncanp-sonar: It's always work checking the VS threading cookbook, which was written by the platform devs and contains recommended best practices. It's less clear whether examples in docs follow best practices or not.
Unfortunately, the cookbook doesn't have a recommendation for calling synchronous code when implementing an async method.
Fingers crossed this is something that is checked by the VS async threading analyzers.

src/IssueViz/Editor/QuickActions/IssueLocationActionsSource.cs
rita-g-sonarsource: Yep possibly. Ok, I'll add a check that it's not already selected.

duncanp-sonar (commented):
I think it's worth not showing a pointless "show" action all the time. Other than that looks ok.

PR #1725, opened 2020-09-17
[Option B] Show primary location in the panel

### LINKED ISSUE #1721: UX improvement: show primary location in the Visualization panel
### Description

There is no way to jump from the Visualization panel to the location of the primary issue. The only way is to find the issue in the error list and double-click on it from there.

TODO: 
Add an option in the panel to jump to the primary location.

rita-g-sonarsource:
Fixes #1721

image
image

image
image


src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -221,15 +236,30 @@
...
+                    <TextBlock Grid.Row="1" 
+                               Grid.Column="1" 
+                               Style="{StaticResource BaseTextBlockStyle}">
                         <Run Text="{Binding Description, Mode=OneWay}"/>
                         <Hyperlink NavigateUri="{Binding RuleHelpLink}" 
-                                   RequestNavigate="Hyperlink_OnRequestNavigate"
-                                   ToolTip="{Binding RuleKey, StringFormat={}(Open description of rule {0}), Mode=OneWay}">
+                                   RequestNavigate="Hyperlink_OnRequestNavigate">
+                            <Hyperlink.ToolTip>
+                                <TextBlock Text="{Binding Path=RuleKey, StringFormat=Open description of rule {0}}" />
```
rita-g-sonarsource, line 261: Unrelated fix: turns out that the tooltip string format wasn't working.

src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -68,6 +69,25 @@ internal sealed class IssueVisualizationViewModel : INotifyPropertyChanged, IDis
...
+                if (issueSpan == null || issueSpan.Value.IsEmpty)
+                {
+                    return null;
+                }
+
+                var position = issueSpan.Value.Start;
+                var line = position.GetContainingLine();
+                var fileName = Path.GetFileName(CurrentIssue.CurrentFilePath);
+
+                return $"{fileName}:{line.LineNumber}";
```
rita-g-sonarsource, line 87: I have a very bizarre behavior: the line number is (-1).
This is the exact same code that we use in IssuesSnapshot, yet there the line number is correct.
I debugged and it seems like when the span changes:
1. the error list gets notified like 20 times
2. the panel gets notified
3. the error list gets notified _again_, this time with the updated line.

I'm not sure what is happening.

duncanp-sonar: What do you mean the line number is (-1)? It's one lower than you expect?

I think the line number is useful information, and the label is a bit confusing: `Source1.cpp, line 3` would be clearer.

rita-g-sonarsource: One lower, yes. So the error list would say "line 5" - and panel "line 4"

duncanp-sonar: It looks a bit weird with primary location above the description; it's a location, but it's separated from the other locations, and doesn't look the same (no file type icon).

duncanp-sonar:
The changes look ok from a technical point of view.

Some comments on the UI:
* It's not obvious that the primary location is clickable. It's physically separated from the other clickable locations and it doesn't look the same. It looks like a title. Unless you happen to move the mouse over it you probably won't realise it's clickable.
* the secondary locations are keyboard-navigable in the pane but the primary isn't (minor: there are others way of navigating so it shouldn't be an accessibility issue, but it's inconsistent).

PR #1727, opened 2020-09-18
Fix tooltip bug

### LINKED ISSUE #1720: Tooltips do not appear for primary locations
Repro steps:

1. Analyze a file that creates an issue.
2. Hover over the squiggly in the editor

Expected ==> Tooltip is shown.
Actual ==> No tooltip.

duncanp-sonar:
Fixes #1720

### DIFF
--- a/src/IssueViz/Editor/LocationTagging/LocationTagger.cs
+++ b/src/IssueViz/Editor/LocationTagging/LocationTagger.cs
@@ -134,7 +134,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging
             // Find any tags in that overlap with that range
             foreach (var tagSpan in TagSpans)
             {
-                if (OverlapsExists(tagSpan.Span, spans))
+                if (IntersectionExists(tagSpan.Span, spans))
                 {
                     yield return tagSpan;
                 }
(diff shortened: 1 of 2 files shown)


src/IssueViz/Editor/LocationTagging/LocationTagger.cs
```diff
@@ -172,8 +172,13 @@ private void TranslateTagSpans(ITextSnapshot newSnapshot)
...
         }
 
-        private static bool OverlapsExists(Span span, NormalizedSnapshotSpanCollection spans) =>
-            spans.Any(x => x.OverlapsWith(span));
+        private static bool IntersectionExists(Span span, NormalizedSnapshotSpanCollection spans) =>
+            // We're using "IntersectsWith" rather than "OverlapsWith" so that zero-length spans are handled correctly.
+            // Given two spans A = [10 -> 15] and B = [12 -> 12],
+            // * A.IntersectsWith(B) == true
+            // * A.OverlapsWith(B) == false.
+            // This matters because when VS is requesting tags to show tooltips it uses a zero-length span.
```
rita-g-sonarsource, line 180: It uses a zero-length span?? Ohhh so that's why!

rita-g-sonarsource (approved):

src/IssueViz/Editor/LocationTagging/LocationTagger.cs
duncanp-sonar: Yep, and an empty span overlaps with nothing, even itself.  However, it will intersect.

rita-g-sonarsource: 🤦 🤯

PR #1747, opened 2020-09-22
Fix selection bugs when issue is selected using suggested action lightbulb

### LINKED ISSUE #1746: Selecting an issue with suggested action lightbulb causes selection side effects
### Description

The lightbulb suggested action sets selected issue without going through the error list. This causes several undesired side effects.

#### repro scenario 1
1. Analyze a file with secondary locations
2. Open visualization panel using the lightbulb
3. Re-analyze the file

Expected ==> visualization panel is cleared.
Actual ==> old issue is still shown in the panel but adornments have disappeared. 

#### repro scenario 2
1. Analyze a file with secondary locations
2. Open visualization panel using the lightbulb
3. Edit the code and delete the code that causes the issue (without saving the file)

Expected ==> visualization panel is cleared
Actual ==> issue is still shown. When clicking on the items, the locations become non navigable.

#### repro scenario 3
1. Analyze a file with secondary locations
2. Open visualization panel using the lightbulb
3. Edit the code and add/remove lines before the primary location

Expected ==> line number is updated in the panel
Actual ==> line number is not updated

#### repro scenario 4
1. Analyze a file with secondary locations
2. Open visualization panel using the lightbulb
3. Rename the file

Expected ==> visualization panel is cleared
Actual ==> visualization panel still shows issue with the old file path.

#### repro scenario 5
1. Analyze a file with secondary locations
2. Open visualization panel using the lightbulb
3. Close the file

Expected ==> visualization panel is cleared
Actual ==> visualization panel still shows the issue. Clicking on an item causes the a new editor to be opened in place of the panel window.

rita-g-sonarsource:
Fixes #1746

### DIFF
--- a/src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
+++ b/src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
@@ -29,6 +29,7 @@ using SonarLint.VisualStudio.Core.Helpers;
 using SonarLint.VisualStudio.IssueVisualization.Editor;
 using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
 using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Selection;
 using SonarLint.VisualStudio.IssueVisualization.TableControls;
 
 namespace SonarLint.VisualStudio.Integration.Vsix.ErrorList
@@ -43,11 +44,14 @@ namespace SonarLint.VisualStudio.Integration.Vsix.ErrorList
         IDisposable
     {
         private readonly IFileRenamesEventSource fileRenamesEventSource;
+        private readonly IAnalysisIssueSelectionService selectionService;
         private readonly ISet<ITableDataSink> sinks = new HashSet<ITableDataSink>();
         private readonly ISet<IIssuesSnapshotFactory> factories = new HashSet<IIssuesSnapshotFactory>();
 
         [ImportingConstructor]
-        internal SonarErrorListDataSource(ITableManagerProvider tableManagerProvider, IFileRenamesEventSource fileRenamesEventSource)
+        internal SonarErrorListDataSource(ITableManagerProvider tableManagerProvider, 
+            IFileRenamesEventSource fileRenamesEventSource,
+            IAnalysisIssueSelectionService selectionService)
         {
             if (tableManagerProvider == null)
             {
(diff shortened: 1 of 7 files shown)


src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -43,18 +44,22 @@ internal sealed class SonarErrorListDataSource :
...
         private readonly IFileRenamesEventSource fileRenamesEventSource;
+        private readonly IAnalysisIssueSelectionService selectionService;
         private readonly ISet<ITableDataSink> sinks = new HashSet<ITableDataSink>();
         private readonly ISet<IIssuesSnapshotFactory> factories = new HashSet<IIssuesSnapshotFactory>();
 
         [ImportingConstructor]
-        internal SonarErrorListDataSource(ITableManagerProvider tableManagerProvider, IFileRenamesEventSource fileRenamesEventSource)
+        internal SonarErrorListDataSource(ITableManagerProvider tableManagerProvider, 
+            IFileRenamesEventSource fileRenamesEventSource,
+            IAnalysisIssueSelectionService selectionService)
```
duncanp-sonar, line 54: This doesn't feel right. The responsibility of the data source class is to handle the integration with the VS Error List. It shouldn't need to know about the selection service. If we need more complicated logic to synchronize the selection service/panel/error list it should be handled elsewhere.

rita-g-sonarsource: I agree. Currently, the analysis results are reported directly to the Snapshot classes. That should've been the place to synchronize the selection service, however I felt that it was too big of a change to do in the last minute. I agree though; I also had to copy logic into the panel, which I really didn't want to do. It was just to minimize the scope of the changes.

rita-g-sonarsource: Btw, I thought about doing a hacky solution and in the lightbulb just selecting the item in error list. However, the error list might not yet be loaded (and the lightbulb is), and also the error list might have a filter, in which case there is nothing to select.

duncanp-sonar: If you're knowingly doing something hacky it's worth capturing that information somehow, either in comments in the code or the PR, otherwise it's very hard later to work out later if there was a good reason for something that looks wrong.

I agree about the underlying problem, but I don't think a different state service is the answer. The error list data source is still the single source of truth about the set of active issues and it raises an event when the set of issues changes.
The behaviour regarding the visualisation of the current selection is a UI issue that should be handled at a higher level.

PR #1755, opened 2020-09-28
Use callerSnapshot instead of currentSnapshot when calling GetSpans

### LINKED ISSUE #1752: Debug assertion when changing a span at the start of a secondary location highlight
#### Repro steps
1. Analyze code with secondary locations, i.e.
```
class OneSecondary {
    void f();
    ~OneSecondary() {
        f();
    }
};
```
2. Select an issue with secondary locations --> adornments and highlights appear 
3. Change the span **right before the highlight**, i.e. comment-out the line, drop to new line

Expected --> No debug exception
Actual --> Multiple debug exceptions are thrown

image

rita-g-sonarsource:
Fixes #1752

### DIFF
--- a/src/IssueViz/Editor/FilteringTaggerBase.cs
+++ b/src/IssueViz/Editor/FilteringTaggerBase.cs
@@ -138,18 +138,11 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor
 
             var filteredTagSpans = Filter(trackedTagSpans);
 
-            var currentSnapshot = GetSnapshot();
             var callerSnapshot = spans[0].Snapshot;
-
-            if (callerSnapshot.Version != currentSnapshot.Version)
-            {
-                Debug.Fail($"Expecting the tags to reference the correct snapshot. Caller snapshot version: {callerSnapshot.Version}, tag snapshot version: {currentSnapshot.Version}");
-                yield break;
-            }
 
             foreach (var dataTagSpan in filteredTagSpans)
             {
-                var filterSpans = dataTagSpan.Span.GetSpans(currentSnapshot);
+                var filterSpans = dataTagSpan.Span.GetSpans(callerSnapshot);
                 if (spans.IntersectsWith(filterSpans))
                 {
                     yield return CreateTagSpan(dataTagSpan.Tag, spans);


rita-g-sonarsource:
@duncanp-sonar I'm not sure about this fix.
It seems that when we modify the buffer right at the beginning of the span, `GetTags` was being called with a newer snapshot but `GetSnapshot` would still return an older version, which caused an exception to be thrown due to Debug.Fail.
I removed the Debug.Fail check, and instead called `dataTagSpan.Span.GetSpans(callerSnapshot)` (instead of the previous currentSnapshot). `GetSpans` is supposed to make the translation to the given snapshot, and it's similar to [roslyn's code](https://github.com/dotnet/roslyn/blob/9cf5d95c611aa430c9599f824e1aa785c53d5d15/src/EditorFeatures/Core.Wpf/InlineParameterNameHints/InlineParameterNameHintsTagger.cs#L123).
I'm not sure of the effect of this change. It seems to work correctly but I'm not sure.

src/IssueViz/Editor/FilteringTaggerBase.cs
```diff
@@ -138,18 +138,11 @@ public IEnumerable<ITagSpan<TTagType>> GetTags(NormalizedSnapshotSpanCollection
...
-            if (callerSnapshot.Version != currentSnapshot.Version)
-            {
-                Debug.Fail($"Expecting the tags to reference the correct snapshot. Caller snapshot version: {callerSnapshot.Version}, tag snapshot version: {currentSnapshot.Version}");
-                yield break;
-            }
-
             foreach (var dataTagSpan in filteredTagSpans)
             {
-                var filterSpans = dataTagSpan.Span.GetSpans(currentSnapshot);
+                var filterSpans = dataTagSpan.Span.GetSpans(callerSnapshot);
```
duncanp-sonar, line 145: This change should be fine. The spans returned by the tag aggregator seem to a type of tracking span, so they should be translated correctly when using `callerShapshot`.

I'm surprised our view tagger is being asked for tags for a newer version of the snapshot than the one currently the view is pointing to; I would have expected the view to be pointing at the latest version. 
However, that is what is happening. Our tagger is being called by the VS OutliningManager in response to a buffer change event, and this is happening before the text view updates the snapshot it is pointing to.

duncanp-sonar (approved):

PR #1756, opened 2020-09-30
Update embedded CFamily analyzer to v6.13.0.22261

Godin:


src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -276,7 +276,7 @@ public void GetKeyValueOptionsList_UsingRealEmbeddedRulesJson()
             CheckHasOption("S1578.format=^[A-Za-z_-][A-Za-z0-9_-]+\\.(c|m|cpp|cc|cxx)$");
 
             options.Count().Should()
-                .BeGreaterOrEqualTo(39); // basic sanity check: v6.6 has 39 - not expecting options to be removed
+                .BeGreaterOrEqualTo(37); // basic sanity check: v6.13 has 37 - not expecting options to be removed
```
Godin, line 279: > not expecting options to be removed

well, only unless it happens 😉 
rule `CommentRegularExpression` with `2` parameters was removed in `6.13` - https://jira.sonarsource.com/browse/CPP-2609

src/Integration.Vsix/CFamily/PortedFromJava/Protocol.cs
```diff
@@ -50,6 +50,8 @@ public static void Write(BinaryWriter writer, Request request)
             Write(writer, request.Macros);
             WriteUTF(writer, request.TargetTriple);
             WriteUTF(writer, request.File);
+            WriteUTF(writer, /* pchDir */ "");
```
rita-g-sonarsource, line 53: @Godin what are these new fields? I couldn't find anything in your release notes that relate to this.

Godin: @rita-g-sonarsource you're looking on the wrong commits - here is commits from version 6.12 to 6.13:
https://github.com/SonarSource/sonar-cpp/compare/6.12.0.20255...6.13.0.22261

And these fields related to https://jira.sonarsource.com/browse/CPP-1602 , which is relevant only for SonarQube and SonarClound, should be empty (`""`) in case of SonarLint.

Issue #1304, opened 2020-03-27
Visual Studio 2015, 2017 and 2019  Extension "Cannot analyze header files"


Juan-Ramirez-Alvarez:
### Description

Visual Studio versions 2015, 17 and 19 do not analyze header files in C++.
It does not matter if using local user settings files or grabbing rules from SonarQube server
Using extension version 4.20.0.16339

### Repro steps
1. Begin a C++ project in any of the above listed visual studio versions
2. Install the appropriate SonarLint extension to the Visual Studio version being run
3. Create a .h file
4. See Visual Studio SonarLint window give error "Cannot analyze header files. File: <filename>"
5. Verify in .h source window that no linting was done with SonarLint

Is this a known defect or missing feature?

### Expected behavior

Linting should happen in .h files
Although not much logic is typically implemented, some orgs have implemented full logic in header.
This is the limitation I'm currently running into, while trying to verify if my changes to a header are fixing security vulnerabilities shown in SonarQube.

### Actual behavior

No linting happens in C++ headers
See output regarding this in linter output window

### Known workarounds

NA

### Related information

* SonarLint for VS Version 4.20.0.16339
* Visual Studio Version Professional/Enterprise 2015-2019
* SonarQube server version - UNKNOWN/ Not relevant
* Languages in your solution - C++

Juan-Ramirez-Alvarez:
https://github.com/SonarSource/sonarlint-visualstudio/pull/677
Does this mean it is not supported?

duncanp-sonar:
@Juan-Ramirez-Alvarez that's correct, analysing header files in SLVS is not currently supported.

PR #1757, opened 2020-10-04
Support header files + Improve unsupported files error messages

### LINKED ISSUE #1547: Improve error messages about files that cannot be analyzed
### Description

Currently we only support CFamily analysis for "native" VS projects. However, VS has limited supports for other types of project configuration e.g. [MakeFile](https://docs.microsoft.com/en-us/cpp/build/reference/creating-a-makefile-project?view=vs-2019) and [CMake](https://docs.microsoft.com/en-us/cpp/build/reference/creating-a-makefile-project?view=vs-2019).

For non-native project types, we can't collect the necessary configuration information. Depending on the project type, our code throws different errors from different locations, some of which are more cryptic than others.

Also, there are so many types of files in the Visual Studio project that are not meant to be analyzed. We should print a clear message for these, and only try to analyze "C/C++ compiler" as source files and "C/C++ header" as header files.

We could improve the UX by reporting a more friendly message for project types we know are not currently supported.

### Related tickets
#903 Support analysis of CMake projects in VS2019
#1546 Unable to collect C/C++ configuration in MakeFile-project

abbas-sabra-sonarsource:
Fixes #1547 and #1304

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
@@ -37,24 +37,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
     [TestClass]
     public class CFamilyHelperTests
     {
-        private const string ValidPlatformName = "Win32";
-
-        private static readonly IDictionary<string, string> MandatoryProjectConfigProperties = new Dictionary<string, string>
-        {
-            ["PlatformToolset"] = "v140_xp"
-        };
-
-        private static readonly IDictionary<string, string> MandatoryFileConfigProperties = new Dictionary<string, string>
-        {
-            ["PrecompiledHeader"] = "NotUsing",
-            ["CompileAs"] = "CompileAsCpp",
-            ["CompileAsManaged"] = "false",
-            ["EnableEnhancedInstructionSet"] = "",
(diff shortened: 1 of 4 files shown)


abbas-sabra-sonarsource:
Hello @duncanp-sonar and @rita-g-sonarsource,
May one of you review my PR? Short description:

- For the error messages this should fix three cases:
1. Unsupported project types, like the reported MakeFile-project
2. Unsupported file Types. This happens often when you add files to your project for navigation/editing without compiling them
3. Custom built files. These are source files that are built in a special way so we cannot load their options
I made sure the fix work on Vs2015/2017/2019. Of course, there might be more cases that I'm not aware of :D

- I also disabled request logging on PCH process since it adds a lot of noise when navigating unsupported files and it is not needed since the same `unsupported message` should appear in normal analysis

- For header files support, This is the change that we have discussed in our last meeting and that we have been dogfooding for a while

rita-g-sonarsource:
@abbas-sabra-sonarsource as we discussed offline, I've split this PR into 3 smaller PRs:
1. Error messages https://github.com/SonarSource/sonarlint-visualstudio/pull/1765
2. Header files https://github.com/SonarSource/sonarlint-visualstudio/pull/1767 (which relies on #1765)
3. PCH noise https://github.com/SonarSource/sonarlint-visualstudio/pull/1766 (which relies on #1765)

PR #1765, opened 2020-10-07
Improve error messages about files that cannot be analyzed

### LINKED ISSUE #1547: Improve error messages about files that cannot be analyzed

rita-g-sonarsource:
Fixes #1547

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
@@ -24,7 +24,6 @@ using System.Linq;
 using EnvDTE;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using Microsoft.VisualStudio.VCProjectEngine;
 using Moq;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Analysis;
@@ -63,7 +62,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
             // Arrange
             var loggerMock = new Mock<ILogger>();
 
-            var projectItemMock = CreateProjectItemWithProject("c:\\foo\\SingleFileISense\\xxx.vcxproj");
+            var projectItemMock = CreateMockProjectItem("c:\\foo\\SingleFileISense\\xxx.vcxproj");
             var rulesConfigProviderMock = new Mock<ICFamilyRulesConfigProvider>();
 
             // Act
@@ -82,7 +81,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
             // Arrange
             var loggerMock = new Mock<ILogger>();
 
-            var projectItemMock = CreateProjectItemWithProject("c:\\foo\\xxx.vcxproj");
+            var projectItemMock = CreateMockProjectItem("c:\\foo\\xxx.vcxproj");
             // Note: we want the exception to be thrown from inside the FileConfig::TryGet
             projectItemMock.Setup(x => x.Object).Throws(new InvalidOperationException("xxx"));
 
@@ -203,7 +202,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         public void IsFileInSolution_SingleFileIntelliSense_ReturnsFalse()
         {
             // Arrange
-            var projectItemMock = CreateProjectItemWithProject("c:\\foo\\SingleFileISense\\xxx.vcxproj");
+            var projectItemMock = CreateMockProjectItem("c:\\foo\\SingleFileISense\\xxx.vcxproj");
 
             // Act
             var result = CFamilyHelper.IsFileInSolution(projectItemMock.Object);
@@ -328,12 +327,24 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
                 .Setup(x => x.GetRulesConfiguration(It.IsAny<string>()))
                 .Returns(rulesConfig);
 
-            var projectItemMock = CreateProjectItemWithProject("c:\\foo\\file.cpp");
+            var projectItemMock = CreateMockProjectItem("c:\\foo\\file.cpp");
 
             var request = CFamilyHelper.CreateRequest(loggerMock.Object, projectItemMock.Object, "c:\\foo\\file.cpp",
                 rulesConfigProviderMock.Object, analyzerOptions);
 
             return request;
+        }
+
+        internal static void AssertMessageLogged(Mock<ILogger> loggerMock, string message)
+        {
+            loggerMock.Verify(x => x.WriteLine(It.Is<string>(
+                s => s.Equals(message))), Times.Once);
+        }
+
+        internal static void AssertPartialMessageLogged(Mock<ILogger> loggerMock, string message)
+        {
+            loggerMock.Verify(x => x.WriteLine(It.Is<string>(
+                s => s.Contains(message))), Times.Once);
         }
     }
 }
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
@@ -37,24 +37,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
     [TestClass]
     public class CFamilyHelperTests
     {
-        private const string ValidPlatformName = "Win32";
-
-        private static readonly IDictionary<string, string> MandatoryProjectConfigProperties = new Dictionary<string, string>
-        {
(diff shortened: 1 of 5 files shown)


src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -30,19 +30,36 @@ partial class CFamilyHelper
...
-
                 var vcConfig = vcProject.ActiveConfiguration;
+                var projectKind = vcConfig.ConfigurationType;
+                // Unknown projects represent all the unspported project type like makefile projects
+                if (projectKind == ConfigurationTypes.typeUnknown)
+                {
+                    logger.WriteLine(@"Project's ""Configuration type"" is not supported.");
+                    return null;
+                }
+                // "ClCompile" for source files and "ClCompile" for header files
```
rita-g-sonarsource, line 49: Is that a mistake in the comment? did you mean to say "ClInclude"?

src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -30,19 +30,36 @@ partial class CFamilyHelper
...
+                if (file.ItemType != "ClCompile")
+                {
+                    logger.WriteLine($"File's \"Item type\" is not supported. File: '{absoluteFilePath}'");
```
rita-g-sonarsource, line 52: Might be useful to output the item type in the log message.

src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -30,19 +30,36 @@ partial class CFamilyHelper
...
+                    return null;
+                }
                 var platformName = ((VCPlatform)vcConfig.Platform).Name; // "Win32" or "x64"
                 var vcFileConfig = file.GetFileConfigurationForProjectConfiguration(vcConfig);
-
+                // We don't support custom build tools. VCCLCompilerTool is needed for all the necessary compilation options to be present
+                if (!(vcFileConfig.Tool is VCCLCompilerTool))
+                {
+                    logger.WriteLine($"Custom built files are not supported. File: '{absoluteFilePath}'");
```
rita-g-sonarsource, line 60: I'm not a c++ dev so I don't know if this message is clear enough. Perhaps something like "File {0} is built using custom build tools. SonarLint supports only VCCLCompilerTool". 🤔

src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -30,19 +30,36 @@ partial class CFamilyHelper
...
             {
                 var vcProject = dteProjectItem.ContainingProject.Object as VCProject;
                 var file = dteProjectItem.Object as VCFile;
                 if (vcProject == null || file == null)
                 {
                     return null;
                 }
-
                 var vcConfig = vcProject.ActiveConfiguration;
+                var projectKind = vcConfig.ConfigurationType;
```
rita-g-sonarsource, line 42: This method is already quite long. I think it's worth extracting all the validation checks into a separate method.

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
```diff
@@ -205,6 +188,60 @@ public void TryGetConfig_ErrorsAreLogged()
             }
         }
 
+        [TestMethod]
```
rita-g-sonarsource, line 191: Is there a reason why these tests are here and not in FileConfigTests?

src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -30,19 +30,36 @@ partial class CFamilyHelper
     {
         internal class FileConfig
         {
-            public static FileConfig TryGet(ProjectItem dteProjectItem, string absoluteFilePath)
+            public static FileConfig TryGet(ILogger logger, ProjectItem dteProjectItem, string absoluteFilePath)
             {
                 var vcProject = dteProjectItem.ContainingProject.Object as VCProject;
                 var file = dteProjectItem.Object as VCFile;
                 if (vcProject == null || file == null)
```
rita-g-sonarsource, line 37: I think it's worth logging this failure as well.

rita-g-sonarsource (commented):
@abbas-sabra-sonarsource Please see my comments

src/Integration.Vsix/CFamily/FileConfig.cs
abbas-sabra-sonarsource: yes it is

src/Integration.Vsix/CFamily/FileConfig.cs
abbas-sabra-sonarsource: The "Item type" is the field in the file option. From my quick look, I didn't find a way to get the exact string in the UI(There are many possibilities) and most of them are mapped to something generic or unknown. Printing the actual `file.ItemType` will not be understandable

src/Integration.Vsix/CFamily/FileConfig.cs
abbas-sabra-sonarsource: VCCLCompilerTool is an internal name that won't resonate with C++ dev.  I prefer when I can to keep the variable size name at the end to avoid bad messages when the absolute path is big. 
maybe:
Custom build tools aren't supported. Custom-built file: '{absoluteFilePath}'
What do you think?

src/Integration.Vsix/CFamily/FileConfig.cs
abbas-sabra-sonarsource: Yes, I believe it is. But I don't really know in which case functionally we can reach this point with VCProject and VCFile are null.
Have you faced it before?

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
abbas-sabra-sonarsource: They are all calling/testing  `CreateRequest` which is in `CfamilyHelper.cs`

src/Integration.Vsix/CFamily/FileConfig.cs
rita-g-sonarsource: Yea that can work 👍 but whatever you think will be most understandable to c++ devs.

src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs
rita-g-sonarsource: I understand, but they don't have to. The logic that was changed is in `FileConfig.TryGet`, so that's the method that should be tested.

src/Integration.Vsix/CFamily/FileConfig.cs
rita-g-sonarsource: Hmm, not sure. Perhaps a CMake file is not a VCProject? I'm not sure if it's here for defensive checks or if there was a real use case.

abbas-sabra-sonarsource: No, I already checked that:
- If you mean CMake project then it will return with: `Not expecting to be called for files that are not in the solution` as they don't have a project
- if you mean Cmake files then they will return early as the `IsCFamilyDocument` will fail

src/Integration.Vsix/CFamily/FileConfig.cs
```diff
@@ -77,8 +77,9 @@ public static FileConfig TryGet(ILogger logger, ProjectItem dteProjectItem, stri
                     // Not supported
                     return null;
                 }
-
-                var (vcProject, vcConfig, vcFileSettings) = loadedConfig.Value;
```
abbas-sabra-sonarsource: @rita-g-sonarsource I reverted this part and used the ugly old syntax as it failed to compile on the CI. Is there something that we can do about it?

abbas-sabra-sonarsource: Great, Thanks!
`we use PascalCase for method names and properties`
Maybe it is a good idea to enable: https://rules.sonarsource.com/csharp/tag/convention/RSPEC-100 ?

rita-g-sonarsource: Good idea! Done 👍

Issue #788, opened 2019-01-10
[Technical] Enable different logging levels


duncanp-sonar:
Currently the _ILogger_ interface in SLVS doesn't support different levels of verbosity.
This means it is not possible to log sufficient information to help diagnose issues, since that would overload the user with too much information in the output window in normal use.

### Related information
The daemon supports different levels of logging, configurable through the UI.

![image](https://user-images.githubusercontent.com/10757559/50960324-09f5f800-14bd-11e9-9a77-16f250966553.png)

### Related issues
#663: [Technical] Add logging in SonarQube.Client

TODO:
- [ ] Consider writing detailed logging / exception stack trace into an external file (would also be useful for troubleshooting, we could ask the users to provide us with the file)

duncanp-sonar:
https://github.com/SonarSource/sonarlint-visualstudio/issues/881 is an example of a bug for which verbose logging would be useful.

duncanp-sonar:
Closing as obsolete. Three levels of logging are now supported.

PR #1766, opened 2020-10-07
Avoid logging on PCH request to eliminate the noise while navigating unsupported files

### LINKED ISSUE #1770: Logging when navigating unsupported file types is too verbose
### Description

It is common for C++ projects to contain file types that cannot be analyzed. Navigating these produces a lot of output messages in the Output Pane that clutter the pane and make it difficult to see more important/relevant output.

#### Current behaviour
1) When an unsupported file is opened, two "unsupported file type" messages are logged: one from the PCH analysis request, and one for the "normal" analysis request.
2) When a previously opened file is activated again, another PCH request is created which logs another "unsupported file type" message.

It is the second behaviour that is generating the noise.

Ignoring the crash of the unsupported file as it will be fixed in another ticket, we see the same message for the same unsupported file:
image

#### Proposed behaviour
Don't log "unsupported" messages when creating PCH analysis requests, so there would be only only one "unsupported type" message logged in case (1) and none for case (2).

Related PR: #1766

Related ticket: #788 _Enable different logging levels_

rita-g-sonarsource:
Fixes #1770

### DIFF
--- a/src/Integration.Vsix/CFamily/CFamilyHelper.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyHelper.cs
@@ -50,6 +50,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
             {
                 // We can't analyze header files currently because we can't get all
                 // of the required configuration information
+                logger.WriteLine($"Cannot analyze header files. File: '{absoluteFilePath}'");
                 return null;
             }
 
@@ -193,7 +194,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
                 // Note: if the C++ tools are not installed then it's likely an exception will be thrown when
                 // the framework tries to JIT-compile the TryGet method (since it won't be able to find the MS.VS.VCProjectEngine
                 // types).
-                return FileConfig.TryGet(projectItem, absoluteFilePath);
+                return FileConfig.TryGet(logger, projectItem, absoluteFilePath);
             }
             catch (Exception e)
             {
--- a/src/Integration.Vsix/CFamily/CFamilyHelper.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyHelper.cs
@@ -44,26 +44,26 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
         private static readonly string analyzerExeFilePath = Path.Combine(
             CFamilyFilesDirectory, "subprocess.exe");
 
-        internal class PCHRequestLogger : ILogger
+        internal class NoOpLogger : ILogger
         {
             public void WriteLine(string message)
             {
-                // We don't log in PCH request to avoid noise while navigating files
             }
 
             public void WriteLine(string messageFormat, params object[] args)
             {
-                // We don't log in PCH request to avoid noise while navigating files
             }
         }
 
-        private static readonly PCHRequestLogger pchLogger = new PCHRequestLogger();
+        private static readonly NoOpLogger noOpLogger = new NoOpLogger();
 
         public static Request CreateRequest(ILogger logger, ProjectItem projectItem, string absoluteFilePath, ICFamilyRulesConfigProvider cFamilyRulesConfigProvider, IAnalyzerOptions analyzerOptions)
         {
             if (analyzerOptions is CFamilyAnalyzerOptions cFamilyAnalyzerOptions && cFamilyAnalyzerOptions.CreatePreCompiledHeaders)
             {
-                logger = pchLogger;
+                // In case the requeset is coming from PCH generation, we don't log failures.
+                // This is to avoid redundant messages while navigating unsupoported files.
+                logger = noOpLogger;
             }
 
             if (IsHeaderFile(absoluteFilePath))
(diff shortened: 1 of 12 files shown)


src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -44,8 +44,28 @@ internal static partial class CFamilyHelper
...
+            {
+                // We don't log in PCH request to avoid noise while navigating files
+            }
+        }
+
+        private static readonly PCHRequestLogger pchLogger = new PCHRequestLogger();
+
         public static Request CreateRequest(ILogger logger, ProjectItem projectItem, string absoluteFilePath, ICFamilyRulesConfigProvider cFamilyRulesConfigProvider, IAnalyzerOptions analyzerOptions)
         {
+            if (analyzerOptions is CFamilyAnalyzerOptions cFamilyAnalyzerOptions && cFamilyAnalyzerOptions.CreatePreCompiledHeaders)
```
rita-g-sonarsource, line 64: @abbas-sabra-sonarsource this only disables the new "not supported" notifications, right?
If so, I'm not sure if it really matters that the users sees them or not. The user is gonna see that we don't support it, and then just stop looking at the SonarLint pane, no? WDYT?

abbas-sabra-sonarsource: Yes, it is to disable not supported notifications. In most cases, there will be files that we don't support in the same projects. This usually makes my SonarLint pane Full of irrelevant information of files that I'm just navigating. I know they are not supported, I'm not triggering any analysis they still fill my pane

rita-g-sonarsource: Ok, I understand.
I think this issue should be handled differently. Overriding the logger feels wrong: it would block us from logging meaningful messages, such as real failures (this PR now actually hides one of them -- the try-catch inside TryGetConfig). 
I think the approach should be different, such as perhaps creating log levels and setting a different verbosity for PCH, instead of disabling it completely (https://github.com/SonarSource/sonarlint-visualstudio/issues/788). It's a much bigger change but I think if we want to control the output, that should be the way to go.
@duncanp-sonar what do you think?

abbas-sabra-sonarsource: `would block us from logging meaningful messages`
True, but **only while navigating** which sounded fine functionally. Changing the verbosity in the case of PCH would solve the problem too

duncanp-sonar: @abbas-sabra-sonarsource  I'm confused; by "navigating" do you mean "opening", in which case won't we still show the "not supported" when we run the analysis on file open anyway?

duncanp-sonar: FYI I've created ticket #1770 for this issue (we have an informal policy that any code changes should relate to a ticket). @abbas-sabra-sonarsource could you paste an example of verbose output in the ticket please? Thanks.

abbas-sabra-sonarsource: @duncanp-sonar by navigating I don't mean opening, I mean focusing on a file or jumping between already opened files.
Showing the "not supported" on a file open is fine. It is actually useful. The problem is having "file not supported" for the same file as many times as you focus on it

duncanp-sonar: Thanks @abbas-sabra-sonarsource, makes sense. If it's just the "unsupported type" message we want to avoid for PCH requests then I think it would be better to conditionally suppress the logging in `TryGetConfig` for PCH requests, rather than disabling PCH logging completely.

abbas-sabra-sonarsource: @duncanp-sonar,  I'm not disabling logging for PCH completely, I'm only disabling it in fetching the Request. Yes, the only concern is "unsupported type". Currently, all the logging done in CreateRequest is related to unsupported types. In principle, I think if CreateRequest is logging something, it must be about "request failure" => "unsupported type" => you don't want to see the message each time you focus on the file. 
logging through a function that checks if it is a PCH request before logging would work, but I thought that the overriding approach is nicer since it is future proof. I expect that we always want to hide logs that will be added later.

For example, the log mentioned by @rita-g-sonarsource : `Unable to collect C/C++ configuration for..`
This log implies that the file type isn't supported and we want to see it only when we analyze the file **instead of** every time we focus on it.

Let me know what do you think and if you still prefer the conditional suppressing.

duncanp-sonar: Ok, makes sense. In that case, I don't have a preference between the two approaches, although if you use the no-op logger approach I think it should be renamed. `PCHRequestLogger` suggests that it is used for the whole PCH request, not just when collecting the configuration. Maybe something like "NoOpLogger" instead, with an appropriate comment at line 66 in place of the two comments in line 51 and 56.

abbas-sabra-sonarsource: Great! I'll update the PR

abbas-sabra-sonarsource:
fixes #1770

PR #1774, opened 2020-10-08
Output FileConfig to allow easier troubleshooting

### LINKED ISSUE #1769: Output FileConfig to allow easier troubleshooting
#### Description

When users experience C/C++ analysis failures, it is difficult to understand what is the problem.
A current way is to ask them for a reproducer file, however sometimes that is not enough, e.g. if the problem is on SonarLint's side in the way that we collect the configuration (for example: #1630, #1537).
Providing a way to output the collected configuration (FileConfig) would help to troubleshoot these problems.

TODO: 
- [x] check the FileConfig class can be serialized easily
- [x] change the "CreateReproducer" command to also serialize the FileConfig in the same location as the reproducer file.
- [x] update the wiki documentation: https://github.com/SonarSource/sonarlint-visualstudio/wiki/Troubleshooting-problems-with-C-or-Cpp-analysis

rita-g-sonarsource:
Fixes #1769

image


src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -35,7 +37,12 @@ internal static partial class CFamilyHelper
         public const string CPP_LANGUAGE_KEY = "cpp";
         public const string C_LANGUAGE_KEY = "c";
 
-        internal static readonly string PchFilePath = Path.Combine(Path.GetTempPath(), "SonarLintForVisualStudio.PCH.preamble");
+        internal static string WorkingDirectory { get; } = Path.GetTempPath();
```
rita-g-sonarsource, line 40: Cosmetics, we kept repeating `Path.GetTempPath()` everywhere.

src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -35,7 +37,12 @@ internal static partial class CFamilyHelper
...
+        internal static readonly string PchFilePath = Path.Combine(WorkingDirectory, "SonarLintForVisualStudio.PCH.preamble");
+        internal static readonly string RequestConfigFilePath = Path.Combine(WorkingDirectory, "sonar-cfamily.request.config");
```
rita-g-sonarsource, line 42: @abbas-sabra-sonarsource @duncanp-sonar do you have a suggestion for a better name than `sonar-cfamily.request.config`?

duncanp-sonar: I think both files should start with the same prefix - it's simpler to ask a user to "send us the two files starting with X" than to have two completely unrelated file names. Also, the name should make it clear that the file relates to SLVS.

Maybe `SonarLintForVisualStudio.cfamily.request.config`?

src/Integration.Vsix/CFamily/CFamilyHelper.cs
duncanp-sonar: Strictly speaking you've changed the behaviour so the temp path is stored the first time it is used and never updated.
`internal static string WorkingDirectory => Path.GetTempPath();`
...would fetch the current temp path each time it is needed.

src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -35,7 +37,12 @@ internal static partial class CFamilyHelper
...
+        internal static readonly string ReproducerFilePath = Path.Combine(WorkingDirectory, "sonar-cfamily.reproducer");
```
duncanp-sonar, line 43: And the two file paths probably shouldn't be readonly.

duncanp-sonar (approved):
Just a couple of comments about read-only-ness,

src/Integration.Vsix/CFamily/CFamilyHelper.cs
rita-g-sonarsource: Yes I thought that was ok, we don't really need it to be calculated each time. It's unlikely that the temp file would change, especially not mid-VS session.

src/Integration.Vsix/CFamily/CFamilyHelper.cs
rita-g-sonarsource: I think you're looking at the PCH file (`SonarLintForVisualStudio.PCH.preamble`) - the reproducer is named `sonar-cfamily.reproducer`. So what do you think? 😕

src/Integration.Vsix/CFamily/CFamilyHelper.cs
rita-g-sonarsource: Why not? 🤔 I don't think it's likely for the temp file to change mid VS-session (or at all)

src/Integration.Vsix/CFamily/CFamilyHelper.cs
abbas-sabra-sonarsource: the name isn't that bad. I think as long as the name starts with the same prefix it is fine(`sonar-cfamily`).
Maybe other suggestion `sonar-cfamily.request.reproducer` as we are trying to reproduce the request. No strong opinion.

src/Integration.Vsix/CFamily/CFamilyHelper.cs
duncanp-sonar: Unlikely but not impossible. The implementation effort is the same so there doesn't seem to be a good reason to choose the less dynamic option.

BTW adding a review comment saying a change is cosmetic is a bit misleading if you've deliberately changed the behaviour!

rita-g-sonarsource: You're right, I didn't think of it a behavior change, sorry 🙈

src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -196,7 +195,7 @@ private static void SaveFileConfig(FileConfig fileConfig, ILogger logger)
             runner.Execute(args);
         }
   
-        internal static FileConfig TryGetConfig(ILogger logger, ProjectItem projectItem, string absoluteFilePath)
+        internal static FileConfig TryGetConfig(ILogger logger, ProjectItem projectItem, string absoluteFilePath, IAnalyzerOptions analyzerOptions)
```
rita-g-sonarsource, line 198: @duncanp-sonar I realized that I might be logging the fileConfig a bit too late. In the previous version, it was possible that Request was null due to some other failure, and so we wouldn't have reached a point where we log the file config.
So now I log it directly after we get it.

PR #1776, opened 2020-10-09
Throw exception on unsupported options

### LINKED ISSUE #1775: Log additional unsupported C++ configurations
#### Description

There are configurations that we don't support, yet we don't log that we don't support them. This makes troubleshooting users' analysis failures very difficult.

TODO:
- [x] Log unsupported "/ZW" "/clr" (https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs#L124)
- [x] Log unsupported "/Tc" (https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs#L231)
- [ ] Log `Absolute` method usages ? 
- [x] For troubleshooting: log the number of collected files?

rita-g-sonarsource:
Fixes #1775


src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs
```diff
@@ -228,17 +226,9 @@ public static Request ToRequest(CFamilyHelper.Capture[] captures)
...
                     }
-                    else if (arg.StartsWith("/Tc"))
-                    {
-                        // TODO completely skipped for now - used in ReactOS only for compilation of resources
-                        files.Clear();
-                        break;
-
-                    }
-                    else if (arg.StartsWith("/Tp"))
+                    else if (arg.StartsWith("/Tc") || arg.StartsWith("/Tp"))
```
abbas-sabra-sonarsource, line 229: As discussed, It is safe to treat `/Tc` like `/TC` and `/Tp` like `/TP`

rita-g-sonarsource: Opened an issue: https://github.com/SonarSource/sonarlint-visualstudio/issues/1778

src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs
```diff
@@ -436,9 +426,17 @@ public static Request ToRequest(CFamilyHelper.Capture[] captures)
                   // TODO can be passed directly to Clang as PreprocessorOptions.Includes
                   + includes.ToString();
                 request.TargetTriple = probe.Isx64 ? "x86_64-pc-windows" : "i686-pc-windows";
-                if (files.Count == 1)
+
+             
+                switch (files.Count)
```
abbas-sabra-sonarsource, line 431: I think it makes sense to move this part to the end of the switch. As `files` isn't modified after the switch => avoid executing all the previous code for nothing

rita-g-sonarsource: Not sure I follow - can you clarify please?

abbas-sabra-sonarsource: Checking the `files` count is now happening at the end of the function `ToRequest` for no real reason. To fail early you can move it just after the switch (L271)

abbas-sabra-sonarsource (approved):

Issue #1783, opened 2020-10-10
Simplify the MsvcDriver logic by using compilation option directly instead of reparsing them


abbas-sabra-sonarsource:
We are currently building the command line from the loaded compilation options to mimic the sonar-scanner driver.
This is not needed in the SLVS context since options are already parsed: no need to go through the MsvcDriver while loop.
For example, instead of adding the `AbsoluteFilePath `to `Cmd` : `c.Cmd.Add(fileConfig.AbsoluteFilePath);` and then rematching the filename is `ToRequest`. We can directly add it to "request.File =  AbsoluteFilePath" This is way less error-prone and it simplifies the big while loop in the MsvcDriver.
Note: this applies for all options but `AdditionalOptions`. To be checked if there is a way to access its already parsed content.

PR #1785, opened 2020-10-11
Fix request with no file bug by ignoring the argument of ignored option

### LINKED ISSUE #1784: Analysis fails when an ignored option has an argument
### Repro steps

In your C++ project, use an ignored option that requires an argument. This can be done either from the **project** or from the **file**:
image
In this case `/AI` will be ignored by the driver but "C:\Users\Abbas\Desktop" will be interpreted as a file to be analyzed. This leads to an empty request generation.

### Expected behavior
The file is analyzed successfully.

### Actual behavior

The analysis fails with `no such file or directory` message

abbas-sabra-sonarsource:
Fixes #1784

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
@@ -165,24 +165,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         }
 
         [TestMethod]
-        public void MoreThanOneAnalyzedFile_InvalidOperationException()
-        {
-            Action action = () => MsvcDriver.ToRequest(new[] {
-                compiler,
-                new CFamilyHelper.Capture()
-                {
-                    Executable = "",
-                    Cwd = "basePath",
-                    Env = new List<string>(),
-                    Cmd = new List<string> { "cl.exe", "c:\\file1.cpp", "c:\\file2.cpp" },
-                }
-            });
-
-            
-            action.Should().ThrowExactly<InvalidOperationException>().And.Message.Should().StartWith("Cannot analyze more than 1 file");
-        }
-
-        [TestMethod]
         public void Include_Directories()
         {
             Request req = MsvcDriver.ToRequest(new CFamilyHelper.Capture[] {
@@ -291,6 +273,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
                         "/I", "bar",
                         "/I", "subdir2\\relativePath1.txt", // \ should be converted to /
                         "/I", "c:\\absPath1",                // Absolute path should not be changed
+                        "c:\\file.cpp"
                     }
                 }
             });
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs
```diff
@@ -256,10 +256,15 @@ public static Request ToRequest(CFamilyHelper.Capture[] captures)
                     }
                     else
                     {
-                        string file = Absolute(capture.Cwd, arg);
-                        if (file != null)
+                        // We know that the file to be analyzed is the last element.
+                        // If it is not the last element, it is the argument of an ignored option and it shouldn't be treated as a file
+                        if (args.Index == (capture.Cmd.Count - 1))
```
rita-g-sonarsource, line 261: I suggest a few changes:
1. Make the check after we leave the while loop (so then `args.Index == (capture.Cmd.Count - 1)` will always be true), instead of checking it for every iteration
2. Perhaps we should log the "else" and what option was it that fell through this if-else block.
3. There's no reason to use the list anymore. This code and the one below can be simplified to 

```
while (args.Index < capture.Cmd.Count) {..}

var file = Absolute(capture.Cwd, capture.Cmd[args.Index-1]);

if (file == null)
{
    throw new InvalidOperationException("No files to analyze");
}

request.File = file;

if (cpp == null)
{
    cpp = request.File.EndsWith(".c");
}
```

src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
```diff
@@ -682,6 +665,27 @@ public void unsupported_std_version()
             });
         }
 
+        [TestMethod]
+        public void unsupported_option_with_argument()
```
rita-g-sonarsource, line 669: I'm not sure that this is the behavior that we want.
From a user's standpoint, wouldn't you rather get an exception and fix your configuration? rather than us randomly picking the last item.
What if a user did something like `/AI c:\\file.cpp a`?
I think it would be better to just tell him that we've identified multiple files and leave the fix to him.

rita-g-sonarsource:
@abbas-sabra-sonarsource I'm not sure that we should fix this. The user can do something like `/AI c:\\file.cpp a` by mistake or by not understanding, and we would attempt to analyze the wrong thing. I think it's better to just say that we've identified multiple files and leave the fix to the user.

src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
abbas-sabra-sonarsource: The user cannot do that, we are building the command and we are explicitly putting the saved file as the last element. Look at `ToCaptures`. We want to analyze the saved file and the saved file is always the last element.

src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs
abbas-sabra-sonarsource: 1. The last branch will very rarely be hit (Only in case of this bug). the one that will be hit the most is:
```
                    else if (arg.StartsWith("/"))
                    {
                        // skip
                        args.Index++;

                    }
```
While putting the check outside the while will be cleaner, it will go against the functional logic that we are extracting all arguments from this switch. In fact, most of the branches can and should be extracted at some point that why I created [this](https://github.com/SonarSource/sonarlint-visualstudio/issues/1783).

2- No as we don't log the option that we ignore. We ignore them because they don't impact the analysis. I agree that we should log them if they aren't supported. like we did for `/ZW`

3- Yes, that makes sense

Issue #1790, opened 2020-10-14
Create new hotspots list


rita-g-sonarsource:
#### Description

Create a new tool window that will host the hotspots.

- [ ] New toolwindow
- [ ] New command (under Tools -> Other WIndows)
- [ ] TBD: Which vstc?

PR #1805, opened 2020-10-15
Add initial tool window

rita-g-sonarsource:
Part of #1790


src/IssueViz.UnitTests/Helpers/ToolWindowCommandTests.cs
```diff
@@ -0,0 +1,139 @@
...
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using Moq.Protected;
+using SonarLint.VisualStudio.Integration;
+using ThreadHelper = SonarLint.VisualStudio.Integration.UnitTests.ThreadHelper;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Helpers
+{
+    public abstract class ToolWindowCommandTests<T> where T : ToolWindowPane
```
rita-g-sonarsource, line 33: The tests (and the code) of hotspot command were identical to the existing issue viz tests, so I extracted them to a common suite.

rita-g-sonarsource:
@duncanp-sonar this PR creates a blank toolwindow under View -> Other Windows.
In the next PR I'll turn it into IWpfTableControl.

src/IssueViz/IssueVizPackage.cs
```diff
@@ -37,6 +39,7 @@ namespace SonarLint.VisualStudio.IssueVisualization
     [ProvideMenuResource("Menus.ctmenu", 1)]
     [ProvideToolWindow(typeof(IssueVisualizationToolWindow), MultiInstances = false, Style = VsDockStyle.Tabbed, Window = ToolWindowGuids.SolutionExplorer, Width = 325, Height = 400)]
     [ProvideToolWindowVisibility(typeof(IssueVisualizationToolWindow), Constants.UIContextGuid)]
+    [ProvideToolWindow(typeof(HotspotsToolWindow), MultiInstances = false, Style = VsDockStyle.Tabbed, Window = ToolWindowGuids.SolutionExplorer, Width = 325, Height = 400)]
```
duncanp-sonar, line 42: I think adding the commands to this package is fine.
We should have a chat about whether we want to create a separate package for the http listener package or not.

src/IssueViz.UnitTests/Hotspots/HotspotsToolWindowCommandTests.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            act = () => new HotspotsToolWindowCommand(package, null, logger);
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("commandService");
+
+            act = () => new HotspotsToolWindowCommand(package, commandService, null);
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("logger");
+        }
+
+        [TestMethod]
+        public void Ctor_CommandAddedToMenu()
```
duncanp-sonar, line 67: I'd have thought that this would have been a common test too.

duncanp-sonar (approved):

src/IssueViz.UnitTests/Hotspots/HotspotsToolWindowCommandTests.cs
rita-g-sonarsource: IssueViz adds multiple commands so I left them separated. I could pass to base CommandSet + list of commands ids, WDYT?

duncanp-sonar: That's the sort of thing I had in mind, or even just having a common assertion method like `CheckCommandRegistered(guid, id)`, rather than repeating the `commandService.Verify` code in each test.

rita-g-sonarsource: Ok, this PR is merged so I'll do that in separate PR

PR #1811, opened 2020-10-16
Add versioned Microsoft.VisualStudio.Shell

rita-g-sonarsource:


src/SonarLint.VSSpecificAssemblies.props
```diff
@@ -155,4 +155,36 @@
 
   </ItemGroup>
 
+  <!-- ******************************************************************************* -->
```
rita-g-sonarsource, line 158: The final solution is a bit tricky.

1. We use shell 14 everywhere, and by importing IssueViz into IssueViz.Security, IssueViz.Security also has a "built-in" reference to shell 14, no matter which VS version it's built on.

Here I re-add shell 14, this time with an alias `shell`. I do the same for shell 15.
This allows me to force import `IWpfTableControlProvider` from the right shell assembly by doing:
```
extern alias shell;

using IWpfTableControlProvider = shell::Microsoft.Internal.VisualStudio.Shell.TableControl.IWpfTableControlProvider;
```
... while all the other classes from shell, like `Microsoft.VisualStudio.Shell.PackageRegistrationAttribute`, just get imported from shell 14 via IssueViz project reference. If IssueViz project reference is removed, the compilation fails due to missing Shell dll -- which either needs to be directly imported, or used via `extern alias`.
Hence why I decided to directly import it in IssueViz.Security.csproj, instead of relying on the IssueViz dependency.

2. Unfortunately, shell 15 brings with it `Microsoft.VisualStudio.Shell.Framework`, which re-declares the class `ServiceProgressData` (from package InitializeAsync method), that already exists in `Microsoft.VisualStudio.Shell.Immutable.14.0`. So in order to "silence" this conflict, I add an alias to `Microsoft.VisualStudio.Shell.Framework`. This way it cannot be used without the alias, there is no conflict, and `ServiceProgressData` is taken from Shell.Immutable.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -109,6 +110,7 @@
     <PackageReference Include="Microsoft.VisualStudio.Text.Data" Version="14.0.23205" />
     <PackageReference Include="Microsoft.VisualStudio.Text.Logic" Version="14.0.23205" />
     <PackageReference Include="Microsoft.VisualStudio.Text.UI" Version="14.0.23205" />
+    <PackageReference Include="Microsoft.VisualStudio.Threading" Version="14.0.50702" />
```
rita-g-sonarsource, line 113: This package was not directly imported before and instead was taken from Shell. If I don't directly import it, it will be taken from Shell 15, so I preferred to explicitly import the 14 version.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -62,6 +62,7 @@
     <RootNamespace>SonarLint.VisualStudio.Integration.Vsix</RootNamespace>
     <AssemblyName>SonarLint.$(VersionSpecificSuffix)</AssemblyName>
     <RequiresSigning>true</RequiresSigning>
+    <RequiresShell>true</RequiresShell>
```
rita-g-sonarsource, line 65: When IssueViz.Security dependency is added to the Vsix project, it imports Shell 15 without an alias. Since Vsix already contains shell 14, the entire project is broken due to ambiguous assembly references.
So I "re-import" shell 15 with an alias as a way to silence the ambiguity.

PR #1814, opened 2020-10-19
Add versioned Microsoft.VisualStudio.Shell

### LINKED ISSUE #1809: Create assembly for security hotspots
#### Description

In order to display hotspots in a native-VS TableControl, we need to use an internal MS interface `Microsoft.Internal.VisualStudio.Shell.TableControl.IWpfTableControlProvider`. To minimize potential additional references to internal interfaces, it's best to place the hotspots control in a separate assembly.
In addition, `IWpfTableControlProvider` requires different versions of `Microsoft.VisualStudio.Shell` - VS 2015 requires shell 14, whereas VS 2017/2019 require shell 15, so the assembly needs to be built against the relevant shell version.

TODO:
- [x] Create new assembly `SonarLint.VisualStudio.IssueVisualization.Security`
- [x] Create test project
- [x] Conditionally reference `Microsoft.VisualStudio.Shell` of the relevant version.

rita-g-sonarsource:
Fixes #1809


src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -245,7 +245,7 @@ Relative path: 'SonarLint.VisualStudio.IssueVisualization.Security.dll'
 Referenced assemblies:
 - 'Microsoft.VisualStudio.Shell.14.0, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
 - 'Microsoft.VisualStudio.Shell.Immutable.14.0, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
-- 'Microsoft.VisualStudio.Threading, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
+- 'Microsoft.VisualStudio.Threading, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```
rita-g-sonarsource, line 248: This is coming from shell 15 dependency.
I tried to manually add `Microsoft.VisualStudio.Threading` version 14 + mark version 15 with an alias, but it still doesn't work. VS complains about package downgrade and won't let me add version 14.

src/IssueViz.Security/IssueViz.Security.csproj
```diff
@@ -8,6 +8,15 @@
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!-- Hack to prevent VS2017 from automatically attempting to upgrade the project -->
+    <MinimumVisualStudioVersion>$(VisualStudioVersion)</MinimumVisualStudioVersion>
+    <RequiresShell>true</RequiresShell>
```
duncanp-sonar, line 14: This name is a bit misleading; `RequiresVersionSpecifiShell` would be better.

src/IssueViz.Security/IssueViz.Security.csproj
```diff
@@ -36,6 +45,7 @@
   <ItemGroup Label="Package references">
     <PackageReference Include="Microsoft.VisualStudio.ComponentModelHost" Version="14.0.25424" />
     <PackageReference Include="Microsoft.VisualStudio.Language.Intellisense" Version="14.0.23205" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.14.0" Version="14.3.25407" />
```
duncanp-sonar, line 48: I think this should be in the props file too, rather than splitting the logic between the two files.

src/SonarLint.VSSpecificAssemblies.props
```diff
@@ -152,7 +152,34 @@
...
+  <ItemGroup Label="Shell versioned references" Condition="$(RequiresShell)=='true' AND '$(VsTargetVersion)'!='2015'">
+    <PackageReference Include="Microsoft.VisualStudio.Shell.15.0"
+                      Version="15.0.26201"
+                      GeneratePathProperty="true"
+                      ExcludeAssets="all"
+                      PrivateAssets="all">
+    </PackageReference>
+    <ReferencePath Include="$(NuGetPackageRoot)\Microsoft.VisualStudio.Shell.15.0\15.0.26201\lib\Microsoft.VisualStudio.Shell.15.0.dll">
+      <Private>$(CopyVSAssembliesToOutput)</Private>
+      <Aliases>shell</Aliases>
```
duncanp-sonar, line 181: I think it would be clearer if the alias name was more descriptive e.g. `versionSpecificShell`.

duncanp-sonar (approved):
Some comments about naming and splitting the functionality across files.

src/IssueViz.Security/IssueViz.Security.csproj
rita-g-sonarsource: It's not exactly the same. The props file is when you know you need a VS specific version of shell, and it's an addition to your existing references. Here the shell is the basic shell for the entire project. If we remove the WpfTableControlProvider, we only need to remove the RequiresShell flag, and everything else would be the same.
Also, there are additional shell references, like shell interop and shell immutable, which we do/will specify directly. So to me it makes sense to have the original shell right next to them. WDYT?

PR #1817, opened 2020-10-19
[1] Move hotspots from IssueViz to IssueViz.Security

rita-g-sonarsource:
Moved hotspots code into IssueViz.Security.


src/IssueViz.Security/Commands/IssueVizSecurityPackage.vsct
```diff
@@ -19,14 +19,14 @@
     </Buttons>
 
     <Bitmaps>
-      <Bitmap guid="guidImages" href="..\..\Resources\sonarlint.png"/>
+      <Bitmap guid="guidImages" href=".\Resources\sonarlint.png"/>
```
rita-g-sonarsource, line 22: The resources folder is now under Commands

src/IssueViz/Commands/IssueVizPackage.vsct
```diff
@@ -63,7 +61,7 @@
     </Buttons>
 
     <Bitmaps>
-      <Bitmap guid="guidImages" href="..\Resources\sonarlint.png"/>
+      <Bitmap guid="guidImages" href=".\Resources\sonarlint.png"/>
```
rita-g-sonarsource, line 64: The resources folder is now under Commands. I moved it outside of Commands when hotspots were in this project as well, so now I could return it back.

src/TestInfrastructure/ToolWindowCommandTests.cs
```diff
@@ -0,0 +1,137 @@
...
+using FluentAssertions;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using Moq.Protected;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests
+{
+    public abstract class ToolWindowCommandTests<T> where T : ToolWindowPane
```
rita-g-sonarsource, line 31: This used to be in the IssueViz.UnitTests.
I'm a bit on the fence if this test suite should even exist.

duncanp-sonar: Why? Is it because you think none of the tests are useful?

duncanp-sonar (approved):

src/TestInfrastructure/ToolWindowCommandTests.cs
rita-g-sonarsource: I think it adds complexity to the tests. The alternative is repetition but in a test context it might be more readable and easier to maintain. I'll keep the suite for now but I'm on the fence about its usefulness in the long run.

PR #1819, opened 2020-10-20
[3] Extract command test to common suite

rita-g-sonarsource:
Fix for previous PR comment


src/IssueViz.Security.UnitTests/Commands/HotspotsToolWindowCommandTests.cs
```diff
@@ -34,15 +35,19 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Commands
     [TestClass]
     public class HotspotsToolWindowCommandTests : ToolWindowCommandTests<HotspotsToolWindow>
     {
-        public HotspotsToolWindowCommandTests() 
-            : base(ExecuteCommand)
+        public HotspotsToolWindowCommandTests()
+            : base(ExecuteCommand, CreateCommand,
+                new Dictionary<Guid, IEnumerable<int>>
```
duncanp-sonar, line 40: Why a `Dictionary`? Each tool window only supports commands from a single command set, don't they?

src/TestInfrastructure/ToolWindowCommandTests.cs
```diff
@@ -31,13 +33,19 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
...
+        private readonly Func<IMenuCommandService, object> createCommand;
+        private readonly IDictionary<Guid, IEnumerable<int>> commandsInCommandSet;
 
         private Mock<AsyncPackage> package;
         private Mock<ILogger> logger;
 
-        protected ToolWindowCommandTests(Action<AsyncPackage, ILogger> executeCommand)
+        protected ToolWindowCommandTests(Action<AsyncPackage, ILogger> executeCommand, 
+            Func<IMenuCommandService, object> createCommand, 
+            IDictionary<Guid, IEnumerable<int>> commandsInCommandSet)
```
duncanp-sonar, line 44: Passing multiple delegates into the constructor is pretty ugly in general, and seems a particularly odd thing to do when it's an abstract class. Abstract methods/properties would look cleaner.

duncanp-sonar (approved):

Issue #1791, opened 2020-10-14
Add new package with HttpListener


rita-g-sonarsource:
#### Description

Create a new package under IssueViz that starts an HttpListener on a predefined port.

- [x] The package should load early (on shell initialization)
- [x] Cycle through a range of predefined ports and find an available port

PR #1820, opened 2020-10-20
Added Open in IDE package and http listener

duncanp-sonar:
Relates to #1791


src/IssueViz.Security/OpenInIDE/OpenInIDEPackage.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        private ILogger logger;
+        private HttpListener listener;
+
+        protected override async System.Threading.Tasks.Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            IComponentModel componentModel = await GetServiceAsync(typeof(SComponentModel)) as IComponentModel;
+            logger = componentModel.GetExtensions<ILogger>().First();
+
+            IListenerFactory listenerFactory = new ListenerFactory(logger);
```
duncanp-sonar, line 55: I thought about having the factory return an interface, but I've decided to keep it as simple as possible in this first commit.

src/IssueViz.Security/OpenInIDE/OpenInIDEPackage.cs
```diff
@@ -0,0 +1,67 @@
...
+        private const int EndPort = 64123;
+
+        private HttpListener listener;
+
+        protected override async System.Threading.Tasks.Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            IComponentModel componentModel = await GetServiceAsync(typeof(SComponentModel)) as IComponentModel;
+            var logger = componentModel.GetExtensions<ILogger>().First();
+
+            IListenerFactory listenerFactory = new ListenerFactory(logger);
```
rita-g-sonarsource, line 54: Why not MEF-import IListenerFactory? It can import ILogger in the CTOR (also, why GetExtensions? we use GetService<ILogger> in the other places, e.g. https://github.com/SonarSource/sonarlint-visualstudio/blob/branch-hotspots/src/IssueViz.Security/Commands/HotspotsToolWindowCommand.cs#L64)

src/IssueViz.Security/IssueViz.Security.csproj
```diff
@@ -60,13 +60,24 @@
...
+      <DependentUpon>OpenInIDEResources.resx</DependentUpon>
+    </Compile>
 
     <Compile Update="Resources.Designer.cs">
       <DependentUpon>Resources.resx</DependentUpon>
       <DesignTime>True</DesignTime>
       <AutoGen>True</AutoGen>
     </Compile>
 
+    <EmbeddedResource Update="OpenInIDE\OpenInIDEResources.resx">
```
rita-g-sonarsource, line 76: Is there a reason to create a different resource file? I'm not against, just wondering what's our convention.

src/IssueViz.Security.UnitTests/ListenerFactoryTests.cs
```diff
@@ -0,0 +1,146 @@
...
+        private static int[] CreateRange (int start, int end)
+        {
+            if (start > end)
+            {
+                return Array.Empty<int>();
+            }
+            return Enumerable.Range(start, end - start + 1).ToArray();
+        }
+
+        private static bool AreAllPortsAvailable(params int[] ports) =>
```
rita-g-sonarsource, line 91: Cosmetics, maybe `VerifyTestPortsAreAvailable`? I can't put my finger on why but `AreAllPortsAvailable` is somehow harder to understand for me, not sure why 🤔

src/IssueViz.Security.UnitTests/ListenerFactoryTests.cs
```diff
@@ -0,0 +1,146 @@
...
+            return true;
+        }
+
+        private static string GetPrefix(int port) => $"http://localhost:{port}/";
+
+        /// <summary>
+        /// Helper class to listen on the specified port(s) until the
+        /// object is disposed. Throws if any of the ports are not free.
+        /// </summary>
+        private sealed class PortGrabber : IDisposable
```
rita-g-sonarsource, line 117: I'm a bit worried that this might be fragile. It's more of an integration test than a unit, and it depends on the machine that the test is running on.
If it causes problems, then it will be worth wrapping HttpListener with our own class, and mock the Start method to either fail/succeed, so we won't need to actually grab the ports on the machine.

rita-g-sonarsource (approved):

src/IssueViz.Security.UnitTests/ListenerFactoryTests.cs
duncanp-sonar: The method is being called twice, first to check that all of the ports being used in the test are available, and then to check that the ports above the free port are still available.
I originally had a method with a name like `VerifyTestPortsAreAvailable`; it worked ok for the first case but didn't fit for the second.
Also, I deliberately moved the `Verify`/`Check` part out of the method as I want to take different actions. If all of the required ports are not free at the start then the test result should be Inconclusive rather than a failure. If the remaining ports aren't free after the listener is created then the test should fail.

src/IssueViz.Security/IssueViz.Security.csproj
duncanp-sonar: There's no harm in having multiple, and it makes it easier to move/refactor the code. If you have one resource for the assembly it can get very big and it's more awkward to work with (which is what happened with the Integration and Integration.VSIX projects).

src/IssueViz.Security/OpenInIDE/OpenInIDEPackage.cs
duncanp-sonar: I copied from the implementation of GetMefService, but on looking at it again there's a comment that says it's using GetExtensions so it doesn't fail if there aren't any matching imports. We don't care about that here. I'll change it.

MEF importing the factory: I'll see how the code evolves. I didn't think it was worth worrying about at this stage.

src/IssueViz.Security.UnitTests/ListenerFactoryTests.cs
duncanp-sonar: I considered wrapping the listener, or looking for an existing NuGet package that does it, but decided not to do it at this stage. I suspect it might become necessary as we add features.

Issue #1807, opened 2020-10-15
Create IWpfTableControl for hotspots list


rita-g-sonarsource:
#### Description

Hook hotspots list into IWpfTableControl with predefined columns (project, file, line, column) and empty ITableDataSource.

PR #1822, opened 2020-10-21
Import only shell 14 or 15 in IssueViz.Security

rita-g-sonarsource:
Part of #1807


src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithoutStrongNames.txt
```diff
@@ -244,8 +244,8 @@ Relative path: 'SonarLint.VisualStudio.IssueVisualization.Security.dll'
 
 Referenced assemblies:
 - 'Microsoft.VisualStudio.ComponentModelHost, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
-- 'Microsoft.VisualStudio.Shell.14.0, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
-- 'Microsoft.VisualStudio.Shell.Immutable.14.0, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
+- 'Microsoft.VisualStudio.Shell.15.0, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
+- 'Microsoft.VisualStudio.Shell.Framework, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```
rita-g-sonarsource, line 248: ShellFramework is the equivalent of Shell.Immutable in the shell.15 version.

src/IssueViz.Security.UnitTests/Commands/HotspotsToolWindowCommandTests.cs
```diff
@@ -19,32 +19,33 @@
...
-using SonarLint.VisualStudio.Integration.UnitTests;
 using SonarLint.VisualStudio.IssueVisualization.Security.Commands;
 using SonarLint.VisualStudio.IssueVisualization.Security.HotspotsControl;
+using ThreadHelper = SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Helpers.ThreadHelper;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Commands
 {
     [TestClass]
-    public class HotspotsToolWindowCommandTests : ToolWindowCommandTests<HotspotsToolWindow>
+    public class HotspotsToolWindowCommandTests
```
rita-g-sonarsource, line 37: I ended up deleting the common suite.
In order to use it, I would've had to add the shell-specific tricks to TestsInfrastructure, which spread to the other test projects. It wasn't worth it IMO.

src/IssueViz.Security.UnitTests/Helpers/ThreadHelper.cs
```diff
@@ -0,0 +1,17 @@
+﻿using FluentAssertions;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Helpers
+{
+    internal class ThreadHelper
```
rita-g-sonarsource, line 5: The alternative was adding shell tricks to TestsInfrastructure and the other test projects.

src/IssueViz.Security/HotspotsControl/HotspotsControl.xaml
```diff
@@ -1,10 +1,8 @@
 ﻿<UserControl x:Class="SonarLint.VisualStudio.IssueVisualization.Security.HotspotsControl.HotspotsControl"
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-             xmlns:vsShell="clr-namespace:Microsoft.VisualStudio.Shell;assembly=Microsoft.VisualStudio.Shell.14.0"
```
rita-g-sonarsource, line 4: We don't need UI themes yet but sadly that's gonna be another problem in the future. We might be able to solve it with extern alias, but I'm not sure that works for xaml. We'll see when we get to it.

src/IssueViz.Security/IssueVizSecurityPackage.cs
```diff
@@ -50,10 +52,27 @@ protected override WindowPane InstantiateToolWindow(Type toolWindowType)
...
             {
+                TestShellLoad();
+
                 return new HotspotsToolWindow();
             }
 
             return base.InstantiateToolWindow(toolWindowType);
         }
+
+        private void TestShellLoad()
```
rita-g-sonarsource, line 63: I'll remove this after the PR changes - this is for testing the artifacts in 2015.

src/IssueViz.UnitTests/Commands/IssueVisualizationToolWindowCommandTests.cs
```diff
@@ -19,114 +19,201 @@
...
 using SonarLint.VisualStudio.IssueVisualization.Commands;
 using SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl;
-using IVsMonitorSelection = Microsoft.VisualStudio.Shell.Interop.IVsMonitorSelection;
+using ThreadHelper = SonarLint.VisualStudio.Integration.UnitTests.ThreadHelper;
 
 namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Commands
 {
     [TestClass]
-    public class IssueVisualizationToolWindowCommandTests : ToolWindowCommandTests<IssueVisualizationToolWindow>
+    public class IssueVisualizationToolWindowCommandTests
```
rita-g-sonarsource, line 38: I reverted the code back to master version.

src/SonarLint.VSSpecificAssemblies.props
```diff
@@ -157,29 +157,29 @@
...
-                      PrivateAssets="all">
-    </PackageReference>
-    <ReferencePath Include="$(NuGetPackageRoot)\Microsoft.VisualStudio.Shell.15.0\15.0.26201\lib\Microsoft.VisualStudio.Shell.15.0.dll">
-      <Private>$(CopyVSAssembliesToOutput)</Private>
-      <Aliases>versionSpecificShell</Aliases>
-    </ReferencePath>
+  <ItemGroup Label="Shell 15 references" Condition="$(RequiresShell15)=='true'">
+    <PackageReference Include="Microsoft.VisualStudio.Shell.15.0" Version="15.0.26201"/>
   </ItemGroup>
+  <Target Name="SilenceShell14" Condition="$(RequiresShell15)=='true'" BeforeTargets="FindReferenceAssembliesForReferences;ResolveReferences">
```
rita-g-sonarsource, line 171: SilenceShell14 will be used by IssueViz.Security and IssueViz.Security.Tests - it's to avoid conflicts caused by shell 14 coming from IssueViz assembly.

src/SonarLint.VSSpecificAssemblies.props
```diff
@@ -157,29 +157,29 @@
...
+    <ItemGroup>
+      <ReferencePath Remove="@(ReferencePath)" Condition="'%(FileName)' == 'Microsoft.VisualStudio.Shell.14.0'" />
+      <ReferencePath Remove="@(ReferencePath)" Condition="'%(FileName)' == 'Microsoft.VisualStudio.Shell.Immutable.14.0'" />
+    </ItemGroup>
+  </Target>
+  <Target Name="SilenceShell15" Condition="'$(RequiresShell15)'!='true'" BeforeTargets="FindReferenceAssembliesForReferences;ResolveReferences">
```
rita-g-sonarsource, line 177: SilenceShell15 will be used by Integration.Vsix and Integration.Vsix.Tests - it's to avoid conflicts caused by shell 15 coming from IssueViz.Security.

src/SonarLint.VSSpecificAssemblies.props
duncanp-sonar: It's probably adding a comment at the top of this section explaining the overall scenario e.g.

>Projects won't build if they reference both Shell 14 and Shell 15. Each project will only directly reference one Shell version. However, it could transitively reference the other Shell assembly e.g. X depends on Shell14 and Y, Y depends on Shell15. In this case, we need to remove the transitive references to Shell15 when building X.

src/SonarLint.VSSpecificAssemblies.props
```diff
@@ -157,29 +157,29 @@
...
+    <ItemGroup>
+      <ReferencePath Remove="@(ReferencePath)" Condition="'%(FileName)' == 'Microsoft.VisualStudio.Shell.15.0'" />
+      <ReferencePath Remove="@(ReferencePath)" Condition="'%(FileName)' == 'Microsoft.VisualStudio.Shell.Framework'" />
+      <ReferencePath Remove="@(ReferencePath)" Condition="$([System.String]::Copy(%(FullPath)).Contains('Microsoft.VisualStudio.Threading\15.0.240'))" />
```
duncanp-sonar, line 181: Probably worth commenting why this uses `Contains`

src/IssueViz.Security/HotspotsControl/HotspotsControl.xaml
duncanp-sonar: Presumably you'll hit this problem soon? It would be better to make sure we can work round it sooner rather than later.

duncanp-sonar (approved):

Issue #1795, opened 2020-10-14
Create hotspots ITableDataSource


rita-g-sonarsource:
#### Description

Add ITableDataSource for the new hotspots list. 
Implement IIssueLocationStore to allow future hooking into the tagging infrastructure (#1796)

PR #1826, opened 2020-10-22
Add empty HotspotsTableDataSource

rita-g-sonarsource:
Part of #1795


rita-g-sonarsource:
@duncanp-sonar this is infrastructure for the table data source; it doesn't handle yet the IssueLocationStore.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableConstants.cs
```diff
@@ -22,6 +22,9 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableD
 {
     internal static class HotspotsTableConstants
     {
+        public static readonly string TableDisplayName = Resources.HotspotsToolWindowCaption;
         public const string TableManagerIdentifier = nameof(TableManagerIdentifier);
+        public const string TableIdentifier = nameof(TableIdentifier);
```
duncanp-sonar, line 27: Am I missing something - the identifier for the table is the string "TableIdentifier"? Same for the type identifier.

Those aren't very descriptive names and there's nothing to suggest they relate to SonarLint. I'm guessing those identifiers are ones we're providing to VS so consumers can get hold of the table. If so, then we also want to reduce the possibility of name conflicts.
Either way, I think the identifiers should have some reference to SonarLint.

src/IssueViz.Security/IssueViz.Security.csproj
```diff
@@ -45,6 +45,9 @@
   <ItemGroup Label="Package references">
     <PackageReference Include="Microsoft.VisualStudio.ComponentModelHost" Version="14.0.25424" />
     <PackageReference Include="Microsoft.VisualStudio.Language.Intellisense" Version="14.0.23205" />
+    <PackageReference Include="Microsoft.VisualStudio.Imaging" Version="15.0.26201" />
```
duncanp-sonar, line 48: It looks like this is the first time we've used this package in product code. Why is it version 15?
The only existing ref I could find was in test code to v14.

src/IssueViz.Security/IssueViz.Security.csproj
```diff
@@ -45,6 +45,9 @@
...
+    <PackageReference Include="Microsoft.VisualStudio.ImageCatalog" Version="14.3.25407" />
+    <PackageReference Include="Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime" Version="16.7.30328.74" />
```
duncanp-sonar, line 50: I see we're already referencing this in IssueViz, but I don't understand why it's version 16.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.Shell.TableManager;
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableDataSource
+{
+    internal interface IHotspotsTableDataSource : ITableDataSource, IDisposable
```
duncanp-sonar, line 29: Just an observation: the equivalent `ISonarErrorListDataSource' interface doesn't directly expose the `ITableDataSource`.
It depends what the purpose of this interface is for, but it's not providing any level of abstraction at present.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntry.cs
```diff
@@ -0,0 +1,128 @@
...
+                case StandardTableColumnDefinitions.DocumentName:
+                    content = originalIssue.FilePath;
+                    break;
+
+                case StandardTableColumnDefinitions.Text:
+                    content = originalIssue.Message;
+                    break;
+
+                case StandardTableColumnDefinitions.Line:
+                    content = originalIssue.StartLine;
```
duncanp-sonar, line 58: What's the plan for showing the "live" position as the location is updated by the tagger?

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntry.cs
```diff
@@ -0,0 +1,128 @@
...
+
+                default:
+                    content = null;
+                    return false;
+            }
+
+            return true;
+        }
+
+        #region ITableEntry
```
duncanp-sonar, line 73: The label on this region is a bit misleading - it only contains some of the interface members. Is it to group the "unimplemented" members?

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntry.cs
```diff
@@ -0,0 +1,128 @@
...
+            this.hotspot = hotspot;
+        }
+
+        public object Identity => hotspot;
+
+        public bool TryGetValue(string keyName, out object content)
+        {
+            var originalIssue = hotspot.Issue;
+
+            switch (keyName)
```
duncanp-sonar, line 43: Presumably hotspots have a severity too?

duncanp-sonar: Perhaps we should check with the PM(s) which columns they want to see in the list.

duncanp-sonar (commented):
A few questions

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntry.cs
rita-g-sonarsource: Good question. I wasn't sure if we would even want to show line/column information, as it's not exactly like the error list (e.g. we also don't show project name). If we take the data from the span, then what should we do if the span is invalid? In the error list we return "null" in GetValue which stops the issue from being rendered, but here we would probably want to keep the item in the list.
Do you think it's worth showing line/column in the first place?

src/IssueViz.Security/IssueViz.Security.csproj
rita-g-sonarsource: My bad, I didn't notice that it was 15. 
I had to import it to reference ImageMoniker in HotspotTableEntry.

src/IssueViz.Security/IssueViz.Security.csproj
rita-g-sonarsource: I'll see what happens if I change it.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
rita-g-sonarsource: I'll remove it for now

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntry.cs
rita-g-sonarsource: Yep, I'll rename

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntry.cs
rita-g-sonarsource: No idea, do we get that in the API?
I think Alex said that the priority (high/medium/low) is taken from a spreadsheet, so I'm not sure we get that back. I'll ask Damien,

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntry.cs
duncanp-sonar: It's possible that there could be multiple hotspots of the same type in the same file. The user will need something to help disambiguate between them.

rita-g-sonarsource: Ok I'll handle that in a separate PR then, along with Severity column.

duncanp-sonar (approved):

PR #1828, opened 2020-10-23
Add priority column + take line/column from issue span

rita-g-sonarsource:
Part of #1795


src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableColumns.cs
```diff
@@ -29,8 +29,9 @@ internal static class HotspotsTableColumns
         public static IReadOnlyList<ColumnState> InitialStates { get; } = new[]
         {
             new ColumnState(StandardTableColumnDefinitions.ErrorCode, true, 50),
+            new ColumnState(StandardTableColumnDefinitions.Priority, true, 100),
+            new ColumnState(StandardTableColumnDefinitions.Text, true, 350),
```
rita-g-sonarsource, line 33: Minor: changed the order, so that the message/text would appear before the file name

src/IssueViz.Security.UnitTests/HotspotsList/HotspotTableEntryTests.cs
```diff
@@ -191,15 +227,37 @@ public void TryCreateStringContent_False()
...
-            var tryGetValue = new HotspotTableEntry(issueViz.Object).TryGetValue(column, out object value);
+            priorityConverter ??= Mock.Of<IAnalysisSeverityToPriorityConverter>();
+
+            var tryGetValue = new HotspotTableEntry(issueViz.Object, priorityConverter).TryGetValue(column, out var value);
             tryGetValue.Should().BeTrue();
 
             return value;
         }
+
+        private IAnalysisIssueVisualization CreateIssueVizWithSpan()
```
rita-g-sonarsource, line 243: Copied from IssuesSnapshotTests_GetValue

src/IssueViz.Security/AnalysisSeverityToPriorityConverter.cs
```diff
@@ -0,0 +1,82 @@
...
+        AnalysisIssueSeverity Convert(string priority);
+    }
+
+    internal class AnalysisSeverityToPriorityConverter : IAnalysisSeverityToPriorityConverter
+    {
+        private readonly IDictionary<string, AnalysisIssueSeverity> map = new Dictionary<string, AnalysisIssueSeverity>
+        {
+            {"high", AnalysisIssueSeverity.Blocker},
+            {"medium", AnalysisIssueSeverity.Major},
+            {"low", AnalysisIssueSeverity.Minor}
```
duncanp-sonar, line 47: Hotspots and Issue have different Severities and Types so we can't conveniently represent a hotspot as an 'IAnalysisIsue'.
We either need to have a re-think about how to use the visualization/issue interfaces so they work nicely with hotspots and issues, or we hack around with the hotspots representation to pretend they are issues too. This looks like the second approach. We should have a chat about the best way to address this problem.

duncanp-sonar: I think a clean solution will be reasonably straightforward as far as the taggers go. As far as I remember the taggers are only interested in location, and whether the location is primary or secondary.

The data sources should be ok too since the Error List data source only cares about issues, and the hotspot data source only cares about hotspots (at least at the moment).

That leaves the flow pane which shows an icon based on the severity. With a bit of luck we'll be able to get away with tweaking the visualization interfaces so they don't need to use the hotspot/issue severity directly.

PR #1829, opened 2020-10-23
Converted http listener to use Owin

duncanp-sonar:


src/IssueViz.Security/IssueViz.Security.csproj
```diff
@@ -89,6 +91,29 @@
...
 
+  <!--**********************************************************************-->
+  <!-- Workarounds -->
+  <!--**********************************************************************-->
+
+  <!-- The OWIN assemblies aren't being included in the VSIX for some reason.	
+       This workaround adds metadata to three artifacts so that the standard VSSDK targets will	
+       include them.	
+       NB the target is based on the one from Microsoft.VisualStudio.SDK.EmbedInteropTypes.
+       We have a similar hack in Intergration.csproj -->
```
duncanp-sonar, line 102: I've added #1832 to the infrastructure backlog to investigate whether we can clean up the VSIX authoring.

src/IssueViz.Security/OpenInIDE/ListenerFactory.cs
```diff
@@ -28,11 +30,11 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE
     internal interface IListenerFactory
     {
         /// <summary>
-        /// Attempts to create and return a new HTTP listener listening
+        /// Attempts to create and return a new HTTP listener listening(
```
duncanp-sonar, line 33: I've made the minimum changes required to use the OWIN classes that wrap the HttpListener so it will be easier to test the rest of the implementation. However, these changes don't make the existing factory class any more testable.

src/IssueViz.Security/OpenInIDE/ListenerFactory.cs
```diff
@@ -73,5 +73,20 @@ HttpListener IListenerFactory.Create(int startPort, int endPort)
             }
             return listener;
         }
+
+        private static IDisposable CreateListener(int port)
+        {
+            var appProperties = new AppProperties(new Dictionary<string, object>())
```
duncanp-sonar, line 79: Owin provides strongly-typed wrapper classes to make working with the untyped property bags easier.

src/IssueViz.Security/OpenInIDE/ListenerFactory.cs
```diff
@@ -73,5 +73,20 @@ HttpListener IListenerFactory.Create(int startPort, int endPort)
...
+            {
+                Addresses = AddressCollection.Create()
+            };
+
+            var address = Address.Create();
+            address.Port = port.ToString();
+            appProperties.Addresses.Add(address);
+
+            var openInIDEListener = new OpenInIDEHttpListener();
+            return OwinServerFactory.Create(openInIDEListener.ProcessRequest, appProperties.Dictionary);
```
duncanp-sonar, line 89: At present we don't need an OWIN pipeline of components to handle auth, routing etc (and we probably never will). So here I'm directly invoking the specific server factory and passing it our handler class that will handle the requests as they arrive.

src/IssueViz.Security/OpenInIDE/ListenerFactory.cs
rita-g-sonarsource, line 33: Minor: extra '(' in the comment

rita-g-sonarsource (approved):

PR #1830, opened 2020-10-26
Take line/column from issue span

rita-g-sonarsource:
Part of #1795


src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableColumns.cs
```diff
@@ -29,8 +29,8 @@ internal static class HotspotsTableColumns
         public static IReadOnlyList<ColumnState> InitialStates { get; } = new[]
         {
             new ColumnState(StandardTableColumnDefinitions.ErrorCode, true, 50),
-            new ColumnState(StandardTableColumnDefinitions.DocumentName, true, 200),
             new ColumnState(StandardTableColumnDefinitions.Text, true, 450),
+            new ColumnState(StandardTableColumnDefinitions.DocumentName, true, 200),
```
rita-g-sonarsource, line 33: Minor: changed the order, so that the message/text would appear before the file name

src/IssueViz.Security.UnitTests/HotspotsList/HotspotTableEntryTests.cs
```diff
@@ -201,5 +222,25 @@ private static object GetValue(IAnalysisIssue issue, string column)
 
             return value;
         }
+
+        private IAnalysisIssueVisualization CreateIssueVizWithSpan()
```
rita-g-sonarsource, line 226: Copied from IssuesSnapshotTests_GetValue

duncanp-sonar: I assumed "Copied from X" meant the whole method was copied so I was a bit confused that that isn't the case - looks like it's part of the constructor.

I think's worth changing the method name to emphasise that the returned object is shared and the individual properties matter e.g. `CreateWellKnownIssueVizAndSpan` - it's more of a warning to future maintainers that changing anything in the function might impact multiple tests.

In general I try to avoid sharing a complex object over multiple tests where the tests check specific individual values; it breaks test isolation and makes tests harder to read. If I do have to shared objects between tests then I try to make sure the tests don't depend on its properties i.e. the test just needs a valid snapshot but doesn't care about the content.

duncanp-sonar (approved):

PR #1835, opened 2020-10-26
Hotspots table: support empty span

rita-g-sonarsource:
Part of #1795


rita-g-sonarsource:
As part of the store testing, I realized I didn't support the "empty span" use-case.

duncanp-sonar (approved):

PR #1836, opened 2020-10-26
Implement IIssueLocationStore

### LINKED ISSUE #1795: Create hotspots ITableDataSource

rita-g-sonarsource:
Fixes #1795


src/IssueViz.Security.UnitTests/HotspotsList/HotspotTableEntryTests.cs
```diff
@@ -167,16 +166,6 @@ public void TryGetValue_UnknownColumn_ReturnsNull()
             content.Should().BeNull();
         }
 
-        [TestMethod]
```
rita-g-sonarsource, line 170: Deleted tests that were part of IWpfTableEntry

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntry.cs
```diff
@@ -19,14 +19,13 @@
...
-using System.Windows;
 using Microsoft.VisualStudio.Shell.TableControl;
+using Microsoft.VisualStudio.Shell.TableManager;
 using SonarLint.VisualStudio.IssueVisualization.Models;
-using ImageMoniker = Microsoft.VisualStudio.Imaging.Interop.ImageMoniker;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableDataSource
 {
-    internal class HotspotTableEntry : IWpfTableEntry
+    internal class HotspotTableEntry : ITableEntry
```
rita-g-sonarsource, line 28: As we discussed offline, I'm having trouble with `IWpfTableEntry` interface due to method `bool TryCreateImageContent(string columnName, bool singleColumnView, out ImageMoniker content);`. 
`ImageMoniker` is in `Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime` assembly, but I couldn't get it to work: the MEF import just kept failing on missing assembly
```Microsoft.VisualStudio.Composition.PartDiscoveryException: Failure while scanning type "SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableDataSource.HotspotsTableDataSource". ---> Microsoft.VisualStudio.Composition.PartDiscoveryException: Part discovery failed at member tableEntries. ---> System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
```
For now I changed it to `ITableEntry` and it works, but it doesn't support icons or tooltips. Once we need those, I'll invest more into the assembly problem.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
```diff
@@ -20,12 +20,25 @@
...
+using SonarLint.VisualStudio.Core.Helpers;
+using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+using SonarLint.VisualStudio.IssueVisualization.Models;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableDataSource
 {
-    internal sealed class HotspotsTableDataSource : ITableDataSource, IDisposable
+    internal interface IHotspotsStore
+    {
+        void Add(IAnalysisIssueVisualization hotspot);
```
rita-g-sonarsource, line 35: This is the method that the HttpRequest handler can call.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
```diff
@@ -67,5 +81,55 @@ public void Dispose()
...
+
+            lock (sinks)
+            {
+                foreach (var sink in sinks)
+                {
+                    sink.AddEntries(new[] {entry});
+                }
+            }
+
+            var hotspotFilePaths = hotspot.GetAllLocations().Select(x => x.CurrentFilePath);
```
rita-g-sonarsource, line 98: GetAllLocations = new extension method in `AnalysisIssueVisualizationExtensionsTests`

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntry.cs
duncanp-sonar: Ok. It's interesting that the missing assembly has `DesignTime` in the name; perhaps it contains interop types that should be embedded in our assembly.

duncanp-sonar: Just an observation: this seems like a logically separable change; I'd have probably kept it separate from the new interface implementation. It's not worth changing now though.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
duncanp-sonar: Yep - except it won't be the HTTP request handler itself; it'll be a class that knows nothing about HTTP.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
duncanp-sonar: Can there be duplicate file paths in the list, or will they be de-duplicated in the event args class?

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
```diff
@@ -67,5 +81,55 @@ public void Dispose()
...
+                .Where(locationViz => PathHelper.IsMatchingPath(locationViz.CurrentFilePath, filePath));
+
+            return matchingLocations;
+        }
+
+        public void Refresh(IEnumerable<string> affectedFilePaths)
+        {
+            var changedEntries = tableEntries
+                .Where(entry => affectedFilePaths.Any(p =>
+                    PathHelper.IsMatchingPath(p, (entry.Identity as IAnalysisIssueVisualization).CurrentFilePath)))
```
duncanp-sonar, line 118: Using 'as' implies we think the object might not be an 'IAnalysisIssueVisualization'.
If we're certain of the type it would be clearer to use a hard cast.
If we're not certain of the type then we should use 'as' and handle the possibility that it could be null e.g. '?.CurrentFilePath' (assuming the PathHelper handles nulls...)

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
```diff
@@ -67,5 +81,55 @@ public void Dispose()
...
+                .ToList();
+
+            if (!changedEntries.Any())
```
duncanp-sonar, line 121: Minor: you've turned the result into a List so you can use the 'Count' property rather than the Any extension method (I thought there was a analysis rule for this?)

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntry.cs
rita-g-sonarsource: Hmm, your comment about interop reminded me something: IssueViz imports `Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime`, which I tried to import here and it didn't work either, but it also imports `Microsoft.VisualStudio.SDK.EmbedInteropTypes` -- and this might be the missing piece. I'll try it and see if it fixes it.

src/IssueViz.Security.UnitTests/HotspotsList/HotspotsTableDataSourceTests.cs
```diff
@@ -79,6 +107,178 @@ public void Unsubscribe_RemovesEntriesFromSink()
...
+
+        [TestMethod]
+        public void GetLocations_HasTableEntriesForGivenFilePath_ReturnsMatchingLocations()
+        {
+            var locationViz = new Mock<IAnalysisIssueLocationVisualization>();
+            locationViz.Setup(x => x.CurrentFilePath).Returns("somefile.cpp");
+
+            var issueViz1 = CreateIssueViz("somefile.cpp");
+            var issueViz2 = CreateIssueViz("someotherfile.cpp", locationViz.Object);
+            var issueViz3 = CreateIssueViz("somefile.cpp");
```
duncanp-sonar, line 194: You could change the casing of `issueViz3` to test that casing is handled correctly.

duncanp-sonar (approved):
Some comments inline

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
rita-g-sonarsource: I can't find any rule for this 🤔

duncanp-sonar:
Ah. I have a third-party analyzer VSIX installed locally. It might be from
that.

Duncan Pocklington | SonarSource

Developer

https://sonarsource.com <http://sonarsource.com/>

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
rita-g-sonarsource: Fixed, added ToLower + Distinct

duncanp-sonar: Use `.Distinct(StringComparer.OrdinalIgnoreCase)`

Some languages don't have a one-to-one mapping between upper and lower case characters (I think Turkish is an example) so forcing to upper/lower and comparing isn't the right solution.

rita-g-sonarsource: Huh, good to know 👍

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
duncanp-sonar: Ah. I have a third-party analyzer VSIX installed locally. It might be from that.

Issue #1763, opened 2020-10-07
Convert SQ issue into IssueVisualization


rita-g-sonarsource:
### Description

Convert the returned issue from SQ into AnalysisIssueVisualization object, so that it could be used in the visualization framework.

- [ ] ~~calculate line hashes~~ --> will be handled when secondary locations are added
- [x]  translate file paths (#1874)
- [ ] ~~modify rule key? (https://github.com/duncanp-sonar/SLVS_SecondaryLocationsProto/blob/master/src/VSSecondaryLocationsProto/Issues/IssueAdapter.cs#L14)~~ --> will not do for now, in order to remain consistent with the rule keys that are displayed for CFamily analysis.
- [ ] ~~reverse locations? (https://github.com/duncanp-sonar/SLVS_SecondaryLocationsProto/blob/master/src/VSSecondaryLocationsProto/Issues/IssuesProvider.cs#L156)~~  --> will be handled when secondary locations are added

PR #1838, opened 2020-10-27
Implement IHotspot

rita-g-sonarsource:
Prep for #1763


rita-g-sonarsource:
Basic implementation; part of the stuff I needed to test the table integration and that we will need in the future.

duncanp-sonar (approved):

Issue #1825, opened 2020-10-22
Implement request listener plumbing


duncanp-sonar:
### Description

- [x] decide what to base the http listener code on - needs to be testable e.g. HttpListener/shim around HttpListener/third-party-shim/some other library
- [x] implement basic request handling infrastructure: routing, error handling, handling unrecognized requests
- [x] unit tests

PR #1840, opened 2020-10-27
Added pipeline message processing

duncanp-sonar:
Relates to #1825

### DIFF
--- a/src/IssueViz.Security.UnitTests/OpenInIDE/Http/HttpListenerFactoryTests.cs
+++ b/src/IssueViz.Security.UnitTests/OpenInIDE/Http/HttpListenerFactoryTests.cs
@@ -19,8 +19,11 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.Owin.Host.HttpListener;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
@@ -46,6 +49,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
         }
 
         [TestMethod]
+        [TestCategory("Integration")]
         [DataRow(10000, 10003, 10000)]
         [DataRow(10000, 10003, 10001)]
         [DataRow(10000, 10003, 10002)]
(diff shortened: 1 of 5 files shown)


src/IssueViz.Security/OpenInIDE/OpenInIDEResources.resx
```diff
@@ -118,18 +118,30 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="Factory_CheckingPort" xml:space="preserve">
-    <value>Checking availability of port {0}</value>
+    <value>[Open in IDE] Checking availability of port {0}</value>
```
duncanp-sonar, line 121: Miscellaneous: I've added a common prefix to the open in IDE messages to see if that makes the output easier to read when they are mixed up with errors from other sources.

src/IssueViz.Security/OpenInIDE/Http/OwinPipelineProcessor.cs
```diff
@@ -58,10 +61,29 @@ public OwinPipelineProcessor([ImportMany]IEnumerable<IOwinPathRequestHandler> pa
         /// <summary>
         /// Called for each low-level HTTP request received by the underlying HTTPListener
         /// </summary>
-        public Task ProcessRequest(IDictionary<string, object> environment)
+        public async Task ProcessRequest(IDictionary<string, object> environment)
```
duncanp-sonar, line 64: Simple implementation to handle urecognized messages and errors:
* all requests are logged
* unrecognised requests return a 200 status code
* exceptions are logged and rethrown. We expect the calling component turn this into a 500 error code (and there's an integration test to check this is the case)

src/IssueViz.Security.UnitTests/OpenInIDE/Http/HttpListenerFactoryTests.cs
```diff
@@ -46,6 +49,7 @@ public void MefCtor_CheckIsExported()
         }
 
         [TestMethod]
+        [TestCategory("Integration")]
```
duncanp-sonar, line 52: Miscellaneous: added a category to help identify/exclude tests that are dependent on the environment.

rita-g-sonarsource (approved):

Issue #1839, opened 2020-10-27
Implement "show hotspot" HTTP request handler


duncanp-sonar:
### Description

Implement the low-level HTTP request handler
- [x] extract and validate parameters
- [x] create strongly-typed request
- [x] unit tests

PR #1843, opened 2020-10-28
Implement show hotpot HTTP handler

duncanp-sonar:
Relates to #1839

### DIFF
--- a/src/IssueViz.Security.UnitTests/OpenInIDE/Http/ShowHotspotRequestHandlerTests.cs
+++ b/src/IssueViz.Security.UnitTests/OpenInIDE/Http/ShowHotspotRequestHandlerTests.cs
@@ -30,7 +30,7 @@ using SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Http;
 namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE.Http
 {
     [TestClass]
-    public class OpenHotspotRequestHandlerTests
+    public class ShowHotspotRequestHandlerTests
     {
         [TestMethod]
         public void MefCtor_CheckIsExported()
@@ -40,10 +40,11 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
             var loggerExport = MefTestHelpers.CreateExport<ILogger>(Mock.Of<ILogger>());
 
             // Act & Assert
-            MefTestHelpers.CheckTypeCanBeImported<OpenHotspotRequestHandler, IOwinPathRequestHandler>(null, new[] { apiRequestHandler, loggerExport });
+            MefTestHelpers.CheckTypeCanBeImported<ShowHotspotRequestHandler, IOwinPathRequestHandler>(null, new[] { apiRequestHandler, loggerExport });
         }
 
         [TestMethod]
+        [DataRow("", "server;project;hotspot")]
         [DataRow("server=s&hotspot=h&organization=o", "project")]
         [DataRow("project=p&hotspot=h&organization=o", "server")]
         [DataRow("project=p&server=s&organization=o", "hotspot")]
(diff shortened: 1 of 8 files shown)


src/IssueViz.Security.UnitTests/OpenInIDE/Http/ShowHotspotRequestHandlerTests.cs
```diff
@@ -0,0 +1,121 @@
...
+using SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Contract;
+using SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Http;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE.Http
+{
+    [TestClass]
+    public class ShowHotspotRequestHandlerTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
```
duncanp-sonar, line 36: FYI this test is unchanged. The other tests are new.

duncanp-sonar:
I renamed the classes from `OpenHotspotXXX` to `ShowHotspotXXX`. Unfortunately, that's messed up how the GitHub web UI is showing the diff. However, the previous implementation was just a skeleton implementation and the new code is additive. I'll point out the bits that haven't changed.

src/IssueViz.Security/OpenInIDE/Http/ShowHotspotRequestHandler.cs
```diff
@@ -0,0 +1,117 @@
...
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Http
+{
+    /// <summary>
+    /// Handles low-level HTTP request to open a hotspot
+    /// </summary>
+    [Export(typeof(IOwinPathRequestHandler))]
+    internal class ShowHotspotRequestHandler : IOwinPathRequestHandler
+    {
+        private const string ParamName_Server = "server";
```
duncanp-sonar, line 37: The constants are new and the implementation of `ProcessRequest` is new. The constructor and MEF attributes haven't changed.

src/IssueViz.Security/OpenInIDE/Http/ShowHotspotRequestHandler.cs
```diff
@@ -0,0 +1,117 @@
...
+        private const string ParamName_Project = "project";
+        private const string ParamName_Hotspot = "hotspot";
```
duncanp-sonar, line 39: I looked at the existing SQ API parameter names, and the standard form appears to be e.g. 'hotspot' rather than 'hotspotKey'.

There's an exception in the SQ API `api/hotspots/search` which has `projectKey`, but that appears to be an anomaly.

src/IssueViz.Security/OpenInIDE/Http/ShowHotspotRequestHandler.cs
```diff
@@ -0,0 +1,117 @@
...
+        public Task ProcessRequest(IOwinContext context)
+        {
+            var request = BuildRequest(context);
+            if (request == null)
+            {
+                context.Response.StatusCode = (int)HttpStatusCode.BadRequest;
+                return Task.CompletedTask;
+            }
+
+            // TODO - make async?
```
duncanp-sonar, line 63: As discussed offline, I'll do this in a separate PR.

src/IssueViz.Security/OpenInIDE/Http/ShowHotspotRequestHandler.cs
```diff
@@ -0,0 +1,117 @@
...
+        private const string ParamName_Server = "server";
+        private const string ParamName_Project = "project";
+        private const string ParamName_Hotspot = "hotspot";
+        private const string ParamName_Organization = "organization";
+
+        private readonly IOpenInIDERequestHandler openInIDERequestHandler;
+        private readonly ILogger logger;
+
+        [ImportingConstructor]
+        internal ShowHotspotRequestHandler(IOpenInIDERequestHandler openInIDERequestHandler, ILogger logger)
```
rita-g-sonarsource, line 46: The names are slightly confusing, both `ShowHotspotRequestHandler` and `IOpenInIDERequestHandler` are "request handlers", and it's not clear what distinction between them.

Perhaps `ShowHotspotRequestHandler` should be named `ShowHotspotHttpRequestHandler` or something of the sort.

rita-g-sonarsource (approved):

src/IssueViz.Security/OpenInIDE/Http/ShowHotspotRequestHandler.cs
duncanp-sonar: I'll change `ShowHotspotRequestHandler` to `ShowHotspotOwinRequestHandler`.

I don't think `IOpenInIDERequestHandler` should change to `IShowHotspotRequestHandler` since it's intended to handle all API calls, which we know will include handling vulnerabilities by the time of the LTS.

I was debating using "HTTP"/"Owin" in all of the zweb-aware components and "Api" in all of the web-agnostic components. I couldn't decide what to change `IOpenInIDERequestHandler` to, though. Maybe something like `ISonarLintApiHandler`.

PR #1844, opened 2020-10-28
Move IInfoBarManager to Core

### LINKED ISSUE #1793: Provide UI that displays request failures to the user
#### Description

There are multiple possible failure scenarios when handling a request from SQ.

Implement a way to display the messages to the user.

https://docs.microsoft.com/en-us/visualstudio/extensibility/ux-guidelines/notifications-and-progress-for-visual-studio?view=vs-2019

TODO: 
- [ ] New modal dialog? -> Not for now
- [x] Goldbar? 
- [ ] Log in SonarLint output pane?

rita-g-sonarsource:
Fixes #1793

### DIFF
--- a/src/Core/SonarLintImageMoniker.cs
+++ b/src/Core/SonarLintImageMoniker.cs
@@ -1,0 +1,39 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarLint.VisualStudio.Core
+{
+    /// <summary>
+    /// Wrapper class for Microsoft.VisualStudio.Imaging.Interop.ImageMoniker struct
+    /// </summary>
+    public readonly struct SonarLintImageMoniker
+    {
+        public SonarLintImageMoniker(Guid guid, int id)
+        {
+            Guid = guid;
+            Id = id;
+        }
+
+        public Guid Guid { get; }
+        public int Id { get; }
+    }
+}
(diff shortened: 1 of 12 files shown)


src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -204,6 +204,8 @@ public void InfoBarManager_DetachInfoBar()
 
         #region Test helpers
 
+        private static SonarLintImageMoniker CreateFromVsMoniker(dynamic imageMoniker) => new SonarLintImageMoniker(imageMoniker.Guid, imageMoniker.Id);
```
rita-g-sonarsource, line 207: `KnownMonikers` is a very problematic class to pass as argument, as it creates ambiguity between `ImageMoniker` from `Microsoft.VisualStudio.ImageCatalog` assembly and `ImageMoniker` from `Microsoft.VisualStudio.Imaging.Interop`. Since it's test code, I opted to just use dynamic.

duncanp-sonar: You don't need 'dynamic'; just replace it here with the fully qualified type name i.e. `Microsoft.VisualStudio.Imaging.Interop.ImageMoniker`.

rita-g-sonarsource: It's the exact same namespace, but different assemblies.

src/Core/SonarLintImageMoniker.cs
```diff
@@ -0,0 +1,39 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarLint.VisualStudio.Core
+{
+    /// <summary>
+    /// Wrapper class for Microsoft.VisualStudio.Imaging.Interop.ImageMoniker struct
```
duncanp-sonar, line 26: `Wrapper` is a bit misleading; it`s duplicating the type, rather than wrapping it.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
duncanp-sonar: The test built and ran fine for me locally using the fully qualified type name instead of dynamic. It also worked if I added 'using Microsoft.VisualStudio.Imaging.Interop;'.
Does it fail for you locally, or in a CI build?

rita-g-sonarsource: Odd; it fails for me locally. And it makes sense, since both assemblies are indeed referenced.
Weird that it works for you 🤔 
image

rita-g-sonarsource: Edit: I see now that despite this error, the code actually compiles and the tests pass, that's very odd 🤔 
So I will check-in it with the namespace and just ignore the editor errors.

duncanp-sonar (approved):

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
duncanp-sonar: No. Could be an incorrect IntelliSense cache on your machine; that would cause those symptoms.

Issue #1797, opened 2020-10-14
Handle SQ request failures


rita-g-sonarsource:
#### Description

There are multiple potential failure scenarios when attempting to handle a request coming from SQ.

- [x] Add checks for the failure use cases https://docs.google.com/document/d/1MOG8172q5yiTkkFcupwT7_gEGRo83nCsmvAWqidWOfs/edit#heading=h.8n0g60dkbrwz
- [x] Display the failure to the user (#1793)

PR #1847, opened 2020-10-29
Split IConfigurationProvider and IConfigurationPersister implementation

rita-g-sonarsource:
Prep for #1797


src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -336,8 +336,24 @@ private ILocalService GetConfigurationProvider()
 
             var bindingSerializer = new SolutionBindingSerializer(sccFileSystem, bindingFileLoader, credentialsLoader);
 
-            return new ConfigurationProvider(legacyConfigPathProvider, connectedModeConfigPathProvider, bindingSerializer,
-                legacyConfigFolderItemAdder);
+            return new ConfigurationPersister(legacyConfigPathProvider, connectedModeConfigPathProvider, bindingSerializer, legacyConfigFolderItemAdder);
```
rita-g-sonarsource, line 339: This copy-paste will go away in future PRs

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -336,8 +336,24 @@ private ILocalService GetConfigurationProvider()
...
+        }
+
+        private ILocalService GetConfigurationProvider()
```
rita-g-sonarsource, line 342: GetConfigurationProvider and GetConfigurationPersister are identical as the moment. This copy-paste will go away in future PRs, as I split more dependencies.

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -1,7 +1,7 @@
 ﻿---
 ################################
 # Assembly references report
-# Report date/time: 2020-10-27T14:54:08.8664584Z
+# Report date/time: 2020-10-29T09:42:23.0046223Z
```
duncanp-sonar, line 4: It doesn't look like anything has changed in the references, which is what I'd expect. However, in that case I wouldn't have expected the AsmRef targets to have updated the baseline in this case.

Committing these unnecessary changes is fine; I'm just confused why the tooling produced them. Did you make reference changes locally then revert them?

duncanp-sonar (approved):

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
rita-g-sonarsource: Hm, I can't recall. I'll try to notice it if it happens again.

PR #1850, opened 2020-10-29
Remove dependency on ISolutionRuleSetsInformationProvider

rita-g-sonarsource:
Prep for #1797

### DIFF
--- a/src/Integration/Persistence/LegacySolutionBindingPathProvider.cs
+++ b/src/Integration/Persistence/LegacySolutionBindingPathProvider.cs
@@ -20,28 +20,31 @@
 
 using System;
 using System.IO;
-using SonarLint.VisualStudio.Core.Binding;
+using Microsoft.VisualStudio.Shell.Interop;
 
 namespace SonarLint.VisualStudio.Integration.Persistence
 {
     internal class LegacySolutionBindingPathProvider : ISolutionBindingPathProvider
     {
-        private readonly IServiceProvider serviceProvider;
-
+        private readonly IVsSolution vsSolution;
         public const string LegacyBindingConfigurationFileName = "SolutionBinding.sqconfig";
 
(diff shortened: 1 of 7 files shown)


src/Integration/Persistence/LegacySolutionBindingPathProvider.cs
```diff
@@ -20,28 +20,31 @@
...
     {
-        private readonly IServiceProvider serviceProvider;
-
+        private readonly IVsSolution vsSolution;
         public const string LegacyBindingConfigurationFileName = "SolutionBinding.sqconfig";
 
-        public LegacySolutionBindingPathProvider(IServiceProvider serviceProvider)
-        { 
-            this.serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+        public LegacySolutionBindingPathProvider(IVsSolution vsSolution)
```
rita-g-sonarsource, line 32: Prep work to be able to consume this class without being dependent on LocalServices - `ISolutionRuleSetsInformationProvider` was in the way, and the method GetSolutionSonarQubeRulesFolder was used only here. I changed it to work with `IVsSolution` instead of DTE.

duncanp-sonar: How are you planning to create this path provider and the other one going forwards? 

`IVsSolution` isn't directly MEF-importable so they will have to be explicitly created by something else. If you make the constructor import `SVSServiceProvider` then they could be imported directly.

duncanp-sonar (approved):

src/Integration/Persistence/LegacySolutionBindingPathProvider.cs
rita-g-sonarsource: Yep, that's the plan.
Once this PR and  #1851 are done, I'll new-up dependencies of ConfigurationProvider and to them IServiceProvider.

PR #1851, opened 2020-10-29
MEF export ICredentialStoreService

rita-g-sonarsource:
Prep for #1797

### DIFF
--- a/src/Integration/LocalServices/CredentialStore.cs
+++ b/src/Integration/LocalServices/CredentialStore.cs
@@ -20,6 +20,7 @@
 
 using System;
 using System.ComponentModel;
+using System.ComponentModel.Composition;
 using Microsoft.Alm.Authentication;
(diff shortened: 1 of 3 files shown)


src/Integration/LocalServices/CredentialStore.cs
```diff
@@ -28,6 +29,7 @@ namespace SonarLint.VisualStudio.Integration
     /// <summary>
     /// Simple wrapper around SecretStore to allow it to be registered as a ILocalService
     /// </summary>
+    [Export(typeof(ICredentialStoreService))]
```
rita-g-sonarsource, line 32: Prep work to be able to MEF-import IConfigurationProvider, which depends on SolutionBindingDataReader -> SolutionBindingCredentialsLoader.
The rest of the classes could then be new-ed.

duncanp-sonar (approved):

PR #1852, opened 2020-10-29
MEF export ConfigurationProvider

rita-g-sonarsource:
Prep for #1797


src/Integration/NewConnectedMode/ConfigurationProvider.cs
```diff
@@ -19,19 +19,34 @@
...
 using System;
+using System.ComponentModel.Composition;
 using System.IO;
+using Microsoft.VisualStudio.Shell;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Integration.Persistence;
 
 namespace SonarLint.VisualStudio.Integration.NewConnectedMode
 {
+    [Export(typeof(Core.Binding.IConfigurationProvider))]
```
rita-g-sonarsource, line 30: The relevant change

src/Integration/NewConnectedMode/IConfigurationProvider.cs
```diff
@@ -18,12 +18,9 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using SonarLint.VisualStudio.Core.Binding;
-
 namespace SonarLint.VisualStudio.Integration.NewConnectedMode
 {
-    public interface IConfigurationProvider : ILocalService
+    public interface IConfigurationProvider : Core.Binding.IConfigurationProvider, ILocalService
```
rita-g-sonarsource, line 23: I was really debating if I should just remove `ILocalService` and the assertion, but I didn't want to further changes to VsSessionHost and changing the services dictionary to `object`, it didn't feel right.
So I opted for an interface that implements both  `IConfigurationProvider` and `ILocalService`.
WDYT?

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -340,17 +339,8 @@ private ILocalService GetConfigurationPersister()
 
         private ILocalService GetConfigurationProvider()
```
rita-g-sonarsource, line 340: I left `GetConfigurationPersister` method as-is: the ctor of `ConfigurationPersister` contains dependencies that can't be easily new-ed, so I just left the construction here as it was.

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -340,17 +339,8 @@ private ILocalService GetConfigurationPersister()
...
-            var legacyConfigPathProvider = new LegacySolutionBindingPathProvider(solution);
-
             var store = this.GetService<ICredentialStoreService>();
-            var credentialsLoader = new SolutionBindingCredentialsLoader(store);
-            var bindingFileLoader = new SolutionBindingFileLoader(Logger);
-
-            var solutionBindingDataReader = new SolutionBindingDataReader(bindingFileLoader, credentialsLoader);
-
-            return new ConfigurationProvider(legacyConfigPathProvider, connectedModeConfigPathProvider, solutionBindingDataReader);
+            return new ConfigurationProvider(this, Logger, store);
```
duncanp-sonar, line 343: The credential store is a thin wrapper that just requires a logger to be constructed.
Does it need to be passed in to the ConfigurationProvider, or could it be new-ed up in the internal constructor?

Also, the convention with parameter-ordering is to put the logger last.

src/Integration/NewConnectedMode/IConfigurationProvider.cs
duncanp-sonar: I think creating an interface that creates both is an ok compromise, but I don't think it should be called `IConfigurationProvider`. It's very confusing.

We're effectively trying to hide the fact that we have two unrelated approaches to constructing objects, but we can't do it cleanly. All of the service-based usages have to include `using IConfigurationProvider = SonarLint.VisualStudio.Integration.NewConnectedMode.IConfigurationProvider;`, and the MEF export of the implementing class is confusing.

It would be less confusing to make it more explicit what we're doing by giving the new interface a different name e.g. `IConfigurationProviderService`.

src/Integration/NewConnectedMode/ConfigurationProvider.cs
duncanp-sonar: It's very easy to miss that the export is of interface X whereas the class implements interface Y because the interface names are the same.

I don't think we should try to hide this; it will just cause problems later.

src/Integration/MefServices/VsSessionHost.cs
rita-g-sonarsource: It could be new-ed up but then it won't be the same store instance as used in ConfigurationPersister. They both use ALM's `SecretStore`, that has a cache in the implementation, and a locking mechanism surrounding it. We don't do a lot of read-write operations so perhaps it's not a problem, but I think it's still best to use the same instance. I'll change the order of the parameters.

duncanp-sonar (approved):

PR #1853, opened 2020-10-30
Added IDE state validation interface

duncanp-sonar:
* added skeleton show hotspot api consumer


src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+        
+        void IOpenInIDERequestHandler.ShowHotspot(IShowHotspotRequest request)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            // TODO: change IShowHostpotRequest server property to be a URI
```
duncanp-sonar, line 56: I'll change the property to be a URI and the IOpenInIDERequestHandler methods to be Task<X> in a separate PR.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+    internal class OpenInIDERequestHandler : IOpenInIDERequestHandler
+    {
+        private readonly IOpenInIDEStateValidator ideStateValidator;
+        private readonly ISonarQubeService sonarQubeService;
+        private readonly ILogger logger;
+
+        [ImportingConstructor]
+        public OpenInIDERequestHandler(IOpenInIDEStateValidator ideStateValidator, ISonarQubeService sonarQubeService, ILogger logger)
+        {
+            this.ideStateValidator = ideStateValidator ?? throw new ArgumentNullException(nameof(ideStateValidator));
```
rita-g-sonarsource, line 39: I've been thinking: are the Null validation checks necessary? It seems that MEF will fail if there are no contracts, so it will never be possible to get a null in a MEF ctor.

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        private static Mock<IOpenInIDEStateValidator> CreateValidator(string uri, string projectKey, string orgKey, bool canHandleRequest)
+        {
+            var mockValidator = new Mock<IOpenInIDEStateValidator>();
+            mockValidator.Setup(x => x.CanHandleOpenInIDERequest(It.Is<Uri>(x => x.ToString() == uri), projectKey, orgKey))
+                .Returns(canHandleRequest);
+
+            return mockValidator;
+        }
+
+        private class TestShowHotspotRequest : IShowHotspotRequest
```
rita-g-sonarsource, line 82: Why not use a mock and setup the return values?

rita-g-sonarsource (commented):
Please see my comments

rita-g-sonarsource (approved):

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
duncanp-sonar: No strong reason, except I find the mock Setup syntax clumsier and harder to read, and it isn't obvious whether the whole object is being mocked or not.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
duncanp-sonar: As long as we create them via MEF, no. We don't have a consistent approach in the existing. It looks to be about 50/50 between those that check and those that don't.
I'll remove the checks.

PR #1856, opened 2020-11-02
Show infobar with SQ request failures

### LINKED ISSUE #1797: Handle SQ request failures

rita-g-sonarsource:
Fixes #1797

### DIFF
--- a/src/IssueViz.Security/HotspotsList/HotspotsToolWindow.cs
+++ b/src/IssueViz.Security/HotspotsList/HotspotsToolWindow.cs
@@ -28,10 +28,10 @@ using Microsoft.VisualStudio.Shell.TableManager;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList
 {
-    [Guid(Guid)]
+    [Guid(ToolWindowId)]
     public class HotspotsToolWindow : ToolWindowPane
     {
-        public const string Guid = "4BCD4392-DBCF-4AA2-9852-01129D229CD8";
+        public const string ToolWindowId = "4BCD4392-DBCF-4AA2-9852-01129D229CD8";
 
         public HotspotsToolWindow(IServiceProvider serviceProvider)
         {
--- a/src/IssueViz.Security/HotspotsList/HotspotsToolWindow.cs
+++ b/src/IssueViz.Security/HotspotsList/HotspotsToolWindow.cs
@@ -28,9 +28,11 @@ using Microsoft.VisualStudio.Shell.TableManager;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList
 {
-    [Guid("4BCD4392-DBCF-4AA2-9852-01129D229CD8")]
+    [Guid(Guid)]
     public class HotspotsToolWindow : ToolWindowPane
     {
+        public const string Guid = "4BCD4392-DBCF-4AA2-9852-01129D229CD8";
+
         public HotspotsToolWindow(IServiceProvider serviceProvider)
         {
             Caption = Resources.HotspotsToolWindowCaption;
--- a/src/IssueViz.Security/OpenInIDE/Api/IOpenInIDEStateValidator.cs
+++ b/src/IssueViz.Security/OpenInIDE/Api/IOpenInIDEStateValidator.cs
@@ -75,7 +75,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Api
                 return true;
             }
 
-            AddInfoBar(failureMessage);
+            AddInfoBar();
             logger.WriteLine(failureMessage);
 
             return false;
@@ -88,21 +88,21 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Api
 
             if (configuration.Mode == SonarLintMode.Standalone)
             {
-                reason = OpenInIDEResources.Infobar_InvalidStateReason_NotInConnectedMode;
+                reason = OpenInIDEResources.RequestValidator_InvalidStateReason_NotInConnectedMode;
             }
             else if (!configuration.Project.ServerUri.Equals(serverUri))
             {
-                reason = string.Format(OpenInIDEResources.Infobar_InvalidStateReason_WrongServer, configuration.Project.ServerUri);
+                reason = string.Format(OpenInIDEResources.RequestValidator_InvalidStateReason_WrongServer, configuration.Project.ServerUri);
             }
             else if (!string.IsNullOrEmpty(organizationKey) &&
                 (configuration.Project.Organization == null ||
                  !organizationKey.Equals(configuration.Project.Organization.Key, StringComparison.OrdinalIgnoreCase)))
             {
-                reason = string.Format(OpenInIDEResources.Infobar_InvalidStateReason_WrongOrganization, configuration.Project.Organization?.Key);
+                reason = string.Format(OpenInIDEResources.RequestValidator_InvalidStateReason_WrongOrganization, configuration.Project.Organization?.Key);
             }
             else if (!configuration.Project.ProjectKey.Equals(projectKey, StringComparison.OrdinalIgnoreCase))
             {
-                reason = string.Format(OpenInIDEResources.Infobar_InvalidStateReason_WrongProject, configuration.Project.ProjectKey);
+                reason = string.Format(OpenInIDEResources.RequestValidator_InvalidStateReason_WrongProject, configuration.Project.ProjectKey);
             }
 
             if (reason == null)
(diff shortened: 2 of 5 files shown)


src/IssueViz.Security/OpenInIDE/Api/IOpenInIDEStateValidator.cs
```diff
@@ -41,4 +46,103 @@ internal interface IOpenInIDEStateValidator
...
+            }
+            else if (!configuration.Project.ServerUri.Equals(serverUri))
+            {
+                reason = string.Format(OpenInIDEResources.Infobar_InvalidStateReason_WrongServer, configuration.Project.ServerUri);
+            }
+            else if (!string.IsNullOrEmpty(organizationKey) &&
+                (configuration.Project.Organization == null ||
+                 !organizationKey.Equals(configuration.Project.Organization.Key, StringComparison.OrdinalIgnoreCase)))
+            {
+                reason = string.Format(OpenInIDEResources.Infobar_InvalidStateReason_WrongOrganization, configuration.Project.Organization?.Key);
```
rita-g-sonarsource, line 101: I decided to use the same message for incorrect and null organization. Null organization is not supposed to happen, since you would have to be bound to the correct server (=sonarcloud) but with no organization, or we would have to get a request from SQ with SQ server and not-null organization. Both cases are not supposed to happen, so I am handling them but the user-message would be partial (`you must connect to the right organization. Currently connected to: `)

src/IssueViz.Security/OpenInIDE/OpenInIDEResources.resx
```diff
@@ -132,6 +132,26 @@
   <data name="Factory_Succeeded" xml:space="preserve">
     <value>[Open in IDE] Request listener created successfully. Listening on port {0}.</value>
   </data>
+  <data name="Infobar_InvalidStateReason_NotInConnectedMode" xml:space="preserve">
```
rita-g-sonarsource, line 135: I'm constructing the final message from multiple resources, so they would be:
`Could not handle Open in IDE request: <reason>. <insturctions>`
E.g.
`Could not handle Open in IDE request: you must connect to the right server. Currently connected to: https://sq-server.com. Please connect to https://sonarcloud.io/, organization sonarsource, and bind to project 'someproject'.`

Please review the messages and I'll share them with the rest of the team.

src/IssueViz.Security/HotspotsList/HotspotsToolWindow.cs
```diff
@@ -28,9 +28,11 @@
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList
 {
-    [Guid("4BCD4392-DBCF-4AA2-9852-01129D229CD8")]
+    [Guid(Guid)]
     public class HotspotsToolWindow : ToolWindowPane
     {
+        public const string Guid = "4BCD4392-DBCF-4AA2-9852-01129D229CD8";
```
duncanp-sonar, line 34: Minor, but Id/ToolWindowId would be a better name.

src/IssueViz.Security/OpenInIDE/Api/IOpenInIDEStateValidator.cs
```diff
@@ -41,4 +46,103 @@ internal interface IOpenInIDEStateValidator
...
+                : OpenInIDEResources.Inforbar_Instructions_SonarCloud;
+
+            var fullMessage = string.Format(OpenInIDEResources.Inforbar_InvalidState, reason, instructions);
+
+            return fullMessage;
+        }
+
+        private void AddInfoBar(string fullMessage)
+        {
+            currentInfoBar = infoBarManager.AttachInfoBar(new Guid(HotspotsToolWindow.Guid), fullMessage, default);
```
duncanp-sonar, line 124: Could 'Debug.Assert` here that the current bar is null

src/IssueViz.Security/OpenInIDE/Api/IOpenInIDEStateValidator.cs
```diff
@@ -41,4 +46,103 @@ internal interface IOpenInIDEStateValidator
...
+            if (reason == null)
+            {
+                return null;
+            }
+
+            var instructions = string.IsNullOrEmpty(organizationKey)
+                ? OpenInIDEResources.Inforbar_Instructions_SonarQube
+                : OpenInIDEResources.Inforbar_Instructions_SonarCloud;
+
+            var fullMessage = string.Format(OpenInIDEResources.Inforbar_InvalidState, reason, instructions);
```
duncanp-sonar, line 117: Typo: `Infor`...

src/IssueViz.Security/OpenInIDE/OpenInIDEResources.resx
duncanp-sonar: That looks like a lot of text to fit in the goldbar. How does it look in the IDE?

src/IssueViz.Security/OpenInIDE/OpenInIDEResources.resx
```diff
@@ -132,6 +132,27 @@
...
+  <data name="Infobar_InvalidStateReason_WrongOrganization" xml:space="preserve">
+    <value>you must connect to the right organization. Currently connected to: {0}</value>
+  </data>
+  <data name="Infobar_InvalidStateReason_WrongProject" xml:space="preserve">
+    <value>you must bind your solution to the right project. Currently bound to: {0}</value>
+  </data>
+  <data name="Infobar_InvalidStateReason_WrongServer" xml:space="preserve">
+    <value>you must connect to the right server. Currently connected to: {0}</value>
+  </data>
+  <data name="Inforbar_InvalidState" xml:space="preserve">
```
duncanp-sonar, line 147: This resource and the following two all have the typo mentioned previously ("Infor...")

src/IssueViz.Security/OpenInIDE/OpenInIDEResources.resx
```diff
@@ -132,6 +132,27 @@
...
+    <value>Could not handle Open in IDE request: {0}. {1}</value>
+  </data>
+  <data name="Inforbar_Instructions_SonarCloud" xml:space="preserve">
+    <value>Please connect to {0}, organization {1}, and bind to project '{2}'.</value>
```
duncanp-sonar, line 151: It looks a bit inconsistent having the project key in single quotes but not the organization key since they are both Sonar keys.
I think not quoting the server address is ok.

src/IssueViz.Security/OpenInIDE/Api/IOpenInIDEStateValidator.cs
```diff
@@ -41,4 +46,103 @@ internal interface IOpenInIDEStateValidator
...
+        }
+
+        private string GetFailureMessage(Uri serverUri, string projectKey, string organizationKey)
+        {
+            var configuration = configurationProvider.GetConfiguration();
+            string reason = null;
+
+            if (configuration.Mode == SonarLintMode.Standalone)
+            {
+                reason = OpenInIDEResources.Infobar_InvalidStateReason_NotInConnectedMode;
```
duncanp-sonar, line 91: Is this the message that will be shown if there isn't an open solution, and if so will it make sense to the user?

src/IssueViz.Security/OpenInIDE/OpenInIDEResources.resx
duncanp-sonar: If the solution is currently in connected mode but not bound to the requested server/org/project, then isn't it more likely that the action the user will need to take is to open a different solution, rather than change the current binding? At the moment, the wording seems to imply changing the binding.

rita-g-sonarsource: Regarding the wording: hmm I can see that now. What would you suggest? 
Maybe instead of `Please connect to..` - `Please open a solution that is connected to..`?

rita-g-sonarsource:

src/IssueViz.Security/OpenInIDE/Api/IOpenInIDEStateValidator.cs
```diff
@@ -41,4 +47,114 @@ internal interface IOpenInIDEStateValidator
...
+                : string.Format(OpenInIDEResources.RequestValidator_Instructions_SonarCloud, serverUri, organizationKey, projectKey);
+
+            var fullMessage = string.Format(OpenInIDEResources.RequestValidator_InvalidState, reason, instructions);
+
+            return fullMessage;
+        }
+
+        private void AddInfoBar()
+        {
+            currentInfoBar = infoBarManager.AttachInfoBarWithButton(new Guid(HotspotsToolWindow.ToolWindowId), OpenInIDEResources.RequestValidator_InfoBarMessage, "Show Output Window", default);
```
duncanp-sonar, line 130: Could assert that the currentInfoBar is null to start with.

src/IssueViz.Security/OpenInIDE/OpenInIDEResources.resx
```diff
@@ -132,6 +132,28 @@
...
     <value>[Open in IDE] Request listener created successfully. Listening on port {0}.</value>
   </data>
+  <data name="RequestValidator_InvalidStateReason_NotInConnectedMode" xml:space="preserve">
+    <value>you must be in Connected Mode</value>
+  </data>
+  <data name="RequestValidator_InvalidStateReason_WrongOrganization" xml:space="preserve">
+    <value>you must connect to the right organization. Currently connected to: {0}</value>
+  </data>
+  <data name="RequestValidator_InvalidStateReason_WrongProject" xml:space="preserve">
+    <value>you must bind your solution to the right project. Currently bound to: {0}</value>
```
duncanp-sonar, line 142: "correct" rather than "right" sounds better. Same for the server message.

src/IssueViz.Security/OpenInIDE/Api/IOpenInIDEStateValidator.cs
rita-g-sonarsource: Ah, my bad, apologies. I missed the PR comment.

src/IssueViz.Security/OpenInIDE/OpenInIDEResources.resx
duncanp-sonar, line 142: "you must bind your solution to..." still sounds to me like an instruction to change the current solution.

Something like "The current solution is not bound to the requested Sonar project. Requested Sonar project: foo. Bound Sonar project: bar" would be less directive.

Also, "project" is used by MSBuild/MS devs and SonarSource to mean different things. It's always worth specifying which one we mean to avoid confusion.

duncanp-sonar (approved):
Looks ok, just some comments about the messages.

More generally, I don't think the messages need to be so specific i.e. the server is wrong, the project is wrong etc. It would be enough to say "the requested issue belongs to a different Sonar project. Requested Sonar project: server/org/project. Currently bound project: server/org/project."

However, maybe we should just wait for feedback from the dogfood users.

Issue #1858, opened 2020-11-02
Modify SonarQubeClient and add hotspot text range


rita-g-sonarsource:
#### Description

The current API returns SonarQubeHotspot object that does not have all the text range properties. They are returned from the API in the following format:
```
	"textRange": {
		"startLine": 10,
		"endLine": 10,
		"startOffset": 28,
		"endOffset": 34
	},
```

TODO:
- [x] Add the missing properties (end line, start offset, end offset) to SonarQubeHotspot 
- [ ] Check what about line hashes

Issue #1857, opened 2020-11-02
Refactor IAnalysisIssueVisualizationConverter and separate conversion from span calculations


rita-gorokhod:
#### Description

IAnalysisIssueVisualization was used up until now only for live analysis issues, where the analyzed document was already open. This allowed us to calculate the relevant span when creating IAnalysisIssueVisualization.

However, hotspots are fetched from the server and the analyzed document doesn't have to be open. Therefore, there is no text buffer, and we cannot calculate the spans.

The current implementation of IAnalysisIssueVisualizationConverter does two things:
1. Creating an issue viz object 
2. Assigning spans to all locations in a given text buffer

The second part is not relevant for hotspots. In order to support hotspots, conditions were added to the code to disable this functionality. 
The class should instead be refactored and the span calculations logic should be taken out.

duncanp-sonar:
Closing as obsolete; there is now no special-case handling for hotspots in the converter.

PR #1859, opened 2020-11-02
Convert SQ hotspot into IAnalysisIssueVisualization

rita-g-sonarsource:
Part of #1763


src/IssueViz.Security/OpenInIDE/Api/IHotspotToIssueVisualizationConverter.cs
```diff
@@ -0,0 +1,98 @@
...
+            // todo: calculate file path
+            var filePath = sonarQubeHotspot.ComponentPath;
+            var priority = GetPriority(sonarQubeHotspot.VulnerabilityProbability);
+
+            var hotspot = new Hotspot(
+                filePath: filePath,
+                message: sonarQubeHotspot.Message,
+                startLine: sonarQubeHotspot.Line,
+                endLine: sonarQubeHotspot.Line,
+                startLineOffset: 0,
```
rita-g-sonarsource, line 68: Missing properties ticket: #1858

src/IssueViz.Security/OpenInIDE/Api/IHotspotToIssueVisualizationConverter.cs
```diff
@@ -0,0 +1,98 @@
...
+        public HotspotToIssueVisualizationConverter(IAnalysisIssueVisualizationConverter issueVisualizationConverter, IDocumentNavigator documentNavigator)
+        {
+            this.issueVisualizationConverter = issueVisualizationConverter;
+            this.documentNavigator = documentNavigator;
+        }
+
+        public IAnalysisIssueVisualization Convert(SonarQubeHotspot sonarQubeHotspot)
+        {
+            var hotspot = ConvertToHotspot(sonarQubeHotspot);
+            var textView = documentNavigator.Open(hotspot.FilePath);
```
rita-g-sonarsource, line 51: Ticket for making the span non-mandatory: #1857

src/IssueViz/Models/AnalysisIssueVisualizationConverter.cs
```diff
@@ -48,19 +48,26 @@ public AnalysisIssueVisualizationConverter(IIssueSpanCalculator issueSpanCalcula
             this.issueSpanCalculator = issueSpanCalculator;
         }
 
-        public IAnalysisIssueVisualization Convert(IAnalysisIssue issue, ITextSnapshot textSnapshot)
+        public IAnalysisIssueVisualization Convert(IAnalysisIssueBase issue, ITextSnapshot textSnapshot = null)
         {
-            var span = issueSpanCalculator.CalculateSpan(issue, textSnapshot);
+            var issueSpan = textSnapshot == null
```
rita-g-sonarsource, line 53: This is hacky - the real solution should be the separate between IssueViz creation, and span calculations, and this class currently does both. I created #1857 to handle this.

src/IssueViz.Security/OpenInIDE/Api/IHotspotToIssueVisualizationConverter.cs
```diff
@@ -0,0 +1,94 @@
...
+            var hotspot = ConvertToHotspot(sonarQubeHotspot);
+            var issueViz = issueVisualizationConverter.Convert(hotspot);
+
+            return issueViz;
+        }
+
+        private Hotspot ConvertToHotspot(SonarQubeHotspot sonarQubeHotspot)
+        {
+            // todo: calculate file path
+            var filePath = sonarQubeHotspot.ComponentPath;
```
rita-g-sonarsource, line 56: Since the file path is not translated, this will be the value that is shown in the data table.
When we attempt to navigate to it, LocationNavigator should fail because the file doesn't exist. The span should be then set to Empty and we could mark the location as non-navigable.

rita-g-sonarsource:
@duncanp-sonar as we discussed, this doesn't handle yet file paths translation.
Also, I have no idea what the coverage is like 😞

duncanp-sonar:
> Also, I have no idea what the coverage is like 😞

Probably more than zero ;-)

duncanp-sonar (approved):
LGTM

Issue #1865, opened 2020-11-04
Add IToolWindowShow service


duncanp-sonar:
#### Description

We now have multiple places in the code where we want to show a specific tool window: the "show secondary locations" command, the output window service, and in the Open in IDE request handler.

- [x] Create a simple MEF service to wrap showing an arbitrary tool window based on the GUID id
- [x] refactor the existing code and tests  

Required by #1798

PR #1867, opened 2020-11-04
Modified OutputWindowService to use IToolWindowService

duncanp-sonar:
Relates to #1865


src/Integration.UnitTests/Helpers/OutputWindowServiceTests.cs
```diff
@@ -68,61 +71,19 @@ public void Show_FailedToActivateSonarLintPane_OutputWindowNotShown()
...
-            {
-                testSubject.Show();
-            }
-
-            serviceProvider.VerifyAll();
-            serviceProvider.VerifyNoOtherCalls();
-        }
-
-        [TestMethod]
-        public void Show_ActivatedSonarLintPane_FailedToGetOutputWindow_OutputWindowNotShown()
```
duncanp-sonar, line 83: The tests are simpler now; they don't need to set up IVsUIShell since that is now hidden inside the tool window service.
The only thing the tests need to check now is whether the tool window service is called or not.

rita-g-sonarsource (approved):
🚀

PR #1868, opened 2020-11-04
Fix web listener ApiPath bug

duncanp-sonar:
Relates to #1825

### DIFF
--- a/src/IssueViz.Security/OpenInIDE/Http/OwinPipelineProcessor.cs
+++ b/src/IssueViz.Security/OpenInIDE/Http/OwinPipelineProcessor.cs
@@ -21,6 +21,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
+using System.Diagnostics;
 using System.Net;
 using System.Threading.Tasks;
 using Microsoft.Owin;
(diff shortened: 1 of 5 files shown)


src/IssueViz.Security/OpenInIDE/Http/OwinPipelineProcessor.cs
```diff
@@ -54,6 +55,7 @@ public OwinPipelineProcessor([ImportMany]IEnumerable<IOwinPathRequestHandler> pa
 
             foreach(var handler in pathRequestHandlers)
             {
+                Debug.Assert(handler.ApiPath.StartsWith("/"), "Expecting the handler path to start with /");
```
duncanp-sonar, line 58: This is to fix an issue I found when manually testing; the "show" request wasn't being passed to the handler class.

The Owin HttpListener is listening to all request on `http://localhost:[port]/sonarlint/api/` and it strips off that prefix before forwarding the request to the next handler in the pipeline. However, the `context.Request.Path` value it forwards always starts with '/'.

Originally, I had the HttpListener configured to listen to all requests on `localhost:[port]` - it was a late change to make it listen only to `sonarlint/api/`, and I can't have restarted manually in the IDE...

rita-g-sonarsource (approved):

PR #1870, opened 2020-11-05
Refactored the show Hotspot list command to use tool window service

duncanp-sonar:
* modified tool window service to force-create the window

Relates to #1865

### DIFF
--- a/src/Integration/Helpers/ToolWIndowService.cs
+++ b/src/Integration/Helpers/ToolWIndowService.cs
@@ -42,7 +42,10 @@ namespace SonarLint.VisualStudio.Integration.Helpers
 
         public void Show(Guid toolWindowId)
         {
-            var hr = shell.FindToolWindow(0, toolWindowId, out var windowFrame);
+            // We want VS to ask the package to create the tool window if it doesn't already exist
+            const uint flags = (uint)__VSFINDTOOLWIN.FTW_fForceCreate;
+
+            var hr = shell.FindToolWindow(flags, toolWindowId, out var windowFrame);
             Debug.Assert(ErrorHandler.Succeeded(hr), $"Failed to find tool window. Guid: {toolWindowId}, hr: {hr} ");
 
             if (ErrorHandler.Succeeded(hr))
--- a/src/IssueViz.Security.UnitTests/Commands/HotspotsToolWindowCommandTests.cs
+++ b/src/IssueViz.Security.UnitTests/Commands/HotspotsToolWindowCommandTests.cs
@@ -21,12 +21,11 @@
 using System;
 using System.ComponentModel.Design;
 using FluentAssertions;
-using Microsoft.VisualStudio.Shell;
-using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
-using Moq.Protected;
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.UnitTests;
 using SonarLint.VisualStudio.IssueVisualization.Security.Commands;
 using SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList;
 using ThreadHelper = SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Helpers.ThreadHelper;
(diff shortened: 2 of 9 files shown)


src/Integration/Helpers/ToolWIndowService.cs
```diff
@@ -42,7 +42,10 @@ public ToolWindowService([Import(typeof(SVsServiceProvider))] IServiceProvider s
 
         public void Show(Guid toolWindowId)
         {
-            var hr = shell.FindToolWindow(0, toolWindowId, out var windowFrame);
+            // We want VS to ask the package to create the tool window if it doesn't already exist
+            const uint flags = (uint)__VSFINDTOOLWIN.FTW_fForceCreate;
+
+            var hr = shell.FindToolWindow(flags, toolWindowId, out var windowFrame);
```
duncanp-sonar, line 48: Previously the service was passing 0 in the call to VS, which the documentation doesn't list as a valid value. I thought it was odd at the time, but it's what the previous code was doing so I didn't change it.

However, it does make a difference now, and "force create" is the behaviour we want.

It works like this: well-known tool window X may or may not exist. The code calling the shell doesn't know whether the window exists yet or not, and shouldn't  have to care. The calling code also doesn't know which package implements the tool window, and again it shouldn't have to.

Instead, the calling code calls shell with "force create" set. VS will check whether the window already exists. If it does, VS will just show it. If it doesn't, it will work out which package implements the tool window and ask it to create it. Then VS will show it.

I've manually tested it and it works the way we want. On the first call, VS will ask the package to create the window. On subsequent calls it doesn't.

src/IssueViz.Security.UnitTests/Commands/HotspotsToolWindowCommandTests.cs
```diff
@@ -82,90 +75,50 @@ public void Ctor_CommandAddedToMenu()
         }
 
         [TestMethod]
-        public void Execute_WindowNotCreated_NoException()
+        public void Execute_ServiceCalled()
```
duncanp-sonar, line 78: This and the following tests are completely new tests; I suggest ignoring the old tests completely and review them as new code.

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIdeStateValidatorTests.cs
```diff
@@ -112,7 +112,7 @@ public void CanHandleOpenInIDERequest_HasPreviousInfoBar_ConnectionIsNowCorrect_
             var infoBar = new Mock<IInfoBar>();
             var infoBarManager = new Mock<IInfoBarManager>();
             infoBarManager
-                .Setup(x => x.AttachInfoBarWithButton(new Guid(HotspotsToolWindow.ToolWindowId), It.IsAny<string>(), It.IsAny<string>(), default))
+                .Setup(x => x.AttachInfoBarWithButton(HotspotsToolWindow.ToolWindowId, It.IsAny<string>(), It.IsAny<string>(), default))
```
duncanp-sonar, line 115: Miscellaneous: I change the toolwindow to expose a Guid id property since that's what's needed in multiple places.

src/IssueViz.Security/Commands/HotspotsToolWindowCommand.cs
```diff
@@ -37,12 +36,13 @@ internal sealed class HotspotsToolWindowCommand
         
         public static HotspotsToolWindowCommand Instance { get; set; }
 
-        private readonly AsyncPackage package;
+        private readonly IToolWindowService toolWindowService;
```
duncanp-sonar, line 39: The package is now only used in the static initialization method to fetch the services and MEF components.

duncanp-sonar:
The coverage is showing as zero, but that's because of a bug in the version of our marketplace extension that was rolled out yesterday. It's just been rolled back to the previous version so coverage should be calculated correctly on future PRs.

rita-g-sonarsource (approved):

rita-g-sonarsource:
LGTM, but the analysis on SC seems odd - it lists the entire OpenInIDEStateValidator class with 0% coverage 🤔 
https://sonarcloud.io/component_measures?id=sonarlint-visualstudio&metric=new_coverage&pullRequest=1870&selected=sonarlint-visualstudio%3Asrc%2FIssueViz.Security%2FOpenInIDE%2FApi%2FIOpenInIDEStateValidator.cs&view=list

duncanp-sonar:
See my review comment: 
> The coverage is showing as zero, but that's because of a bug in the version of our marketplace extension that was rolled out yesterday. It's just been rolled back to the previous version so coverage should be calculated correctly on future PRs.

PR #1871, opened 2020-11-05
Modified "show flow window" command to use tool window service

duncanp-sonar:
Relates to #1865 

Apply the same changes as #1870 but this time for the "Show secondary locations commands"


src/IssueViz.UnitTests/Commands/IssueVisualizationToolWindowCommandTests.cs
```diff
@@ -127,75 +128,53 @@ public void ErrorListQueryStatus_CommandVisibilityIsSetToContextActivation(bool
         }
 
         [TestMethod]
-        public void Execute_WindowNotCreated_NoException()
+        public void Execute_ServiceCalled()
```
duncanp-sonar, line 131: Same as the other PR: the old tests are redundant.
The new tests were cut and paste from the PR and the names of the types being tested etc were changed. Other than that the tests are the same as the other PR.

I've manually tested these changes too.

PR #1877, opened 2020-11-06
Create AbsoluteFilePathLocator service

### LINKED ISSUE #1874: Create file path translation service
In order to support #1763, the relative path that we get back from the server should be converted to an absolute file path on disk.

rita-g-sonarsource:
Fixes #1874

### DIFF
--- a/src/Core/IAbsoluteFilePathLocator.cs
+++ b/src/Core/IAbsoluteFilePathLocator.cs
@@ -23,7 +23,7 @@ namespace SonarLint.VisualStudio.Core
     public interface IAbsoluteFilePathLocator
     {
         /// <summary>
-        /// Searches for all files in the current solution and returns an absolute file path that ends with the given relative path
+        /// Searches all files in the current solution and returns the absolute path of the first file in the solution that ends with the given relative path.
         /// </summary>
         /// <returns>Null if there is no solution open, or if there are no files in the solution that end with the given relative path</returns>
         string Locate(string relativeFilePath);
--- a/src/Core/IAbsoluteFilePathLocator.cs
+++ b/src/Core/IAbsoluteFilePathLocator.cs
@@ -1,0 +1,31 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Core
+{
+    public interface IAbsoluteFilePathLocator
+    {
+        /// <summary>
+        /// Searches for all files in the current solution and returns an absolute file path that ends with the given relative path
+        /// </summary>
+        /// <returns>Null if there is no solution open, or if there are no files in the solution that end with the given relative path</returns>
+        string Locate(string relativeFilePath);
+    }
+}
--- a/src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
+++ b/src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
@@ -785,12 +785,12 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
              *       file3
              * file4
              */
-            project.SetProperty((uint)VSConstants.VSITEMID.Root, (int)__VSHPROPID.VSHPROPID_FirstChild, file1);
-            project.SetProperty(file1, (int)__VSHPROPID.VSHPROPID_NextSibling, folder1);
-            project.SetProperty(folder1, (int)__VSHPROPID.VSHPROPID_NextSibling, file4);
-            project.SetProperty(folder1, (int)__VSHPROPID.VSHPROPID_FirstChild, file2);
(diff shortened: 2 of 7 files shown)


src/Integration/LocalServices/ProjectSystemHelper.cs
```diff
@@ -56,24 +56,19 @@ internal class ProjectSystemHelper : IProjectSystemHelper
...
         public ProjectSystemHelper(IServiceProvider serviceProvider)
         {
-            if (serviceProvider == null)
-            {
-                throw new ArgumentNullException(nameof(serviceProvider));
-            }
+            this.serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
 
-            this.serviceProvider = serviceProvider;
+            solution = this.serviceProvider.GetService<SVsSolution, IVsSolution>();
```
rita-g-sonarsource, line 65: I decided to fetch this in the ctor, so `EnumerateProjects` wouldn't have to get parameters

src/Integration/LocalServices/ProjectSystemHelper.cs
```diff
@@ -455,7 +450,7 @@ private string GetSolutionItemsFolderName()
             return solutionItemsFolderName;
         }
 
-        private static IEnumerable<IVsHierarchy> EnumerateProjects(IVsSolution solution)
+        public IEnumerable<IVsHierarchy> EnumerateProjects()
```
rita-g-sonarsource, line 453: I'm using this now in the new service

src/TestInfrastructure/Framework/ConfigurableVsProjectSystemHelper.cs
```diff
@@ -197,6 +198,21 @@ public bool IsLegacyProjectSystem(Project dteProject)
             return this.isLegacyProjectSystem;
         }
 
+        public IEnumerable<IVsHierarchy> EnumerateProjects()
```
rita-g-sonarsource, line 201: These aren't used in a test context

src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
```diff
@@ -52,7 +52,7 @@ public void TestInitialize()
         #region Tests
 
         [TestMethod]
-        public void ProjectSystemHelper_ArgCheck()
```
rita-g-sonarsource, line 55: Misc: I removed the `ProjectSystemHelper_` prefix from the tests

src/Core/IAbsoluteFilePathLocator.cs
```diff
@@ -0,0 +1,31 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Core
+{
+    public interface IAbsoluteFilePathLocator
+    {
+        /// <summary>
+        /// Searches for all files in the current solution and returns an absolute file path that ends with the given relative path
```
duncanp-sonar, line 26: I think it's worth calling out that the first match will be returned e.g.

> Returns the absolute path of the first file in the solution that ends with the given relative path

src/Integration/Helpers/AbsoluteFilePathLocator.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        public string Locate(string relativeFilePath)
+        {
+            if (relativeFilePath == null)
+            {
+                throw new ArgumentNullException(nameof(relativeFilePath));
+            }
+
+            foreach (var vsHierarchy in projectSystemHelper.EnumerateProjects())
```
duncanp-sonar, line 53: Does this need to be done on the UI thread?

src/Integration/LocalServices/ProjectSystemHelper.cs
```diff
@@ -502,5 +492,47 @@ public bool IsLegacyProjectSystem(Project dteProject)
...
+
+        private static IEnumerable<VSConstants.VSITEMID> ChildrenOf(IVsHierarchy hierarchy, uint rootId)
+        {
+            var result = new List<VSConstants.VSITEMID>();
+
+            for (var itemId = FirstChild(hierarchy, rootId);
+                itemId != (uint)VSConstants.VSITEMID.Nil;
+                itemId = NextSibling(hierarchy, itemId))
+            {
+                result.Add((VSConstants.VSITEMID) itemId);
```
duncanp-sonar, line 510: Did you decide against making this enumeration lazy?

duncanp-sonar: Ah - so the top-level projects are enumerated lazily, but all of the items in a project are fetched at once?
It's worth documenting this as it's not obvious.

duncanp-sonar (approved):

src/Integration/Helpers/AbsoluteFilePathLocator.cs
rita-g-sonarsource: Doesn't seem to be so; there are no warnings, and I ran it on a background thread and was able to access the objects without errors.

src/Integration/LocalServices/ProjectSystemHelper.cs
rita-g-sonarsource: I added a comment 👍

src/Integration/LocalServices/ProjectSystemHelper.cs
```diff
@@ -502,5 +492,47 @@ public bool IsLegacyProjectSystem(Project dteProject)
...
+        private static uint GetItemId(IVsHierarchy hierarchy, uint itemId, __VSHPROPID propId)
+        {
+            hierarchy.GetProperty(itemId, (int)propId, out var idObj);
+
+            return idObj == null ? (uint)VSConstants.VSITEMID.Nil : (uint)(int) idObj;
+        }
+
+        public string GetItemFilePath(IVsProject vsProject, VSConstants.VSITEMID itemId)
+        {
+            var hr = vsProject.GetMkDocument((uint)itemId, out var filePath);
```
rita-g-sonarsource, line 532: @duncanp-sonar I did some testing in a c++ project and I get a debug exception here.
It seems to be that in c++ projects, `GetAllItems` returns also all included files, such as framework `C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\shared\basetsd.h`. I get hundreds of files in my tiny project, whilst with DTE I get only the relevant ones. I'm not sure how to fix `GetAllItems` to ignore all the c++ dependencies. I'll look into it and won't merge this PR for now.

PR #1881, opened 2020-11-06
Implemented show hotspot api handler

duncanp-sonar:
Initial implementation of the show hotspots orchestrator.


src/IssueViz/Editor/LocationNavigator.cs
```diff
@@ -28,7 +28,7 @@
 
 namespace SonarLint.VisualStudio.IssueVisualization.Editor
 {
-    internal interface ILocationNavigator
+    public interface ILocationNavigator
```
duncanp-sonar, line 31: Now needs to be visible to the security assembly

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -20,26 +20,51 @@
...
-        public OpenInIDERequestHandler(IOpenInIDEStateValidator ideStateValidator, ISonarQubeService sonarQubeService, ILogger logger)
+        public OpenInIDERequestHandler(
+            IToolWindowService toolWindowService,
+            IOpenInIDEStateValidator ideStateValidator,
+            ISonarQubeService sonarQubeService,
+            IHotspotToIssueVisualizationConverter converter,
+            ILocationNavigator navigator,
+            IHotspotsStore hotspotsStore,
+            IOpenInIDEFailureInfoBar failueInfoBar,
+            ILogger logger)
```
duncanp-sonar, line 58: This component uses a lot of other services. Fortunately, the processing is relatively simple:
* always clear the previous info bar, if there was one
* always switch to the hotspots tool window
* stop processing the request in the event of any failure and show a gold bar

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -55,16 +80,71 @@ Task IOpenInIDERequestHandler.ShowHotspotAsync(IShowHotspotRequest request)
...
+            if (hotspot == null)
+            {
+                ShowFailureInfoBar();
+                return;
+            }
+
+            var hotspotViz = converter.Convert(hotspot);
+            if (hotspotViz == null)
+            {
+                logger.WriteLine(OpenInIDEResources.OpenHotspot_UnableToConvertHotspotData);
```
duncanp-sonar, line 114: This isn't a particularly useful message for the user. However, it will probably never occur in practice, and having the message removes any doubt about which stage the problem occurred at.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -55,16 +80,71 @@ Task IOpenInIDERequestHandler.ShowHotspotAsync(IShowHotspotRequest request)
...
-            // * set the selection
-            return Task.CompletedTask;
+            // Add to store regardless of whether navigation succeeded
+            hotspotsStore.Add(hotspotViz);
         }
+
+
+        private void ShowFailureInfoBar()
+        {
+            RunOnUIThread.Run(() => failureInfoBar.Show());
```
duncanp-sonar, line 132: Trying to show the gold bar on a background thread causes an exception (makes sense, because VS is trying to create WPF elements).
I'll add "CheckAccess" calls to the info bar manager in a separate PR.
Going forwards, we should consider making any new service we write that requires a UI thread/calls a VS service that does async.

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -19,72 +19,252 @@
...
+            stateValidatorMock = new Mock<IOpenInIDEStateValidator>();
+            serverMock = new Mock<ISonarQubeService>();
+            converterMock = new Mock<IHotspotToIssueVisualizationConverter>();
+            navigatorMock = new Mock<ILocationNavigator>();
+            storeMock = new Mock<IHotspotsStore>();
+            failureInfoBarMock = new Mock<IOpenInIDEFailureInfoBar>();
+            logger = new TestLogger(logToConsole: true);
+
+            testSubject = new OpenInIDERequestHandler(toolWindowService.Object, stateValidatorMock.Object, serverMock.Object,
+                converterMock.Object, navigatorMock.Object, storeMock.Object, failureInfoBarMock.Object, logger);
```
duncanp-sonar, line 106: Extra whitespace ...

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -19,72 +19,252 @@
...
+        private Mock<IHotspotToIssueVisualizationConverter> converterMock;
+        private Mock<ILocationNavigator> navigatorMock;
+        private Mock<IHotspotsStore> storeMock;
+        private Mock<IOpenInIDEFailureInfoBar> failureInfoBarMock;
+        private TestLogger logger;
+
+        private IOpenInIDERequestHandler testSubject;
+
+        [TestInitialize]
+        public void TestInitialize()
```
duncanp-sonar, line 89: These tests require a lot of setup, but the actual test cases are quite simple.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -55,16 +80,71 @@ Task IOpenInIDERequestHandler.ShowHotspotAsync(IShowHotspotRequest request)
...
+        }
+
+        private async Task ShowHotspotAsync(IShowHotspotRequest request)
+        {
+            logger.WriteLine(OpenInIDEResources.OpenHotspot_ProcessingRequest, request.ServerUrl, request.ProjectKey,
+                request.OrganizationKey ?? OpenInIDEResources.OpenHotspot_NullOrganization, request.HotspotKey);
+
+            // Always show the Hotspots tool window. If we can't successfully process the
+            // request we'll show a gold bar in the window
+            toolWindowService.Show(HotspotsToolWindow.ToolWindowId);
```
rita-g-sonarsource, line 93: `OpenInIDEFailureInfoBar` knows that it's attaching itself to the Hotspots ToolWindow, so I think it makes sense that it's also the one that opens it as well. Otherwise the logic is split: if we change the infobar to attach to a different window, we will need to remember to change this place as well.

Alternatively, we can change the infobar to receive the id of the toolwindow, so the orchestrator will be the only one who knows where the infobar is shown.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -55,16 +80,71 @@ Task IOpenInIDERequestHandler.ShowHotspotAsync(IShowHotspotRequest request)
...
             {
-                return Task.CompletedTask;
+                // We're assuming the validator will have output an explanantion of why the IDE
+                // isn't in the correct state
+                ShowFailureInfoBar();
+                return;
+            }
+
+            var hotspot = await TryGetHotspotData(request.HotspotKey);
+            if (hotspot == null)
```
rita-g-sonarsource, line 105: Now that I'm done with the converter, I can say that it won't return a null but it can throw an exception (e.g. if the SQ data is wrong, such as unsupported priority). 
I think it's best that the converter catch it and handle it. This way we're following the pattern that the callees are responsible for handling exceptions and the orchestrator only displays an infobar. WDYT? 
(If you agree I'll change the converter. Otherwise we should probably catch an exception here)

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -55,16 +80,71 @@ Task IOpenInIDERequestHandler.ShowHotspotAsync(IShowHotspotRequest request)
...
+            {
+                ShowFailureInfoBar();
+                return;
+            }
+
+            var hotspotViz = converter.Convert(hotspot);
```
rita-g-sonarsource, line 111: LocationNavigator has no logging at the moment, and I'm debating if we should add it, at least for opening the document. For hotspots, it's possible that we couldn't resolve the file path -- I guess we could guess it based on the fact that this log will output a partial file path, but maybe we should have something extra in locationNavigator, and then here just display an infobar. WYDT?

rita-g-sonarsource, line 111: Now that I'm done with the converter I can say that it won't return a null, but it can throw an exception (e.g. if the SQ data is wrong, such as unsupported priority). I think it's best that the converter catches it and handles it. This way we're following the pattern that the callees are responsible for handling exceptions and the orchestrator only displays an infobar.
WDYT?
If you agree, I'll change the converter, otherwise we'll need to handle exceptions here.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -55,16 +80,71 @@ Task IOpenInIDERequestHandler.ShowHotspotAsync(IShowHotspotRequest request)
...
+            if (hotspotViz == null)
+            {
+                logger.WriteLine(OpenInIDEResources.OpenHotspot_UnableToConvertHotspotData);
+                ShowFailureInfoBar();
+                return;
+            }
+
+            if (!navigator.TryNavigate(hotspotViz))
+            {
+                logger.WriteLine(OpenInIDEResources.ApiHandler_FailedToNavigateToHotspot, hotspotViz.FilePath, hotspotViz.StartLine);
```
rita-g-sonarsource, line 121: LocationNavigator has no logging at the moment, and I'm debating if we should add it, at least for opening the document. For hotspots, it's possible that we couldn't resolve the file path -- and I guess we could guess it based on the fact that this log will output a partial file path, but maybe we should instead do the logging in locationNavigator, and then here just display an infobar. WYDT?
It might be a bit noisy for secondary locations, when code is being modified, so that's why I'm thinking of adding a log only for opening the document.

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -19,72 +19,251 @@
...
+using SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList;
+using SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableDataSource;
 using SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Api;
 using SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Contract;
 using SonarQube.Client;
+using SonarQube.Client.Models;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE.Api
 {
     [TestClass]
```
rita-g-sonarsource, line 41: This is a new convention, any reason for having separate test classes?

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -19,72 +19,251 @@
...
         }
 
-        private static Mock<IOpenInIDEStateValidator> CreateValidator(string uri, string projectKey, string orgKey, bool canHandleRequest)
+        [TestMethod]
+        public async Task ShowHotspot_ConversionFailed_NoFurtherProcessing()
         {
-            var mockValidator = new Mock<IOpenInIDEStateValidator>();
-            mockValidator.Setup(x => x.CanHandleOpenInIDERequest(It.Is<Uri>(x => x.ToString() == uri), projectKey, orgKey))
+            InitializeStateValidator(ValidRequest, true);
+            SetServerResponse(ValidRequest, ValidServerHotspot);
```
rita-g-sonarsource, line 151: I guess mock-wise it's redundant to say
`SetConversionResponse(ValidServerHotspot, null);`  but I think it's worth adding anyway, for readability.

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -19,72 +19,251 @@
...
+        }
+
+        [TestMethod]
+        public async Task ShowHotspot_DataIsValid_NavigationSucceeded_And_IssueAddedToStore()
+        {
+            // Note: this test needs a viz mock that doesn't have any members mocked using Setup(...).
+            // This is because we don't expect any of the members to be called (and if the mock has any
+            // members mocked then the call to converterMock.VerifyAll() will fail when the library tries
+            // to transitively verify the viz members).
+            var hotspotVizMock = new Mock<IAnalysisIssueVisualization>();
```
rita-g-sonarsource, line 170: I'm a bit confused by the comment.
What would happen if we use `Mock.Of<IAnalysisIssueVisualization>()` ?

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -19,72 +19,251 @@
...
+            SetStoreExpectedItem(hotspotVizMock.Object);
+
+            // Act
+            await testSubject.ShowHotspotAsync(ValidRequest)
+                .ConfigureAwait(false);
+
+            CheckInfoBarCleared();
+            CheckInfoBarNotShown();
+            CheckCalled(toolWindowService, stateValidatorMock, serverMock, converterMock, navigatorMock, storeMock);
+            CheckNotCalled(hotspotVizMock); // shouldn't have accessed any of the members
```
rita-g-sonarsource, line 185: I'm not sure it really matters that it was accessed, no?

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -19,72 +19,251 @@
...
+            CheckInfoBarCleared();
+            CheckInfoBarShown();
+            CheckCalled(toolWindowService, stateValidatorMock, serverMock, converterMock, navigatorMock, storeMock);
+            logger.AssertPartialOutputStringExists(hotspotFilePath, hotspotStartLine.ToString());
+        }
+
+        private static IAnalysisIssueVisualization CreateHotspotVisualization(string filePath, int startLine)
+        {
+            var vizMock = new Mock<IAnalysisIssueVisualization>();
+            // Note: these properties are not marked as Verifiable() since they won't always be read
```
rita-g-sonarsource, line 216: I'm not sure why we need to verify this class at all

rita-g-sonarsource (commented):
@duncanp-sonar left some comments

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
duncanp-sonar: Just something I was trying out; organising tests based on shared setup i.e. if a test class has a TestInitialize then only tests that need that initialization should be in the class. Tests that are completely isolated are grouped into a class that doesn't have a TestInitialize.
It seems like a reasonable approach, except that it seems to be the case that the only completely isolated test is the MEF construction test...

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
duncanp-sonar: That would work; the only reason to have a mock was to check it wasn't used.

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
duncanp-sonar: So we can assert in the navigation failure case that the info about the failed navigation is logged.

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
duncanp-sonar: It's a weak check; the only time we expect the orchestrator to access the properties is to provide a detailed failure i.e. we don't expect the orchestrator to need read the data in this case.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
duncanp-sonar: I prefer the second option since we always show the hotspots window, even in success cases.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
duncanp-sonar: I don't have a strong preference, and the orchestrator is already handling exceptions from the Sonar web client.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
duncanp-sonar: I was expecting we'd have to change the logging once we've started dogfood and can see how it works in practice. I saw there wasn't any logging in the navigator, and for now I just wanted to log something to give us a starting point.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
duncanp-sonar: The converter doesn't currently log anything, and getting bad data from the server does seem like an exceptional condition. Maybe it's better if we keep the converter simple and let it throw.
I'll add exception handling to the orchestrator. We can always change it later if necessary.

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
duncanp-sonar: I'll make that change in a separate PR.

PR #1882, opened 2020-11-09
Create AbsoluteFilePathLocator service

### LINKED ISSUE #1874: Create file path translation service

rita-g-sonarsource:
Fixes #1874

### DIFF
--- a/src/Integration/Helpers/AbsoluteFilePathLocator.cs
+++ b/src/Integration/Helpers/AbsoluteFilePathLocator.cs
@@ -20,7 +20,6 @@
 
 using System;
 using System.ComponentModel.Composition;
-using System.IO;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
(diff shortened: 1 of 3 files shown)


src/Integration/Helpers/AbsoluteFilePathLocator.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        public string Locate(string relativeFilePath)
+        {
+            if (relativeFilePath == null)
+            {
+                throw new ArgumentNullException(nameof(relativeFilePath));
+            }
+
+            relativeFilePath = relativeFilePath.TrimStart('\\');
```
duncanp-sonar, line 48: Do we need to support forward slashes as well, or does the API handle them? I'm not sure what we'll get back from SonarQube for a C#/C++ file that is analysed on Linux machine.

duncanp-sonar (approved):

src/Integration/Helpers/AbsoluteFilePathLocator.cs
rita-g-sonarsource: I'd expect the API to handle it, similarly to the way it does for issues, but we could verify it with SQ team when the hotspots api is finalized. 

https://github.com/SonarSource/sonarlint-visualstudio/blob/5fb393022542266ae2e8f93b362539ecfc127796/src/Integration/Suppression/SonarQubeIssuesProvider.cs#L128

Issue #1798, opened 2020-10-14
Handle SQ open issue request


rita-g-sonarsource:
#### Description

Handle a request coming from SQ:

- [x] Verify that the request can be handled #1797
- [x] Fetch SQ hotspot #1771
~~- [ ] Fetch file hashes ?~~
~~- [ ] Fetch source code ? #1762~~
- [x] Convert SQ issue into IssueViz #1763
- [x] Push IssueViz into the hotspots data source #1795
- [x] Bring the hotspot list to front

PR #1886, opened 2020-11-09
Changed failure info toolbar to accept the target tool window id

duncanp-sonar:
Related to #1798

### DIFF
--- a/src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
+++ b/src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
@@ -128,7 +128,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Api
 
         private void ShowFailureInfoBar()
         {
-            RunOnUIThread.Run(() => failureInfoBar.Show());
+            RunOnUIThread.Run(() => failureInfoBar.Show(HotspotsToolWindow.ToolWindowId));
         }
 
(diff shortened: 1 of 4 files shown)


src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -128,7 +128,7 @@ private async Task ShowHotspotAsync(IShowHotspotRequest request)
 
         private void ShowFailureInfoBar()
         {
-            RunOnUIThread.Run(() => failureInfoBar.Show());
+            RunOnUIThread.Run(() => failureInfoBar.Show(HotspotsToolWindow.ToolWindowId));
```
rita-g-sonarsource, line 131: I saw the issue regarding the Clear.
I think that maybe the failureInfoBar should return a task and switch to the UI thread itself, instead of relying on the callers to do it.

rita-g-sonarsource (approved):

src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
duncanp-sonar: Yep; as I said in one of the PRs yesterday, I think we should consider making any new service that consumes VS services async.

As it happens, the failure info bar doesn't directly consume any VS services - it only consumes our services, so ideally our existing info bar manager should handle the threading requirements. I'll have a look, but that might involve too much change.

PR #1889, opened 2020-11-10
Add ServerFilePath property to Hotspot

rita-g-sonarsource:


src/IssueViz.Security/OpenInIDE/Api/IHotspotToIssueVisualizationConverter.cs
```diff
@@ -55,7 +55,7 @@ public IAnalysisIssueVisualization Convert(SonarQubeHotspot sonarQubeHotspot)
 
         private Hotspot ConvertToHotspot(SonarQubeHotspot sonarQubeHotspot)
         {
-            var filePath = absoluteFilePathLocator.Locate(sonarQubeHotspot.ComponentPath);
+            var filePath = absoluteFilePathLocator.Locate(sonarQubeHotspot.ComponentPath) ?? sonarQubeHotspot.ComponentPath;
```
rita-g-sonarsource, line 58: Noticed this bug yesterday.
I always set the value to whatever we get back from the locator, which can be null if the file is not found.
We then try to navigate to it, and fail, but the message is not descriptive ("Failed to open file ''", or something of the sort), and there is no file path in the hotspots table.
So if it's null, I just keep the original value.
The navigation will still fail and it will be more descriptive.

duncanp-sonar, line 58: Didn't we discuss the possibility of having separate properties for the server path and the local path (I think when we discussing the retry handling)? Overloading a single property is potentially confusing.

duncanp-sonar (approved):
Looks ok for a bug fix, but separate properties seems cleaner.

src/IssueViz.Security/OpenInIDE/Api/IHotspotToIssueVisualizationConverter.cs
rita-g-sonarsource: I thought about it but I didn't find a real use case for it. Either we find it, in which case we will always use the absolute path, or we don't, and then we will always use the relative path. I couldn't find a use case where we'd want both, expect maybe to distinguish between multiple results - but even then, it would only for troubleshooting and can easily be retrieved from SQ UI.

duncanp-sonar: I think it makes a difference for the consumer. With this change, the caller can't easily tell if the path is supposed to be a usable local path or not. Separate properties for `ServerPath` and `LocalPath` would make the distinction clearer; I wouldn't accidentally try to use `ServerPath` to access a local file, and it's clearer that that a `null` `LocalPath` means the file couldn't be resolved.

We could keep one property and document what it means e.g. `The local path to the file if the file could be resolved, otherwise the relative server path`. However, even writing that comment feels wrong.

rita-g-sonarsource: What would the consumer do with it?

If it's just for logging, I think it's a bit odd. The orchestrator could do something like:
```
var hotspot = Converter.Convert(sqHotspot);
if (string.IsNullOrEmpty(hotspot.LocalPath)) 
{
    logger.WriteLine(... hotspot.LocalPath, hotspot.ServerPath);
}
```
and other than that, `ServerPath` will never be used.

If it's not just for logging, then the orchestrator might as well abort the navigation, since it already knows it's not going to work. But that feels a bit too low-level. I don't think the orchestrator should even know that the converter does a file path translation -- it might very well be that `LocalPath` is null because the navigator will be the one to calculate it, for example.

So anyway I don't know. Maybe instead I should add logging in the converter? `Resolved SQ hotspot file path '{0}' to '{1}'`

duncanp-sonar: > Then again, I really see your point that it's odd that a field can be 2 things. If it can be 2 things, it should be 2 fields...

For me, that's the most important consideration.  It's a design smell that's hard to see and makes the code harder to understand and maintain. We already have other cases where it's not obvious what a "path" property represents; local, server, relative, absolute, Window, Linux... We shouldn't have to spend time looking at a property try to work out what the data means.

If we want to keep things simple for UI purposes, we could add a calculated 'DisplayPath' property. 
If we're using a column in the hotspots table for two different purposes then that's a design smell too, and we should fix it rather than trying to hide it at a lower level.

rita-g-sonarsource: > If we're using a column in the hotspots table for two different purposes then that's a design smell too, and we should fix it rather than trying to hide it at a lower level.

How would you fix it? And how would you handle the navigation logging?

I don't mind adding the field, but when I try to visualize the usages, they don't make sense.

duncanp-sonar: The logging is straightforward; either log both path fields, or if it's documented as part of the converter contract that "local path will be null if it can't be resolved" then just the server path.

For the table, we haven't got as far as looking at how to properly handle non-navigable issues yet, so I'm not sure what the correct way to handle it will be. For now, it's purely an UI issue, so I'd either have the "if empty" logic or add a calculated property to return a display string.

Adding a calculated display string property seems like a reasonable approach for callers who don't care if the path is valid or not, but it preserves the underlying data for cases where it does matter e.g. the "retry" scenario.

rita-g-sonarsource: Regarding the retry scenario - We would be running the entire process from scratch, so as long as the SQ hotspot is unchanged, we can re-run the converter. The converted IssueViz is of no use here.

PR #1898, opened 2020-11-10
Add navigation functionality to hotspots list

### LINKED ISSUE #1794: Add navigation functionality to hotspots list
#### Description

Add functionality to the hotspots list:
- [x] Selecting an item should navigate to the issue primary location
- [ ] ~~Check if the file/line hashes are the same?~~ --> will be done as part of the navigation logic
- [ ] ~~If the IssueViz has secondary locations - selecting an item should set the selection in the selection service.~~ --> will be done in #1903

rita-g-sonarsource:
Fixes #1794


src/IssueViz.Security/HotspotsList/TableDataSource/Events/HotspotsTableEventProcessor.cs
```diff
@@ -0,0 +1,54 @@
...
+    internal class HotspotsTableEventProcessor : TableControlEventProcessorBase
+    {
+        private readonly ILocationNavigator locationNavigator;
+
+        public HotspotsTableEventProcessor(ILocationNavigator locationNavigator)
+        {
+            this.locationNavigator = locationNavigator;
+        }
+
+        public override void PreprocessNavigate(ITableEntryHandle entry, TableEntryNavigateEventArgs e)
```
rita-g-sonarsource, line 37: @duncanp-sonar I'm about to lose my sanity.

This is a super convenient [API](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.tablecontrol.iwpftablecontrol2.navigationbehavior?view=visualstudiosdk-2019#Microsoft_VisualStudio_Shell_TableControl_IWpfTableControl2_NavigationBehavior). All you need to do is cast the [created IWpfTableControl](https://github.com/SonarSource/sonarlint-visualstudio/blob/branch-hotspots/src/IssueViz.Security/HotspotsList/HotspotsControl.xaml.cs#L47) to `IWpfTableControl2`, and set the navigation behavior, and then this method will be magically called when the nagivation behavior is triggered.

But does it work in 2015? Of course it doesn't 😠 

```
Error CS1061: 'IWpfTableControl2' does not contain a definition for 'NavigationBehavior' and no accessible extension method 'NavigationBehavior' accepting a first argument of type 'IWpfTableControl2' could be found (are you missing a using directive or an assembly reference?)
```

And why is that? Because the god damn internal control that we use is implementing in shell 14 `Microsoft.Internal.VisualStudio.Shell.TableControl.IWpfTableControl2`, and not `Microsoft.VisualStudio.Shell.TableControl.IWpfTableControl2`. So I cannot set this property in 2015 since the internal control is not implementing the right interface.

However - the default behavior does seem to be what we want anyway (double click+enter key). I tested that in 2015 and 2019 and it works even though I'm not specifying it explicitly.

I don't know. An alternative is to catch key button down + mouse click events, like I did in a [previous commit](https://github.com/SonarSource/sonarlint-visualstudio/pull/1898/commits/9e2ffc8ca5a2f82f04f696613c89303bdea8d838).
That however had problems of its own: I cannot mock a mouse double click event without using reflection.

So what do you think? Rely on the default non-documented behavior? Or use plain events?

duncanp-sonar: I think it's ok to rely on the default behaviour; it's very unlikely to change.

src/IssueViz.Security/HotspotsList/TableDataSource/Events/HotspotsTableEventProcessor.cs
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+using Microsoft.VisualStudio.Shell.TableControl;
+using Microsoft.VisualStudio.Shell.TableManager;
+using SonarLint.VisualStudio.IssueVisualization.Editor;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableDataSource.Events
+{
+    internal class HotspotsTableEventProcessor : TableControlEventProcessorBase
```
duncanp-sonar, line 28: Separate issue: we should change our existing error list event processor to use this base class.

src/IssueViz.Security/HotspotsList/TableDataSource/Events/HotspotsTableEventProcessor.cs
rita-g-sonarsource: Ok, then please review 🙂

src/IssueViz.Security.UnitTests/HotspotsList/HotspotsControlTests.cs
```diff
@@ -45,6 +46,18 @@ public void Ctor_HotspotsListIsWpfTableControl()
             testSubject.hotspotsList.Child.Should().Be(uiControl);
         }
 
+        [TestMethod]
```
rita-g-sonarsource, line 49: I had tests for the navigation as well but had to take them out.. left this one though. It's unrelated to the PR.

duncanp-sonar (approved):

PR #1899, opened 2020-11-10
Added RunOnUIThread.RunAsync

### LINKED ISSUE #1895: Change RunOnUIThread to return the awaitable task
### Description

The class `RunOnUIThread` switches to VS's UI thread but does not return the awaitable task.
This doesn't allow the callers to use async/await on the result.

TODO:
- [x] Investigate how to return the awaitable task.

duncanp-sonar:
Fixes #1895

### DIFF
--- a/src/IssueViz.UnitTests/Helpers/RunOnUIThreadTests.cs
+++ b/src/IssueViz.UnitTests/Helpers/RunOnUIThreadTests.cs
@@ -1,0 +1,172 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.Threading;
+using Moq;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Helpers;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Helpers
+{
+    [TestClass]
+    public class RunOnUIThreadTests
+    {
+        // These tests are slightly hacky in that they use "ThreadHelper.SetCurrentThreadAsUIThread()"
+        // to fool the VS thread helper into thinking the current thread is the UI thread.
+        // 
+        // However, these tests are quite limited as we can only test the cases where the
+        // caller is already on the main thread - we're not hacking/mocking/faking enough 
+        // of the VS setup for a call to "ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync()"
+        // to succeed.
+
+        [TestMethod]
+        public void Run_StartsOnUIThread_ExecuteOnUIThread()
+        {
+            ThreadHelper.SetCurrentThreadAsUIThread();
+
+            var initialThread = Thread.CurrentThread.ManagedThreadId;
+            int executionThreadId = -1;
+
+            // Act
+            RunOnUIThread.Run(() => executionThreadId = Thread.CurrentThread.ManagedThreadId);
+
+            executionThreadId.Should().Be(initialThread);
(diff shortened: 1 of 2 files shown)


src/IssueViz.UnitTests/Helpers/RunOnUIThreadTests.cs
```diff
@@ -0,0 +1,172 @@
...
+using Moq;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Helpers;
+
+namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Helpers
+{
+    [TestClass]
+    public class RunOnUIThreadTests
+    {
+        // These tests are slightly hacky in that they use "ThreadHelper.SetCurrentThreadAsUIThread()"
```
rita-g-sonarsource, line 36: Cosmetics: this might be better as a class doc comment.
Otherwise it looks a bit like it's relevant only for the first test case.

src/IssueViz.UnitTests/Helpers/RunOnUIThreadTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+
+    [TestClass]
+    public class RunOnUIThreadTests_Hacky
+    {
+        // These tests are quite hacky in that they use "ThreadHelper.SetCurrentThreadAsUIThread()",
+        // and also use reflection to set private data fields in the VS ThreadHelper hierarchy.
+        // As such, they will break if the underlying VS implementation changes.
+        // to fool the VS thread helper into thinking the current thread is the UI thread.
```
rita-g-sonarsource, line 81: This line is in the wrong place

src/IssueViz.UnitTests/Helpers/RunOnUIThreadTests.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            executionThreadId.Should().Be(initialThread);
+            Thread.CurrentThread.ManagedThreadId.Should().Be(initialThread); // should still be on the "UI" thread
+        }
+    }
+
+    [TestClass]
+    public class RunOnUIThreadTests_Hacky
+    {
+        // These tests are quite hacky in that they use "ThreadHelper.SetCurrentThreadAsUIThread()",
```
rita-g-sonarsource, line 78: same cosmetics: can be a doc comment

src/IssueViz.UnitTests/Helpers/RunOnUIThreadTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            // Act
+            await RunOnUIThread.RunAsync(() => executionThreadId = Thread.CurrentThread.ManagedThreadId);
+
+            executionThreadId.Should().Be(initialThread);
+            Thread.CurrentThread.ManagedThreadId.Should().Be(initialThread); // should still be on the "UI" thread
+        }
+    }
+
+    [TestClass]
+    public class RunOnUIThreadTests_Hacky
```
rita-g-sonarsource, line 76: Maybe "_VsMocking"?

src/IssueViz.UnitTests/Helpers/RunOnUIThreadTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        // "ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync()" won't fail.
+        // It won't actually cause a thread transition, but we can use the fact that our
+        // scheduler mock was called to detect whether a thread transition was requested.
+
+        private Mock<IVsTaskSchedulerService2> schedulerMock;
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            ThreadHelper.SetCurrentThreadAsUIThread();
```
rita-g-sonarsource, line 93: Do we need this for all the tests?

For tests like `Run_StartsOnBackgroundThread_ExecuteOnUIThread` we can avoid starting a new task if we don't do this.

rita-g-sonarsource (approved):

src/IssueViz.UnitTests/Helpers/RunOnUIThreadTests.cs
duncanp-sonar: Unfortunately not. Each test passes if you run it individually, but not if you run them as a group. It's because the 'ThreadHelper' test code is setting a static, any we currently have no way of resetting the effect. So to maintain isolation it's better if each test explicitly sets what is considered the UI thread for that test.

I can improve the test isolation slightly for the really hacky tests by making them reset the field they are setting, so I'll do that. We might be able to do something similar for the test ThreadHelper, but I won`t do that now.

src/IssueViz.UnitTests/Helpers/RunOnUIThreadTests.cs
duncanp-sonar: "Mocking" is a normal, safe thing to do in tests. I want it to be clear from the name that these tests are doing something non-standard and ugly. `RunUIThreadTests_HackVsInternals`?

rita-g-sonarsource: Yep sounds good

src/IssueViz.UnitTests/Helpers/RunOnUIThreadTests.cs
rita-g-sonarsource: Ah, I see. 
Maybe we should change ThreadHelper to return a disposable scope, like AssertIgnoreScope. I'll open a ticket.

duncanp-sonar: FYI I've added a disposable scope that deals with the `cachedServiceInstance` field, but I haven't tried to unset whatever is set by the ThreadHelper call.

PR #1907, opened 2020-11-11
Select requested hotspot in the hotspots list

### LINKED ISSUE #1802: Auto-navigate to location and auto-select in hotspot list
#### Description

When a hotspot request is received, 
- [x] Navigate automatically to the hotspot location
- [x] Select the hotspot in the hotspots list

rita-g-sonarsource:
Fixes #1802

### DIFF
--- a/src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
+++ b/src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
@@ -19,6 +19,7 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
@@ -33,6 +34,7 @@ using SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList;
 using SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableDataSource;
 using SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Api;
 using SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Contract;
+using SonarLint.VisualStudio.IssueVisualization.Security.SelectionService;
 using SonarQube.Client;
 using SonarQube.Client.Models;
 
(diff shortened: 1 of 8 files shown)


src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -237,17 +243,41 @@ public async Task ShowHotspot_DataIsValid_NavigationFailed_IssueStillAddedToStor
...
 
             CheckInfoBarCleared();
             CheckInfoBarShown();
-            CheckCalled(toolWindowService, stateValidatorMock, serverMock, converterMock, navigatorMock, storeMock);
+            CheckCalled(toolWindowService, stateValidatorMock, serverMock, converterMock, navigatorMock, storeMock, selectionServiceMock);
             logger.AssertPartialOutputStringExists(hotspotFilePath, hotspotStartLine.ToString());
         }
 
+        [TestMethod]
+        public async Task ShowHotspot_DataIsValid_IssueFirstAddedToStoreAndThenSelected()
```
rita-g-sonarsource, line 259: I'm not a fan of this.
I don't know if it's better to use instead something like `IHotspotsStore.AddAndSelect`.

duncanp-sonar: Yep; depends on what the overall selection/sync big picture looks like.

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml.cs
```diff
@@ -54,8 +62,30 @@ private static IWpfTableControl CreateWpfTableControl(ITableManagerProvider tabl
...
         }
 
+        private async void SelectionService_SelectionChanged(object sender, SelectionService.SelectionChangedEventArgs e)
+        {
+            if (e.SelectedHotspot == null)
+            {
+                return;
+            }
+
+            await wpfTableControl.ForceUpdateAsync();
```
duncanp-sonar, line 72: This needs a comment to explain why it's necessary (particularly since the docs say _"There should be no need to call this method (updates will happen automatically if there is a change). It is exposed here primarily to simplify testing."_)

PR #1908, opened 2020-11-11
Add navigability column to hotspots list

### LINKED ISSUE #1804: Display navigability status in the hotspots list
e.g. can't find the file
can't find the issue in the file

rita-g-sonarsource:
Fixes #1804 

image


src/IssueViz.Security/HotspotsList/TableDataSource/CustomColumns/NavigabilityTableColumnDefinition.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public override string DisplayName { get; } = "";
+
+        public override double DefaultWidth { get; } = Width;
+
+        public override bool IsFilterable { get; } = false;
+
+        public override bool IsSortable { get; } = false;
+
+        public override bool TryCreateToolTip(ITableEntryHandle entry, out object toolTip)
```
rita-g-sonarsource, line 47: I preferred to place the entire navigability logic here, instead of in the general HotspotTableEntry. This way the logic is isolated and HotspotTableEntry doesn't repeatedly call these methods for the wrong columns.

src/IssueViz.Security/HotspotsList/TableDataSource/CustomColumns/PriorityTableColumnDefinition.cs
```diff
@@ -29,13 +29,12 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableD
     internal class PriorityTableColumnDefinition : TableColumnDefinitionBase
     {
         public const string ColumnName = "ReviewPriority";
+        public const int Width = 120;
```
rita-g-sonarsource, line 32: Minor change: the column was a bit too small for the current values

src/IssueViz.Security/HotspotsList/TableDataSource/CustomColumns/PriorityTableColumnDefinition.cs
```diff
@@ -29,13 +29,12 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableD
...
 
         public override string Name { get; } = ColumnName;
 
-        public override double DefaultWidth => 50;
+        public override string DisplayName { get; } = "Priority";
 
-        public override string DisplayName => "Priority";
-
-        public override bool IsFilterable => true;
```
rita-g-sonarsource, line 39: IsFilterable=true is the default value, so no need to specify it.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableColumns.cs
```diff
@@ -29,8 +29,9 @@ internal static class HotspotsTableColumns
     {
         public static IReadOnlyList<ColumnState> InitialStates { get; } = new[]
         {
-            new ColumnState(StandardTableColumnDefinitions.ErrorCode, true, 50),
-            new ColumnState(PriorityTableColumnDefinition.ColumnName, true, 100),
+            new ColumnState(NavigabilityTableColumnDefinition.ColumnName, true, NavigabilityTableColumnDefinition.Width),
```
rita-g-sonarsource, line 32: No copy-paste of the width

src/IssueViz.Security/HotspotsList/TableDataSource/CustomColumns/PriorityTableColumnDefinition.cs
```diff
@@ -29,13 +29,12 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableD
     internal class PriorityTableColumnDefinition : TableColumnDefinitionBase
     {
         public const string ColumnName = "ReviewPriority";
+        public const int Width = 120;
 
         public override string Name { get; } = ColumnName;
 
-        public override double DefaultWidth => 50;
+        public override string DisplayName { get; } = "Priority";
```
rita-g-sonarsource, line 36: Minor unrelated change: changed to single initialization as the value will never change.

src/IssueViz.Security/Resources.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="ERR_CannotNavigate" xml:space="preserve">
+    <value>Cannot navigate to step. The source code is different from the analyzed version</value>
```
duncanp-sonar, line 121: I think `Location` would be better than `step`, and since there is more than one sentence the second sentence should have a full stop tool.
Also, it helps to prefix the resource name with the component that uses it, particularly since this is in the top-level resource file.

src/IssueViz.Security/HotspotsList/TableDataSource/CustomColumns/PriorityTableColumnDefinition.cs
duncanp-sonar: Perhaps not from a functionality point of view, but it doesn't help the reader unless they happen to know it is the default...

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableColumns.cs
duncanp-sonar: I don't understand your comment.

rita-g-sonarsource: The width value was copy-pasted: here and in the column definition. So now I simply reference the value in the column definition.

src/IssueViz.Security/HotspotsList/TableDataSource/CustomColumns/PriorityTableColumnDefinition.cs
rita-g-sonarsource: There are other properties, such as IsSortable, IsResizable, IsCopyable - it's noisy to repeat them and just set to default values.

src/IssueViz.Security/HotspotsList/TableDataSource/CustomColumns/NavigabilityTableColumnDefinition.cs
duncanp-sonar: Localising the code makes sense, but why would the HotspotTableEntry "repeatedly call these methods for the wrong columns"?

rita-g-sonarsource: It checks for each column if it has a tooltip/image. Currently the base class has an empty implementation for them.
If I were to put this code in the hotspots table, then it'd have to have a break `if (columnName != "Navigability column name") return;`. Just noisy. This is cleaner.

duncanp-sonar (approved):

Issue #1888, opened 2020-11-09
Modify InfoBarManager to check only called on UI thread


duncanp-sonar:
### Description

- [x] change the info bar manager to throw if it is not called on the UI thread
- [x] change the Open In IDE request handler to call "Clear" from the UI thread

#### Notes
Calling "Clear" from a background thread can throw the following exception:

```
[Open in IDE] Processing request. Server: https://peach.sonarsource.com/, project: blazor, organization: {none}, hotspot: AWqAIyITze4-wR2-4IJ8
[Open in IDE] Error handling request: This type of CollectionView does not support changes to its SourceCollection from a thread different from the Dispatcher thread.
[Open in IDE] Detailed error: System.NotSupportedException: This type of CollectionView does not support changes to its SourceCollection from a thread different from the Dispatcher thread.
   at System.Windows.Data.CollectionView.OnCollectionChanged(Object sender, NotifyCollectionChangedEventArgs args)
   at System.Collections.ObjectModel.ReadOnlyObservableCollection`1.OnCollectionChanged(NotifyCollectionChangedEventArgs args)
   at System.Collections.ObjectModel.ReadOnlyObservableCollection`1.HandleCollectionChanged(Object sender, NotifyCollectionChangedEventArgs e)
   at System.Collections.ObjectModel.ObservableCollection`1.OnCollectionChanged(NotifyCollectionChangedEventArgs e)
   at Microsoft.Internal.VisualStudio.PlatformUI.SuspendableObservableCollection`1.OnCollectionChanged(NotifyCollectionChangedEventArgs e)
   at System.Collections.ObjectModel.ObservableCollection`1.RemoveItem(Int32 index)
   at System.Collections.ObjectModel.Collection`1.RemoveAt(Int32 index)
   at Microsoft.Internal.VisualStudio.PlatformUI.CollectionAdapter`2.RemoveSourceItem(Int32 index)
   at Microsoft.Internal.VisualStudio.PlatformUI.CollectionAdapter`2.OnInnerCollectionChanged(IEnumerable innerCollection, NotifyCollectionChangedEventArgs e)
   at Microsoft.Internal.VisualStudio.PlatformUI.CollectionAdapter`2.System.Windows.IWeakEventListener.ReceiveWeakEvent(Type managerType, Object sender, EventArgs e)
   at System.Windows.WeakEventManager.ListenerList.DeliverEvent(Listener& listener, Object sender, EventArgs args, Type managerType)
[...]

PR #1909, opened 2020-11-11
Added UI thread checks to InfoBarManager

duncanp-sonar:
Relates to #1888

### DIFF
--- a/src/Integration.Vsix/InfoBar/InfoBarManager.cs
+++ b/src/Integration.Vsix/InfoBar/InfoBarManager.cs
@@ -214,6 +214,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.InfoBar
                     return;
                 }
 
+                ThreadHelper.ThrowIfNotOnUIThread();
                 this.InfoBarUIElement.Close();
 
                 Debug.Assert(!this.cookie.HasValue, "Expected to be unadvised");
--- a/src/Integration.Vsix/InfoBar/InfoBarManager.cs
+++ b/src/Integration.Vsix/InfoBar/InfoBarManager.cs
@@ -70,11 +70,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.InfoBar
             return AttachInfoBarImpl(toolWindowGuid, message, null, imageMoniker);
         }
 
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Reliability",
-            "S3215:\"interface\" instances should not be cast to concrete types",
-            Justification = "We want to hide the concrete implementation but also handle instance that were created by this class in the first place",
-            Scope = "member",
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -70,26 +70,21 @@ public IInfoBar AttachInfoBar(Guid toolWindowGuid, string message, SonarLintImag
             return AttachInfoBarImpl(toolWindowGuid, message, null, imageMoniker);
         }
 
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Reliability",
```
duncanp-sonar, line 73: Miscellaneous: code smell said this wasn't needed

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -70,26 +70,21 @@ public IInfoBar AttachInfoBar(Guid toolWindowGuid, string message, SonarLintImag
...
         public void DetachInfoBar(IInfoBar currentInfoBar)
         {
             if (currentInfoBar == null)
             {
                 throw new ArgumentNullException(nameof(currentInfoBar));
             }
 
-            PrivateInfoBarWrapper wrapper = currentInfoBar as PrivateInfoBarWrapper;
-            if (wrapper == null)
+            if (!(currentInfoBar is PrivateInfoBarWrapper wrapper))
```
duncanp-sonar, line 80: Miscellaneous: code smell

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -70,26 +70,21 @@ public IInfoBar AttachInfoBar(Guid toolWindowGuid, string message, SonarLintImag
...
             {
                 throw new ArgumentException(Strings.InvalidInfoBarInstance, nameof(currentInfoBar));
             }
 
             currentInfoBar.Close();
 
+            ThreadHelper.ThrowIfNotOnUIThread();
```
duncanp-sonar, line 87: There are two possible paths through the code and I've added the thread helper checker in both places.
However, it hasn't stopped the VSX analyzer complaining, though - there are still half-a-dozen other analysis issues in this file, even though they are in lower-level methods that are guarded by a check in an upstream method.

duncanp-sonar:
I've manually tested that the "ruleset out of date" gold bar still works correctly i.e. doesn't throw.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
rita-g-sonarsource: Is the `currentInfoBar.Close();` safe to run on non-UI thread?

rita-g-sonarsource (approved):
Just a question about IVsInfoBarUIElement.Close running on non-UI thread

Issue #1799, opened 2020-10-14
Handle SQ status request


duncanp-sonar:
- [x] implement http status request handler
- [x] implement status API request handler
- [x] add service to fetch status info

PR #1911, opened 2020-11-11
Added http handler for the /status request

duncanp-sonar:
Relates to #1799


src/IssueViz.Security/OpenInIDE/Api/OpenInIDERequestHandler.cs
```diff
@@ -74,7 +74,8 @@ internal class OpenInIDERequestHandler : IOpenInIDERequestHandler
 
         Task<IStatusResponse> IOpenInIDERequestHandler.GetStatusAsync()
         {
-            throw new NotImplementedException();
+            // TODO: fetch the real data
+            return Task.FromResult<IStatusResponse>(new StatusResponse("Visual Studio", "{unknown}"));
```
duncanp-sonar, line 78: This is enough so that the end-to-end status request now works.

src/IssueViz.Security/OpenInIDE/Contract/IStatusResponse.cs
```diff
@@ -20,27 +20,10 @@
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Contract
 {
-    // TODO:
```
duncanp-sonar, line 23: I changed the names to match those used in SLI.
The icon information will be returned separately. I created another ticket for that.

src/IssueViz.Security/OpenInIDE/Http/StatusOwinRequestHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+        internal StatusOwinRequestHandler(IOpenInIDERequestHandler openInIDERequestHandler)
+        {
+            this.openInIDERequestHandler = openInIDERequestHandler;
+        }
+
+        public string ApiPath => "/status";
+
+        public async Task ProcessRequest(IOwinContext context)
+        {
+            var status = await openInIDERequestHandler.GetStatusAsync();
```
duncanp-sonar, line 48: There are no parameters to validate so this class doesn't have much to do. It doesn't even have to log anything; the pipeline will log that the request has been received.

src/IssueViz.Security/OpenInIDE/Api/StatusResponse.cs
```diff
@@ -0,0 +1,41 @@
...
+        public StatusResponse(string ideName, string description)
+        {
+            IdeName = ideName ?? throw new ArgumentNullException(nameof(ideName));
+            Description = description ?? throw new ArgumentNullException(nameof(description));
+        }
+
+        [JsonProperty("ideName")]
+        public string IdeName { get; }
+
+        [JsonProperty("description")]
```
rita-g-sonarsource, line 38: If we want to change the name of the field, we should have a chat with the team sooner than later

rita-g-sonarsource (approved):

Issue #1801, opened 2020-10-14
Clear hotspot list when solution closes


duncanp-lseg:
TODO: decide whether to do this

pavel-mikula-sonarsource:
Moved to [SLVS-2276](https://sonarsource.atlassian.net/browse/SLVS-2276)

[SLVS-2276]: https://sonarsource.atlassian.net/browse/SLVS-2276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1913, opened 2020-11-12
Implement Clear functionality on hotspots list

rita-g-sonarsource:


src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntryFactory.cs
```diff
@@ -29,21 +30,21 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableD
...
     {
-        ITableEntry Create(IAnalysisIssueVisualization issueVisualization);
+        ITableEntry Create(IAnalysisIssueVisualization issueVisualization, Action removeCallback);
     }
 
     [Export(typeof(IHotspotTableEntryFactory))]
     internal class HotspotTableEntryFactory : IHotspotTableEntryFactory
     {
-        private readonly IServiceProvider serviceProvider;
+        private readonly IVsUIShell2 vsUiShell;
```
rita-g-sonarsource, line 39: Minor: fetching the vsUiShell only once

src/IssueViz.Security.UnitTests/HotspotsList/HotspotTableEntryTests.cs
```diff
@@ -187,32 +187,15 @@ public void TryCreateColumnContent_UnsupportedColumn_ReturnsNull(string column)
         }
 
         [TestMethod]
-        public void TryCreateColumnContent_IssueVizIsNavigable_ReturnsNull()
```
rita-g-sonarsource, line 190: This test is not relevant anymore: a framework element is always created now with a context menu, and it either has navigability style or no style. I added the equivalent of this test in the WpfElementFactory tests

rita-g-sonarsource:
@duncanp-sonar it might be useful to look at the commits separately:
1. General plumbing to add context menu to the control.
2. Implementing a general "Remove" of table entries (not used yet).
3. Gluing 1+2: context menu callback now calls Remove.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
```diff
@@ -33,6 +33,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.TableD
     internal interface IHotspotsStore
     {
         void Add(IAnalysisIssueVisualization hotspot);
+        void Remove(IAnalysisIssueVisualization hotspot);
```
rita-g-sonarsource, line 36: The interface method is not used externally, but I think it still makes sense to have it (and soon we will probably add "RemoveAll()" for #1801)

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -248,6 +248,7 @@ Referenced assemblies:
 - 'Microsoft.VisualStudio.ComponentModelHost, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
 - 'Microsoft.VisualStudio.CoreUtility, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
 - 'Microsoft.VisualStudio.ImageCatalog, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
+- 'Microsoft.VisualStudio.Imaging, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```
duncanp-sonar, line 251: I wonder whether we should look at generating a baseline report for the VS2015 build too, now that our assembly referencing story is getting more complicated.
When you were getting the conditional VSShell stuff to work, didn't you hit problems where the VS2015 version built, but failed at runtime because an assembly from a newer version of VS was referenced? If so, a report might help spot that kind of error. Alternatively, we could write unit test that uses reflection to check referenced assembly versions don't exceed the target VS version (which might be simpler and more reliable).

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntryWpfElementFactory.cs
```diff
@@ -18,34 +18,72 @@
...
+            this.vsUiShell = vsUiShell;
+            this.issueVisualization = issueVisualization;
         }
 
         public FrameworkElement Create(string content)
         {
-            var control = new TextBlock
+            var control = new Border
+            {
+                ContextMenu = CreateContextMenu(),
```
duncanp-sonar, line 58: This probably needs some explanation: we can't create a context menu for the row or the table as a whole, so we have to create the same context menu for each column element in the row?

src/IssueViz.Security.UnitTests/HotspotsList/HotspotTableEntryWpfElementFactoryTests.cs
```diff
@@ -27,17 +26,55 @@
...
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.HotspotsList
 {
     [TestClass]
-    public class NonNavigableFrameworkElementFactoryTests
+    public class HotspotTableEntryWpfElementFactoryTests
     {
         [TestMethod]
-        public void Create_CreatesStyledElement()
+        public void Create_CreatesElementWithContextMenu()
```
duncanp-sonar, line 42: Strictly speaking this only tests that navigable elements have context menus. It wouldn't hurt to make it data driven so it checks both states.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotsTableDataSource.cs
```diff
@@ -97,12 +98,29 @@ public void Add(IAnalysisIssueVisualization hotspot)
...
+        public void Remove(IAnalysisIssueVisualization hotspot)
+        {
+            var entries = tableEntries.Where(x => x.Identity == hotspot).ToArray();
+
+            if (entries.Length == 0)
+            {
+                return;
+            }
+
+            tableEntries.RemoveAll(x=> x.Identity == hotspot);
```
duncanp-sonar, line 113: It would be safer to change this condition to `x in entries` rather than repeating the same condition as in line 106. I'd also rename `entries` to `entriesToRemove`.

I guess we're not expecting the array to ever have more than one entry, are we? Assuming that's the case, I'd add an assertion.

src/IssueViz.Security.UnitTests/HotspotsList/HotspotsTableDataSourceTests.cs
```diff
@@ -164,6 +164,111 @@ public void Add_HasSubscribersToIssuesChangedEvent_SubscribersNotified()
...
+            var testSubject = CreateTestSubject();
+            testSubject.Add(issueViz1);
+            testSubject.Add(issueViz2);
+
+            var sink = new Mock<ITableDataSink>();
+            testSubject.Subscribe(sink.Object);
+
+            sink.Reset();
+
+            testSubject.Remove(issueViz1);
```
duncanp-sonar, line 237: It doesn't hurt to vary the test inputs between otherwise identical tests.
In all three of the "remove with matches" tests you're removing the first item in the list each time. Logically, these tests would still all pass if the product code only ever checked the first item in the list i.e. the tests don't demonstrate that the product code iterates through the list of entries.

You could add a separate test for that case, but that can get pretty tedious. In any event, the general point stands that by varying the data you might catch something you weren't expecting.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntryWpfElementFactory.cs
```diff
@@ -64,8 +69,7 @@ private ContextMenu CreateContextMenu()
                 Header = "Remove",
                 Icon = new CrispImage { Moniker = KnownMonikers.DeleteTableRow }
             };
-            // TODO: implement the click
-            //removeEntryMenuItem.Click += (sender, args) => removeCallback();
+            removeEntryMenuItem.Click += (sender, args) => removeCallback();
```
duncanp-sonar, line 72: This is starting to feel a bit ugly and non-WPF-y :-(
I hope we don't have to add any more commands or we might need to have a re-think about how to wire up the commands.

src/IssueViz.Security/HotspotsList/TableDataSource/HotspotTableEntryWpfElementFactory.cs
```diff
@@ -18,34 +18,72 @@
...
+        {
+            var removeEntryMenuItem = new MenuItem
+            {
+                Header = "Remove",
+                Icon = new CrispImage { Moniker = KnownMonikers.DeleteTableRow }
+            };
+            // TODO: implement the click
+            //removeEntryMenuItem.Click += (sender, args) => removeCallback();
+
+            return new ContextMenu
```
duncanp-sonar, line 74: Do context menus need to be themed?

duncanp-sonar (approved):
A few questions/comments

rita-g-sonarsource:
@duncanp-sonar as we discussed offline, I'm closing this PR for now without merging and reassessing the WPF control.

PR #1916, opened 2020-11-12
Add "bring to front" interface; added calls to it

duncanp-sonar:
Relates #1800

TODO: service implementation

### DIFF
--- a/src/Integration/Helpers/IDEWindowService.cs
+++ b/src/Integration/Helpers/IDEWindowService.cs
@@ -1,0 +1,34 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.Integration.Helpers
+{
+    [Export(typeof(IIDEWindowService))]
+    internal class IDEWindowService : IIDEWindowService
+    {
+        public void BringToFront()
+        {
+            // TODO
+        }
+    }
+}
--- a/src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
+++ b/src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
@@ -127,7 +127,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
             await testSubject.ShowHotspotAsync(ValidRequest)
                 .ConfigureAwait(false);
 
-            CheckInvariants();
+            CheckInvariantBehaviours();
             CheckInfoBarShown();
 
             CheckCalled(stateValidatorMock);
@@ -144,7 +144,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
             await testSubject.ShowHotspotAsync(ValidRequest)
                 .ConfigureAwait(false);
 
-            CheckInvariants();
+            CheckInvariantBehaviours();
             CheckInfoBarShown();
 
             CheckCalled(stateValidatorMock, serverMock);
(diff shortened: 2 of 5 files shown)


src/Integration/Helpers/IDEWindowService.cs
```diff
@@ -0,0 +1,34 @@
...
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.Integration.Helpers
+{
+    [Export(typeof(IIDEWindowService))]
+    internal class IDEWindowService : IIDEWindowService
+    {
+        public void BringToFront()
+        {
+            // TODO
```
duncanp-sonar, line 31: There is an existing VS "BringToFront" command, but it doesn't seem to work reliably i.e. the IDE window isn't always brought to the top. Mind you, directly calling the WinApi doesn't seem always work either...

rita-g-sonarsource: Hm, I wonder if it's what I experienced, that I told you about before: that sometimes I click on "Open in explorer" and the window is not brought to front 🤔 I think it could be some windows limitation.

src/Integration/Helpers/IDEWindowService.cs
```diff
@@ -0,0 +1,34 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.Integration.Helpers
+{
+    [Export(typeof(IIDEWindowService))]
+    internal class IDEWindowService : IIDEWindowService
```
duncanp-sonar, line 27: I considered merging this service with `IToolWindowService` which also only has one method; both methods could be logically grouped into a "window services" component. I think I prefer keeping them separate, but I don't have a strong feeling either way.

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -95,9 +94,10 @@ public void TestInitialize()
             ThreadHelper.SetCurrentThreadAsUIThread();
 
             // The tool window should always be called with the same argument
-            toolWindowService = new Mock<IToolWindowService>();
-            toolWindowService.Setup(x => x.Show(HotspotsToolWindow.ToolWindowId));
+            toolWindowServiceMock = new Mock<IToolWindowService>();
```
duncanp-sonar, line 97: Miscellaneous; renamed for consistency

src/Integration/Helpers/IDEWindowService.cs
duncanp-sonar: Also, I added this empty implementation for now otherwise it would break the dogfood experience.

duncanp-sonar: No, it doesn't flash - there's no indication that anything has happened :-(

duncanp-sonar:
Code coverage - 1 line not covered (the empty method with the TODO). I'm not going to add a junk test just to cover it here.

src/Integration/Helpers/IDEWindowService.cs
rita-g-sonarsource: No I agree, they serve a different functionality, I'd keep them separated

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -46,21 +46,19 @@ public class OpenInIDERequestHandlerTests_MEF
...
-            var storeExport = MefTestHelpers.CreateExport<IHotspotsStore>(Mock.Of<IHotspotsStore>());
-            var failureInfoBarExport = MefTestHelpers.CreateExport<IOpenInIDEFailureInfoBar>(Mock.Of<IOpenInIDEFailureInfoBar>());
-            var hotspotSelectionServiceExport = MefTestHelpers.CreateExport<IHotspotsSelectionService>(Mock.Of<IHotspotsSelectionService>());
-            var loggerExport = MefTestHelpers.CreateExport<ILogger>(Mock.Of<ILogger>());
-
-            // Act & Assert
             MefTestHelpers.CheckTypeCanBeImported<OpenInIDERequestHandler, IOpenInIDERequestHandler>(null,
-                new[] { toolWindowServiceExport, validatorExport, sonarQubeServiceExport, converterExport,
-                        navigatorExport, storeExport, failureInfoBarExport, hotspotSelectionServiceExport, loggerExport });
+                new[] {
```
rita-g-sonarsource, line 50: Btw I like this convention more, it's easier to write

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -314,6 +321,14 @@ private static IAnalysisIssueVisualization CreateHotspotVisualization(string fil
         private void SetSelectionServiceExpectedItem(IAnalysisIssueVisualization expected) =>
             selectionServiceMock.Setup(x => x.Select(expected));
 
+        private void CheckInvariants()
```
rita-g-sonarsource, line 324: Maybe VerifyPersistentBehavior ? CheckConstantBehavior? 
I think Invariants give me a "variables" feeling, as in, you're asserting numbers or strings (probably due to `StringComparer.InvariantCulture`...)

rita-g-sonarsource (approved):

rita-g-sonarsource (approved):

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
duncanp-sonar: To me, "Persistent" suggests serialization and "Constant" suggests data. I've changed it to "CheckInvariantBehaviours".

PR #1917, opened 2020-11-13
Implement bring IDE to front

duncanp-sonar:


duncanp-sonar:
WIP  - working version, but code needs to re-arranged, needs tests etc

src/Integration.Vsix.UnitTests/Helpers/IDEWindowServiceTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        public void BringToFront_GetWindowPlacementFails_IsStillBroughtToFront()
+        {
+            var testSubject = new IDEWindowService(nativeMock.Object, ValidHandle, logger);
+
+            SetGetPlacementResponse(false, new WINDOWPLACEMENT());
+
+            // Act
+            testSubject.BringToFront();
+
+            CheckGetPlacementIsCalled();
```
duncanp-sonar, line 74: We can't reliably set the WINAPI last error code so I'm not testing it.

src/Integration.Vsix/Helpers/IDEWindowService.cs
```diff
@@ -0,0 +1,104 @@
...
+
+using System;
+using System.ComponentModel.Composition;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Integration.Vsix.Native;
+using SonarLint.VisualStudio.Integration.Vsix.Resources;
+
+namespace SonarLint.VisualStudio.Integration.Vsix
```
duncanp-sonar, line 29: I moved the implementation to the `Integration.VSIX` project since that is where the native methods interface is currently.

src/Integration.Vsix/Helpers/IDEWindowService.cs
```diff
@@ -0,0 +1,104 @@
...
+    [Export(typeof(IIDEWindowService))]
+    internal class IDEWindowService : IIDEWindowService
+    {
+        private readonly INativeMethods nativeMethods;
+        private readonly IntPtr handle;
+        private readonly ILogger logger;
+
+        [ImportingConstructor]
+        public IDEWindowService(ILogger logger)
+            : this(new NativeMethods(), Process.GetCurrentProcess().MainWindowHandle, logger)
```
duncanp-sonar, line 40: Fetching the window handle here so that we can inject it into the other internal constructor for testing.

src/Integration.Vsix/Helpers/IDEWindowService.cs
```diff
@@ -0,0 +1,104 @@
...
+            this.handle = handle;
+            this.logger = logger;
+        }
+
+        public void BringToFront()
+        {
+            try
+            {
+                logger.WriteLine(Strings.IDEWindowService_BringingToFront);
+                if (handle == IntPtr.Zero)
```
duncanp-sonar, line 56: As it happens, the current process doesn't have a windows handle when running the tests, so we have to have a way of injecting the handle in the tests.

src/Integration.Vsix/Helpers/IDEWindowService.cs
```diff
@@ -0,0 +1,104 @@
...
+            var placement = new WINDOWPLACEMENT();
+            placement.length = (uint)Marshal.SizeOf(placement);
+            var success = nativeMethods.GetWindowPlacement(handle, ref placement);
+            if (!success)
+            {
+                LogWinApiFailure(nameof(nativeMethods.GetWindowPlacement));
+                return;
+            }
+
+            if (placement.showCmd == NativeMethods.SW_SHOWMINIMIZED)
```
rita-g-sonarsource, line 82: Out of curiosity: why not always call `nativeMethods.ShowWindow(handle, NativeMethods.SW_RESTORE)` ? Worst case it would just do nothing, no?

src/Integration.Vsix.UnitTests/Helpers/IDEWindowServiceTests.cs
```diff
@@ -0,0 +1,148 @@
...
+            CheckSetForegroundIsCalled();
+        }
+
+        [TestMethod]
+        public void BringToFront_NotMinimized_IsBroughtToFrontButNotRestored()
+        {
+            var testSubject = new IDEWindowService(nativeMock.Object, ValidHandle, logger);
+
+            SetGetPlacementResponse(result: true,
+                new WINDOWPLACEMENT { showCmd = 0 });
```
rita-g-sonarsource, line 84: Can this be a readonly? 
`WINDOWPLACEMENT NotMinimizedPlacement` 
It would make the tests slightly easier to read (same for the MinimizedPlacement below)

rita-g-sonarsource (approved):
Just some cosmetics

src/Integration.Vsix/Native/INativeMethods.cs
```diff
@@ -37,20 +43,62 @@ internal class NativeMethods : INativeMethods
...
             public static extern IntPtr GetAncestor(IntPtr hwnd, uint flags);
-            
+
             [DllImport("user32.dll", CharSet = CharSet.Auto)]
             public static extern int SendMessage(IntPtr hWnd, UInt32 Msg, IntPtr wParam, IntPtr lParam);
+
+            [DllImport("user32.dll")]
+            public static extern bool SetForegroundWindow(IntPtr hWnd);
+            [DllImport("user32.dll", SetLastError = true)]
+            public static extern bool ShowWindow(IntPtr hWnd, int nCmdShow);
```
rita-g-sonarsource, line 73: If you're making any additional commits then can you please add an empty line here? 😄 (like with the others)

src/Integration.Vsix/Helpers/IDEWindowService.cs
duncanp-sonar: No; it would reset a maximized window to non-maximized. And I think it would change the size of a normal window that the user had resized back to its initial size.

PR #1919, opened 2020-11-13
Return real data in SQ status request

### LINKED ISSUE #1799: Handle SQ status request

rita-g-sonarsource:
Fixes #1799

### DIFF
--- a/src/IssueViz.Security/OpenInIDE/Api/StatusRequestHandler.cs
+++ b/src/IssueViz.Security/OpenInIDE/Api/StatusRequestHandler.cs
@@ -18,12 +18,19 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.ComponentModel.Composition;
 using System.Threading.Tasks;
+using SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Contract;
 
-namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Contract
+namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Api
 {
-    public interface IOpenInIDERequestHandler
+    [Export(typeof(IStatusRequestHandler))]
+    internal class StatusRequestHandler : IStatusRequestHandler
     {
-        Task ShowHotspotAsync(IShowHotspotRequest request);
+        Task<IStatusResponse> IStatusRequestHandler.GetStatusAsync()
+        {
+            // TODO: fetch the real data
+            return Task.FromResult<IStatusResponse>(new StatusResponse("Visual Studio", "{unknown}"));
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


src/IssueViz.Security/OpenInIDE/Api/StatusRequestHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+        public StatusRequestHandler([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider)
+        {
+            vsShell = serviceProvider.GetService(typeof(SVsShell)) as IVsShell;
+            vsSolution = serviceProvider.GetService(typeof(SVsSolution)) as IVsSolution;
+        }
+
+        Task<IStatusResponse> IStatusRequestHandler.GetStatusAsync()
+        {
+            vsShell.GetProperty((int) __VSSPROPID5.VSSPROPID_AppBrandName, out var ideName);
+            ideName = ideName ?? "Microsoft Visual Studio";
```
rita-g-sonarsource, line 47: I don't really care about the hr result of the `vsShell.GetProperty`. If it fails, we will return the default name, so no harm done.

src/IssueViz.Security/OpenInIDE/Api/StatusRequestHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            vsSolution.GetProperty((int) __VSPROPID.VSPROPID_SolutionBaseName, out var ideInstance);
```
rita-g-sonarsource, line 49: Same comment as above regarding the hr result failure.
Also, the name will be null if there is no solution open, so in both cases I'm returning an empty string.

In a success case, this will return the solution name, e.g. `SonarLint.VisualStudio.Integration` (without .sln).

src/IssueViz.Security/OpenInIDE/Contract/IStatusRequestHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Threading.Tasks;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Contract
+{
+    internal interface IStatusRequestHandler
```
rita-g-sonarsource, line 25: I wasn't sure how to call it. IStatusService seemed misleading.

src/IssueViz.Security/OpenInIDE/Api/StatusRequestHandler.cs
duncanp-sonar: It might be better to return `(no solution)` / `(no open solution)`. I guess we can wait for dogfood feedback.

duncanp-sonar (approved):

PR #1921, opened 2020-11-17
Add CORS-related headers to /status and /hotspot/show responses

### LINKED ISSUE #1918: Bug: fix /status request message headers
### Description

The headers for the status request are wrong:

* `Content-Type` should be `application/json`
* `Access-Control-Allow-Origin` header should be set to the request's Origin

jblievremont:
Fixes #1918

### DIFF
--- a/src/IssueViz.Security.UnitTests/OpenInIDE/Http/ShowHotspotOwinRequestHandlerTests.cs
+++ b/src/IssueViz.Security.UnitTests/OpenInIDE/Http/ShowHotspotOwinRequestHandlerTests.cs
@@ -34,8 +34,6 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
     [TestClass]
     public class ShowHotspotOwinRequestHandlerTests
     {
-        const string ORIGIN = "http://origin";
-
         [TestMethod]
         public void MefCtor_CheckIsExported()
         {
@@ -126,7 +124,6 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
             await testSubject.ProcessRequest(context);
 
             context.Response.StatusCode.Should().Be(200);
-            context.Response.Headers["Access-Control-Allow-Origin"].Should().Be(ORIGIN);
             CheckApiHandlerCalledWithExpectedValues(apiHandlerMock, expectedServer, expectedProject, expectedHotspot, expectedOrganization);
         }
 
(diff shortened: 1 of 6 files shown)


src/IssueViz.Security/OpenInIDE/Http/StatusOwinRequestHandler.cs
```diff
@@ -45,10 +45,11 @@ internal StatusOwinRequestHandler(IStatusRequestHandler statusRequestHandler)
 
         public async Task ProcessRequest(IOwinContext context)
         {
+            context.Response.Headers["Access-Control-Allow-Origin"] = context.Request.Headers["Origin"];
```
rita-g-sonarsource, line 48: @jblievremont is there a reason to do it before processing the request? why is it not next to ContentType assignment?

src/IssueViz.Security.UnitTests/OpenInIDE/Http/ShowHotspotOwinRequestHandlerTests.cs
```diff
@@ -34,6 +34,8 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
     [TestClass]
     public class ShowHotspotOwinRequestHandlerTests
     {
+        const string ORIGIN = "http://origin";
```
rita-g-sonarsource, line 37: Just FYI, c# naming convention for constants is PascalCase :) we have a rule for that in the QP but it's test code so unfortunately it doesn't get picked up... I'll fix it for you 🙂

src/IssueViz.Security/OpenInIDE/Http/ShowHotspotOwinRequestHandler.cs
```diff
@@ -53,6 +53,7 @@ internal ShowHotspotOwinRequestHandler(IOpenInIDERequestHandler openInIDERequest
 
         public async Task ProcessRequest(IOwinContext context)
         {
+            context.Response.Headers["Access-Control-Allow-Origin"] = context.Request.Headers["Origin"];
```
rita-g-sonarsource, line 56: It might be better to do it as part of OwinPipelineProcessor.ProcessRequest, instead of in each handler.

rita-g-sonarsource (commented):
@jblievremont see my comments please 🙂

src/IssueViz.Security/OpenInIDE/Http/StatusOwinRequestHandler.cs
jblievremont: Yup, I moved it to `OwinPipelineProcessor` - thanks for the pointer! The idea is that we want this to be done as early as possible in the request handling, for the browser to allow the frontend code to see the response even in case of a failure.

rita-g-sonarsource (approved):

Issue #1922, opened 2020-11-17
Replace WpfTableControl with regular WPF implementation


rita-g-sonarsource:
#### Description

Following #1920 - Implement a replacement control.

PR #1924, opened 2020-11-19
Add new HotspotsList control

rita-g-sonarsource:
Part of #1922


src/IssueViz.Security/HotspotsList2/ViewModels/HotspotViewModel.cs
```diff
@@ -18,30 +18,78 @@
...
+                NotifyPropertyChanged(nameof(Line));
+                NotifyPropertyChanged(nameof(Column));
+            }
+        }
 
-        // todo: take from span
-        public int Column => Hotspot.Issue.StartLineOffset;
+        public event PropertyChangedEventHandler PropertyChanged;
+
+        [NotifyPropertyChangedInvocator]
```
rita-g-sonarsource: This attribute is a resharper annotation that was added automatically and I kept by mistake. It was removed in the last commit.

rita-g-sonarsource: This attribute is a resharper annotation that was added automatically and I kept by mistake. It was removed in a later commit.

rita-g-sonarsource:
@duncanp-sonar a few notes:

1. This PR adds a new control that is not yet used. It has all the existing functionality but no style (that would come in PR #1928, I've tested the style changes separately).
2. I've made no changes to the old control except removing `IHotspotsStore` implementation from it so that `OpenInIDERequestHandler` would import the new code. In PR #1925 all the old code will be deleted and the new control will be used instead.
3. I've added each functionality bit in a separate commit so it might help to look at them separately.

src/IssueViz.Security/HotspotsList2/ViewModels/HotspotsControlViewModel.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        public ObservableCollection<IHotspotViewModel> Hotspots { get; } = new ObservableCollection<IHotspotViewModel>();
+
+        public IHotspotViewModel SelectedHotspot { get; set; }
+
+        public ICommand NavigateCommand { get; }
+
+        public HotspotsControlViewModel(IHotspotsStore hotspotsStore, ICommand navigateCommand, IHotspotsSelectionService selectionService)
+        {
+            BindingOperations.EnableCollectionSynchronization(Hotspots, Lock);
```
rita-g-sonarsource, line 59: To allow the hotspotsStore to be changed on a different thread.

src/IssueViz.Security/IssueViz.Security.csproj
```diff
@@ -58,6 +58,11 @@
 
   <ItemGroup>
     <None Remove="HotspotsList\HotspotsControl.xaml" />
+    <None Remove="HotspotsList2\HotspotsControl2.xaml" />
```
rita-g-sonarsource, line 61: Will be renamed in #1925

src/IssueViz.Security/Store/HotspotsStore.cs
```diff
@@ -0,0 +1,78 @@
...
+        public void Add(IAnalysisIssueVisualization hotspot)
+        {
+            Hotspots.Add(hotspot);
+
+            var hotspotFilePaths = hotspot
+                .GetAllLocations()
+                .Select(x => x.CurrentFilePath)
+                .Distinct(StringComparer.OrdinalIgnoreCase);
+
+            IssuesChanged?.Invoke(this, new IssuesChangedEventArgs(hotspotFilePaths));
```
duncanp-sonar, line 55: Minor; there's no point in calculating the event arguments if there aren't any listeners. But it probably doesn't make any difference since we expect there will always be a listener.

src/IssueViz.Security/Store/HotspotsStore.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        public ReadOnlyObservableCollection<IAnalysisIssueVisualization> GetAll()
+        {
+            return new ReadOnlyObservableCollection<IAnalysisIssueVisualization>(Hotspots);
+        }
+
+        public event EventHandler<IssuesChangedEventArgs> IssuesChanged;
```
duncanp-sonar, line 63: It's confusing reading this code and seeing an observable collection and specific event; it's not clear why both are needed. Perhaps implementing the interfaces explicitly would make it clearer.

src/IssueViz.Security/HotspotsList2/ViewModels/HotspotViewModel.cs
```diff
@@ -0,0 +1,104 @@
...
+            Hotspot = hotspot;
+            Hotspot.PropertyChanged += Hotspot_PropertyChanged;
+        }
+
+        public IAnalysisIssueVisualization Hotspot { get; }
+
+        public int Line =>
+            CanUseSpan()
+                ? Hotspot.Span.Value.Start.GetContainingLine().LineNumber
+                : Hotspot.Issue.StartLine;
```
duncanp-sonar, line 54: Are the `Span` and `Issue` positions values both 0-based (or both 1-based?)

src/IssueViz.Security/HotspotsList2/ViewModels/HotspotViewModel.cs
```diff
@@ -0,0 +1,104 @@
...
+                    return Hotspot.Issue.StartLineOffset;
+                }
+
+                var position = Hotspot.Span.Value.Start;
+                var line = position.GetContainingLine();
+                return position.Position - line.Start.Position;
+            }
+        }
+
+        public string FileName =>
```
duncanp-sonar, line 71: Is it correct that this path could contain either an absolute path (if we've been able to resolve the location) or a relative path (if we couldn't)?
Presumably we're not expecting the caller to need to know this because the value is only used for display purposes. If so, then it should be clear that the caller shouldn't be using the value for anything else e.g. calling the property something like `DisplayPath`, or a doc comment.

src/IssueViz.Security/HotspotsList2/ViewModels/HotspotsControlViewModel.cs
duncanp-sonar: This is unusual enough to be worth calling out in comment.
The docs also say that it should be called on the UI thread, so it would be worth checking that with the ThreadHelper too.

duncanp-sonar (approved):

src/IssueViz.Security/HotspotsList2/ViewModels/HotspotViewModel.cs
rita-g-sonarsource: Hmm, good question, I didn't think about it.
For line there is a difference: SQ's is one-based and VS is zero.
But for column, I get the same values, and it seems that SQ is zero based. I'll check that with the team.

PR #1929, opened 2020-11-19
Add ContextMenu with Remove command

### LINKED ISSUE #1792: Implement Clear functionality on hotspots list
#### Description

Provide a way for the user to remove a hotspot from the list, e.g. right click menu.

rita-g-sonarsource:
Fixes #1792


src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -17,6 +17,16 @@
             <wpf:BoolToVisibilityConverter TrueValue="Collapsed" FalseValue="Visible" />
         </wpf:AggregatorValueConverter>
 
+        <ContextMenu x:Key="RowMenu" DataContext="{Binding PlacementTarget.DataContext, RelativeSource={RelativeSource Self}}">
+            <MenuItem Header="Delete"
```
duncanp-sonar, line 21: `Delete` sounds a bit final. Could it mean the issue is deleted on the server? 
`Remove` sounds less final and less likely to be misinterpreted.

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -144,6 +155,58 @@
             <Setter Property="TextWrapping" Value="Wrap"/>
             <Setter Property="TextAlignment" Value="Left"/>
         </Style>
+        <Style TargetType="ContextMenu">
```
duncanp-sonar, line 158: Is this based on an existing/reference implementation?

src/IssueViz.Security/HotspotsList/ViewModels/HotspotsControlViewModel.cs
```diff
@@ -65,22 +73,49 @@ public HotspotsControlViewModel(IHotspotsStore hotspotsStore, ICommand navigateC
...
-        private void UpdateHotspotsList(IEnumerable<IAnalysisIssueVisualization> hotspots)
+        private void UpdateHotspotsList(IEnumerable<IAnalysisIssueVisualization> addedHotspots, IEnumerable<IAnalysisIssueVisualization> deletedHotspots)
         {
-            foreach (var addedHotspot in hotspots)
+            foreach (var addedHotspot in addedHotspots)
             {
                 Hotspots.Add(new HotspotViewModel(addedHotspot));
             }
+
+            var viewModelsToDelete = Hotspots.Where(x => deletedHotspots.Contains(x.Hotspot)).ToList();
```
duncanp-sonar, line 103: I'm guessing the list needs to be reified here because the hotspots collection is being mutated below.
This might be worth calling out in a comment since it's not entirely obvious - I sometimes add a comment like that, but not always.

I was going to say that since the collection is only used within this method and then discarded, it makes sense to use the cheapest data structure that will do the job i.e. an array instead of a list. However, we don't know if the implementation of ToArray is more or less efficient that that of ToList, so there's no point.

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
rita-g-sonarsource: No, I tweaked it until I thought it looked similar enough.

PR #1942, opened 2020-11-25
Add support for Roslyn in SonarLanguageRecognizer

### LINKED ISSUE #1923: Change ITaggableBufferIndicator and add support for C# files
#### Description

The tagging infrastructure does not work for C# files due to `ITaggableBufferIndicator` not recognizing them. Therefore, `IAnalysisIssueVisualization.Span` will not be updated as the buffer changes.

rita-g-sonarsource:
Fixes #1923

### DIFF
--- a/src/Core/Analysis/AnalysisLanguage.cs
+++ b/src/Core/Analysis/AnalysisLanguage.cs
@@ -26,6 +26,7 @@ namespace SonarLint.VisualStudio.Core.Analysis
     public enum AnalysisLanguage
     {
         Javascript,
-        CFamily
+        CFamily,
+        Roslyn
     }
 }
--- a/src/Core/Analysis/AnalysisLanguage.cs
+++ b/src/Core/Analysis/AnalysisLanguage.cs
@@ -27,6 +27,6 @@ namespace SonarLint.VisualStudio.Core.Analysis
     {
         Javascript,
         CFamily,
-        Roslyn
+        RoslynFamily
     }
 }
--- a/src/IssueViz.UnitTests/Editor/LanguageDetection/SonarLanguageRecognizerTests.cs
+++ b/src/IssueViz.UnitTests/Editor/LanguageDetection/SonarLanguageRecognizerTests.cs
@@ -114,7 +114,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.LanguageDet
         [DataRow("JavaScript", AnalysisLanguage.Javascript)]
         [DataRow("Vue", AnalysisLanguage.Javascript)]
         [DataRow("C/C++", AnalysisLanguage.CFamily)]
-        [DataRow("Roslyn Languages", AnalysisLanguage.Roslyn)]
+        [DataRow("Roslyn Languages", AnalysisLanguage.RoslynFamily)]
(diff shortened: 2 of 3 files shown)


src/IssueViz/Editor/LanguageDetection/SonarLanguageRecognizer.cs
```diff
@@ -93,5 +100,8 @@ private IEnumerable<IContentType> GetExtensionContentTypes(string fileExtension,
 
         private static bool IsCFamilyDocument(IEnumerable<IContentType> contentTypes) =>
             contentTypes.Any(type => type.IsOfType("C/C++"));
+
+        private static bool IsRoslynDocument(IEnumerable<IContentType> contentTypes) =>
+            contentTypes.Any(type => type.IsOfType("Roslyn Languages"));
```
rita-g-sonarsource, line 105: Based on https://github.com/dotnet/roslyn/blob/14b7626db71faee5df16a4db4b339046bf54c95f/src/EditorFeatures/Text/ContentTypeNames.cs#L11

Works for .cs files and .vb files - can you think of anything else I should check? I'm not sure what other files do we analyze.

src/IssueViz.UnitTests/Editor/LanguageDetection/SonarLanguageRecognizerTests.cs
```diff
@@ -97,78 +97,62 @@ public void Detect_WhenFileExtensionIsJavascriptRelated_ReturnsJavascript()
         }
 
         [TestMethod]
-        public void Detect_WhenExtensionNotRegisteredAndBufferContentTypeIsJsRelated_ReturnsJavascript()
```
rita-g-sonarsource, line 100: Changed to DataRow

src/IssueViz/Editor/LanguageDetection/SonarLanguageRecognizer.cs
duncanp-sonar: C, C++, C#, VB.NET and JavaScript currently. So I don't think there's anything else we need to cover at the moment.

src/Core/Analysis/AnalysisLanguage.cs
```diff
@@ -26,6 +26,7 @@ namespace SonarLint.VisualStudio.Core.Analysis
     public enum AnalysisLanguage
     {
         Javascript,
-        CFamily
+        CFamily,
+        Roslyn
```
duncanp-sonar, line 30: Seems odd calling `Roslyn` a language. Still, we can always change it later to identify the individual languages if necessary.

rita-g-sonarsource: Yea I wasn't sure, maybe "RoslynFamily"? (like CFamily)

duncanp-sonar (approved):

src/Core/Analysis/AnalysisLanguage.cs
duncanp-sonar: That might be slightly better. 

Although returning CFamily also seems odd since we do care whether the language is C or C++, so it seems odd that the language recognizer isn't the thing that calculates it. Still, that's a separate issue.

rita-g-sonarsource: 😆 true.

Ok, I renamed to `RoslynFamily` for now.

PR #1947, opened 2020-11-25
Enable debug logging using environment variable

duncanp-sonar:


src/Core/ILoggerExtensions.cs
```diff
@@ -18,20 +18,32 @@
...
+        {
+            shouldLogDebug = settings.LogDebugMessages();
+        }
+
         /// <summary>
-        /// Logs messages only when the code is compiled with DEBUG condition. This is temporary
+        /// Logs messages only when an environment variable is set. This is temporary
         /// solution for not having log verbosity setting.
         /// </summary>
-        [Conditional("DEBUG")]
```
duncanp-sonar, line 31: Previously the lack of tests wasn't a problem because the code didn't exist in the shipping version. However, since it will now ship I've added tests.

src/Core.UnitTests/EnvironmentSettingsTests.cs
```diff
@@ -103,5 +103,23 @@ public void SonarLintDaemonDownloadUrl_ReturnsExpectedValue(string envVarValue,
...
+        [DataRow("0", false)]
+        [DataRow("not a boolean", false)]
+        public void LogDebugMessages_CorrectlyMapped(string envVarValue, bool expected)
+        {
+            using (var scope = new EnvironmentVariableScope())
+            {
+                scope.SetVariable(EnvironmentSettings.LogDebugMessagesEnvVar, envVarValue);
+                new EnvironmentSettings().LogDebugMessages().Should().Be(expected);
+            }
+        }
```
duncanp-sonar, line 123: This test is copied from the `TreatBlockerSeverityAsError_CorrectlyMapped` test above.

src/Core/EnvironmentSettings.cs
```diff
@@ -50,6 +52,15 @@ public int PCHGenerationTimeoutInMs(int defaultValue)
...
 
+        public bool LogDebugMessages()
+        {
+            if (bool.TryParse(Environment.GetEnvironmentVariable(LogDebugMessagesEnvVar), out var result))
+            {
+                return result;
+            }
+            return false;
+        }
+
```
duncanp-sonar, line 63: Same as the `TreatBlockerSeverityAsError` method above.

src/Core/ILoggerExtensions.cs
```diff
@@ -18,20 +18,32 @@
...
     {
+        private static bool shouldLogDebug;
+
+        static ILoggerExtensions() => Initialize(new EnvironmentSettings());
+
+        internal /* for testing*/ static void Initialize(IEnvironmentSettings settings)
+        {
+            shouldLogDebug = settings.LogDebugMessages();
+        }
+
```
duncanp-sonar, line 35: This is a bit more complicated now to make it testable. Perhaps I should just have written the property as
```
internal /* for testing */ static bool shouldLogDebug = new EnvironmentSettings().LogDebugMessages();
```

src/Integration.Vsix/Helpers/IDEWindowService.cs
```diff
@@ -99,6 +100,6 @@ private void BringToFront(IntPtr handle)
         }
 
         private void LogWinApiFailure(string apiName) =>
-            logger.WriteLine(Strings.IDEWindowService_WinAPICallFailed, apiName, Marshal.GetLastWin32Error());
+            logger.LogDebug(Strings.IDEWindowService_WinAPICallFailed, apiName, Marshal.GetLastWin32Error());
```
duncanp-sonar, line 103: I changed the WinApi failure logging to be debug-only.

src/Integration/MefServices/MefSonarQubeService.cs
```diff
@@ -56,8 +56,8 @@ public LoggerAdapter(ILogger logger)
             }
 
             public void Debug(string message) =>
-                // This will be executed only on DEBUG build.
-                logger.LogDebug($"DEBUG: {message}");
```
duncanp-sonar, line 60: The `LogDebug` method now prefixes all strings with `DEBUG` so it isn't needed here.

src/Core/IEnvironmentSettings.cs
```diff
@@ -46,5 +45,10 @@ public interface IEnvironmentSettings
         /// Download location for the SonarLint daemon additional zip file
         /// </summary>
         string SonarLintDaemonDownloadUrl();
+
+        /// <summary>
+        /// Returns true if debug-level messages should be logged to the output window, otherwise false
+        /// </summary>
+        bool LogDebugMessages();
```
rita-g-sonarsource, line 52: Maybe something like `ShouldLogDebugMessages`, to clarify that it's a boolean.

src/Core/EnvironmentSettings.cs
rita-g-sonarsource: Minor: we have a `ParseInt` method, it might be worth adding `ParseBoolean` just for consistency and to make it a bit shorter. It's already pretty clear though so it's not that important.

rita-g-sonarsource (approved):

PR #1950, opened 2020-11-27
Add Hotspot to the list only once

### LINKED ISSUE #1944: [Hotspots] Hotspot should only be added to the list once
### Description

If a hotspot is already in the hotspot list and the user executes an Open In IDE request for the same hotspot again then the existing list entry should be selected.

Currently, duplicate entries are added to the list:

image

rita-g-sonarsource:
Fixes #1944

### DIFF
--- a/src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
+++ b/src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
@@ -19,7 +19,6 @@
  */
 
 using System;
-using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
@@ -225,8 +224,10 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
             SetServerResponse(ValidRequest, ValidServerHotspot);
             SetConversionResponse(ValidServerHotspot, hotspotVizMock.Object);
             SetNavigationRespone(hotspotVizMock.Object, true);
-            SetStoreExpectedItem(hotspotVizMock.Object);
-            SetSelectionServiceExpectedItem(hotspotVizMock.Object);
+
+            var addedHotspotVizMock = new Mock<IAnalysisIssueVisualization>();
+            SetStoreExpectedItem(hotspotVizMock.Object, addedHotspotVizMock.Object);
+            SetSelectionServiceExpectedItem(addedHotspotVizMock.Object);
 
             // Act
             await testSubject.ShowHotspotAsync(ValidRequest)
@@ -252,7 +253,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
             SetServerResponse(ValidRequest, ValidServerHotspot);
             SetConversionResponse(ValidServerHotspot, hotspotViz);
             SetNavigationRespone(hotspotViz, false);
-            SetStoreExpectedItem(hotspotViz);
+            SetStoreExpectedItem(hotspotViz, hotspotViz);
             SetSelectionServiceExpectedItem(hotspotViz);
 
             // Act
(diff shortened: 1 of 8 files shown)


src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDERequestHandlerTests.cs
```diff
@@ -266,29 +267,6 @@ await testSubject.ShowHotspotAsync(ValidRequest)
             logger.AssertPartialOutputStringExists(hotspotFilePath, hotspotStartLine.ToString());
         }
 
-        [TestMethod]
-        public async Task ShowHotspot_DataIsValid_IssueFirstAddedToStoreAndThenSelected()
```
rita-g-sonarsource, line 270: Test is no longer relevant since it is now possible to instead test the dependency between mocks, as done in `ShowHotspot_DataIsValid_NavigationSucceeded_And_IssueAddedToStoreAndSelected`

src/IssueViz.Security/Store/HotspotsStore.cs
```diff
@@ -45,10 +50,19 @@ internal class HotspotsStore : IHotspotsStore, IIssueLocationStore
     {
         private ObservableCollection<IAnalysisIssueVisualization> Hotspots { get; } = new ObservableCollection<IAnalysisIssueVisualization>();
 
-        void IHotspotsStore.Add(IAnalysisIssueVisualization hotspot)
+        IAnalysisIssueVisualization IHotspotsStore.GetOrAdd(IAnalysisIssueVisualization hotspot)
```
duncanp-sonar, line 53: The parameter name `hotspot` is a bit misleading since it's a visualization that we expect to contain a hotspot.
I think the visualization params normally have "Viz" in the name.

src/IssueViz.Security/Store/HotspotsStore.cs
```diff
@@ -72,6 +86,15 @@ private void NotifyHotspotChanged(IAnalysisIssueVisualization changedHotspot)
             IssuesChanged.Invoke(this, new IssuesChangedEventArgs(hotspotFilePaths));
         }
 
+        private IAnalysisIssueVisualization Exists(IAnalysisIssueVisualization hotspot)
+        {
+            var matches = Hotspots
+                .ToLookup(x => ((IHotspot)x.Issue).HotspotKey)
```
duncanp-sonar, line 92: Is the lookup necessary - wouldn't a simple `FirstOrDefault` on the collection do? e.g.

```
            var key = ((IHotspot)hotspot.Issue).HotspotKey;
            return Hotspots.FirstOrDefault(x => ((IHotspot)x.Issue).HotspotKey == key);
```

rita-g-sonarsource: You're right; I was originally implementing IEquatable, then changed the lookup, and completely missed the obvious solution. Thanks!

src/IssueViz.Security/Store/HotspotsStore.cs
```diff
@@ -72,6 +86,15 @@ private void NotifyHotspotChanged(IAnalysisIssueVisualization changedHotspot)
             IssuesChanged.Invoke(this, new IssuesChangedEventArgs(hotspotFilePaths));
         }
 
+        private IAnalysisIssueVisualization Exists(IAnalysisIssueVisualization hotspot)
```
duncanp-sonar, line 89: I'd expect 'Exists' to return a boolean. `FindExisting`?

src/IssueViz.Security/Store/HotspotsStore.cs
```diff
@@ -26,14 +26,19 @@
...
 namespace SonarLint.VisualStudio.IssueVisualization.Security.Store
 {
     internal interface IHotspotsStore
     {
         ReadOnlyObservableCollection<IAnalysisIssueVisualization> GetAll();
 
-        void Add(IAnalysisIssueVisualization hotspot);
+        /// <summary>
+        /// Adds a given hotspot to the list if it does not already exist.
+        /// Returns the given hotspot, or the existing hotspot if the key exists.
```
duncanp-sonar, line 39:
```
...
/// Returns the given visualization, or the existing visualization with the same hotspot key.
```

src/IssueViz.Security.UnitTests/Store/HotspotsStoreTests.cs
```diff
@@ -80,12 +81,77 @@ public void GetAll_ReturnsReadOnlyObservableWrapper()
...
 
+        [TestMethod]
+        public void GetOrAdd_NoExistingHotspots_HotspotAdded()
+        {
+            var testSubject = new HotspotsStore() as IHotspotsStore;
+
+            var hotspotToAdd = CreateIssueViz(hotspotKey:"some hotspot");
+            var addedHotspot = testSubject.GetOrAdd(hotspotToAdd);
+
+            addedHotspot.Should().Be(hotspotToAdd);
```
duncanp-sonar, line 92: We're expecting the same object instance: 'BeSameAs' is a more accurate check.

src/IssueViz.Security.UnitTests/Store/HotspotsStoreTests.cs
```diff
@@ -80,12 +81,77 @@ public void GetAll_ReturnsReadOnlyObservableWrapper()
...
+        {
+            var testSubject = new HotspotsStore() as IHotspotsStore;
+
+            var someOtherHotspot = CreateIssueViz(hotspotKey: "some hotspot 1");
+            testSubject.GetOrAdd(someOtherHotspot);
+
+            var hotspotToAdd = CreateIssueViz(hotspotKey: "some hotspot 2");
+            var addedHotspot = testSubject.GetOrAdd(hotspotToAdd);
+
+            addedHotspot.Should().Be(hotspotToAdd);
```
duncanp-sonar, line 108: 'BeSameAs'

src/IssueViz.Security/Store/HotspotsStore.cs
duncanp-sonar: Could be worse; when I thought about this yesterday, I missed the fact that there was a unique `HotspotKey` and thought we'd have to compare the issue fields...

duncanp-sonar (approved):

src/IssueViz.Security/Store/HotspotsStore.cs
rita-g-sonarsource: That's exactly what I did originally...

duncanp-sonar: 🤦‍♂️

PR #1955, opened 2020-11-30
Separated rule metadata from IHotspot

duncanp-sonar:
Added data fields for rule help

The web API returns additional information such as the "help" fields that we currently don't handle but will need later.
This commit changes the SLVS data structure to more closely match the data structure returned by the API i.e. a separate interface to contain the rule metadata.

### DIFF
--- a/src/IssueViz.Security.UnitTests/Models/HotspotRuleTests.cs
+++ b/src/IssueViz.Security.UnitTests/Models/HotspotRuleTests.cs
@@ -1,0 +1,43 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.IssueVisualization.Security.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Models
+{
+    [TestClass]
+    public class HotspotRuleTests
+    {
+        public void Ctor_PropertiesSet()
+        {
+            IHotspotRule rule = new HotspotRule("key", "name", "sec cat", HotspotPriority.Medium, "risk", "vuln", "fix");
+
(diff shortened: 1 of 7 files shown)


src/IssueViz.Security/Models/IHotspot.cs
```diff
@@ -82,9 +73,9 @@ internal class Hotspot : IHotspot
         public int StartLineOffset { get; }
         public int EndLineOffset { get; }
         public string LineHash { get; }
-        public string RuleKey { get; }
+        public string RuleKey => Rule.RuleKey;
```
duncanp-sonar, line 76: `IAnalysisIssueBase` has a `RuleKey` property so `IHotspot` has to provide an implementation.
However, the new `IHotspotRule` interface has a `RuleKey` property, so the `IHotspot` implementation can just delegate to it.

src/IssueViz.Security/Models/IHotspot.cs
```diff
@@ -28,21 +28,14 @@ internal interface IHotspot : IAnalysisIssueBase
...
         string ServerFilePath { get; }
     }
 
-    public enum HotspotPriority
-    {
-        High,
-        Medium,
-        Low
-    }
-
```
duncanp-sonar, line 45: Moved to the `IHotspotRule` file.

src/IssueViz.Security.UnitTests/Models/HotspotRuleTests.cs
```diff
@@ -0,0 +1,43 @@
...
+    [TestClass]
+    public class HotspotRuleTests
+    {
+        public void Ctor_PropertiesSet()
+        {
+            IHotspotRule rule = new HotspotRule("key", "name", "sec cat", HotspotPriority.Medium, "risk", "vuln", "fix");
+
+            rule.RuleKey.Should().Be("key");
+            rule.RuleName.Should().Be("name");
+            rule.SecurityCategory.Should().Be("sec cat");
```
duncanp-sonar, line 36: I don't think we currently display the security category, but I can't think of a reason not to.

src/IssueViz.Security.UnitTests/Models/HotspotRuleTests.cs
```diff
@@ -0,0 +1,43 @@
...
+            rule.Priority.Should().Be(HotspotPriority.Medium);
+            rule.RiskDescription.Should().Be("risk");
+            rule.VulnerabilityDescription.Should().Be("vuln");
+            rule.FixRecommendations.Should().Be("fix");
```
duncanp-sonar, line 40: These are the three "help" fields.

src/IssueViz.Security/Models/IHotspot.cs
rita-g-sonarsource: @duncanp-sonar note that this field is also used in the xaml and I don't see a fix for it.
Does the colour of the priority column still work?

src/IssueViz.Security.UnitTests/Models/HotspotRuleTests.cs
rita-g-sonarsource: Just that it's not user friendly. 
Damien has raised the point that they do a copy-paste of the friendly values in SLI (https://github.com/SonarSource/sonarqube/blob/master/server/sonar-web/src/main/js/helpers/standards.json#L3622).

I'm not a fan of this though so as long as it wasn't requested, there's no reason to create more maintance.

rita-g-sonarsource (commented):
@duncanp-sonar looks ok but see my comment regarding the xaml.

src/IssueViz.Security/Models/IHotspot.cs
duncanp-sonar: It does now ;-) Good catch.

PR #1956, opened 2020-12-01
Fix datagrid double-click sensitivity

### LINKED ISSUE #1953: Hotspots list double-click is not sensitive enough
#### Description

The double-click on the hotspots list doesn't always work.
It sometimes requires multiple double-clicks.

TODO:
- [ ] Fix the double click

rita-g-sonarsource:
Fixes #1953

### DIFF
--- a/src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
+++ b/src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
@@ -221,7 +221,7 @@
             <ColumnDefinition Name="Table" Width="*"/>
         </Grid.ColumnDefinitions>
         <DataGrid Grid.Column="0" 
-                  IsReadOnly="False" 
+                  IsReadOnly="True" 
                   CanUserAddRows="False"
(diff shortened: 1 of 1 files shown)


src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -221,7 +221,7 @@
             <ColumnDefinition Name="Table" Width="*"/>
         </Grid.ColumnDefinitions>
         <DataGrid Grid.Column="0" 
-                  IsReadOnly="False" 
+                  IsReadOnly="True"
```
rita-g-sonarsource, line 224: Based on an elusive comment in this thread:
https://stackoverflow.com/a/19808349/339720

"Note, that if IsReadOnly="False" for your DataGrid, then the first click may be stolen the selection mechanism. Set IsReadOnly="True" to fix that up." 🤦

duncanp-sonar (approved):

PR #1959, opened 2020-12-01
Display security category in Hotspots list

### LINKED ISSUE #1957: Display security category in Hotspots list
Display security category in Hotspots list

TODO
- [x] convert and display security category
https://github.com/SonarSource/sonarqube/blob/master/server/sonar-web/src/main/js/helpers/standards.json#L3622

- [x] check other columns displayed in other SL flavours e.g. assignee --> See #1962

rita-g-sonarsource:
Fixes #1957

### DIFF
--- a/src/IssueViz.Security/HotspotsList/ViewModels/HotspotViewModel.cs
+++ b/src/IssueViz.Security/HotspotsList/ViewModels/HotspotViewModel.cs
@@ -36,6 +36,8 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.ViewMo
         int Column { get; }
 
         string DisplayPath { get; }
+
+        string CategoryDisplayName { get; }
     }
 
     internal sealed class HotspotViewModel : IHotspotViewModel
@@ -77,6 +79,12 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.ViewMo
 
         public string DisplayPath =>
             Path.GetFileName(Hotspot.CurrentFilePath ?? ((IHotspot)Hotspot.Issue).ServerFilePath);
+
+        public string CategoryDisplayName =>
+            SecurityCategoryDisplayNames.Mapping.TryGetValue(
+                ((IHotspot) Hotspot.Issue).Rule.SecurityCategory, out var displayName)
+                ? displayName
+                : string.Empty;
 
         private bool CanUseSpan()
         {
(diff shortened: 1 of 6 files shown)


src/IssueViz.Security/HotspotsList/ViewModels/HotspotViewModel.cs
```diff
@@ -78,6 +80,12 @@ public int Column
         public string DisplayPath =>
             Path.GetFileName(Hotspot.CurrentFilePath ?? ((IHotspot)Hotspot.Issue).ServerFilePath);
 
+        public string CategoryDisplayName =>
+            SecurityCategoryDisplayNames.Mapping.TryGetValue(
+                ((IHotspot) Hotspot.Issue).Rule.SecurityCategory, out var displayName)
+                ? displayName
+                : string.Empty;
```
duncanp-sonar, line 87: I'm not sure _Empty_ is the most useful value we could return here. 
The short forms are fairly readable on their own, so returning that in the hopefully very rare case of us encountering a new mapping would at least give the user some useful information.
Alternatively, something like `{unrecognized}` or `{unknown}` would provide the user with some information about why the category name is not displayed.

rita-g-sonarsource: I was wary of using the unfriendly name, as that might cause users to ignore it and we will miss the potential problems.
{unknown} sounds good 👍

src/IssueViz.Security/HotspotsList/ViewModels/SecurityCategoryDisplayNames.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Generic;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.HotspotsList.ViewModels
+{
+    /// <summary>
+    /// Based on https://github.com/SonarSource/sonarqube/blob/master/server/sonar-web/src/main/js/helpers/standards.json#L3622
+    /// </summary>
+    internal static class SecurityCategoryDisplayNames
+    {
+        public static readonly IDictionary<string, string> Mapping = new Dictionary<string, string>
```
duncanp-sonar, line 30: Seems odd for this to be a public dictionary with the lookup/conversion logic somewhere else. Why not just have a public static method and hide the dictionary?

rita-g-sonarsource: Hmm yep. I wanted to minimize the code of this class, so it would be clear that it's a copy-paste. 
I'll add the method.

duncanp-sonar (approved):

PR #1964, opened 2020-12-02
Better navigability indication

### LINKED ISSUE #1963: [Hotspot list] Improve navigation indicator column UX
### Description

The first column of the hotspot list shows a icon indicating when the location is not navigable.  However, it's not clear what the column unless there is an entry with an icon as the column header doesn't have a description.

Possible UX tweaks:
- make the column fixed size (no point in it being resizable)
 - add a column header / tooltip to describe the columnd (cf the `Severity` column the `Error List` is fixed size and has an explanatory tool tip)
- show icons for locations that are navigable e.g. a small doc with a green tick)
- ?different icon for file that can't be found e.g. doc with red cross?

rita-g-sonarsource:
Fixes #1963 

image
image
image

### DIFF
--- a/src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
+++ b/src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
@@ -2,7 +2,7 @@
              x:ClassModifier="internal"
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-             xmlns:wpf="clr-namespace:SonarLint.VisualStudio.Core.WPF;assembly=SonarLint.VisualStudio.Core"
+             xmlns:core="clr-namespace:SonarLint.VisualStudio.Core.WPF;assembly=SonarLint.VisualStudio.Core"
              xmlns:issueVisualizationControl="clr-namespace:SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl;assembly=SonarLint.VisualStudio.IssueVisualization"
              xmlns:vsShell="clr-namespace:Microsoft.VisualStudio.Shell;assembly=Microsoft.VisualStudio.Shell.14.0"
              xmlns:vsCatalog="clr-namespace:Microsoft.VisualStudio.Imaging;assembly=Microsoft.VisualStudio.ImageCatalog"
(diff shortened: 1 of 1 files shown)


src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -252,12 +256,23 @@
...
                         <DataTemplate>
-                            <vsImaging:CrispImage Height="16" Width="16"
-                                                  Moniker="{x:Static vsCatalog:KnownMonikers.DocumentWarning}" 
-                                                  Visibility="{Binding Hotspot.Span, Converter={StaticResource SpanToInvertedVisibilityConverter}, Mode=OneWay}"/>
+                            <Grid>
+                                <vsImaging:CrispImage Height="16" Width="16"
+                                                      Moniker="{x:Static vsCatalog:KnownMonikers.DocumentWarning}" 
+                                                      Visibility="{Binding Hotspot.Span, Converter={StaticResource SpanToInvertedVisibilityConverter}, Mode=OneWay}"/>
+                                <vsImaging:CrispImage Height="16" Width="16"
+                                                      ToolTipService.ToolTip="Location is navigable. Use double-click or Enter to open the document."
```
rita-g-sonarsource, line 272: Note that the other image doesn't have an explicit tooltip.
This is because when a location is not navigable, there is a tooltip on the entire row (set by style https://github.com/SonarSource/sonarlint-visualstudio/blob/6a322aaa3d75612961a0f9077e65fd0b2853e2c9/src/IssueViz.Security/HotspotsList/HotspotsControl.xaml#L131)

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -252,12 +256,23 @@
                     CommandParameter="{Binding Hotspots/}"/>
             </DataGrid.InputBindings>
             <DataGrid.Columns>
-                <DataGridTemplateColumn CanUserSort="False" Width="25">
+                <DataGridTemplateColumn CanUserSort="False" CanUserResize="False" Width="30">
```
rita-g-sonarsource, line 259: Note the new CanUserResize

rita-g-sonarsource:
@duncanp-sonar let me know what you think of the icon. I'm slightly worried that it implies that the hotspot is fixed or something ("the document is ok" == no hotspot?). I think that something blue might be better than green, e.g. "go to source code":
image

or maybe "document source", although they're not the same size but it might be fixable:
image

Fun fact: there are "go to hotspot" and "hotspot" icons:
image

duncanp-sonar:
> let me know what you think of the icon. I'm slightly worried that it implies that the hotspot is fixed
Overall I think it's an improvement; let's see what feedback we get (if any). I agree that blue rather than green might be better, but unless it's very easy to change I wouldn't bother doing anything at this point.

Also, I wouldn't make any more changes until we've decide if and how we are going to display the real "status" field.

duncanp-sonar (approved):

rita-g-sonarsource:
> > let me know what you think of the icon. I'm slightly worried that it implies that the hotspot is fixed
> > Overall I think it's an improvement; let's see what feedback we get (if any). I agree that blue rather than green might be better, but unless it's very easy to change I wouldn't bother doing anything at this point.
> 
> Also, I wouldn't make any more changes until we've decide if and how we are going to display the real "status" field.

So I'll change to the blue then ("document source"), it's a one liner.

PR #1966, opened 2020-12-03
Tweak hotspot list column widths

duncanp-sonar:


src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -240,7 +240,8 @@
                   IsSynchronizedWithCurrentItem="True"
                   SelectedItem="{Binding SelectedHotspot, Mode=TwoWay}"
                   ItemsSource="{Binding Hotspots}"
-                  vsTheming:ImageThemingUtilities.ImageBackgroundColor="{Binding Background, RelativeSource={RelativeSource Self}, Converter={StaticResource BrushToColorConverter}}">
+                  vsTheming:ImageThemingUtilities.ImageBackgroundColor="{Binding Background, RelativeSource={RelativeSource Self}, Converter={StaticResource BrushToColorConverter}}"
+                  HorizontalScrollBarVisibility="Auto">
```
duncanp-sonar, line 244: It's possible to manually resize the columns and tool window so the columns aren't all visible, in which case we'll want a scrollbar to appear.

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -256,7 +257,7 @@
                     CommandParameter="{Binding Hotspots/}"/>
             </DataGrid.InputBindings>
             <DataGrid.Columns>
-                <DataGridTemplateColumn CanUserSort="False" CanUserResize="False" Width="30">
+                <DataGridTemplateColumn CanUserSort="False" CanUserResize="False" Width="25">
```
duncanp-sonar, line 260: Can be slightly smaller and still show the icon - leaves more space for the variable-length columns.

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -277,55 +278,55 @@
                         </DataTemplate>
                     </DataGridTemplateColumn.CellTemplate>
                 </DataGridTemplateColumn>
-                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="Hotspot.Issue.RuleKey" Header="Code" Width="130">
+                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="Hotspot.Issue.RuleKey" Header="Code" Width="130" MinWidth="50">
```
duncanp-sonar, line 281: I've set reasonable minimum widths for the columns so they are still more or less usable with the default font size.

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -277,55 +278,55 @@
...
+                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="CategoryDisplayName" Header="Category" Width="Auto"  MinWidth="50">
                     <DataGridTemplateColumn.CellTemplate>
                         <DataTemplate>
                             <TextBlock Text="{Binding CategoryDisplayName}" 
                                        Style="{StaticResource CellTextBlockStyle}"/>
                         </DataTemplate>
                     </DataGridTemplateColumn.CellTemplate>
                 </DataGridTemplateColumn>
-                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="Hotspot.Issue.Rule.Priority" Header="Priority" Width="100">
+                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="Hotspot.Issue.Rule.Priority" Header="Priority" Width="90"  MinWidth="50">
```
duncanp-sonar, line 297: Another default width that can be a bit smaller but still show the longest priority string with the default font.

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -277,55 +278,55 @@
...
+                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="Hotspot.Issue.Message" Header="Description" Width="*" MinWidth="150">
                     <DataGridTemplateColumn.CellTemplate>
                         <DataTemplate>
                             <TextBlock Text="{Binding Hotspot.Issue.Message}" 
                                        Style="{StaticResource CellTextBlockStyle}"/>
                         </DataTemplate>
                     </DataGridTemplateColumn.CellTemplate>
                 </DataGridTemplateColumn>
-                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="DisplayPath" Header="File" Width="200">
+                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="DisplayPath" Header="File" Width="180"  MinWidth="80">
```
duncanp-sonar, line 313: See how you think this and the changes to `Line` and `Column` look. With the examples I tried they all had more whitespace than necessary so I reduced them slightly.

I was doing this to maximize the space available for the help pane, but it still makes sense not to allocate more space than is needed by.

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
rita-g-sonarsource: It [used to be 25](https://github.com/SonarSource/sonarlint-visualstudio/pull/1964#discussion_r534268798) and it is ever so slightly cut:
image

Also, I see now that the icons are not aligned:
image

They never aligned due to the way they are in VS so I added a margin on one of them to sync them. But that margin is now out of date due to the column change. So for width 25 I'd change the icon's margin to -2.5.

I did some experiments, I think width 27 and margin -3.5 is looking good as well. The icon is cut by 1-2 pixels and it's hardly noticeable:
image

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -277,55 +278,55 @@
...
+                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="Hotspot.Issue.RuleKey" Header="Code" Width="130" MinWidth="50">
                     <DataGridTemplateColumn.CellTemplate>
                         <DataTemplate>
                             <TextBlock Text="{Binding Hotspot.Issue.RuleKey}" 
                                        Style="{StaticResource CellTextBlockStyle}" />
                         </DataTemplate>
                     </DataGridTemplateColumn.CellTemplate>
                 </DataGridTemplateColumn>
-                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="CategoryDisplayName" Header="Category" Width="Auto">
+                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="CategoryDisplayName" Header="Category" Width="Auto"  MinWidth="50">
```
rita-g-sonarsource, line 289: MinWidth is a bit small

image

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
```diff
@@ -277,55 +278,55 @@
...
+                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="DisplayPath" Header="File" Width="180"  MinWidth="80">
                     <DataGridTemplateColumn.CellTemplate>
                         <DataTemplate>
                             <TextBlock Text="{Binding DisplayPath}" 
                                        Style="{StaticResource CellTextBlockStyle}"/>
                         </DataTemplate>
                     </DataGridTemplateColumn.CellTemplate>
                 </DataGridTemplateColumn>
-                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="Line" Header="Line" Width="50">
+                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="Line" Header="Line" Width="40"  MinWidth="25">
```
rita-g-sonarsource, line 321: MinWidth is a bit small
image

rita-g-sonarsource (commented):
@duncanp-sonar please see my comments.

Also, I noticed that the MinWidth does not account for the sorting arrow. I has a margin of 5 and it disappears if the column width is too smal.
Compared to the error list, they give priority to the sorting arrow and it always appears:
image

I think we should just set the minimal width to always include the arrow (e.g. +5-10 to the width). It's simpler than the more sophisticated "..." that VS does.

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
rita-g-sonarsource: See my general comment please - it does not leave space for the sorting arrow (not sure it did before either)

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
duncanp-sonar: Hhmm. Thanks for trying those variations. I'll switch it back to 30 to avoid it being cut for now. We can tweak it again later if/when we add a help pane.

src/IssueViz.Security/HotspotsList/HotspotsControl.xaml
duncanp-sonar: I was setting the minimum widths so the columns are "more or less usable" if set to a minimum width. I wasn't trying to set them so that all of the text was always visible, or the sort arrows.
However, I'll increase the minimums slightly.

duncanp-sonar:
> I think we should just set the minimal width to always include the arrow

Ok. In the case of Line and Col it means increasing the default width too.

Issue #1989, opened 2020-12-07
Support tagging of taint issues


rita-g-sonarsource:
Implement IIssueLocationStore

PR #2005, opened 2020-12-09
Add common IIssueLocationStore for hotspots and taint

rita-g-sonarsource:
Part of #1989


src/IssueViz.Security.UnitTests/IssueVizsStoreTests.cs
```diff
@@ -0,0 +1,188 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests
+{
+    [TestClass]
+    public class IssueVizsStoreTests
```
rita-g-sonarsource, line 15: Relevant tests copied from HotspotsStoreTests

src/IssueViz.Security/IssueVizsStore.cs
```diff
@@ -0,0 +1,101 @@
...
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security
+{
+    internal interface IIssueVizsStore : IIssueLocationStore, IDisposable
+    {
+        ReadOnlyObservableCollection<IAnalysisIssueVisualization> GetAll();
+    }
+
+    internal sealed class IssueVizsStore : IIssueVizsStore
```
rita-g-sonarsource, line 36: This is a repeating identical logic between hotspots and taint, and we already had a bug in it (the null path, which will be relevant for taint as well). I created a common Store for both of them to use.
Take a look at how the usage will be like:
https://github.com/SonarSource/sonarlint-visualstudio/pull/2006/files#diff-1c9f171be0a53939909d03bedba2f81a5e7782a0a80c2c0b78dc913d74d360b7

src/IssueViz.Security/ObservingIssueLocationStore.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Linq;
+using SonarLint.VisualStudio.Core.Helpers;
+using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security
+{
+    internal interface IObservingIssueLocationStore : IIssueLocationStore, IDisposable
+    {
+        void Register(ObservableCollection<IAnalysisIssueVisualization> issueVisualizations);
```
rita-g-sonarsource, line 34: I'm debating a bit if it should be `ReadOnlyObservableCollection`. It makes more sense and would protect the data from accidental changes, but it makes the usage much more complicated, as the callers have to call Unregister with the exact same wrapper instance that they used to call Register (which isn't a trivial assumption). I can add tests for the callers that verify that it's the same ref instance.
WDYT?

src/IssueViz.Security/ObservingIssueLocationStore.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections.ObjectModel;
+using System.ComponentModel.Composition;
+using System.Linq;
+using SonarLint.VisualStudio.Core.Helpers;
+using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security
+{
+    internal interface IObservingIssueLocationStore : IIssueLocationStore, IDisposable
```
duncanp-sonar, line 32: Does this interface need to inherit from 'IIssueLocationStore'?

The consumers of the additional methods on this interface (the individual issue stores) don't need to know anything about the `IIssueLocationStore`, do they? From their point of view, they just interested in registering themselves with an issue store observer/sink - they don't need to know what the observer/sink does.

Something like `IIssueStoreObserver`, `IIssueStoreSink` instead?

src/IssueViz.Security/ObservingIssueLocationStore.cs
```diff
@@ -0,0 +1,123 @@
...
+                .SelectMany(issueVizsCollection => issueVizsCollection )
+                .SelectMany(issueViz => issueViz.GetAllLocations())
+                .Where(locationViz => !string.IsNullOrEmpty(locationViz.CurrentFilePath) &&
+                                      PathHelper.IsMatchingPath(locationViz.CurrentFilePath, filePath));
+
+            return matchingLocations;
+        }
+
+        public void Refresh(IEnumerable<string> affectedFilePaths)
+        {
```
duncanp-sonar, line 81: Worth commenting why an implementation isn`t required.

src/IssueViz.Security/ObservingIssueLocationStore.cs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    [Export(typeof(IObservingIssueLocationStore))]
+    [Export(typeof(IIssueLocationStore))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal sealed class ObservingIssueLocationStore : IObservingIssueLocationStore
+    {
+        private IList<ObservableCollection<IAnalysisIssueVisualization>> ObservableIssueVisualizations { get; } = new List<ObservableCollection<IAnalysisIssueVisualization>>();
+
+        public void Register(ObservableCollection<IAnalysisIssueVisualization> issueVisualizations)
```
duncanp-sonar, line 45: Assuming `IObservingIssueLocationStore` doesn't need to implement `IIssueLocationStore`, could you make it clearer which methods belong to which interface? i.e. explicit implementation or regions.

That would make it easier to see the two distinct faces of this class - the "input" side used by the stores that provide the data, and the "output" side used by the consumers.

src/IssueViz.Security/ObservingIssueLocationStore.cs
```diff
@@ -0,0 +1,123 @@
...
+        void Register(ObservableCollection<IAnalysisIssueVisualization> issueVisualizations);
+        void Unregister(ObservableCollection<IAnalysisIssueVisualization> issueVisualizations);
+    }
+
+    [Export(typeof(IObservingIssueLocationStore))]
+    [Export(typeof(IIssueLocationStore))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal sealed class ObservingIssueLocationStore : IObservingIssueLocationStore
+    {
+        private IList<ObservableCollection<IAnalysisIssueVisualization>> ObservableIssueVisualizations { get; } = new List<ObservableCollection<IAnalysisIssueVisualization>>();
```
duncanp-sonar, line 43: List or HashSet? We're only expecting an item to be registered once.

src/IssueViz.Security/ObservingIssueLocationStore.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        public void Register(ObservableCollection<IAnalysisIssueVisualization> issueVisualizations)
+        {
+            if (issueVisualizations == null)
+            {
+                throw new ArgumentNullException(nameof(issueVisualizations));
+            }
```
duncanp-sonar, line 50: Perhaps assert that the object hasn't already been registered? If it has then there's something wrong with the logic in the calling code.

duncanp-sonar (approved):
Nice - much simpler! Just a few questions/comments.

src/IssueViz.Security/ObservingIssueLocationStore.cs
duncanp-sonar: Doesn't the caller have to provide the same instance anyway? Otherwise `ObservableIssueVisualizations.Remove(issueVisualizations)` won't be able to remove it, or to unregister from the event.

src/IssueViz.Security/ObservingIssueLocationStore.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        public void Unregister(ObservableCollection<IAnalysisIssueVisualization> issueVisualizations)
+        {
+            if (issueVisualizations == null)
+            {
+                throw new ArgumentNullException(nameof(issueVisualizations));
+            }
+
+            ObservableIssueVisualizations.Remove(issueVisualizations);
```
duncanp-sonar, line 63: Should this check that the item is actually in the collection?

src/IssueViz.Security/ObservingIssueLocationStore.cs
rita-g-sonarsource: You're right. I changed Register to
`IDisposable Register(ReadOnlyObservableCollection<IAnalysisIssueVisualization> issueVisualizations);`

src/IssueViz.Security/ObservingIssueLocationStore.cs
rita-g-sonarsource: Good idea, I went with `IIssueStoreObserver`

src/IssueViz.Security/ObservingIssueLocationStore.cs
rita-g-sonarsource: I changed the Register to return unregister callback, so now it will definitely be there.

PR #2006, opened 2020-12-09
Use common IssueStoreObserver in HotspotsStore

### LINKED ISSUE #1989: Support tagging of taint issues

rita-g-sonarsource:
Part of #1989


src/IssueViz.Security/Hotspots/HotspotsStore.cs
```diff
@@ -46,10 +43,23 @@ internal interface IHotspotsStore
     [Export(typeof(IHotspotsStore))]
```
rita-g-sonarsource, line 43: I'm not quite satisfied with the MEF imports.
HotspotsStore is heavily aware of the inner implementation of IIssueVizsStore and it's underlying interfaces.

duncanp-sonar: Imports, or exports (there aren't any imports...).

What does the hotspots store know about the implementation of contained store that it shouldn't? It knows it implements a specific interface; anything else?

Having said that, it doesn't feel right. It looks like you are trying too hard to avoid using inheritance when that might be a cleaner implementation.

src/IssueViz.Security/Hotspots/HotspotsStore.cs
```diff
@@ -97,25 +80,11 @@ private IAnalysisIssueVisualization FindExisting(IAnalysisIssueVisualization hot
...
-        IEnumerable<IAnalysisIssueLocationVisualization> IIssueLocationStore.GetLocations(string filePath)
-        {
-            var matchingLocations = Hotspots
-                .SelectMany(hotspotViz => hotspotViz.GetAllLocations())
-                .Where(locationViz => !string.IsNullOrEmpty(locationViz.CurrentFilePath) &&
-                                      PathHelper.IsMatchingPath(locationViz.CurrentFilePath, filePath));
-
-            return matchingLocations;
-        }
+        public ReadOnlyObservableCollection<IAnalysisIssueVisualization> GetAll() => ReadOnlyWrapper;
```
duncanp-sonar, line 83: I wasn't very comfortable with the way the old code return a new wrapper for every call - this feels nicer.

src/IssueViz.Security/Hotspots/HotspotsStore.cs
```diff
@@ -44,13 +41,20 @@ internal interface IHotspotsStore
     }
 
     [Export(typeof(IHotspotsStore))]
-    [Export(typeof(IIssueLocationStore))]
     [PartCreationPolicy(CreationPolicy.Shared)]
-    internal class HotspotsStore : IHotspotsStore, IIssueLocationStore
+    internal sealed class HotspotsStore : IHotspotsStore
```
duncanp-sonar, line 45: This class is a bit complicated - could you make it simpler? ;-)

duncanp-sonar (approved):
Looks good - much simpler.

Issue #1988, opened 2020-12-07
Glue-code orchestator


rita-g-sonarsource:
1. Fetch issues (depends on #1990)
2. Convert SQ issues to IssueViz
3. ~~Fetch source code~~ -> will be needed for secondary locations
4. ~~Calculate line hashes~~ -> will be needed for secondary locations
5. Populate list (depends on #1989)

PR #2007, opened 2020-12-10
Add TaintStore

### LINKED ISSUE #1989: Support tagging of taint issues

rita-g-sonarsource:
Fixes #1989

### DIFF
--- a/src/IssueViz.Security.UnitTests/Taint/TaintStoreTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintStoreTests.cs
@@ -99,6 +99,16 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
             readOnlyWrapper.First().Should().Be(issueViz2);
         }
 
+        [TestMethod]
+        public void Initialize_NullCollection_ArgumentNullException()
+        {
+            var testSubject = CreateTestSubject();
+
+            Action act = () => testSubject.Initialize(null);
+
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("issueVisualizations");
+        }
+
         private ITaintStore CreateTestSubject(IIssueStoreObserver issueStoreObserver = null)
         {
             issueStoreObserver ??= Mock.Of<IIssueStoreObserver>();
--- a/src/IssueViz.Security.UnitTests/Taint/TaintStoreTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintStoreTests.cs
@@ -1,0 +1,108 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Security.Taint;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
+{
+    [TestClass]
+    public class TaintStoreTests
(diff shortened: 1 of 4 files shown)


src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (issueVisualizations == null)
+            {
+                throw new ArgumentNullException(nameof(issueVisualizations));
+            }
+
+            TaintVulnerabilities.Clear();
+
+            foreach (var issueViz in issueVisualizations)
+            {
+                TaintVulnerabilities.Add(issueViz);
```
rita-g-sonarsource, line 66: This is slightly annoying: there is no AddRange method for an observable collection.
I thought about calling `issueStoreObserver.Register` only in Initialize, instead of the ctor, but the xaml view model will likely be another consumer of the observable (by calling GetAll()). So in the end I just left it with Add() and I hope the performance won't be awful with the multiple redraws.

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        [ImportingConstructor]
+        public TaintStore(IIssueStoreObserver issueStoreObserver)
+        {
+            unregisterCallback = issueStoreObserver.Register(ReadOnlyWrapper);
+        }
+
+        public ReadOnlyObservableCollection<IAnalysisIssueVisualization> GetAll() => ReadOnlyWrapper;
+
+        public void Initialize(IEnumerable<IAnalysisIssueVisualization> issueVisualizations)
```
rita-g-sonarsource, line 55: I expect Initialize to be called in the glue-code orchestrator (#1988) when the binding changes, and perhaps every X time if we decide to add a periodic pull.

src/IssueViz.Security.UnitTests/Taint/TaintStoreTests.cs
```diff
@@ -0,0 +1,108 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Security.Taint;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
+{
+    [TestClass]
+    public class TaintStoreTests
```
rita-g-sonarsource, line 35: Mostly copy-paste of the HotspotsStoreTests

src/IssueViz.Security/Taint/TaintStore.cs
duncanp-sonar: 'Initialize' methods are generally called only once whereas we expect this to be called multiple times, which feels a bit odd.
I'm not sure what else I'd call it, though - `SetAll`? `Reset`?

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -0,0 +1,75 @@
...
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
+{
+    internal interface ITaintStore : IDisposable
+    {
+        ReadOnlyObservableCollection<IAnalysisIssueVisualization> GetAll();
+
+        /// <summary>
+        /// Removes all existing visualizations and initializes the store to the given collection.
```
duncanp-sonar, line 34: Worth commenting that this can be called multiple times.

src/IssueViz.Security/Taint/TaintStore.cs
duncanp-sonar: Hhmm. Good point. We'll have to try it and see. It would be worth adding a comment here to remind us there are perf implications.

If there's a problem we could call `Unregister` / `Register` in this method.

Thinking about it, the common store should notify listeners when a new collection is registered or unregistered, which it doesn't currently do.

src/IssueViz.Security.UnitTests/Taint/TaintStoreTests.cs
duncanp-sonar: In future, it would help if you called out the bits that aren't copy-pasted, otherwise the reviewer still has to look at the whole file.

src/IssueViz.Security.UnitTests/Taint/TaintStoreTests.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            unregisterMock.VerifyNoOtherCalls();
+
+            testSubject.Dispose();
+
+            unregisterMock.Verify(x => x.Dispose(), Times.Once);
+        }
+
+        [TestMethod]
+        public void GetAll_ReturnsReadOnlyObservableWrapper()
```
duncanp-sonar, line 80: Is it worth testing that `GetAll` returns the same instance each time? I`m don't think it matters to the caller, but it is what we now expect from the implementation.

duncanp-sonar (approved):
Looks ok; although it makes me think that there is an issue in the store observer (it should notify listeners when register/unregister is called).

src/IssueViz.Security/Taint/TaintStore.cs
rita-g-sonarsource: Oh you're right, definitely. I'll add it.

src/IssueViz.Security/Taint/TaintStore.cs
rita-g-sonarsource: Maybe just 'Set'?

src/IssueViz.Security.UnitTests/Taint/TaintStoreTests.cs
rita-g-sonarsource: I don't think it matters. There's no difference in the store nor in the callers, it can be implemented either way.

PR #2008, opened 2020-12-10
Move IActiveSolutionBoundTracker to Core

rita-g-sonarsource:
Part of #1988.
Will be needed in order to trigger fetching taint issues when the binding changes.


src/Core/Binding/IActiveSolutionBoundTracker.cs
```diff
@@ -44,4 +43,14 @@ public interface IActiveSolutionBoundTracker
 
         event EventHandler SolutionBindingUpdated;
     }
+
+    public class ActiveSolutionBindingEventArgs : EventArgs
```
rita-g-sonarsource, line 47: This was in another file, I just placed it along-side the interface

duncanp-sonar (approved):

PR #2011, opened 2020-12-10
Extracted display position calculation from the hotspot view model

duncanp-sonar:
Extracting common code that will be used by the taint view model

### DIFF
--- a/src/IssueViz.Security.UnitTests/Hotspots/HotspotsList/HotspotViewModelTests.cs
+++ b/src/IssueViz.Security.UnitTests/Hotspots/HotspotsList/HotspotViewModelTests.cs
@@ -97,47 +97,33 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Hotspots.
         }
 
         [TestMethod]
-        [DataRow(true)]
-        [DataRow(false)]
-        public void Line_NoSpan_ReturnsHotspotStartLine(bool spanIsNull)
+        public void Line_ReturnsDisplayPosition()
         {
-            const int originalLineNumber = 123;
-            var issueViz = CreateIssueVizWithoutSpan(spanIsNull, originalLineNumber: originalLineNumber);
+            var issueViz = Mock.Of<IAnalysisIssueVisualization>();
 
-            var testSubject = CreateTestSubject(issueViz);
-            testSubject.Line.Should().Be(originalLineNumber);
+            var positionCalculatorMock = new Mock<IIssueVizDisplayPositionCalculator>();
+            positionCalculatorMock.Setup(x => x.GetLine(issueViz)).Returns(123);
+
+            // Act
+            var testSubject = CreateTestSubject(issueViz, positionCalculator: positionCalculatorMock.Object);
+
+            testSubject.Line.Should().Be(123);
+            positionCalculatorMock.VerifyAll();
         }
 
         [TestMethod]
-        [DataRow(true)]
-        [DataRow(false)]
-        public void Column_NoSpan_ReturnsOneBasedHotspotStartLineOffset(bool spanIsNull)
+        public void Column_ReturnsDisplayPosition()
         {
-            const int originalColumnNumber = 456;
-            var issueViz = CreateIssueVizWithoutSpan(spanIsNull, originalColumnNumber: originalColumnNumber);
+            var issueViz = Mock.Of<IAnalysisIssueVisualization>();
 
-            var testSubject = CreateTestSubject(issueViz);
-            testSubject.Column.Should().Be(originalColumnNumber + 1);
-        }
+            var positionCalculatorMock = new Mock<IIssueVizDisplayPositionCalculator>();
+            positionCalculatorMock.Setup(x => x.GetColumn(issueViz)).Returns(999);
 
-        [TestMethod]
-        public void Line_HasSpan_ReturnsOneBasedSpanStartLine()
-        {
-            const int lineNumber = 12;
-            var issueViz = CreateIssueVizWithSpan(lineNumber: lineNumber);
+            // Act
+            var testSubject = CreateTestSubject(issueViz, positionCalculator: positionCalculatorMock.Object);
 
-            var testSubject = CreateTestSubject(issueViz);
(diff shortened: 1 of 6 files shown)


src/IssueViz.Security.UnitTests/Hotspots/HotspotsList/HotspotViewModelTests.cs
```diff
@@ -97,47 +97,33 @@ public void HotspotPropertyChanged_SpanProperty_RaisesPropertyChangedForLineAndC
         }
 
         [TestMethod]
-        [DataRow(true)]
```
duncanp-sonar, line 100: The deleted tests were moved to the new test class.

src/IssueViz.Security.UnitTests/Hotspots/HotspotsList/HotspotViewModelTests.cs
```diff
@@ -97,47 +97,33 @@ public void HotspotPropertyChanged_SpanProperty_RaisesPropertyChangedForLineAndC
...
-        [DataRow(false)]
-        public void Line_NoSpan_ReturnsHotspotStartLine(bool spanIsNull)
+        public void Line_ReturnsDisplayPosition()
```
duncanp-sonar, line 100: Two new tests that check the position calculator is being used.

src/IssueViz.Security.UnitTests/Hotspots/HotspotsList/HotspotsControlViewModelTests.cs
```diff
@@ -35,6 +36,13 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Hotspots.
     [TestClass]
     public class HotspotsControlViewModelTests
     {
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            // HotspotsControlViewModel needs to be created on the UI thread
```
duncanp-sonar, line 42: Miscellaneous change to this and the tool window tests.
We must have been lucky with the test order execution before i.e. one of the earlier tests must have already called `SetCurrentThreadAsUIThread`.

src/IssueViz.Security.UnitTests/IssueVizDisplayPositionCalculatorTests.cs
```diff
@@ -0,0 +1,117 @@
...
+        [TestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public void Line_NoSpan_ReturnsHotspotStartLine(bool spanIsNull)
+        {
+            const int originalLineNumber = 123;
+            var issueViz = CreateIssueVizWithoutSpan(spanIsNull, originalLineNumber: originalLineNumber);
+
+            var testSubject = CreateTestSubject();
+            testSubject.GetLine(issueViz).Should().Be(originalLineNumber);
```
duncanp-sonar, line 42: Cut and pasted from the view model tests with a few changes... but I'm not going to tell you what they are - MWAA-HA-HA-HA-HA!!!! ;-)

Ok, 'CreateTestSubject' is obviously different, and the calls to the `Line` and `Column` properties are now calls to `GetLine` and `GetProperty`. Everything else is identical.

src/IssueViz.Security/IssueVizDisplayPositionCalculator.cs
```diff
@@ -0,0 +1,63 @@
...
+    internal interface IIssueVizDisplayPositionCalculator
+    {
+        int GetLine(IAnalysisIssueVisualization issueViz);
+
+        int GetColumn(IAnalysisIssueVisualization issueViz);
+    }
+
+    internal class IssueVizDisplayPositionCalculator : IIssueVizDisplayPositionCalculator
+    {
+        int IIssueVizDisplayPositionCalculator.GetColumn(IAnalysisIssueVisualization issueViz)
```
duncanp-sonar, line 35: Code extracted from the view model. No changes to the code, other than to make it use the supplied `issueViz` parameter instead of a property.

src/IssueViz.Security/IssueVizDisplayPositionCalculator.cs
```diff
@@ -0,0 +1,63 @@
...
+                zeroBasedColumn = issueViz.Issue.StartLineOffset;
+            }
+            else
+            {
+                var position = issueViz.Span.Value.Start;
+                var line = position.GetContainingLine();
+                zeroBasedColumn = position.Position - line.Start.Position;
+            }
+
+            // both SQ hotspot column and VS column are zero-based
```
rita-g-sonarsource, line 50: Comment is now out of date (says "hotpost").
Are we sure that this is the case for regular issues as well? (I'd assume that yes but it's worth to double check)

src/IssueViz.Security.UnitTests/IssueVizDisplayPositionCalculatorTests.cs
rita-g-sonarsource: touché!

src/IssueViz.Security.UnitTests/IssueVizDisplayPositionCalculatorTests.cs
```diff
@@ -0,0 +1,117 @@
...
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests
+{
+    [TestClass]
+    public class IssueVizDisplayPositionCalculatorTests
+    {
+        [TestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public void Line_NoSpan_ReturnsHotspotStartLine(bool spanIsNull)
```
rita-g-sonarsource, line 36: Misc: Hotspot should be IssueViz

rita-g-sonarsource (approved):
Just a comment regarding names ("Hotspot" => "IssueViz") + to check that the issue's column is also zero-based like in hotspot.

src/IssueViz.Security/IssueVizDisplayPositionCalculator.cs
duncanp-sonar: I checked, they are (it would be very surprising if they weren't, given that hotspots were originally treated as normal issues).

PR #2012, opened 2020-12-10
Add converter from SonarQubeIssue to AnalysisIssueVisualization

rita-g-sonarsource:
Part of #1988

### DIFF
--- a/src/IssueViz.Security/Taint/SonarQubeIssueToIssueVisualizationConverter.cs
+++ b/src/IssueViz.Security/Taint/SonarQubeIssueToIssueVisualizationConverter.cs
@@ -1,0 +1,127 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Linq;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
+{
+    interface ISonarQubeIssueToIssueVisualizationConverter
+    {
+        IAnalysisIssueVisualization Convert(SonarQubeIssue sonarQubeIssue);
+    }
+
+    [Export(typeof(ISonarQubeIssueToIssueVisualizationConverter))]
+    internal class SonarQubeIssueToIssueVisualizationConverter : ISonarQubeIssueToIssueVisualizationConverter
+    {
+        private readonly IAnalysisIssueVisualizationConverter issueVisualizationConverter;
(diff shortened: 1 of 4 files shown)


src/IssueViz.Security/Taint/SonarQubeIssueToIssueVisualizationConverter.cs
```diff
@@ -0,0 +1,127 @@
...
+                    location.TextRange.EndLine,
+                    location.TextRange.StartOffset,
+                    location.TextRange.EndOffset,
+                    null);
+            }).ToArray();
+
+        /// <summary>
+        /// Converts from the sonarqube issue severity enum to the standard AnalysisIssueSeverity
+        /// </summary>
+        internal /* for testing */ static AnalysisIssueSeverity Convert(SonarQubeIssueSeverity issueSeverity)
```
rita-g-sonarsource, line 107: Identical to https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/SonarLintDaemon/SonarLintDaemon.cs#L389

src/IssueViz.Security/Taint/SonarQubeIssueToIssueVisualizationConverter.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            var analysisIssue = ConvertToAnalysisIssue(sonarQubeIssue);
+            var issueViz = issueVisualizationConverter.Convert(analysisIssue);
+
+            return issueViz;
+        }
+
+        private IAnalysisIssueBase ConvertToAnalysisIssue(SonarQubeIssue sonarQubeIssue)
+        {
+            if (sonarQubeIssue.TextRange == null)
```
rita-g-sonarsource, line 60: I don't think it's possible to have a vulnerability without a span or to have a location without a span, but I'll check tomorrow with Alex/Marco.

src/IssueViz.Security/Taint/SonarQubeIssueToIssueVisualizationConverter.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                throw new ArgumentNullException(nameof(sonarQubeIssue.TextRange));
+            } 
+
+            var localFilePath = absoluteFilePathLocator.Locate(sonarQubeIssue.FilePath);
+
+            return new AnalysisIssue(
```
rita-g-sonarsource, line 67: I'm using the regular AnalysisIssue for now (like we do in the deamon and cfamily code). I didn't see the need for a dedicated Taint class (yet).

src/IssueViz.Security/Taint/SonarQubeIssueToIssueVisualizationConverter.cs
```diff
@@ -0,0 +1,127 @@
...
+                sonarQubeIssue.RuleId,
+                Convert(sonarQubeIssue.Severity),
+                AnalysisIssueType.Vulnerability,
```
duncanp-sonar, line 70: There's nothing in the naming of this interface/class to suggest that it's specific to vulnerabilities. If it's a general purpose class it should handle all issue types.

src/IssueViz.Security/Taint/SonarQubeIssueToIssueVisualizationConverter.cs
duncanp-sonar: I don't think it's worth introducing a separate converter for this. However, the duplication isn't good either - a shared static class+method would be better.

duncanp-sonar (commented):
There are a couple of things I think should be tweaked; the class shouldn't have a general-purpose name if it's hard-coded to assume the issue is a vulnerability, and the code duplication could be avoided.

src/IssueViz.Security/Taint/SonarQubeIssueToIssueVisualizationConverter.cs
rita-g-sonarsource: I'm sorry I should've clarified. It's identical in terms of logic, it's different enums. SonarQubeIssueSeverity vs DaemonIssueType.

duncanp-sonar: And the 'DaemonIssueType' is part of the gRPC message format so we can't share a single enum. Ok.

src/IssueViz.Security/Taint/SonarQubeIssueToIssueVisualizationConverter.cs
rita-g-sonarsource: You're right. I renamed it to `TaintIssueToIssueVisualizationConverter` for now, as it also verifies that there is a location (so it's definitely not right for all issues).

duncanp-sonar (approved):

Issue #1985, opened 2020-12-07
Add Taint list control


rita-g-sonarsource:
### Description

Add basic list and columns, with basic functionality.

- [x] themes
- [x] sorting
- [x] nav icons/tooltips
- [x] navigate to code functionality

Not included in this ticket:
* synchronizing with the flow pane

PR #2014, opened 2020-12-11
Added initial taint view model implementations

duncanp-sonar:
Relates to #1985

### DIFF
--- a/src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
@@ -1,0 +1,167 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.ObjectModel;
+using System.Linq;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Editor;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Security.Taint;
+using SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.ViewModels;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint.TaintList
+{
+    [TestClass]
+    public class TaintIssuesControlViewModelTests
+    {
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            // The ViewModel needs to be created on the UI thread
+            ThreadHelper.SetCurrentThreadAsUIThread();
+        }
+
+        [TestMethod]
+        public void Ctor_RegisterToSourceCollectionChanges()
+        {
+            var storeCollection = new ObservableCollection<IAnalysisIssueVisualization>();
+            var testSubject = CreateTestSubject(storeCollection);
+
+            var issueViz1 = Mock.Of<IAnalysisIssueVisualization>();
+            var issueViz2 = Mock.Of<IAnalysisIssueVisualization>();
+            var issueViz3 = Mock.Of<IAnalysisIssueVisualization>();
+
+            storeCollection.Add(issueViz1);
+
+            testSubject.Issues.Count.Should().Be(1);
+            testSubject.Issues[0].TaintIssueViz.Should().Be(issueViz1);
+
+            storeCollection.Add(issueViz2);
+            storeCollection.Add(issueViz3);
+
+            testSubject.Issues.Count.Should().Be(3);
+            testSubject.Issues[0].TaintIssueViz.Should().Be(issueViz1);
+            testSubject.Issues[1].TaintIssueViz.Should().Be(issueViz2);
+            testSubject.Issues[2].TaintIssueViz.Should().Be(issueViz3);
+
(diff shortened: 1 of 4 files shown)


src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -0,0 +1,167 @@
...
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Editor;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Security.Taint;
+using SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.ViewModels;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint.TaintList
+{
+    [TestClass]
+    public class TaintIssuesControlViewModelTests
```
duncanp-sonar, line 35: Copied from the equivalent hotspots tests. Only the types and variable names have changed.

Same for the other test class.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssueViewModel.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.IO;
+using System.Runtime.CompilerServices;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.ViewModels
+{
+    internal interface ITaintIssueViewModel : INotifyPropertyChanged, IDisposable
+    {
+        IAnalysisIssueVisualization TaintIssueViz { get; }
```
duncanp-sonar, line 32: The volume of code was getting too large to manage conveniently for review purposes so I changed my approach. I'm adding the necessary classes for a minimal working tool window that will display something, then add functionality incrementally (additional columns, selection, filtering etc).

Breaking up the work should make it easier to review and to identify opportunities for refactoring. We'll also have a working end to end sooner.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssueViewModel.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        int Line { get; }
+
+        int Column { get; }
+
+        string DisplayPath { get; }
+    }
+
+    internal sealed class TaintIssueViewModel : ITaintIssueViewModel
```
duncanp-sonar, line 41: Copied from the hotspot view model with a few bits deleted.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssueViewModel.cs
```diff
@@ -0,0 +1,93 @@
...
+                NotifyPropertyChanged(nameof(Line));
+                NotifyPropertyChanged(nameof(Column));
+            }
+            else if (e.PropertyName == nameof(IAnalysisIssueVisualization.CurrentFilePath))
+            {
+                NotifyPropertyChanged(nameof(DisplayPath));
+            }
+        }
+
+        public event PropertyChangedEventHandler PropertyChanged;
```
duncanp-sonar, line 81: This event and `NotifyPropertyChanged` aren't needed yet. I`ll delete them.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -0,0 +1,116 @@
...
+using Microsoft.VisualStudio.Shell;
+using SonarLint.VisualStudio.IssueVisualization.Editor;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.ViewModels
+{
+    internal interface ITaintIssuesControlViewModel : IDisposable
+    {
+        // WIP: add selection handling
+        // WIP: add any other necessary commands
```
duncanp-sonar, line 38: Again, copied from the hotspots control VM with bits taken out.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -0,0 +1,120 @@
...
+        ICommand NavigateCommand { get; }
+
+        ObservableCollection<ITaintIssueViewModel> Issues { get; }
+    }
+
+    /// <summary>
+    /// View model that surfaces the data in the TaintStore
+    /// </summary>
+    /// <remarks>Monitors the taint store for changes and updates the view model accordingly.</remarks>
+    internal sealed class TaintIssuesControlViewModel : ITaintIssuesControlViewModel
```
rita-g-sonarsource, line 49: Looking at the code, I think the sheer amount of duplication is worthy of a generic component.

I was thinking that maybe we can do something like this:
```
IIssueViewModel
{
 	IAnalysisIssueVisualization TaintIssueViz { get; }
        int Line { get; }
        int Column { get; }
        string DisplayPath { get; }
}

IIssuesControlViewModel<T> where T : IIssueViewModel
{
	ICommand NavigateCommand { get; }
        ObservableCollection<T> Issues { get; }
}

IHotspotViewModel : IIssueViewModel
{
	string CategoryDisplayName { get; }
}

IHotspotsControlViewModel : IIssuesControlViewModel<IHotspotViewModel>
{
	IHotspotViewModel SelectedHotspot { get; }
	ICommand RemoveCommand { get; }
}
```

And create a base class that can contain essentially everything that is in TaintIssuesControlViewModel.
WDYT?

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
rita-g-sonarsource: See my comment regarding the generic component. I think it's too much duplication.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
duncanp-sonar: Yes, I think there is too much duplication and there should be a couple of view model base classes (that's what I was alluding to when I mentioned identifying opportunities for refactoring above - perhaps I should have been more explicit). My plan was to make the changes to the taint tool window and control to get a working end-to-end where both view models are actually being consumed and then refactor to extract the common behaviour.

rita-g-sonarsource (approved):

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
duncanp-sonar: FYI one part of the potentially common view model behaviour that I'm not currently sure about is selection, more specifically integration with the selection service. We currently have two selection services, `IAnalysisIssueSelectionService` and `IHotspotSelectionService`. I don't think we want to add a third one for taint. We might end up with one service that can be consumed in a consistent way by the editor light bulbs, flow window and error/hotspot/taint lists.

PR #2015, opened 2020-12-11
Add TaintIssuesSynchronizer

rita-g-sonarsource:
Part of #1988


src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -0,0 +1,90 @@
...
+using SonarQube.Client;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
+{
+    internal interface ITaintIssuesSynchronizer
+    {
+        /// <summary>
+        /// Fetches taint vulnerabilities from the server, converts them into visualizations and populates <see cref="ITaintStore"/>.
+        /// </summary>
+        Task Sync();
```
duncanp-sonar, line 38: `SynchronizeWithServer`?

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -0,0 +1,90 @@
...
+                return;
+            }
+            try
+            {
+                var projectKey = bindingConfiguration.Project.ProjectKey;
+                var taintVulnerabilities = await sonarQubeService.GetTaintVulnerabilitiesAsync(projectKey, CancellationToken.None);
+
+                logger.WriteLine(TaintResources.Synchronizer_NumberOfServerIssues, taintVulnerabilities.Count);
+
+                var taintIssueVizs = taintVulnerabilities.Select(converter.Convert).ToList();
```
duncanp-sonar, line 81: `ToArray` rather than `ToList`?

It's an interesting question whether there's any value in realising the list in this case; one the one hand, we're only going to iterate through it once, so by create a list/array we're just creating an extra object that needs to be garbage collected.
One the other hand, knowing we're only going to iterate through it once is dependent on the current implementation of the store, which could change.
In practice it won't make much difference as this code isn't performance or memory critical.

src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
```diff
@@ -0,0 +1,178 @@
...
+            var converter = new Mock<ITaintIssueToIssueVisualizationConverter>();
+            converter.Setup(x => x.Convert(serverIssue)).Throws<ArgumentException>();
+            var taintStore = new Mock<ITaintStore>();
+
+            var testSubject = CreateTestSubject(taintStore.Object, converter.Object, serverIssue);
+
+            Func<Task> act = async () => await testSubject.Sync();
+            await act.Should().NotThrowAsync();
+
+            taintStore.Invocations.Count.Should().Be(0);
```
duncanp-sonar, line 105: We'd normally check that the exception is logged.

duncanp-sonar (approved):
A couple of minor comments

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
rita-g-sonarsource: Well, it's more for the tests really. Otherwise I can't assert conversion failures 😛 
Probably won't make a difference.

PR #2017, opened 2020-12-11
Add TaintIssuesBindingMonitor

rita-g-sonarsource:
Part of #1988

### DIFF
--- a/src/IssueViz.Security/Taint/TaintIssuesBindingMonitor.cs
+++ b/src/IssueViz.Security/Taint/TaintIssuesBindingMonitor.cs
@@ -25,12 +25,15 @@ using SonarLint.VisualStudio.Core.Binding;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
 {
+    /// <summary>
+    /// Listens to binding changes and triggers fetching of taint vulnerabilities from the connected server.
+    /// </summary>
     internal interface ITaintIssuesBindingMonitor : IDisposable
     {
-        void ListenToBindingChanges();
     }
 
     [Export(typeof(ITaintIssuesBindingMonitor))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
     internal sealed class TaintIssuesBindingMonitor : ITaintIssuesBindingMonitor
     {
         private readonly IActiveSolutionBoundTracker activeSolutionBoundTracker;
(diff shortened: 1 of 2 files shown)


rita-g-sonarsource:
Nearly last PR of this issue (almost done ™).
The next and last PR will be to call the new services - which is currently a problem, since IssueViz package is not initialized at the start. I will need to create a new package for that, similar to the IDE listener one 😞

src/IssueViz.Security/Taint/TaintIssuesBindingMonitor.cs
```diff
@@ -0,0 +1,76 @@
...
+        private readonly ITaintIssuesSynchronizer taintIssuesSynchronizer;
+
+        [ImportingConstructor]
+        public TaintIssuesBindingMonitor(IActiveSolutionBoundTracker activeSolutionBoundTracker, ITaintIssuesSynchronizer taintIssuesSynchronizer)
+        {
+            this.activeSolutionBoundTracker = activeSolutionBoundTracker;
+            this.taintIssuesSynchronizer = taintIssuesSynchronizer;
+        }
+
+        public void ListenToBindingChanges()
```
duncanp-sonar, line 46: I'm not clear on what the lifecycle of this object is;
* who is going to create it and when, and
* when will the `ListenToBindingChanges` method will be called, and will it be called more than once?

The un-registration and re-registration suggests the method is expected to be called more than once, but I don't understand why that would be necessary (also, I thought re-registering the same event handler callback multiple times was a no-op, so I'm not sure the un-registration is necessary).

If it's only going to be called once on construction, could it be done in the constructor?

src/IssueViz.Security/Taint/TaintIssuesBindingMonitor.cs
```diff
@@ -0,0 +1,76 @@
...
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
+{
+    internal interface ITaintIssuesBindingMonitor : IDisposable
+    {
+        void ListenToBindingChanges();
+    }
+
+    [Export(typeof(ITaintIssuesBindingMonitor))]
+    internal sealed class TaintIssuesBindingMonitor : ITaintIssuesBindingMonitor
```
duncanp-sonar, line 34: Presumably this is a singleton?

duncanp-sonar, line 34: It would be useful to have a doc comment explaining the purpose of the class e.g. `Fetches the taint issues from connected server when the binding changes`

duncanp-sonar:
Apologies, somehow I missed the review request notification and have just looked at this PR. Presumably some of the question (e.g. about package splitting/loading) are no longer relevant following our discussions this morning?

PR #2019, opened 2020-12-14
Added simple columns to taint list

duncanp-sonar:
Relates to #1985

### DIFF
--- a/src/IssueViz.Security.UnitTests/Taint/TaintList/TaintToolWindowTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintList/TaintToolWindowTests.cs
@@ -19,9 +19,15 @@
  */
 
 using System;
+using System.Collections.ObjectModel;
+using Microsoft.VisualStudio.ComponentModelHost;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Editor;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Security.Taint;
 using SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint.TaintList
@@ -29,6 +35,13 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint.Tai
     [TestClass]
     public class TaintToolWindowTests
     {
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            // TiantViewModel needs to be created on the UI thread
+            ThreadHelper.SetCurrentThreadAsUIThread();
+        }
+
         [TestMethod]
         public void Dispose_FrameIsClosed()
         {
@@ -44,7 +57,25 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint.Tai
 
         private TaintToolWindow CreateTestSubject(Mock<IVsWindowFrame> frameMock)
         {
+            var componentModelMock = new Mock<IComponentModel>();
+
+            var store = new Mock<ITaintStore>();
+            store.Setup(x => x.GetAll())
+                .Returns(new ReadOnlyObservableCollection<IAnalysisIssueVisualization>(
+                    new ObservableCollection<IAnalysisIssueVisualization>()));
+
+            componentModelMock
+                .Setup(x => x.GetService<ITaintStore>())
+                .Returns(store.Object);
+
+            componentModelMock
+                .Setup(x => x.GetService<ILocationNavigator>())
+                .Returns(Mock.Of<ILocationNavigator>());
+
             var serviceProviderMock = new Mock<IServiceProvider>();
+            serviceProviderMock
+                .Setup(x => x.GetService(typeof(SComponentModel)))
+                .Returns(componentModelMock.Object);
 
             frameMock ??= new Mock<IVsWindowFrame>();
 
--- a/src/IssueViz.Security/Hotspots/HotspotsList/HotspotsControl.xaml
+++ b/src/IssueViz.Security/Hotspots/HotspotsList/HotspotsControl.xaml
@@ -53,21 +53,9 @@
             <ColumnDefinition Name="Table" Width="*"/>
         </Grid.ColumnDefinitions>
         <DataGrid Grid.Column="0" 
-                  IsReadOnly="True" 
-                  CanUserAddRows="False"
-                  CanUserDeleteRows="False"
-                  CanUserReorderColumns="True"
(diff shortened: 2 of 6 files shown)


src/IssueViz.Security.UnitTests/Taint/TaintList/TaintToolWindowTests.cs
```diff
@@ -44,7 +57,25 @@ public void Dispose_FrameIsClosed()
 
         private TaintToolWindow CreateTestSubject(Mock<IVsWindowFrame> frameMock)
         {
+            var componentModelMock = new Mock<IComponentModel>();
```
duncanp-sonar, line 60: Added services required to create the view model; same as the equivalent hotspot tests.

This files is now almost identical to the hotspot tool window tests, but they each only have one test (of Dispose). I'll have a look at this later as part of the refactoring to see if it's worth removing the duplicated product/test code.

src/IssueViz.Security/Hotspots/HotspotsList/HotspotsControl.xaml
```diff
@@ -2,8 +2,6 @@
              x:ClassModifier="internal"
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-             xmlns:core="clr-namespace:SonarLint.VisualStudio.Core.WPF;assembly=SonarLint.VisualStudio.Core"
```
duncanp-sonar, line 5: Moved some more resources to the shared resources. I manually tested that hotspots navigability still works as expected.

src/IssueViz.Security/Taint/TaintList/TaintVulnerabilitiesControl.xaml
```diff
@@ -1,15 +1,124 @@
...
             <ResourceDictionary.MergedDictionaries>
                 <ResourceDictionary Source="../../SharedUI/SharedResources.xaml" />
             </ResourceDictionary.MergedDictionaries>
+
+            <!-- Extend the shared data grid row style to add the context menu and navigability data bindings -->
+            <Style BasedOn="{StaticResource BaseDataGridRow}"
+                       TargetType="DataGridRow">
+                <Setter Property="Background" Value="{DynamicResource {x:Static vsTheming:TreeViewColors.BackgroundBrushKey}}"/>
+                <Style.Triggers>
+                    <DataTrigger Binding="{Binding TaintIssueViz.Span, Converter={StaticResource SpanToNavigabilityConverter}}" Value="False">
```
duncanp-sonar, line 21: This is almost identical to derived style in the hotspot control. The only difference is the path to the property in the Binding. I'm not sure what the best way of removing this duplication is, or whether it's a priority at the moment.
If you can suggest a simple way to do it I'll do it now, otherwise I'll add a sub-task to the backlog.

src/IssueViz.Security/Taint/TaintList/TaintVulnerabilitiesControl.xaml
```diff
@@ -1,15 +1,124 @@
...
+                        <DataTemplate>
+                            <TextBlock Text="{Binding TaintIssueViz.Issue.RuleKey}" 
+                                       Style="{StaticResource CellTextBlockStyle}" />
+                        </DataTemplate>
+                    </DataGridTemplateColumn.CellTemplate>
+                </DataGridTemplateColumn>
+                <DataGridTemplateColumn CanUserSort="True" SortMemberPath="TaintIssueViz.Issue.Severity" Header="Severity" Width="68"  MinWidth="68">
+                    <DataGridTemplateColumn.CellTemplate>
+                        <DataTemplate>
+                            <TextBlock Text="{Binding TaintIssueViz.Issue.Severity}"
```
duncanp-sonar, line 84: I'm displaying the Sonar Severity directly for now. We can add other columns later if necessary. There's not much point displaying the Type since it will always be "Vulnerability". I think there is a security sub-category we could display, but it would require more work to fetch and process it.

The hotspot list has "Category" and "Priority", but neither has a direct equivalent in "normal" issues.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintToolWindowTests.cs
rita-g-sonarsource: I'm not sure that the Dispose code is needed.

I originally copied it from the secondary locations code (IssueVisualizationToolWindow), where it was needed due to a weird integration with ProvideToolWindowVisibility.
I thought it would be good for the 2 new windows as well, but it wasn't, and I later added `ProvideToolWindow(Transient = true)` which worked for them.
But I didn't remove this code and I think it was a mistake. `Transient = true` should be enough if we're not using ProvideToolWindowVisibility.
I'll open a ticket to investigate it and if it's not needed, we could remove these tests entirely.

src/IssueViz.Security/Taint/TaintList/TaintVulnerabilitiesControl.xaml
rita-g-sonarsource: Hm, a dependency property might do it, but I'm not sure if we'd want the exact same behavior. Let's wait and see what the final UX looks like.

src/IssueViz.Security/Taint/TaintList/TaintVulnerabilitiesControl.xaml
```diff
@@ -1,15 +1,124 @@
...
     <UserControl.Resources>
         <ResourceDictionary>
             <ResourceDictionary.MergedDictionaries>
                 <ResourceDictionary Source="../../SharedUI/SharedResources.xaml" />
             </ResourceDictionary.MergedDictionaries>
+
+            <!-- Extend the shared data grid row style to add the context menu and navigability data bindings -->
+            <Style BasedOn="{StaticResource BaseDataGridRow}"
+                       TargetType="DataGridRow">
+                <Setter Property="Background" Value="{DynamicResource {x:Static vsTheming:TreeViewColors.BackgroundBrushKey}}"/>
```
duncanp-sonar, line 19: FYI setting the background here is the same hack as in the hotspot control.

I haven't worked out why it's required, but I do know a bit more context: the `DataGridCell` has `<Setter Property="Padding" Value="6,2,2,2"/>`.
This means the data cell isn't filling the whole cell, which is exposing part of the control behind it, which is a border that is a child of the `DataGridRow`. For some reason, the border isn't picking up the background set for its parent unless it's set here.

src/IssueViz.Security/Taint/TaintList/TaintVulnerabilitiesControl.xaml
duncanp-sonar: Sounds good. A dependency property would required both view models to use the same property name, which seems a bit fragile given they are currently independent and invalid bindings aren`t detected until runtime.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintToolWindowTests.cs
duncanp-sonar: Ok. It would have be worth commenting in the Dispose method why it was there.

rita-g-sonarsource (approved):

Issue #2018, opened 2020-12-14
Start taint services when bound solution is open


rita-g-sonarsource:
#### Description

Start the taint services (TaintIssuesBindingMonitor, TaintIssuesSynchronizer) when a solution in connected mode is loaded.

TODO:
- [x] Investigate custom UI context that the binding service can set+ ~~a new~~ package that can AutoLoad with this.

PR #2020, opened 2020-12-14
Add VS UI context for connected mode indication

rita-g-sonarsource:
Part of #2018

### DIFF
--- a/src/Core/Binding/BoundSolutionVsUiContext.cs
+++ b/src/Core/Binding/BoundSolutionVsUiContext.cs
@@ -1,0 +1,31 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarLint.VisualStudio.Core.Binding
+{
+    public static class BoundSolutionVsUiContext
+    {
+        public const string GuidString = "C10779BB-5071-427D-9BE1-6F60EE064469";
+
+        public static Guid Guid = new Guid(GuidString);
+    }
+}
--- a/src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
+++ b/src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
@@ -25,6 +25,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Threading;
 using FluentAssertions;
+using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.ComponentModelHost;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
@@ -33,12 +34,15 @@ using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Integration.NewConnectedMode;
 using SonarQube.Client.Models;
 using SonarQube.Client;
+using IVsMonitorSelection = Microsoft.VisualStudio.Shell.Interop.IVsMonitorSelection;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests
 {
     [TestClass]
     public class ActiveSolutionBoundTrackerTests
     {
+        private uint boundSolutionUiContextCookie = 999;
+
         private readonly Expression<Func<ISonarQubeService, Task>> connectMethod = x => x.ConnectAsync(It.IsAny<ConnectionInformation>(), It.IsAny<CancellationToken>());
         private readonly Expression<Action<ISonarQubeService>> disconnectMethod = x => x.Disconnect();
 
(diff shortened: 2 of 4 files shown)


src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -56,27 +60,34 @@ public class ActiveSolutionBoundTrackerTests
         [TestInitialize]
         public void TestInitialize()
         {
-            this.serviceProvider = new ConfigurableServiceProvider(false);
-            this.host = new ConfigurableHost(this.serviceProvider, Dispatcher.CurrentDispatcher);
-            var mefExport1 = MefTestHelpers.CreateExport<IHost>(this.host);
+            serviceProvider = new ConfigurableServiceProvider(false);
```
rita-g-sonarsource, line 63: Cosmetics: removed redundant "this."

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -56,27 +60,34 @@ public class ActiveSolutionBoundTrackerTests
...
-            this.serviceProvider.RegisterService(typeof(IErrorListInfoBarController), this.errorListController);
+            configProvider = new ConfigurableConfigurationProvider();
+            serviceProvider.RegisterService(typeof(IConfigurationProviderService), configProvider);
 
-            this.configProvider = new ConfigurableConfigurationProvider();
-            this.serviceProvider.RegisterService(typeof(IConfigurationProviderService), this.configProvider);
+            loggerMock = new Mock<ILogger>();
 
-            this.loggerMock = new Mock<ILogger>();
+            vsMonitorMock = new Mock<IVsMonitorSelection>();
```
rita-g-sonarsource, line 85: new code

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -48,36 +51,26 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
...
-            }
-            if (activeSolutionTracker == null)
-            {
-                throw new ArgumentNullException(nameof(activeSolutionTracker));
-            }
-            if (logger == null)
-            {
-                throw new ArgumentNullException(nameof(logger));
-            }
+            extensionHost = host ?? throw new ArgumentNullException(nameof(host));
```
rita-g-sonarsource, line 54: Cosmetics: removed redundant "this." + combined initializations

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -48,36 +51,26 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
...
+            solutionTracker = activeSolutionTracker ?? throw new ArgumentNullException(nameof(activeSolutionTracker));
+            this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
-            this.extensionHost = host;
-            this.solutionTracker = activeSolutionTracker;
-            this.logger = logger;
+            vsMonitorSelection = host.GetService<SVsShellMonitorSelection, IVsMonitorSelection>();
```
rita-g-sonarsource, line 58: new code

src/Core/Binding/BoundSolutionVsUiContext.cs
```diff
@@ -0,0 +1,31 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarLint.VisualStudio.Core.Binding
+{
+    public static class BoundSolutionVsUiContext
```
duncanp-sonar, line 25: Casing: VS uses `UIContext` rather than `Ui...`.

I don't think it needs to have "Vs" in the name.

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -135,20 +128,27 @@ private void OnBindingStateChanged(object sender, BindingStateEventArgs e)
...
                 SolutionBindingUpdated?.Invoke(this, EventArgs.Empty);
             }
-            else
+
+            SetBoundSolutionUiContext();
+        }
+
+        private void SetBoundSolutionUiContext()
+        {
+            if (vsMonitorSelection.GetCmdUIContextCookie(ref BoundSolutionVsUiContext.Guid, out var cookie) == VSConstants.S_OK)
```
duncanp-sonar, line 148: The VS [documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsmonitorselection.getcmduicontextcookie?view=visualstudiosdk-2019) suggests that `GetCmdUIContextCookie` should only be called once per GUID, implying the caller should hold on to the cookie and re-use it.
The behaviour of calling the method multiple times doesn't seem to be documented.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -541,5 +559,13 @@ private void VerifyServiceDisconnect(Times expected)
         {
             sonarQubeServiceMock.Verify(disconnectMethod, expected);
         }
+
+        private void VerifyBoundSolutionUiContextIsSet(bool isActive)
+        {
+            var isActiveInt = isActive ? 1 : 0;
+            vsMonitorMock.Verify(x => x.SetCmdUIContext(boundSolutionUiContextCookie, isActiveInt), Times.Once);
+            vsMonitorMock.Verify(x => x.SetCmdUIContext(It.IsAny<uint>(), It.IsAny<int>()), Times.Once);
+            vsMonitorMock.Invocations.Clear();
```
duncanp-sonar, line 568: This is making a stateful change to the mock which will affect the behaviour of this method when it's called multiple times in the same test.
That's not a problem per se but it's not obvious and it's not how most of the test verification methods work so it deserves to be called out somehow, either by changing the method name or adding a comment (although I can't think of a good name that would do).

duncanp-sonar (commented):
See comments: I don't think we should be calling `GetCmdUIContextCookie` multiple times.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
rita-g-sonarsource: How's `VerifyAndResetBoundSolutionUiContextMock`? 🤔

PR #2022, opened 2020-12-14
Split IssueVizSecurityPackage to Hotspots and Taint packages

rita-g-sonarsource:


src/IssueViz.Security/IssueVizSecurityPackage.cs
```diff
@@ -35,44 +31,11 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security
     [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid("D7D54E08-45E1-49A6-AA53-AF1CFAA6EBDC")]
     [ProvideMenuResource("Menus.ctmenu", 1)]
```
rita-g-sonarsource, line 33: This file still exists just for the `ProvideMenuResource` attribute. Is there a way around this? Or do we need to have this?

PR #2026, opened 2020-12-14
Added input bindings for taint list navigation

duncanp-sonar:


src/IssueViz.Security/Taint/TaintList/TaintIssuesControl.xaml
```diff
@@ -37,6 +37,16 @@
...
+            <DataGrid.InputBindings>
+                <MouseBinding
+                    MouseAction="LeftDoubleClick"
+                    Command="{Binding NavigateCommand}"
+                    CommandParameter="{Binding Issues/}"/>
+                <KeyBinding
+                    Key="Enter"
+                    Command="{Binding NavigateCommand}"
+                    CommandParameter="{Binding Issues/}"/>
+            </DataGrid.InputBindings>
```
duncanp-sonar, line 49: Added the missing mouse and keyboard bindings to make taint navigation work - same as the hotspot control, except for the binding path.

I noticed one weird behaviour with both controls when testing:
1) select an item in the list
2) double-click on the background of the control i.e. below any of the list items.
The selected item is still navigated to, even though that isn't where I double-clicked.

I've logged a bug: https://github.com/SonarSource/sonarlint-visualstudio/issues/2027

rita-g-sonarsource (approved):

PR #2028, opened 2020-12-15
Start taint services when bound solution is open

### LINKED ISSUE #2018: Start taint services when bound solution is open

rita-g-sonarsource:
Fixes #2018

### DIFF
--- a/src/IssueViz.Security/IssueVizSecurityPackage.cs
+++ b/src/IssueViz.Security/IssueVizSecurityPackage.cs
@@ -23,9 +23,13 @@ using System.ComponentModel.Design;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.InteropServices;
 using System.Threading;
+using Microsoft.VisualStudio.ComponentModelHost;
 using Microsoft.VisualStudio.Shell;
+using SonarLint.VisualStudio.Core.Binding;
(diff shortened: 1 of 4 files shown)


src/IssueViz.Security/IssueVizSecurityPackage.cs
```diff
@@ -58,6 +69,11 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
             await ShowToolWindowCommand.CreateAsync(this,
                 new CommandID(Constants.CommandSetGuid, Constants.TaintToolWindowCommandId),
                 TaintToolWindow.ToolWindowId);
+
+            bindingMonitor = componentModel.GetService<ITaintIssuesBindingMonitor>();
```
duncanp-sonar, line 73: I think the doc comment on `ITaintIssuesBindingMonitor` should make it clear that it doesn't do an initial sync; it's not obvious, and it's important from an implementation point of view.

duncanp-sonar (approved):

Issue #2032, opened 2020-12-16
Filter Taint list to display issues only for current file


rita-g-sonarsource:
#### Description

Display in the taint list only the issues where the **sink** is in the opened file.

PR #2035, opened 2020-12-16
Add Infrastructure.VS assembly

rita-g-sonarsource:
Prep for #2032


SonarLint.VisualStudio.Integration.sln
```diff
@@ -86,7 +86,13 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "IssueViz.Security", "IssueV
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IssueViz.Security", "src\IssueViz.Security\IssueViz.Security.csproj", "{8D0D5AF0-D5B7-41ED-9EC6-9E10C66B37E0}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "IssueViz.Security.UnitTests", "src\IssueViz.Security.UnitTests\IssueViz.Security.UnitTests.csproj", "{F5BE0E99-066F-470D-BB4B-A136B1F26425}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IssueViz.Security.UnitTests", "src\IssueViz.Security.UnitTests\IssueViz.Security.UnitTests.csproj", "{35AB5B51-BC2A-44A6-BE46-EF41CDF1398D}"
```
rita-g-sonarsource, line 89: I don't understand why `IssueViz.Security.UnitTests` was referenced incorrectly and why it still showed up just fine in my solution -- this update is correct, `35AB5B51-BC2A-44A6-BE46-EF41CDF1398D` is indeed the guid of the project, there is no project with the guid `F5BE0E99-066F-470D-BB4B-A136B1F26425` like it was referenced before.

rita-g-sonarsource:
@duncanp-sonar MEF registration and manifests will be updated in the next PR.

SonarLint.VisualStudio.Integration.sln
duncanp-sonar: VS is fairly tolerant about missing/incorrect info in solution files. As long as it can resolve the project somehow it won't complain.

duncanp-sonar:
Just to confirm: is the new assembly included in the VSIX?

duncanp-sonar (approved):

rita-g-sonarsource:
> Just to confirm: is the new assembly included in the VSIX?

Yep it's there. I also verified in the next PR that everything is pulled properly.

PR #2039, opened 2020-12-17
Move ActiveDocumentLocator to Infrastructure.VS

rita-g-sonarsource:
Prep for #2032 

No product changes.
Affected & tested areas: CFamily reproducer command.

### DIFF
--- a/src/Infrastructure.VS/ActiveDocumentLocator.cs
+++ b/src/Infrastructure.VS/ActiveDocumentLocator.cs
@@ -24,11 +24,10 @@ using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Text;
-using SonarLint.VisualStudio.Infrastructure.VS;
 
-namespace SonarLint.VisualStudio.Integration.Vsix
+namespace SonarLint.VisualStudio.Infrastructure.VS
 {
-    internal interface IActiveDocumentLocator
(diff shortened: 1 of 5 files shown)


src/Infrastructure.VS/TextDocumentProvider.cs
```diff
@@ -26,7 +26,7 @@
 
 namespace SonarLint.VisualStudio.Infrastructure.VS
 {
-    public interface ITextDocumentProvider
+    internal interface ITextDocumentProvider
```
rita-g-sonarsource, line 29: This is used only by the 2 ActiveDocument services and now that they're in the same assembly, there is no need for this to be public anymore.

src/Infrastructure.VS/ActiveDocumentLocator.cs
```diff
@@ -24,11 +24,10 @@
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Text;
-using SonarLint.VisualStudio.Infrastructure.VS;
 
-namespace SonarLint.VisualStudio.Integration.Vsix
+namespace SonarLint.VisualStudio.Infrastructure.VS
 {
-    internal interface IActiveDocumentLocator
+    public interface IActiveDocumentLocator
```
rita-g-sonarsource, line 30: I realized I need this too, for the initial ViewModel filter initialization.

duncanp-sonar (approved):

PR #2040, opened 2020-12-17
Add ActiveDocument filter to Taint list

### LINKED ISSUE #2032: Filter Taint list to display issues only for current file

rita-g-sonarsource:
Fixes #2032


src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -282,7 +282,8 @@ Referenced assemblies:
 - 'System.ComponentModel.Composition, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
 - 'System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
 - 'System.Xaml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
-# Number of references: 22
+- 'WindowsBase, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
```
rita-g-sonarsource, line 285: I don't understand where this is suddenly coming from, and why I don't see this on the main branch-taint.
It was always there...
https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/IssueViz.Security/IssueViz.Security.csproj#L41

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -48,24 +50,55 @@ internal interface ITaintIssuesControlViewModel : IDisposable
...
+        }
+
+        private bool ActiveDocumentFilter(object viewModel)
+        {
+            if (string.IsNullOrEmpty(activeDocumentFilePath))
+            {
+                return false;
+            }
+
+            return ((ITaintIssueViewModel)viewModel).DisplayPath.Equals(activeDocumentFilePath);
```
duncanp-sonar, line 101: File paths should be compared using `OrdinalIgnoreCase`.

Using `DisplayPath` feels like a bit hacky. It's called `DisplayPath` because there are no guarantees about what it contains, other than it's a user-friendly string. Relying on it being ,an absolute path if the relative path from the server has been resolved successfully is assuming more than the view model is promising.

At the very least this assumption should be documented.

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
duncanp-sonar: Maybe it wasn't actually needed before so VS wasn't including the reference in the assembly?

Have you checked whether the assembly actually contains a reference to WindowsBase (using e.g. `ildasm`)?

I have a similar problem with my latest changes: I'm seeing `- 'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'` appearing as a reference for `SonarLint.VisualStudio.Infrastructure.VS`, and I haven't made any changes to that assembly and it doesn't depend on anything I have changed. I haven't investigated using `ildasm` yet.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -151,17 +226,80 @@ public void NavigateCommand_Execute_LocationNavigated()
...
+            var storeCollection = new ObservableCollection<IAnalysisIssueVisualization> { issueViz1, issueViz2 };
+
+            var activeDocumentTracker = new Mock<IActiveDocumentTracker>();
+
+            var testSubject = CreateTestSubject(storeCollection, activeDocumentTracker: activeDocumentTracker.Object);
+
+            var activeDocument = new Mock<ITextDocument>();
+            activeDocument.Setup(x => x.FilePath).Returns("test2.cpp");
+
+            activeDocumentTracker.Raise(x => x.OnDocumentFocused += null, new DocumentFocusedEventArgs(activeDocument.Object));
```
duncanp-sonar, line 263: Minor: the tests that handle the raised a doc focused event would be more readable if there was a helper method that created the `ActiveDocument` and raised the event e.g.
`RaiseDocFocusedEvent(Mock<IActiveDocumentTracker> mock, string activeDocName)`
... where `activeDocName` could be null.

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
duncanp-sonar: Update: my build problem turned out to be due a difference between Debug and Release builds (building locally means the Infra assembly includes Debug.Assert calls, building on the server means it doesn't, and presumably the code doesn't reference anything else from that assembly).

I wouldn't expect it to make a difference here, but you might want to try it.

rita-g-sonarsource: Seems like it indeed wasn't needed before - it's used now for `CollectionViewSource.GetDefaultView`.  
I don't know why it was there before then (in the csproj) -- I guess it was not needed anymore at some point but the csproj was not updated.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
rita-g-sonarsource: Good point, I changed it use `CurrentFilePath` from the underlying issue viz.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
rita-g-sonarsource: It's not the document name that is null, it's the document itself.
So `activeDocName` would have a double meaning - that if it's null, it means the document itself is null (and not just the name).
It'd have to be something like:
```
void RaiseDocFocusedEvent(Mock<IActiveDocumentTracker> mock, string activeDocName) 
{
    ITextDocument document = null;
    if (activeDocName != null) 
     {
            var documentMock = new Mock<ITextDocument>();
            documentMock.Setup(x => x.FilePath).Returns(activeDocName);
            document = documentMock.Object;
     }
     ...
}
```
and that's a bit confusing IMO.

duncanp-sonar: Yes, I realise that, and that's exactly the sort of thing I'd want to hide to raise the level of abstraction at which the main test can be read.

Issue #2042, opened 2020-12-17
Display creation date file in taint list


duncanp-sonar:
### Description

The MMF says the list should be sorted by mostly recently created issue so the creation date needs to be available in the list, and the list needs to have a default sort set.

TODO:
* decide what to display: just date, or date time{)?
* consider also whether to display the last updated timestamp.

Issue #1976, opened 2020-12-07
Add command to navigate to vulnerability on the server


rita-g-sonarsource:
#### Description

Add a link to open the vulnerability in SQ.
Similar to #1974.

Depends on https://github.com/SonarSource/sonarqube-webclient-dotnet/issues/72

PR #2043, opened 2020-12-17
Updated submodule to get additional SonarQubeIssue fields

duncanp-sonar:
* Fixed up tests.
* No product changes

Prep for #1976 and #2042

### DIFF
--- a/src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
+++ b/src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
@@ -211,9 +211,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
         public void GetSuppressedIssues_WhenProjectHasModulesAndIssueIsModuleLevelFound_ReturnsExpectedIssue()
         {
             // Arrange
-            var sonarQubeIssue1 = new SonarQubeIssue(null, null, "message", "sqkey:sqkey:projectId2", "S1", true, SonarQubeIssueSeverity.Blocker, null, flows: null);
-            var sonarQubeIssue2 = new SonarQubeIssue(null, null, "message", "sqkey:sqkey:projectId", "S2", true, SonarQubeIssueSeverity.Blocker, null, flows: null);
-            var sonarQubeIssue3 = new SonarQubeIssue("foo\\bar.cs", "hash", "message", "sqkey:sqkey:projectId", "S3", true, SonarQubeIssueSeverity.Blocker, new IssueTextRange(12,1,1,1), flows: null);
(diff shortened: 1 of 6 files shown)


src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -211,9 +211,9 @@ public void GetIssues_NoIssuesOnServer_ReturnsEmptyList()
         public void GetSuppressedIssues_WhenProjectHasModulesAndIssueIsModuleLevelFound_ReturnsExpectedIssue()
         {
             // Arrange
-            var sonarQubeIssue1 = new SonarQubeIssue(null, null, "message", "sqkey:sqkey:projectId2", "S1", true, SonarQubeIssueSeverity.Blocker, null, flows: null);
```
duncanp-sonar, line 214: There are only three important fields used by these tests, so I created a helper function that also handles setting the new fields appropriately.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -281,36 +281,27 @@ public void GetSuppressedIssues_WhenProjectHasModulesAndIssueIsFileLevelAndIsFou
...
-                WaitForInitialFetchTaskToStart();
+            var issuesProvider = new SonarQubeIssuesProvider(mockSqService.Object, "sqkey", mockTimerFactory.Object, testLogger);
+            WaitForInitialFetchTaskToStart();
 
-                VerifyServiceGetIssues(Times.Exactly(1)); // issues should be fetched on creation
+            VerifyServiceGetIssues(Times.Exactly(1)); // issues should be fetched on creation
 
-                // Act
-                IEnumerable<SonarQubeIssue> matches;
-                // We're deliberately faking SonarQube returning paths with \ instead of / which
```
duncanp-sonar, line 305: This comment was no longer accurate, and the AssertIgnoreScope is no longer required.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -334,31 +325,28 @@ public void GetSuppressedIssues_WhenProjectHasModulesAndIssueIsFileLevelAndIsNot
 
             VerifyServiceGetIssues(Times.Exactly(1)); // issues should be fetched on creation
 
-            // We're deliberately faking SonarQube returning paths with \ instead of / which
```
duncanp-sonar, line 337: This comment is no longer accurate either so the AssertIgnoreScope isn't required for all of the test cases.

However, it is still required for the input parameter in the second case, so I moved the AssertIgnoreScope to just cover that case.

rita-g-sonarsource (approved):

PR #2044, opened 2020-12-18
Don't try to fetch taint issues when SQ server is not connected

### LINKED ISSUE #2034: Fetching taint vulnerabilities sometimes throws an exception
#### Description

There is a race condition between the initialization of SonarQubeService and the call to GetTaintVulnerabilitiesAsync, which results in an exception being shown in the output window: 
```
ERROR: The service is expected to be connected.
Failed to synchronize taint vulnerabilities with the connected server: System.InvalidOperationException: This operation expects the service to be connected.
   at SonarQube.Client.SonarQubeService.EnsureIsConnected()
   at SonarQube.Client.SonarQubeService.<InvokeRequestAsync>d__18`2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at SonarQube.Client.SonarQubeService.<GetTaintVulnerabilitiesAsync>d__36.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintIssuesSynchronizer.<SynchronizeWithServer>d__6.MoveNext().
```

Repro steps:
1. Open a VS solution
2. Open the SQ team explorer pane
3. Close the solution
4. Re-open the solution

Expected ==> taint vulnerabilities are fetched
Actual ==> exception is written to the output window

Notes: 
* This doesn't seem to happen when the SQ team explorer pane is closed, but the race condition is probably still there.
* The taint vulnerabilities are still fetched, because there are 2 binding events: 1) BindingChanged (which is raised when the binding changes to Connected or Standalone), and 2) BindingUpdated. When the team explorer pane is open, BindingChanged fails but then next BindingUpdated succeeds.

rita-g-sonarsource:
Fixes #2034

### DIFF
--- a/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
+++ b/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
@@ -73,13 +73,15 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
                 ClearStore();
                 return;
             }
+
+            if (!sonarQubeService.IsConnected)
+            {
+                logger.WriteLine(TaintResources.Synchronizer_ServerNotConnected);
+                return;
+            }
+
             try
             {
-                if (!sonarQubeService.IsConnected)
-                {
-                    return;
-                }
-
                 var projectKey = bindingConfiguration.Project.ProjectKey;
                 var taintVulnerabilities = await sonarQubeService.GetTaintVulnerabilitiesAsync(projectKey, CancellationToken.None);
 
--- a/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
+++ b/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
@@ -75,6 +75,11 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
             }
             try
             {
+                if (!sonarQubeService.IsConnected)
+                {
+                    return;
+                }
+
                 var projectKey = bindingConfiguration.Project.ProjectKey;
                 var taintVulnerabilities = await sonarQubeService.GetTaintVulnerabilitiesAsync(projectKey, CancellationToken.None);
 
--- a/src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
@@ -78,6 +78,19 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
         }
 
         [TestMethod]
+        public async Task SynchronizeWithServer_SonarQubeServerNotYetConnected_NoChanges()
+        {
+            var converter = new Mock<ITaintIssueToIssueVisualizationConverter>();
+            var taintStore = new Mock<ITaintStore>();
+
+            var testSubject = CreateTestSubject(taintStore.Object, converter.Object, isServerConnected: false);
+            await testSubject.SynchronizeWithServer();
+
+            converter.Invocations.Count.Should().Be(0);
+            taintStore.Invocations.Count.Should().Be(0);
+        }
+
+        [TestMethod]
         public async Task SynchronizeWithServer_NoIssues_StoreSynced()
         {
             var converter = new Mock<ITaintIssueToIssueVisualizationConverter>();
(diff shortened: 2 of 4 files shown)


src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -75,6 +75,11 @@ public async Task SynchronizeWithServer()
             }
             try
             {
+                if (!sonarQubeService.IsConnected)
+                {
+                    return;
+                }
```
rita-g-sonarsource, line 81: The fetch attempt will only happen if the binding configuration is "Connected Mode". So if we got to this point, we should be in connected mode yet SQ client might not be initialized yet.

This code can be called as a response to 2 events:

1. [ActiveSolutionBoundTracker.SolutionBindingChanged](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration/MefServices/ActiveSolutionBoundTracker.cs#L143) event.
This event is raised in 2 use cases:
1.1 When the active solution is changed [IActiveSolutionTracker.ActiveSolutionChanged](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration/MefServices/ActiveSolutionBoundTracker.cs#L78). If the opened solution has connected mode configured, `SolutionBindingChanged` event will always be preceded with [awaiting a connection to SQ server](https://github.com/SonarSource/sonarlint-visualstudio/blob/4813fca6e9133af9b64c3c1171f878c5f6c55003/src/Integration/MefServices/ActiveSolutionBoundTracker.cs#L123).
So in this case, TaintIssuesSynchronizer will always succeed.

1.2 When team explorer SQ pane is first initialized/changed (https://github.com/SonarSource/sonarlint-visualstudio/blob/4813fca6e9133af9b64c3c1171f878c5f6c55003/src/Integration/MefServices/ActiveSolutionBoundTracker.cs#L75). 
It originates in StateManager.SetProjects, which is called by the team explorer pane's controller.
In this case, TaintIssuesSynchronizer will always fail, since the SQ server is not yet initialized at this point.

1.1 and 1.2 have an interaction between them: if the team explorer pane is already open when a bound solution is opened, 1.2 will always win and 1.1 will never be called due to [this IF](https://github.com/SonarSource/sonarlint-visualstudio/blob/4813fca6e9133af9b64c3c1171f878c5f6c55003/src/Integration/MefServices/ActiveSolutionBoundTracker.cs#L140).

2. [ActiveSolutionBoundTracker.SolutionBindingUpdated](https://github.com/SonarSource/sonarlint-visualstudio/blob/4813fca6e9133af9b64c3c1171f878c5f6c55003/src/Integration/MefServices/ActiveSolutionBoundTracker.cs#L147).
This is always called by the team explorer SQ pane's controller. If the pane is never opened, it will never be called.

So unless I missed something, it seems that the execution would depend on whether or not the user has opened the team explorer pane. 
If they did, we always fail on `SolutionBindingChanged` event and rely on `SolutionBindingUpdated` to be raised. 
If the pane has not been opened, then only `SolutionBindingChanged` will be called and it would always be preceded with awaiting the connection to the server.
So it seems that checking if the server is connected is enough to handle the two cases, but I must admit that I'm worried that I missed something.

I've also tried to check what happens if a user manually opens the tool window during the solution load (e.g. after opening the solution but before `SolutionBindingChanged` is raised). However that doesn't seem to be possible: `IActiveSolutionTracker.ActiveSolutionChanged` is raised on the UI thread, and it immediately [disconnects the SQ server](https://github.com/SonarSource/sonarlint-visualstudio/blob/4813fca6e9133af9b64c3c1171f878c5f6c55003/src/Integration/MefServices/ActiveSolutionBoundTracker.cs#L103). So I can't seem to open the taint tool window after the solution is opened and yet before this block of code is executed - either the tool window is opened before the solution is loaded (in which case we are in standalone mode), or it is opened after VS raises this event and SQ is already disconnected and `SolutionBindingChanged` will be fired later. 

I tried to reproduce various cases and this block of code seems to work but this entire thing feels wrong.
Why do we even need to wait for the SQ client to be "connected"? We have the credentials, and every call should be stateless and just try to fetch the data. If it fails then it fails. I don't see why we would need a global IsConnected flag that gives us so much trouble.

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -75,6 +75,11 @@ public async Task SynchronizeWithServer()
             }
             try
             {
+                if (!sonarQubeService.IsConnected)
+                {
+                    return;
```
duncanp-sonar, line 80: I`d log that it wasn`t possible to sync otherwise it will be hard to diagnose what`s happening. I think it would be ok to log it with non-debug verbosity.

Also, this check doesn`t need to be inside the `try` block.

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
duncanp-sonar: As discussed offline, I suggest opening a ticket about the sync order and copy your notes above into it.

duncanp-sonar (approved):

duncanp-sonar (approved):

PR #2045, opened 2020-12-18
Add TaintIssue class

rita-g-sonarsource:
Part of #1976

### DIFF
--- a/src/IssueViz.Security.UnitTests/Taint/TaintIssueToIssueVisualizationConverterTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintIssueToIssueVisualizationConverterTests.cs
@@ -29,6 +29,7 @@ using SonarLint.VisualStudio.Core.Analysis;
 using SonarLint.VisualStudio.Integration.UnitTests;
 using SonarLint.VisualStudio.IssueVisualization.Models;
 using SonarLint.VisualStudio.IssueVisualization.Security.Taint;
+using SonarLint.VisualStudio.IssueVisualization.Security.Taint.Models;
 using SonarQube.Client.Models;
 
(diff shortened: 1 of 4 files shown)


src/IssueViz.Security/Taint/Models/ITaintIssue.cs
```diff
@@ -0,0 +1,75 @@
...
+
+using System;
+using System.Collections.Generic;
+using SonarLint.VisualStudio.Core.Analysis;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint.Models
+{
+    internal interface ITaintIssue : IAnalysisIssueBase
+    {
+        string IssueKey { get; }
```
rita-g-sonarsource, line 29: I thought about adding `IssueKey` to AnalysisIssue but it didn't feel right; AnalysisIssue is used in cfamily and deamon issues, so it's more representative of "local issues" that by definition don't have an id.

src/IssueViz.Security.UnitTests/Taint/TaintIssueToIssueVisualizationConverterTests.cs
```diff
@@ -100,7 +101,8 @@ public void Convert_IssueVizConverterCalledWithCorrectParameters_ReturnsConverte
             result.Should().BeSameAs(expectedConvertedIssueViz);
 
             issueVizConverter.Verify(x => x.Convert(
-                    It.Is((AnalysisIssue issueBase) =>
+                    It.Is((TaintIssue issueBase) =>
```
duncanp-sonar, line 104: Minor; the variable name isn't accurate any more

duncanp-sonar (approved):

PR #2051, opened 2020-12-21
Sort taint list by creation date

duncanp-sonar:
Related to #2042

### DIFF
--- a/src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
+++ b/src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
@@ -22,6 +22,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
+using System.ComponentModel;
 using System.Linq;
 using System.Windows.Data;
 using System.Windows.Input;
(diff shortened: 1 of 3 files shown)


src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -114,6 +116,14 @@ private bool ActiveDocumentFilter(object viewModel)
             return PathHelper.IsMatchingPath(filePath, activeDocumentFilePath);
         }
 
+        private void SetDefaultSortOrder()
+        {
+            var collectionView = CollectionViewSource.GetDefaultView(Issues);
+
+            collectionView.SortDescriptions.Add(
+                new SortDescription("TaintIssueViz.Issue.CreationTimestamp", ListSortDirection.Descending));
```
rita-g-sonarsource, line 124: @duncanp-sonar I'm catching up on old PRs; there seems to be a way to do this via xaml:
https://stackoverflow.com/a/21045727/339720

Have you tried to do something like this? It's a bit neater than the string

duncanp-sonar: Just setting the `SortDirection` in the `DataGridTemplateColumn` doesn`t work (or at least it isn't easily unit testable).
Switching to use the CollectionViewSource directly in the XAML should work. However, I didn't want to have some of the CollectionViewSource properties being set in XAML and some (for the filtering) set in code, and I also didn't want to refactor the existing code as I was trying to keep the changes to a minimum, partly to finish the sprint and partly because I didn't want to make more extensive changes in the absence of a reviewer.

I think we should change both the filtering and sorting code in one of two ways: either (1) explicitly expose the CollectionViewSource from the view model and have the control bind directly to it, or (2) move all of the CollectionViewSource stuff to XAML, and expose a `Filter` method on the view model so we can bind to it from XAML.

Issue #1980, opened 2020-12-07
Refactor the integration between the error list, the lightbulb and the secondary locations selection service


rita-g-sonarsource:
#### Description

SonarErrorListDataSource used to be the only service that could set the selection on AnalysisIssueSelectionService.
This meant that when SonarErrorListDataSource changed (i.e. a new analysis was triggered, file was closed, etc) - AnalysisIssueSelectionService was syncronized and could never contain data that wasn't in SonarErrorListDataSource.

However, with the introduction of the lightbulb, it meant that there was another source that could update AnalysisIssueSelectionService. And since the lightbulb could potentially be out of sync with SonarErrorListDataSource (#1803), it meant that AnalysisIssueSelectionService could contain data that is not available in SonarErrorListDataSource and cause selection side effects (see #1746).

A hacky fix was done in PR #1747 however this deserves a proper fix: a single source-of-truth of the currently available issues, that would be updated when a new analysis is done, issue is removed (via code changes), file is closed/renamed, etc.

Note: a similar problem will arise with hotspots/taint vulnerabilities with secondary locations (#1978, #1904).

PR #2061, opened 2021-01-05
Clear secondary locations panel when it contains an obsolete issue viz

rita-g-sonarsource:
Part of #1980


src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -178,16 +155,16 @@ public void RemoveFactory(IIssuesSnapshotFactory factory)
...
                 foreach (var sink in sinks)
                 {
                     SafeOperation(sink, "RemoveFactory", () => sink.RemoveFactory(factory));
                 }
 
-                if (factory.CurrentSnapshot.Issues.Contains(selectionService.SelectedIssue))
+                if (wasRemoved)
                 {
-                    selectionService.SelectedIssue = null;
+                    NotifyIssuesChanged(factory);
```
rita-g-sonarsource, line 167: This is a bit different behavior: RemoveFactory is called when a file is closed, and we will now notify the taggers that the closed file has changed (as they listen to IssuesChanged). In practice this won't make a difference since the taggers filter by current file path, so they won't do anything. 

https://github.com/SonarSource/sonarlint-visualstudio/blob/e5f23e84787d2bf93dac139c700750b23ffeec25/src/IssueViz/Editor/LocationTagging/LocationTagger.cs#L65

src/IssueViz/IssueVizPackage.cs
```diff
@@ -45,6 +47,9 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
 
             await IssueVisualizationToolWindowCommand.InitializeAsync(this);
             await NavigationCommands.InitializeAsync(this);
+
+            var componentModel = await GetServiceAsync(typeof(SComponentModel)) as IComponentModel;
+            componentModel.GetService<ISelectedVisualizationValidityMonitor>();
```
rita-g-sonarsource, line 52: The monitor will only start when the package was loaded, so until the panel is opened, the service won't run. At the moment this is ok, since the problem is in the panel's state to begin with, but there might be some future problems since technically the SelectionService can contain old issues this way. But since we're only using it for the panel at the moment, it should be ok.

rita-g-sonarsource:
@duncanp-sonar As we discussed offline, I've created a new class, `SelectedVisualizationValidityMonitor`, that listens to IssuesChanged event. When the event is raised, the class queries all the location stores (using the aggregator) and calls their new `Contains` method to know if the issue still exists. If it doesn't, it clears the selection. I went over the original use cases (https://github.com/SonarSource/sonarlint-visualstudio/issues/1746) and it all works.

It might help to look at the commits separately:
1. First commit adds the new Contains method to all the stores.
2. Second commit adds the new SelectedVisualizationValidityMonitor service that uses the Contains method above.
3. Third commit replaces the old code with the new SelectedVisualizationValidityMonitor.

src/IssueViz/Selection/SelectedVisualizationValidityMonitor.cs
```diff
@@ -0,0 +1,74 @@
...
+
+using System;
+using System.ComponentModel.Composition;
+using System.Linq;
+using SonarLint.VisualStudio.Core.Helpers;
+using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Selection
+{
+    internal interface ISelectedVisualizationValidityMonitor : IDisposable
```
duncanp-sonar, line 29: The interface/class needs a doc comment explaining what it's responsible for. It's not obvious without reading all of the code.

src/IssueViz/Selection/SelectedVisualizationValidityMonitor.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        private void LocationStoreAggregator_IssuesChanged(object sender, IssuesChangedEventArgs e)
+        {
+            if (selectionService.SelectedIssue == null)
+            {
+                return;
+            }
+
+            var issuesChangedInSelectedFile = e.AnalyzedFiles.Any(x => PathHelper.IsMatchingPath(x, selectionService.SelectedIssue.CurrentFilePath));
```
duncanp-sonar, line 56: `AffectedFiles` might be a better name than `AnalyzedFiles` now - `AnalyzedFiles` suggests that the changed was caused by an analysis which isn't necessarily the case.

duncanp-sonar:
For me the main value in this change is removing the hacky code from `SonarErrorListDataSource`.

However, I think the selection code is still too complicated: we have `IAnalysisIssueSelectionService`, 
`IHotspotsSelectionService`, `IIssueTablesSelectionMonitor`, we don't yet handle taint selection, and now we have `SelectedVisualizationValidityMonitor`. Still, this is a step in the right direction.

src/IssueViz/Selection/SelectedVisualizationValidityMonitor.cs
duncanp-sonar: I was trying to think of a comment that describes what this class is for, and in fact it seems it just exists to patch an incorrect selection status that could be created by the interaction of all of the other classes. That makes it seem like a design smell in its current form (hence my other comment about selection overall needing more work). Hopefully this class will disappear at some point if we can simplify the rest of the selection logic.

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
```diff
@@ -327,49 +263,68 @@ public void RemoveFactory_ExistingSinkManagersAreNotified()
         }
 
         [TestMethod]
-        public void RemoveFactory_SelectedIssueIsInFactory_IssueDeselected()
+        public void Contains_NullArg_Throws()
```
duncanp-sonar, line 266: Should the `Contains` tests be in SonarErrorListDataSource_IssueLocationStoreTests.cs` instead?

duncanp-sonar (approved):

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
rita-g-sonarsource: Good point, I'll move them

Issue #2037, opened 2020-12-17
Taint list is not cleared when solution binding changes


rita-g-sonarsource:
#### Description

When the solution is closed/changed to standalone, the taint list is not cleared.

TODO:
- [ ] Clear the list when the binding changes.

duncanp-sonar:
The store is cleared, but the taint list isn't. The screen shot below shows the effect of opening and closing a bound solution three times. with an open file that has one issue:

![image](https://user-images.githubusercontent.com/10757559/102910710-e2628500-4472-11eb-9140-23baa7bf047b.png)

The two obsolete entries are shown as non-navigable while the latest entry is navigable.

Issue #1958, opened 2020-12-01
Exception when opening a hotspot when there are navigable and non-navigable hotspots 


rita-g-sonarsource:
Repro scenario:
1. Open a hotspot for a file that doesn't exist, e.g. https://next.sonarqube.com/sonarqube/security_hotspots?id=sonarlint-visualstudio&hotspots=AWcuhGPcUzMSG3VG2V6j
1. Open a hotspot for a file that does exist https://next.sonarqube.com/sonarqube/security_hotspots?id=sonarlint-visualstudio&hotspots=AWcuhGNIUzMSG3VG2V6g

Expected ==> both hotspots are added to the list, one navigable and one not.
Actual ==> Exception is thrown

Notes:
`IIssueLocationStore.GetLocations` returns both hotspots, including the one where hotspot.CurrentFilePath is null.

```
   at System.IO.Path.GetFullPathInternal(String path)
   at System.IO.Path.GetFullPath(String path)
   at SonarLint.VisualStudio.Core.Helpers.PathHelper.IsMatchingPath(String path1, String path2) in C:\Projects\sonarlint-visualstudio\src\Core\Helpers\PathHelper.cs:line 145
   at SonarLint.VisualStudio.IssueVisualization.Security.Store.HotspotsStore.<>c__DisplayClass11_0.<SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging.IIssueLocationStore.GetLocations>b__1(IAnalysisIssueLocationVisualization locationViz) in C:\Projects\sonarlint-visualstudio\src\IssueViz.Security\Store\HotspotsStore.cs:line 107
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.Enumerable.<SelectManyIterator>d__17`2.MoveNext()
   at SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging.LocationTagger.EnsureUpdatedSpansExist(IEnumerable`1 locVizs, ITextSnapshot currentSnapshot) in C:\Projects\sonarlint-visualstudio\src\IssueViz\Editor\LocationTagging\LocationTagger.cs:line 108
   at SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging.LocationTagger.CreateTagSpans(ITextSnapshot textSnapshot) in C:\Projects\sonarlint-visualstudio\src\IssueViz\Editor\LocationTagging\LocationTagger.cs:line 88
[...]

PR #2065, opened 2021-01-06
Replace Clear() with extension method that removes all and raises NotifyCollectionChanged

### LINKED ISSUE #2064: Taint tags are not cleared when solution binding changes
#### Description

Taint tags are not removed from the editor when the binding changes.

Steps to reproduce:
1. Open a solution in connected mode and navigate to a file with taint vulnerability
2. Re-sync the binding and make the issue disappear

Expected ==> tags are removed
Actual ==> tags are still there

Note: similar to #2037

rita-g-sonarsource:
Fixes #2064 #2037

### DIFF
--- a/src/Core/Helpers/ObservableCollectionExtensions.cs
+++ b/src/Core/Helpers/ObservableCollectionExtensions.cs
@@ -1,0 +1,41 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.ObjectModel;
+using System.Linq;
+
+namespace SonarLint.VisualStudio.Core.Helpers
+{
+    public static class ObservableCollectionExtensions
+    {
+        /// <summary>
+        /// Removes all items from the collection and raises NotifyCollectionChanged for every removed item
+        /// </summary>
+        public static void RemoveAll<T>(this ObservableCollection<T> collection)
+        {
+            var allItems = collection.ToList();
+
+            foreach (var item in allItems)
+            {
+                collection.Remove(item);
+            }
+        }
+    }
+}
--- a/src/IssueViz.Security/Taint/TaintStore.cs
+++ b/src/IssueViz.Security/Taint/TaintStore.cs
@@ -22,6 +22,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core.Helpers;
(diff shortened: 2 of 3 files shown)


src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -62,7 +63,7 @@ public void Set(IEnumerable<IAnalysisIssueVisualization> issueVisualizations)
                 throw new ArgumentNullException(nameof(issueVisualizations));
             }
 
-            taintVulnerabilities.Clear();
+            taintVulnerabilities.RemoveAll();
```
rita-g-sonarsource, line 66: There is a problem with the Clear method:
It raises NotifyCollectionChanged with `e.OldItems = null`, and `e.Action = NotifyCollectionChangedAction.Reset`. 
It would be fine for the view model (as we discussed, it might not need the observable to begin with), but `IssueStoreObserver` relies on it too: https://github.com/SonarSource/sonarlint-visualstudio/blob/branch-taint/src/IssueViz.Security/IssueStoreObserver.cs#L116

If we don't supply the removed items to `IssueStoreObserver`, the tags don't get removed (see #2064)

src/Core/Helpers/ObservableCollectionExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace SonarLint.VisualStudio.Core.Helpers
+{
+    public static class ObservableCollectionExtensions
+    {
+        /// <summary>
+        /// Removes all items from the collection and raises NotifyCollectionChanged for every removed item
+        /// </summary>
+        public static void RemoveAll<T>(this ObservableCollection<T> collection)
+        {
+            var allItems = collection.ToList();
```
rita-g-sonarsource, line 33: I remember that you mentioned something about deferring the updates to introduce a bulk-change notification. I couldn't find it, except writing our own ObservableCollection that implements it. Is that what you meant?

duncanp-sonar: No, it`s on the collection view: `CollectionView.DeferRefresh()`

src/IssueViz.Security/Taint/TaintStore.cs
duncanp-sonar: Removing one item at a time and triggering a chain of further update events feels wrong.

Taking a step back, the data structures and event forwarding just feel too complicated. When reviewing your other PR earlier, I was trying to remember what the point of the `IssueStoreObserver` was. It looks like an adapter that implements `IIssueLocationStore` over an `ObservableCollection`.

If an `ObservableCollection` isn`t the right data structure for the `TaintStore`, we shouldn`t need to care about `IssueStoreObserver`. Wouldn`t it be simpler to just to pick the right data structure, implement `IIssueLocationStore` directly on `TaintStore` and forget about `IssueStoreObserver`?

If I remember correctly, `IssueStoreObserver` was introduced to avoid duplicating code between the hotspots store and the taint store. However, the code is starting to feel over-complicated.

The update scenarios of the hotspot and taint stores are different; hotspots can be added incrementally as the user chooses to "open in IDE", whereas the taint list is only bulk-updated. However, we can't do bulk notifications with the current implementation; we have to raise one notification for each item added as part of a bulk operation, and now one notification for each item removed.

I think should have an offline discussion about this tomorrow to discuss whether the current implementation is still the right one, and if not when/how to change.

rita-g-sonarsource: I think `IssueStoreObserver` is an important component: it has a sizable chunk of code and that code already had bugs in it (#1958). It's identical for hotspots and taint and I don't think it's right to duplicate it. Also, we discussed that we could have a list of all the hotspots from the server (just like taint), in which case the hotspots could also be bulk-updated.

Perhaps we need to look into writing our own `ObservableCollection`, one that properly implements Clear() and bulk-notifications. Maybe there's even a package for it already.

rita-g-sonarsource: Hey look at [that](https://github.com/dotnet/runtime/issues/18087#issuecomment-359197102), we might get it if we're lucky 😆

rita-g-sonarsource: FYI: https://github.com/dotnet/runtime/issues/18087#issuecomment-359197102

PR #2068, opened 2021-01-06
Dp/taint UI context

### LINKED ISSUE #2059: Show/hide the Taint Issues tool window automatically
### Description

Currently the taint issues tool window does not appear unless the user specifically asks for it to be created, making the feature very difficult to discover. We want the feature to be more discoverable without it being too intrusive.

#### Proposed UX
* the taint tool window will appear automatically when the user opens a bound solution that has  at least one taint issue, and disappear when the bound solution is closed. The window will not automatically be brought to the front or have focus.
* The tool window caption will update automatically to indicate whether there are any taint issues in the current file (see #2058).

Notes:
* the taint issues window will not appear for projects that are bound to the Community edition of SonarQube.
* if the user manually closes the tool window it will no longer appear/disappear automatically (i.e. normal VS behaviour for this type of tool window).

#### TODO:
- [x] set a new UIContext when a bound solution is opened/clear it when the solution closes
- [x] set the visibility of the tool window to be conditional on the new UIContext being active

Follows on from #2055.

duncanp-sonar:
Fixes #2059

### DIFF
--- a/src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
@@ -199,7 +199,8 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
             IConfigurationProvider configurationProvider,
             ILogger logger)
         {
-            return new TaintIssuesSynchronizer(taintStore, sonarQubeServer, converter, configurationProvider, logger);
+            return new TaintIssuesSynchronizer(taintStore, sonarQubeServer, converter, configurationProvider,
+                Mock.Of<IServiceProvider>(), logger);
         }
 
         private class TestSonarQubeIssue : SonarQubeIssue
--- a/src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
@@ -22,8 +22,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
-using System.Threading.Tasks;
 using FluentAssertions;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Core.Binding;
@@ -33,12 +34,15 @@ using SonarLint.VisualStudio.IssueVisualization.Models;
 using SonarLint.VisualStudio.IssueVisualization.Security.Taint;
 using SonarQube.Client;
 using SonarQube.Client.Models;
+using Task = System.Threading.Tasks.Task;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
 {
     [TestClass]
     public class TaintIssuesSynchronizerTests
     {
+        private const string SharedProjectKey = "test";
+
         [TestMethod]
         public void MefCtor_CheckIsExported()
         {
@@ -48,11 +52,12 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
                 MefTestHelpers.CreateExport<ISonarQubeService>(Mock.Of<ISonarQubeService>()),
                 MefTestHelpers.CreateExport<ITaintIssueToIssueVisualizationConverter>(Mock.Of<ITaintIssueToIssueVisualizationConverter>()),
                 MefTestHelpers.CreateExport<IConfigurationProvider>(Mock.Of<IConfigurationProvider>()),
+                MefTestHelpers.CreateExport<SVsServiceProvider>(Mock.Of<IServiceProvider>()),
                 MefTestHelpers.CreateExport<ILogger>(Mock.Of<ILogger>())
             });
         }
 
-        [TestMethod]    
+        [TestMethod]
         public async Task SynchronizeWithServer_NotInConnectedMode_StoreCleared()
         {
             var sonarQubeServer = new Mock<ISonarQubeService>();
@@ -62,11 +67,12 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
 
             configurationProvider.Setup(x => x.GetConfiguration()).Returns(BindingConfiguration.Standalone);
 
-            var testSubject = CreateTestSubject(
+            var testSubject = new TaintIssuesSynchronizer(
                 taintStore.Object,
                 sonarQubeServer.Object,
                 converter.Object,
-                configurationProvider.Object, 
+                configurationProvider.Object,
+                Mock.Of<IServiceProvider>(),
                 Mock.Of<ILogger>());
 
             await testSubject.SynchronizeWithServer();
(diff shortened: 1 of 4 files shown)


src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -49,18 +51,26 @@ internal sealed class TaintIssuesSynchronizer : ITaintIssuesSynchronizer
         private readonly IConfigurationProvider configurationProvider;
         private readonly ILogger logger;
 
+        private readonly VSShellInterop.IVsMonitorSelection vsMonitorSelection;
```
duncanp-sonar, line 54: The product change is relatively simple; the taint synchronizer sets/clears the UIContext when the binding changes. The context is only set if taint issues exist (so it won`t ever be set for SonarQube Community edition).

src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
```diff
@@ -62,11 +67,12 @@ public async Task SynchronizeWithServer_NotInConnectedMode_StoreCleared()
 
             configurationProvider.Setup(x => x.GetConfiguration()).Returns(BindingConfiguration.Standalone);
 
-            var testSubject = CreateTestSubject(
```
duncanp-sonar, line 65: Miscellaneous change: there were previously two `CreateTestSubject` overloads, which I've modified to support the new tests. That made it simpler to change this existing test just to call the constructor directly.

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -101,5 +114,10 @@ private void ClearStore()
         {
             taintStore.Set(Enumerable.Empty<IAnalysisIssueVisualization>());
         }
+
+        private void SetHasTaintIssuesUIContext(bool hasTaintIssues)
```
rita-g-sonarsource, line 118: I kept reading it as 'SetHasTaintIssues', as in, a collection (set) has taint issues... Maybe `UpdateTaintUiContext` ?

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -101,5 +114,10 @@ private void ClearStore()
...
+        {
+            vsMonitorSelection?.SetCmdUIContext(contextCookie, hasTaintIssues ? 1 : 0);
```
rita-g-sonarsource, line 120: why the `?`? in other places we just assume that we managed to get the service. Can it ever really be null?
https://github.com/SonarSource/sonarlint-visualstudio/blob/e5f23e84787d2bf93dac139c700750b23ffeec25/src/IssueViz/Selection/AnalysisIssueSelectionService.cs#L47

duncanp-sonar: It shouldn`t be (and if it is then VS has bigger problems so it won't matter!).

I think this is left over from the prototype as a convenient way to make the existing tests pass without having to supply a valid configured service provider. I'll remove it.

src/IssueViz.Security/IssueVizSecurityPackage.cs
```diff
@@ -41,7 +42,9 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security
     [Guid("D7D54E08-45E1-49A6-AA53-AF1CFAA6EBDC")]
     [ProvideMenuResource("Menus.ctmenu", 1)]
     [ProvideToolWindow(typeof(HotspotsToolWindow), MultiInstances = false, Transient = true, Style = VsDockStyle.Tabbed, Window = VsWindowKindErrorList, Width = 700, Height = 250)]
-    [ProvideToolWindow(typeof(TaintToolWindow), MultiInstances = false, Transient = true, Style = VsDockStyle.Tabbed, Window = VsWindowKindErrorList, Width = 700, Height = 250)]
+    [ProvideToolWindow(typeof(TaintToolWindow), MultiInstances = false, Transient = false, // Note: Transient must false when using ProvideToolWindowVisibility
+        Style = VsDockStyle.Tabbed, Window = VsWindowKindErrorList, Width = 700, Height = 250)]
+    [ProvideToolWindowVisibility(typeof(TaintToolWindow), TaintIssuesExistUIContext.GuidString)]
```
rita-g-sonarsource, line 47: When we made this change in secondary locations panel, it introduced a problem with the persistence that we had to fix using the frame close:
https://github.com/SonarSource/sonarlint-visualstudio/commit/e0e78e54041c67ff3260573922f39d06d4947c28#diff-9f42bc19311338b66ca2e7dd10a3d6fd54f56ae33f85544277fd064861e59e04

Is that needed?
What happens if you force-open the tool window and close VS? will it disappear next time VS launches? (it didn't for secondary locations so we had to force close the frame)

src/IssueViz.Security/IssueVizSecurityPackage.cs
```diff
@@ -41,7 +42,9 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security
     [Guid("D7D54E08-45E1-49A6-AA53-AF1CFAA6EBDC")]
     [ProvideMenuResource("Menus.ctmenu", 1)]
     [ProvideToolWindow(typeof(HotspotsToolWindow), MultiInstances = false, Transient = true, Style = VsDockStyle.Tabbed, Window = VsWindowKindErrorList, Width = 700, Height = 250)]
-    [ProvideToolWindow(typeof(TaintToolWindow), MultiInstances = false, Transient = true, Style = VsDockStyle.Tabbed, Window = VsWindowKindErrorList, Width = 700, Height = 250)]
+    [ProvideToolWindow(typeof(TaintToolWindow), MultiInstances = false, Transient = false, // Note: Transient must false when using ProvideToolWindowVisibility
```
rita-g-sonarsource, line 45: `Transient = false` is the default, maybe it's not worth setting it at all and avoid the comment this way.

duncanp-sonar: I think it's clearer with the comment as it's not obvious and not state in the VS docs.

rita-g-sonarsource (approved):
@duncanp-sonar looks good, just see my comment regarding the frame persistence. It was a problem with secondary locations panel.

src/IssueViz.Security/IssueVizSecurityPackage.cs
rita-g-sonarsource: Ok (it is in the docs though https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.providetoolwindowattribute.transient?view=visualstudiosdk-2019#Microsoft_VisualStudio_Shell_ProvideToolWindowAttribute_Transient)

src/IssueViz.Security/IssueVizSecurityPackage.cs
duncanp-sonar: Good catch!

duncanp-sonar: FYI I added this code, but it didn't seem to stop the window being reopened next time VS started. I'll push this code as it is.
I have the "close frame" code plus tests in a separate commit and will test it again tomorrow.

rita-g-sonarsource: Hmm, have you tried to reset the hive? Just to make sure that it's not caching something improperly.

src/IssueViz.Security/IssueVizSecurityPackage.cs
duncanp-sonar: It's not in the docs that the value has to be false, otherwise it seems to mess up the operation of the `ProvideToolWindowVisibilityAttribute`.

PR #2069, opened 2021-01-08
Initialize taint sync monitor in a separate package

duncanp-sonar:
* avoids reentrant calls on UI thread causing VS to lock up

Relates to #2059

### DIFF
--- a/src/Core/Binding/IActiveSolutionBoundTracker.cs
+++ b/src/Core/Binding/IActiveSolutionBoundTracker.cs
@@ -38,6 +38,8 @@ namespace SonarLint.VisualStudio.Core.Binding
         /// <summary>
         /// Event to notify subscribers when the binding status of a solution have changed.
         /// This occurs when a new solution is opened, or the SonarQube binding status of the solution changes.
+        /// The event is raised after the <see cref="ISonarQubeService"/> connection has finished
+        /// opening/closing.
         /// </summary>
         event EventHandler<ActiveSolutionBindingEventArgs> SolutionBindingChanged;
 
--- a/src/IssueViz.Security/IssueVizSecurityPackage.cs
+++ b/src/IssueViz.Security/IssueVizSecurityPackage.cs
@@ -23,10 +23,8 @@ using System.ComponentModel.Design;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.InteropServices;
 using System.Threading;
-using Microsoft.VisualStudio;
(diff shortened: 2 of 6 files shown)


src/Core/Binding/IActiveSolutionBoundTracker.cs
```diff
@@ -38,6 +38,8 @@ public interface IActiveSolutionBoundTracker
         /// <summary>
         /// Event to notify subscribers when the binding status of a solution have changed.
         /// This occurs when a new solution is opened, or the SonarQube binding status of the solution changes.
+        /// The event is raised after the <see cref="ISonarQubeService"/> connection has finished
```
duncanp-sonar, line 41: Miscellaneous: added documentation to the interface and class to make it clear that the connection has changed by the time the event is raised.

duncanp-sonar:
This change prevents the UI lock-up I was seeing in some circumstances.
I've also tested that the sync monitor can force-create the taint tool window without causing a lock-up (which it will need to do so the tool window caption can be updated dynamically). I'll send a separate PR for that change, and the force-close change if required.

src/IssueViz.Security/IssueVizSecurityPackage.cs
```diff
@@ -36,9 +34,17 @@
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security
 {
+    /*
+        This is a simple package that just provides the basic VS plumbing for the security commands
+        and tool windows.
+
+        It doesn't provide any other services, and doesn't need to be auto-loadable. It will be loaded
+        by VS only when needed i.e. when it has received a reqeust to to invoke a command or display a
```
rita-g-sonarsource, line 42: reqeust  = typo.

But I don't think this comment is needed, we have other packages without AutoLoad (i.e. IssueVizPackage) so it's not uncommon.

rita-g-sonarsource (approved):

src/IssueViz.Security/IssueVizSecurityPackage.cs
duncanp-sonar: I don't think we have a clear enough idea what specific packages are for or what there load characteristics are expected to be. I'm not even sure how many packages we have any more. However, my feeling is that we have too many, we don't understand what the relationships between them are, and some of them will be loading unnecessarily/sooner than necessary. This isn't something we can address in the short term. However, I'm going to start adding comments to packages as part of an effort to understand what we currently have.

PR #2070, opened 2021-01-08
Remove ObservableCollection from TaintStore

### LINKED ISSUE #2064: Taint tags are not cleared when solution binding changes

rita-g-sonarsource:
Fixes #2064 #2037

### DIFF
--- a/src/IssueViz.Security.UnitTests/IssuesStoreToIssueLocationStoreAdapterTests.cs
+++ b/src/IssueViz.Security.UnitTests/IssuesStoreToIssueLocationStoreAdapterTests.cs
@@ -1,0 +1,297 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.ComponentModel.Composition.Hosting;
+using System.Linq;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Security.IssuesStore;
+using IssuesChangedEventArgs = SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging.IssuesChangedEventArgs;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests
+{
+    [TestClass]
+    public class IssuesStoreToIssueLocationStoreAdapterTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckExports()
+        {
+            var batch = new CompositionBatch();
+
+            batch.AddExport(MefTestHelpers.CreateExport<IIssuesStore>(Mock.Of<IIssuesStore>()));
+            batch.AddExport(MefTestHelpers.CreateExport<IIssuesStore>(Mock.Of<IIssuesStore>()));
+            batch.AddExport(MefTestHelpers.CreateExport<IIssuesStore>(Mock.Of<IIssuesStore>()));
+
+            var adapterImport = new SingleObjectImporter<IIssuesStoreToIssueLocationStoreAdapter>();
+            batch.AddPart(adapterImport);
+
+            var locationStoreImport = new SingleObjectImporter<IIssueLocationStore>();
+            batch.AddPart(locationStoreImport);
+
+            using var catalog = new TypeCatalog(typeof(IssuesStoreToIssueLocationStoreAdapter));
+            using var container = new CompositionContainer(catalog);
+            container.Compose(batch);
+
+            adapterImport.Import.Should().NotBeNull();
+            locationStoreImport.Import.Should().NotBeNull();
+        }
+
+        [TestMethod]
+        public void IssueStoresChanged_NoSubscribersToIssuesChangedEvent_NoException()
+        {
+            var issueStore = new Mock<IIssuesStore>();
+            CreateTestSubject(issueStore.Object);
+
+            var act = new Action(() => RaiseIssuesChangedEvent(issueStore, newIssues: new[] {CreateIssueViz()}));
+            act.Should().NotThrow();
+        }
+
+        [TestMethod]
+        public void IssueStoresChanged_HasSubscribersToIssuesChangedEvent_SubscribersNotified()
+        {
+            var issueStore1 = new Mock<IIssuesStore>();
+            var issueStore2 = new Mock<IIssuesStore>();
+
+            var testSubject = CreateTestSubject(issueStore1.Object, issueStore2.Object);
+
+            IssuesChangedEventArgs suppliedArgs = null;
+            var eventCount = 0;
+            testSubject.IssuesChanged += (sender, args) => { suppliedArgs = args; eventCount++; };
+
+            var location1 = CreateLocation("b.cpp");
+            var location2 = CreateLocation("B.cpp");
+            var issueViz1 = CreateIssueViz("a.cpp", location1, location2);
+
+            RaiseIssuesChangedEvent(issueStore1, newIssues: new[] {issueViz1});
+
+            eventCount.Should().Be(1);
+            suppliedArgs.Should().NotBeNull();
+            suppliedArgs.AnalyzedFiles.Should().BeEquivalentTo("a.cpp", "b.cpp");
+
+            var issueViz2 = CreateIssueViz("c.cpp");
+
+            eventCount = 0;
+            RaiseIssuesChangedEvent(issueStore2, newIssues: new[] { issueViz2 });
+
+            eventCount.Should().Be(1);
+            suppliedArgs.Should().NotBeNull();
+            suppliedArgs.AnalyzedFiles.Should().BeEquivalentTo("c.cpp");
+
+            eventCount = 0;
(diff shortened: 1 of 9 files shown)


src/IssueViz.Security/IssuesStore/IssuesStoreToIssueLocationStoreAdapter.cs
```diff
@@ -0,0 +1,126 @@
...
+    {
+    }
+
+    /// <summary>
+    /// MEF-imports <see cref="IIssuesStore"/> and implements <see cref="IIssueLocationStore"/> on top of the given stores.
+    /// </summary>
+    [Export(typeof(IIssuesStoreToIssueLocationStoreAdapter))]
+    [Export(typeof(IIssueLocationStore))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal sealed class IssuesStoreToIssueLocationStoreAdapter : IIssuesStoreToIssueLocationStoreAdapter
```
rita-g-sonarsource, line 44: Same code as from IssueStoreObserver, except it uses IssuesStore instead of Observable collections

src/IssueViz.Security.UnitTests/IssuesStoreToIssueLocationStoreAdapterTests.cs
```diff
@@ -0,0 +1,297 @@
...
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Security.IssuesStore;
+using IssuesChangedEventArgs = SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging.IssuesChangedEventArgs;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests
+{
+    [TestClass]
+    public class IssuesStoreToIssueLocationStoreAdapterTests
```
rita-g-sonarsource, line 39: Adjusted tests from Observer. 
You can see a diff here: https://gist.github.com/rita-g-sonarsource/ec0b3d0a13887fcc2cbde931064069ea/revisions

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -32,17 +29,14 @@
...
-using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Security.IssuesStore;
 using SonarLint.VisualStudio.IssueVisualization.Security.SharedUI;
 using SonarLint.VisualStudio.IssueVisualization.Security.Taint.Models;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.ViewModels
 {
     internal interface ITaintIssuesControlViewModel : IDisposable
     {
-        // WIP: add selection handling
```
rita-g-sonarsource, line 43: These comments can be removed: there are tickets for the selection and other things that we might need.

rita-g-sonarsource:
@duncanp-sonar I left a modified version of the observer/adapter and I'm using it in the taint store. I think it looks cleaner than extension methods: I started creating them but they're not general purpose, and it's very clear that they're all part of a unified logic. I don't think extension methods are the best way to avoid the code duplication between Taint and Hotspots, and a service is more appropriate. Hence, the new adapter service.
Let me know what you think.

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -62,17 +54,10 @@ public void Set(IEnumerable<IAnalysisIssueVisualization> issueVisualizations)
...
 
-            taintVulnerabilities.Clear();
-
-            foreach (var issueViz in issueVisualizations)
-            {
-                taintVulnerabilities.Add(issueViz);
-            }
-        }
+            var oldIssues = taintVulnerabilities;
+            taintVulnerabilities = issueVisualizations;
```
duncanp-sonar, line 58: It would be safer to store a copy of the enumerable so this class won't be impacted by changes to the caller (e.g. if it's passed a non-reified enumerator).

src/IssueViz.Security/IssuesStore/IIssuesStore.cs
```diff
@@ -0,0 +1,45 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.IssuesStore
+{
+    internal interface IIssuesStore
```
duncanp-sonar, line 27: Doc comment

src/IssueViz.Security/IssuesStore/IssuesStoreToIssueLocationStoreAdapter.cs
```diff
@@ -0,0 +1,126 @@
...
+using SonarLint.VisualStudio.Core.Helpers;
+using SonarLint.VisualStudio.IssueVisualization.Editor.LocationTagging;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.IssuesStore
+{
+    /// <summary>
+    /// Adapter between <see cref="IIssuesStore"/> and <see cref="IIssueLocationStore"/>
+    /// </summary>
+    internal interface IIssuesStoreToIssueLocationStoreAdapter : IIssueLocationStore, IDisposable
```
duncanp-sonar, line 34: Is this interface required, and if so what is going to import it?

src/IssueViz.Security/IssuesStore/IssuesStoreToIssueLocationStoreAdapter.cs
duncanp-sonar, line 44: From the name I'd expect this to be adapter an `IIssueStore` to an `IIssueStoreLocation`; I wouldn't expect it to be an aggregator too.

Possible names that would convey this:
`AggregatingStoreToLocationStoreAdapter`
`AggregatingIssueLocationStoreAdapter`

I don't think it's necessary to include the full source and target type names in the class name; it's pretty obvious from the context and the interfaces that are implemented what the relevant types are.

src/IssueViz.Security/IssuesStore/IssuesStoreToIssueLocationStoreAdapter.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        public void Refresh(IEnumerable<string> affectedFilePaths)
+        {
+            // Implementation is not required:
+            // This method was originally added in order to notify the error list sinks that the taggers changed the span of the issue visualizations.
+            // The newer issue stores display their data in plain xaml lists, and the span changes are tracked via NotifyPropertyChanged mechanism in IAnalysisIssueVisualization.
+        }
+
+        public bool Contains(IAnalysisIssueVisualization issueVisualization)
```
duncanp-sonar, line 79: Thinking about it, `Contains` was added to fix the selection bug in the Error List, wasn't it? If the problem only exists for the Error List, does the `Contains` method actually need to be on the `IIssueLocation` interface?

Assuming that is the case, we should fix it separately (although it would mean there might be no point in implementing or testing the method here).

src/IssueViz.Security.UnitTests/IssuesStoreToIssueLocationStoreAdapterTests.cs
```diff
@@ -0,0 +1,297 @@
...
+    {
+        [TestMethod]
+        public void MefCtor_CheckExports()
+        {
+            var batch = new CompositionBatch();
+
+            batch.AddExport(MefTestHelpers.CreateExport<IIssuesStore>(Mock.Of<IIssuesStore>()));
+            batch.AddExport(MefTestHelpers.CreateExport<IIssuesStore>(Mock.Of<IIssuesStore>()));
+            batch.AddExport(MefTestHelpers.CreateExport<IIssuesStore>(Mock.Of<IIssuesStore>()));
```
duncanp-sonar, line 48: The test isn't checking whether these types are actually imported so they aren't adding any value (in fact, we'd expect the export of the aggregating adapter to work even if there were no exported `IIssuesStores`).

I think there are two separate things to check:
1) is the class exportable as an `IIssueLocationStore`? (as commented elsewhere, I'm not sure the other export is required)
2) does the class import multiple `IIssuesStore` instances? (not currently tested)

duncanp-sonar:
It's definitely better than the current version; the code is simpler, the dependencies are not as complicated (no explicit registration/unregistration), and it doesn't have the bugs of the current version.

There are a couple of things that jar: the aggregating adapter feels a bit odd; there's very little left in the TaintStore class, and presumably the same will apply to the HotspotStore, and it will be almost the same as the TaintStore so it's not clear there's enough left to justify their existence as separate classes.

Still, this is clearly better than what is there now so let's go with it and revisit it another time if appropriate.

duncanp-sonar (commented):
See inline comments and questions, all relatively minor.

src/IssueViz.Security/IssuesStore/IssuesStoreToIssueLocationStoreAdapter.cs
rita-g-sonarsource: Hmm, the problem was that the selection service (and the panel) displayed an issue that was no longer there. 
We haven't added secondary locations yet to taint and hotspots, so it's difficult to know if the same problem would exist there or not, but it's a good point.
I can think of 2 similar scenarios:
1. Select a taint issue that has secondary locations and then re-sync the binding so that the issue no longer exists.
2. Select a hotspot that has secondary locations and then remove it from the list.

Depending on the implementation of these two, there might be a selection bug and might not.
As you may recall, the original version of the error list store cleared the selection each time a new analysis was performed. This is similar to the 2 scenarios above; if there's no "monitor" service, then we would need to clear the selection. It doesn't sound so bad honestly... I hope that the monitor and the contains method can both go away when we change the selection.

src/IssueViz.Security/Taint/TaintStore.cs
duncanp-sonar: The class is still storing the original enumerable that was passed in, so it could be mutated externally.

rita-g-sonarsource: Oh that's what you meant; my bad. I thought you meant the passing to the event args.

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -57,7 +57,7 @@ public void Set(IEnumerable<IAnalysisIssueVisualization> issueVisualizations)
             var oldIssues = taintVulnerabilities;
             taintVulnerabilities = issueVisualizations;
 
-            IssuesChanged?.Invoke(this, new IssuesChangedEventArgs(oldIssues, taintVulnerabilities));
+            IssuesChanged?.Invoke(this, new IssuesChangedEventArgs(oldIssues, taintVulnerabilities.ToList()));
```
duncanp-sonar, line 60: I hadn't thought about passing an immutable type in the event, but that makes sense too.

It might be better make the store store an immutable collection, then it could be safely passed in the event args.

PR #2071, opened 2021-01-08
Remove ObservableCollection from HotspotsStore

rita-g-sonarsource:


src/IssueViz.Security/Hotspots/HotspotsStore.cs
```diff
@@ -65,14 +59,22 @@ public IAnalysisIssueVisualization GetOrAdd(IAnalysisIssueVisualization hotspotV
                 return existingHotspot;
             }
 
+            var oldIssues = hotspots.ToArray();
+
             hotspots.Add(hotspotViz);
 
+            NotifyIssuesChanged(oldIssues);
```
duncanp-sonar, line 66: The only change to the collection is the addition of a new issue and it isn't part of the notification. Is this correct?

src/IssueViz.Security/Hotspots/HotspotsStore.cs
```diff
@@ -65,14 +59,22 @@ public IAnalysisIssueVisualization GetOrAdd(IAnalysisIssueVisualization hotspotV
...
             return hotspotViz;
         }
 
         public void Remove(IAnalysisIssueVisualization hotspotViz)
         {
+            var oldIssues = hotspots.ToArray();
+
             hotspots.Remove(hotspotViz);
+
+            NotifyIssuesChanged(oldIssues);
```
duncanp-sonar, line 77: This seems odd too - why are we saying "everything's changed" when only one issue has been removed?

src/IssueViz.Security/Hotspots/HotspotsStore.cs
duncanp-sonar: Ah - I see. Seems a bit weird to return all of the old and new issue in the event arguments. I missed that before - I incorrectly interpreted it as "removed" and "added".

duncanp-sonar (approved):

src/IssueViz.Security/Hotspots/HotspotsStore.cs
rita-g-sonarsource: Ah, but I do realize that there is a bug in the StoreAdapter: I need to notify only on what's changed, so a diff between old and new 🤦 I'll fix that in a separate PR.

PR #2072, opened 2021-01-08
Modify IShowToolWindow to support conditional window activation

duncanp-sonar:
Prep for #2059

### DIFF
--- a/src/Core/IToolWindowService.cs
+++ b/src/Core/IToolWindowService.cs
@@ -27,6 +27,7 @@ namespace SonarLint.VisualStudio.Core
         /// <summary>
         /// Opens specified tool window
         /// </summary>
-        void Show(Guid toolWindowId);
+        /// <param name="activateWindow">True if the window should receive focus, otherwise false</param>
+        void Show(Guid toolWindowId, bool activateWindow);
     }
 }
--- a/src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
+++ b/src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
@@ -64,7 +64,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Helpers
             // Act
             using (new AssertIgnoreScope())
             {
-                testSubject.Show(ValidToolWindowId);
+                testSubject.Show(ValidToolWindowId, true);
             }
 
             // Non-success HResult so shouldn't have attempted to show the frame
@@ -81,12 +81,14 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Helpers
 
             var testSubject = new ToolWindowService(serviceProviderMock.Object);
 
-            Action act = () => testSubject.Show(ValidToolWindowId);
+            Action act = () => testSubject.Show(ValidToolWindowId, true);
             act.Should().NotThrow();
         }
 
         [TestMethod]
-        public void Show_FindWindowSucceededAndNonNullToolWindow_ToolWindowShown()
+        [DataRow(true)]
+        [DataRow(false)]
+        public void Show_FindWindowSucceededAndNonNullToolWindow_ToolWindowShown(bool shouldActivateWindow)
         {
             var serviceProviderMock = new Mock<IServiceProvider>();
             var uiShellMock = new Mock<IVsUIShell>();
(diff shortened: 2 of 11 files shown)


src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
```diff
@@ -97,9 +99,10 @@ public void Show_FindWindowSucceededAndNonNullToolWindow_ToolWindowShown()
             var testSubject = new ToolWindowService(serviceProviderMock.Object);
 
             // Act
-            testSubject.Show(ValidToolWindowId);
+            testSubject.Show(ValidToolWindowId, shouldActivateWindow);
 
-            frameMock.Verify(x => x.Show(), Times.Once);
+            var expectedShowCallCount = shouldActivateWindow ? 1 : 0;
+            frameMock.Verify(x => x.Show(), Times.Exactly(expectedShowCallCount));
```
duncanp-sonar, line 105: Modified test to check `frame.Show` is called correctly.
All of the other test changes are to just to fix up the mock to have the new parameter and to check that all of the existing callers are passing `true`.

src/Core/IToolWindowService.cs
```diff
@@ -27,6 +27,7 @@ public interface IToolWindowService
         /// <summary>
         /// Opens specified tool window
         /// </summary>
-        void Show(Guid toolWindowId);
+        /// <param name="activateWindow">True if the window should receive focus, otherwise false</param>
+        void Show(Guid toolWindowId, bool activateWindow);
```
duncanp-sonar, line 31: Update to the interface.

src/Integration/Helpers/ToolWIndowService.cs
```diff
@@ -40,15 +40,15 @@ public ToolWindowService([Import(typeof(SVsServiceProvider))] IServiceProvider s
...
+        public void Show(Guid toolWindowId, bool activateWindow)
         {
             // We want VS to ask the package to create the tool window if it doesn't already exist
             const uint flags = (uint)__VSFINDTOOLWIN.FTW_fForceCreate;
 
             var hr = shell.FindToolWindow(flags, toolWindowId, out var windowFrame);
             Debug.Assert(ErrorHandler.Succeeded(hr), $"Failed to find tool window. Guid: {toolWindowId}, hr: {hr} ");
 
-            if (ErrorHandler.Succeeded(hr))
+            if (activateWindow && ErrorHandler.Succeeded(hr))
```
duncanp-sonar, line 51: This is the actual product change.
All of the other product code changes are to pass `true` as the additional parameter.

duncanp-sonar:
Prep for changing the taint synchronizer to be able to ensure the taint control exists without showing it.
The actual product change is minimal; the rest of the changes are to the existing consumers of `IToolWindowService` to pass the additional parameter.
I`ve called out the "real" product and test changes in comments.

src/Core/IToolWindowService.cs
```diff
@@ -27,6 +27,7 @@ public interface IToolWindowService
         /// <summary>
         /// Opens specified tool window
         /// </summary>
-        void Show(Guid toolWindowId);
+        /// <param name="activateWindow">True if the window should receive focus, otherwise false</param>
```
rita-g-sonarsource, line 30: Based on the doc comment, maybe `focus` is more appropriate?

duncanp-sonar: `Activate` is the terminology used in the VS interfaces and documentation so I'd rather not introduce a new term.
I'll change the comment.

duncanp-sonar: Previously, calling the method would create the window if it didn`t already exist, bring it to the front and give it the focus.
The new parameter controls whether the window is brought to the front _and_ given the focus.

rita-g-sonarsource: Ok, maybe we should separate it to 2 methods then:

`void CreateAndActivate(Guid toolWindowId)` - which is what we would call everywhere (essentially renaming the Show)
`void Create(Guid toolWindowId)` - which is what we can call in the new taint context.

This way it would be clearer what it does and we won't need to pass `true` param in most cases.

duncanp-sonar: I`ll add another method.

PR #2074, opened 2021-01-11
Change IssuesChangedEventArgs to contain RemovedIssues and AddedIssues

rita-g-sonarsource:


src/IssueViz.Security/IssuesStore/IIssuesStore.cs
```diff
@@ -37,20 +37,14 @@ internal interface IIssuesStore
 
     internal class IssuesChangedEventArgs
```
rita-g-sonarsource, line 38: This was a bug the previous PR https://github.com/SonarSource/sonarlint-visualstudio/pull/2070/files#diff-7af7f940524863076596a6aae92cdb28e1dcb6e9cf0100d75b963f5dc9d126bd

I misunderstood ObservableCollection's event args [NotifyCollectionChangedEventArgs.OldItems](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.notifycollectionchangedeventargs.olditems?view=net-5.0#System_Collections_Specialized_NotifyCollectionChangedEventArgs_OldItems).
As a result, I was updating the tags for all the issues, instead of only what's really changed.

I thought about keeping it OldItems/NewItems and make the changes in the `AggregatingIssueLocationStoreAdapter`, but the names are what confused me in the first place, so I thought it was more clear to just name it Removed/Added and pass the appropriate params in the callers.

duncanp-sonar (approved):

PR #2080, opened 2021-01-11
Changed taint control view model to expose view directly

duncanp-sonar:


src/IssueViz.Security/Taint/TaintList/TaintIssuesControl.xaml
```diff
@@ -46,17 +46,17 @@
         </Grid.ColumnDefinitions>
         <DataGrid Grid.Column="0" 
                   SelectedItem="{Binding SelectedIssue, Mode=TwoWay}"
-                  ItemsSource="{Binding Issues}"
+                  ItemsSource="{Binding IssuesView}"
```
duncanp-sonar, line 49: The control now binds explicitly to the exposed `ICollectionView`, rather than implicitly creating one behind the scenes.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -56,7 +56,10 @@ internal sealed class TaintIssuesControlViewModel : ITaintIssuesControlViewModel
         private readonly object Lock = new object();
         private string activeDocumentFilePath;
 
-        public ObservableCollection<ITaintIssueViewModel> Issues { get; } = new ObservableCollection<ITaintIssueViewModel>();
+        // TODO: duncanp - make collection private
+        internal /* for testing */ ObservableCollection<ITaintIssueViewModel> Issues { get; } = new ObservableCollection<ITaintIssueViewModel>();
```
duncanp-sonar, line 60: If we decide to keep this change I'll change the scope in a separate PR (changing it here would have added quite a bit of noise).

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -109,13 +112,9 @@ private bool ActiveDocumentFilter(object viewModel)
             return PathHelper.IsMatchingPath(issueFilePath, activeDocumentFilePath);
         }
 
-        private void SetDefaultSortOrder()
-        {
-            var collectionView = CollectionViewSource.GetDefaultView(Issues);
-
-            collectionView.SortDescriptions.Add(
+        private void SetDefaultSortOrder() =>
```
rita-g-sonarsource, line 115: Any chance that the SortDescriptions xaml binding will work now?
https://stackoverflow.com/a/1280768/339720

rita-g-sonarsource (approved):

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
duncanp-sonar: Perhaps, but that example is actually creating the collection in XAML so I don't think that particular example would work.

PR #2081, opened 2021-01-12
Add telemetry to Taint vulnerabilities

### LINKED ISSUE #1983: Add telemetry
### Description

_Extracted from the working doc:_
> What we should try to understand is:
Is the feature actually used (i.e. do developers review some taint vulnerabilities in the IDE)?
What we can log in the daily ping is:
Number of times an issue is actually opened in the IDE (this is not a unique count for simplicity, thus if the same issue is opened ten times it will be counted ten times.)
Number of redirects to SQ/SC to review issues

#### Implementation
- [ ] add two new data points to the telemetry payload: a running total of `taint issues opened` and a running total of `taint issues opened on server`
- [ ] increment a running total each time the user "opens" a taint issue i.e. double-clicks on an issue in the Error List, or uses the light bulb to navigate to the issue.
- [ ] increment a running total each time the user executes the "Show in SonarQube/SonarCloud" action

#### Background notes
An earlier draft of the telemetry requirements called for recording telemetry when the tool window was activated, but this requirement was dropped for the initial release.
Should it become necessary, there are couple of ways of ways to be notified of tool window events:
* implement `IVsWindowFrameNotify3` on the `ToolWindowPane`-derived class.
* implement `IVsWindowFrameEvents` and use `IVsUIShell7.AdviseWindowFrameEvents` to register for notifications

rita-g-sonarsource:
Fixes #1983

### DIFF
--- a/src/Core/Telemetry/ITelemetryManager.cs
+++ b/src/Core/Telemetry/ITelemetryManager.cs
@@ -28,7 +28,7 @@ namespace SonarLint.VisualStudio.Core.Telemetry
         void Update();
         void LanguageAnalyzed(string languageKey);
         void ShowHotspotRequested();
-        void TaintIssueOpened();
-        void TaintIssueOpenedInBrowser();
+        void TaintIssueInvestigatedLocally();
+        void TaintIssueInvestigatedRemotely();
     }
 }
--- a/src/Core/Telemetry/ITelemetryManager.cs
+++ b/src/Core/Telemetry/ITelemetryManager.cs
@@ -28,5 +28,7 @@ namespace SonarLint.VisualStudio.Core.Telemetry
         void Update();
         void LanguageAnalyzed(string languageKey);
         void ShowHotspotRequested();
+        void TaintIssueOpened();
+        void TaintIssueRedirected();
     }
 }
--- a/src/Integration.UnitTests/Telemetry/TelemetryDataTests.cs
+++ b/src/Integration.UnitTests/Telemetry/TelemetryDataTests.cs
@@ -55,6 +55,12 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
                 ShowHotspot = new ShowHotspot
                 {
                     NumberOfRequests = 567
+                },
+
+                Taint = new Taint
+                {
+                    NumberOfOpenedOnServer = 88,
+                    NumberOfOpenedIssues = 99
                 }
             };
 
@@ -86,6 +92,10 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
 
             reloadedData.ShowHotspot.Should().NotBeNull();
             reloadedData.ShowHotspot.NumberOfRequests.Should().Be(567);
+
+            reloadedData.Taint.Should().NotBeNull();
+            reloadedData.Taint.NumberOfOpenedOnServer.Should().Be(88);
+            reloadedData.Taint.NumberOfOpenedIssues.Should().Be(99);
         }
     }
 }
--- a/src/Integration.UnitTests/Telemetry/TelemetryDataTests.cs
+++ b/src/Integration.UnitTests/Telemetry/TelemetryDataTests.cs
@@ -57,10 +57,10 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
                     NumberOfRequests = 567
                 },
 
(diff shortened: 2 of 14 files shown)


rita-g-sonarsource:
@duncanp-sonar naming aside, it's ready for review. Once we decide on the final field names I'll make the appropriate changes.

src/Core/Telemetry/ITelemetryManager.cs
```diff
@@ -28,5 +28,7 @@ public interface ITelemetryManager
         void Update();
         void LanguageAnalyzed(string languageKey);
         void ShowHotspotRequested();
+        void TaintIssueOpened();
```
rita-g-sonarsource, line 31: As we discussed online, it might be time to switch to a generic method `void IncrementDailyCounter(string counterName)`, which would reduce a lot of the copy-paste. We can do it next time we need to add telemetry.

src/Core/Telemetry/ITelemetryManager.cs
```diff
@@ -28,5 +28,7 @@ public interface ITelemetryManager
...
+        void TaintIssueOpenedInBrowser();
```
duncanp-sonar, line 32: I think calling the methods something like `TaintIssueOpenedLocally/OnClient` and `TaintIssueOpenedOnServer` would be slightly clearer, but it doesn't matter too much.

src/Integration.UnitTests/Telemetry/TelemetryDataTests.cs
```diff
@@ -55,6 +55,12 @@ public void XmlSerialization_RoundTrips()
                 ShowHotspot = new ShowHotspot
                 {
                     NumberOfRequests = 567
+                },
+
+                Taint = new Taint
+                {
+                    NumberOfOpenedOnServer = 88,
```
duncanp-sonar, line 62: Minor: this is one name that will definitely have to change as it doesn't make sense grammatically.

rita-g-sonarsource: IEnglishSpeakNot

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -626,6 +640,48 @@ public void ShowHotspotRequested_CounterIncremented(int previousCounter)
...
+        [DataRow(1)]
+        [DataRow(100)]
+        public void TaintIssueOpened_CounterIncremented(int previousCounter)
+        {
+            var telemetryData = new TelemetryData { Taint = new Taint { NumberOfOpenedIssues = previousCounter } };
+            telemetryRepositoryMock.Setup(x => x.Data).Returns(telemetryData);
+            var testSubject = CreateManager();
+
+            // Set up the telemetry mock after creating the manager, since the manager will
+            // may saved data on initial creation
```
duncanp-sonar, line 654: The comment needs to be tweaked: "... manager will may saved data..."

rita-g-sonarsource: Darn copy-paste.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -130,6 +134,8 @@ private void SetCommands(ILocationNavigator locationNavigator)
             NavigateCommand = new DelegateCommand(
                 parameter =>
                 {
+                    telemetryManager.TaintIssueOpened();
```
duncanp-sonar, line 137: We haven't implement light-bulb navigation yet, but presumably we'll want to include that in the telemetry too (?). For now, I'd just add a note in the taint lightbulb ticket so we don't forget.

duncanp-sonar (approved):

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
rita-g-sonarsource: Yep, I was thinking about that too, of maybe putting it in a more centralized place (e.g. if we have something that calls the selection service, since that's what both actions will do in addition to navigating).

PR #2082, opened 2021-01-12
Ensure taint tool window is initialized

duncanp-sonar:
Relates to #2059

Modifies the taint synchronizer to ensure the taint tool window has been created so the tool window caption can be set correctly.

### DIFF
--- a/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
+++ b/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
@@ -110,6 +110,9 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
                 UpdateTaintIssuesUIContext(hasTaintIssues);
                 if (hasTaintIssues)
                 {
+                    // We need the tool window content to exist so the issues are filtered and the
+                    // tool window caption is updated. See the "EnsureToolWindowExists" method comment
+                    // for more information.
                     toolWindowService.EnsureToolWindowExists(TaintToolWindow.ToolWindowId);
                 }
             }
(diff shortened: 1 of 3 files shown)


src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -101,7 +105,13 @@ public async Task SynchronizeWithServer()
 
                 var taintIssueVizs = taintVulnerabilities.Select(converter.Convert).ToArray();
                 taintStore.Set(taintIssueVizs);
-                UpdateTaintIssuesUIContext(taintIssueVizs.Any());
+
+                var hasTaintIssues = taintIssueVizs.Length > 0;
+                UpdateTaintIssuesUIContext(hasTaintIssues);
+                if (hasTaintIssues)
+                {
+                    toolWindowService.EnsureToolWindowExists(TaintToolWindow.ToolWindowId);
```
duncanp-sonar, line 113: Delay creating the tool window as long as possible. It will only be created in situations where we actually want to update the caption i.e. there are taint issues.

rita-g-sonarsource: Why do we need to do this if we're setting the UI context?
It should flip the `ProvideToolWindowVisibility` switch and show the toolwindow - isn't it the same as calling `EnsureToolWindowExists`?

duncanp-sonar: No. VS will draw the tool window tab without actually creating the window content if it can e.g. if the tab isn't top-most.
In that case, the tab will appear with whatever caption it had last time VS shut down, and it won't change dynamically.

I'll add a comment.

rita-g-sonarsource (approved):

Issue #2079, opened 2021-01-11
Taint filtering does not work for single-click navigation in the Solution Explorer


duncanp-sonar:
### Description

The taint issues list is supposed to be filtered to only show issues for the active document. It works if the user navigates by double-clicking in the Solution Explorer, but not if the VS is configured to navigate by single-clicking in the Solution Explorer.

Note: this bug will also affect the shipped pre-compiled headers code. However, that won't be as visible to the user as it is a performance optimisation and there is no visual indication that anything is wrong. The CFamily issues will still be calculated correctly, but won't benefit from the optimisation.

Related: #2091

### Repro steps

1. Enable single-click navigation in the Solution Explorer by allowing files to be previewed in the Error List:

![image](https://user-images.githubusercontent.com/10757559/104219889-9fbf1780-5436-11eb-97b1-1126730c26c8.png)

2. Open a bound solution contain taint issues.
3. Navigate between files by single-clicking in the Solution Explorer.

### Expected behavior

The taint list is filtered to the previewed document.

### Actual behavior

The filter is not updated.

### Known workarounds

Disable single-click preview / navigate by double-clicking

### Related information

* SonarLint for VS Version: v4.32

### Additional notes
The affected component is `ActiveDocumentTracker`. It uses `IVsSelectionEvents` to detect when a new file has selection. In the case of single-click navigation, the file will be opened but the selection remains in the Solution Explorer.

PR #2083, opened 2021-01-12
Change taint window caption dynamically

duncanp-sonar:
Relates to #2059

### DIFF
--- a/src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.Caption.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.Caption.cs
@@ -1,0 +1,260 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.ComponentModel;
+using System.Linq;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.Text;
+using Moq;
+using SonarLint.VisualStudio.Infrastructure.VS;
+using SonarLint.VisualStudio.Infrastructure.VS.DocumentEvents;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.IssueVisualization.Editor;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Security.IssuesStore;
+using SonarLint.VisualStudio.IssueVisualization.Security.SharedUI;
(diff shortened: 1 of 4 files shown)


src/IssueViz.Security/Taint/TaintList/TaintToolWindow.cs
```diff
@@ -52,7 +52,36 @@ public TaintToolWindow(IServiceProvider serviceProvider)
                 componentModel.GetService<ITelemetryManager>()
             );
 
-            Content = new TaintIssuesControl(viewModel);
+            Initialize(viewModel);
+        }
+
+        internal /* for testing */ TaintToolWindow(ITaintIssuesControlViewModel viewModel) =>
```
duncanp-sonar, line 58: I added a second constructor to make it easier to add some unit tests for the tool window class.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -177,9 +195,29 @@ private void UpdateIssues()
             NotifyPropertyChanged(nameof(HasServerIssues));
         }
 
+        private void UpdateCaption()
+        {
+            // We'll show the default caption if:
+            // * there are no underlying issues, or
+            // * there is not ana active document.
```
duncanp-sonar, line 202: Typo

duncanp-sonar:
There are a couple of scenarios in which the caption will not update correctly. However, both of these are due to bugs in the `IActiveDocumentTracker`. I raised #2079 last week (single-click navigation), and found another one today (#2091 - filter not updated when the last document is closed).

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.Caption.cs
```diff
@@ -0,0 +1,261 @@
...
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint.TaintList
+{
+    [TestClass]
+    public class TaintIssuesControlViewModelTestsCaption
+    {
+        private readonly string DefaultCaption = Resources.TaintToolWindowCaption;
+
+        [TestMethod]
+        public void Ctor_NoUnderlyingIssues_DefaultCaptionSet()
```
duncanp-sonar, line 48: There are four different cases to test to check that the correct caption is set when the window is first opened.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -177,9 +195,29 @@ private void UpdateIssues()
...
+            WindowCaption = Resources.TaintToolWindowCaption + suffix;
+        }
+
+        private int GetFilteredIssuesCount() =>
+            IssuesView.OfType<object>().Count();
+
         private void Store_IssuesChanged(object sender, IssuesChangedEventArgs e)
         {
             UpdateIssues();
+            UpdateCaption();
```
rita-g-sonarsource, line 220: Minor: why not set the UpdateCaption inside UpdateIssues? This way we won't need to explicitly call it from the ctor as well.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.Caption.cs
```diff
@@ -0,0 +1,261 @@
...
+        {
+            var testCase = new TestCase()
```
rita-g-sonarsource, line 50: Very neat syntax, I like it.

rita-g-sonarsource (approved):

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
duncanp-sonar: That's what I did originally but it's not enough on its own - the caption needs to be set after the issues have been set *and* the filter has been initialised.
In the end it was simpler just to make explicit calls in a couple of places.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.Caption.cs
duncanp-sonar: Yep, it is similar; the main difference apart from readability is test isolation.

PR #2086, opened 2021-01-13
Add overlay to Taint list when solution is not connected to a supported server

### LINKED ISSUE #2063: Add overlay to Taint list when solution is not connected to a supported SQ server edition
#### Description

The user will never see taint issues when the solution is not bound, or when it is bound to a non-developer edition (or higher) of SQ. In these cases, the user will open the taint list but it will remain empty with no obvious explanation how the feature works.

TODO:
- [ ] ~~Add an overlay when the solution is not in connected mode~~
- [ ] ~~Add an overlay when the solution is not connected to a supported SQ edition~~
- [ ] ~~Message for SQ version < 8.6 ?~~
- [ ] ~~Message for no open solution ?~~
- [x] Documentation link 
- [x] General message that no taint issues were found

rita-g-sonarsource:
Fixes #2063 

image

### DIFF
--- a/src/IssueViz.Security/Taint/TaintList/TaintIssuesControl.xaml
+++ b/src/IssueViz.Security/Taint/TaintList/TaintIssuesControl.xaml
@@ -69,9 +69,7 @@
                     <LineBreak/>
                 </TextBlock>
                 <TextBlock Style="{StaticResource OverlayTextBlockStyle}">
-                    <Run>SonarLint Taint Vulnerabilities are available in Connected Mode for SonarCloud and SonarQube version 8.6 or higher.</Run>
-                    <LineBreak/>
-                    <Run>Please ensure that you are connected to a supported server and that taint vulnerabilities exist in your Sonar project.</Run>
+                    <Run>Please ensure that your solution is bound to a Sonar project that contains taint vulnerabilities.</Run>
                 </TextBlock>
                 <TextBlock Style="{StaticResource OverlayTextBlockStyle}">
                     <Hyperlink Command="{Binding ShowDocumentationCommand}">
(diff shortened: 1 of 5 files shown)


src/IssueViz.Security/Taint/TaintList/TaintIssuesControl.xaml
```diff
@@ -36,107 +37,149 @@
...
 
     </UserControl.Resources>
-    
     <Grid UseLayoutRounding="True">
-        <Grid.ColumnDefinitions>
-            <ColumnDefinition Name="Table" Width="*"/>
-        </Grid.ColumnDefinitions>
-        <DataGrid Grid.Column="0" 
+
+        <Grid x:Name="Overlay" Style="{StaticResource OverlayGridStyle}" Panel.ZIndex="1000">
```
rita-g-sonarsource, line 65: As we discussed offline, this should be a more generalized message.
Maybe something like:

```
No Taint Vulnerabilities were found.

Please ensure that your project is bound to a Sonar project that contains taint vulnerabilities.

Click here to go online and find out more about this and other SonarLint features.
```

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -173,5 +187,12 @@ public void Dispose()
             store.IssuesChanged -= Store_IssuesChanged;
             activeDocumentTracker.OnDocumentFocused -= ActiveDocumentTracker_OnDocumentFocused;
         }
+
+        public event PropertyChangedEventHandler PropertyChanged;
```
rita-g-sonarsource, line 191: I'll open a new PR for that

duncanp-sonar (approved):

Issue #2023, opened 2020-12-14
Refactor selection handling so there is a single selection service


duncanp-sonar:
### Description

Note: potential overlap with #1980

The flow window can show flows from normal issues, hotspots (or at least it will when the server-side code is fixed), and taint issues. Similarly, the light build in the editor can set the active selection for normal issues, hotspots (future) and taint issues.

This suggests that either we have one selection service, or one per issue type (or one per type with a singleton aggregator).

Currently there doesn't seem to be a need to have multiple separate services.

- [x] investigate merging the existing selection services
- [x] implement any required changes

PR #2088, opened 2021-01-14
Add single selection service

rita-g-sonarsource:
Part of #2023

### DIFF
--- a/src/IssueViz/Selection/IssueSelectionService.cs
+++ b/src/IssueViz/Selection/IssueSelectionService.cs
@@ -1,0 +1,90 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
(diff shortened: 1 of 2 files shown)


src/IssueViz/Selection/IssueSelectionService.cs
```diff
@@ -0,0 +1,87 @@
...
+                    if (flowStepSelectionService.SelectedIssue != value)
+                    {
+                        var hasSecondaryLocations = value != null && value.Flows.SelectMany(x => x.Locations).Any();
+                        flowStepSelectionService.SelectedIssue = hasSecondaryLocations ? value : null;
+                    }
+                }
+            }
+        }
+
+        private void FlowStepSelectionService_SelectionChanged(object sender, SelectionChangedEventArgs e)
```
rita-g-sonarsource, line 77: We might not need to subscribe to this event after everything is done.
We can update the IssueViz using the old service, so that's why I'm syncing the changes, but in the future I hope that this would be the only service that can set the IssueViz, and the detailed one could only read it, and set the flow/step. When that would be the case, we won't need to listen to this event.

duncanp-sonar: We shouldn't need to - if we do, then there's probably something wrong with the design.
I think it's worth putting a design comment in the code saying this should become redundant.

If I remember correctly, we thought the flow panel only sets the selection to null if the primary span becomes null, which is a hack that should disappear.

It looks like a fair chunk of the tests are about ensure this hacky behaviour still works for now.
I think it would have been ok not to handle this case here, even if it means breaking that particularly hacky scenario in the short term.

duncanp-sonar (approved):

src/IssueViz/Selection/IssueSelectionService.cs
rita-g-sonarsource: Yep I expect all these tests to go away as well. I was debating if I should just skip them, but in the end decided to keep them for now, so we would have an option to delay the fixing the hack if we'd need to.

PR #2094, opened 2021-01-14
Use IssueSelectionService in SelectedVisualizationValidityMonitor

rita-g-sonarsource:
Part of #2023


rita-g-sonarsource:
@duncanp-sonar part of the prep of removing the `Set` from the detailed selection service.

duncanp-sonar (approved):

PR #2096, opened 2021-01-15
Use IssueSelectionService in VSTableEventProcessorProvider

rita-g-sonarsource:
Part of #2023


src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
```diff
@@ -39,12 +40,12 @@ namespace SonarLint.VisualStudio.IssueVisualization.TableControls
     [DataSource(SonarLintTableControlConstants.ErrorListDataSourceIdentifier)]
     internal class VSTableEventProcessorProvider : ITableControlEventProcessorProvider
     {
-        private readonly IIssueTablesSelectionMonitor issueTableSelectionMonitor;
+        private readonly IIssueSelectionService selectionService;
```
rita-g-sonarsource, line 43: The old interface had the exact "Select" functionality and it was also a general purpose one, so I just replaced it with the new one.

src/IssueViz.UnitTests/TableControls/VSTableEventProcessorTests.cs
```diff
@@ -131,48 +130,25 @@ public void SelectionChanged_SingleItem_TryGetValueReturnsTrueButValueIsNotIssue
 
             SimulatePostProcessEvent();
 
-            CheckMonitorCalledOnce(null);
+            CheckSelectionServiceCalledOnce(null);
 
             SanityCheckSnapshotAccessedOnce();
         }
 
         [TestMethod]
```
rita-g-sonarsource, line 138: Simplified the two tests: passing now simply `Mock.Of<IAnalysisIssueVisualization>()`

duncanp-sonar (approved):

PR #2097, opened 2021-01-15
Don't clear secondary locations panel when primary location is non-navigable

rita-g-sonarsource:
Part of #2023

### DIFF
--- a/src/Core/WPF/NullToBooleanConverter.cs
+++ b/src/Core/WPF/NullToBooleanConverter.cs
@@ -1,0 +1,43 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Globalization;
+using System.Windows.Data;
+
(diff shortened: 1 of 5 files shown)


src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs
```diff
@@ -83,16 +83,15 @@ internal sealed class IssueVisualizationViewModel : IIssueVisualizationViewModel
...
         {
             get
             {
                 var issueSpan = CurrentIssue?.Span;
 
                 if (issueSpan == null || issueSpan.Value.IsEmpty)
                 {
-                    selectionService.SelectedIssue = null;
-                    return 0;
+                    return null;
```
rita-g-sonarsource, line 94: As we discussed offline, this is part of the clean-up as preparation for the new selection service.
I opened #2098 to handle the previous behavior.

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -245,13 +250,16 @@
...
                                           Moniker="{Binding Severity, Converter={StaticResource SeverityToMonikerConverter}, Mode=OneWay}" 
                                           Margin="0,0,5,0"/>
-                    <TextBlock Grid.Row="0" 
-                               Grid.Column="1" 
-                               Margin="0,0,0,5"
-                               Style="{StaticResource BaseTitleTextBlockStyle}">
-                        <Run Text="{Binding FileName, Mode=OneWay}"/>
-                        <Run Text="{Binding LineNumber, StringFormat=(line {0}), Mode=OneWay}"/>
-                    </TextBlock>
+                    <StackPanel Grid.Row="0"
```
rita-g-sonarsource, line 253: The `line (x)` part is now conditional, depending if the line number is null.

src/Core/WPF/NullToBooleanConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Globalization;
+using System.Windows.Data;
+
+namespace SonarLint.VisualStudio.Core.WPF
+{
+    /// <summary>
+    /// Converter that returns true when given value is null, and false otherwise.
+    /// </summary>
+    [ValueConversion(typeof(object), typeof(bool))]
+    public class NullToBooleanConverter : IValueConverter
```
rita-g-sonarsource, line 31: I really tried to bind to `LineNumber.HasValue` but it just doesn't work. It doesn't seem to be supported in WPF and I couldn't find anyone doing it.
I also tried using `TargetNullValue` but that didn't work either.
This converter feels completely redundant but I couldn't make it work purely in the xaml.

duncanp-sonar (approved):

PR #2099, opened 2021-01-15
Change IssueSelectionService not to listen to AnalysisIssueSelectionService event

rita-g-sonarsource:
Part of #2023

### DIFF
--- a/src/IssueViz.UnitTests/Selection/IssueSelectionServiceTests.cs
+++ b/src/IssueViz.UnitTests/Selection/IssueSelectionServiceTests.cs
@@ -40,35 +40,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Selection
                     MefTestHelpers.CreateExport<IAnalysisIssueSelectionService>(Mock.Of<IAnalysisIssueSelectionService>())
                 });
         }
-
-        [TestMethod]
-        public void Ctor_RegisterToFlowStepSelectionEvent()
-        {
-            var flowStepSelectionService = new Mock<IAnalysisIssueSelectionService>();
-            flowStepSelectionService.SetupAdd(x => x.SelectionChanged += null);
-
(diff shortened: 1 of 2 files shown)


rita-g-sonarsource:
@duncanp-sonar I couldn't remove the Set from the detailed selection service interface, because the outer service is the one setting it.. https://github.com/SonarSource/sonarlint-visualstudio/blob/branch-taint/src/IssueViz/Selection/IssueSelectionService.cs#L71

I looked into merging the two services into 1, but it becomes a bit messy: I need to store 2 issue vizs, ("SelectedIssue" and "SelectedIssueWithSecondaryLocations") and the code gets a bit hard to follow.
I also tried having 1 class that implements both interfaces but delegates the implementation, e.g.:
```
    [Export(typeof(IIssueSelectionService))]
    [Export(typeof(IAnalysisIssueSelectionService))]
    internal sealed class IssueSelectionService : IIssueSelectionService, IAnalysisIssueSelectionService
    {
        private readonly AnalysisIssueSelectionService flowStepSelectionService;
      
        public IssueSelectionService()
        {
            flowStepSelectionService = new AnalysisIssueSelectionService(); // use the real class and call its Set that is now internal and not part of the interface
        }
```
... but that is messy as well. So I'm not sure. WDYT?

duncanp-sonar:
From the point of view of the "primary" selection, isn't the detailed selection service a sink?
i.e. the detailed selection service should be listening to the events from the "primary" selection service and reacting appropriately. 

The primary selection service is more general; it shouldn't need to know that the more specialised detailed selection service exists.

duncanp-sonar (approved):

rita-g-sonarsource:
> The primary selection service is more general; it shouldn't need to know that the more specialised detailed selection service exists.

Doh! You're right!

duncanp-sonar:
So today's score for obvious design issues spotted by the reviewer is currently 1-1 :-)

PR #2103, opened 2021-01-17
Add secondary locations selection to taint list

### LINKED ISSUE #1977: Vulnerabilities list: set selection in AnalysisIssueSelectionService when vulnerability has secondary locations
#### Description

Set AnalysisIssueSelectionService selection when selected vulnerability has flows. Similar to #1903.

rita-g-sonarsource:
Part of #2023. Fixes #1977.

### DIFF
--- a/src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
@@ -38,6 +38,7 @@ using SonarLint.VisualStudio.IssueVisualization.Security.SharedUI;
 using SonarLint.VisualStudio.IssueVisualization.Security.Taint;
 using SonarLint.VisualStudio.IssueVisualization.Security.Taint.Models;
 using SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.ViewModels;
+using SonarLint.VisualStudio.IssueVisualization.Selection;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint.TaintList
 {
@@ -79,6 +80,18 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint.Tai
             issues.Count.Should().Be(2);
             issues[0].TaintIssueViz.Should().Be(issueViz1);
             issues[1].TaintIssueViz.Should().Be(issueViz3);
+        }
+
+        [TestMethod]
+        public void Ctor_RegisterToSelectionChangedEvent()
+        {
+            var selectionService = new Mock<IIssueSelectionService>();
+            selectionService.SetupAdd(x => x.SelectedIssueChanged += null);
+
+            CreateTestSubject(selectionService: selectionService.Object);
+
+            selectionService.VerifyAdd(x => x.SelectedIssueChanged += It.IsAny<EventHandler>(), Times.Once());
+            selectionService.VerifyNoOtherCalls();
         }
 
         [TestMethod]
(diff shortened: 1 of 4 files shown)


src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -48,6 +49,8 @@ internal interface ITaintIssuesControlViewModel : INotifyPropertyChanged, IDispo
 
         ICollectionView IssuesView { get; }
 
+        ITaintIssueViewModel SelectedIssue { get; set; }
```
rita-g-sonarsource, line 52: This was in the xaml but there was no actual property in the view model to match it:
https://github.com/SonarSource/sonarlint-visualstudio/blob/branch-taint/src/IssueViz.Security/Taint/TaintList/TaintIssuesControl.xaml#L91

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -439,14 +466,155 @@ public void HasServerIssues_IssuesChanged_RaisesPropertyChanged()
...
+        [TestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public void SelectionChanged_SelectedIssueIsNotInList_SelectionSetToNull(bool isSelectedNull)
+        {
+            var selectionService = new Mock<IIssueSelectionService>();
+
+            var testSubject = CreateTestSubject(selectionService: selectionService.Object);
+
+            testSubject.SelectedIssue.Should().BeNull();
```
duncanp-sonar, line 495: This test starts with a null selection and finishes with a null selection, so it doesn't show that the test subject is correctly setting the selection to null - it could be doing nothing at all. The test subject should start with a selection.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -439,14 +466,155 @@ public void HasServerIssues_IssuesChanged_RaisesPropertyChanged()
...
+
+            var selection = new TaintIssueViewModel(Mock.Of<IAnalysisIssueVisualization>());
+            testSubject.SelectedIssue = selection;
+
+            selectionService.VerifySet(x => x.SelectedIssue = selection.TaintIssueViz, Times.Once);
+            selectionService.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public void SetSelectedIssue_ValueIsTheSame_SelectionServiceNotCalled()
```
duncanp-sonar, line 594: Is there a test that the selection service is called when the service changes from (something) -> null?

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -439,14 +466,155 @@ public void HasServerIssues_IssuesChanged_RaisesPropertyChanged()
...
+
+            var selection = new TaintIssueViewModel(Mock.Of<IAnalysisIssueVisualization>());
+            testSubject.SelectedIssue = selection;
+
+            selectionService.Reset();
+
+            testSubject.SelectedIssue = selection;
+
+            selectionService.VerifyNoOtherCalls();
+        }
```
duncanp-sonar, line 607: "One check per test" is a reasonable rule of thumb, but I find in practice there are scenarios it makes the tests harder to read, and it harder to see if all of the appropriate cases have been covered. 
This "selection service only called if selected object has changed" scenario is an example where I'd have considered testing multiple state transitions in the same test.

duncanp-sonar (commented):
Just some comments on the tests.

PR #2106, opened 2021-01-18
Updated webclient to include additional hotspot fields

duncanp-sonar:


duncanp-sonar:
I added a couple of timestamp fields to the hotspot class in the webclient at the end of last year and we haven't updated SLVS since then. This PR SLVS for those hotspot changes. I'll do another update to pull the SonarCloud changes.

rita-g-sonarsource (approved):

rita-g-sonarsource:
> I added a couple of timestamp fields to the hotspot class in the webclient at the end of last year and we haven't updated SLVS since then. This PR SLVS for those hotspot changes. I'll do another update to pull the SonarCloud changes.

Oh damn, I thought we already were on master.

duncanp-sonar:
Those fields are already part of the taint issue, but not for hotspots.

PR #2120, opened 2021-01-20
Change ActiveDocumentFilter to track secondary locations

### LINKED ISSUE #2117: Secondary locations are removed when switching to a location in another file
#### Steps to reproduce:

1. Open a taint issue that has secondary locations in another file.
2. Select the issue.
3. In secondary locations panel, navigate to a location in another file.

Expected ==> location navigated and secondary locations panel is still on.
Actual ==> location navigated but the panel is cleared.

Notes:
This is due to `ActiveDocumentFilter` on the taint issues list.

rita-g-sonarsource:
Fixes #2117


rita-g-sonarsource:
image

@duncanp-sonar We might want to change "File" column to "Sink File", WDYT? Otherwise it's a bit confusing why am I seeing this issue (it becomes clear once you click on it, but until then it looks odd).

duncanp-sonar (approved):
Looks good, just a comment about the test.

gh-devtility:
"Sink file" - not sure. It's not immediately obvious what a "Sink file" is either, as it's not a term we use anywhere else.

It might be better to leave it as "File" and add a tooltip to the column e.g. "File in which tainted data is passed to a sink/passed to a sensitive function/used in an unsafe way"

PR #2123, opened 2021-01-21
Display missing file names in Secondary Locations panel

### LINKED ISSUE #2114: Secondary locations panel doesn't render correctly when it contains files that don't exist
#### Description

When a taint/hotspot refers to a file that doesn't exist, secondary locations panel doesn't show that file name at all.

image

rita-g-sonarsource:
Fixes #2114

### DIFF
--- a/src/IssueViz.Security/Taint/TaintIssueToIssueVisualizationConverter.cs
+++ b/src/IssueViz.Security/Taint/TaintIssueToIssueVisualizationConverter.cs
@@ -53,7 +53,19 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
             var analysisIssue = ConvertToAnalysisIssue(sonarQubeIssue);
             var issueViz = issueVisualizationConverter.Convert(analysisIssue);
 
+            CalculateLocalFilePaths(issueViz);
+
             return issueViz;
+        }
+
+        private void CalculateLocalFilePaths(IAnalysisIssueVisualization issueViz)
+        {
+            var allLocations = issueViz.GetAllLocations();
+
+            foreach (var location in allLocations)
+            {
+                location.CurrentFilePath = absoluteFilePathLocator.Locate(location.Location.FilePath);
+            }
         }
 
         private IAnalysisIssueBase ConvertToAnalysisIssue(SonarQubeIssue sonarQubeIssue)
@@ -63,11 +75,9 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
                 throw new ArgumentNullException(nameof(sonarQubeIssue.TextRange));
             } 
 
-            var localFilePath = absoluteFilePathLocator.Locate(sonarQubeIssue.FilePath);
-
             return new TaintIssue(
                 sonarQubeIssue.IssueKey,
-                localFilePath,
+                sonarQubeIssue.FilePath,
                 sonarQubeIssue.RuleId,
                 sonarQubeIssue.Message,
                 sonarQubeIssue.TextRange.StartLine,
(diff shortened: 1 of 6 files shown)


src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -66,6 +67,7 @@
             <Setter Property="Background" Value="Transparent"/>
             <Setter Property="BorderThickness" Value="0"/>
             <Setter Property="SelectionMode" Value="Single"/>
+            <Setter Property="ScrollViewer.HorizontalScrollBarVisibility" Value="Disabled"/>
```
rita-g-sonarsource, line 70: I noticed that for taint, some locations have a very long text, which causes the "navigability warning" to be less visible.
I removed the scroll so the text would be wrapped.

image

src/IssueViz/IssueVisualizationControl/ViewModels/FileNameLocationListItem.cs
```diff
@@ -68,8 +70,18 @@ private void Location_PropertyChanged(object sender, PropertyChangedEventArgs e)
         private void UpdateState()
         {
             FullPath = location.CurrentFilePath;
-            FileName = Path.GetFileName(FullPath);
-            Icon = GetImageMonikerForFile(FullPath);
+
+            if (FileExists)
```
duncanp-sonar, line 74: If it is really the case that the existence/non-existence of a file can be determined from whether `IAnalysisIssueLocationVisualization.CurrentFilePath" is null or not then the doc comment for that interface method should say so to make it's clear it's part of the interface contract.

src/IssueViz.Security/Taint/TaintIssueToIssueVisualizationConverter.cs
```diff
@@ -53,21 +53,31 @@ public IAnalysisIssueVisualization Convert(SonarQubeIssue sonarQubeIssue)
...
+        private void CalculateLocalFilePaths(IAnalysisIssueVisualization issueViz)
+        {
+            var allLocations = issueViz.GetAllLocations();
+
+            foreach (var location in allLocations)
+            {
+                location.CurrentFilePath = absoluteFilePathLocator.Locate(location.Location.FilePath);
+            }
+        }
+
```
duncanp-sonar, line 70: Presumably we're going to need to make the same change in the hotspot converter when we get round to handling hotspots with secondary locations? If so, it would be useful if we could create a ticket for it so we don't forget.

Or could the work be pushed to the `IAnalysisIssueVisualizationConverter`?

src/IssueViz/IssueVisualizationControl/ViewModels/FileNameLocationListItem.cs
rita-g-sonarsource: I checked what happens for regular issues: we don't support it.
If you delete a file for a secondary location, nothing will change in the panel, and any attempt to navigate will simply become non-navigable.
This is because we don't actually do anything when it happens: https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/IssueViz/Editor/FileRenamesEventSource.cs#L102

So for live issues this is not relevant: CurrentFilePath might not be up-to-date (i.e. if it was deleted), but it will never be null.

For both hotspots and taint, it can definitely be null if the file cannot be found in the solution -- it doesn't mean that the file doesn't exist on disk, just that it wasn't found in the solution (e.g. the project could've been unloaded).

However, I do think that having a null CurrentFilePath makes sense: if we were to support file deletions for live issues, it would be null. So I will change the doc comment to say so. Let me know if you have objections.

src/IssueViz.Security/Taint/TaintIssueToIssueVisualizationConverter.cs
rita-g-sonarsource: It could create a performance hit for `IAnalysisIssueVisualizationConverter` - it is used by live issues where the CurrentFilePath is exactly the same as FilePath (since it was just calculated).
I will create a ticket for hotspots.

duncanp-sonar (approved):

PR #2130, opened 2021-01-22
Invalidate spans of locations in files that could not be resolved

### LINKED ISSUE #2124: Secondary locations for non-existent files are not marked as non-navigable until the user clicks on them
#### Description

When a taint issue has secondary locations in a non-existing file, the secondary locations panel still displays those issues as navigable. They will become non-navigable only when the user clicks on them. 

TODO:
- [ ] Mark all issues of non-existing files as non-navigable.

rita-g-sonarsource:
Fixes #2124

### DIFF
--- a/src/IssueViz.UnitTests/Models/AnalysisIssueLocationVisualizationTests.cs
+++ b/src/IssueViz.UnitTests/Models/AnalysisIssueLocationVisualizationTests.cs
@@ -51,9 +51,63 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Models
         [TestMethod]
         public void Ctor_InitialSpanIsNull()
         {
-            var testSubject = CreateTestSubject();
+            var testSubject = CreateTestSubject(filePath:"test");
 
             testSubject.Span.Should().BeNull();
+        }
+
+        [TestMethod]
+        public void SetCurrentFilePath_FilePathIsNull_SpanIsInvalidated()
+        {
+            // Arrange
+            var oldFilePath = "oldpath.txt";
+            var oldSpan = CreateSpan();
+
+            var testSubject = CreateTestSubject(filePath: oldFilePath);
+            testSubject.Span = oldSpan;
+            testSubject.Span.Should().Be(oldSpan);
+            testSubject.CurrentFilePath.Should().Be(oldFilePath);
+
+            var propertyChangedEventHandler = new Mock<PropertyChangedEventHandler>();
+            testSubject.PropertyChanged += propertyChangedEventHandler.Object;
+
+            // Act
+            testSubject.CurrentFilePath = null;
+
+            // Assert
+            testSubject.Span.Value.IsEmpty.Should().BeTrue();
+            testSubject.CurrentFilePath.Should().BeNull();
+
+            VerifyPropertyChangedRaised(propertyChangedEventHandler, nameof(testSubject.CurrentFilePath));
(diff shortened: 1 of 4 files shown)


src/IssueViz.UnitTests/Models/AnalysisIssueLocationVisualizationTests.cs
```diff
@@ -51,11 +51,65 @@ public void Ctor_InitialFilePathIsTakenFromIssueLocation()
         [TestMethod]
         public void Ctor_InitialSpanIsNull()
         {
-            var testSubject = CreateTestSubject();
+            var testSubject = CreateTestSubject(filePath:"test");
 
             testSubject.Span.Should().BeNull();
         }
 
+        [TestMethod]
```
rita-g-sonarsource, line 59: Very blatant copy-paste of the tests in AnalysisIssueVisualizationTests...

src/IssueViz.UnitTests/Models/AnalysisIssueLocationVisualizationTests.cs
```diff
@@ -51,11 +51,65 @@ public void Ctor_InitialFilePathIsTakenFromIssueLocation()
         [TestMethod]
         public void Ctor_InitialSpanIsNull()
         {
-            var testSubject = CreateTestSubject();
+            var testSubject = CreateTestSubject(filePath:"test");
```
rita-g-sonarsource, line 54: It was previously null, which according to the logic, will now set the span to Empty. This way the test checks the previous behavior, of initial state when the file name is not null.

src/IssueViz/Models/AnalysisIssueLocationVisualization.cs
```diff
@@ -85,6 +85,11 @@ public string CurrentFilePath
             {
                 filePath = value;
                 NotifyPropertyChanged();
+
+                if (string.IsNullOrEmpty(filePath))
+                {
+                    this.InvalidateSpan();
```
duncanp-sonar, line 91: `InvalidateSpan` is also going to raise a change notification. Wouldn't it be better to make all the state changes first before any of the notifications are raised so that the properties are in a consistent state when any listeners are notified?

i.e. move the notification in line 87 to after this new block of code, so the sequence of operations is
1. set filePath
2. set span
3. notify span
4. notify filePath

src/IssueViz/Models/AnalysisIssueVisualization.cs
```diff
@@ -70,6 +70,11 @@ public string CurrentFilePath
             {
                 currentFilePath = value;
                 NotifyPropertyChanged();
+
+                if (string.IsNullOrEmpty(currentFilePath))
+                {
+                    this.InvalidateSpan();
+                }
```
duncanp-sonar, line 77: Same comment about notification order applies here.

src/IssueViz.UnitTests/Models/AnalysisIssueLocationVisualizationTests.cs
duncanp-sonar: It's almost as if the class structure is wrong, and `IAnalysisIssueVisualization` should _have_ an `IAnalysisIssueLocationVisualiztion` rather than _be_ an `IAnalysisIssueLocationVisualization`...

rita-g-sonarsource: https://github.com/SonarSource/sonarlint-visualstudio/issues/2132

gh-devtility (approved):

PR #2133, opened 2021-01-25
Support setting Error List selection for C# issues

### LINKED ISSUE #2125: Secondary locations panel is not cleared when selecting a C# issue in the Error List
#### Steps to reproduce:

1. Open a C# file that has a taint issue and a regular issue.
2. Select a taint issue from the taint list -> secondary locations panel is populated
3. Go to the error list and select the regular issue.

Expected ==> secondary locations panel is cleared.
Actual ==> no changes to the panel.

Notes:
This works correctly for c++ issues.
For c#, `VSTableEventProcessorProvider` is not being called.

rita-g-sonarsource:
Fixes #2125

### DIFF
--- a/src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
+++ b/src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
@@ -37,7 +37,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.TableControls
     // We can specify multiple data source types and data source i.e. the same provider can be used for
     // multiple table controls/sources
     [DataSourceType(StandardTableDataSources.ErrorTableDataSource)]
-    [DataSource(SonarLintTableControlConstants.ErrorListDataSourceIdentifier)]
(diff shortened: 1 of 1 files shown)


src/IssueViz/TableControls/VSTableEventProcessorProvider.cs
```diff
@@ -37,7 +37,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.TableControls
     // We can specify multiple data source types and data source i.e. the same provider can be used for
     // multiple table controls/sources
     [DataSourceType(StandardTableDataSources.ErrorTableDataSource)]
-    [DataSource(SonarLintTableControlConstants.ErrorListDataSourceIdentifier)]
+    [DataSource(StandardTableDataSources.AnyDataSource)]
```
duncanp-sonar, line 40: Would be worth commenting why we need to listen to other data sources.

duncanp-sonar (approved):

PR #2134, opened 2021-01-25
Support regional datetime settings in Taint List

### LINKED ISSUE #2126: Taint list creation date ignores machine regional format settings
#### Description

The taint list contains `Created` column that displays the creation date of the taint issue. However, this column ignores the regional format overrides:

image
image

Notes:
* See also https://github.com/dotnet/wpf/issues/1946#issuecomment-534564980 and https://github.com/microsoft/WindowsTemplateStudio/issues/3667
* It might be easier to just use a verbal format such as F https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#FullDateLongTime

rita-g-sonarsource:
Fixes #2126

### DIFF
--- a/src/Core/WPF/DateTimeConverter.cs
+++ b/src/Core/WPF/DateTimeConverter.cs
@@ -24,19 +24,23 @@ using System.Windows.Data;
 
 namespace SonarLint.VisualStudio.Core.WPF
 {
+    /// <summary>
+    /// Workaround for https://github.com/dotnet/wpf/issues/1946
+    /// Fixes SLVS #2126
+    /// </summary>
     [ValueConversion(typeof(DateTime), typeof(string))]
     public class DateTimeConverter : IValueConverter
     {
-        private readonly Func<CultureInfo> getMachineCulture;
+        private readonly CultureInfo userCulture;
 
         public DateTimeConverter()
-            : this(() => CultureInfo.CurrentUICulture)
+            : this(CultureInfo.CurrentUICulture)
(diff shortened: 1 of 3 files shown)


src/Core/WPF/DateTimeConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+using System;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace SonarLint.VisualStudio.Core.WPF
+{
+    [ValueConversion(typeof(DateTime), typeof(string))]
+    public class DateTimeConverter : IValueConverter
```
duncanp-sonar, line 28: It's not obvious why this converter is required. A comment linking to the MS bug would be useful.

src/Core/WPF/DateTimeConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        private readonly Func<CultureInfo> getMachineCulture;
+
+        public DateTimeConverter()
+            : this(() => CultureInfo.CurrentUICulture)
+        {
+        }
+
+        internal DateTimeConverter(Func<CultureInfo> getMachineCulture)
```
duncanp-sonar, line 37: Why does this parameter need to be a function?

rita-g-sonarsource: It doesn't; changed to a regular obj.

PR #2136, opened 2021-01-29
Add disclaimer to taint list

### LINKED ISSUE #2062: Add disclaimer to Taint list that the issues come from the server
#### Description

In order to avoid confusion with the live issues, the taint list should display a note to the user, stating that the issues are coming from a remote server and are not analyzed locally.

TODO:
- [ ] Add a note, similar to the one in SLI
- [ ] Add a help link to the wiki?
- [ ] Consider including information about the analysis results that are being displayed eg. _29 issues detected in branch foo, analyzed on 01/06/2020. Click here to see the results in SonarQube/SonarCloud._

rita-g-sonarsource:
Fixes #2062

### DIFF
--- a/src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
@@ -19,7 +19,6 @@
  */
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using FluentAssertions;
@@ -75,7 +74,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
 
             await testSubject.SynchronizeWithServer();
 
-            taintStore.Verify(x => x.Set(Enumerable.Empty<IAnalysisIssueVisualization>()), Times.Once());
+            taintStore.Verify(x => x.Set(Enumerable.Empty<IAnalysisIssueVisualization>(), null), Times.Once());
 
             sonarQubeServer.Invocations.Count.Should().Be(0);
             converter.Invocations.Count.Should().Be(0);
@@ -86,8 +85,9 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
         {
             var converter = new Mock<ITaintIssueToIssueVisualizationConverter>();
             var taintStore = new Mock<ITaintStore>();
+            var sonarQubeService = CreateSonarQubeService(isConnected: false);
 
-            var testSubject = CreateTestSubject(taintStore.Object, converter.Object, isServerConnected: false);
+            var testSubject = CreateTestSubject(taintStore.Object, converter.Object, sonarService: sonarQubeService.Object);
             await testSubject.SynchronizeWithServer();
 
             converter.Invocations.Count.Should().Be(0);
@@ -95,7 +95,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
         }
 
         [TestMethod]
-        public async Task SynchronizeWithServer_NoIssues_StoreSynced()
+        public async Task SynchronizeWithServer_ConnectedModeWithNoIssues_StoreCleared()
         {
             var converter = new Mock<ITaintIssueToIssueVisualizationConverter>();
             var taintStore = new Mock<ITaintStore>();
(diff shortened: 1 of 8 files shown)


src/IssueViz.Security/Taint/TaintList/TaintIssuesControl.xaml
```diff
@@ -83,105 +94,119 @@
                 </TextBlock>
             </StackPanel>
         </Grid>
-
-        <Grid x:Name="TaintList">
-            <Grid.ColumnDefinitions>
-                <ColumnDefinition Name="Table" Width="*"/>
-            </Grid.ColumnDefinitions>
-            <DataGrid Grid.Column="0" 
+        <Grid x:Name="Content">
```
rita-g-sonarsource, line 97: image

I looked into adding a yellow-ish background like in SLI, but it looked too much like a goldbar.

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -100,16 +100,24 @@ public async Task SynchronizeWithServer()
...
-                UpdateTaintIssuesUIContext(hasTaintIssues);
-                if (hasTaintIssues)
+                if (!hasTaintIssues)
+                {
+                    ClearStore();
+                    UpdateTaintIssuesUIContext(false);
+                }
+                else
                 {
+                    var analysisInformation = await GetAnalysisInformation(projectKey);
```
rita-g-sonarsource, line 115: Fetching the project branches only if we really have issues.
If we don't, then we won't show the taint window anyway, and it will contain the empty overlay. So no point in fetching the information.

src/IssueViz.Security.UnitTests/Taint/TaintStoreTests.cs
```diff
@@ -60,7 +60,7 @@ public void Set_NullCollection_ArgumentNullException()
         {
             var testSubject = CreateTestSubject();
 
-            Action act = () => testSubject.Set(null);
+            Action act = () => testSubject.Set(null, null);
```
rita-g-sonarsource, line 63: All the existing tests check the issues collection, the new property is standalone so I just passed null everywhere and added new dedicated tests.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -455,29 +455,24 @@ public void HasServerIssues_IssuesChanged_RaisesPropertyChanged()
             var store = new Mock<ITaintStore>();
             var testSubject = CreateTestSubject(store: store);
 
-            var callCount = 0;
-            PropertyChangedEventArgs suppliedArgs = null;
-            testSubject.PropertyChanged += (sender, args) => { callCount++; suppliedArgs = args; };
+            var eventHandler = new Mock<PropertyChangedEventHandler>();
```
rita-g-sonarsource, line 458: suppliedArgs was no longer suitable; it was capturing the first NotifyPropertyChanged, and now we are raising two (one for the new property and one for the old `HasServerIssues`).

src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
```diff
@@ -306,14 +323,12 @@ public async Task SynchronizeWithServer_ConnectedModeWithIssues_UIContextIsSetAn
                 toolWindowService, serviceProvider, logger);
         }
 
-        private static Mock<ISonarQubeService> CreateSonarServerMock(bool isServerConnected, params SonarQubeIssue[] serverIssues)
```
rita-g-sonarsource, line 309: I decided to remove this method and instead use the explicit `CreateSonarQubeService`, `SetupTaintIssues`, `SetupAnalysisInformation`.
The method would have had another optional parameter (analysis information) and it just became too complicated to read.

src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
```diff
@@ -358,5 +373,25 @@ public TestSonarQubeIssue()
             {
             }
         }
+
+        private void SetupAnalysisInformation(Mock<ISonarQubeService> sonarQubeService, AnalysisInformation analysisInformation)
+        {
+            var projectBranches = new[]
```
rita-g-sonarsource, line 379: Creating 3 branches: 2 dummy non-main and 1 main.

src/IssueViz.Security/Taint/TaintList/TaintIssuesControl.xaml
duncanp-sonar: Looks good.
I think it would better to have the branch name in quotes or italics - something to make it stand out more from the boiler plate text.

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -123,9 +131,17 @@ public async Task SynchronizeWithServer()
             }
         }
 
+        private async Task<AnalysisInformation> GetAnalysisInformation(string projectKey)
+        {
+            var branches = await sonarQubeService.GetProjectBranchesAsync(projectKey, CancellationToken.None);
+            var issuesBranch = branches.First(x => x.IsMain);
```
duncanp-sonar, line 137: How defensive do we need to be about bugs on the SQ/SC side?
If we throw an exception here, will it bring down VS, or just give a gold bar?

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -33,7 +33,13 @@ internal interface ITaintStore : IIssuesStore
         /// Removes all existing visualizations and initializes the store to the given collection.
         /// Can be called multiple times.
         /// </summary>
-        void Set(IEnumerable<IAnalysisIssueVisualization> issueVisualizations);
+        void Set(IEnumerable<IAnalysisIssueVisualization> issueVisualizations, AnalysisInformation analysisInformation);
+
+        /// <summary>
+        /// Returns additional analysis information for the existing visualizations in the store.
+        /// </summary>
+        /// <returns></returns>
```
duncanp-sonar, line 41: Empty <returns>

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
duncanp-sonar: The user can still manually open the window. What will it show in that case?

src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
```diff
@@ -358,5 +373,25 @@ public TestSonarQubeIssue()
             {
             }
         }
+
+        private void SetupAnalysisInformation(Mock<ISonarQubeService> sonarQubeService, AnalysisInformation analysisInformation)
```
duncanp-sonar, line 377: It would be clearer if this parameter was called something like `mainBranchInfo`

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
rita-g-sonarsource: No, it will contain an empty overlay.

duncanp-sonar (approved):
Looks good, just a question about the use of `First`.

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
rita-g-sonarsource: No need; the entire block is still covered with a try-catch, so we should be ok (I threw an exception and it was fine).

Also, if the user opens the window in an unbound project, we won't reach this code at all (since there are no issues).
So we could fail only if there _are_ issues, but for some reason there is no main branch. In which case, we will get to the catch block and log the exception.

src/IssueViz.Security/Taint/TaintList/TaintIssuesControl.xaml
rita-g-sonarsource: I went with single quotes, we can change after dogfood feedback

Issue #2131, opened 2021-01-23
Investigate adding help links to C and C++ editor tooltips


rita-g-sonarsource:
#### Description

Editor tooltips for C# contain help links that lead to rules.sonarsource.com. C and C++ issues do not contain a link, which doesn't allow users to understand more about the issue. Community thread: https://community.sonarsource.com/t/sonarlint-no-effect/37359/12

C# tooltip:
![image](https://user-images.githubusercontent.com/60586879/105564964-c62d5e80-5d24-11eb-8be6-af328f425b44.png)

C and C++ tooltip:
![image](https://user-images.githubusercontent.com/60586879/105564978-dba28880-5d24-11eb-9766-101fb2557682.png)

Workaround:
The user can see the help link in the error list
![image](https://user-images.githubusercontent.com/60586879/105564991-e8bf7780-5d24-11eb-9c34-f248078c9cc1.png)

Related issue: #2054

TODO:
- [ ] Investigate adding tooltips for C and C++ issues.

PR #2137, opened 2021-01-29
Add help link to editor tooltip

### LINKED ISSUE #2054: Editor tooltip or taint issue does not contain a help link
#### Description

The tool tip that appears in the editor for security issues does not contain a link to the rule help.
_cf_ the appearance of "normal" (non-taint) SonarC# issues:

image

rita-g-sonarsource:
Fixes #2054 and #2131

image


src/Infrastructure.VS/IBrowserService.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// Testable wrapper for <see cref="VsShellUtilities.OpenBrowser(string)"/>
+    /// </summary>
+    public interface IBrowserService
+    {
+        void Navigate(string url);
+    }
+
+    [Export(typeof(IBrowserService))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class BrowserService : IBrowserService
```
rita-g-sonarsource, line 36: This could be later used in 2 additional places: `IShowInBrowserService` and `IssueVisualizationControl.xaml.cs`

src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// Creates a tooltip object for an IErrorTag
+    /// </summary>
+    internal interface IErrorTagTooltipProvider
+    {
+        object Create(IAnalysisIssueBase analysisIssueBase);
+    }
+
+    [Export(typeof(IErrorTagTooltipProvider))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class ErrorTagTooltipProvider : IErrorTagTooltipProvider
```
rita-g-sonarsource, line 43: If (and when) we drop VS 2015, we can do what roslyn do:
https://github.com/dotnet/roslyn/blob/afd10305a37c0ffb2cfb2c2d8446154c68cfa87a/src/EditorFeatures/Core/Implementation/Diagnostics/AbstractDiagnosticsAdornmentTaggerProvider.cs#L69

Unfortunately those classes are not available in 2015 and I couldn't find anything else.

src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+            this.browserService = browserService;
+            this.vsThemeColorProvider = vsThemeColorProvider;
+            this.ruleHelpLinkProvider = ruleHelpLinkProvider;
+        }
+
+        public object Create(IAnalysisIssueBase analysisIssueBase)
+        {
+            var ruleUrl = ruleHelpLinkProvider.GetHelpLink(analysisIssueBase.RuleKey);
+
+            var hyperLink = new Hyperlink
```
rita-g-sonarsource, line 66: The end result looks ok, but for the life of me - I couldn't add a style that adds an underline when hovering over the link (which is the default VS behavior for the built in tooltips).
So the link is blue but there is no hover effect. I've spent an hour on this and in the ended I decided to drop it.
Let me know if you can think of a way.

src/Infrastructure.VS/IBrowserService.cs
duncanp-sonar: The name is confusingly similar to `ShowInBrowserService`. 
`IVSBrowserService` might be clearer.

src/Infrastructure.VS/IBrowserService.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        public void Navigate(string url)
+        {
+            VsShellUtilities.OpenBrowser(url);
```
duncanp-sonar, line 40: Elsewhere we're calling `VsShellUtilities.OpenSystemBrowser`. I'm not sure it matters too much which one we pick, but we should be consistent.

src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs
duncanp-sonar: I think it's good enough to ship as it is. 

What did you try? I'd have thought using a style trigger bound to the `IsMouseOver` property of the hyperlink would have worked, perhaps with two versions of the `hyperlink.Inlines`, one with an `Underline` run and one without.

duncanp-sonar (approved):

src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs
rita-g-sonarsource: That's what I tried, it doesn't work. I'll leave it as-is.

PR #2141, opened 2021-02-01
Simplified flow combo styling

duncanp-sonar:


src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -242,159 +242,13 @@
...
-                        <ControlTemplate.Triggers>
-                            <Trigger SourceName="Popup" Property="Popup.AllowsTransparency" Value="true">
-                                <Setter TargetName="DropDownBorder" Property="CornerRadius" Value="0"/>
-                                <Setter TargetName="DropDownBorder" Property="Margin" Value="0,2,0,0"/>
-                            </Trigger>
-                        </ControlTemplate.Triggers>
-                    </ControlTemplate>
-                </Setter.Value>
-            </Setter>
+        <Style x:Key="FlowsComboBox" TargetType="{x:Type ComboBox}" BasedOn="{StaticResource {x:Static vsShell:VsResourceKeys.ComboBoxStyleKey}}">
```
duncanp-sonar, line 245: I think this is the missing piece we were looking for: `vsShell:VsResourceKeys` has a number of styles defined.

rita-g-sonarsource: Nicely done; I hope we can apply it to the datagrid and context menu controls as well (that we use in hotspots and taint).

rita-g-sonarsource:
@duncanp-sonar I took a look at the tests that are failing -- I don't think we should attempt to fix the style problem. Instead, it's a good opportunity to move all the commands that are in `IssueVisualizationControl` to the VM, like we do in hotspots and taint (https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/IssueViz.Security/HotspotsList/ViewModels/HotspotsControlViewModel.cs#L96).
The only test that will remain is the dispose one, which is not that important anyway since the dispose will occur only on VS shutdown.

PR #2143, opened 2021-02-02
[2] Add Secondary Locations panel commands

rita-g-sonarsource:


rita-g-sonarsource:
I decided not to pollute the view model with the commands' logic (unlike in taint/hotspots) and instead created dedicated classes with tests. It will also allow them to be reusable.

duncanp-sonar (approved):

PR #2144, opened 2021-02-02
[3] Replace code-behind logic with VM commands

rita-g-sonarsource:


src/IssueViz.UnitTests/IssueVisualizationControl/IssueVisualizationViewModelTests.cs
```diff
@@ -347,41 +347,6 @@ public void RuleKey_CurrentIssueHasAnalysisIssue_IssueRuleKey()
 
         #endregion
 
-        #region RuleHelpLink
```
rita-g-sonarsource, line 350: No longer needed; it was used in the hyperlink

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -127,8 +127,6 @@
         <!-- Data grid items styles -->
 
         <Style x:Key="IssueDescriptionBorderStyle" TargetType="Border">
-            <EventSetter Event="MouseLeftButtonDown" Handler="IssueDescription_OnMouseLeftButtonDown"/>
```
rita-g-sonarsource, line 130: Replaced with command

src/IssueViz/IssueVisualizationControl/IssueVisualizationControl.xaml
```diff
@@ -149,8 +147,6 @@
         </Style>
 
         <Style x:Key="LocationsListViewItemStyle" TargetType="ListViewItem">
-            <EventSetter Event="PreviewMouseLeftButtonDown" Handler="LocationsList_OnMouseLeftButtonDown"/>
```
rita-g-sonarsource, line 152: It was actually never needed; clicking or navigating in the list also sets the selected item, which automatically navigates to the location https://github.com/SonarSource/sonarlint-visualstudio/blob/cb70ace4f1784c42e45f9586bf0aa9d5b9c19b47/src/IssueViz/IssueVisualizationControl/ViewModels/IssueVisualizationViewModel.cs#L184

duncanp-sonar (approved):

Issue #1714, opened 2020-09-15
Fix Visualization tool window auto show/hide + gold bar


rita-g-sonarsource:
Make sure that the auto show/hide works + there is no goldbar

![image](https://user-images.githubusercontent.com/60586879/109790202-74280480-7c11-11eb-8860-c7aaa5c003bd.png)

Clicking on the `Manage performance` button shows the following dialogue:

![image](https://user-images.githubusercontent.com/10757559/109847749-79a04180-7c47-11eb-9eca-a8861a87dd90.png)

#### Occurrence and impact
The tool window will appear/disappear automatically based on the state of the UIContext as long as the user does not do a manual action that "breaks" the tracking. Once the tracking is broken it will stay broken until the user takes a manual action to restore it e.g.
* using the menu command to show the tool window when the UIContext is active, or
* manually closing the tool window when the UIContext is not active.

The only scenario in which the gold bar will appear is if tracking is a broken, and the tool window was open when VS shuts down.
In that case, the tool window will be shown again on startup, potentially leading to the appearance of the gold bar.

However, this appearance of this gold bar is not a serious as the gold bars which suggest that the offending extension should be disable. This gold bar merely effects the specific tool window. The gold bar can also appear for built-in VS tool windows.

#### Interaction between ToolWindowVisibility and "Transient" setting on ToolWindow attribute
The ToolWindow attribute has a "Transient" boolean property. If set to true, the tool window will never be open at startup, even if it was open when VS closed.

However, the "transient" setting appears to be ignored if the ToolWindowVisibility attribute is also present.

duncanp-sonar:
Closing as won't fix. The occurrence should be low, the impact is relatively low (in that the user is not prompted to disable the whole extension), and there is isn't an simple, reliable fix.

PR #2145, opened 2021-02-02
Remove redundant dispose method from IssueVisualizationToolWindow

rita-g-sonarsource:


src/IssueViz/IssueVisualizationControl/IssueVisualizationToolWindow.cs
```diff
@@ -65,18 +65,5 @@ public IssueVisualizationToolWindow(IServiceProvider serviceProvider) : base(nul
 
             Content = new IssueVisualizationControl(viewModel);
         }
-
-        protected override void Dispose(bool disposing)
```
rita-g-sonarsource, line 69: We know that this doesn't work (#1714) and the tests are causing the style branch to fail unnecessarily.

gh-devtility (approved):

duncanp-sonar (approved):

PR #2150, opened 2021-02-05
Clear issue selection when live issue's span becomes non navigable

### LINKED ISSUE #2098: Clear secondary locations panel when live issue's primary location becomes non-navigable
#### Description

Previously, when a live analysis issue had secondary locations and the primary location was removed, the secondary locations panel was cleared and the tags were removed.
Following #2023 that is not the case anymore: the panel is not cleared when the primary location is removed, and the tags remain in place. This is because for hotspots and taint issues, the panel should NOT be cleared.

TODO:
- [ ] Decide on the UX for live issues.

rita-g-sonarsource:
Fixes #2098

### DIFF
--- a/src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_IssueLocationStoreTests.cs
+++ b/src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_IssueLocationStoreTests.cs
@@ -219,8 +219,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.ErrorList
         public void LocationsUpdated_SelectedIssueIsNotInFactory_SelectedIssueNotChanged(bool hasSelectedIssue)
         {
             var selectedIssue = hasSelectedIssue ? Mock.Of<IAnalysisIssueVisualization>() : null;
-            var selectionService = new Mock<IIssueSelectionService>();
-            selectionService.Setup(x => x.SelectedIssue).Returns(selectedIssue);
+            var selectionService = CreateSelectionService(selectedIssue);
 
             var factory = CreateFactoryWithIssues(Mock.Of<IAnalysisIssueVisualization>());
             var testSubject = CreateTestSubject(selectionService.Object);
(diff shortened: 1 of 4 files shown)


src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -207,11 +212,19 @@ public void Refresh(IEnumerable<string> affectedFilePaths)
...
                 {
                     var snapshot = factory.CurrentSnapshot;
+                    
                     if (snapshot.FilesInSnapshot.Any(snapshotPath => affectedFilePaths.Any(affected => PathHelper.IsMatchingPath(snapshotPath, affected))))
                     {
                         snapshot.IncrementVersion();
                         InternalRefreshErrorList(factory);
                     }
+
+                    var selectedIssue = issueSelectionService.SelectedIssue;
```
rita-g-sonarsource, line 222: Disclaimer: I think this is the wrong way to fix this.

The real fix should be deeper and get rid of the following places as well:
* Hiding the issue in the error list when it is not navigable https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/ErrorList/IssuesSnapshot.cs#L159
* Clearing the selection when an issue is removed https://github.com/SonarSource/sonarlint-visualstudio/blob/branch-taint/src/IssueViz/Selection/SelectedVisualizationValidityMonitor.cs#L64
* This is now the 3rd "patch": clearing the selection when the span is not navigable, which is unique to live issues and does not apply to hotspots/taint.

I think the full fix should be somewhere in the `IssuesSnapshot` domain:
1. Register to NotifyPropertyChanged of the issue visualizations
2. When a span changes:
	2.1 If the span is not navigable: 
		2.1.1 Remove it from the issues list
		2.1.2 Clear the selection service if this issue was selected
	2.2 Increment the snapshot version and notify the error list (which makes the Refresh method on the `IIssueLocationStore` redundant)
3. When the snapshot is disposed -> clear the selection service if an issue in the snapshot was selected

This will allow us to drop all these patches and will centralize the Error List behavior, which is removing the issues when they are not navigable (as opposed to the taint/hotspots where we keep them). We could then also drop the Validity Monitor.

I didn't make this change now A) because we didn't talk about it yet, and B) I spotted this regression ticket and wanted to have a quick fix  proposal for this release.

duncanp-sonar: I agree this problem is localised to live issues in the Error List so the fix should be contained within the classes that handle that feature. However, I don't think the snapshot should know anything about selection per se.

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_IssueLocationStoreTests.cs
```diff
@@ -211,6 +213,62 @@ public void LocationsUpdated_FactoriesWithMatchingFilesUpdatedAndErrorListRefres
...
+            var selectionService = new Mock<IIssueSelectionService>();
+            selectionService.Setup(x => x.SelectedIssue).Returns(selectedIssue.Object);
+
+            var factory = CreateFactoryWithIssues(selectedIssue.Object);
+            var testSubject = CreateTestSubject(selectionService.Object);
+            testSubject.AddFactory(factory);
+
+            testSubject.Refresh(new[] { "test.txt" });
+
+            selectionService.VerifySet(x => x.SelectedIssue = null, Times.Once);
```
duncanp-sonar, line 269: Minor: the repeated setup boilerplate makes it harder obscures the intent of the test. A helper function would make the intent clearer e.g.

```
var selectionService = CreateSelectionService(selectedIssueViz: null);
...
var selectionService = CreateSelectionService(selectedIssueViz: NaviagableIssueViz);
...
var selectionService = CreateSelectionService(selectedIssueViz: NonNavigableIssueViz);
```

PR #2152, opened 2021-02-05
[Merge to master] Re-use NavigateToDocumentationCommand 

rita-g-sonarsource:
Cleanup


src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -46,7 +47,7 @@ internal interface ITaintIssuesControlViewModel : INotifyPropertyChanged, IDispo
 
         ICommand ShowInBrowserCommand { get; }
 
-        ICommand ShowDocumentationCommand { get; }
+        INavigateToDocumentationCommand ShowDocumentationCommand { get; }
```
duncanp-sonar, line 50: Does the view model need to know that the command is an 'INavigateToDocumentationCommand` rather than just and `ICommand`?

duncanp-sonar (approved):

rita-g-sonarsource:
Replaced with #2162

Issue #2147, opened 2021-02-04
[Telemetry] Report the VS major version and update number


duncanp-sonar:
### Description

Currently we are reporting the file version of `msenv.exe` as the VS version. However, we are not confident that this can be accurately mapped to a particular VS update.

TODO: in addition, report the VS major version and update numbers.

We should be able to retrieve these values from a VS service. A good starting point would be to look at the service used by the "Open in IDE" service to return a user-friendly description of the running VS instance.

PR #2163, opened 2021-02-09
Add VsVersionProvider

rita-g-sonarsource:
Part of #2147


src/Infrastructure.VS/IVsVersionProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        /// <summary>
+        /// See <see cref="__VSSPROPID5.VSSPROPID_AppShortBrandName"/>
+        /// </summary>
+        string ShortName { get; }
+
+        /// <summary>
+        /// See <see cref="__VSSPROPID5.VSSPROPID_ReleaseVersion"/>
+        /// </summary>
+        string BuildVersion { get; }
```
rita-g-sonarsource, line 67: I investigated further the values that are returned from this property.

* In 2015 it's `14.0.25420.1 D14REL` when the about window says `14.0.25431.01 update 3`
* In 2017 it's `15.0.28307.1259 D15.9` when the about window says `15.9.27`
* In 2019 it's `16.0.30914.41 D16.9` when the about window says `Version 16.9.0 Preview 3.0`.

I was able to cross reference the number with the release versions of [2019](https://docs.microsoft.com/en-us/visualstudio/install/visual-studio-build-numbers-and-release-dates?view=vs-2019) and [2017](https://docs.microsoft.com/en-us/visualstudio/install/visual-studio-build-numbers-and-release-dates?view=vs-2017&preserve-view=true) (couldn't find for 2015). 
Notice that the minor field does not match: in both 2019 and 2017 it seems to always be zero:
* `16.0.30914.41 D16.9` matches `16.9.30914.41`
* `15.0.28307.1259 D15.9` matches `15.9.28307.1259`

I tried to cross-reference it with the values that we already have in telemetry -- those don't match. Some of them do, but for instance the telemetry reports `16.0.30907.39` which is nowhere to be found in the release table. So I think that what we're currently sending to the telemetry is definitely not enough and we should be sending this field instead.

The question is: what should we send to telemetry? 
I suppose it would be ok to send the field as-is; in most cases we would be interested in knowing "who is below update x" -- which would require us to go to the release notes, get the number, and then filter telemetry for anyone below that number.
Alternatively, we could "patch" this field and take the minor version from whatever is reported in the last part (e.g. the `D15.9` from `15.0.28307.1259 D15.9`). But that feels kinda hacky.

I think we should just send it as-is. WDYT?

src/Infrastructure.VS/IVsVersionProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    internal class VsVersion : IVsVersion
+    {
+        public string Name { get; }
```
rita-g-sonarsource, line 72: `Name` could be used in the OpenInIDE code.
`ShortName` -- not sure if we would need it. I was thinking of sending that too to the telemetry, for further filtering. Example values: `VS Community 2015`, `VS Community 2017`, `VS Enterprise 2019`

src/Infrastructure.VS/IVsVersionProvider.cs
duncanp-sonar: That's very irritating. At least there are official pages that list the versions for VS2017 and 2019, but it's still going to be a bit of a pain to have to cross-reference them.

I'll have a quick look to see if I can find an alternative, but I'm not expecting to find anything you haven't.

duncanp-sonar (approved):
Looks ok. 
As discussed offline, I think it's worth checking how easy it would be to use `ISetupConfiguration` and whether it returns "cleaner" data. I think that is now the "correct" way to return information about the product used by the installer etc.

rita-g-sonarsource:

src/Infrastructure.VS/VsVersion/IVsVersion.cs
```diff
@@ -0,0 +1,56 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Infrastructure.VS.VsVersion
+{
+    public interface IVsVersion
```
duncanp-sonar, line 23: Do any of the interfaces need to be in the `Infrastructure.VS` assembly? They don't expose any VS-specific types.

src/Infrastructure.VS/VsVersion/IVsVersionProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                logger.WriteLine(Resources.VsVersionDetails,
+                    productName,
+                    productVersion,
+                    productDisplayVersion);
+
+                return new VsVersion(productName, productVersion, productDisplayVersion);
+            }
+            catch (Exception ex)
+            {
+                logger.LogDebug(Resources.FailedToCalculateVsVersion, ex);
```
duncanp-sonar, line 91: I don't think we want to dump the whole exception here; `ex.Message` would be enough.

src/Infrastructure.VS/VsVersion/IVsVersionProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        internal VsVersionProvider(IServiceProvider serviceProvider, ISetupConfiguration setupConfiguration, ILogger logger)
+        {
+            this.setupConfiguration = setupConfiguration;
+            this.logger = logger;
+            vsShell = serviceProvider.GetService(typeof(SVsShell)) as IVsShell;
+        }
+
+        IVsVersion IVsVersionProvider.TryGet()
+        {
+            version ??= CalculateVersion();
```
duncanp-sonar, line 64: Wouldn't it be simpler just to calculate the value once on construction? We don't expect the call to succeed on a subsequent call if it fails first time.

Also, for VS2015 the version will always be null, so with the current implementation it will try and fail to calculate the version each time the method is called.

Currently, that probably doesn't matter too much if the only caller is the telemetry manager which will only call the method once per day. However, this class shouldn't know anything about how it is used.

src/Infrastructure.VS/Resources.resx
```diff
@@ -0,0 +1,129 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="FailedToCalculateVsVersion" xml:space="preserve">
+    <value>Failed to calculate VS version, exception details: {0}</value>
```
duncanp-sonar, line 121: Most error messages are formatted as e.g. `Failed to calculate VS version. Error: {0}`

src/Infrastructure.VS.UnitTests/VsVersionProviderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            var vsVersion2 = testSubject.TryGet();
+
+            vsVersion1.Should().BeSameAs(vsVersion2);
+
+            vsShell.Invocations.Should().BeEmpty();
+            setupConfig.Invocations.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void TryGet_FailureToGetVersion_Null()
```
duncanp-sonar, line 92: I guess this is meant to test that the exception handler is hit. However, it's quite fragile as it's implicitly dependant on both the current test setup code and the product code implementation. It would be possible to change either inadvertently so the method returns null without hitting the exception handler.

It would be better if it made an explicit check e.g. that part of the `FailedToCalculateVsVersion` message got logged.

duncanp-sonar (commented):
Nice; much better than rely on the service properties.
A few implementation questions inline.

src/Infrastructure.VS/VsVersion/IVsVersion.cs
rita-g-sonarsource: Good point; I'll move them to Core and keep the implementation in Infrastructure.VS (I'd like the reuse the class in OpenInIDE after we drop 2015).

src/Infrastructure.VS/VsVersion/IVsVersionProvider.cs
rita-g-sonarsource: It's a log debug so I thought that we might as well see the full VS stack trace too.

src/Infrastructure.VS/VsVersion/IVsVersionProvider.cs
rita-g-sonarsource: I originally placed it in the ctor but then wanted to defer the calculation until it was needed, and thought to cache the result in telemetry manager. But you're right, it's just easier in the ctor, I'll revert.

src/Infrastructure.VS.UnitTests/VsVersionProviderTests.cs
rita-g-sonarsource: I started with that originally, but it's a logDebug, which relies on the environment parameters... https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Core.UnitTests/LoggerExtensionsTests.cs#L62

I'll change it not to be null, and instead throw and exception, and verify that it was called.

PR #2168, opened 2021-02-11
Report additional VS version fields in telemetry

### LINKED ISSUE #2147: [Telemetry] Report the VS major version and update number

rita-g-sonarsource:
Fixes #2147


src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -63,7 +66,7 @@ public void TestInitialize()
         public void Ctor_WhenGivenANullActiveSolutionBoundTracker_ThrowsArgumentNullException()
```
rita-g-sonarsource, line 66: I think we should delete most, if not all, of the `ArgumentNullException` tests. They can't be null in a MEF ctor.

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -676,6 +679,28 @@ public void TaintIssueInvestigatedRemotely_CounterIncremented(int previousCounte
             telemetryRepositoryMock.Verify(x => x.Save(), Times.Once);
         }
 
+        [TestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public void SendPayload_SendsVsVersion(bool isVersionNull)
```
rita-g-sonarsource, line 685: Supplementary test to `TelemetryHelper_CreatePayload.cs`. This test just checks that the version from IVsVersionProvider is passed through to TelemetryHelper, but unfortunately TelemetryHelper is a static class, so I can't easily Verify that it was called with the given argument.

src/Integration.UnitTests/Telemetry/TelemetryClientTests.cs
```diff
@@ -161,6 +161,7 @@ public async Task Dates_Are_Serialized_In_Roundtrip_Format()
   ""sonarlint_product"": ""SonarLint for Visual Studio"",
   ""sonarlint_version"": ""1.2.3.4"",
   ""ide_version"": ""15.16"",
+  ""ide_version_information"": null,
```
rita-g-sonarsource, line 164: Tested better in TelemetryHelper_Serialize.cs

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -37,14 +38,15 @@ public class TelemetryManagerTests
 
         // Fixed time to use in tests in which the specific time is not checked or manipulated
         private readonly DateTimeOffset AnyValidTime = new DateTimeOffset(2020, 06, 04, 11, 01, 02, TimeSpan.FromHours(1));
+        private readonly ICurrentTimeProvider currentTimeProvider = DefaultCurrentTimeProvider.Instance;
```
rita-g-sonarsource, line 41: cosmetics

src/Integration/Telemetry/TelemetryPayload.cs
```diff
@@ -40,6 +40,9 @@ public sealed class TelemetryPayload
         [JsonProperty("ide_version")]
         public string VisualStudioVersion { get; set; }
 
+        [JsonProperty("ide_version_information")]
```
rita-g-sonarsource, line 43: After discussion with Julien, we decided against commonly named fields -- the reasoning is that each IDE has its own version concepts, and they already had situations in the past where they wanted to know more intellij-specific information that doesn't fit into these 3 fields. So there is no point in trying to come up with a common schema.

Hence, `ide_version_information` might be better named as `slvs_ide_version`. I haven't changed it yet, please give your feedback on the PR and the names and I'll do the changes in bulk.

src/Integration/Telemetry/TelemetryHelper.cs
```diff
@@ -80,7 +82,22 @@ public static bool IsSonarCloud(Uri sonarqubeUri)
...
+            if (vsVersion == null)
+            {
+                return null;
+            }
+
+            return new IdeVersionInformation
+            {
+                ProductName = vsVersion.ProductName,
+                ProductBuildVersion = vsVersion.ProductVersion,
+                ProductDisplayVersion = vsVersion.ProductDisplayVersion
```
duncanp-sonar, line 100: Are these the same field names as in other instance of SL?

If not, I think `ProductBuildVersion` would be better named `InstallationVersion` so it's easier to map back to the "official" VS versioning, and also because it isn't a build version.

Given the choice, I'd call the telemetry fields `DisplayName`, `InstallationVersion` and `DisplayVersion`.

src/Integration/Telemetry/TelemetryPayload.cs
duncanp-sonar: `slvs_` would stop the data being accidentally misinterpreted when looking at telemetry. In fact, the "sl" bit is redundant, so just `vs_version_information` would be enough.

src/Integration/Telemetry/TelemetryHelper.cs
duncanp-sonar: ... and rename the `IVsVersion` properties to match.

src/Integration/Telemetry/TelemetryPayload.cs
rita-g-sonarsource: Julien suggested that it be named "slvs_", as a way to filter out flavor-specific properties. Just "vs_" won't be as eye-catching to the rest of the team.

duncanp-sonar (approved):

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
duncanp-sonar: Agreed for the MEF tests. We should remove the null checks too, or change them to assertions.

Issue #1885, opened 2020-11-09
Collect telemetry for Developer notifications usage


rita-g-sonarsource:
See https://jira.sonarsource.com/browse/SLCORE-282

PR #2173, opened 2021-02-12
Add server notifications fields to telemetry payload

rita-g-sonarsource:
Part of #1885

### DIFF
--- a/src/Integration.UnitTests/Telemetry/TelemetryDataRepositoryTests.cs
+++ b/src/Integration.UnitTests/Telemetry/TelemetryDataRepositoryTests.cs
@@ -53,7 +53,7 @@ namespace SonarLint.VisualStudio.Integration.Tests
             var repository = new TelemetryDataRepository(fileSystemMock.Object, watcherFactoryMock.Object);
 
             // Assert
-            RemoveLineEndings(fileContents.ToString()).Should().Be(RemoveLineEndings(@"<?xml version=""1.0"" encoding=""utf-16""?>
+            fileContents.ToString().Should().Be(@"<?xml version=""1.0"" encoding=""utf-16""?>
 <TelemetryData xmlns:xsi=""http://www.w3.org/2001/XMLSchema-instance"" xmlns:xsd=""http://www.w3.org/2001/XMLSchema"">
   <IsAnonymousDataShared>true</IsAnonymousDataShared>
   <NumberOfDaysOfUse>0</NumberOfDaysOfUse>
@@ -68,7 +68,20 @@ namespace SonarLint.VisualStudio.Integration.Tests
     <NumberOfIssuesInvestigatedLocally>0</NumberOfIssuesInvestigatedLocally>
     <NumberOfIssuesInvestigatedRemotely>0</NumberOfIssuesInvestigatedRemotely>
   </TaintVulnerabilities>
-</TelemetryData>"));
+  <ServerNotifications>
+    <IsDisabled>false</IsDisabled>
+    <ServerNotificationCounters>
+      <QualityGateNotificationCounter>
+        <ReceivedCount>0</ReceivedCount>
+        <ClickedCount>0</ClickedCount>
+      </QualityGateNotificationCounter>
+      <NewIssuesNotificationCounter>
+        <ReceivedCount>0</ReceivedCount>
+        <ClickedCount>0</ClickedCount>
+      </NewIssuesNotificationCounter>
+    </ServerNotificationCounters>
+  </ServerNotifications>
+</TelemetryData>");
 
             Mock.VerifyAll(fileSystemMock, watcherFactoryMock);
         }
@@ -91,6 +104,19 @@ namespace SonarLint.VisualStudio.Integration.Tests
     <NumberOfIssuesInvestigatedLocally>66</NumberOfIssuesInvestigatedLocally>
     <NumberOfIssuesInvestigatedRemotely>55</NumberOfIssuesInvestigatedRemotely>
   </TaintVulnerabilities>
+  <ServerNotifications>
+    <IsDisabled>true</IsDisabled>
+    <ServerNotificationCounters>
+        <QualityGateNotificationCounter>
+            <ReceivedCount>11</ReceivedCount>
+            <ClickedCount>22</ClickedCount>
+        </QualityGateNotificationCounter>
+        <NewIssuesNotificationCounter>
+            <ReceivedCount>33</ReceivedCount>
+            <ClickedCount>44</ClickedCount>
+        </NewIssuesNotificationCounter>
+    </ServerNotificationCounters>
+  </ServerNotifications>
 </TelemetryData>");
 
             InitializeMocks(fileContents, fileExists: true, dirExists: true);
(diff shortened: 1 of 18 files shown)


src/Integration/Telemetry/Payload/TelemetryPayload.cs
```diff
@@ -72,57 +73,8 @@ public sealed class TelemetryPayload
 
         [JsonProperty("taint_vulnerabilities")]
         public TaintVulnerabilities TaintVulnerabilities { get; set; }
-    }
-
-    public sealed class IdeVersionInformation
```
rita-g-sonarsource, line 77: Cosmetics: I placed these classes in other files, all under "Payload" folder. It started to get a bit much.

src/Integration/Telemetry/Payload/ServerNotifications.cs
```diff
@@ -0,0 +1,51 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Newtonsoft.Json;
+
+namespace SonarLint.VisualStudio.Integration.Telemetry.Payload
+{
+    public sealed class ServerNotifications
```
rita-g-sonarsource, line 25: New class according to https://jira.sonarsource.com/browse/SLCORE-282

src/Integration/Telemetry/IServerNotificationsTelemetryManager.cs
```diff
@@ -0,0 +1,87 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+using System.Diagnostics;
+
+namespace SonarLint.VisualStudio.Integration.Telemetry
+{
+    internal interface IServerNotificationsTelemetryManager
```
rita-g-sonarsource, line 26: I decided not to place more methods under ITelemetryManager: no consumer needs to access all the methods under it, and this allows a per-feature separation and keeps it contained.

src/Integration/Telemetry/IServerNotificationsTelemetryManager.cs
```diff
@@ -0,0 +1,87 @@
...
+        void QualityGateNotificationReceived();
+        void QualityGateNotificationClicked();
+        void NewIssueNotificationReceived();
+        void NewIssueNotificationClicked();
+        void NotificationsToggled(bool areNotificationsEnabled);
+    }
+
+    [Export(typeof(IServerNotificationsTelemetryManager))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class ServerNotificationsTelemetryManager : IServerNotificationsTelemetryManager
```
rita-g-sonarsource, line 37: Same logic as the interface, I decided to create a new manager. There are actually 2 bits of logic in TelemetryManager: 1) timer that does the send, 2) random "increment counter" methods that belong to different features. This falls into category 2 and since it only updates repository's data (that is a singleton), we can extract this bit out.

src/Integration/Telemetry/TelemetryHelper.cs
```diff
@@ -82,7 +83,8 @@ public static bool IsSonarCloud(Uri sonarqubeUri)
                 InstallDate = telemetryData.InstallationDate,
                 Analyses = telemetryData.Analyses,
                 ShowHotspot = telemetryData.ShowHotspot,
-                TaintVulnerabilities = telemetryData.TaintVulnerabilities
+                TaintVulnerabilities = telemetryData.TaintVulnerabilities,
+                ServerNotifications = isConnected ? telemetryData.ServerNotifications : null
```
rita-g-sonarsource, line 87: This is a bit hacky, but with the code being scattered all over the place, I didn't find a better place for it.

src/Integration.UnitTests/Telemetry/TelemetryDataRepositoryTests.cs
```diff
@@ -53,7 +53,7 @@ public void Ctor_Create_Storage_File()
             var repository = new TelemetryDataRepository(fileSystemMock.Object, watcherFactoryMock.Object);
 
             // Assert
-            RemoveLineEndings(fileContents.ToString()).Should().Be(RemoveLineEndings(@"<?xml version=""1.0"" encoding=""utf-16""?>
```
rita-g-sonarsource, line 56: The line endings were unnecessary as the test still fails when the tabs/spaces don't match. This actually made the test's failure much harder to read.

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -156,47 +157,15 @@ public void WhenFirstCallDelayerAndNewDay_ChangeLastUploadAndSaveAndSendPayload(
...
-            telemetryData.Analyses.Count.Should().Be(0); // should have cleared the list of installed languages
-            telemetryData.ShowHotspot.NumberOfRequests.Should().Be(0); // should have cleared the counter
-            telemetryData.TaintVulnerabilities.NumberOfIssuesInvestigatedRemotely.Should().Be(0); // should have cleared the counter
-            telemetryData.TaintVulnerabilities.NumberOfIssuesInvestigatedLocally.Should().Be(0); // should have cleared the counter
-            telemetryRepositoryMock.Verify(x => x.Save(), Times.Once);
-            telemetryClientMock.Verify(x => x.SendPayloadAsync(It.IsAny<TelemetryPayload>()), Times.Once);
-        }
-
-        [TestMethod]
-        public void WhenTryUploadDataTimerElapsedAndNewDay_ChangeLastUploadAndSaveAndSendPayload()
```
rita-g-sonarsource, line 180: This test was exactly the same as `WhenFirstCallDelayerAndNewDay_ChangeLastUploadAndSaveAndSendPayload`

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -213,6 +182,11 @@ public void WhenTryUploadDataTimerElapsedAndNewDay_ChangeLastUploadAndSaveAndSen
             telemetryData.ShowHotspot.NumberOfRequests.Should().Be(0); // should have cleared the counter
             telemetryData.TaintVulnerabilities.NumberOfIssuesInvestigatedRemotely.Should().Be(0); // should have cleared the counter
             telemetryData.TaintVulnerabilities.NumberOfIssuesInvestigatedLocally.Should().Be(0); // should have cleared the counter
+            telemetryData.ServerNotifications.IsDisabled.Should().BeFalse(); // should have cleared the value
```
rita-g-sonarsource, line 185: Ideally, TelemetryDataRepository would have a "ClearDailyCounters" method (or something) that the manager could call, without knowing the internal structure. But I left this as-is for now, we need to think first about the overall design.

rita-g-sonarsource:
@duncanp-sonar this is part one of the notifications telemetry -- only the new fields in the payload and the manager methods. Nothing calls these methods yet. The PR was already large enough so I decided to stop now.

src/Integration/Telemetry/Payload/TelemetryPayload.cs
duncanp-sonar: Please make this kind of change separately going forwards. It just adds noise and makes the "real" changes harder to review.

duncanp-sonar (approved):

PR #2174, opened 2021-02-15
Change notification counters to be a dictionary

rita-g-sonarsource:
Part of #1885

### DIFF
--- a/src/Integration/Telemetry/Payload/ServerNotifications.cs
+++ b/src/Integration/Telemetry/Payload/ServerNotifications.cs
@@ -18,6 +18,9 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.Collections.Generic;
+using System.Linq;
+using System.Xml.Serialization;
 using Newtonsoft.Json;
 
 namespace SonarLint.VisualStudio.Integration.Telemetry.Payload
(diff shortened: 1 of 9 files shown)


src/Integration/Telemetry/Payload/ServerNotifications.cs
```diff
@@ -28,16 +31,16 @@ public sealed class ServerNotifications
...
         [JsonProperty("count_by_type")]
-        public ServerNotificationCounters ServerNotificationCounters { get; set; } = new ServerNotificationCounters();
-    }
-
-    public sealed class ServerNotificationCounters
-    {
-        [JsonProperty("QUALITY_GATE")]
-        public ServerNotificationCounter QualityGateNotificationCounter { get; set; } = new ServerNotificationCounter();
+        [XmlIgnore]
+        public Dictionary<string, ServerNotificationCounter>  ServerNotificationCounters { get; set; } = new Dictionary<string, ServerNotificationCounter>();
```
rita-g-sonarsource, line 35: My bad, this should've been a dictionary and not a concrete class.

src/Integration/Telemetry/Payload/ServerNotifications.cs
```diff
@@ -28,16 +31,16 @@ public sealed class ServerNotifications
...
 
-        [JsonProperty("NEW_ISSUES")]
-        public ServerNotificationCounter NewIssuesNotificationCounter { get; set; } = new ServerNotificationCounter();
+        [JsonIgnore]
+        [XmlArray(nameof(ServerNotificationCounters))]
```
rita-g-sonarsource, line 38: A dictionary is not xml serializable, this is a workaround (based on https://stackoverflow.com/a/34445868/339720

src/Integration/Telemetry/Payload/ServerNotifications.cs
duncanp-sonar: Why? The json sent to the server was in the correct format, wasn't it?

rita-g-sonarsource: Still is.
This allows us not to know about the specific notification types and support all categories without needed changes.

duncanp-sonar: Do we have any other categories of notification?

It seems weird to add a strongly-typed `IServerNotificationsTelemetryManager` interface just for server notifications, then make it a "weak" wrapper round a property bag by allowing anything as a category.

rita-g-sonarsource: There's a difference between knowing that there are notification types and knowing exactly which types exist on the server. We don't care anywhere else in the code about the specific types, which made the strongly typed structure extremely awkward to use. 
We receive generic notifications from the server and display them as-is -- we should continue with this generic approach in collecting the telemetry as well. Otherwise we're just creating ourselves a tech debt to support any changes made on the server. 
You can take a look at the usage, the new generic structure makes it very simple: https://github.com/SonarSource/sonarlint-visualstudio/pull/2175/files#diff-4f799163db4bdb918950a9a163e5e53202bf34c3ecc4551342c81bf3144f14e9

duncanp-sonar: Making the internal data structure used by the server notifications flexible is one thing - that's an internal implementation issue. Exposing it to consumers in a weakly-typed way is another.

duncanp-sonar: Ok, makes sense. 

The original review comment needed a bit more explanation than "this should have been a dictionary".

duncanp-sonar (approved):

src/Integration/Telemetry/Payload/ServerNotifications.cs
rita-g-sonarsource: Fair point; I'll explain more next time.

PR #2175, opened 2021-02-15
Collect telemetry for server notifications

rita-g-sonarsource:


src/Infrastructure.VS/Infrastructure.VS.csproj
```diff
@@ -15,7 +15,7 @@
 
   <ItemGroup Label="Package references">
     <PackageReference Include="Microsoft.VisualStudio.Setup.Configuration.Interop" Version="1.16.30" />
-    <PackageReference Include="Microsoft.VisualStudio.SDK.EmbedInteropTypes" Version="15.0.34" />
+    <PackageReference Include="Microsoft.VisualStudio.SDK.EmbedInteropTypes" Version="15.0.27" />
```
rita-g-sonarsource, line 18: There was a conflict with `Integration` and `Integration.Vsix`'s dependencies. I decided to downgrade the package instead of upgrading the others.

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -186,11 +186,13 @@ Referenced assemblies:
 - 'Microsoft.VisualStudio.Shell.Immutable.14.0, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
 - 'Microsoft.VisualStudio.Shell.Interop, Version=7.1.40304.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
 - 'Microsoft.VisualStudio.Shell.Interop.8.0, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
+- 'Microsoft.VisualStudio.Utilities, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```
rita-g-sonarsource, line 189: Due to DelegateCommand usage.
I'm a bit surprised with this change -- it was already used in the project: https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration/Connection/UI/ConnectionInfoDialogView.xaml#L12
Perhaps asmref doesn't pick up xaml references.

src/Integration/Notifications/NotificationIndicator.xaml
```diff
@@ -74,7 +74,9 @@
                                     <LineBreak FontSize="16" />
 
                                     <Run Text="{Binding Message, Mode=OneWay}" />
-                                    <Hyperlink NavigateUri="{Binding Link, Mode=OneWay}" RequestNavigate="Hyperlink_RequestNavigate">more...</Hyperlink></TextBlock>
+                                    <Hyperlink Command="{Binding DataContext.NavigateToNotification, RelativeSource={RelativeSource AncestorType={x:Type UserControl}}}"
```
rita-g-sonarsource, line 77: Replaced with DelegateCommand so I could have access to the Notification's category.

src/Integration/Notifications/NotificationIndicatorViewModel.cs
```diff
@@ -60,6 +70,14 @@ internal NotificationIndicatorViewModel(Action<Action> uiThreadInvoker, ITimer a
 
             autocloseTimer.Elapsed += OnAutocloseTimerElapsed;
             ClearUnreadEventsCommand = new RelayCommand(() => HasUnreadEvents = false);
+
+            NavigateToNotification = new DelegateCommand(parameter =>
+            {
+                var notification = (SonarQubeNotification) parameter;
+                vsBrowserService.Navigate(notification.Link.ToString());
+                IsToolTipVisible = false;
```
rita-g-sonarsource, line 78: `IsToolTipVisible = false;` was part of the previous code-behind: https://github.com/SonarSource/sonarlint-visualstudio/blob/6a9c7fd05e657bb2a926413598089b3156fa4504/src/Integration/Notifications/NotificationIndicator.xaml.cs#L43

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
duncanp-sonar: That is odd. If I remember correctly, `asmref` gets the list of references from the assembly manifest metadata, which I would expect would contain all referenced assemblies.
Maybe references that are only in compiled baml don't get included in the assembly manifest.

duncanp-sonar (approved):

PR #2199, opened 2021-02-26
Normalize path of newly created header files

### LINKED ISSUE #2149: Newly added header files do not contain secondary location adornments
#### Steps to reproduce

1. Create a new header file (DO NOT CLOSE IT)
2. Change the header file to contain the following code and save the file (DO NOT CLOSE IT): 
```
#include <memory>
struct TrivCopyPaddingInside { // secondary will be in the file where this class is defined
	int x;
	char color;
	int y;
};
```
3. Create a new cpp \ change existing cpp file \ with the following code:
```
#include "header.h" // name of the header file
int f() {
	TrivCopyPaddingInside trivCopyPaddingInside;
	auto r =memcmp(&trivCopyPaddingInside, &trivCopyPaddingInside, sizeof trivCopyPaddingInside);

	return r;
}
```
4. Select the issue S5000 in the error list
5. In secondary locations panel --> navigate to the location in the header file

Expected ==> code navigated and location highlighted, adornments displayed
Actual ==> adornments are not displayed

No adornment:
image

Expected adornment:
image

#### Workaround
* Closing the header file and re-opening it solves the issue.

#### Notes
* Similar behavior occurs when a file is renamed, or when an existing header file is loaded from disk and added to the project.
* Re-saving the header file after the issue is selected makes the adornments appear.

rita-g-sonarsource:
Fixes #2149

### DIFF
--- a/src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
@@ -140,7 +140,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
         {
             return new AnalysisIssueLocation
             (
-                filePath: cFamilyIssueLocation.Filename.Replace('/', '\\'),
+                filePath: Path.GetFullPath(cFamilyIssueLocation.Filename),
                 message: cFamilyIssueLocation.Text,
                 lineHash: CalculateLineHash(cFamilyIssueLocation, fileContents),
                 startLine: cFamilyIssueLocation.Line,
--- a/src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
@@ -22,6 +22,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Diagnostics;
+using System.IO;
 using System.IO.Abstractions;
 using System.Linq;
 using SonarLint.VisualStudio.Core;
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
```diff
@@ -139,7 +140,7 @@ private AnalysisIssueLocation ToAnalysisIssueLocation(MessagePart cFamilyIssueLo
         {
             return new AnalysisIssueLocation
             (
-                filePath: cFamilyIssueLocation.Filename,
+                filePath: cFamilyIssueLocation.Filename.Replace('/', '\\'),
```
duncanp-sonar, line 143: The best way to get a path in canonical form is to call `Path.GetFullPath...)`.
The path/file doesn't have to exist, and the method will fix this issue and others e.g. duplicate directory separators and paths containing `.` and `..`.

duncanp-sonar (approved):
Looks ok, just the comment about using `Path.GetFullPath`.

src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
rita-g-sonarsource: That's what I started with originally. However, if the file path is absolute, it will be placed under the current directory. I was afraid it would lead to some side effects. WDYT?

duncanp-sonar: That's worth commenting in the code, because it isn't obvious that's why it's done like that.

Does the `subprocess.exe` ever return relative paths?

rita-g-sonarsource: At the moment - no, it shouldn't. 

On second thought, I'll change it to `Path.GetFullPath`: even if the subprocess returns a relative path, we will fail to match the issues. So the behavior would be the same between a relative path and a wrong absolute one.

PR #2201, opened 2021-02-26
Taint issues: output user-friendly message for unsupported SQ version

### LINKED ISSUE #2179: Display a user friendly message when Taint Vulnerabilities are not available in the given SQ version
#### Description

When a solution is connected to SQ versions released prior to 8.6, fetching taint vulnerabilities fails with a message:
```
ERROR: Could not find compatible implementation of 'IGetTaintVulnerabilitiesRequest' for SonarQube 7.9.4.35981.
    [Taint] Failed to synchronize taint vulnerabilities with the connected server: System.InvalidOperationException: Could not find compatible implementation of 'IGetTaintVulnerabilitiesRequest' for SonarQube 7.9.4.35981.
       at SonarQube.Client.Requests.RequestFactory.Create[TRequest](ServerInfo serverInfo)
       at SonarQube.Client.SonarQubeService.<InvokeRequestAsync>d__19`2.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at SonarQube.Client.SonarQubeService.<GetTaintVulnerabilitiesAsync>d__37.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
``` 

The message is expected however it is not clear to the user that this is a "normal" behavior.
TODO:
- [x] Remove the stack trace and display a user-friendly message that the SQ version is not supported
- [x] Link to documentation

Example:
image

duncanp-sonar:
Fixes #2179

Example of new output:

image

### DIFF
--- a/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
+++ b/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
@@ -46,6 +46,8 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
     [PartCreationPolicy(CreationPolicy.Shared)]
     internal sealed class TaintIssuesSynchronizer : ITaintIssuesSynchronizer
     {
+        private static readonly Version MinimumRequiredSonarQubeVersion = new Version(8, 6);
+
         private readonly ITaintStore taintStore;
         private readonly ISonarQubeService sonarQubeService;
         private readonly ITaintIssueToIssueVisualizationConverter converter;
@@ -96,11 +98,19 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
                 return;
             }
 
+            if (!IsFeatureSupported())
+            {
+                logger.WriteLine(TaintResources.Synchronizer_UnsupportedSQVersion, sonarQubeService.ServerInfo.Version);
+                ClearStore();
+                UpdateTaintIssuesUIContext(false);
+                return;
+            }
+
             try
             {
                 var projectKey = bindingConfiguration.Project.ProjectKey;
                 var taintVulnerabilities = await sonarQubeService.GetTaintVulnerabilitiesAsync(projectKey, CancellationToken.None);
-                
+
                 logger.WriteLine(TaintResources.Synchronizer_NumberOfServerIssues, taintVulnerabilities.Count);
 
                 var hasTaintIssues = taintVulnerabilities.Count > 0;
@@ -130,6 +140,10 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
                 ClearStore();
             }
         }
+
+        private bool IsFeatureSupported() =>
+            sonarQubeService.ServerInfo.ServerType == ServerType.SonarCloud ||
+            sonarQubeService.ServerInfo.Version >= MinimumRequiredSonarQubeVersion;
 
         private async Task<AnalysisInformation> GetAnalysisInformation(string projectKey)
         {
--- a/src/IssueViz.Security/Taint/TaintResources.resx
+++ b/src/IssueViz.Security/Taint/TaintResources.resx
@@ -135,4 +135,8 @@
   <data name="SyncPackage_Initializing" xml:space="preserve">
     <value>[Taint] Initializing taint issues synchronization package...</value>
   </data>
+  <data name="Synchronizer_UnsupportedSQVersion" xml:space="preserve">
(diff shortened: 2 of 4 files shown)


duncanp-sonar:
I thought about changing the webclient to throw an "unsupported" exception, but we can do better in this case; we know the minimum version that is required, so we can explicitly check that in the caller and return a more useful error message.

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -96,11 +98,19 @@ public async Task SynchronizeWithServer()
                 return;
             }
 
+            if (!IsFeatureSupported())
+            {
+                logger.WriteLine(TaintResources.Synchronizer_UnsupportedSQVersion, sonarQubeService.ServerInfo.Version);
+                ClearStore();
```
rita-g-sonarsource, line 104: This is a 3rd prerequisite that we have and the code is starting to get a bit messy with the copy paste.

Perhaps add helper methods that log the other 2 conditions, so we could do something like
```
if (IsStandalone() || !IsFeatureSupported() || !ServerConnected())
{
   ClearStore();
   UpdateTaintIssuesUIContext(false);
   return;
}
```
_Technically for `sonarQubeService.IsConnected` we don't clear the store, but I think that's just an oversight. I don't see a reason why we shouldn't._

src/IssueViz.Security/Taint/TaintResources.resx
```diff
@@ -135,4 +135,8 @@
   <data name="SyncPackage_Initializing" xml:space="preserve">
     <value>[Taint] Initializing taint issues synchronization package...</value>
   </data>
+  <data name="Synchronizer_UnsupportedSQVersion" xml:space="preserve">
+    <value>[Taint] Cannot fetch taint vulnerabilities: requires SonarQube v8.6 or later, or SonarCloud. Connected SonarQube version: v{0}
```
rita-g-sonarsource, line 139: `Cannot fetch` makes it sound a bit like an exception, perhaps it's worth phrasing it in a more neutral manner, i.e. 

`[Taint] Taint Vulnerabilities are not supported in the connected SonarQube server version (v{0}). Visit https://github.com/SonarSource/sonarlint-visualstudio/wiki to find out more about this and other SonarLint features.`

duncanp-sonar: Originally I had `Displaying TVs in the IDE requires...`. I changed the working to be consistent with the other condition checks i.e. standalone mode and not yet connected.

rita-g-sonarsource: Hm, I see your point. Perhaps it's worth changing them too to something "less scary"; I'd change this anyway though as it could be a marketing opportunity so it's worth making it sound attractive.

rita-g-sonarsource (approved):
Looks good, nice "workaround" with the version number. Left minor comments.

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
duncanp-sonar: I did most of that and ended up reverting it. I can't remember if there was a good reason for the `IsConnected` check not clearing the store, but there's an explicit test for it. Same for the exception handling; I couldn't remember if there was reason we didn't clear the store in that case.
Making those changes introduced several changes of behaviour that were unrelated to the bug. I'd rather keep them separate; this commit fixes the bug. Changing the other behaviours and refactoring should be separate.

rita-g-sonarsource: In the exception handling we do the opposite: we clear the store but don't reset the UI context. I find that odd.

In the IsConnected case, I believe I did that because I assumed that we will be connected "shortly", and that will either set the store or clear it, so it was more of an optimization. Originally it did nothing except return, and when the UI context was introduced, it suddenly did "something" but we didn't go back and made sure it did the same logic as the other cases.

Either way, I think it's worth doing the cleanup and investigating if there was a reason for the difference. This code was written less than a month ago and we're already "scared" to touch it. We should either clean it up, or understand what are the reasons for the difference and document them. That being said, I'm ok with doing that in a separate commit if you prefer.

duncanp-sonar: I don't have a problem with changing the behaviour, but it's unrelated to the bug fix so it should be done separately.

Issue #2091, opened 2021-01-14
Taint filtering does not work when the last document is closed


duncanp-sonar:
### Description

The taint tool window caption is not updated correctly when the last document in a bound solution is closed.
The same issue can occur when switching from a bound project to an unbound project that doesn't have any open files.

Related: #2079.

Like #2079, the problem is in the underlying `ActiveDocumentTracker`, which doesn't raise an event when the last document is closed. If the event were raised then the taint caption would be updated correctly.

The issue is cosmetic.

### Repro steps

1. Open a solution with taint issues.
2. Close all files exception one that contains taint issues (=> should have data in the taint list)
3. Close that file

### Expected behavior

The taint tool window caption reset to the default and the taint list should be cleared.

### Actual behavior

The caption is not updated.
 
![image](https://user-images.githubusercontent.com/10757559/104601699-72f74400-5672-11eb-8c3a-e5c8fd6105a3.png)

### Known workarounds

Close the solution / open another file

### Related information

* SonarLint for VS Version: 4.32

PR #2203, opened 2021-02-26
Rename ActiveDocumentTracker event class and update comments

duncanp-sonar:
* fix up listeners to handle null event args

Prep for fixing #2091

### DIFF
--- a/src/Infrastructure.VS/DocumentEvents/ActiveDocumentTracker.cs
+++ b/src/Infrastructure.VS/DocumentEvents/ActiveDocumentTracker.cs
@@ -29,12 +29,13 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.DocumentEvents
     public interface IActiveDocumentTracker : IDisposable
     {
         /// <summary>
-        /// Raises an event when a text document is brought into focus.
+        /// Raises an event when the active text document changes
         /// </summary>
         /// <remarks>
-        /// Returned <see cref="DocumentFocusedEventArgs.TextDocument"/> cannot be null.
+        /// Returned <see cref="ActiveDocumentChangedEventArgs.TextDocument"/> can be null if there is not an active text document
+        /// e.g. if the last document has been closed, or if the active document is not a text document
         /// </remarks>
-        event EventHandler<DocumentFocusedEventArgs> OnDocumentFocused;
+        event EventHandler<ActiveDocumentChangedEventArgs> ActiveDocumentChanged;
     }
 
     [Export(typeof(IActiveDocumentTracker))]
@@ -46,7 +47,7 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.DocumentEvents
         private uint cookie;
         private bool disposed;
 
-        public event EventHandler<DocumentFocusedEventArgs> OnDocumentFocused;
+        public event EventHandler<ActiveDocumentChangedEventArgs> ActiveDocumentChanged;
 
         [ImportingConstructor]
         public ActiveDocumentTracker([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider, ITextDocumentProvider textDocumentProvider)
@@ -72,7 +73,7 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.DocumentEvents
 
                 if (textDocument != null)
                 {
-                    OnDocumentFocused?.Invoke(this, new DocumentFocusedEventArgs(textDocument));
+                    ActiveDocumentChanged?.Invoke(this, new ActiveDocumentChangedEventArgs(textDocument));
                 }
             }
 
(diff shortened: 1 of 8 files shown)


duncanp-sonar:
There is only one functional change which I'll call out in a comment.
Everything else is renaming/changing comments.

src/Infrastructure.VS/DocumentEvents/ActiveDocumentTracker.cs
```diff
@@ -29,12 +29,13 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.DocumentEvents
     public interface IActiveDocumentTracker : IDisposable
     {
         /// <summary>
-        /// Raises an event when a text document is brought into focus.
+        /// Raises an event when the active text document changes
         /// </summary>
         /// <remarks>
-        /// Returned <see cref="DocumentFocusedEventArgs.TextDocument"/> cannot be null.
```
duncanp-sonar, line 35: It turns out that this constraint doesn't work - when the last document is closed, the "focused" document will be null.

Also, referring to "focus" isn't strictly accurate. Selecting a tool window changes the focus, so it's possible there isn't a document with the focus.

However, VS will still know which document is "active" i.e. the  document with focus, or the last document that had the focus. So, I renamed the event args class and the method on the interface.

src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PreCompiledHeadersEventListenerTests.cs
```diff
@@ -168,11 +168,19 @@ public void OnDocumentFocused_LanguageIsSupported_SchedulePchGeneration()
                 cancellationToken.Token));
         }
 
-        private void RaiseDocumentFocusedEvent()
+        [TestMethod]
+        public void OnDocumentFocused_NoActiveDocument_NoError()
```
duncanp-sonar, line 172: New test to check that "null" documents don't cause an error.

src/Integration.Vsix/CFamily/PreCompiledHeaders/PreCompiledHeadersEventListener.cs
```diff
@@ -71,12 +71,17 @@ internal sealed class PreCompiledHeadersEventListener : IPreCompiledHeadersEvent
...
         }
 
-        private void OnActiveDocumentFocused(object sender, DocumentFocusedEventArgs e)
+        private void OnActiveDocumentFocused(object sender, ActiveDocumentChangedEventArgs e)
         {
-            var detectedLanguages = sonarLanguageRecognizer.Detect(e.TextDocument.FilePath, e.TextDocument.TextBuffer.ContentType);
+            if (e.ActiveTextDocument == null)
+            {
+                return;
+            }
```
duncanp-sonar, line 82: This is the only substantive code change. Previously, this class relied on the active document not being null, so now it has to handle that case gracefully.

src/Integration.Vsix.UnitTests/CFamily/PreCompiledHeaders/PreCompiledHeadersEventListenerTests.cs
```diff
@@ -168,11 +168,19 @@ public void OnDocumentFocused_LanguageIsSupported_SchedulePchGeneration()
...
         {
-            activeDocumentTrackerMock.Raise(x => x.OnDocumentFocused += null, new DocumentFocusedEventArgs(CreateMockTextDocument()));
+            RaiseDocumentFocusedEvent(null);
+            languageRecognizerMock.Invocations.Count.Should().Be(0);
         }
 
+        private void RaiseDocumentFocusedEvent() =>
+            RaiseDocumentFocusedEvent(CreateMockTextDocument());
+
+        private void RaiseDocumentFocusedEvent(ITextDocument textDocument) =>
```
rita-g-sonarsource, line 181: Method name can be updated as well, `RaiseActivateDocumentChangedEvent`

src/Infrastructure.VS/DocumentEvents/ActiveDocumentTracker.cs
rita-g-sonarsource: I'm confused, when would this happen?
Based on the code, we raise the event only when the document is not null:
https://github.com/SonarSource/sonarlint-visualstudio/blob/f331d67a5894b1f31d14593b3931c7c95ea18e84/src/Infrastructure.VS/DocumentEvents/ActiveDocumentTracker.cs#L72
So the old comment should be correct, no?

duncanp-sonar: This PR is prep work to fix the bug, which will involve changing the behaviour so that the "active document" in the event args can be null.

rita-g-sonarsource (approved):

Issue #2208, opened 2021-03-01
Taint filtering - taint list is not updated when the active document content is not visible


duncanp-sonar:
### Description

The taint issues window filtering is still applied, even when the active document is not visible because it is hidden by a tool window:

![image](https://user-images.githubusercontent.com/10757559/109537555-e801ca00-7ab6-11eb-84ae-68d7fce64a5a.png)

Contrast this with the situation in which a tool window is active but the active document is still visible:

![image](https://user-images.githubusercontent.com/10757559/109537818-4464e980-7ab7-11eb-9c9c-8e144a97ff45.png)

Note: this bug was raised to document an odd corner case observed when fixing the related bugs, not because it has a major impact on the user experience. See the related bugs  #2079 and #2091.

### Repro steps

1. Open a bound solution that has taint issues
2. Open a file with taint issues
3. Open a tool window and dock it in the main frame

### Expected behavior

TODO - decide what the behaviour should be. It looks weird to have a non-filtered taint list when it isn't clear what the active document is. We *could* clear the taint list in this case since there the "active" document is not visible.

However, this is very much a corner case. It will only occur if the user has tool window docked in the main document well.

### Actual behavior

The taint list is not cleared.

### Known workarounds

None.

### Related information

* SonarLint for VS Version: v4.32

PR #2209, opened 2021-03-01
Fix ActiveDocumentTracker bugs

### LINKED ISSUE #2079: Taint filtering does not work for single-click navigation in the Solution Explorer

### LINKED ISSUE #2091: Taint filtering does not work when the last document is closed

duncanp-sonar:
Fixes #2079
Fixes #2091

### DIFF
--- a/src/Infrastructure.VS.UnitTests/ActiveDocumentTrackerTests.cs
+++ b/src/Infrastructure.VS.UnitTests/ActiveDocumentTrackerTests.cs
@@ -33,6 +33,8 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
     [TestClass]
     public class ActiveDocumentTrackerTests
     {
+        private static readonly IVsWindowFrame ValidWindowFrame = Mock.Of<IVsWindowFrame>();
+
         private Mock<IVsMonitorSelection> monitorSelectionMock;
         private Mock<IServiceProvider> serviceProviderMock;
         private Mock<Action<ActiveDocumentChangedEventArgs>> mockEventHandler;
@@ -91,12 +93,12 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
         [TestMethod]
         public void Dispose_ShouldNoLongerRaiseEvents()
         {
-            var selectedFrame = CreateMockFrame(__WindowFrameTypeFlags.WINDOWFRAMETYPE_Document);
+            var selectedFrame = ValidWindowFrame;
 
             testSubject.Dispose();
-            (testSubject as IVsSelectionEvents).OnElementValueChanged((uint)VSConstants.VSSELELEMID.SEID_WindowFrame, "1", selectedFrame.Object);
+            (testSubject as IVsSelectionEvents).OnElementValueChanged((uint)VSConstants.VSSELELEMID.SEID_WindowFrame, "1", selectedFrame);
 
-            mockEventHandler.Verify(x => x(It.IsAny<ActiveDocumentChangedEventArgs>()), Times.Never);
+            CheckEventNotRaised();
         }
 
         [TestMethod]
@@ -105,7 +107,7 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
             var result = (testSubject as IVsSelectionEvents).OnCmdUIContextChanged(1, 1);
             result.Should().Be(VSConstants.S_OK);
 
-            mockEventHandler.VerifyNoOtherCalls();
+            CheckEventNotRaised();
         }
 
         [TestMethod]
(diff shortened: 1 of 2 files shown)


src/Infrastructure.VS.UnitTests/ActiveDocumentTrackerTests.cs
```diff
@@ -91,12 +93,12 @@ public void Dispose_ShouldUnregisterFromSelectionEvents()
         [TestMethod]
         public void Dispose_ShouldNoLongerRaiseEvents()
         {
-            var selectedFrame = CreateMockFrame(__WindowFrameTypeFlags.WINDOWFRAMETYPE_Document);
```
duncanp-sonar, line 94: The product code doesn't need to check the frame type property any more; we are only interested in SEID_DocumentFrame notifications, so we know that the frame must be a document frame,

src/Infrastructure.VS.UnitTests/ActiveDocumentTrackerTests.cs
```diff
@@ -91,12 +93,12 @@ public void Dispose_ShouldUnregisterFromSelectionEvents()
...
+            var selectedFrame = ValidWindowFrame;
 
             testSubject.Dispose();
-            (testSubject as IVsSelectionEvents).OnElementValueChanged((uint)VSConstants.VSSELELEMID.SEID_WindowFrame, "1", selectedFrame.Object);
+            (testSubject as IVsSelectionEvents).OnElementValueChanged((uint)VSConstants.VSSELELEMID.SEID_WindowFrame, "1", selectedFrame);
 
-            mockEventHandler.Verify(x => x(It.IsAny<ActiveDocumentChangedEventArgs>()), Times.Never);
+            CheckEventNotRaised();
```
duncanp-sonar, line 101: Unrelated - added helper methods to make the tests easier to read.

src/Infrastructure.VS.UnitTests/ActiveDocumentTrackerTests.cs
```diff
@@ -114,115 +116,85 @@ public void OnSelectionChanged_DoesNotRaiseEvent()
...
-        public void OnElementValueChanged_NonWindowFrameElement_DoesNotRaiseEvent()
+        [DataRow(VSConstants.VSSELELEMID.SEID_PropertyBrowserSID)]
+        [DataRow(VSConstants.VSSELELEMID.SEID_ResultList)]
+        [DataRow(VSConstants.VSSELELEMID.SEID_StartupProject)]
+        [DataRow(VSConstants.VSSELELEMID.SEID_UndoManager)]
+        [DataRow(VSConstants.VSSELELEMID.SEID_UserContext)]
+        [DataRow(VSConstants.VSSELELEMID.SEID_WindowFrame)]
+        public void OnElementValueChanged_NonDocumentElementId_DoesNotRaiseEvent(VSConstants.VSSELELEMID elementId)
         {
-            var selectedFrame = CreateMockFrame(__WindowFrameTypeFlags.WINDOWFRAMETYPE_Document);
```
duncanp-sonar, line 132: We no longer need the tests that create frames with different kinds of "frame type".

src/Infrastructure.VS.UnitTests/ActiveDocumentTrackerTests.cs
```diff
@@ -114,115 +116,85 @@ public void OnSelectionChanged_DoesNotRaiseEvent()
...
+            var result = SimulateElementValueChanged(elementId, selectedFrame);
             result.Should().Be(VSConstants.S_OK);
 
             mockEventHandler.VerifyNoOtherCalls();
             textDocumentProviderMock.VerifyNoOtherCalls();
         }
 
         [TestMethod]
-        public void OnElementValueChanged_FrameElement_ToolFrame_DoesNotRaiseEvent()
+        public void OnElementValueChanged_DocumentElement_NewValueIsNull_RaisesEventWithNullArg()
```
duncanp-sonar, line 141: We should now raise the event with a null arg. It means the last doc has been closed, or the doc isn't a text doc (e.g. is a graphical doc).

duncanp-sonar:
FYI I raised another related "bug" in the same area: #2208. We can decide separately whether to fix it (the behaviour described in the new bug existed before and isn't changed by this fix).

rita-g-sonarsource (approved):

PR #2210, opened 2021-03-02
Update Taint tool window caption on the UI thread

### LINKED ISSUE #2171: [Taint] Intermittent error synchronising taint issues
### Description

See screenshot:

image

Note: although the issue occurs when synchronizing with the server, the problem is a threading issue on the VS side.

### Repro steps

1. Reset VS hive (? might not be needed)
2. Open a bound solution that has taint vulnerabilities
2. While the solution is still loading, open the taint vulnerabilities tool window
3. Go to the output window and wait for the Taint package to be initialized

### Expected behavior

Issues are synchronized successfully.

### Actual behavior

Synchronization fails silently in the background, with no obvious indication that anything has gone wrong i.e. no dialogue box, no gold bar. However:
* the taint issues window will always be empty
* an exception message is logged to the output window:

```
[Taint] Initializing taint issues synchronization package...
[Taint] Fetched 8 taint vulnerabilities.
[Taint] Failed to synchronize taint vulnerabilities with the connected server: System.InvalidOperationException: The calling thread cannot access this object because a different thread owns it.
   at System.Windows.Threading.Dispatcher.VerifyAccess()
   at System.Windows.Data.CollectionView.VerifyRefreshNotDeferred()
   at System.Windows.Data.ListCollectionView.GetEnumerator()
   at System.Windows.Data.CollectionView.System.Collections.IEnumerable.GetEnumerator()
   at System.Linq.Enumerable.<OfTypeIterator>d__95`1.MoveNext()
   at System.Linq.Enumerable.Count[TSource](IEnumerable`1 source)
   at SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.ViewModels.TaintIssuesControlViewModel.GetFilteredIssuesCount()
   at SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.ViewModels.TaintIssuesControlViewModel.UpdateCaption()
   at SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.ViewModels.TaintIssuesControlViewModel.Store_IssuesChanged(Object sender, IssuesChangedEventArgs e)
   at System.EventHandler`1.Invoke(Object sender, TEventArgs e)
   at SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintStore.NotifyIssuesChanged(IAnalysisIssueVisualization[] oldIssues)
   at SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintStore.Set(IEnumerable`1 issueVisualizations, AnalysisInformation analysisInformation)
   at SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintIssuesSynchronizer.<SynchronizeWithServer>d__9.MoveNext().
[Taint] Finished initializing taint issues synchronization package.

```

### Known workarounds

Close and re-open the solution or update the binding.

### Related information

* SonarLint for VS Version: v4.32
* Visual Studio Version: n/a
* SonarQube server version: n/a (other than it must support taint issues)
* Languages in your solution: n/a

rita-g-sonarsource:
Fixes #2171

### DIFF
--- a/src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
+++ b/src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
@@ -230,21 +230,27 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.Vie
 
         private void UpdateCaption()
         {
-            // We'll show the default caption if:
-            // * there are no underlying issues, or
-            // * there is not an active document.
-            // Otherwise, we'll add a suffix showing the number of issues in the active document.
-            string suffix = null;
-
-            if (unfilteredIssues.Count != 0 && activeDocumentFilePath != null)
+            RunOnUIThread.Run(() =>
             {
-                suffix = $" ({GetFilteredIssuesCount()})";
-            }
+                // We'll show the default caption if:
+                // * there are no underlying issues, or
+                // * there is not an active document.
+                // Otherwise, we'll add a suffix showing the number of issues in the active document.
+                string suffix = null;
 
-            WindowCaption = Resources.TaintToolWindowCaption + suffix;
+                if (unfilteredIssues.Count != 0 && activeDocumentFilePath != null)
+                {
+                    suffix = $" ({GetFilteredIssuesCount()})";
+                }
+
+                WindowCaption = Resources.TaintToolWindowCaption + suffix;
+            });
         }
 
-        private int GetFilteredIssuesCount() =>
+        /// <summary>
+        /// Must run on the UI thread.
+        /// </summary>
+        private int GetFilteredIssuesCount() => 
             IssuesView.OfType<object>().Count();
 
         private void Store_IssuesChanged(object sender, IssuesChangedEventArgs e)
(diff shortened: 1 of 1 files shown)


src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -230,21 +230,27 @@ private void UpdateIssues()
 
         private void UpdateCaption()
```
rita-g-sonarsource, line 231: I was able to reproduce it consistently.

`UpdateCaption` is called from 3 places:
1. CTOR (runs on the UI thread - no problem)
2. `ActiveDocumentTracker_OnDocumentFocused` (runs on the UI thread - no problems)
3. `Store_IssuesChanged` -- might run on background thread, which then causes this problem.

The third scenario gets broken when the window already exists by the time the synchronizer runs (reproducible by opening the tool window quickly while the binding still loads). 

I was considering placing the call to `RunOnUIThread` inside `Store_IssuesChanged`, but then we are at risk for it to be broken again in the future, if we call it from a 4th place that runs on the background thread.
I also thought about wrapping just `GetFilteredIssuesCount` but this seemed easier.
Let me know what you think.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -230,21 +230,27 @@ private void UpdateIssues()
...
+
+                WindowCaption = Resources.TaintToolWindowCaption + suffix;
+            });
         }
 
-        private int GetFilteredIssuesCount() =>
+        /// <summary>
+        /// Must run on the UI thread.
+        /// </summary>
+        private int GetFilteredIssuesCount() =>
```
duncanp-sonar, line 253: This function isn't called from anywhere else, so it would be safer to make it a local function.

duncanp-sonar (approved):

PR #2212, opened 2021-03-02
CFamily: improve messages for unsupported options

### LINKED ISSUE #1841: Improve the handling of compiler unsupported options
### Description
Currently, when an unsupported compiler option is provided we see a stacktrace in the output:
image
This is usually interpreted by the users as SonarLint is failing and is not working as expected. While in reality, this is not the case. We intentionally don't support these options (like "/std:c++latest" and "/ZW").

### Proposed solution:
Log the message when an option is not supported without the stacktrace.

duncanp-sonar:
Fixes #1841

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
@@ -129,25 +129,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         }
 
         [TestMethod]
-        [DataRow("/ZW")]
-        [DataRow("/clr")]
-        public void Should_Skip_Cx_And_Cli(string option)
-        {
-            Action action = () => MsvcDriver.ToRequest(new[] {
-                compiler,
-                new CFamilyHelper.Capture()
-                {
-                    Executable = "",
-                    Cwd = "basePath",
-                    Env = new List<string>(),
-                    Cmd = new List<string> { "cl.exe", option },
-                }
-            });
-
-            action.Should().ThrowExactly<InvalidOperationException>().And.Message.Should().Contain("CX and CLI are not supported");
-        }
-
-        [TestMethod]
         public void NoAnalyzedFiles_InvalidOperationException()
         {
             Action action = () => MsvcDriver.ToRequest(new[] {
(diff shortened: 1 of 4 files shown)


src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
```diff
@@ -128,25 +128,6 @@ public void File_RelativeAndAbsolutePath(string input, string expected)
             req.File.Should().Be(expected);
         }
 
-        [TestMethod]
-        [DataRow("/ZW")]
-        [DataRow("/clr")]
-        public void Should_Skip_Cx_And_Cli(string option)
```
duncanp-sonar, line 134: Merged into a new test below that tests all of the unsupported options.

src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
```diff
@@ -646,9 +627,32 @@ public void Microsoft_Extensions()
...
+                    Env = new List<string>(),
+                    Cmd = new List<string>{ "cl.exe", option, "anyfile.txt" }
+                }
+            });
+
+            act.Should().ThrowExactly<NotSupportedException>().And.Message.Should().Be(expectedMessage);
+        }
+
+        [TestMethod]
+        [DataRow("@foo", "unexpanded response file")]
```
duncanp-sonar, line 652: WIP: checking with Abbas whether this should be "invalid" or "unsupported"

duncanp-sonar:
Examples of new output:
image

src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
duncanp-sonar: Done: Abbas suggests keeping this as invalid for now.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -196,6 +196,12 @@ public void RequestAnalysis(string path, string charset, IEnumerable<AnalysisLan
                             analyzerOptions, cancellationToken),
                     analysisTimeout);
             }
+            catch (NotSupportedException ex)
+            {
+                // Display a simple user-friendly message for options we know are not supported.
+                // See https://github.com/SonarSource/sonarlint-visualstudio/pull/2212
+                logger.WriteLine($"Analysis error: {ex.Message}");
```
rita-g-sonarsource, line 203: Perhaps we should rephrase it not as an error, e.g. "Unsupported analysis option: ..."
WDYT?

rita-g-sonarsource (approved):

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
duncanp-sonar: That doesn't make it clear that the analysis hasn't taken place (could be interpreted as meaning the setting is not taken into account).

I'll change it to "Unable to analyze: ..."

Issue #2183, opened 2021-02-24
Implement Node.js discovery mechanism


duncanp-sonar:
### Description

Locate Node.js implementation on the machine.

#### TODO
- [x] decide strategy(ies) to use: look for version under VS? Version on path? User-specified via SonarLint option?
- [x] handle unsupported/missing version - warn user

#### Notes
* Java implementation: https://github.com/SonarSource/sonarlint-core/blob/612c500489cffb8bc789f0bd3d6144fb310f6367/core/src/main/java/org/sonarsource/sonarlint/core/NodeJsHelper.java#L94

PR #2237, opened 2021-03-19
Add infrastructure for node locators

rita-sonarsource:
Part of #2183 -- infrastructure that supports multiple discovery sources.


src/TypeScript/NodeJSLocator/NodeLocatorAggregator.cs
```diff
@@ -0,0 +1,98 @@
...
+                logger.WriteLine(Resources.ERR_IncompatibleVersion11);
+                return false;
+            }
+
+            return true;
+        }
+
+        private static Version GetNodeVersion(string path)
+        {
+            var version = FileVersionInfo.GetVersionInfo(path);
```
rita-sonarsource, line 93: This is how it's done in the nodejstools:
https://github.com/microsoft/nodejstools/blob/275e85d5cd95cad9122f59a76f9e49bead66101b/Nodejs/Product/Nodejs/Nodejs.cs#L22

I can't mock `FileVersionInfo` so I had to mock the entire method. Unfortunately, system.io.abstractions doesn't provide a wrapper for it.

src/TypeScript/NodeJSLocator/NodeLocatorAggregator.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            return new Version(version.ProductMajorPart, version.ProductMinorPart, version.ProductBuildPart);
```
rita-sonarsource, line 95: The `ProductBuildPart` is not used for comparison but it was useful to output it to the log.

src/TypeScript/NodeJSLocator/NodeLocatorsProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// Returns `node.exe` locator strategies, in precedence order.
+        /// </summary>
+        IReadOnlyCollection<INodeLocator> Get();
+    }
+
+    [Export(typeof(INodeLocatorsProvider))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class NodeLocatorsProvider : INodeLocatorsProvider
+    {
+        public IReadOnlyCollection<INodeLocator> Get() => Array.Empty<INodeLocator>();
```
rita-sonarsource, line 39: This will be populated in future PRs

src/TypeScript/NodeJSLocator/NodeLocatorAggregator.cs
duncanp-sonar: I think it's worth putting a link to the Nodejs code in a comment for future reference (perhaps at the top of this class).

src/TypeScript/NodeJSLocator/NodeLocatorAggregator.cs
```diff
@@ -0,0 +1,98 @@
...
+                }
+            }
+
+            logger.WriteLine(Resources.ERR_NoCompatibleVersion);
+            return null;
+        }
+
+        private bool IsCompatibleVersion(string nodeExePath)
+        {
+            var nodeVersion = getNodeExeVersion(nodeExePath);
```
duncanp-sonar, line 72: Probably worth asserting somewhere that the file exists, since it's an assumption we're making about the plugged-in strategies.

src/TypeScript/NodeJSLocator/Resources.resx
```diff
@@ -0,0 +1,135 @@
...
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="ERR_IncompatibleVersion" xml:space="preserve">
+    <value>[NodeJsLocator] Node version is incompatible, required version: 10 or &gt;= 12.</value>
+  </data>
+  <data name="ERR_IncompatibleVersion11" xml:space="preserve">
+    <value>[NodeJsLocator] Node version is incompatible, node 11 is not supported. Required version: 10 or &gt;= 12.</value>
+  </data>
+  <data name="ERR_NoCompatibleVersion" xml:space="preserve">
+    <value>[NodeJsLocator] Could not find a compatible node.exe</value>
```
duncanp-sonar, line 127: We should probably provide a bit more information and help here
e.g. "JavaScript analysis will not be performed. See here (link) for more information."

src/TypeScript/NodeJSLocator/NodeLocatorAggregator.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            logger.WriteLine(Resources.INFO_NodeVersion, nodeVersion.ToString(), nodeExePath);
+
+            if (nodeVersion.Major < 10)
+            {
+                logger.WriteLine(Resources.ERR_IncompatibleVersion);
```
duncanp-sonar, line 78: I think the messages can be streamlined a bit: there's no need to separately report the node version and path each time, and then have a separate "incompatible" message e.g.

Success case:
```
[NodeJsLocator] Found incompatible Node version: v10. Path: c:\....
[NodeJsLocator] Found incompatible Node version: v11. Path: c:\....
[NodeJsLocator] Found compatible Node version: v12. Path: c:\....
```

Failure case

```
[NodeJsLocator] Found incompatible Node version: v10. Path: c:\....
[NodeJsLocator] Found incompatible Node version: v11. Path: c:\....
[NodeJsLocator] Could not find a compatible node version. Supported versions are v10 and v12+. 
[NodeJsLocator] JavaScript analysis will not be performed. See here (link) for more information.
```

src/TypeScript/NodeJSLocator/Resources.resx
```diff
@@ -0,0 +1,135 @@
...
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="ERR_IncompatibleVersion" xml:space="preserve">
+    <value>[NodeJsLocator] Node version is incompatible, required version: 10 or &gt;= 12.</value>
+  </data>
+  <data name="ERR_IncompatibleVersion11" xml:space="preserve">
+    <value>[NodeJsLocator] Node version is incompatible, node 11 is not supported. Required version: 10 or &gt;= 12.</value>
```
duncanp-sonar, line 124: Using commas in sentences makes the sentence sound clumsy (and I'm not sure it's grammatically correct).

src/TypeScript.UnitTests/NodeJSLocator/NodeLocatorAggregatorTests.cs
```diff
@@ -0,0 +1,152 @@
...
+            foreach (var locator in locators)
+            {
+                VerifyLocatorCalled(locator);
+            }
+        }
+
+        [TestMethod]
+        [DataRow(9)]
+        [DataRow(11)]
+        public void Locate_OneLocatorHasIncompatibleVersion_LocatorSkipped(int incompatibleVersion)
```
duncanp-sonar, line 83: I found this test and the next one confusing; this one in particular, because it's testing two separate things at the same time: (a) incompatible versions are detected correctly, and (b) part of the search algorithm i.e. incompatible versions are ignored. The next test checks another part of (b).

I think it would be simpler if the two aspects where tested separately e.g.
a) a data-driven test that checks that incompatible versions are recognised
b) a data-driven test that checks that compatible versions are recognised (I couldn't see a test that checks that 10 is a valid version)
c) search algorithm - stops on first valid version.

(c) could be tested in a single test e.g. with the following list of locators:
* invalid
* invalid
* valid
* invalid
* valid

src/TypeScript/NodeJSLocator/NodeLocatorAggregator.cs
duncanp-sonar: Is this covered by any of tests? Could be a bit tricky since it directly accesses the file system; we wouldn't really want to have to create a new assembly with a known version on the fly, but if we use an existing assembly the test code would have to calculate the expected version in exactly the same way as the product code.

duncanp-sonar (commented):
Looks ok; comments about the tests and logging.

src/TypeScript/NodeJSLocator/NodeLocatorAggregator.cs
rita-sonarsource: Good suggestion; I ended up using the `System.ComponentModel.Composition.dll`. I couldn't use our own assemblies because it turns out that we don't populate the Product version fields, so it was always 0.0.0.

src/TypeScript/NodeJSLocator/NodeLocatorAggregator.cs
rita-sonarsource: I think the JS analysis log doesn't belong here, it should probably be in the caller class. 
But I fixed the other messages.

PR #2238, opened 2021-03-19
Support specifying node in env var

rita-sonarsource:
Part of #2183


src/TypeScript/NodeJSLocator/Locators/EnvironmentVariableNodeLocator.cs
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+using System.ComponentModel.Composition;
+using System.IO.Abstractions;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.TypeScript.NodeJSLocator.Locators
+{
+    internal interface IEnvironmentVariableNodeLocator : INodeLocator
```
duncanp-sonar, line 28: As discussed offline, it's not obvious that the per-strategy interfaces are adding much value; worth considering whether the provider should just create the instances it needs.

src/TypeScript/NodeJSLocator/Resources.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="ERR_EnvVarFileDoesNotExist" xml:space="preserve">
+    <value>[NodeJsLocator] The specified file `{0}` does not exist</value>
```
duncanp-sonar, line 121: We normally put the path at the end of the message. It makes it easier to read the message, since the path could be quite long.

The message doesn't make it clear where the specification is coming from - is it setting in VS? Is it an environment variable? If so, which one?
Something like the following would make it easier to diagnose support problems, and make it easier for the user to self-diagnose:

Scenario: not set
```
[NodeJSLocator] Checking environment variable SONAR_NODEJS_PATH...
[NodeJSLocator] Environment variable is not set
```

Scenario: set incorrectly
```
[NodeJSLocator] Checking environment variable SONAR_NODEJS_PATH...
[NodeJSLocator] Environment variable value: c:\foo\node.exe
[NodeJSLocator] Specified file does not exist
```

Scenario: set correctly
```
[NodeJSLocator] Checking environment variable SONAR_NODEJS_PATH...
[NodeJSLocator] Environment variable value: c:\foo\node.exe
[NodeJSLocator] Found node executable at c:\foo\node.exe
```

duncanp-sonar (approved):
Comments about the wording of the log output and MEF.

src/TypeScript/NodeJSLocator/Resources.resx
rita-sonarsource: Outputting the var name is a bit problematic since we use `IEnvironmentSettings`, that encapsulates the var's name. I wouldn't want to repeat that.
Do you think it's better not to use IEnvironmentSettings?

duncanp-sonar: The other logs are useful from a diagnostic point of view, so perhaps they should be debug-level.

A documentation link would be ok, but not as good in my opinion. The design is there to serve the user experience, not the other way round - ideally we shouldn't have to compromise the UX because of the design.

PR #2239, opened 2021-03-19
[3] Find nodejs in PATH

rita-sonarsource:
Part of #2183

### DIFF
--- a/src/TypeScript.UnitTests/NodeJSLocator/Locators/GlobalPathNodeLocatorTests.cs
+++ b/src/TypeScript.UnitTests/NodeJSLocator/Locators/GlobalPathNodeLocatorTests.cs
@@ -1,0 +1,131 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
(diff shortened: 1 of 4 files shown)


src/TypeScript.UnitTests/NodeJSLocator/Locators/GlobalPathNodeLocatorTests.cs
```diff
@@ -0,0 +1,131 @@
...
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.TypeScript.NodeJSLocator.Locators;
+
+namespace SonarLint.VisualStudio.TypeScript.UnitTests.NodeJSLocator.Locators
+{
+    [TestClass]
+    public class GlobalPathNodeLocatorTests
+    {
+        private readonly string programFilesPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles), "nodejs\\node.exe");
+        private readonly string programFiles86Path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86), "nodejs\\node.exe");
```
rita-sonarsource, line 37: I wanted originally to put these fields in the product class and make them internal, but `Environment.SpecialFolder.ProgramFilesX86` returns an empty string on x86 machines, which is handled separately in the product code via the `string.IsNullOrEmpty(x86path)`.
So if I were to use this full calculation in the product code, it will be a relative path instead of an absolute. It probably doesn't make much difference in practice (only if under the current directory we have a "nodejs\\node.exe" as well), but I didn't want to make changes to the nodejstools code so it will be easier to compare in the future.

PR #2240, opened 2021-03-19
Find nodejs in PATH

rita-sonarsource:
Part of #2183


src/TypeScript/NodeJSLocator/Resources.resx
```diff
@@ -117,6 +117,12 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="INFO_NotFoundInGlobalPath" xml:space="preserve">
+    <value>[NodeJsLocator] Could not find node.exe in PATH environment variable</value>
```
rita-sonarsource, line 121: Technically this message is partial, as we search in path AND in the program files, but I think it's enough for the user. We _would_ want them to update the path var.

src/TypeScript.UnitTests/NodeJSLocator/Locators/GlobalPathNodeLocatorTests.cs
```diff
@@ -0,0 +1,122 @@
...
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.TypeScript.NodeJSLocator.Locators;
+
+namespace SonarLint.VisualStudio.TypeScript.UnitTests.NodeJSLocator.Locators
+{
+    [TestClass]
+    public class GlobalPathNodeLocatorTests
+    {
+        private readonly string programFilesPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles), "nodejs\\node.exe");
```
rita-sonarsource, line 36: I wanted originally to put these fields in the product class and make them internal, but Environment.SpecialFolder.ProgramFilesX86 returns an empty string on x86 machines, which is handled separately in the product code via the string.IsNullOrEmpty(x86path).
So if I were to use this full calculation in the product code, it would be a relative path instead of an absolute on x86 machines. It probably doesn't make much difference in practice (only if under the current directory we have a "nodejs\node.exe" as well), but I didn't want to make changes to the nodejstools code so it will be easier to compare in the future.

src/TypeScript.UnitTests/NodeJSLocator/Locators/GlobalPathNodeLocatorTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            fileSystem.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public void Locate_FileNotFoundInPath_FileFoundInProgramFiles_FilePath()
+        {
+            using var scope = CreateEnvironmentVariableScope();
+
+            var fileSystem = new Mock<IFileSystem>();
+            fileSystem.Setup(x => x.File.Exists(programFilesPath)).Returns(true);
```
rita-sonarsource, line 76: I have a problem testing the x86 fallback.

On my machine, I have both program files and x86. So I can mock the fileSystem to return false for one and true for the other.
However, on our build agent, only x86 exists -- so the two cases become identical and I can't do proper assertions.
So I'm only testing the first program files, not the second fallback.

src/TypeScript/NodeJSLocator/Locators/GlobalPathNodeLocator.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            logger.WriteLine(Resources.INFO_FoundInGlobalPath, nodeExePath);
+            return nodeExePath;
+        }
+
+        /// <summary>
+        /// Copied from MS nodejstools repo: https://github.com/microsoft/nodejstools/blob/275e85d5cd95cad9122f59a76f9e49bead66101b/Nodejs/Product/Nodejs/Nodejs.cs#L112
+        /// </summary>
+        /// <remarks>
+        /// Copied as-is, except replacing `File.Exists` with <see cref="fileSystem.File.Exists"/> for testability
```
duncanp-sonar, line 64: What does the MS license say for the tools say about reusing the code? Do we need to acknowledge the original source?

src/TypeScript.UnitTests/NodeJSLocator/Locators/GlobalPathNodeLocatorTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            return new GlobalPathNodeLocator(fileSystem, Mock.Of<ILogger>());
+        }
+
+        private EnvironmentVariableScope CreateEnvironmentVariableScope(string path = null)
+        {
+            var scope = new EnvironmentVariableScope();
+
+            // remove any existing node.exe from machine's PATH for testing purposes
+            scope.SetPath(path ?? " ");
```
duncanp-sonar, line 96: Empty environment variables are normally returned as empty strings, aren't they?
Why is the "empty" path a non-empty string?

src/TypeScript/NodeJSLocator/Resources.resx
duncanp-sonar, line 121: It might be slightly clearer to say "... could not find node.exe on the PATH" since we aren't actually looking for it the environment variable itself. Although I'd expect any dev to be able to work out what it meant.

src/TypeScript.UnitTests/NodeJSLocator/Locators/GlobalPathNodeLocatorTests.cs
duncanp-sonar: The only thing I can think of would be change one of the other environment variables like `ProgramFiles(x86)` to a known value to see if that has an effect on what the `GetEnvironmentFolder` returns.

However, if that does work then it could leave the machine in a bad state if the test is interrupted for some reason so it doesn't get to do the Dispose cleanup to reset the variable. This is only likely to happen if we are running the test in the IDE and stop the test before it completes.

duncanp-sonar (commented):
Looks ok. The important question is the one about the copied code.

src/TypeScript.UnitTests/NodeJSLocator/Locators/GlobalPathNodeLocatorTests.cs
rita-sonarsource: I definitely wouldn't want to risk sabotaging our personal machines, so it's probably not worth it.

src/TypeScript/NodeJSLocator/Locators/GlobalPathNodeLocator.cs
rita-sonarsource: They have the generic apache license https://github.com/Microsoft/nodejstools/blob/main/LICENSE
but it's not filled out ("Copyright {yyyy} {name of copyright owner}"...).
I'll update our third-party notices and just point to the repo. I don't think there's any point in copy-pasting the generic apache. E.g.
```
License notice for MS NodeJsTools node.exe discovery code
---------------------------------------------------------
https://github.com/microsoft/nodejstools/blob/275e85d5cd95cad9122f59a76f9e49bead66101b/LICENSE
```

src/TypeScript.UnitTests/NodeJSLocator/Locators/GlobalPathNodeLocatorTests.cs
rita-sonarsource: No, they're returned as [nulls](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariable?view=net-5.0#System_Environment_GetEnvironmentVariable_System_String_).  So the product code fails on the `.Split` call, and since the path is unlikely to be null in reality, I preferred to tweak the test rather than the product code.

rita-sonarsource: I'll change the product code and add the null check. It's probably safer anyway.

PR #2241, opened 2021-03-22
Find bundled nodejs

rita-sonarsource:
Part of #2183


src/TypeScript/NodeJSLocator/Locators/BundledNodeLocator.cs
```diff
@@ -0,0 +1,75 @@
...
+            // e.g. C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\
+            vsShell.GetProperty((int)__VSSPROPID2.VSSPROPID_InstallRootDir, out var installDir);
+
+            msbuildNodeExePath = Path.Combine((string)installDir, MsBuildPath);
+            bundledNodeExePath = Path.Combine((string)installDir, VsBundledPath);
+        }
+
+        public string Locate()
+        {
+            var filePath = fileSystem.File.Exists(msbuildNodeExePath)
```
rita-sonarsource, line 59: As we discussed offline, I'm not a fan of this.
I'm assuming that the msbuild version will be newer but that's not necessarily true.
I think a better solution would be to return both paths and let the aggregator do the version checks.
I'll take a look at the overall code after everything is merged.
Let me know if you have a suggestion.

duncanp-sonar: Agreed.

src/TypeScript/NodeJSLocator/Locators/BundledNodeLocator.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        internal BundledNodeLocator(IServiceProvider serviceProvider, IFileSystem fileSystem, ILogger logger)
+        {
+            this.fileSystem = fileSystem;
+            this.logger = logger;
+            var vsShell = serviceProvider.GetService(typeof(SVsShell)) as IVsShell;
+
+            // e.g. C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\
+            vsShell.GetProperty((int)__VSSPROPID2.VSSPROPID_InstallRootDir, out var installDir);
```
duncanp-sonar, line 51: It would be better to check the return code to see the call succeeded (although I'm not what circumstances it might fail in).

duncanp-sonar (approved):

PR #2244, opened 2021-03-23
[1] Change NodeLocatorsProvider to NodeLocationsProvider

rita-sonarsource:


src/TypeScript/NodeJSLocator/NodeLocator.cs
```diff
@@ -19,57 +19,53 @@
...
-
-                if (string.IsNullOrEmpty(nodeExePath))
-                {
-                    continue;
-                }
-
-                if (!fileSystem.File.Exists(nodeExePath))
+                if (string.IsNullOrEmpty(nodeExePath) || !fileSystem.File.Exists(nodeExePath))
                 {
-                    logger.WriteLine(Resources.ERR_FileNotFound, nodeExePath);
```
rita-sonarsource, line 72: Since I'm now returning all candidate file paths, including the numerous folders in the PATH var, I think it will be too noisy to log all the "not found" attempts.

src/TypeScript/NodeJSLocator/NodeLocationsProvider.cs
```diff
@@ -18,24 +18,44 @@
...
     [PartCreationPolicy(CreationPolicy.Shared)]
-    internal class NodeLocatorsProvider : INodeLocatorsProvider
+    internal class NodeLocationsProvider : INodeLocationsProvider
     {
-        public IReadOnlyCollection<INodeLocator> Get() => Array.Empty<INodeLocator>();
+        private readonly IEnumerable<INodeLocationsProvider> locationsProviders;
+
+        [ImportingConstructor]
+        public NodeLocationsProvider()
+            : this(Enumerable.Empty<INodeLocationsProvider>())
```
rita-sonarsource, line 43: The plan is to change the existing Locators (env, global, bundled) to INodeLocationsProvider.

I originally placed all their logic here but it was getting complicated to test the various combinations. I think it's better to keep them separated and have this class do the aggregation. WDYT?

src/TypeScript.UnitTests/NodeJSLocator/NodeLocationsProviderTests.cs
```diff
@@ -0,0 +1,74 @@
...
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.TypeScript.NodeJSLocator;
+
+namespace SonarLint.VisualStudio.TypeScript.UnitTests.NodeJSLocator
+{
+    [TestClass]
+    public class NodeLocationsProviderTests
```
rita-sonarsource, line 30: Renamed tests class.
Originally it had only the mef test, I added 2 more.

src/TypeScript/NodeJSLocator/NodeLocationsProvider.cs
```diff
@@ -18,24 +18,44 @@
...
+        {
+            this.locationsProviders = locationsProviders;
+        }
+
+        public IReadOnlyCollection<string> Get()
+        {
+            return locationsProviders
+                .SelectMany(x => x.Get() ?? Enumerable.Empty<string>())
+                .Where(x => !string.IsNullOrEmpty(x))
+                .Distinct()
```
duncanp-sonar, line 57: According to the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.distinct?view=net-5.0) `Distinct` returns an unordered list.

In our case I don't think duplicates matter; we'll stop at the first valid entry anyway.
If you do want to de-dupe the list then it needs to be done in an order-preserving way (and one that uses an appropriate string comparer).

src/TypeScript.UnitTests/NodeJSLocator/NodeLocationsProviderTests.cs
duncanp-sonar: Yep; a bit ugly, but a pragmatic compromise.

duncanp-sonar (approved):
Just the remark about distinctness.

src/TypeScript/NodeJSLocator/NodeLocationsProvider.cs
rita-sonarsource: TIL, I'll remove the distinct.

duncanp-sonar: Me too - and I was pleasantly surprised that the behaviour was documented.

rita-sonarsource:
Closing this PR, will merge the other PR

PR #2245, opened 2021-03-23
Change NodeLocatorsProvider to NodeLocationsProvider 

rita-sonarsource:


src/TypeScript.UnitTests/NodeJSLocator/NodeLocationsProviderTests.cs
```diff
@@ -32,7 +36,22 @@ public class NodeLocationsProviderTests
...
-            MefTestHelpers.CheckTypeCanBeImported<NodeLocationsProvider, INodeLocationsProvider>(null, null);
+            MefTestHelpers.CheckTypeCanBeImported<NodeLocationsProvider, INodeLocationsProvider>(null, new []
+            {
+                MefTestHelpers.CreateExport<SVsServiceProvider>(Mock.Of<IServiceProvider>()),
+                MefTestHelpers.CreateExport<ILogger>(Mock.Of<ILogger>())
+            });
+        }
+
+        [TestMethod]
+        public void Ctor_InitializesCorrectProviders()
```
duncanp-sonar, line 47: `Ctor_InitializesCorrectProviders_InPriorityOrder` would be more explicit that the order is important.

src/TypeScript.UnitTests/NodeJSLocator/LocationProviders/BundledNodeLocationsProviderTests.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        [TestMethod]
+        public void Get_FailsToRetrieveInstallDirectory_Exception()
+        {
+            const string installDir = "c:\\test\\";
+
+            var testSubject = CreateTestSubject(installDir, shellHrResult: VSConstants.E_FAIL);
+            Action act = () => testSubject.Get();
+
+            act.Should().Throw<COMException>();
```
duncanp-sonar, line 43: Where do we catch this? We don't want a gold bar to appear complaining SLVS.
It would be simpler to just return an empty list in this case.

duncanp-sonar (approved):

Issue #2246, opened 2021-03-24
Package the required eslint-bridge files


duncanp-sonar:
### Description

As part of the build, locate the additional files required for JavaScript/TypeScript analysis and package them in the SLVS vsix.

Note: this ticket follows on from #2182. The required files will be packaged in a new VSIX which will be added as an embedded dependency of the main SLVS VSIX. This approach avoids the build time "out of memory" error with the VSSDK tooling.

#### TODO
- [x] update `DownloadCFamilyPlugin` to handle fetching and unpacking the TypeScript plugin
- [x] rename `DownloadCFamilyPlugin` project, tasks, and referencing targets
- [x] add a new VSIX project to ship the required files
- [x] modify the SLVS VSIX manifests to embed the extra VSIX
- [x] check new process works on all dev and CI machines
- [x] manually test with VS2015, VS2017 and VS2019

Follows on from #2182

PR #2249, opened 2021-03-24
Tooling: extract common code from download plugin task

duncanp-sonar:
No product code changes, just build-time internal tooling.

Part of #2246

### DIFF
--- a/build/DownloadCFamilyPlugin/Common.cs
+++ b/build/DownloadCFamilyPlugin/Common.cs
@@ -165,13 +165,6 @@ namespace DownloadCFamilyPlugin
 
         public static void ExtractTarToDirectory(string sourceFilePath, string destinationDirectory, TaskLoggingHelper logger)
         {
-            if (Directory.GetDirectories(destinationDirectory).Length > 0)
-            {
-                // Unzipping the jar is slow so skip if possible
-                LogMessage($"Skipping unzipping the tar because the destination folder already contains a sub-folder.", logger);
-                return;
-            }
-
             using (var outputStream = new FileStream(sourceFilePath, FileMode.Open))
             {
                 ICSharpCode.SharpZipLib.Tar.TarArchive tarArchive = ICSharpCode.SharpZipLib.Tar.TarArchive.CreateInputTarArchive(outputStream);
--- a/build/DownloadCFamilyPlugin/Common.cs
+++ b/build/DownloadCFamilyPlugin/Common.cs
@@ -1,0 +1,239 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2019 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Net.Http;
+using Microsoft.Build.Utilities;
+using SharpCompress.Compressors.Xz;
+
+namespace DownloadCFamilyPlugin
+{
+    internal static class Common
+    {
+        public static void EnsureWorkingDirectoryExist(string localWorkingFolder, TaskLoggingHelper logger)
+        {
+            if (!Directory.Exists(localWorkingFolder))
+            {
+                LogMessage($"Creating working folder: {localWorkingFolder}", logger);
+                Directory.CreateDirectory(localWorkingFolder);
+            }
+        }
+
+        public static void DownloadJarFile(string url, string targetFilePath, TaskLoggingHelper logger)
+        {
+            LogMessage($"Download url: {url}", logger);
+
+            if (File.Exists(targetFilePath))
+            {
+                // Downloading the file is slow so skip if possible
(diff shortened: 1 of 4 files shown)


build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.csproj
```diff
@@ -1,5 +1,32 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
+  <!-- This target exists to help test the code in this project.
```
duncanp-sonar, line 3: New target to help with testing/debugging the build tooling. It won't be executed unless you call it directly, so it won't have any impact on the "normal" build.

build/DownloadCFamilyPlugin/Common.cs
```diff
@@ -0,0 +1,239 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Net.Http;
+using Microsoft.Build.Utilities;
+using SharpCompress.Compressors.Xz;
+
+namespace DownloadCFamilyPlugin
```
duncanp-sonar, line 32: These methods have just been moved from the task class. Unless otherwise commented, the only change is to add an additional parameter for the logger.

build/DownloadCFamilyPlugin/Common.cs
```diff
@@ -0,0 +1,239 @@
...
+            }
+
+            using (var outputStream = new FileStream(sourceFilePath, FileMode.Open))
+            {
+                ICSharpCode.SharpZipLib.Tar.TarArchive tarArchive = ICSharpCode.SharpZipLib.Tar.TarArchive.CreateInputTarArchive(outputStream);
+                tarArchive.ExtractContents(destinationDirectory);
+            }
+        }
+
+        public static List<string> FindSingleFiles(string searchRoot, IEnumerable<string> patterns, TaskLoggingHelper logger)
```
duncanp-sonar, line 182: I split the old "FindFiles" method into this method and the next one so then can be reused more easily for the TypeScript plugin.

build/DownloadCFamilyPlugin/DownloadAndExtract.cs
```diff
@@ -84,188 +79,43 @@ public override bool Execute()
...
         private List<string> FindFiles(string searchRoot)
         {
             var files = new List<string>();
 
-            foreach(var file in SingleFilePatterns)
-            {
-                files.Add(FindSingleFile(searchRoot, file));
-            }
+            files.AddRange(Common.FindSingleFiles(searchRoot, SingleFilePatterns, Log));
+            files.AddRange(Common.FindMultipleFiles(searchRoot, MultipleFilesPatterns, Log));
```
duncanp-sonar, line 111: I tested that the results of the old and new methods are the same - I ran a version of the task with both old and new versions of the code with programmatic checks that exactly the same files are returned; then I deleted the old version of the code.

duncanp-sonar:
FYI there is a unit test for the CFamily configuration provider that checks a few things like the expected number of rules.
That will break if the refactoring has changed the set of files that are extracted.

build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.csproj
rita-sonarsource: I'm not sure I understand what exactly is the purpose of this target, how does it help? Does it just invoke the build with hard-coded "dummy" values (folder + jar path?)?

The name Test is a bit confusing, as it sounds like something related to unit tests. Same goes for the parameters, e.g. TestTargetDirectory -- that definitely sounds like something for unit tests. Perhaps something like "DummyParams" ?

build/DownloadCFamilyPlugin/Common.cs
```diff
@@ -0,0 +1,239 @@
...
+            }
+            return destFile;
+        }
+
+        public static void ExtractTarToDirectory(string sourceFilePath, string destinationDirectory, TaskLoggingHelper logger)
+        {
+            if (Directory.GetDirectories(destinationDirectory).Length > 0)
+            {
+                // Unzipping the jar is slow so skip if possible
+                LogMessage($"Skipping unzipping the tar because the destination folder already contains a sub-folder.", logger);
```
rita-sonarsource, line 171: I see that this is new; why do we care that the destination folder has sub folders?

build/DownloadCFamilyPlugin/Common.cs
```diff
@@ -0,0 +1,239 @@
...
+            // The txz file is compressed using XZ compression and zipped using the tar format.
+            // There is no built-in framework support for these format so we are using two
+            // open source libraries, both licensed under the MIT license.
+            EnsureWorkingDirectoryExist(destinationFolder, logger);
+
+            var uncompresssedFile = DecompressGZipFile(tarFilePath, destinationFolder, logger);
+            ExtractTarToDirectory(uncompresssedFile, destinationFolder, logger);
+        }
+
+        public static string DecompressGZipFile(string sourceFilePath, string destinationDirectory, TaskLoggingHelper logger)
```
rita-sonarsource, line 149: I see that this is new too, is that what we will need for eslint? why didn't the existing DecompressXZFile work?

build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.csproj
duncanp-sonar: It makes it simple to debug this targets file and the tasks in it. Yes, it just builds the project and invokes the target the same way it's called from the `SLVS.csproj`.

If the caller doesn't supply specific values for the two required properties, the defaults will be used.
They're not dummy values; the specified jar will be download, unpacked and dumped to the specified output folder.

The target name was originally something longer, but I changed it to "test" to make it easier to remember and shorter to type. I'll change it to `ManualTest`.

rita-sonarsource (approved):
Looks ok, just some questions

build/DownloadCFamilyPlugin/Common.cs
duncanp-sonar: Ah. This addition is from the JS prototype version.

I originally created this file as part of the prototype, which is when I originally extracted to the code. I used that as the basis of this new file, so a couple of extra changes have sneaked in.

The CFamily and TypeScript files use different formats.
CFamily use `.tgx` files, TypeScript uses `.tgz`

build/DownloadCFamilyPlugin/Common.cs
duncanp-sonar: Good spot. Left over from the prototype, when unzipping was being done at runtime.
I'll delete it.

PR #2250, opened 2021-03-24
Internal tooling: add targets to download and extract SonarJS plugin

duncanp-sonar:
Part of #2246

First part: add a new target and tasks to download the required files from the SonarJS plugin.

### DIFF
--- a/build/DownloadCFamilyPlugin/DownloadAndExtractSonarJS.cs
+++ b/build/DownloadCFamilyPlugin/DownloadAndExtractSonarJS.cs
@@ -1,0 +1,119 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2019 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace DownloadCFamilyPlugin
+{
+    /// <summary>
+    /// Downloads the Sonar JavaScript/TypeScript jar and extracts the files that need to be
+    /// embedded in the SLVS vsix
+    /// </summary>
+    /// <remarks>
+    /// Assumptions:
+    /// * we are downloading the sonar-js plugin jar e.g. https://binaries.sonarsource.com/Distribution/sonar-javascript-plugin/sonar-javascript-plugin-6.2.0.12043.jar
+    /// * the jar contains a .tgz archive containing ESLintBridge (a NodeJS app)
+    /// * the jar contains various .json files, some with well-known names.
+    /// 
+    /// We want the task to fail if it cannot locate all of the expected files (otherwise we'll build an invalid VSIX).
+    /// 
+    /// Downloading and extracting the files can be slow so we want to skip those steps if possible.
+    /// </remarks>
+    public class DownloadAndExtractSonarJS: Task
+    {
+        // Structure of the archive file:
+        // 
+        //      eslint-bridge-1.0.0.tgz -> eslint-bridge-1.0.0.tar ->
+        //          package\bin\
+        //          package\lib\
+        //          package\node_modules\
+        //          package\package.json
+
+        // The archive containing the eslintbridge
+        private const string SourceEsLintBridgeFilePattern = "eslint-bridge-*.tgz";
+
+        // Sub-folder into which the tar file should be unzipped
+        public const string TargetEslintBridgeFolderName = "eslint-bridge";
+
+        // List of patterns to match single files in the uncompressed output
+        private readonly string[] SourceSingleFilePatterns = new string[]
+        {
+                "Sonar_way_profile.json",
+                "Sonar_way_recommended_profile.json"
+        };
+
+        private readonly string SourceRelativePackageDirectory = TargetEslintBridgeFolderName + @"\package";
+
+        #region MSBuild input / output properties (set/read in the project file)
+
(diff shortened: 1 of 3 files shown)


build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.csproj
```diff
@@ -1,32 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <!-- This target exists to help test manually testing and debugging this targets file and the tasks in this project.
```
duncanp-sonar, line 3: I've moved this target to the bottom of the file.

build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.csproj
```diff
@@ -47,14 +20,14 @@
 
   <!-- Building the project creates an assembly that contains this custom task.   
        We then need to execute the CopyEmbeddedFiles task to execute the task and copy the files. -->
-  <UsingTask TaskName="DownloadAndExtract" AssemblyFile="$(OutputPath)DownloadCFamilyPlugin.dll" />
-  
-  <Target Name="CopyEmbeddedFiles" >
-    <Message Importance="high" Text="Fetching CFamily files..." />
 
-    <!-- Parameter validation -->
```
duncanp-sonar, line 55: I extracted the parameter validation to a separate target. It's only a couple of lines, so I'm not sure it was worth it.

build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.csproj
```diff
@@ -69,4 +42,96 @@
...
+  <!-- ******************************************** -->
+  <UsingTask TaskName="DownloadAndExtractSonarJS" AssemblyFile="$(OutputPath)DownloadCFamilyPlugin.dll" />
+
+  <Target Name="EmbedTypeScriptFiles" DependsOnTargets="ValidateInputs" >
+    <Message Importance="high" Text="Fetching TypeScript files..." />
+
+    <DownloadAndExtractSonarJS DownloadUrl="$(PluginUrl)">
+      <Output TaskParameter="FilesToEmbed" ItemName="FilesToEmbed" />
+      <Output TaskParameter="PackageDirectoryToEmbed" PropertyName="PackageDirectoryToEmbed" />
+    </DownloadAndExtractSonarJS>
```
duncanp-sonar, line 57: The process for dealing with SonarJS outputs is slightly more complicated than for the CFamily plugin. 

For the CFamily, we can copy everything in to a single flat directory.
For SonarJS, we want to preserve the directory structure of the `packages` folder.

The way the MSBuild `Copy` task works means we need to handle the folder-copying separately, so this task returns two outputs - a list of "flat" files, and the "packages" directory.

Overall, I'm not convinced about using tasks and targets for this - it might be simpler just to write an exe and execute it directly. However, it was easier to following the existing approach. It's internal build tooling, so the main thing for now is that it produces the correct output.

build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.csproj
```diff
@@ -69,4 +42,96 @@
...
+  <!-- ******************************************** -->
+
+  <Target Name="ValidateInputs">
+    <!-- Parameter validation -->
+    <Error Condition="$(PluginUrl)==''" Text="PluginUrl property must be specified (url from which the TypeScript jar can be downloaded)" />
+    <Error Condition="$(TargetDirectory)==''" Text="TargetDirectory property must be specified (i.e. destination to copy the TypeScript files to)" />
+  </Target>
+  
+  
+  <!-- ******************************************** -->
```
duncanp-sonar, line 97: I moved the test target here, and added a separate one for TypeScript.

build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.sln
```diff
@@ -0,0 +1,25 @@
...
+		{37E4D27F-FDAB-4D1D-8269-362ADC06D777}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{37E4D27F-FDAB-4D1D-8269-362ADC06D777}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {12B97225-FE66-4177-8E41-2D51616C13DB}
+	EndGlobalSection
+EndGlobal
```
duncanp-sonar, line 25: VS kept nagging me to save this file, so in the end I saved it and added it to source control

build/DownloadCFamilyPlugin/DownloadAndExtractSonarJS.cs
```diff
@@ -0,0 +1,119 @@
...
+            var tarSubFolder = Path.Combine(perVersionPluginFolder, TargetEslintBridgeFolderName);
+            Common.UncompressAndUnzipTgz(tarFilePath, tarSubFolder, Log);
+
+            // Locate the required files from the uncompressed jar and tar
+            var fileList = FindFiles(perVersionPluginFolder);
+            FilesToEmbed = fileList.ToArray();
+
+            PackageDirectoryToEmbed =   Path.Combine(perVersionPluginFolder, SourceRelativePackageDirectory);
+
+            return !Log.HasLoggedErrors;
```
duncanp-sonar, line 109: This code is logically the same as the CFamily code. The differences are due to:
* different compression formats used,
* different internal layouts in the archive, and
* the file patterns we want to extract.

duncanp-sonar:
This isn't linked in to the main build yet - nothing invokes the new target/task.

build/DownloadCFamilyPlugin/DownloadAndExtractSonarJS.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// Assumptions:
+    /// * we are downloading the sonar-js plugin jar e.g. https://binaries.sonarsource.com/Distribution/sonar-javascript-plugin/sonar-javascript-plugin-6.2.0.12043.jar
+    /// * the jar contains a .tgz archive containing ESLintBridge (a NodeJS app)
+    /// * the jar contains various .json files, some with well-known names.
+    /// 
+    /// We want the task to fail if it cannot locate all of the expected files (otherwise we'll build an invalid VSIX).
+    /// 
+    /// Downloading and extracting the files can be slow so we want to skip those steps if possible.
+    /// </remarks>
+    public class DownloadAndExtractSonarJS: Task
```
rita-sonarsource, line 43: Unrelated to this PR: I just noticed that the project, sln and namespace mention only CFamily. I think it's worth renaming it in the future.

duncanp-sonar: That's one of the sub-items in the ticket: #2246

rita-sonarsource (approved):

build/DownloadCFamilyPlugin/DownloadCFamilyPlugin.csproj
rita-sonarsource: I agree, in typescript's case it feels a bit over-complicated for what we're trying to achieve. We can revisit this in the future. I just hope it never breaks when you're on a holiday :D

PR #2255, opened 2021-03-26
Added another VSIX for additional files

duncanp-sonar:
* new VSIX will be embedded in the main VSIX

Relates to #2246

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -165,3 +165,4 @@ $RECYCLE.BIN/
 /src/Integration.Vsix/.CFamilyEmbedded
 /src/Integration.Vsix/lib
 /build/DownloadCFamilyPlugin/ManualTestOutput
+/src/EmbeddedVsix/ts
--- a/src/EmbeddedVsix/EmbeddedVsix.csproj
+++ b/src/EmbeddedVsix/EmbeddedVsix.csproj
@@ -92,7 +92,15 @@
     -->
     <Exec Command="&quot;$(MSBuildBinPath)\msbuild.exe&quot; ..\..\build\DownloadCFamilyPlugin\DownloadCFamilyPlugin.csproj /p:PluginUrl=$(PluginUrl) /p:TargetDirectory=&quot;$(TypeScriptTargetDirectory)&quot; /p:UseSharedCompilation=false /t:Restore;Build;EmbedTypeScriptFiles" LogStandardErrorAsError="true" IgnoreExitCode="false" />
     <ItemGroup>
-      <!-- Note: the file name has to be relative. The "Include" value is used to set the value of the "vsixSubPath" metadata item.
+      <!-- Note (1): vsixutil.exe will fail with an OutOfMemory error if there are too many files in the VSIX (the size of the files
+                     doesn't matter, it's the number that counts.
+		     So, we are excluding some files that are not required i.e. the *.map files.
+		     If we need to exclude more files then other candidates are:
+		     * the localization files for the TypeScript compiler
+		     * some modules include "test" folders which may not be needed
+		     * *.d.ts files i.e. type definition files. Ones in modules other than the TypeScript compiler/eslint modules *might* not be needed
+      -->
+      <!-- Note (2): the file name has to be relative. The "Include" value is used to set the value of the "vsixSubPath" metadata item.
 			If this is an absolute path the VSSDK tools won't complain, but the file won't be included in the vsix. -->
       <Content Include="$(TypeScriptFolderName)\**\*"
                Exclude="$(TypeScriptFolderName)\**\*.map" >
--- a/src/EmbeddedVsix/EmbeddedVsix.csproj
+++ b/src/EmbeddedVsix/EmbeddedVsix.csproj
@@ -1,0 +1,107 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <MinimumVisualStudioVersion>15.0</MinimumVisualStudioVersion>
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+    <TargetFrameworkProfile />
+  </PropertyGroup>
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <SchemaVersion>2.0</SchemaVersion>
+    <ProjectTypeGuids>{82b43b9b-a64c-4715-b499-d71e9ca2bd60};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
+    <ProjectGuid>{4BBE7B39-E00A-40FE-88FC-A53C7E49731F}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>SonarLint.VisualStudio.AdditionalFiles</RootNamespace>
+    <AssemblyName>SonarLint.VisualStudio.AdditionalFiles</AssemblyName>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
+    <GeneratePkgDefFile>false</GeneratePkgDefFile>
+    <IncludeAssemblyInVSIXContainer>true</IncludeAssemblyInVSIXContainer>
+    <IncludeDebugSymbolsInVSIXContainer>false</IncludeDebugSymbolsInVSIXContainer>
+    <IncludeDebugSymbolsInLocalVSIXDeployment>false</IncludeDebugSymbolsInLocalVSIXDeployment>
+    <CopyBuildOutputToOutputDirectory>false</CopyBuildOutputToOutputDirectory>
+    <CopyOutputSymbolsToOutputDirectory>false</CopyOutputSymbolsToOutputDirectory>
+    <StartAction>Program</StartAction>
+    <StartProgram Condition="'$(DevEnvDir)' != ''">$(DevEnvDir)devenv.exe</StartProgram>
+    <StartArguments>/rootsuffix Exp</StartArguments>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug\</OutputPath>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\</OutputPath>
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -165,3 +165,4 @@ $RECYCLE.BIN/
 /src/Integration.Vsix/.CFamilyEmbedded
 /src/Integration.Vsix/lib
 /build/DownloadCFamilyPlugin/ManualTestOutput
+/src/EmbeddedVsix/ts
```
duncanp-sonar, line 168: We don't want 4.5k+ nodejs files appearing in source control...

src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -0,0 +1,115 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
duncanp-sonar, line 1: I've left this as an old-style project for now. The first 47 lines are boilerplate - the custom stuff is below that.

src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -0,0 +1,115 @@
...
+      <IncludeAssets>runtime; build; native; contentfiles; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <None Include="source.extension.vsixmanifest">
+      <SubType>Designer</SubType>
+    </None>
+    <Content Include="ReadMe.txt" />
+  </ItemGroup>
+  <Import Project="..\EmbeddedSonarAnalyzer.props" />
+  <!-- ********************************************************** -->
```
duncanp-sonar, line 58: This section was copied from the equivalent CFamily section in the Integration.VSIX project.

src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -0,0 +1,115 @@
...
+    
+    <!-- Hook the target that fetches the additional files into to the build process. 
+         This needs to be set after the SDK targets have been imported.
+    -->
+    <PrepareForBuildDependsOn>$(PrepareForBuildDependsOn);GetTypeScriptFiles</PrepareForBuildDependsOn>
+    <!-- By default the SDK targets only zip Release builds. We'll always zip since the file is very large otherwise. -->
+    <ZipPackageCompressionLevel>Normal</ZipPackageCompressionLevel>
+  </PropertyGroup>
+  
+  <ItemGroup Label="GetTypScriptFiles inputs and outputs">
```
duncanp-sonar, line 79: This ItemGroup is new. It doesn't completely stop the VSIX from being rebuilt; for some reason MSBuild always seems to think that some manifest-related cache file has changed so the VSIX still needs to be rebuilt.

I'll look at that separately, along with making matching changes to the CFamily fetching code.

src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -0,0 +1,115 @@
...
+    <Message Importance="high" Text="Fetching the TypeScript files..." />
+    <!-- Note: building the helper project explicitly using MSBuild rather than referencing it via a ProjectReference.
+         This makes is easier for us to control when the build is performed, so we can trigger it early enough
+         to include the downloaded files as normal Content files in this project.
+         Also, we are using the "Exec" task to invoke MSBuild.exe rather the "MSBuild" task when building
+         DownloadCFamilyPlugin.csproj to avoid getting file-locking errors when rebuilding.
+    -->
+    <Exec Command="&quot;$(MSBuildBinPath)\msbuild.exe&quot; ..\..\build\DownloadCFamilyPlugin\DownloadCFamilyPlugin.csproj /p:PluginUrl=$(PluginUrl) /p:TargetDirectory=&quot;$(TypeScriptTargetDirectory)&quot; /p:UseSharedCompilation=false /t:Restore;Build;EmbedTypeScriptFiles" LogStandardErrorAsError="true" IgnoreExitCode="false" />
+    <ItemGroup>
+      <!-- Note (1): vsixutil.exe will fail with an OutOfMemory error if there are too many files in the VSIX (the size of the files
```
duncanp-sonar, line 95: This comment, and the `Exclude` for the map files below are new.

src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -0,0 +1,115 @@
...
+		     If we need to exclude more files then other candidates are:
+		     * the localization files for the TypeScript compiler
+		     * some modules include "test" folders which may not be needed
+		     * *.d.ts files i.e. type definition files. Ones in modules other than the TypeScript compiler/eslint modules *might* not be needed
+      -->
+      <!-- Note (2): the file name has to be relative. The "Include" value is used to set the value of the "vsixSubPath" metadata item.
+			If this is an absolute path the VSSDK tools won't complain, but the file won't be included in the vsix. -->
+      <Content Include="$(TypeScriptFolderName)\**\*"
+               Exclude="$(TypeScriptFolderName)\**\*.map" >
+        <CopyToOutputDirectory>Never</CopyToOutputDirectory>
```
duncanp-sonar, line 107: Setting this to `Never` doesn't seem to make any difference - I still seem to end up with the files in the output directory, and the copied to the Integration.Vsix output directory. Not sure why, but it doesn't cause a problem other than potentially slowing the build, so it's not urgent.

src/EmbeddedVsix/source.extension.vsixmanifest
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
+    <Metadata>
+        <Identity Id="SonarLint.AdditionalFiles.95521f6c-c3e6-460d-b0e6-aa64acd2fdb1" Version="1.0" Language="en-US" Publisher="SonarSource" />
+        <DisplayName>SonarLint.AdditionalFiles</DisplayName>
+        <Description xml:space="preserve">VSIX to package additional files required by non-Roslyn analyzers</Description>
+    </Metadata>
+    <Installation>
+        <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[14.0, 17.0)" />
```
duncanp-sonar, line 9: This range spans VS2015-2019 to be on the safe side. I'm not sure what would happen at installation time if the embedded VSIX could not be installed but the outer VSIX could.

src/Integration.Vsix.UnitTests/packages.lock.json
```diff
@@ -413,14 +413,6 @@
         "resolved": "7.10.6070",
         "contentHash": "p7xNzxIoZWDqSGlCi84u1CsNSL9YL7HWBdurfBUW8TZOsfW/0DS8/2MQcicQZav2Csq1f7Pep+VRdZJZkmQKJA=="
       },
-      "Microsoft.VisualStudio.SDK.Analyzers": {
```
duncanp-sonar, line 416: This project never needed the VSSDK analyzers. I think they used to appear here via a transitive reference from the Integration.Vsix project.
However, all of the VSSDK build tools assets are now marked as `private` in the Integration.VSIX project so they don't flow downstream.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -128,7 +128,10 @@
     <PackageReference Include="NuGet.VisualStudio" Version="3.3.0" />
     <PackageReference Include="System.IO.Abstractions" Version="7.1.10" />
     <PackageReference Include="System.Reflection.Metadata" Version="1.2.0" />
-    <PackageReference Include="Microsoft.VSSDK.BuildTools" Version="16.3.2099" />
+    <PackageReference Include="Microsoft.VSSDK.BuildTools" Version="16.8.3038">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
```
duncanp-sonar, line 134: I upgraded the version of the build tools to match the one in the `EmbeddedVsix` project.
The UI added the `PrivateAssets` and `IncludeAssets` settings, which are actually what we want - this project will use all of the included assets, but the references won't flow to child projects.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -169,6 +172,13 @@
       <IncludeOutputGroupsInVSIX>BuiltProjectOutputGroup%3bBuiltProjectOutputGroupDependencies%3bGetCopyToOutputDirectoryItems%3bSatelliteDllsProjectOutputGroup%3bPkgdefProjectOutputGroup%3b</IncludeOutputGroupsInVSIX>
       <IncludeOutputGroupsInVSIXLocalOnly>DebugSymbolsProjectOutputGroup%3b</IncludeOutputGroupsInVSIXLocalOnly>
     </ProjectReference>
+    <ProjectReference Include="..\EmbeddedVsix\EmbeddedVsix.csproj">
```
duncanp-sonar, line 175: Reference the embedded VSIX project.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -169,6 +172,13 @@
...
+      <Name>EmbeddedVsix</Name>
+      <Project>{4BBE7B39-E00A-40FE-88FC-A53C7E49731F}</Project>
+      <VSIXSubPath>EmbeddedVsix</VSIXSubPath>
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
```
duncanp-sonar, line 179: The new VSIX project doesn't build an assembly yet. That will be added in a later PR.

src/Integration.Vsix/Manifests/VS2015_v3Manifest/source.extension.vsixmanifest
```diff
@@ -22,6 +22,7 @@
   </Installation>
   <Dependencies>
     <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" d:Source="Manual" Version="[4.6,)" />
+    <Dependency d:Source="Project" d:ProjectName="EmbeddedVsix" Version="[1.0]" d:InstallSource="Embed" d:VsixSubPath="EmbeddedVsix" Location="|EmbeddedVsix;VSIXContainerProjectOutputGroup|" DisplayName="|EmbeddedVsix;VSIXNameProjectOutputGroup|" Id="|EmbeddedVsix;VSIXIdentifierProjectOutputGroup|" />
```
duncanp-sonar, line 25: And this is the magic that embeds one VSIX inside another.

The version is hard-coded at the moment. I'll add a ticket to deal with that later.

src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -0,0 +1,115 @@
...
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <SchemaVersion>2.0</SchemaVersion>
+    <ProjectTypeGuids>{82b43b9b-a64c-4715-b499-d71e9ca2bd60};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
+    <ProjectGuid>{4BBE7B39-E00A-40FE-88FC-A53C7E49731F}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>SonarLint.VisualStudio.AdditionalFiles</RootNamespace>
```
rita-sonarsource, line 17: The name AdditionalFiles is slightly roslyn-like, I think something like "Analyzers" or "ExternalAnalyzers" is a bit more appropriate, but it doesn't matter much. I don't know if it's worth changing.

rita-sonarsource (approved):
Great work! 🎉

src/EmbeddedVsix/EmbeddedVsix.csproj
duncanp-sonar: I thought about other names but wasn't happy with any of them. We have multiple analyzers and this VSIX only contains one of them so calling it "Analyzers" seems misleading. Ditto "ExternalAnalyzers" - we might be able to put the CFamily analyzer files in this VSIX, but that's hundreds of extra files so that seems risky. Also, it not clear what makes an analyzer "internal" or "external".
In the end I decided to name the VSIX based on its function, which is just as a way of packing any extra stuff that doesn't fit in the main package.

Issue #2184, opened 2021-02-24
Implement ESLint bridge wrapper


duncanp-sonar:
### Description

Implement wrapper class to launch and communicate with `eslintbridge`

#### TODO
- [ ] ~~Locate `eslintbridge` startup script~~ -> moved to #2256
- [x] Capture log and error output
- [x] Implement handlers for public `eslintbridge` methods

Issue #2256, opened 2021-03-27
Implement eslint-bridge process start up


rita-sonarsource:
TODO:
- [ ] Find an available port
- [ ] Locate eslintbridge startup script 
- [ ] Start the node process
- [ ] Check if it's required to call the endpoint "Close", or if killing the process is enough.

PR #2257, opened 2021-03-27
Implement eslint-bridge http wrapper

rita-sonarsource:
Part of #2184


src/Integration.UnitTests/Telemetry/TelemetryClientTests.cs
```diff
@@ -27,28 +27,13 @@
...
 using Moq;
 using Moq.Protected;
+using SonarLint.VisualStudio.Integration.UnitTests;
 
 namespace SonarLint.VisualStudio.Integration.Tests
 {
     [TestClass]
     public class TelemetryClientTests
     {
-        private class FakeHttpMessageHandler : HttpMessageHandler
```
rita-sonarsource, line 36: Extracted as-is to TestsInfrastructure project and re-used

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
```diff
@@ -0,0 +1,96 @@
...
+namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
+{
+    internal interface IEslintBridgeHttpWrapper : IDisposable
+    {
+        Task<string> PostAsync(string serverEndpoint, object request = null);
+    }
+
+    [Export(typeof(IEslintBridgeHttpWrapper))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal sealed class EslintBridgeHttpWrapper : IEslintBridgeHttpWrapper
```
rita-sonarsource, line 40: This logic is common to all the API calls so I decided to place it in a dedicated class. It makes it easier to test the caller (https://github.com/SonarSource/sonarlint-visualstudio/pull/2258)

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
+{
+    // TODO: https://github.com/SonarSource/sonarlint-visualstudio/issues/2256
+
+    internal interface IEslintBridgeStartUp : IDisposable
+    {
+        /// <summary>
+        /// Ensures that the process is running. Returns the port that server is running on.
+        /// </summary>
+        Task<int> Start();
```
rita-sonarsource, line 33: Not so sure yet about this method and the "close" functionality. Created a dedicated ticket #2256

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
```diff
@@ -0,0 +1,96 @@
...
+                var response = await httpClient.PostAsync($"http://localhost:{port}/{serverEndpoint}", content);
+                var responseString = await response.Content.ReadAsStringAsync();
+
+                logger.LogDebug(Resources.INFO_ResponseDetails, serverEndpoint, responseString);
+
+                return responseString;
+            }
+            catch (AggregateException ex) 
+            {
+                var exceptions = string.Join(Environment.NewLine, ex.InnerExceptions);
```
duncanp-sonar, line 79: Is it worth dumping the stack trace if an exception occurs? The exception will be happening in framework code so I'm not sure it will tell us much.
I think it would better just to log the error type and message, and if we want the stack trace to log it as a debug message.
Same for the non-aggregate exception handler below.

duncanp-sonar (approved):

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
duncanp-sonar: Update: see my comment in the other PR about whether it's worth catching exceptions here.

rita-sonarsource: I ended up leaving the try-catch in both places. The http wrapper catches aggregate exceptions and prints them nicely, and it's not something that I'd want to repeat in every API call, but the client still needs to handle its own exceptions, e.g. deserialization failures.

PR #2258, opened 2021-03-29
Implement EslintBridgeClient with AnalyzeJs endpoint

rita-sonarsource:
Part of #2184.
I'll add "init-linter" in a separate PR.


src/TypeScript/EslintBridgeClient/DefaultTsConfigPathProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// Returns a stringified json of a tsconfig with default parameters
+        /// </summary>
+        /// <remarks>
+        /// Java-side code: https://github.com/SonarSource/SonarJS/blob/0dda9105bab520569708e230f4d2dffdca3cec74/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensor.java#L89
+        /// </remarks>
+        private string GenerateDefaultTsConfig()
+        {
+            var tsConfig = new TsConfig
+            {
+                Files = Array.Empty<string>(), // todo: what should be here?
```
rita-sonarsource, line 70: TBD: I need to ask the team what should be here.
This seems to be code on the java side https://github.com/SonarSource/SonarJS/blob/0dda9105bab520569708e230f4d2dffdca3cec74/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensor.java#L147
... and I'm not sure what is the final output. I'll update the PR once I get a response.

src/TypeScript/EslintBridgeClient/DefaultTsConfigPathProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
+{
+    internal interface IDefaultTsConfigPathProvider
+    {
+        string GetFilePath();
+    }
+
+    internal class DefaultTsConfigPathProvider : IDefaultTsConfigPathProvider
+    {
+        internal static readonly string DefaultTsConfigFilePath = Path.GetFullPath(Path.Combine(Path.GetTempPath(), "tsconfig.json"));
```
rita-sonarsource, line 36: On the java side they store it under the temp folder. If the file is always the same, perhaps we can save it under appdata, but I'm not sure if it really makes any difference.

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -0,0 +1,99 @@
...
+                    FilePath = filePath,
+                    IgnoreHeaderComments = true,
+                    TSConfigFilePaths = new[] { defaultTsConfigPathProvider.GetFilePath() }
+                };
+
+                var responseString = await httpWrapper.PostAsync("analyze-js", analysisRequest);
+
+                return responseString == null ? null : JsonConvert.DeserializeObject<AnalysisResponse>(responseString);
+            }
+            catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
```
rita-sonarsource, line 74: There is a try-catch in the http wrapper as well. I thought about removing it from there and keeping only here, but I think the other logs are more specific and can be more meaningful in troubleshooting. Maybe it's an overkill, let me know what you think.

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -0,0 +1,99 @@
...
+        private Task Close()
+        {
+            return httpWrapper.PostAsync("close");
+        }
+
+        public async void Dispose()
+        {
+            try
+            {
+                await Close();
```
rita-sonarsource, line 90: A) It's a bit odd that we have a dedicated "Close" api method. I'd expect that killing the process is enough.
B) At the moment I'm calling "Close" only during dispose (= VS shutdown). I don't see another use case (yet) that we would want to call it.

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        private Task Close()
+        {
+            return httpWrapper.PostAsync("close");
+        }
+
+        public async void Dispose()
+        {
+            await Close();
```
rita-sonarsource, line 88: A) It's a bit odd that we have a dedicated "Close" api method. I'd expect that killing the process is enough.  
B) At the moment I'm calling "Close" only during dispose (= VS shutdown). I don't see another use case (yet) that we would want to call it, hence there's no public method for it.

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -0,0 +1,92 @@
...
+                var analysisRequest = new AnalysisRequest
+                {
+                    FilePath = filePath,
+                    IgnoreHeaderComments = true,
+                    TSConfigFilePaths = new[] { defaultTsConfigPathProvider.GetFilePath() }
+                };
+
+                var responseString = await httpWrapper.PostAsync("analyze-js", analysisRequest);
+
+                return responseString == null ? null : JsonConvert.DeserializeObject<AnalysisResponse>(responseString);
```
rita-sonarsource, line 72: I'm not handling parsing errors yet. I think it should be in the caller class, e.g. `TypeScriptAnalyzer.cs`, with this class being responsible to execute the call and deserialize the output.

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -0,0 +1,92 @@
...
+        internal EslintBridgeClient(IEslintBridgeHttpWrapper httpWrapper,
+            IDefaultTsConfigPathProvider defaultTsConfigPathProvider,
+            ILogger logger)
+        {
+            this.httpWrapper = httpWrapper;
+            this.defaultTsConfigPathProvider = defaultTsConfigPathProvider;
+            this.logger = logger;
+        }
+
+        public async Task<AnalysisResponse> AnalyzeJs(string filePath)
```
rita-sonarsource, line 59: Only accepting filePath at the moment.
FileContent could be added later if we add on-the-fly support, and the other parameters are static.

src/TypeScript/EslintBridgeClient/DefaultTsConfigPathProvider.cs
rita-sonarsource: EDIT: This whole class is not needed.
Talked to JB, in other flavors they return an empty list of ts configs.
https://github.com/SonarSource/SonarJS/blob/0dda9105bab520569708e230f4d2dffdca3cec74/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TsConfigProvider.java#L140

I'll delete the class and its usages.

src/TypeScript/EslintBridgeClient/Contract/AnalysisRequest.cs
```diff
@@ -0,0 +1,40 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Newtonsoft.Json;
+
+namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract
+{
+    // Note: corresponding TS class is called "AnalysisInput"
+    public class AnalysisRequest
```
duncanp-sonar, line 26: Do these data classes need to be public for the serializer to work, or could they be internal?

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
duncanp-sonar: Are the lower-level logs more specific? If we only catch the exceptions here then the only difference will be one extra entry in the in stack trace, won't it?

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
duncanp-sonar: A) we should be treating the eslint bridge as a black box. If it has a close method I think we should call it.
B) for the SonarLint Java daemon, we were stopping the server when the solution closed. That was partly so the end user is only paying the code of having the JS analyser process running if they actually had JS to analyse, but it was also a safety mechanism. The external process memory grew over time, so recycling the process was crude way of recovering that memory.

We don't know whether we'll have the same problem with the Node server yet. It might be worth creating a separate ticket to remind us to look at it later.

duncanp-sonar (approved):

PR #2260, opened 2021-03-29
Add InitLinter endpoint

rita-sonarsource:
Part of #2184

### DIFF
--- a/src/TypeScript.UnitTests/EslintBridgeClient/AnalysisConfigurationTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/AnalysisConfigurationTests.cs
@@ -32,6 +32,9 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
         [TestMethod]
         public void GetGlobals_NoEnvVar_ReturnsDefaults()
         {
+            using var environmentsScope = new EnvironmentVariableScope();
+            environmentsScope.SetVariable(AnalysisConfiguration.GlobalsVarName, null);
+
             var testSubject = CreateTestSubject();
             var result = testSubject.GetGlobals();
 
@@ -53,6 +56,9 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
         [TestMethod]
         public void GetEnvironments_NoEnvVar_ReturnsDefaults()
         {
+            using var environmentsScope = new EnvironmentVariableScope();
+            environmentsScope.SetVariable(AnalysisConfiguration.EnvironmentsVarName, null);
+
             var testSubject = CreateTestSubject();
             var result = testSubject.GetEnvironments();
 
--- a/src/TypeScript.UnitTests/EslintBridgeClient/AnalysisConfigurationTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/AnalysisConfigurationTests.cs
@@ -1,0 +1,76 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Linq;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.Integration.UnitTests;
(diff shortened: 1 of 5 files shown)


src/TypeScript/EslintBridgeClient/IAnalysisConfiguration.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    internal interface IAnalysisConfiguration
+    {
+        string[] GetGlobals();
+        string[] GetEnvironments();
+    }
+
+    internal class AnalysisConfiguration : IAnalysisConfiguration
+    {
+        internal const string EnvironmentsVarName = "SONARLINT_JAVASCRIPT_ENVIRONMENTS";
```
rita-sonarsource, line 34: Since we will store the values in an env var, I changed the values to be inline with our existing env var names.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
```diff
@@ -34,35 +34,64 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
     [TestClass]
     public class EslintBridgeClientTests
     {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
```
rita-sonarsource, line 38: Forgot to add that in the previous PR

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
```diff
@@ -131,11 +155,26 @@ public void Dispose_DisposesHttpWrapper()
...
+            IAnalysisConfiguration analysisConfiguration = null,
+            ILogger logger = null)
         {
             logger ??= Mock.Of<ILogger>();
+            analysisConfiguration ??= Mock.Of<IAnalysisConfiguration>();
+
+            return new TypeScript.EslintBridgeClient.EslintBridgeClient(httpWrapper, analysisConfiguration, logger);
+        }
+
+        private static Mock<IEslintBridgeHttpWrapper> SetupHttpWrapper(string endpoint, string response = null, Action<object> assertReceivedRequest = null)
```
rita-sonarsource, line 168: Helper method to avoid the repetition

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
```diff
@@ -34,35 +34,64 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
...
+
         [TestMethod]
         public async Task AnalyzeJs_HttpWrapperCalledWithCorrectArguments()
         {
-            var httpWrapper = new Mock<IEslintBridgeHttpWrapper>();
-            httpWrapper
-                .Setup(x => x.PostAsync("analyze-js", It.IsAny<object>()))
-                .ReturnsAsync((string)null);
+            var httpWrapper = SetupHttpWrapper("analyze-js", assertReceivedRequest: receivedRequest =>
+            {
```
rita-sonarsource, line 76: I'm using a helper method to do the assertion instead of `It.Is` -- It.Is was cumbersome and when it fails, it just says that there was no invocation with X params and it doesn't tell you what went wrong. This is friendlier for debugging.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
```diff
@@ -131,11 +155,26 @@ public void Dispose_DisposesHttpWrapper()
             httpWrapper.VerifyNoOtherCalls();
         }
 
-        private TypeScript.EslintBridgeClient.EslintBridgeClient CreateTestSubject(IEslintBridgeHttpWrapper httpWrapper = null, ILogger logger = null)
+        private TypeScript.EslintBridgeClient.EslintBridgeClient CreateTestSubject(IEslintBridgeHttpWrapper httpWrapper = null,
```
duncanp-sonar, line 158: Using the same name for a namespace and a type makes referring to the type a bit clumsy. Why not make them different? e.g. change the namespace to `EslintBridge`?

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -30,6 +32,8 @@ namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
 {
     internal interface IEslintBridgeClient : IDisposable
     {
+        Task InitLinter(IEnumerable<Rule> rules);
```
duncanp-sonar, line 35: It would be useful to have a link to the corresponding Java and JS implementations  at the top of this file.

src/TypeScript.UnitTests/EslintBridgeClient/AnalysisConfigurationTests.cs
```diff
@@ -0,0 +1,76 @@
...
+
+namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
+{
+    [TestClass]
+    public class AnalysisConfigurationTests
+    {
+        [TestMethod]
+        public void GetGlobals_NoEnvVar_ReturnsDefaults()
+        {
+            var testSubject = CreateTestSubject();
```
duncanp-sonar, line 35: This test assumes the environment variable is not set, which might not be the case when running on a dev machine.

Any tests that use environment variables directly should use an `EnvironmentVariableScope`.

duncanp-sonar (commented):
Couple of comments e.g. a couple of the environment tests need to be more robust.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
rita-sonarsource: I'll open a separate PR rename 👍

PR #2262, opened 2021-03-29
Start eslint-bridge process

rita-sonarsource:
Fixes #2262

### DIFF
--- a/src/TestInfrastructure/Helpers/MefTestHelpers.cs
+++ b/src/TestInfrastructure/Helpers/MefTestHelpers.cs
@@ -29,9 +29,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
 {
     public static class MefTestHelpers
     {
-        public static Export CreateExport<T>(object exportInstance)
+        public static Export CreateExport<T>(object exportInstance, string contractName = null)
         {
-            return CreateExport<T>(exportInstance, new Dictionary<string, object>());
+            return CreateExport<T>(exportInstance, new Dictionary<string, object>(), contractName);
         }
 
         public static Export CreateExportWithMetadata<T>(object exportInstance, string metadataKey, string metadataValue)
@@ -42,12 +42,12 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             return CreateExport<T>(exportInstance, metadata);
         }
 
-        public static Export CreateExport<T>(object exportInstance, IDictionary<string, object> metadata)
+        public static Export CreateExport<T>(object exportInstance, IDictionary<string, object> metadata, string contractName = null)
         {
             // Add the required export ID so that MEF knows which contract it exports.
             metadata.Add("ExportTypeIdentity", AttributedModelServices.GetTypeIdentity(typeof(T)));
 
-            string contractName = AttributedModelServices.GetContractName(typeof(T));
+            contractName ??= AttributedModelServices.GetContractName(typeof(T));
             Export export = new Export(contractName, metadata, () => exportInstance);
 
             return export;
--- a/src/TestInfrastructure/TestInfrastructure.csproj
+++ b/src/TestInfrastructure/TestInfrastructure.csproj
@@ -14,6 +14,8 @@
     <ProjectGuid>{FBD8024A-9795-4E5B-938E-AFE44FCA240C}</ProjectGuid>
 
     <SonarQubeTestProject>true</SonarQubeTestProject>
+
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
   
   <ItemGroup Label="References">
--- a/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeStartUpTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeStartUpTests.cs
@@ -1,0 +1,215 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.IO;
+using System.IO.Abstractions;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.TypeScript.EslintBridgeClient;
+using SonarLint.VisualStudio.TypeScript.NodeJSLocator;
+
+namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
+{
+    [TestClass]
+    public class EslintBridgeStartUpTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<EslintBridgeStartUp, IEslintBridgeStartUp>(null, new[]
+            {
+                MefTestHelpers.CreateExport<string>("some path", EslintBridgeStartUp.EslintBridgeDirectoryMefContractName),
+                MefTestHelpers.CreateExport<INodeLocator>(Mock.Of<INodeLocator>()),
+                MefTestHelpers.CreateExport<ILogger>(Mock.Of<ILogger>())
+            });
+        }
+
+        [TestMethod]
+        public void Start_FailsToFindNode_FileNotFoundException()
+        {
+            var nodeLocator = SetupNodeLocator(null);
+
+            var testSubject = CreateTestSubject(nodeLocator: nodeLocator.Object);
+            Func<Task> act = async () => await testSubject.Start();
+
+            act.Should().ThrowExactly<FileNotFoundException>().And.Message.Should().Contain("node.exe");
+        }
+
+        [TestMethod]
+        public void Start_FailsToFindStartupScript_FileNotFoundException()
+        {
+            const string filePath = "somefile.txt";
+
+            var fileSystem = SetupStartupScriptFile(filePath, false);
+            var testSubject = CreateTestSubject(filePath, fileSystem: fileSystem.Object);
+            Func<Task> act = async () => await testSubject.Start();
+
+            act.Should().ThrowExactly<FileNotFoundException>().And.Message.Should().Contain(filePath);
+        }
+
+        [TestMethod]
+        public async Task Start_ServerStarts_ReturnsPortNumber()
+        {
+            var scriptPath = CreateScriptThatPrintsPortNumber(123);
+            var startupScriptPath = " "; // pass whitespace so it would not fail NullOrEmpty validation but also do nothing
+
+            var nodeLocator = SetupNodeLocator(scriptPath);
+            var testSubject = CreateTestSubject(startupScriptPath, nodeLocator: nodeLocator.Object);
+
+            var port = await testSubject.Start();
+            port.Should().Be(123);
+        }
+
+        [TestMethod]
+        public async Task Start_TaskSucceeds_NextCallDoesNotStartAnotherServer()
+        {
+            var scriptPath = CreateScriptThatPrintsPortNumber(123);
+            var startupScriptPath = " "; // pass whitespace so it would not fail NullOrEmpty validation but also do nothing
+
+            var nodeLocator = SetupNodeLocator(scriptPath);
+            var testSubject = CreateTestSubject(startupScriptPath, nodeLocator: nodeLocator.Object);
+
+            var port = await testSubject.Start();
+            port.Should().Be(123);
+
+            port = await testSubject.Start();
+            port.Should().Be(123);
+
+            nodeLocator.Verify(x => x.Locate(), Times.Once);
+        }
+
+        [TestMethod]
+        public async Task Start_TaskFails_NextCallAttemptsAgain()
+        {
+            var scriptPath = CreateScriptThatPrintsPortNumber(123);
+            var startupScriptPath = " "; // pass whitespace so it would not fail NullOrEmpty validation but also do nothing
+
+            var nodeLocator = new Mock<INodeLocator>();
+            nodeLocator.SetupSequence(x => x.Locate())
+                .Throws(new NotImplementedException("some exception"))
+                .Returns(scriptPath);
+
+            var testSubject = CreateTestSubject(startupScriptPath, nodeLocator: nodeLocator.Object);
+
+            Func<Task> act = async () => await testSubject.Start();
+            act.Should().Throw<NotImplementedException>().And.Message.Should().Be("some exception");
+
+            var port = await testSubject.Start();
+            port.Should().Be(123);
+        }
+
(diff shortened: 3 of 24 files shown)


src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -29,9 +29,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
 {
     public static class MefTestHelpers
     {
-        public static Export CreateExport<T>(object exportInstance)
+        public static Export CreateExport<T>(object exportInstance, string contractName = null)
```
rita-sonarsource, line 32: Supporting import by contract name

src/TestInfrastructure/TestInfrastructure.csproj
```diff
@@ -14,6 +14,8 @@
     <ProjectGuid>{FBD8024A-9795-4E5B-938E-AFE44FCA240C}</ProjectGuid>
 
     <SonarQubeTestProject>true</SonarQubeTestProject>
+
```
rita-sonarsource, line 17: Upgraded to use "??", it's a test project so no harm.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
```diff
@@ -19,17 +19,183 @@
...
         /// Ensures that the process is running. Returns the port that server is running on.
         /// </summary>
         Task<int> Start();
     }
+
+    [Export(typeof(IEslintBridgeStartUp))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal sealed class EslintBridgeStartUp : IEslintBridgeStartUp
+    {
+        internal const string EslintBridgeDirectoryMefContractName = "SonarLint.TypeScript.EsLintBridgeServerPath";
```
rita-sonarsource, line 46: Took this from the draft PR, can change to whatever we decide.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
```diff
@@ -19,17 +19,183 @@
...
+                FileName = nodePath,
+                RedirectStandardError = true,
+                RedirectStandardOutput = true,
+                UseShellExecute = false, // required if we want to capture the error output
+                ErrorDialog = false,
+                CreateNoWindow = true,
+                Arguments = serverStartupScriptLocation
+            };
+
+            psi.EnvironmentVariables.Add("NODE_PATH", @"%APPDATA%\npm\node_modules");
```
rita-sonarsource, line 116: I'm having a trouble to test this. I thought about creating a file that outputs this to another file but that's just over-complicated.
An alternative is to inherit from Process and wrap `Process.Start`, but I'm not sure that it's worth it.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
```diff
@@ -19,17 +19,183 @@
...
+        }
+
+        private void OnOutputDataReceived(object sender, DataReceivedEventArgs e)
+        {
+            if (string.IsNullOrEmpty(e.Data))
+            {
+                return;
+            }
+
+            logger.LogDebug("ESLINT-BRIDGE OUTPUT: " + e.Data);
```
rita-sonarsource, line 137: I left the log-debug strings as plain strings, didn't think it's worth putting them in resources. WDYT?

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
```diff
@@ -19,17 +19,183 @@
...
+            Process = null;
+        }
+
+        private string FindNodePath()
+        {
+            var nodePath = nodeLocator.Locate();
+
+            if (string.IsNullOrEmpty(nodePath))
+            {
+                throw new FileNotFoundException("Could not find node.exe");
```
rita-sonarsource, line 185: Not sure that FileNotFoundException is the best one, maybe a custom exception would be better. This is so the task will fail and the caller could retry.

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
```diff
@@ -65,8 +65,6 @@ public async Task<string> PostAsync(string serverEndpoint, object request = null
                 logger.LogDebug(Resources.INFO_RequestDetails, serverEndpoint, serializedRequest);
 
                 var port = await eslintBridgeStartUp.Start();
-                logger.LogDebug(Resources.INFO_ServerStarted, port);
```
rita-sonarsource, line 68: Moved to the startup code.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeStartUpTests.cs
```diff
@@ -0,0 +1,215 @@
...
+            Func<Task> act = async () => await testSubject.Start();
+
+            act.Should().ThrowExactly<FileNotFoundException>().And.Message.Should().Contain(filePath);
+        }
+
+        [TestMethod]
+        public async Task Start_ServerStarts_ReturnsPortNumber()
+        {
+            var scriptPath = CreateScriptThatPrintsPortNumber(123);
+            var startupScriptPath = " "; // pass whitespace so it would not fail NullOrEmpty validation but also do nothing
```
rita-sonarsource, line 76: No idea how else to simulate this 🙈

rita-sonarsource:
@duncanp-sonar please review. I'm not sure about the "unit" tests; I hope they won't be flaky. Let me know if you got another idea. Maybe it's worth wrapping `Process`.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeStartUpTests.cs
```diff
@@ -0,0 +1,215 @@
...
+            var nodeLocator = SetupNodeLocator(scriptPath);
+            var testSubject = CreateTestSubject(startupScriptPath, nodeLocator: nodeLocator.Object);
+
+            var port = await testSubject.Start();
+            port.Should().Be(123);
+
+            port = await testSubject.Start();
+            port.Should().Be(123);
+
+            nodeLocator.Verify(x => x.Locate(), Times.Once);
```
rita-sonarsource, line 100: Not the most reliable way to test this but I don't see another option besides wrapping Process.Start :/

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeStartUpTests.cs
```diff
@@ -0,0 +1,215 @@
...
+            var content = "echo port " + portNumber;
+
+            if (isHanging)
+            {
+                content += Environment.NewLine + "set /p asd=\"Waiting for process to be killed\"";
+            }
+
+            var fileName = Path.GetTempFileName() + ".bat";
+
+            File.WriteAllText(fileName, content);
```
rita-sonarsource, line 210: FYI originally I was deleting this file at the end of the test, but it was causing the tests to fail from time to time due to the file still being in use by another process. It could be the antivirus on my machine (like the problem I had with the cfamily subprocess). So in the end I decided to just leave the temp files and let the operating system clean them... Not the best solution, I know, but it's the same thing that we do in `ProcessRunnerTests` and I don't have another idea.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
duncanp-sonar: Not sure; it might not be necessary. There was a lot of temporary/experimental stuff in the prototype, including some bits that were ported from the Java code.

I think we should drop it, and add it only if it turns out to be necessary.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
duncanp-sonar: Ok; the other messages should be resources, though, with a standard message prefix.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
```diff
@@ -19,17 +19,183 @@
...
     {
         /// <summary>
         /// Ensures that the process is running. Returns the port that server is running on.
         /// </summary>
         Task<int> Start();
     }
+
+    [Export(typeof(IEslintBridgeStartUp))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal sealed class EslintBridgeStartUp : IEslintBridgeStartUp
```
duncanp-sonar, line 44: The class name doesn't really convey what the class is doing; it's a wrapper that persists for the lifetime of the process so it can monitor the process and capture the output.

The class/interface name suggests that it just starts the process, which makes it non-obvious why it implements `IDisposable`, or that disposing the class will terminate the external process.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
```diff
@@ -19,17 +19,183 @@
...
     internal interface IEslintBridgeStartUp : IDisposable
     {
         /// <summary>
         /// Ensures that the process is running. Returns the port that server is running on.
         /// </summary>
         Task<int> Start();
     }
+
+    [Export(typeof(IEslintBridgeStartUp))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
```
duncanp-sonar, line 43: I'm not clear how the lifetime of the various objects is going to work. How do you see this class and the `EsLintBridgeHttpWrapper` being consumed by the analyzers?

How would we stop the node process when a solution is closed, or have separate processes for TS and JS?

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
```diff
@@ -19,17 +19,183 @@
...
+            var psi = new ProcessStartInfo
+            {
+                FileName = nodePath,
+                RedirectStandardError = true,
+                RedirectStandardOutput = true,
+                UseShellExecute = false, // required if we want to capture the error output
+                ErrorDialog = false,
+                CreateNoWindow = true,
+                Arguments = serverStartupScriptLocation
+            };
```
duncanp-sonar, line 114: The `ts\bin\server` script can handle multiple other arguments. Do we need to supply any of them (e.g. working directory, shouldUseTypeScriptParserForJS)?

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
duncanp-sonar: Leaving aside the type of the exception for now, what overall UX do we want if node can't be found?

Should we should this message only once then stop trying to analyse JS files, or should we keep triggering analyses and display this message multiple times?

It's not directly the responsibility of this class, but it would affect how this class and the http wrapper are created and called.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeStartUpTests.cs
duncanp-sonar: I don't think it's worth worrying about. The files are tiny and will get cleaned up eventually.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeStartUpTests.cs
duncanp-sonar: The `startupScriptPath` is being passed to your dummy node script, isn't it? Does it matter what the arguments are since the script will ignore them anyway?

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeStartUpTests.cs
```diff
@@ -0,0 +1,215 @@
...
+            var testSubject = CreateTestSubject(filePath, fileSystem: fileSystem.Object);
+            Func<Task> act = async () => await testSubject.Start();
+
+            act.Should().ThrowExactly<FileNotFoundException>().And.Message.Should().Contain(filePath);
+        }
+
+        [TestMethod]
+        public async Task Start_ServerStarts_ReturnsPortNumber()
+        {
+            var scriptPath = CreateScriptThatPrintsPortNumber(123);
```
duncanp-sonar, line 75: Could you change `scriptPath` to something more descriptive and less like `strartupScriptPath`?
e.g. `fakeNodeExePath`

duncanp-sonar:
> I'm not sure about the "unit" tests; I hope they won't be flaky. Let me know if you got another idea. Maybe it's worth wrapping Process.

FYI the S4MSB has used this approach for a long time and the tests are reliable.

duncanp-sonar (commented):
We're starting to hit lower-level details that weren't handled in the prototype e.g. what arguments do we actually need to pass to node exe.

Also, I'm not clear what the bigger picture is regarding object lifetime, overall UX, usage by the analyzers so I've some questions about that. Let me know if you'd rather discuss them offline.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
rita-sonarsource: Good question. Originally I thought that it's better to retry, so the user could install node in the background and won't have to restart VS. However, if you're a mixed C# and JS dev, and you don't care about JS, you're gonna get spammed with this message each and every time you open/save a JS file. On the other hand, if it's shown only once, you might miss it in the sea of logs... So I'm really not sure. Perhaps we can start with retries and see if anyone complains? WDYT?

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
rita-sonarsource: I took the inspiration from the Owin classes, where you have a StartUp class that wraps the running app and is responsible for terminating it.
What do you think would be a better name? IEslintBridgeHost? IEslintBridgeProcess?

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
rita-sonarsource: If we decide to go with 2 processes then this will have to be changed, probably to a Factory class (that is a singleton), that receives the dependencies in the ctor and news-up the instances.
Stopping when a solution is closed: I haven't exposed a Stop method yet, but it's something that we could do if we decide to terminate the process when a solution is closed.
Do you think that it's worth doing the changes now?

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
rita-sonarsource: I'll check 👍 I assumed that was all based on the prototype, I haven't checked what else is supported.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
duncanp-sonar: Yeugh - that's not a helpful naming convention.
Either of those names would be better, and "process" is probably more accurate.

> I wanted to avoid giving away the fact that it's a process
Any specific reason? On the one hand it's an implementation detail, but it's a pretty important one that conveys useful information about what to expect about the behaviour of the class.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
duncanp-sonar: As it is, the user would get errors every time a JS file is opened or saved (with the full call stack, or has that changed)? I'd say that was too noisy.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
duncanp-sonar: Not now you've written it like this, but I think it's worth thinking about the overall UX and how it's going to work end-to-end with the analysers before doing much more to avoid unnecessary rework later.

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
rita-sonarsource: Yep, just that it's an implementation detail. Renamed to `IEslintBridgeProcess`

src/TypeScript/EslintBridgeClient/EslintBridgeStartUp.cs
duncanp-sonar: I agree we should have a retry process to handle cases where the process has crashed/been terminated; we should just handle that behind the scenes.
I don't think we need a retry process for cases where node isn't installed. It's not unreasonable for the user to have to restart VS after they have installed node.

PR #2264, opened 2021-03-30
Support cancellation in EslintBridgeClient

### LINKED ISSUE #2259: Implement cancellation of previous analysis when a new analysis is triggered
#### Description

When a new JS/TS analysis is requested, the old analysis should be cancelled, similarly to the way it works for CFamily.

Note: should also abort the http request.

rita-sonarsource:
Fixes #2259

### DIFF
--- a/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeHttpWrapperTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeHttpWrapperTests.cs
@@ -21,14 +21,17 @@
 using System;
 using System.Collections.Generic;
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
+using Moq.Protected;
 using Newtonsoft.Json;
 using SonarLint.VisualStudio.Integration;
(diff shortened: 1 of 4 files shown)


src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeHttpWrapperTests.cs
```diff
@@ -70,11 +73,42 @@ public async Task PostAsync_ExecutesRequestsOnTheCorrectPort()
...
+
+            httpMessageHandler
+                .Protected()
+                .Setup<Task<HttpResponseMessage>>("SendAsync",
+                    ItExpr.IsAny<HttpRequestMessage>(),
+                    ItExpr.IsAny<CancellationToken>())
+                .ReturnsAsync(new HttpResponseMessage {Content = new StringContent("response")})
+                .Callback((HttpRequestMessage message, CancellationToken receivedToken) =>
+                {
+                    // SendAsync uses `CancellationTokenSource.CreateLinkedTokenSource`
```
rita-sonarsource, line 96: I couldn't do a simple comparison of receivedToken and the original token, as they're actually not the same token -- decompiled HttpClient shows that it creates CancellationTokenSource.CreateLinkedTokenSource under the covers, so I had to get creative.

duncanp-sonar (approved):

Issue #2271, opened 2021-03-31
Implement new JavaScript analyzer


rita-sonarsource:
#### Description

- [x] Implement IAnalyzer
- [x] Call EslintBridgeClient
- [x] Handle parsing errors
- [x] Convert returned issues to AnalysisIssue
- [ ] ~~Display analysis logs (started / cancelled / failed / finished)~~ --> Handled in #2280
- [x] Use the new analyzer instead of DaemonAnalyzer

Issue #2191, opened 2021-02-24
Implement rule key converter


duncanp-sonar:
### Description

There are two different sets of rule keys in use: the Sxxx rule keys used in SonarQube/SonarCloud, and the ESLint rule key.
Quality profiles will use the Sxxx for SonarSource rules.
The SonarJS plugin will contain a metadata file containing the mapping between Sxxx and ESLint rule keys.

#### TODO
- [x] decide whether to implement rule key translation in the first version - may not be required until Connected Mode is supported? UPDATE: the required metadata is in the file supplied with SonarJS, so implementing the mapping is simple.
- [x] check how bundled/external ESLint rules are handled - do they also have Sxxx numbers? UPDATE: not an issue; we'll use the metadata from the file
- [x] decide whether to include or strip the language prefix from the rule key e.g. `typescript:S123`. See below for screenshots from VS Code and IntelliJ, which do show the language. However, the layout is quite different from the VS Error List: DECISION - we'll start by stripping the repo prefix so the error code shown in the Error List in the same form as for other errors/warnings (including the SonarC# and SonarVB rules, and IDE/compiler error codes).
- [x] implement key mapper

#### VS Code
![image](https://user-images.githubusercontent.com/10757559/113830921-fb234c00-977e-11eb-91af-f81f28f0f571.png)

#### IntelliJ
![image](https://user-images.githubusercontent.com/10757559/113831228-49384f80-977f-11eb-8b08-33f4dc51bdd0.png)

Issue #2189, opened 2021-02-24
Implement retrieval of rule metadata


duncanp-sonar:
### Description
Extract the rule configuration file(s) from the jar so we can pass them to the `eslintbridge`.
There is one file that contains the rule metadata for both JavaScript and TypeScript.

TODO
- [x] locate and load metadata file
- [x] add export for metadata file path
- [x] add data classes for deserialized rules

~~- [ ] add data classes for rules parameters~~
UPDATE: supporting parameterised rules turns out to be problematic so it's been moved to a separate ticket. See #2284.

PR #2275, opened 2021-03-31
Add EslintBridgeIssueConverter

rita-sonarsource:
Part of #2271

### DIFF
--- a/src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
+++ b/src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
@@ -1,0 +1,83 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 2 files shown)


src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
```diff
@@ -0,0 +1,83 @@
...
+    internal class EslintBridgeIssueConverter : IEslintBridgeIssueConverter
+    {
+        public IAnalysisIssue Convert(string filePath, Issue issue)
+        {
+            // todo: get values from rule configuration
+            var ruleSeverity = AnalysisIssueSeverity.Info;
+            var ruleType = AnalysisIssueType.Vulnerability;
+
+            // todo: calculate line hash
+            string lineHash = null;
```
rita-sonarsource, line 43: I'll open a separate task for this.

src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        IAnalysisIssue Convert(string filePath, Issue issue);
+    }
+
+    internal class EslintBridgeIssueConverter : IEslintBridgeIssueConverter
+    {
+        public IAnalysisIssue Convert(string filePath, Issue issue)
+        {
+            // todo: get values from rule configuration
+            var ruleSeverity = AnalysisIssueSeverity.Info;
```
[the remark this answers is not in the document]
duncanp-sonar: It'll be a combination of #2189 and the standalone config ticket, I think.

src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
```diff
@@ -0,0 +1,83 @@
...
+            var ruleType = AnalysisIssueType.Vulnerability;
+
+            // todo: calculate line hash
+            string lineHash = null;
+
+            return new AnalysisIssue(
+                issue.RuleId,
```
duncanp-sonar, line 46: This is the eslint-bridge rule identity, isn't it?

If it is: 
* maybe we should change the property name to make it clearer; and
* add another "TODO" to handle the rule key conversion (#2191)

duncanp-sonar (approved):

PR #2278, opened 2021-03-31
Fetch TypeScript rules config from embedded file

duncanp-sonar:
Part of #2189


src/TypeScript.UnitTests/Rules/SampleRuleMetadata.json
```diff
@@ -0,0 +1,56 @@
...
...onst express \\u003d require(\\u0027express\\u0027);\\n\\nlet csrfProtection \\u003d csrf({ cookie: { secure: true }}); // Compliant\\n\\u003c/pre\\u003e\\n\\u003ch2\\u003eSee\\u003c/h2\\u003e\\n\\u003cul\\u003e\\n  \\u003cli\\u003e \\u003ca href\\u003d'https://www.owasp.org/index.php/Top_10-2017_A3-Sensitive_Data_Exposure'\\u003eOWASP Top 10 2017 Category A3\\u003c/a\\u003e - Sensitive Data Exposure\\n  \\u003c/li\\u003e\\n  \\u003cli\\u003e \\u003ca href\\u003d'http://cwe.mitre.org/data/definitions/311'\\u003eMITRE, CWE-311\\u003c/a\\u003e - Missing Encryption of Sensitive Data \\u003c/li\\u003e\\n  \\u003cli\\u003e \\u003ca href\\u003d'http://cwe.mitre.org/data/definitions/315'\\u003eMITRE, CWE-315\\u003c/a\\u003e - Cleartext Storage of Sensitive Information in a Cookie \\u003c/li\\u003e\\n  \\u003cli\\u003e \\u003ca href\\u003d'http://cwe.mitre.org/data/definitions/614'\\u003eMITRE, CWE-614\\u003c/a\\u003e - Sensitive Cookie in HTTPS Session Without \\u0027Secure\\u0027 Attribute \\u003c/li\\u003e\\n  \\u003cli\\u003e \\u003ca href\\u003d'https://www.sans.org/top25-software-errors/#cat3'\\u003eSANS Top 25\\u003c/a\\u003e - Porous Defenses \\u003c/li\\u003e\\n\\u003c/ul\\u003e",
```
duncanp-sonar, line 6: We don't currently need the HTML description so I haven't bothered loading and converting it.

src/EmbeddedVsix.UnitTests/TypeScriptPathsProviderTests.cs
```diff
@@ -31,7 +31,7 @@ public class TypeScriptPathsProviderTests
     {
         [TestMethod]
         [DataRow("SonarLint.TypeScript.EsLintBridgeServerPath", @"\ts\bin\server")]
-        [DataRow("SonarLint.TypeScript.TypeScriptRulesMetadataFilePath", @"\ts\sonarlint-metadata.json")]
```
duncanp-sonar, line 34: The file contains data for both languages, not just TypeScript.

src/TypeScript/Rules/RuleDefinitionsProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+        IEnumerable<RuleDefinition> GetDefinitions();
+    }
+
+    internal interface IJavaScriptRuleDefinitionsProvider
+    {
+        /// <summary>
+        /// Returns the metadata descriptions for all JavaScript rules
+        /// </summary>
+        IEnumerable<RuleDefinition> GetDefinitions();
+    }
```
duncanp-sonar, line 44: Separate interfaces for TS and JS.
It's not possible for those interfaces to inherit from a common interface; the compiler complains, even if you try to implement the interfaces explicitly.

In fact, if the interfaces do inherit from a common base you can't explicitly implement `ITypeScriptXXX.GetX` and `IJavaScriptXXX.GetX` separately - you have to implement `IBaseInterface.GetX` i.e. both methods would have the same implementation, which isn't what we want.

duncanp-sonar:
I'll look at fetching the parameters in a separate PR.

src/TypeScript/Rules/RuleDefinitionsProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+        private readonly IReadOnlyCollection<RuleDefinition> tsRules;
+        private readonly IReadOnlyCollection<RuleDefinition> jsRules;
+
+        private const string RepoPrefix_TypeScript = "typescript:";
+        private const string RepoPrefix_JavaScript = "javascript:";
+
+        [ImportingConstructor]
+        public RuleDefinitionsProvider([Import(RuleDefinitionsFilePathContractName)] string typeScriptMetadataFilePath)
+        {
+            var allRules = Load(typeScriptMetadataFilePath);
```
rita-sonarsource, line 63: Since this is loaded in the CTOR, if we fail to deserialize the json the user will see the extension load failure popup.

I think it's better to catch it and return empty rules. WDYT?

src/TypeScript/Rules/RuleDefiniton.cs
```diff
@@ -0,0 +1,85 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Newtonsoft.Json;
+
+// Data classes used to deserialize rule metadata from the json file embedded the SonarJS 
+
+namespace SonarLint.VisualStudio.TypeScript.Rules
+{
+    internal class RuleDefinition
```
rita-sonarsource, line 27: Minor: typo in the file's name (definiton)

src/TypeScript.UnitTests/Rules/RuleDefinitionsProviderTests.cs
```diff
@@ -0,0 +1,122 @@
...
+            var testSubject = new RuleDefinitionsProvider(jsonFilePath);
+
+            // 1. TypeScript
+            var tsProvider = (ITypeScriptRuleDefinitionsProvider)testSubject;
+            var tsRuleKeys = tsProvider.GetDefinitions()
+                .Select(x => x.RuleKey);
+
+            tsRuleKeys.Should().BeEquivalentTo("typescript:S2092", "typescript:S3524", "TypeSCRIPT:S1135");
+
+            // 2. TypeScript
```
rita-sonarsource, line 69: Wrong comment

rita-sonarsource (approved):

rita-sonarsource:
@duncanp-sonar just see my comment regarding the failure in the MEF ctor.

src/TypeScript/Rules/RuleDefinitionsProvider.cs
duncanp-sonar: Good point. Returning an empty list isn't going to give a great UX either, though. How will the user know that something is wrong? The other components will continue to work, the log will show that analysis is being triggered and no issues were found etc.
We'd have to make the caller check for an empty set of rules and warn the user somehow.

Something would have had to go badly wrong for us to ship a version with invalid json (both with the SonarJS code that generates the file, and our manual testing). We could also write more integration level tests to make sure the embedded file is valid, in the same way we do for the CFamily data files.

Issues could also occur at runtime on the user's machine e.g. due to file corruption/ a bad installation. I don't think we should try to carry on in that case, though.

For now, I'd rather let the extension fail if the file can't be deserialized.
I'll add a task to write a sanity check test for the embedded file.

How does that sound?

rita-sonarsource: Yep, we could do something like we do for CFamily. Maybe not just the rules, but just that the json is deserialized and the list is not empty for JS and TS. Sounds good.

PR #2279, opened 2021-04-01
Implement JavaScriptAnalyzer

rita-sonarsource:
Part of #2271


src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -79,24 +79,16 @@ public Task InitLinter(IEnumerable<Rule> rules, CancellationToken cancellationTo
...
+                IgnoreHeaderComments = true,
+                TSConfigFilePaths = Array.Empty<string>() // eslint-bridge generates a default tsconfig for JS analysis
+            };
 
-                var responseString = await httpWrapper.PostAsync("analyze-js", analysisRequest, cancellationToken);
+            var responseString = await httpWrapper.PostAsync("analyze-js", analysisRequest, cancellationToken);
 
-                return responseString == null ? null : JsonConvert.DeserializeObject<AnalysisResponse>(responseString);
-            }
-            catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
```
rita-sonarsource, line 95: I decided to remove the try-catch here. I need an outer try-catch in the caller (JavaScriptAnalyzer) anyway and this one doesn't add anything.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            ExecuteAnalysis(path, consumer, cancellationToken).Forget(); // fire and forget
+        }
+
+        internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, CancellationToken cancellationToken)
+        {
+            // Switch to a background thread
+            await TaskScheduler.Default;
+
+            try
```
rita-sonarsource, line 79: In [CLangAnalyzer](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/CFamily/CLangAnalyzer.cs#L132) we have a status bar notifier but I think it should move to AnalyzerController in some form. 
This way all the analyzers could benefit from it, instead of repeating it in the new JS and TS analyzers.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                var analysisResponse = await eslintBridgeClient.AnalyzeJs(filePath, cancellationToken);
```
rita-sonarsource, line 81: I'm not calling InitLinter yet in any place, I'm not sure where it should be called. Probably in the class that will be responsible for loading the user's settings (and call init-linter again if they change).

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -0,0 +1,130 @@
...
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.TypeScript.EslintBridgeClient;
+using SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract;
+
+namespace SonarLint.VisualStudio.TypeScript.Analyzer
+{
+    [Export(typeof(IAnalyzer))]
+    internal class JavaScriptAnalyzer : IAnalyzer
```
rita-sonarsource, line 38: I'll remove the deamon analyzer in the next PR.

src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
```diff
@@ -0,0 +1,298 @@
...
+
+            logger.AssertPartialOutputStringExists(TypeScript.Analyzer.Resources.ERR_ParsingError_UnsupportedTypescript);
+        }
+
+        [TestMethod]
+        [DataRow(ParsingErrorCode.FAILING_TYPESCRIPT)]
+        [DataRow(ParsingErrorCode.GENERAL_ERROR)]
+        [DataRow(ParsingErrorCode.PARSING)]
+        [DataRow((ParsingErrorCode)1234)]
+        public async Task ExecuteAnalysis_ResponseWithParsingError_ParsingErrorLogged(int errorCode)
```
rita-sonarsource, line 150: I can't receive ParsingErrorCode because it's internal; if I change the method to internal, it's not being executed, and if I change the enum to public, it changes a bunch of the contract classes. So I just get it as int.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
duncanp-sonar: I'd create a ticket for the status notification and leave it until later. I don't think we'd delay shipping for that.
However, we need at least a minimal UX to indicate something is happening, so we should log something.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
duncanp-sonar: We need to call init-linter:
* when eslint-bridge is started, and
* when the user settings change _if_ eslint-bridge is already running.

I think the classes loading the user's settings should be as thin as possible; their job should be to work out what the active settings are when asked. I don't think they should need to know about the eslint-bridge or track its status.

Maybe the `EslinttBridgeProcess`?

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
rita-sonarsource: Done #2280

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
rita-sonarsource: I'll create a ticket for that.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
duncanp-sonar: It might be neater to strip out all of the redundant daemon code at once.
It doesn't have to be done immediately; just commenting out the export of the old analyzer would be enough to disable it.

duncanp-sonar (approved):

PR #2283, opened 2021-04-06
Added sanity check unit test for TS/JS rules metadata

duncanp-sonar:


duncanp-sonar:
I'll squash this as-is because I want to use it to validate the new SonarJS jar provided by Tibor. We can change it later if required.

PR #2285, opened 2021-04-07
[1] Added eslint->sonar key mappers

duncanp-sonar:
Relates to #2191


src/TypeScript/Rules/RuleDefinitionProviders.cs
```diff
@@ -0,0 +1,40 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+
+namespace SonarLint.VisualStudio.TypeScript.Rules
```
duncanp-sonar, line 23: No changes - just moved from the RulesDefinitionsProvider class to reduce clutter.

src/TypeScript/Rules/RuleDefinitionsProvider.cs
```diff
@@ -20,33 +20,22 @@
...
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text;
 using Newtonsoft.Json;
 
 namespace SonarLint.VisualStudio.TypeScript.Rules
 {
```
duncanp-sonar, line 30: This class now implements four interfaces. I put the interfaces in separate files to reduce the noise here.

src/TypeScript/Rules/RuleDefinitionsProvider.cs
```diff
@@ -69,12 +58,38 @@ public RuleDefinitionsProvider([Import(RuleDefinitionsFilePathContractName)] str
...
+        private static string GetRuleKey(string eslintRuleKey, IEnumerable<RuleDefinition> rules)
+        {
+            var repoAndKey = rules.FirstOrDefault(x => eslintRuleKey.Equals(x.EslintKey, System.StringComparison.OrdinalIgnoreCase))
+                ?.RuleKey;
+            return repoAndKey == null ? null : StripRepoPrefix(repoAndKey);
+        }
+
+        private static string StripRepoPrefix(string ruleKey)
+        {
+            // Expecting the key to be in the form [repo id]:[rule id]
```
duncanp-sonar, line 84: I decided to strip the repo prefix so we'll display e.g. `S123` in the Error List rather than `javascript:S123`. It felt more in keeping with what VS developers would expect.

In contrast, the other SL flavours don't strip the repo prefix, but then they don't display the error codes in the same way either.

rita-sonarsource: It's not what we currently do for taint/hotspots/cfamily but I agree that it's neater. Do you think it's worth opening an issue(s) to align the other 3?

src/TypeScript/Rules/RuleDefinitionsProvider.cs
```diff
@@ -20,33 +20,22 @@
...
-        IEnumerable<RuleDefinition> GetDefinitions();
-    }
-
     [Export(typeof(ITypeScriptRuleDefinitionsProvider))]
     [Export(typeof(IJavaScriptRuleDefinitionsProvider))]
+    [Export(typeof(ITypeScriptRuleKeyMapper))]
+    [Export(typeof(IJavaScriptRuleKeyMapper))]
     [PartCreationPolicy(CreationPolicy.Shared)]
-    internal class RuleDefinitionsProvider : ITypeScriptRuleDefinitionsProvider, IJavaScriptRuleDefinitionsProvider
+    internal class RuleDefinitionsProvider :
```
rita-sonarsource, line 36: It's a bit confusing that we have 2 file names that are nearly identical. Maybe we should rename this one, i.e. something like `EslintBridgeRulesRepository`, or `JsTsRulesContainer`, etc.

rita-sonarsource (approved):

PR #2286, opened 2021-04-07
[2] Modified JS issue converter to use key mapper

duncanp-sonar:
Relates to #2271 and #2191


src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
```diff
@@ -31,17 +31,24 @@ internal interface IEslintBridgeIssueConverter
         IAnalysisIssue Convert(string filePath, Issue issue);
     }
 
+    internal delegate string ConvertToSonarRuleKey(string eslineRuleKey);
+
     internal class EslintBridgeIssueConverter : IEslintBridgeIssueConverter
     {
+        private readonly ConvertToSonarRuleKey keyMapper;
+
+        public EslintBridgeIssueConverter(ConvertToSonarRuleKey keyMapper)
```
duncanp-sonar, line 40: We have a mis-match here; currently, we only have one EsLintBridgeConverter, but we have two key mappers (one for TS one for JS).

We could perhaps get away with one key mapper, if it is the case that all rules that have the same eslint rule key have the same Sonar `Sxxx` rule number (i.e. only the repo name is different). However, we can always refactor it later if necessary.

There's another outstanding TODO in the EslintBridgeIssueConverter for fetching rule severities and types from the current rules config, which will also affect how the class is constructed.

rita-sonarsource: Hmm good catch, we'll need to pass that too. And we had a problem to create a unified interface, right? I'll take a look at everything.

rita-sonarsource (approved):

src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
duncanp-sonar: Yes: we could _define_ a common interface, but we couldn't then have a single class that implemented both derived interfaces.

Issue #2289, opened 2021-04-08
[JavaScript] Implement active rule configuration provider


duncanp-sonar:
### Description

We need to pass `eslintbridge:init-linter` the configuration for the set of rules to run and the rule parameters to use.

The active configuration should be based on the default rules configuration, modified by the local user settings in the case of standalone mode, and by the QP configuration from the server in the case of connected mode.

However, user-configuration (#771) and connected mode (#770) will be handled separately. The scope of this ticket is limited to providing the appropriate configuration based on the default rules configuration only.

#### Related tickets:
* user-configuration in standalone mode #771
* connected mode configuration #770 
* parameterised rules #2284 

#### TODO
- [ ] implement active configuration provider, per language

PR #2294, opened 2021-04-08
Added JavaScript active rules config provider

duncanp-sonar:
Relates to #2289

### DIFF
--- a/src/TypeScript/Rules/ActiveJavaScriptRulesProvider.cs
+++ b/src/TypeScript/Rules/ActiveJavaScriptRulesProvider.cs
@@ -1,0 +1,78 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


src/TypeScript/Rules/ActiveJavaScriptRulesProvider.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        /// <summary>
+        /// Returns the eslint configuration for the currently active rules
+        /// </summary>
+        IEnumerable<Rule> Get();
+    }
+
+    [Export(typeof(IActiveJavaScriptRulesProvider))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class ActiveJavaScriptRulesProvider : IActiveJavaScriptRulesProvider
```
duncanp-sonar, line 39: This is a minimal implementation to allow us to get an end-to-end analysis working.

It works directly against the list of rule definitions and does some simple filtering for rules we know we don't want to run i.e. not active, hotspot.

I haven't bothered with any perf optimisations at this stage so the list is recalculated each time, even though it will currently return the same list.
That will change once we start implementing some of the TODOs.

rita-sonarsource (approved):

PR #2295, opened 2021-04-09
[1] Use rule definition in issue conversion

rita-sonarsource:
Part of #2189

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueToAnalysisIssueConverterTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueToAnalysisIssueConverterTests.cs
@@ -369,9 +369,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
         public void ConvertFromIssueType(IssueType cfamilyIssueType, AnalysisIssueType analysisIssueType)
         {
             CFamilyIssueToAnalysisIssueConverter.Convert(cfamilyIssueType).Should().Be(analysisIssueType);
-
-            Action act = () => CFamilyIssueToAnalysisIssueConverter.Convert((IssueType)(-1));
-            act.Should().ThrowExactly<ArgumentOutOfRangeException>().And.ParamName.Should().Be("issueType");
         }
 
         [TestMethod]
--- a/src/TypeScript.UnitTests/Analyzer/EslintBridgeIssueConverterTests.cs
+++ b/src/TypeScript.UnitTests/Analyzer/EslintBridgeIssueConverterTests.cs
@@ -18,18 +18,22 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Generic;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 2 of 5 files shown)


src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueToAnalysisIssueConverterTests.cs
```diff
@@ -369,9 +369,6 @@ public void ConvertFromIssueSeverity_InvalidValue_Throws()
         public void ConvertFromIssueType(IssueType cfamilyIssueType, AnalysisIssueType analysisIssueType)
         {
             CFamilyIssueToAnalysisIssueConverter.Convert(cfamilyIssueType).Should().Be(analysisIssueType);
-
```
rita-sonarsource, line 372: Misc: noticed this when blatantly copying the tests; it's a redundant assertion that belongs to the test below.

src/TypeScript.UnitTests/Analyzer/EslintBridgeIssueConverterTests.cs
```diff
@@ -18,18 +18,22 @@
...
 using SonarLint.VisualStudio.TypeScript.Analyzer;
 using SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract;
+using SonarLint.VisualStudio.TypeScript.Rules;
 
 namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
 {
     [TestClass]
     public class EslintBridgeIssueConverterTests
     {
+        private const string TestEslintBridgeRuleId = "rule id";
```
rita-sonarsource, line 35: A bit ugly but I don't have a neat way to mock the collection and it's simpler.

src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
```diff
@@ -32,28 +33,29 @@ internal interface IEslintBridgeIssueConverter
...
     internal delegate string ConvertToSonarRuleKey(string eslineRuleKey);
+    internal delegate IEnumerable<RuleDefinition> GetRuleDefinitions();
 
     internal class EslintBridgeIssueConverter : IEslintBridgeIssueConverter
     {
         private readonly ConvertToSonarRuleKey keyMapper;
+        private readonly GetRuleDefinitions ruleDefinitionProvider;
 
-        public EslintBridgeIssueConverter(ConvertToSonarRuleKey keyMapper)
+        public EslintBridgeIssueConverter(ConvertToSonarRuleKey keyMapper, GetRuleDefinitions ruleDefinitionProvider)
```
duncanp-sonar, line 43: Strictly speaking this should be using the current active rule configuration (assuming we allow the user to change the severity; they can for C#, and I think they can for C++).
However, we can change that when we support user configuration.

src/TypeScript.UnitTests/Analyzer/EslintBridgeIssueConverterTests.cs
duncanp-sonar: I think it's ok, but I'd change the name to better indicate that the value isn't important as it isn't checked; it just needs to be valid e.g. `ValidEsLintKey`, `AnyEslintKey`.

duncanp-sonar (approved):

Issue #2281, opened 2021-04-01
Call init-linter for JS analyzer


rita-sonarsource:
TODO:

Call Eslint-bridge's `init-linter` method in the following use cases:
- [x] After the server starts and before the first analysis is triggered
- [ ] ~~After any change to the user settings~~ --> will be done after supporting rules configuration #771

PR #2296, opened 2021-04-09
[2] Move EslintBridgeProcess to JavaScriptAnalyzer

rita-sonarsource:
Prep for #2281

### DIFF
--- a/src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
+++ b/src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
@@ -81,7 +81,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
             var client = SetupEslintBridgeClient(null);
             var clientFactory = SetupEslintBridgeClientFactory(123, client.Object);
             
-            var testSubject = CreateTestSubject(clientFactory, serverProcess.Object);
+            var testSubject = CreateTestSubject(clientFactory.Object, serverProcess.Object);
 
             await testSubject.ExecuteAnalysis("some path", Mock.Of<IIssueConsumer>(), CancellationToken.None);
 
@@ -138,6 +138,31 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
 
             client1.Verify(x => x.Dispose(), Times.Once);
             client2.Verify(x => x.Dispose(), Times.Never);
+        }
+
+        [TestMethod]
+        public async Task ExecuteAnalysis_EslintBridgePortUnchanged_DoesNotCreateNewClient()
+        {
+            const int port = 123;
+
+            var serverProcess = SetupServerProcess(port);
+            var client = SetupEslintBridgeClient(null);
+            var clientFactory = SetupEslintBridgeClientFactory(port, client.Object);
+
+            var testSubject = CreateTestSubject(clientFactory.Object, serverProcess.Object);
+
+            await testSubject.ExecuteAnalysis("some path", Mock.Of<IIssueConsumer>(), CancellationToken.None);
+
+            client.VerifyAll();
+            clientFactory.Verify(x => x.Create(port), Times.Once);
+            clientFactory.VerifyNoOtherCalls();
+
+            await testSubject.ExecuteAnalysis("some path", Mock.Of<IIssueConsumer>(), CancellationToken.None);
+
+            client.VerifyAll();
+            clientFactory.VerifyNoOtherCalls();
+            client.Verify(x => x.Dispose(), Times.Never);
+            client.VerifyNoOtherCalls();
         }
 
         [TestMethod]
@@ -329,7 +354,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
             var client = SetupEslintBridgeClient(null);
             var clientFactory = SetupEslintBridgeClientFactory(123, client.Object);
 
-            var testSubject = CreateTestSubject(clientFactory, serverProcess.Object);
+            var testSubject = CreateTestSubject(clientFactory.Object, serverProcess.Object);
 
             Action act = () => testSubject.Dispose();
             act.Should().NotThrow();
@@ -345,7 +370,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
             var client = SetupEslintBridgeClient(null);
             var clientFactory = SetupEslintBridgeClientFactory(123, client.Object);
 
-            var testSubject = CreateTestSubject(clientFactory, serverProcess.Object);
+            var testSubject = CreateTestSubject(clientFactory.Object, serverProcess.Object);
 
             await testSubject.ExecuteAnalysis("some path", Mock.Of<IIssueConsumer>(), CancellationToken.None);
             testSubject.Dispose();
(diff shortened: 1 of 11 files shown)


src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
```diff
@@ -34,40 +33,34 @@ namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
...
 
-        public async Task<string> PostAsync(string serverEndpoint, object request, CancellationToken cancellationToken)
+        public async Task<string> PostAsync(Uri serverEndpoint, object request, CancellationToken cancellationToken)
         {
             try
             {
                 var serializedRequest = request == null ? string.Empty : JsonConvert.SerializeObject(request, Formatting.Indented);
                 logger.LogDebug(Resources.INFO_RequestDetails, serverEndpoint, serializedRequest);
 
-                var port = await eslintBridgeProcess.Start();
```
rita-sonarsource, line 68: As we discussed offline, the wrapper is now dumb and accepts the given uri, without starting the process.
JavaScriptAnalyzer is the one that will call `eslintBridgeProcess.Start();`

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -40,21 +40,22 @@ internal interface IEslintBridgeClient : IDisposable
...
     internal sealed class EslintBridgeClient : IEslintBridgeClient
     {
+        private readonly Uri baseServerUri;
         private readonly IEslintBridgeHttpWrapper httpWrapper;
         private readonly IAnalysisConfiguration analysisConfiguration;
 
-        [ImportingConstructor]
-        public EslintBridgeClient(IEslintBridgeHttpWrapper httpWrapper)
-            : this(httpWrapper, new AnalysisConfiguration())
+        public EslintBridgeClient(Uri baseServerUri, ILogger logger)
```
rita-sonarsource, line 49: This is now a utility class, similar to the HttpWrapper.
Since the process and can die and start on a new port, it receives the server uri in the ctor, so you need to create a new one if the port has changed. This is done by JavaScriptAnalyzer.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -36,29 +36,36 @@
...
         private readonly IEslintBridgeIssueConverter issuesConverter;
         private readonly ILogger logger;
 
+        private IEslintBridgeClient javascriptClient;
+        private int javascriptServerPort;
+
         [ImportingConstructor]
-        public JavaScriptAnalyzer(IEslintBridgeClient eslintBridgeClient,
+        public JavaScriptAnalyzer(IEslintBridgeClientFactory eslintBridgeClientFactory,
+            IEslintBridgeProcess eslintBridgeProcess,
```
rita-sonarsource, line 51: IEslintBridgeProcess is still a singleton, I will change that in the next PR and create IEslintBridgeProcessFactory that it will call and hold reference to.

src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
```diff
@@ -73,6 +74,72 @@ public void IsAnalysisSupported_HasJavascript_True()
...
+            var testSubject = CreateTestSubject(clientFactory, serverProcess.Object);
+
+            await testSubject.ExecuteAnalysis("some path", Mock.Of<IIssueConsumer>(), CancellationToken.None);
+
+            serverProcess.Verify(x => x.Start(), Times.Once);
+            serverProcess.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public async Task ExecuteAnalysis_FailsToStartsEslintBridgeServer_ExceptionCaughtAndLogged()
```
rita-sonarsource, line 93: This test might be redundant, we already have `ExecuteAnalysis_NonCriticalException_ExceptionCaughtAndLogged` and `ExecuteAnalysis_CriticalException_ExceptionThrown`, so I'm not sure it adds much value.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -113,6 +121,20 @@ internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, Ca
...
         }
 
+        private async Task<IEslintBridgeClient> GetJavaScriptClient()
+        {
+            var port = await eslintBridgeProcess.Start();
+
+            if (port != javascriptServerPort)
+            {
+                javascriptServerPort = port;
+                javascriptClient?.Dispose();
```
rita-sonarsource, line 131: There is a problem at the moment with the [EslintBridgeClient.Dispose implementation](https://github.com/SonarSource/sonarlint-visualstudio/blob/branch-replacejs/src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs#L95) -- it attempts to call the "Close" endpoint, but if the process died, the client will fail and not dispose the httpWrapper. I think I should either remove the Close from the dispose, or wrap it with try-catch. Leaning towards removing the Close, WDYT?

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -113,6 +121,20 @@ internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, Ca
             }
         }
 
+        private async Task<IEslintBridgeClient> GetJavaScriptClient()
```
rita-sonarsource, line 124: I think this method should be wrapped with a lock, to avoid a race condition of multiple file analysis that can happen if the process not started yet or died.
E.g.: open a few files, kill the process, change settings.json --> will spawn multiple analysis that each will attempt to start the server. Perhaps can also happen with first analysis (e.g. open file and immediately save it --> 2 analysis)

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
duncanp-sonar: Hhmmm. What is going to responsible for telling the eslintbridge to close down cleanly? Something should - we can't just terminate the process.
To put it another way, what is responsible for managing the lifetime of the eslintbridge?

I think there is still something missing, but I'm not sure quite what.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
duncanp-sonar: What's the purpose of `IEslintBridgeProcessFactory` i.e. what would call it and when?
Currently, `EslintBridgeProcess` handles starting a new process if the previous one exited. What does the factory add?

src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
duncanp-sonar: It's not adding a lot from a technical point of view. It is testing a specific end-user scenario i.e. a known type of failure, but it's not adding to the overall test coverage.
I probably wouldn't keep it, but I don't have a strong feeling about it.

duncanp-sonar (commented):
It still feels like there is something wrong with the design, but moving the process handling out of the http wrapper is definitely a step in the right direction.
I'd commit it as-is and we can refactor more later as necessary.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
duncanp-sonar: I don't think we need to worry about the "open then save" case; I think the TextBufferIssueTracker should handle that for us (via the slow-memory-leaking-job-scheduler code).

Changing the settings might cause an issue (when we get round to supporting it).

However, perhaps the `eslintBridgeProcess` should handle it. It's returning a task that could be awaited by multiple callers, so it should protect itself against race conditions so the callers don't have to.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
rita-sonarsource: A new instance for `EslintBridgeProcess`, for typescript. On second thought, I will do it when we get to typescript.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
rita-sonarsource: At the moment, nothing. We discussed that we would want to terminate the process on solution close, but it's not implement yet so nothing calls Close except Dispose.

PR #2297, opened 2021-04-09
[3] Call init-linter

### LINKED ISSUE #2281: Call init-linter for JS analyzer

rita-sonarsource:
Fixes #2281

### DIFF
--- a/src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
+++ b/src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
@@ -91,21 +91,6 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
         }
 
         [TestMethod]
-        public async Task ExecuteAnalysis_FailsToStartsEslintBridgeServer_ExceptionCaughtAndLogged()
-        {
-            var serverProcess = SetupServerProcess(exceptionToThrow: new NotImplementedException("this is a test"));
-            var clientFactory = new Mock<IEslintBridgeClientFactory>();
-            var logger = new TestLogger();
-
-            var testSubject = CreateTestSubject(clientFactory.Object, serverProcess.Object, logger: logger);
-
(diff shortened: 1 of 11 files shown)


src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
```diff
@@ -153,27 +154,52 @@ public async Task ExecuteAnalysis_EslintBridgePortUnchanged_DoesNotCreateNewClie
 
             await testSubject.ExecuteAnalysis("some path", Mock.Of<IIssueConsumer>(), CancellationToken.None);
 
-            client.VerifyAll();
             clientFactory.Verify(x => x.Create(port), Times.Once);
-            clientFactory.VerifyNoOtherCalls();
+            clientFactory.Verify(x => x.Create(It.IsAny<int>()), Times.Once);
```
rita-sonarsource, line 158: There is something odd happening with the mocks: `clientFactory` is a mock that returns the `client` mock, and when I call `clientFactory.VerifyNoOtherCalls();`, it fails on unverified expectations on `client`. That's why I previously had `client.VerifyAll();`, but now it fails because I didn't setup & verify InitLinter (I didn't because it's void and I don't care in this test).
So in the end I just opted to check that Create was called once and only once.

duncanp-sonar: I encountered something similar in the past; I can't remember the precise situation, but it did seem to make sense once I'd worked out what was happening. I think it might have been to do with "nested" objects e.g. calling `VerifyNoOtherCalls` on a "parent" mock also verified "child" mocks.
I think I probably did what you have done - just verify the specific call I cared about.

duncanp-sonar (approved):

PR #2304, opened 2021-04-12
Handle EslintBridgeProcess failures

### LINKED ISSUE #2300: EslintBridgeProcess does not start a new process when the previous process is killed
#### Description

EslintBridgeProcess does not check if the process is still running. So if the process was killed/existed, it will not create a new one.

TODO:
- [x] Handle process failures.

rita-sonarsource:
Fixes #2300

### DIFF
--- a/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
@@ -24,6 +24,7 @@ using System.IO.Abstractions;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.Threading;
 using Moq;
 using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.Integration.UnitTests;
@@ -55,6 +56,54 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
             Func<Task> act = async () => await testSubject.Start();
 
             act.Should().ThrowExactly<FileNotFoundException>().And.Message.Should().Contain("node.exe");
+        }
+
+        [TestMethod]
+        public async Task Start_SecondCall_ProcessFailed_StartsNewProcess()
+        {
+            var fakeNodeExePath = CreateScriptThatPrintsPortNumber(123, isHanging: true);
+            var startupScriptPath = "dummy path";
+
+            var nodeLocator = SetupNodeLocator(fakeNodeExePath);
+            var testSubject = CreateTestSubject(startupScriptPath, nodeLocator: nodeLocator.Object);
+
+            await testSubject.Start();
+
+            var oldProcess = testSubject.Process;
+
+            oldProcess.HasExited.Should().BeFalse();
+            oldProcess.Kill();
+            await oldProcess.WaitForExitAsync();
+            oldProcess.HasExited.Should().BeTrue();
+
+            await testSubject.Start();
+
+            var newProcess = testSubject.Process;
+            newProcess.Should().NotBeSameAs(oldProcess);
+            newProcess.HasExited.Should().BeFalse();
+            newProcess.Kill();
+        }
+
+        [TestMethod]
+        public async Task Start_SecondCall_ProcessIsStillRunning_DoesNotStartNewProcess()
+        {
+            var fakeNodeExePath = CreateScriptThatPrintsPortNumber(123, isHanging: true);
+            var startupScriptPath = "dummy path";
+
+            var nodeLocator = SetupNodeLocator(fakeNodeExePath);
+            var testSubject = CreateTestSubject(startupScriptPath, nodeLocator: nodeLocator.Object);
+
+            await testSubject.Start();
+
+            var oldProcess = testSubject.Process;
+            oldProcess.HasExited.Should().BeFalse();
+
+            await testSubject.Start();
+
+            var newProcess = testSubject.Process;
+            newProcess.Should().BeSameAs(oldProcess);
+            newProcess.HasExited.Should().BeFalse();
+            newProcess.Kill();
         }
 
         [TestMethod]
--- a/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
@@ -67,21 +67,29 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
             var nodeLocator = SetupNodeLocator(fakeNodeExePath);
             var testSubject = CreateTestSubject(startupScriptPath, nodeLocator: nodeLocator.Object);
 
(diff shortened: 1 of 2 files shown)


src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -74,27 +75,38 @@ internal sealed class EslintBridgeProcess : IEslintBridgeProcess
 
         public Task<int> Start()
         {
-            if (startTask != null && !startTask.Task.IsFaulted)
+            lock (Lock)
```
rita-sonarsource, line 78: Should prevent spawning multiple processes during first time analysis/after failure.

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -74,27 +75,38 @@ internal sealed class EslintBridgeProcess : IEslintBridgeProcess
...
             {
-                return startTask.Task;
-            }
+                var shouldSpawnNewProcess = startTask == null ||
+                                            startTask.Task.IsFaulted ||
+                                            Process == null ||
```
rita-sonarsource, line 82: `Process == null || Process.HasExited` was not handled before

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -162,17 +180,23 @@ private void Stop()
         {
             logger.LogDebug(Resources.INFO_TerminatingServer);
 
+            if (Process != null)
```
rita-sonarsource, line 183: Wasn't this a memory leak? we have another one in the cfamily ProcessRunner.

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -162,17 +180,23 @@ private void Stop()
...
+            {
+                Process.ErrorDataReceived -= OnErrorDataReceived;
+                Process.OutputDataReceived -= OnOutputDataReceived;
+                Process.Exited -= Process_Exited;
+            }
             if (Process == null || Process.HasExited)
             {
                 logger.LogDebug(Resources.INFO_ServerAlreadyTerminated);
+            }
+            else
```
rita-sonarsource, line 193: Calling `Process?.Dispose();` on a terminated process can actually throw, so we shouldn't have been calling it all the time

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
duncanp-sonar: Why - because the event handlers weren't explicitly unregistered?

It shouldn't have been; the only thing that has a reference to the Process instance is this class, so once this class sets its `Process` variable to null the process instance is unreachable and eligible for collection.
The fact that the process instance holds references to the `EslintBridgeProcess` won't stop the process instance from being collected. It doesn't hurt to explicitly unregister the event handlers but it isn't strictly necessary here.

The only occasion where it does matter is if the publisher lives longer than the subscriber, in which case the event registrations can keep the subscriber alive.

Or did you mean something else?

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
duncanp-sonar: There's still a race condition here, isn't there? The process could have finished after the `HasExited` check but before the code in the `else` is executed ([the CFamily process-handling code](https://github.com/SonarSource/sonarlint-visualstudio/blob/6b9a8a7623906def168229d9089bf6e17a9c00a4/src/Integration.Vsix/CFamily/ProcessRunner.cs#L167) guards against this).

I don't think there is a safe way of cleaning up a process that doesn't involve exception handling.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
```diff
@@ -57,6 +58,54 @@ public void Start_FailsToFindNode_FileNotFoundException()
...
+            oldProcess.Kill();
+            await oldProcess.WaitForExitAsync();
+            oldProcess.HasExited.Should().BeTrue();
+
+            await testSubject.Start();
+
+            var newProcess = testSubject.Process;
+            newProcess.Should().NotBeSameAs(oldProcess);
+            newProcess.HasExited.Should().BeFalse();
+            newProcess.Kill();
```
duncanp-sonar, line 84: The cleanup code won't be executed if any of the assertions fail. It would be safer to always do the cleanup in a `finally` block.

I think only the `newProcess` has to be conditionally cleaned up in the `finally` block. There don't look to be any paths by which there could be an error that leaves the `oldProcess` running, although it would be safer to make sure.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
```diff
@@ -57,6 +58,54 @@ public void Start_FailsToFindNode_FileNotFoundException()
...
+
+            await testSubject.Start();
+
+            var oldProcess = testSubject.Process;
+            oldProcess.HasExited.Should().BeFalse();
+
+            await testSubject.Start();
+
+            var newProcess = testSubject.Process;
+            newProcess.Should().BeSameAs(oldProcess);
```
duncanp-sonar, line 104: Same cleanup problem: if the assertions fail the processes will be left running.

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -125,6 +138,11 @@ private void StartServer()
             Process.BeginOutputReadLine();
         }
 
+        private void Process_Exited(object sender, EventArgs e)
+        {
+            Stop();
```
duncanp-sonar, line 143: Is it necessary to do anything here except set `Process` to null?

It seems a bit odd to be calling a method called `Stop` when we know the process has already stopped.
If the idea is to put common cleanup code in one place then it might be better to rename the `Stop` method to e.g. `Cleanup` or `EnsureProcessStopped`

duncanp-sonar (commented):
I think both the tests and product code need some tweaking; see comments.

rita-sonarsource:

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
```diff
@@ -57,6 +58,70 @@ public void Start_FailsToFindNode_FileNotFoundException()
...
+
+                var newProcess = testSubject.Process;
+                newProcess.Should().NotBeSameAs(oldProcess);
+                newProcess.HasExited.Should().BeFalse();
+                newProcess.Kill();
+            }
+            finally
+            {
+                // Kill spawned process
+                testSubject.Dispose();
```
duncanp-sonar, line 91: This is indirect cleanup; it's relying on the testSubject being correctly implemented and doing the cleanup for us. The test doesn't need to do that - it created the test processes, so it can ensure that they are cleaned up by killing/disposing them directly.
Same for the other test.

duncanp-sonar (approved):
Just a comment about the test process cleanup.

PR #2306, opened 2021-04-12
Delete daemon environment settings and resources

duncanp-sonar:


src/Core/EnvironmentSettings.cs
```diff
@@ -29,8 +29,6 @@ public class EnvironmentSettings : IEnvironmentSettings
         internal const string PchGenerationTimeoutEnvVar = "SONARLINT_INTERNAL_PCH_GENERATION_TIMEOUT_MS";
         internal const string LogDebugMessagesEnvVar = "SONARLINT_INTERNAL_LOG_DEBUG";
 
-        public const string SonarLintDownloadUrlEnvVar = "SONARLINT_DAEMON_DOWNLOAD_URL";
```
duncanp-sonar, line 32: Simple cleanup of daemon environment settings and unused resources.

src/Integration.Vsix/Resources/Strings.resx
```diff
@@ -129,16 +129,6 @@
     <value>Cannot find tool window with the specified guid: {0}</value>
     <comment>Exception message, {0} guid</comment>
   </data>
-  <data name="Daemon_Downloaded" xml:space="preserve">
```
duncanp-sonar, line 132: There are still a number of strings prefixed with `Daemon_` in this resource file, but they are all still in use. In so far as possible I'll keep deletions and renames in separate commits to make the changes easier to review.

duncanp-sonar:
Some info about the remaining work:
* the "disable rule" only applies to CFamily rules, so it doesn't need to be touched at the moment
* we communicated with the daemon via protobuf, with the protobuf generator being called at build time to create the C# wrapper classes. We obviously don't need that generated code any more, but we might still need to ship the protobuf assemblies (some of the old SQ APIs returned protobuf-serialized data).

rita-sonarsource (approved):

PR #2310, opened 2021-04-13
Remove daemon protobuf files

duncanp-sonar:


duncanp-sonar:
Previously, the VSIX communicated with the SonarLint daemon using GRPC, using protobuf-serialized data.
The protobuf data classes were generated as part of the build.

We no longer need the GRPC or protobuf files for the daemon, but we do still need the protobuf assembly for the SonarQube.Client assembly.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -338,51 +320,9 @@
...
-    </ItemGroup>
-
-  </Target>
-
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
   <Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" Condition="'$(VSToolsPath)' != ''" />
 
-
-  <!-- ********************************************************** -->
-  <!-- Reference Grpc and System.Interactive.Async.dll -->
```
duncanp-sonar, line 361: We no longer need any of this nastiness to include the correct Grpc binaries and dependencies.

The resulting VSIX still includes the protobuf dll, but that is because it is included as a dependency of SonarQube.Client assembly. We don't need to do anything special to embed the correct protobuf assembly in the VSIX.

src/Integration.Vsix/packages.lock.json
```diff
@@ -283,6 +247,16 @@
           "stdole": "7.0.3300"
         }
       },
+      "Google.Protobuf": {
+        "type": "Transitive",
```
duncanp-sonar, line 251: These are now included as transitive dependencies, used by SonarQube.Client.

rita-sonarsource (approved):

Issue #2251, opened 2021-03-25
Remove old JS daemon code + update the options pane


rita-sonarsource:
#### Description

Remove old JS daemon functionality.
TODO:
- [x] Remove daemon code
- [x] Remove Grpc-related code
- [x] Remove daemon-related protobuf code. Do we still need the protobuf code for anything else? (e.g. old SQ client APIs?): UPDATE: yes, but it is still correctly included in the VSIX as a transitive dependency
- [x] Update the options pane and remove the "Activate more.." option
- [x] Remove unused resources
- [x] logger options should always be visible
~~- [ ] Delete daemon files on existing installations? (user might have multiple VS instances with different SLVS versions installed...)~~ Separate ticket created: #2307

PR #2311, opened 2021-04-13
Remove "activate more..." from options page

duncanp-sonar:
Part of #2251 

The Tools Options page used to have a "enabled additional languages" checkbox, with the "log level" combo being conditionally shown/hidden depending on the state of the checkbox. In fact, the log level setting is used by the CFamily analyzer so it should be visible all of the time.

The page is now much simpler - fewer controls, no dynamic showing/hiding.


src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml
```diff
@@ -39,29 +35,8 @@
                    VerticalAlignment="Center"
                    Margin="5,8,0,5" TextWrapping="Wrap" Text="Note: per-user settings are ignored for solutions using connected mode." />
         </StackPanel>
-        <Button Grid.Column="0" Grid.Row="2" Content="Activate more..." Click="OnActivateMoreClicked"
```
duncanp-sonar, line 42: Delete all of the unnecessary rows and put the verbosity panel in the new last row.

src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml.cs
```diff
@@ -32,96 +29,16 @@ namespace SonarLint.VisualStudio.Integration.Vsix
     /// </summary>
     public partial class GeneralOptionsDialogControl : UserControl
     {
-        private readonly ISonarLintSettings settings;
```
duncanp-sonar, line 35: Removed the code that showed/hid the controls depending on the state of the text box. That meant the control no longer needs the settings instance or the logger.

src/TestInfrastructure/Framework/ConfigurableIntegrationSettings.cs
```diff
@@ -25,7 +25,5 @@ public class ConfigurableSonarLintSettings : ISonarLintSettings
         public DaemonLogLevel DaemonLogLevel { get; set; }
 
         public bool IsActivateMoreEnabled { get; set; }
-
-        public bool SkipActivateMoreDialog { get; set; }
```
duncanp-sonar, line 29: Miscellaneous change: the `SkipActiveMoreDialog` was removed from the public interface a long time ago, but was left behind in this test class.

PR #2316, opened 2021-04-13
Handle init-linter request failures

### LINKED ISSUE #2299: init-linter request failures leave the server in an unusable state
#### Description

Currently there is no retry/recovery mechanism on init-linter request. So if eslint-bridge server starts successfully, but init-linter request fails, it leaves the server in an unusable state. The analyzer will not start a new server, yet the existing server would not be usable.

TODO:
- [ ] Handle init-linter request failures.

rita-sonarsource:
Fixes #2299

### DIFF
--- a/src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
+++ b/src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
@@ -48,7 +48,7 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
         private readonly ITelemetryManager telemetryManager;
         private readonly ILogger logger;
 
-        private IEslintBridgeClient javascriptClient;
+        private IEslintBridgeClient eslintBridgeClient;
         private int javascriptServerPort;
         private bool shouldInitLinter;
 
@@ -109,7 +109,8 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
 
             try
             {
-                var eslintBridgeClient = await GetJavaScriptClient(cancellationToken);
+                await EnsureEslintBridgeClientIsInitialized(cancellationToken);
+
                 var analysisResponse = await eslintBridgeClient.AnalyzeJs(filePath, cancellationToken);
 
                 if (analysisResponse == null)
@@ -136,7 +137,7 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
             }
         }
 
-        private async Task<IEslintBridgeClient> GetJavaScriptClient(CancellationToken cancellationToken)
+        private async Task EnsureEslintBridgeClientIsInitialized(CancellationToken cancellationToken)
         {
             var port = await eslintBridgeProcess.Start();
 
@@ -148,13 +149,13 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
                 {
                     javascriptServerPort = port;
                     shouldInitLinter = true;
-                    javascriptClient?.Dispose();
-                    javascriptClient = eslintBridgeClientFactory.Create(javascriptServerPort);
+                    eslintBridgeClient?.Dispose();
+                    eslintBridgeClient = eslintBridgeClientFactory.Create(javascriptServerPort);
                 }
 
                 if (shouldInitLinter)
                 {
-                    await javascriptClient.InitLinter(activeRulesProvider.Get(), cancellationToken);
+                    await eslintBridgeClient.InitLinter(activeRulesProvider.Get(), cancellationToken);
                     shouldInitLinter = false;
                 }
             }
@@ -162,8 +163,6 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
             {
                 serverInitLocker.Set();
             }
-
-            return javascriptClient;
         }
 
         private IEnumerable<IAnalysisIssue> ConvertIssues(string filePath, IEnumerable<Issue> analysisResponseIssues) =>
@@ -191,7 +190,7 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
 
         public void Dispose()
         {
-            javascriptClient?.Dispose();
+            eslintBridgeClient?.Dispose();
             eslintBridgeProcess?.Dispose();
             serverInitLocker?.Dispose();
         }
--- a/src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
+++ b/src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
@@ -39,6 +39,8 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
     [Export(typeof(IAnalyzer))]
     internal sealed class JavaScriptAnalyzer : IAnalyzer, IDisposable
     {
+        private readonly EventWaitHandle serverInitLocker = new EventWaitHandle(true, EventResetMode.AutoReset);
+
         private readonly IEslintBridgeClientFactory eslintBridgeClientFactory;
         private readonly IEslintBridgeProcess eslintBridgeProcess;
         private readonly IActiveJavaScriptRulesProvider activeRulesProvider;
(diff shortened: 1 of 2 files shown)


src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -133,17 +136,31 @@ internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, Ca
...
         {
             var port = await eslintBridgeProcess.Start();
 
-            if (port != javascriptServerPort)
+            try
             {
-                javascriptServerPort = port;
-                javascriptClient?.Dispose();
-                javascriptClient = eslintBridgeClientFactory.Create(javascriptServerPort);
+                serverInitLocker.WaitOne();
```
rita-sonarsource, line 145: I'm breaking my head trying to find a way to test this.
I wrote a test that launches two tasks with two EventWaitHandles that simulate a hang in the `InitLinter`.
However -- my tests still pass even when the second task hasn't even started yet. I'm trying to find a way to make sure that the second task is actually blocked, but I can't think of anything. Some stackoverflow posts recommend adding an internal var "waitingThreadsCount", but I don't want to pollute the product code for this.
I've tried using the `eslintBridgeProcess.Start` but again, the tests pass even when the second task hasn't reached the "WaitOne" block. 
Got any ideas? 🤔

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -133,17 +136,31 @@ internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, Ca
...
-                await javascriptClient.InitLinter(activeRulesProvider.Get(), token);
+                if (shouldInitLinter)
+                {
+                    await javascriptClient.InitLinter(activeRulesProvider.Get(), cancellationToken);
+                    shouldInitLinter = false;
+                }
+            }
+            finally
+            {
+                serverInitLocker.Set();
```
rita-sonarsource, line 163: Because it's an async method, I can't use a regular lock.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -133,17 +136,31 @@ internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, Ca
             }
         }
 
-        private async Task<IEslintBridgeClient> GetJavaScriptClient(CancellationToken token)
+        private async Task<IEslintBridgeClient> GetJavaScriptClient(CancellationToken cancellationToken)
         {
             var port = await eslintBridgeProcess.Start();
```
duncanp-sonar, line 141: Should the call to `Start()` be inside the lock too?

Scenario:
Initial analysis request has completed successfully, process on port 888, init-linter called successfully. Then:
1. Thread A: first analysis request: Start() returns port 888
2. process crashes
3. Thread B: second analysis request: Start() returns port 999
4. Thread A resumes - compares port result, decides (incorrectly) that init-linter call is not required.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
duncanp-sonar: > I guess in a way it doesn't make sense to test this, because it's an equivalent of attempting to test a lock statement, just with an async call.

Well, we want to test how our code uses the lock statement rather than the lock statement itself i.e. whether the lock guarding the correct section of code.
That doesn't it's possible/easy to test though ;-)

Moving the call to Start inside the lock might make it easier to test.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -133,17 +136,31 @@ internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, Ca
...
 
-            if (port != javascriptServerPort)
+            try
```
duncanp-sonar, line 143: It's not immediately obvious what this code is trying to achieve i.e. _Ensure init-linter is called successfully once and only once per process_.
I think at least a comment is required, unless it can be made obvious by renaming/restructuring.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
duncanp-sonar: Actually, I'm not sure it will be any easier test e.g.

Task 1 - start analysis 1
 - block thread in Start call
Task 2 - start analysis 2

Task 2 should be blocked until Task 1 has finished.
- wait for Task2 for a short timeout
- check the wait timed out
- check Start hasn't been called twice

- unblock Start call and check Task 2 runs to completion.

Ugly, and not that different from using a Sleep.

Can you think of anything neater?

rita-sonarsource: No :( and in the end it still relies on the Sleep. It's very similar to the test that I wrote, but without the sleep, there's no way to really test the behavior. And I don't really want to use a sleep...

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
rita-sonarsource: The call to init-linter has to be made on the port that the Start() returned, so in this use case, if thread A got 888 and that process has died, any call on thread A will fail -- either the init-linter or the analyze-js, since both will call 888.
So if the process died at any point after a thread already got hold of the port number, I think that thread is gonna crash either way and the user would have to reanalyze. WDYT?

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
rita-sonarsource: Hmm, it's not just the init-linter. Thinking out loud: this method ensures that you get an instance of IEslintBridgeClient that is initialized to a (hopefully still) running port of the eslint-bridge process. So the method wraps 2 things: 1) that you have a running process, and 2) that process was initialized with init-linter. 

So perhaps renaming it from `GetJavaScriptClient` to `GetInitializedEslintBridgeClient` ? 
Or, change it to void and name it `EnsureEslintBridgeClientIsInitialized` - I originally returned an instance of client to ensure that you're getting an initialized instance, but I think there is a problem either way (that you spotted in the other comment):
1) Thread A calls EnsureEslintBridgeClientIsInitialized, client instance is initialized to 888
2) Before thread A manages to call analyze-js, the process dies
3) Thread B calls EnsureEslintBridgeClientIsInitialized, client instance is initialized to 999 but init-linter was not called yet.
4) Thread A attempts to call analyze-js and fails since thread B haven't called init-linter yet.

In either case, I think if the process died in the middle of an analysis, that analysis is gonna fail one way or another.

So if it will be a void, there is a potential for a different problem:
1) Thread A calls GetJavaScriptClient, returns client initialized to 888
2) Before thread A manages to call analyze-js, the process dies
3) Thread B calls GetJavaScriptClient, returns client initialized to 999
4) Thread A attempts to call analyze-js and fails since thread B disposed this instance

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
duncanp-sonar: Yep, the only way to test the locking is either to use a sleep-like approach and test if from the outside, or to put in more test hooks so that the behaviour can be controlled non-deterministically. Or not test it at all.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
duncanp-sonar: We are going to have a more complicated situation with TypeScript, where we'll need to check which `ts-config` was last used to initialize the eslintbridge. But that's a problem for another day.

Renaming the method would do.

Let's not over-complicate it for now. From what you said, the other SL flavours don't recover and it doesn't seem to have caused major problems so far.

duncanp-sonar (approved):

Issue #2284, opened 2021-04-06
Support user-configurable rules in JS/TS


duncanp-sonar:
### Description

The translation from Sonar rule parameters to the corresponding ESLint rule configuration is not metadata-driven; it's currently done in Java code e.g. [here](https://github.com/SonarSource/SonarJS/blob/0dda9105bab520569708e230f4d2dffdca3cec74/javascript-checks/src/main/java/org/sonar/javascript/checks/ArrowFunctionConventionCheck.java).

There isn't enough information in the metadata to automatically produce the required ESLint configuration.
See the Google doc _SonarLint / ESLintBridge integration_ for more information.

We need to decide how to handle this issue in both the short-term and the longer term.

#### Short-term options
* don't support parameterised rules in v1
* use hard-coded config in v1 i.e. don't allow the user configuration
* duplicate the Java code

#### Longer-term options
* add metadata to allow the transformation to be done automatically
* simplify/standardise the ESLint config to make the automated mapping easier

PR #2317, opened 2021-04-14
Use default params to set rule config

duncanp-sonar:


src/EmbeddedSonarAnalyzer.props
```diff
@@ -5,7 +5,7 @@
   <PropertyGroup>
     <EmbeddedSonarAnalyzerVersion>8.19.0.28253</EmbeddedSonarAnalyzerVersion>
     <EmbeddedSonarCFamilyAnalyzerVersion>6.17.0.27551</EmbeddedSonarCFamilyAnalyzerVersion>
-    <EmbeddedSonarJSAnalyzerVersion>7.4.0.15161</EmbeddedSonarJSAnalyzerVersion>
+    <EmbeddedSonarJSAnalyzerVersion>7.4.0.15221</EmbeddedSonarJSAnalyzerVersion>
```
duncanp-sonar, line 8: New version with the `defaultParams` metadata. The rest of the data is the same.

src/EmbeddedVsix.UnitTests/TypeScriptRuleDefinitions_SanityChecks.cs
```diff
@@ -58,10 +58,13 @@ private static void CheckRules(string language, IEnumerable<RuleDefinition> rule
...
             var ruleCount = rules.Count();
             var parameterisedRulesCount = rules.Count(RuleHasParameters);
-            Console.WriteLine($"{language}: rules: {ruleCount}, parameterised rules: {parameterisedRulesCount}");
+            var rulesWithDefaultParamsCount = rules.Count(HasDefaultParameters);
+
+            Console.WriteLine($"{language}: rules: {ruleCount}, parameterised rules: {parameterisedRulesCount}, non-null default params: {rulesWithDefaultParamsCount}");
 
             ruleCount.Should().BeGreaterThan(200);
             parameterisedRulesCount.Should().BeGreaterThan(10);
+            rulesWithDefaultParamsCount.Should().BeGreaterThan(5);
```
duncanp-sonar, line 67: Sanity check that we're getting some rules with default params back.

src/EmbeddedVsix.UnitTests/TypeScriptRuleDefinitions_SanityChecks.cs
```diff
@@ -85,5 +88,8 @@ private static void CheckRules(string language, IEnumerable<RuleDefinition> rule
             !string.IsNullOrEmpty(ruleDefinition.EslintKey) ||
             // Special case - "JavaScript parser failure"
             ruleDefinition.RuleKey.EndsWith(":S2260");
+
+        private static bool HasDefaultParameters(RuleDefinition ruleDefinition) =>
+                ruleDefinition.DefaultParams.Length > 0;
```
duncanp-sonar, line 93: Rules without default params will have an empty array, rather than null.
We're not expecting `DefaultParams` to be null.

src/TypeScript.UnitTests/Rules/RuleDefinitionsProviderTests.cs
```diff
@@ -89,36 +88,55 @@ public void GetAllRules_SameProviderInstances_ReturnsExpectedRulesForLanguage()
...
             result[1].Tags.Should().BeEquivalentTo("convention", "es2015");
             result[1].Params.Should().HaveCount(2);
+            result[1].DefaultParams.Should().HaveCount(1);
             result[1].Scope.Should().Be(RuleScope.BOTH);
             result[1].EslintKey.Should().Be("arrow-function-convention");
             result[1].ActivatedByDefault.Should().BeFalse();
         }
 
+        [TestMethod]
+        public void GetAllRules_WithDefaultParams_ReturnsExpectedConfigs()
```
duncanp-sonar, line 123: I'm not sure this test adds much value, but it does at least explicitly check that our understanding of how the default params work is correct.

src/TypeScript/Rules/ActiveJavaScriptRulesProvider.cs
```diff
@@ -66,12 +65,13 @@ private static bool IncludeRule(RuleDefinition ruleDefinition)
 
         private static Rule Convert(RuleDefinition ruleDefinition)
         {
+            Debug.Assert(ruleDefinition.DefaultParams != null, $"JavaScript rule default params should not be null: {ruleDefinition.RuleKey}");
             return new Rule
             {
                 Key = ruleDefinition.EslintKey,
 
                 // TODO: handle parameterised rules #2284
```
duncanp-sonar, line 73: #2284 is still TODO - it's for user-parameterised rules

src/TypeScript.UnitTests/Rules/RuleDefinitionsProviderTests.cs
rita-sonarsource: At the moment, is defaultParams always an array? Can it be an object?
I know it doesn't matter in the product code either way, just wondering since we discussed that there are different types.

rita-sonarsource (approved):

src/TypeScript.UnitTests/Rules/RuleDefinitionsProviderTests.cs
duncanp-sonar: It's always an array of objects. It might have contain a single simple element, but it will always be an array.

PR #2319, opened 2021-04-14
Add support for C++20 ("/std:c++latest")

### LINKED ISSUE #2318: Support C++ projects compiling with C++20 (/std:c++latest)
Currently the /std:c++latest is not supported. with the standardization of C++20, more people are specifying this option to use C++20 features.

abbas-sabra-sonarsource:
Fixes #2318
Note: this should be merged after 6.18 that why I'm setting "branch-updateAnalyzers" as a base branch

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
@@ -627,7 +627,30 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         }
 
         [TestMethod]
-        [DataRow("/std:latest", "/std:latest is not supported")]
+        public void CPP_Latest()
+        {
+            Request req = MsvcDriver.ToRequest(new CFamilyHelper.Capture[] {
+                compiler,
+                new CFamilyHelper.Capture()
+                {
+                    Executable = "",
+                    Cwd = "basePath",
+                    Env = new List<string>(),
+                    Cmd = new List<string>() {
+                      "cl.exe",
+                      "/std:c++latest",
+                      "c:\\file.cpp"
+                    },
+                }
+            });
+            req.Flags.Should().Be(Request.CPlusPlus | Request.CPlusPlus11 | Request.CPlusPlus14 |
+                Request.CPlusPlus17 | Request.CPlusPlus20 | Request.SonarLint | Request.MS);
+            req.Predefines.Should().Contain("#define __cplusplus 202002L\n");
+            req.Predefines.Should().Contain("#define _MSVC_LANG 202002L\n");
+        }
+
+        [TestMethod]
+        [DataRow("/std:c++20", "/std:c++20 is not supported")]
         [DataRow("/arch:foo", "/arch:foo is not supported")]
         [DataRow("/ZW", "/ZW: CX and CLI are not supported")]
         [DataRow("/clr", "/clr: CX and CLI are not supported")]
--- a/src/Integration.Vsix/CFamily/PortedFromJava/Analyzer.cs
+++ b/src/Integration.Vsix/CFamily/PortedFromJava/Analyzer.cs
@@ -48,6 +48,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
         public const int SonarLint = 1 << 18;
         public const int BuildPreamble = 1 << 19;
         public const int MainFileIsHeader = 1 << 20;
+        public const int CPlusPlus20 = 1 << 21;
 
         public string[] Options { get; set; } = Array.Empty<string>();
         public long Flags { get; set; }
--- a/src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs
+++ b/src/Integration.Vsix/CFamily/PortedFromJava/MsvcDriver.cs
@@ -412,6 +412,12 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
                                 predefines.Append("#define __cplusplus 201703L\n");
                                 predefines.Append("#define _MSVC_LANG 201703L\n");
                                 break;
+                            case "c++latest":
+                                request.Flags |= Request.CPlusPlus11 | Request.CPlusPlus14 |
+                                    Request.CPlusPlus17 | Request.CPlusPlus20;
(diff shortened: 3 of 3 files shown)


src/Integration.Vsix.UnitTests/CFamily/MsvcDriverTest.cs
```diff
@@ -627,7 +627,30 @@ public void Microsoft_Extensions()
...
+                }
+            });
+            req.Flags.Should().Be(Request.CPlusPlus | Request.CPlusPlus11 | Request.CPlusPlus14 |
+                Request.CPlusPlus17 | Request.CPlusPlus20 | Request.SonarLint | Request.MS);
+            req.Predefines.Should().Contain("#define __cplusplus 202002L\n");
+            req.Predefines.Should().Contain("#define _MSVC_LANG 202002L\n");
+        }
+
+        [TestMethod]
+        [DataRow("/std:c++20", "/std:c++20 is not supported")]
```
duncanp-sonar, line 653: I'm confused: what's the difference between `/std:c++latest` and `/std:c++20`?

Is `c++latest` still the working draft, or has the standard been ratified? And if it's been ratified, why can't we support `c++20`?

abbas-sabra-sonarsource: Currently there is no /std:c++20 in visual studio. They don't add it until they reach a full support of the new standard. So at this moment "latest" is C++17 plus some features from C++20 that are supported by the compiler. Once they support all the C++20 features they add /std:c++20 and /std:c++latest becomes C++20 plus features from the next standard(C++23). Once /std:c++20 is added to VS we have to update the test and adjust some option. After this change, our analyzer will support the C++20 features that are supported by the compiler. 
Note: The meaning of latest is continously changing. So we should keep monitoring it: for example, in 2 years we will probably have to start mapping it to C++23.

rita-sonarsource: @abbas-sabra-sonarsource I'm not sure I understand; does this mean that unless we actively monitor the changes on MS's side and adjust, our analysis will suddenly start to fail when latest becomes 23? I see that in [MsvcDriver](https://github.com/SonarSource/sonarlint-visualstudio/pull/2319/files#diff-3c2d8f1cfa7ec1666a416a65dd67a45a8c9cea67cf6721230b9c239ed93e26fd) you specify that we support CPlusPlus20.
I'm confused what will happen when MS make changes and officially release c++20.

abbas-sabra-sonarsource: @rita-sonarsource Yes we need to monitor it but, actively is a strong word. It is with the release and partial support of a new C++ standard. Which is usually every 2-3 years.
let me try to simplify it. `/std:c++latest` means latest **implemented** feature from the latest C++ standard. `/std:c++20` means **all** C++20 features and no features introduced after C++20.

Before C++17 was fully supported by msvc compiler:
- `/std:c++latest` =` C++14` + some  `C++17` features.
-  no `/std:c++17` available

When C++17 was fully supported by msvc compiler:
-  `/std:c++17` is available
- `/std:c++latest` == `/std:c++17`

When C++20 starting to be standardized: standard draft/ standard release**(we are here)**:
- `/std:c++latest` =` C++17` + some  `C++20` features.
-  no `/std:c++20` available

When C++20 is fully supported:
- `/std:c++20` available
- `/std:c++latest` ==``/std:c++20`

Let's say in 2 years when C++23 starting to be standardized(standard draft/ standard release):`

- `/std:c++latest` =` C++20` + some  `C++23` features. At this point `/std:c++latest`  should be changed to be mapped to C++23 
-  no `/std:c++23` available

Let me know if it is clearer now!

duncanp-sonar: Thanks @abbas-sabra-sonarsource, that's clearer.

How will this behave in earlier versions of VS? As far as I can tell, if the user specifies "latest" in VS2015 or VS2017, the CFamily analyser will analyse the code as if the user was targeting C++20. Is that what will happen, and will it cause a problem?

(the MS [docs](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-140) are slightly contradictory on the support for /std; in one place it says it's available in VS2017+, in other VS2015 Update 3+).

abbas-sabra-sonarsource: @duncanp-sonar @rita-sonarsource ,
**Summary:** 
Most of the users using /std::c++latest are using it to get some C++20 features. Otherwise, they can simply use /std:c++14/17 explicitly. It might be a little bit inconvenient in some edge conditions but supporting will bring value for most users and is currently aligned with the scanner behavior(we will face the same situation with VS2017/2015).

**Long answer:** 
The last time I checked, it is available only in the latest update of VS2015.

Depending on the compiler version/std::c++latest will have some extra features. And since the compiler that you can install depends on the VS version this might lead to different behavior between 2015/2017/2019.

**Why it is not a common case:**
if you are using /std::c++latestbut not expecting C++20. I would suggest that the user use /std::c++17 or /std:c++14. probably they are not aware that they are using the latest.

What happens if we analyze code as C++20 while they want to comply with, for example, C++17:
- We should be able to parse understand the code normally. The quality of the analysis will not be impacted. Unless the user is writing standard-aware code. in that case will analyze the C++20 code. This is not a common case:
```
#if C++20
generate some code
#else
generate other code
#endif
```
- From a rules perspective, this might be inconvenient since they will not be able to apply C++20 rules that suggest using the new features. This will be a problem even in SQ.SC context and the workaround would be to disable the rule with the C++20 tag.


So we map /std::c++latest to bring value to the common cause instead of failing. For the unusual cases, there will be some inconvenience but even, in that case, it will be a small inconvenience: definitely better than no analysis.

It is a baby step, If we ever discover an issue with this approach we can always try to study the behavior of each version of the compiler and try to customize the meaning of /std::c++latest based on it.

duncanp-sonar: Thanks @abbas-sabra-sonarsource, makes sense.

duncanp-sonar (approved):

PR #2320, opened 2021-04-14
Use IAnalysisStatusNotifier in JavascriptAnalyzer

### LINKED ISSUE #2280: Use IAnalysisStatusNotifier in JS/TS analyzers
#### Description

IAnalysisStatusNotifier provides useful abstraction over VS status bar and the output window logger. It is used in the CLangAnalyzer.

TODO:
- [ ] Investigate adding it in a generic way to all the analyzers
- [x] Add it to JS/TS analyzers

rita-sonarsource:
Fixes #2280


src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
```diff
@@ -54,30 +52,18 @@ internal EslintBridgeHttpWrapper(HttpMessageHandler httpHandler, ILogger logger)
 
         public async Task<string> PostAsync(Uri serverEndpoint, object request, CancellationToken cancellationToken)
         {
```
rita-sonarsource, line 54: I wanted to support notifying cancellation of the analysis, and for that I needed the TaskCancelledException to propagate instead of returning a null.
Also, since I'm handling analysis failures in the JavascriptAnalyzer, it made the logs here redundant and spammy.

So I opted to remove all the exception handling from here and from EslintBridgeClient, and let JavascriptAnalyzer catch all of them and notify accordingly. 
This class might not be needed anymore at all (since its purpose was mostly for error handling), but I left it for now.

src/Integration.Vsix.UnitTests/Analysis/AnalysisStatusNotifierTests.cs
```diff
@@ -141,6 +142,24 @@ public void AnalysisFailed_LogToOutputWindow()
             logger.OutputStrings.Count.Should().Be(1);
         }
 
+        [TestMethod]
+        public void AnalysisFailed_AggregateException_LogToOutputWindow()
```
rita-sonarsource, line 146: We had a special case in EslintBridgeHttpWrapper that collected the ex.Message of the inner exceptions.
However, in AnalysisStatusNotifier we always outputted ex.ToString(), and that works also for aggregate exceptions.

I think we that in general we might want to handle specific exceptions in JavascriptAnalyzer. For instance, if the process was killed, an WebException is thrown with message "Unable to connect to the remote server", and it contains an inner exception "SocketException: No connection could be made because the target machine actively refused it", which is a lot more useful from a troubleshooting standpoint. So we should probably catch WebExceptions and report the ex.Message of the underlying exception.

For now I left it with toString() and we could change to Message if we want to.

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -83,7 +83,12 @@ public async Task<AnalysisResponse> AnalyzeJs(string filePath, CancellationToken
 
             var responseString = await httpWrapper.PostAsync(BuildServerUri("analyze-js"), analysisRequest, cancellationToken);
 
-            return responseString == null ? null : JsonConvert.DeserializeObject<AnalysisResponse>(responseString);
+            if (string.IsNullOrEmpty(responseString))
+            {
+                throw new ArgumentNullException(nameof(responseString));
+            }
+
+            return JsonConvert.DeserializeObject<AnalysisResponse>(responseString);
```
rita-sonarsource, line 91: Not too sure about this. If the responseString is an empty string, DeserializeObject will return a null AnalysisResponse.
So if I don't throw an exception here, I'd have to check that the response is null in JavascriptAnalyzer and manually call AnalysisFailure on the notifier (instead of having it only in the catch block).

src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
```diff
@@ -93,11 +94,9 @@ public async Task ExecuteAnalysis_AlwaysCollectsTelemetry()
         [TestMethod]
         public async Task ExecuteAnalysis_AlwaysStartsEslintBridgeServer()
         {
-            var serverProcess = SetupServerProcess(123);
-            var client = SetupEslintBridgeClient(null);
-            var clientFactory = SetupEslintBridgeClientFactory(123, client.Object);
+            var serverProcess = SetupServerProcess();
```
rita-sonarsource, line 97: I changed the helper method and it simplified the test, we don't really care about the created client in this test.

src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
```diff
@@ -367,22 +366,6 @@ public async Task ExecuteAnalysis_ResponseWithIssues_ConsumerCalled()
             consumer.Verify(x => x.Accept("some path", convertedIssues));
         }
 
-        [TestMethod]
-        public async Task ExecuteAnalysis_NonCriticalException_ExceptionCaughtAndLogged()
```
rita-sonarsource, line 371: This is now handled by the `ExecuteAnalysis_AnalysisFailed_NotifiesThatAnalysisFailed` test.

src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
```diff
@@ -398,13 +381,13 @@ public void ExecuteAnalysis_CriticalException_ExceptionThrown()
         }
 
         [TestMethod]
-        public void Dispose_AnalysisNeverRan_NoException()
+        public void Dispose_AnalysisNeverRan_DisposesServerProcess()
```
rita-sonarsource, line 384: Renamed to be more accurate

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeHttpWrapperTests.cs
```diff
@@ -126,48 +125,15 @@ public async Task PostAsync_HasContent_SendsRequestWithSerializedContent()
...
-            logger.AssertPartialOutputStringExists("this is a test1");
-            logger.AssertPartialOutputStringExists("this is a test2");
-        }
-
-        [TestMethod]
-        public async Task PostAsync_CriticalException_ExceptionNotCaught()
-        {
-            var httpMessageHandler = new FakeHttpMessageHandler(message =>
-                throw new StackOverflowException());
+                throw new NotImplementedException());
```
rita-sonarsource, line 131: I removed all the exception handling tests and changed the exception to a non critical one (as the critical ones were already thrown beforehand).

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -93,7 +98,15 @@ private Task Close()
 
         public async void Dispose()
         {
-            await Close();
+            try
```
rita-sonarsource, line 101: As we discussed in a previous PR, JavascriptAnalyzer will dispose this class when the eslint-bridge's port changes. This will then attempt to call Close on process that is no longer running, fail, and leave httpWrapper behind.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
```diff
@@ -145,6 +147,22 @@ public void Dispose_DisposesHttpWrapper()
             httpWrapper.VerifyNoOtherCalls();
         }
 
+        [TestMethod]
+        public void Dispose_FailsToCallClose_HttpWrapperStillDisposed()
```
rita-sonarsource, line 151: Test for the try-catch in the dispose (probably should've split it to another PR... my bad).

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
duncanp-sonar: If null/empty is an exceptional (i.e. error) response then throwing seems ok.

However, I'm not sure about the type of the exception. It's not really an argument exception so that seems a bit misleading. What gets logged? Ideally we'd want some general like "JavaScript analysis failed".

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -107,33 +112,38 @@ internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, Ca
             // Switch to a background thread
             await TaskScheduler.Default;
 
+            analysisStatusNotifier.AnalysisStarted(filePath);
+
             try
             {
                 await EnsureEslintBridgeClientIsInitialized(cancellationToken);
-
+                var analysisStartTime = DateTime.Now;
```
duncanp-sonar, line 120: I don't know why we didn't originally use a `Stopwatch` in the CFamily version, and then `Stopwatch.Elapsed` to get the accurate `TimeSpan`.  Was there a reason?

duncanp-sonar (approved):
Looks good. Just a question about logging a misleading "argument null" exception if the analysis returns null/empty.

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
rita-sonarsource: Right now it's after the catch 👍

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
rita-sonarsource: No idea, good call.

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
rita-sonarsource: Hm, perhaps just ArgumentException?
In the end it will reach the notifier and output the generic "analysis failed" message, along with this exception.

rita-sonarsource: I merged for now with ArgumentException, I will make a change tomorrow if needed.

Issue #2308, opened 2021-04-13
Stop EslintBridgeProcess when solution closes


rita-sonarsource:
#### Description

To avoid consuming unneeded resources + potential memory leaks, we should stop the running EslintBridgeProcess.

TODO:
- [ ] Listen to solution closed/changed events and call EslintBridgeProcess.Stop

PR #2328, opened 2021-04-15
Add ActiveSolutionChangedCallback

rita-sonarsource:
Part of #2308

### DIFF
--- a/src/Integration.Vsix/SonarLintDaemonPackage.cs
+++ b/src/Integration.Vsix/SonarLintDaemonPackage.cs
@@ -23,10 +23,9 @@ using System.Runtime.InteropServices;
 using System.Threading;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
-using Microsoft.VisualStudio.Shell.Interop;
+using SonarLint.VisualStudio.Integration.MefServices;
 using SonarLint.VisualStudio.Integration.Vsix.Analysis;
 using SonarLint.VisualStudio.Integration.Vsix.CFamily;
-using SonarLint.VisualStudio.Integration.Vsix.Resources;
 
 namespace SonarLint.VisualStudio.Integration.Vsix
 {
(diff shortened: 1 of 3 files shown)


src/Integration/MefServices/ActiveSolutionChangedCallback.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// </summary>
+    public interface IActiveSolutionChangedCallback : IDisposable
+    {
+    }
+
+    [Export(typeof(IActiveSolutionChangedCallback))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal sealed class ActiveSolutionChangedCallback : IActiveSolutionChangedCallback
+    {
+        internal const string CallbackContractName = "ActiveSolutionChangedCallback";
```
rita-sonarsource, line 38: This will be repeated in every consumer but I don't know where to put it. Perhaps it's better to import an interface, WDYT?

src/Integration.Vsix/SonarLintDaemonPackage.cs
```diff
@@ -90,6 +88,7 @@ private async System.Threading.Tasks.Task InitAsync()
                 await CFamilyReproducerCommand.InitializeAsync(this, logger);
 
                 cFamilyPreCompiledHeadersEventListener = await this.GetMefServiceAsync<IPreCompiledHeadersEventListener>();
+                activeSolutionChangedCallback = await this.GetMefServiceAsync<IActiveSolutionChangedCallback>();
```
rita-sonarsource, line 91: Another item in the dumping ground... I remembered why we used this for the CFamily stuff: because the integration package only applies for .net stuff, not cfamily/js.

Side note: the callbacks won't be called until this package loads, so it is possible that a solution is opened/closed at the start of VS before this package loads, and the eslint-bridge will be left behind.

duncanp-sonar: Hhmm. The daemon package loads when the shell is initialized, but since it's done in the background there can be quite a delay before VS actually gets round to loading it.

I was hoping to avoid having the analyzers act as observers too. However, I've hit a similar problem with disabling rules - we need to call `init-linter` again when the user settings change i.e. something needs to monitor the user settings file for changes :-(

PR #2330, opened 2021-04-15
Move IActiveSolutionTracker to Core

rita-sonarsource:
Prep for #2308

### DIFF
--- a/src/Core/IActiveSolutionTracker.cs
+++ b/src/Core/IActiveSolutionTracker.cs
@@ -19,7 +19,6 @@
  */
 
 using System;
-using Microsoft.VisualStudio.Shell.Interop;
 
 namespace SonarLint.VisualStudio.Integration
 {
@@ -33,16 +32,6 @@ namespace SonarLint.VisualStudio.Integration
         public bool IsSolutionOpen { get; }
     }
 
-    public class ProjectOpenedEventArgs : EventArgs
-    {
-        public ProjectOpenedEventArgs(IVsHierarchy pHierarchy)
-        {
-            ProjectHierarchy = pHierarchy;
-        }
-
-        public IVsHierarchy ProjectHierarchy { get; }
-    }
-
     public interface IActiveSolutionTracker
     {
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Core/IActiveSolutionTracker.cs
```diff
@@ -19,7 +19,6 @@
  */
 
 using System;
-using Microsoft.VisualStudio.Shell.Interop;
 
 namespace SonarLint.VisualStudio.Integration
```
rita-sonarsource, line 23: Moved to Core so I could use it in Typescript proj, I'll open a separate PR for namespace change

src/Core/IActiveSolutionTracker.cs
```diff
@@ -33,16 +32,6 @@ public ActiveSolutionChangedEventArgs(bool isSolutionOpen)
         public bool IsSolutionOpen { get; }
     }
 
-    public class ProjectOpenedEventArgs : EventArgs
```
rita-sonarsource, line 36: Was not used

src/Integration/MefServices/ActiveSolutionTracker.cs
```diff
@@ -24,6 +24,8 @@
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
+using SonarLint.VisualStudio.Core;
+using ErrorHandler = Microsoft.VisualStudio.ErrorHandler;
```
rita-sonarsource, line 28: Changed due to Core dependency, although we probably want to use Core's ErrorHandler, no?

duncanp-sonar: Unfortunately, we're not consistent in which one we use (although Core is supposed to work the same way as the VS version). In fact, we're so inconsistent it probably doesn't matter which one we pick.

However, we could try to establish a rule of thumb to follow: if the assembly reference VS assemblies, use the VS one; otherwise use Core.
It's a bit arbitrary, but at least it would be consistent. What do you think?

duncanp-sonar (approved):

src/Integration/MefServices/ActiveSolutionTracker.cs
rita-sonarsource: So basically use the VS one everywhere, except in Core? 🤔 Can do, and then make it internal.

PR #2331, opened 2021-04-15
Stop EslintBridgeProcess when solution closes

### LINKED ISSUE #2308: Stop EslintBridgeProcess when solution closes

rita-sonarsource:
Fixes #2308

### DIFF
--- a/src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
+++ b/src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
@@ -38,6 +38,11 @@ namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
         /// Ensures that the process is running. Returns the port that server is running on.
         /// </summary>
         Task<int> Start();
+
+        /// <summary>
+        /// Terminates running process.
+        /// </summary>
+        void Stop();
     }
 
     [Export(typeof(IEslintBridgeProcess))]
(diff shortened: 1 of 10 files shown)


src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -102,14 +107,19 @@ public Task<int> Start()
...
-        public void Dispose()
+        public void Stop()
         {
             lock (Lock)
             {
                 TerminateRunningProcess();
             }
         }
 
+        public void Dispose()
```
rita-sonarsource, line 118: Right now this class is MEF-exportable, so it seems odd to call dispose on it, even though it behaves the same as Stop(). 
So I added the stop but it's redundant functionality-wise. 
When we make it non-MEF (for typescript), the dispose will make sense and we could remove the Stop.
WDYT? should I just call dispose now?

duncanp-sonar: Agreed, it is odd - we should never call Dispose on a MEF-exported object. It should just be there for VS to call when it is shutting down (although I'm not sure there is a strong guarantee by the runtime that Dispose will be called when a managed process is shutting down).

I'd either leave it as MEF-exportable and make the implementation of IDisposable explicit (so Dispose isn't part of the "normal" visible API for the class), or stop exporting it. Whichever is simpler - we might end up changing it for TypeScript anyway.

duncanp-sonar (approved):

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
rita-sonarsource: I'd leave it as-is for now, if that's ok. I'm quite sure we will change it when we do typescript, and then I'd want the dispose method public on it.

PR #2332, opened 2021-04-16
Cleanup - moved frequently-used Sonar artefact constants to a single …

duncanp-sonar:
…file


src/Core/SonarArtefactKeys.cs
```diff
@@ -0,0 +1,73 @@
+﻿/*
```
duncanp-sonar, line 1: I renamed `SonarLanguageKeys` to `SonarArtefactKeys` and added the other constants and comments.

We have a couple of problems with constants:
* the same constants are defined in multiple places
* confusion between different types of artefact that happen to have the same values e.g. the C++ language and repo keys.

There's more to do, but this is a first step.

src/Core/SonarArtefactKeys.cs
```diff
@@ -0,0 +1,73 @@
...
+    // * rule definition: part of a single repository. The full rule id is given by "[repo id]:[rule id]"
+
+    /// <summary>
+    /// Language keys for languages supported by SonarQube/Cloud plugins
+    /// </summary>
+    /// <remarks>A full list of languages keys can be obtained by calling https://sonarcloud.io/api/languages/list
+    /// </remarks>
+    public static class SonarLanguageKeys
+    {
+        // TODO: the SonarQube.Client assembly has a class that defines the SonarQube language keys:
```
duncanp-sonar, line 46: This type is unchanged (except for the addition of the TODO comment - perhaps we should move the definition of all of the server constants to the SonarQube.Client assembly).

The other types are new.

rita-sonarsource (approved):

Issue #771, opened 2018-11-30
Provide standalone-mode ruleset configuration mechanism for JavaScript


duncanp-sonar:
### Description

Currently there is no way for users to configure the set of SonarLint rules that are run in standalone mode for languages other than C# and VB.

Users should be able to configure the set of rules that are run.

### Related information

* SonarLint for VS Version: all (current versin 4.7)

Related ticket about configuring rules in Connected Mode: #770 

UPDATE: support for disabling C++ rules was added in v4.13. See #1015.

duncanp-sonar:
JavaScript rules can be enable/disabled via the `settings.json` file.
However, rule parameters cannot set in the settings file.

PR #2333, opened 2021-04-16
Add ConfigChanged event

rita-sonarsource:
Prep for #771

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
@@ -56,7 +56,33 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
         }
 
         [TestMethod]
-        public void WhenSupressionsUpdated_AnalysisIsRequested()
+        public void WhenUserSettingsChange_StandaloneMode_HasSubscribersToConfigChangedEvent_SubscribersNotified()
+        {
+            var builder = new TestEnvironmentBuilder(SonarLintMode.Standalone);
+            var eventHandler = new Mock<EventHandler>();
+            builder.TestSubject.ConfigChanged += eventHandler.Object;
+
+            builder.SimulateUserSettingsChanged();
+
+            eventHandler.Verify(x=> x(builder.TestSubject, EventArgs.Empty), Times.Once);
+        }
+
+        [TestMethod]
+        [DataRow(SonarLintMode.Connected)]
+        [DataRow(SonarLintMode.LegacyConnected)]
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
```diff
@@ -56,7 +56,33 @@ public void WhenUserSettingsChange(SonarLintMode bindingMode, bool shouldAnalysi
...
+            var eventHandler = new Mock<EventHandler>();
+            builder.TestSubject.ConfigChanged += eventHandler.Object;
+
+            builder.SimulateUserSettingsChanged();
+
+            eventHandler.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public void WhenSuppressionsUpdated_AnalysisIsRequested()
```
rita-sonarsource, line 85: typo in existing test's name

src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
```diff
@@ -56,7 +56,33 @@ public void WhenUserSettingsChange(SonarLintMode bindingMode, bool shouldAnalysi
         }
 
         [TestMethod]
-        public void WhenSupressionsUpdated_AnalysisIsRequested()
+        public void WhenUserSettingsChange_StandaloneMode_HasSubscribersToConfigChangedEvent_SubscribersNotified()
```
rita-sonarsource, line 59: The "no subscribers - no exception" is already covered by the existing tests

src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
```diff
@@ -66,6 +92,18 @@ public void WhenSupressionsUpdated_AnalysisIsRequested()
             builder.Logger.AssertOutputStringExists(AnalysisStrings.ConfigMonitor_SuppressionsUpdated);
         }
 
+        [TestMethod]
+        public void WhenSuppressionsUpdated_HasSubscribersToConfigChangedEvent_SubscribersNotified()
```
duncanp-sonar, line 96: I hadn't thought about suppressions; strictly speaking we don't need to re-calculate the effective rules or call init-linter in this case, but it won't happen often so it won't matter too much.

duncanp-sonar (approved):
Thanks!

src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
rita-sonarsource: It would happen if it's a proj in connected mode that also happens to have JS files, but yeah it would be every 10 min + on binding update.

PR #2334, opened 2021-04-16
Added SonarCompositeRuleId helper class

duncanp-sonar:
Prep for part of #771

### DIFF
--- a/src/Core/SonarCompositeRuleId.cs
+++ b/src/Core/SonarCompositeRuleId.cs
@@ -1,0 +1,55 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


src/Core/SonarCompositeRuleId.cs
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+using System;
+
+namespace SonarLint.VisualStudio.Core
+{
+    public class SonarCompositeRuleId
+    {
+        /// <summary>
+        /// Attempts to parse a VS error code as a Sonar rule in the form "[repo key]:[rule key]"
```
duncanp-sonar, line 28: As part of changing the "disable rule" command I need both parts of the composite rule ID so I've created this helper.

src/Core/SonarCompositeRuleId.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// </summary>
+        public static bool TryParse(string vsErrorCode, out SonarCompositeRuleId ruleInfo)
```
rita-sonarsource, line 30: vsErrorCode is a bit misleading, it sounds like something VS would throw. Maybe `errorListRuleCode`?

rita-sonarsource (approved):

src/Core/SonarCompositeRuleId.cs
duncanp-sonar: The column in the Error List is called "ErrorCode" - `ErrorListErrorCode`?

rita-sonarsource: Yeah that works 👍

PR #2336, opened 2021-04-16
Stop EslintBridge process when analysis config changes

rita-sonarsource:
Part of #771

### DIFF
--- a/src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
+++ b/src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
@@ -48,6 +48,7 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
         private readonly ITelemetryManager telemetryManager;
         private readonly IAnalysisStatusNotifier analysisStatusNotifier;
         private readonly IActiveSolutionTracker activeSolutionTracker;
+        private readonly IAnalysisConfigMonitor analysisConfigMonitor;
         private readonly ILogger logger;
 
         private IEslintBridgeClient eslintBridgeClient;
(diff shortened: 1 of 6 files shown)


src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -63,13 +64,15 @@ internal sealed class JavaScriptAnalyzer : IAnalyzer, IDisposable
             ITelemetryManager telemetryManager,
             IAnalysisStatusNotifier analysisStatusNotifier,
             IActiveSolutionTracker activeSolutionTracker,
+            IAnalysisConfigMonitor analysisConfigMonitor,
```
rita-sonarsource, line 67: The CTOR of this class is getting ginormous but I think there's a way to do some tidying up: a lot of the dependencies are part of the EslintBridgeProcess's management and initialization of the EslintBridgeClient; I think we can extract `EnsureEslintBridgeClientIsInitialized` to another class that returns an initialized instance of the client + maintains the lifetime of the process. We could make it generic and re-use in TypescriptAnalyzer.
I suggest we take a look at this when we write the TS stuff.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -218,6 +225,11 @@ private void ActiveSolutionTracker_ActiveSolutionChanged(object sender, ActiveSo
             StopServer();
         }
 
+        private void AnalysisConfigMonitor_ConfigChanged(object sender, EventArgs e)
+        {
+            StopServer();
```
rita-sonarsource, line 230: Minor logging issue: our "server started" logs are written always, but the "stopped" logs are debug only. Do you think it's worth changing? I had to turn off my debug logs due to noise and then didn't understand why I wasn't seeing any stop logs.

duncanp-sonar: It would be better to be consistent, and it does look weird only getting "start" messages. I'd make it non-debug.
Also, do we log the process id? Knowing it makes it easier to check e.g. memory usage.

duncanp-sonar (approved):

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
rita-sonarsource: Yep we log it. I'll open a PR for the non debug then.

PR #2337, opened 2021-04-16
Change EslintBridgeProcess log verbosity

rita-sonarsource:
Changing LogDebug to Writeline, so we could see the termination logs in non-debug mode.

### DIFF
--- a/src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
+++ b/src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
@@ -189,8 +189,6 @@ namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
 
         private void TerminateRunningProcess()
         {
-            logger.LogDebug(Resources.INFO_TerminatingServer);
-
             try
             {
                 if (Process == null || Process.HasExited)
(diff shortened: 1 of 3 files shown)


src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -189,8 +189,6 @@ private void OnErrorDataReceived(object sender, DataReceivedEventArgs e)
 
         private void TerminateRunningProcess()
         {
-            logger.LogDebug(Resources.INFO_TerminatingServer);
```
rita-sonarsource, line 192: Since we now call this when solution closes & settings change, this was getting a bit spammy as maybe we never analyzed a JS file and therefore there was no process running.
So I moved it to the block where there definitely is a process and logged the process id as well.

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -199,9 +197,11 @@ private void TerminateRunningProcess()
                 }
                 else
                 {
+                    var processId = Process.Id;
```
rita-sonarsource, line 200: I can't use the Process.Id after it was killed, so I'm grabbing it beforehand,

src/TypeScript/EslintBridgeClient/Resources.resx
```diff
@@ -129,21 +129,21 @@
...
   <data name="INFO_ServerProcessId" xml:space="preserve">
-    <value>[eslint-bridge] Started eslint-bridge server. Process id: {0}</value>
+    <value>[eslint-bridge] Started eslint-bridge process. Process id: {0}</value>
   </data>
   <data name="INFO_ServerStarted" xml:space="preserve">
     <value>[eslint-bridge] Server running on port: {0}</value>
   </data>
   <data name="INFO_ServerTerminated" xml:space="preserve">
-    <value>[eslint-bridge] EslintBridge process terminated.</value>
+    <value>[eslint-bridge] Terminated eslint-bridge process. Process id: {0}</value>
```
rita-sonarsource, line 144: Just some string changes, to be consistent with the "eslint-bridge process" terminology.

duncanp-sonar (approved):

PR #2338, opened 2021-04-16
Allow disabling JS rules in standalone and connected mode

duncanp-sonar:
Part of #771

### DIFF
--- a/src/Core.UnitTests/SonarCompositeRuleIdTests.cs
+++ b/src/Core.UnitTests/SonarCompositeRuleIdTests.cs
@@ -48,5 +48,15 @@ namespace SonarLint.VisualStudio.Core.UnitTests
             result.RepoKey.Should().Be(expectedRepo);
             result.RuleKey.Should().Be(expectedRule);
         }
+
+
+        [TestMethod]
+        public void ToString_ReturnsExpected()
+        {
+            const string input = "xxx:YYY";
+
+            SonarCompositeRuleId.TryParse(input, out var output);
+            output.ToString().Should().Be(input);
+        }
     }
 }
--- a/src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
@@ -38,6 +38,10 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
     [TestClass]
     public class DisableRuleCommandTests
     {
+        private const int VisibleAndEnabled = (int)(OLECMDF.OLECMDF_SUPPORTED | OLECMDF.OLECMDF_ENABLED);
+        private const int VisibleButDisabled = (int)(OLECMDF.OLECMDF_SUPPORTED);
+        private const int InvisbleAndDisabled = (int)(OLECMDF.OLECMDF_SUPPORTED | OLECMDF.OLECMDF_INVISIBLE);
+
         [TestMethod]
         public void Ctor_NullArguments()
         {
@@ -80,13 +84,16 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
         }
 
         [TestMethod]
-        public void CheckStatusAndExecute_SingleCFamilyIssue_StandaloneMode_VisibleAndEnabled()
+        [DataRow("cpp:S111", "S111")]
+        [DataRow("c:S222", "S222")]
+        [DataRow("javascript:S333", "S333")]
+        public void CheckStatusAndExecute_SingleIssue_SupportedRepo_StandaloneMode_VisibleAndEnabled(string errorCode, string expectedRuleKey)
         {
             // Arrange
             var issueHandle = CreateIssueHandle(111, new Dictionary<string, object>
             {
                 { StandardTableKeyNames.BuildTool, "SonarLint" },
-                { StandardTableKeyNames.ErrorCode, "cpp:S123" }
+                { StandardTableKeyNames.ErrorCode,  errorCode}
             });
             var errorList = CreateErrorList(issueHandle);
             var mockUserSettingsProvider = new Mock<IUserSettingsProvider>();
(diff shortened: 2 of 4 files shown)


src/Core.UnitTests/SonarCompositeRuleIdTests.cs
```diff
@@ -48,5 +48,15 @@ public void TryParse_Valid_ReturnExpecteed(string errorCode, string expectedRepo
...
+
+
+        [TestMethod]
+        public void ToString_ReturnsExpected()
+        {
+            const string input = "xxx:YYY";
+
+            SonarCompositeRuleId.TryParse(input, out var output);
+            output.ToString().Should().Be(input);
+        }
```
duncanp-sonar, line 60: I forgot to check this test in before

src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
```diff
@@ -38,6 +38,10 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
     [TestClass]
     public class DisableRuleCommandTests
     {
+        private const int VisibleAndEnabled = (int)(OLECMDF.OLECMDF_SUPPORTED | OLECMDF.OLECMDF_ENABLED);
```
duncanp-sonar, line 41: I turned the expected values into constants to make the tests easier to read.

src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
```diff
@@ -99,26 +106,31 @@ public void CheckStatusAndExecute_SingleCFamilyIssue_StandaloneMode_VisibleAndEn
...
         [TestMethod]
-        [DataRow(SonarLintMode.Connected)]
-        [DataRow(SonarLintMode.LegacyConnected)]
-        public void CheckStatus_SingleCFamilyIssue_ConnectedMode_VisibleButNotEnabled(SonarLintMode bindingMode)
+        [DataRow("c:S111", SonarLintMode.Connected, false)]
+        [DataRow("c:S111", SonarLintMode.LegacyConnected, false)]
+        [DataRow("cpp:S111", SonarLintMode.Connected, false)]
+        [DataRow("cpp:S111", SonarLintMode.LegacyConnected, false)]
+        [DataRow("javascript:S111", SonarLintMode.Connected, true)]
+        [DataRow("javascript:S111", SonarLintMode.LegacyConnected, true)]
```
duncanp-sonar, line 125: Special casing for JS - can still disable in connected modes for now.

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
```diff
@@ -95,10 +94,12 @@ private void QueryStatus(object sender, EventArgs args)
         {
             try
             {
-                // Show the menu item for Sonar errors...
-                bool isVisible = TryGetErrorCodeSync(errorList, out _);
```
duncanp-sonar, line 99: Previously the behaviour was:
* we didn't supported disabling rules in all Sonar languages,
* but if disabling was supported it only worked in standalone mode.

Now the behaviour is slightly more complicated:
* we still don't support disabling rules in all Sonar languages;
* JS rules can always be disabled, regardless of mode,
* other supported languages (i.e. C, C++) can only be disabled in standalone mode.


`TryGetErrorCodeSync`

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
```diff
@@ -149,22 +150,18 @@ private void Execute(object sender, EventArgs e)
         /// 1) there is only one selected error in the error list and
         /// 2) it is a SonarLint error
         /// </summary>
-        internal /* for testing */ static  bool TryGetErrorCodeSync(IErrorList errorList, out string errorCode)
```
duncanp-sonar, line 152: The old comment was inaccurate: there was a third condition: "3) the language supports disabling".
The comment is now accurate, as the enable/disable check is done elsewhere.

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
```diff
@@ -191,11 +188,28 @@ private static string FindErrorCodeForEntry(ITableEntryHandle handle)
             SonarRuleRepoKeys.Cpp
         };
 
-        private static bool IsDisablingRulesSupported(string errorCode)
+        private static void CalculateStatuses(SonarCompositeRuleId rule, SonarLintMode mode, out bool isVisible, out bool isEnabled)
         {
+            // Special case: JavaScript
+            // We don't support connected mode for JavaScript at the moment so we allow disabling
+            // in standalone mode. See #770.
```
duncanp-sonar, line 195: Should be "... in connected mode."

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
rita-sonarsource: Well, we could simplify the JS stuff, and combine it with C++. As in, if it's not standalone, return false.

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
```diff
@@ -191,11 +188,28 @@ private static string FindErrorCodeForEntry(ITableEntryHandle handle)
...
+            if (SonarRuleRepoKeys.AreEqual(SonarRuleRepoKeys.JavaScript, rule.RepoKey))
```
rita-sonarsource, line 196: Is this block really needed? There's no connected mode in JS, so it should always work in the following block.

rita-sonarsource (approved):
Looks good, just the comment about JS's connected mode.

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
duncanp-sonar: Yep, we could. That would mean you could disable JS rules in standalone mode, but you'd always have the default settings in connected mode i.e. if a user is already using connected mode and the JS rules are noisy, there's nothing they can do to stop it apart from stop using connected mode or uninstall SLVS.

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
duncanp-sonar: Connected mode is set at solution level, so it is possible to have JS files in connected mode.

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
rita-sonarsource: Ah I see, ok 👍

PR #2342, opened 2021-04-17
Added MEF test for TelemetryManager

duncanp-sonar:


duncanp-sonar:
The MEF test creates an instance of the test subject without any stub. The test instance tries to interact with the real VS UIContext handling code, leaving it in a bad state that causes other tests to fail.

Issue #2340, opened 2021-04-17
Add setting to disable all rules for a language


duncanp-lseg:
### Description

At present the only way to completely disable analysis for a language is to individually disable all of the rules for that language, either in the settings file (standalone mode) or in the Quality Profile (connected mode). This isn't a practical solution, given that each language has hundreds of rules.

It should be possible to disable all of the rules for a specific language via a single setting in the user settings file.

#### Out of scope
The following features are out of scope for now:
* UI to disable/enable languages

#### Should the setting apply in Connected mode, or only in Standalone mode?
At present, the rule configurations in the user settings file are ignored in Connected mode, the thinking being that the Quality Profile settings should take precedence in that case.

However, the individual developer should be in control of the development experience in their IDE, so they should be able to completely disable analysis for a language if they wish.

#### TODO
- [x] add appropriate settings to the `settings.json` config file
- [x] update the JavaScript rules provider to respect the new setting
- [ ] update the CFamily rules provider to respect the new setting
- [ ] add telemetry to record disabled languages
- [ ] update documentation

#### File format
The json format of the new setting will be as follows:
```
{
  "sonarlint.general": {
    "disabledLanguages": ["cpp", "c", "js"]
  },

  "sonarlint.rules": {
...
  }
}
```

The values stored will be the Sonar language keys i.e. cpp, c, js, ts etc.

Bosch-Eli-Black:
@duncanp-sonar Has any of this feature already been shipped as part of the SonarLint VS extension? I'm looking to disable SonarLint for JS and TS files, e.g.

```
  "sonarlint.general": {
    "disabledLanguages": ["js", "ts"]
  },
```

Thanks! 🙂

duncanp-sonar:
Hi @Bosch-Eli-Black. No, it's not shipped. We ran out of time to implement it in the sprint and it was cut. It's still on the backlog, but not currently prioritised.

cc @marco-comi-sonarsource for info.

Bosch-Eli-Black:
@duncanp-sonar Okay, thanks 🙂

I don't think this is a priority for us, either. I was hoping to be able to temporarily disable JS warnings while I tested out the SonarLint extension, but long-term, we probably want to use connected mode and/or configure which JS errors we want to see, anyway.

duncanp-sonar:
@Bosch-Eli-Black FYI JS/TS rule configuration is currently only supported in standalone mode. Support for connected mode is covered in #770.

[...]

PR #2343, opened 2021-04-17
 Added disabledLanguages data to telemetry payload

duncanp-sonar:
Part of #2340

### DIFF
--- a/src/Core.UnitTests/Binding/CompositeBindingConfigProviderTests.cs
+++ b/src/Core.UnitTests/Binding/CompositeBindingConfigProviderTests.cs
@@ -80,7 +80,7 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Binding
             testSubject.IsLanguageSupported(Language.VBNET).Should().BeTrue();
             testSubject.IsLanguageSupported(Language.CSharp).Should().BeTrue();
 
-            // 2. Unsupported langauge
+            // 2. Unsupported language
             testSubject.IsLanguageSupported(Language.Cpp).Should().BeFalse();
         }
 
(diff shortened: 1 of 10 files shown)


src/Core.UnitTests/Binding/CompositeBindingConfigProviderTests.cs
```diff
@@ -80,7 +80,7 @@ public void IsSupported_ReturnsTrueIfSupportedByAny()
             testSubject.IsLanguageSupported(Language.VBNET).Should().BeTrue();
             testSubject.IsLanguageSupported(Language.CSharp).Should().BeTrue();
 
-            // 2. Unsupported langauge
+            // 2. Unsupported language
```
duncanp-sonar, line 83: Miscellaneous - fix typo

src/Integration.UnitTests/Telemetry/TelemetryHelper_CreatePayload.cs
```diff
@@ -34,13 +35,15 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
     [TestClass]
     public class TelemetryHelper_CreatePayload
     {
+        private static readonly BindingConfiguration ValidBindingConfiguration = CreateConfiguration(SonarLintMode.Connected, "https://sonarcloud.io");
```
duncanp-sonar, line 38: Miscellaneous - added symbolic value to make reduce duplication and make the tests easier to read

src/Integration/Telemetry/TelemetryManager.cs
```diff
@@ -44,19 +45,25 @@ public sealed class TelemetryManager : ITelemetryManager, IDisposable
         private readonly ICurrentTimeProvider currentTimeProvider;
         private readonly IVsVersion vsVersion;
 
+        // Note: there isn't a MEF test for this constructor.Writing one causes other tests to fail.
```
rita-sonarsource, line 48: I think it would be better to create a Mef "test", mark it as Ignored, and paste that comment there. The comment doesn't really belong in product code.

PR #2350, opened 2021-04-21
Update text in options window

### LINKED ISSUE #2349: Update text in options pane
### Description

The text in the general options pane needs to be updated to reflect the new behaviours:
* the `settings.json` file can be now be used to enable/disable JS rules.
* the JS settings will apply even in standalone, as connected mode is not yet supported (see #770)

duncanp-sonar:
Added hyperlink to show the wiki in a browser.

Fixes #2349

### DIFF
--- a/src/Integration.Vsix/Settings/GeneralOptionsDialogPage.cs
+++ b/src/Integration.Vsix/Settings/GeneralOptionsDialogPage.cs
@@ -25,12 +25,15 @@ using System.Linq;
 using System.Windows;
 using Microsoft.VisualStudio.Shell;
 using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Infrastructure.VS;
+using SonarLint.VisualStudio.Integration.WPF;
 
 namespace SonarLint.VisualStudio.Integration.Vsix
 {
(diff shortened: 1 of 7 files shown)


src/Integration.Vsix/Settings/GeneralOptionsDialogPage.cs
```diff
@@ -43,10 +46,12 @@ protected override UIElement Child
                 {
                     Debug.Assert(this.Site != null, "Expecting the page to be sited");
                     var userSettingsProvider = this.Site.GetMefService<IUserSettingsProvider>();
+                    var browserService = this.Site.GetMefService<IVsBrowserService>();
                     var logger = this.Site.GetMefService<ILogger>();
 
                     var openSettingsFileCmd = new OpenSettingsFileWpfCommand(this.Site, userSettingsProvider, this, logger);
-                    dialogControl = new GeneralOptionsDialogControl(openSettingsFileCmd);
+                    var showWikiInBrowserCmd = new RelayCommand(() => browserService.Navigate(WikiUrl));
```
rita-sonarsource, line 53: FYI we have `INavigateToDocumentationCommand` that could've been used here :)

Issue #2351, opened 2021-04-21
Error List displays a empty line if the span of the selected issue is invalidated


duncanp-sonar:
### Description

After editing a file to remove an issue, the Error List shows an empty line if the edited issue was previously selected in the Error List.

Only applies to C/C++ and JavaScript i.e. non-Roslyn issues

### Repro steps

1. Analyse a JS/CFamily file
2. [UPDATE] Select an item in the Error List
3. Edit the file to change a character in the squiggle for the selected issue

The issue was reported with the sample file given below,.
Analyse the file, start at the end of line 51, backspace to remove the `d` in `var id`

Note: the problem only occurs if the issue is selected in the Error List.

#### Sample file (change extension to .js to use)

[Spript1.js.txt](https://github.com/SonarSource/sonarlint-visualstudio/files/6350401/Spript1.js.txt)

#### Before editing:
![Before](https://user-images.githubusercontent.com/10757559/115545904-4881dc00-a29c-11eb-8c74-78b2754fd27f.png)

#### After editing
![After](https://user-images.githubusercontent.com/10757559/115545920-4e77bd00-a29c-11eb-9000-214443d38f72.png)

### Expected behavior

The line for the issue disappears from the Error List.

### Actual behavior

The line for the issue remains in the Error List, but has no data i.e. looks like a blank line.

### Known workarounds

Save the file to trigger a new analysis

### Related information

* SonarLint for VS Version: 4.33 (and possibly earlier versions)
* Languages in your solution: Javascript, C, C++ i.e. not Roslyn

duncanp-sonar:
PM reports that the issue still repros on his VM.
The code change still makes sense, so it doesn't need to be rolled back.

PR #2352, opened 2021-04-21
Fix for #2351 - blank line in error list after editing the highlighted text

duncanp-sonar:


src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests.cs
```diff
@@ -112,8 +114,8 @@ public void IndexOf_SameSnapshotId_ReturnExpectedIndex()
             var revised = original.CreateUpdatedSnapshot("unimportant change");
 
             // Should be able to map issues between two snapshots with the same snapshot id
-            original.IndexOf(999, revised)
-                .Should().Be(999);
+            original.IndexOf(ValidIndexInValidIssueList, revised)
```
duncanp-sonar, line 117: The product code is now using the supplied index to check whether the issue should be hidden, so the index needs to be valid.

Issue #2358, opened 2021-04-26
[TypeScript] Implement active rules config provider


duncanp-sonar:
### Description

Equivalent to `IActiveJavaScriptRulesProvider` - calculates the active set of rules, based on the default settings and any user configuration in `settings.json`.

PR #2365, opened 2021-04-27
Added TypeScript active rules provider

duncanp-sonar:
Relates to #2358

### DIFF
--- a/src/TypeScript.UnitTests/Rules/ActiveRulesCalculatorTests.cs
+++ b/src/TypeScript.UnitTests/Rules/ActiveRulesCalculatorTests.cs
@@ -1,0 +1,233 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract;
+using SonarLint.VisualStudio.TypeScript.Rules;
+
+namespace SonarLint.VisualStudio.TypeScript.UnitTests.Rules
+{
+    [TestClass]
+    public class ActiveRulesCalculatorTests
+    {
+        private static readonly IUserSettingsProvider EmptyUserSettingsProvider = new UserSettingsBuilder();
+
+        [TestMethod]
+        public void Get_NoUserOverrides_InactiveRulesAreNotReturned()
+        {
+            var ruleDefns = new RuleDefinitionsBuilder()
+                .AddRule("active 1", activeByDefault: true)
+                .AddRule("inactive AAA", activeByDefault: false)
+                .AddRule("active 2", activeByDefault: true)
+                .AddRule("inactive BBB", activeByDefault: false);
+
+            var testSubject = CreateTestSubject(ruleDefns, EmptyUserSettingsProvider);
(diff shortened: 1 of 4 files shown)


duncanp-sonar:
I've split the existing class into two logical parts: a language-agnostic helper that does the actual work, and the types needed for MEF exports.

src/TypeScript.UnitTests/Rules/ActiveRulesCalculatorTests.cs
```diff
@@ -23,29 +23,17 @@
...
 using SonarLint.VisualStudio.Core;
-using SonarLint.VisualStudio.Integration.UnitTests;
 using SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract;
 using SonarLint.VisualStudio.TypeScript.Rules;
 
 namespace SonarLint.VisualStudio.TypeScript.UnitTests.Rules
 {
     [TestClass]
-    public class ActiveJavaScriptRulesProviderTests
+    public class ActiveRulesCalculatorTests
```
duncanp-sonar, line 33: These tests now cover the helper class.
The only real change is to remove the MEF test.

src/TypeScript.UnitTests/Rules/ActiveRulesProviderTests.cs
```diff
@@ -0,0 +1,52 @@
...
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.TypeScript.Rules;
+
+namespace SonarLint.VisualStudio.TypeScript.UnitTests.Rules
+{
+    [TestClass]
+    public class ActiveRulesProviderTests
+    {
+        [TestMethod]
+        public void MefCtor_JavaScript_CheckIsExported()
```
duncanp-sonar, line 33: Tests for the JavaScript and TypeScript exports

src/TypeScript/Rules/ActiveRulesProviders.cs
```diff
@@ -0,0 +1,75 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract;
+
+namespace SonarLint.VisualStudio.TypeScript.Rules
```
duncanp-sonar, line 26: This file contains the MEF interfaces and exports.

We should have a look at consolidating some of the rules types: we have `IActiveJavaScriptRulesProvider`, `IJavaScriptRuleDefinitionsProvider` and `IJavaScriptRuleKeyMapper`, and the same again for TypeScript.
I'm not sure that they all required or all need to be MEF-exportable.

src/TypeScript/Rules/ActiveRulesProviders.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// Returns the eslint configuration for the currently active rules
+        /// </summary>
+        IEnumerable<Rule> Get();
+    }
+
+    [Export(typeof(IActiveJavaScriptRulesProvider))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class ActiveJavaScriptRulesProvider : IActiveJavaScriptRulesProvider
+    {
+        private readonly ActiveRulesCalculator calc;
```
duncanp-sonar, line 48: The first commit used inheritance rather than composition, but it seemed messy.
This approach makes it easier to see which types existing only to wire things together via MEF, which in turn should make it easier to refactor the construction if necessary.

src/TypeScript/Rules/ActiveRulesProviders.cs
```diff
@@ -0,0 +1,75 @@
...
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class ActiveTypeScriptRulesProvider : IActiveTypeScriptRulesProvider
+    {
+        private readonly ActiveRulesCalculator calc;
+
+        [ImportingConstructor]
+        public ActiveTypeScriptRulesProvider(ITypeScriptRuleDefinitionsProvider tsRuleDefinitions,
+            IUserSettingsProvider userSettingsProvider)
+        {
+            calc = new ActiveRulesCalculator(tsRuleDefinitions.GetDefinitions(), userSettingsProvider);
```
rita-sonarsource, line 70: Question: why not have ActiveRulesCalculator implement and MEF-export both interfaces? i.e.
```
internal class ActiveRulesCalculator : IActiveJavaScriptRulesProvider, IActiveTypeScriptRulesProvider
    {
    IEnumerable<Rule> IActiveJavaScriptRulesProvider.Get()
        {
            throw new System.NotImplementedException();
        }

        IEnumerable<Rule> IActiveTypeScriptRulesProvider.Get()
        {
            throw new System.NotImplementedException();
        }
}
```

duncanp-sonar: That's an option, but I'm not sure it solves the problem. We're building up parallel hierarchies for JS and TS stemming from the fact that the underlying `RulesRepository` class implements multiple interfaces and is MEF-created.
I'll have a look at simplifying the RuleDefinitionsProviders and then see what impact that has on these classes.

PR #2366, opened 2021-04-27
Merge IxxxKeyMapper interfaces into IxxxRuleDefinitionsProvider

duncanp-sonar:


src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -67,7 +66,7 @@ internal sealed class JavaScriptAnalyzer : IAnalyzer, IDisposable
             IAnalysisConfigMonitor analysisConfigMonitor,
             ILogger logger)
             : this(eslintBridgeClientFactory, eslintBridgeProcess, activeRulesProvider,
-                new EslintBridgeIssueConverter(keyMapper.GetSonarRuleKey,
```
duncanp-sonar, line 70: Previously, this class was importing two different single-method interfaces, then passing both methods as delegates to the issue converter.

It was always the case that the two interfaces were implemented on the same underlying class, but it wasn't as obvious before. Now it's easier to see as we are passing _(ruleDefinitionsProvider.X, ruleDefinitionsProvider.Y)_.

We should just be able to pass the _ruleDefinitionsProvider_ instance to the issue converter, but we still can't do that. Still, getting rid of the key mapper interfaces simplifies things a bit.

src/TypeScript/Rules/RuleDefinitionProviders.cs
```diff
@@ -28,6 +28,11 @@ internal interface ITypeScriptRuleDefinitionsProvider
         /// Returns the metadata descriptions for all TypeScript rules
         /// </summary>
         IEnumerable<RuleDefinition> GetDefinitions();
+
+        /// <summary>
+        /// Returns the rule key to display in VS for the specified ESLint rule
+        /// </summary>
+        string GetSonarRuleKey(string eslintRuleKey);
```
duncanp-sonar, line 35: Moved the single method from the key mapper interface to this class.
It seems reasonable that the class that is the source of the rule definitions should know how to map between the keys.

rita-sonarsource (approved):

Issue #2362, opened 2021-04-26
[EslintBridge] Implement tsconfig-files method


duncanp-sonar:
### Description

Implement request and response for `eslintbridge:tsconfig-files`.

PR #2368, opened 2021-04-28
Implement tsconfig-files method

duncanp-sonar:
Relates to #2362

### DIFF
--- a/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
@@ -164,6 +164,52 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
         }
 
         [TestMethod]
+        public async Task TSConfigFiles_HttpWrapperCalledWithCorrectArguments()
+        {
+            var httpWrapper = SetupHttpWrapper("tsconfig-files",
+                response: JsonConvert.SerializeObject(new TSConfigResponse()),
+                assertReceivedRequest: receivedRequest => receivedRequest.Should().Be("some path"));
+            var token = new CancellationToken();
+
+            var testSubject = CreateTestSubject(httpWrapper.Object);
+            await testSubject.TsConfigFiles("some path", token);
+
+            httpWrapper.Verify(x => x.PostAsync(BuildFullUri("tsconfig-files"), It.IsAny<object>(), token), Times.Once);
+            httpWrapper.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        [DataRow("")]
+        [DataRow(null)]
+        public void TSConfigFiles_EmptyResponse_InvalidOperationException(string response)
+        {
+            var httpWrapper = SetupHttpWrapper("tsconfig-files", response);
+            var testSubject = CreateTestSubject(httpWrapper.Object);
+            Func<Task> act = async () => await testSubject.TsConfigFiles("some path", CancellationToken.None);
+
+            act.Should().ThrowExactly<InvalidOperationException>();
+        }
+
+        [TestMethod]
+        public async Task TSConfigFiles_HasResponse_DeserializedResponse()
+        {
+            var response = new TSConfigResponse
+            {
+                Files = new[] { "file1", "file2" },
+                ProjectReferences = new[] { "ref1", "ref2" },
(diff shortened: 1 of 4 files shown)


src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
```diff
@@ -163,6 +163,52 @@ public void NewTsConfig_IncorrectResponse_Throws()
             act.Should().ThrowExactly<InvalidOperationException>();
         }
 
+        [TestMethod]
+        public async Task TSConfigFiles_HttpWrapperCalledWithCorrectArguments()
```
duncanp-sonar, line 167: Same set of tests as for `AnalyseJs` (cut-and-paste then modified)

src/TypeScript/EslintBridgeClient/Contract/AnalysisResponse.cs
```diff
@@ -26,7 +26,7 @@ namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract
     internal class AnalysisResponse
     {
         // Other fields omitted
-        // i.e. Hightlight[], HighlightedSymbol[], Metrics, CpdToken[]
+        // i.e. Highlight[], HighlightedSymbol[], Metrics, CpdToken[]
```
duncanp-sonar, line 29: Miscellaneous - fixing typo

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -115,6 +121,18 @@ public async Task NewTsConfig(CancellationToken cancellationToken)
             }
         }
 
+        public async Task<TSConfigResponse> TsConfigFiles(string tsConfigFilePath, CancellationToken cancellationToken)
+        {
+            var responseString = await httpWrapper.PostAsync(BuildServerUri("tsconfig-files"), tsConfigFilePath, cancellationToken);
+
+            if (string.IsNullOrEmpty(responseString))
```
duncanp-sonar, line 128: Same error handling as for `AnalyzeJS`. This wrapper will throw if we get an invalid response. The response could contain error codes, but it's up to the caller to handle them.

src/TypeScript/EslintBridgeClient/Contract/TSConfigResponse.cs
```diff
@@ -0,0 +1,46 @@
...
+    internal class TSConfigResponse
+    {
+        /// <summary>
+        /// File paths to source files referenced by the tsconfig file
+        /// </summary>
+        [JsonProperty("files")]
+        public IEnumerable<string> Files { get; set; }
+
+        /// <summary>
+        /// File paths of referenced projects that should be resolved by calling "tsconfig-files"
```
rita-sonarsource, line 35: I'm not sure I understand this comment. Will we ever need to use this property? Or do we only care about "Files"?

src/TypeScript/EslintBridgeClient/Contract/TSConfigResponse.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract
+{
+    internal class TSConfigResponse
+    {
+        /// <summary>
+        /// File paths to source files referenced by the tsconfig file
+        /// </summary>
+        [JsonProperty("files")]
+        public IEnumerable<string> Files { get; set; }
```
rita-sonarsource, line 32: Maybe "ReferencedFiles"? Not sure if it's better.

rita-sonarsource (approved):

src/TypeScript/EslintBridgeClient/Contract/TSConfigResponse.cs
duncanp-sonar: I'd rather stick to the same names as in the TS and Java code unless there's a strong reason to change them.

src/TypeScript/EslintBridgeClient/Contract/TSConfigResponse.cs
duncanp-sonar: I was going to exclude the property, but I had a look at the Java code and it is used. The Java code parses the response and calls `tsconfig-files` for each value in `ProjectReferences` (which suggests that a "project reference" is another tsconfig file).

PR #2369, opened 2021-04-28
Added IRulesProvider and factory

duncanp-sonar:
No changes to existing code - purely additive

### DIFF
--- a/src/TypeScript.UnitTests/Rules/RuleProviderTests.cs
+++ b/src/TypeScript.UnitTests/Rules/RuleProviderTests.cs
@@ -1,0 +1,71 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.TypeScript.Rules;
+
+namespace SonarLint.VisualStudio.TypeScript.UnitTests.Rules
+{
+    [TestClass]
+    public class RuleProviderTests
+    {
+        [TestMethod]
+        public void Ctor_InvalidArg_Throws()
+        {
+            Action act = () => new RuleDefinitionProvider(null);
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("ruleDefinitions");
+        }
+
+        [TestMethod]
+        public void GetDefinitions_ReturnsExpectedDefinitions()
+        {
+            var defns = new RuleDefinition[]
+            {
+                new RuleDefinition { RuleKey = "key1" },
+                new RuleDefinition { RuleKey = "key2" }
+            };
+
+            var testSubject = new RuleDefinitionProvider(defns);
+
+            testSubject.GetDefinitions().Should().BeEquivalentTo(defns);
+        }
+
+        [TestMethod]
+        [DataRow(null, null)]
+        [DataRow("unknown es lint key", null)]
+        [DataRow("eslint common", "typescript:common")]
+        [DataRow("eslint ts S1135", "typescript:S1135")]
+        [DataRow("ESLINT TS S1135", "typescript:S1135")] // case-insensitive
+        public void GetSonarRuleKey_TypeScript_ReturnsExpected(string eslintRuleKey, string expected)
+        {
+            var defns = new RuleDefinition[]
+            {
+                new RuleDefinition { EslintKey = "eslint common", RuleKey = "typescript:common"},
+                new RuleDefinition { EslintKey = "eslint ts S1135", RuleKey = "typescript:S1135"},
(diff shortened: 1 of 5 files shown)


duncanp-sonar:
I was trying to find a way to refactor the rules provider code as discussed this morning that didn't involve changing everything at once, which turned out to be rather difficult. This is the best I could come up with - adding a new set of types, and leave the existing code untouched.
I'll change the existing code to use the new classes in a separate commit.

src/TypeScript/Rules/RulesProviderFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+            }
+
+            Debug.Assert(!repoKey.EndsWith(":"), "Not expecting the repoKey to end with the colon separator character");
+
+            var allRules = Load(ruleMetadataFilePath);
+            var filteredRules = FilterByRepo(repoKey + ":", allRules);
+            return new RuleDefinitionProvider(filteredRules);
+        }
+
+        private static List<RuleDefinition> Load(string filePath) =>
```
duncanp-sonar, line 69: This method and the next one were copied from `RulesRepository`.

src/TypeScript/Rules/RulesProviderFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+        public IRuleProvider Create(string repoKey)
+        {
+            if (string.IsNullOrEmpty(repoKey))
+            {
+                throw new ArgumentNullException(nameof(repoKey));
+            }
+
+            Debug.Assert(!repoKey.EndsWith(":"), "Not expecting the repoKey to end with the colon separator character");
+
+            var allRules = Load(ruleMetadataFilePath);
```
duncanp-sonar, line 64: I thought about just loading the rules once on construction and storing them, but decided against it. Loading should be fast and we only expect to call the factory once per language.

src/TypeScript/Rules/RulesProviders.cs
```diff
@@ -0,0 +1,55 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SonarLint.VisualStudio.TypeScript.Rules
+{
+    internal interface IRuleProvider
```
duncanp-sonar, line 27: Exactly the same methods as `ITypeScriptRulesDefinitionProvider` and `IJavaScriptRulesDefinitionProvider`.

src/TypeScript/Rules/RulesProviders.cs
```diff
@@ -0,0 +1,55 @@
...
+        private IEnumerable<RuleDefinition> ruleDefinitions;
+
+        public RuleDefinitionProvider(IEnumerable<RuleDefinition> ruleDefinitions)
+        {
+            this.ruleDefinitions = ruleDefinitions ?? throw new ArgumentNullException(nameof(ruleDefinitions));
+        }
+
+        public IEnumerable<RuleDefinition> GetDefinitions() => ruleDefinitions;
+
+        public string GetSonarRuleKey(string eslintRuleKey) =>
```
duncanp-sonar, line 51: Copied from `RulesRepository`

src/TypeScript.UnitTests/Rules/RulesProviderFactoryTests.cs
```diff
@@ -0,0 +1,153 @@
...
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.TypeScript.Rules;
+
+namespace SonarLint.VisualStudio.TypeScript.UnitTests.Rules
+{
+    [TestClass]
+    public class RulesProviderFactoryTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
```
duncanp-sonar, line 35: Copied from the `RulesRepository` tests.

src/TypeScript.UnitTests/Rules/RulesProviderFactoryTests.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            var testSubject = new RulesProviderFactory(jsonFilePath);
+
+            Action act = () => testSubject.Create(prefix);
+
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("repoKey");
+        }
+
+        [TestMethod]
+        public void Create_ReturnsExpectedRulesForLanguage()
```
duncanp-sonar, line 60: This test and the two below were copied from the `RulesRepository` tests.

src/TypeScript/Rules/RulesProviderFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+            var filteredRules = FilterByRepo(repoKey + ":", allRules);
+            return new RuleDefinitionProvider(filteredRules);
+        }
+
+        private static List<RuleDefinition> Load(string filePath) =>
+            JsonConvert.DeserializeObject<List<RuleDefinition>>(File.ReadAllText(filePath, Encoding.UTF8));
```
duncanp-sonar, line 70: I started changing the class to use `IFileSystem` rather than reading from the file system directly but stopped: moving the sample json from the test files into string makes the test json harder to work with and the tests harder to read.

src/TypeScript/Rules/RulesProviders.cs
rita-sonarsource: Yay more identical interfaces! (kidding)

src/TypeScript.UnitTests/Rules/RuleProviderTests.cs
```diff
@@ -0,0 +1,71 @@
...
+            testSubject.GetDefinitions().Should().BeEquivalentTo(defns);
+        }
+
+        [TestMethod]
+        [DataRow(null, null)]
+        [DataRow("unknown es lint key", null)]
+        [DataRow("eslint common", "typescript:common")]
+        [DataRow("eslint ts S1135", "typescript:S1135")]
+        [DataRow("ESLINT TS S1135", "typescript:S1135")] // case-insensitive
+        public void GetSonarRuleKey_TypeScript_ReturnsExpected(string eslintRuleKey, string expected)
```
rita-sonarsource, line 58: The "TypeScript" in the middle can be removed

rita-sonarsource (approved):

src/TypeScript/Rules/RulesProviders.cs
rita-sonarsource, line 27: Minor: naming - `IRulesProvider` ? (plural)

PR #2370, opened 2021-04-28
Make EslintBridgeProcess multi-instance

### LINKED ISSUE #2356: Make EslintBridgeProcess multi-instance
### Description

JavaScript and TypeScript analysis will be handled separately i.e. separate analyzer classes talking to separate EsLintBridge NodeJS processes.

The `EslintBridgeProcess` process wraps the extenal Node process. Currently it a MEF-imported singleton consumed directly by the JavaScript analyzer. A separate instance will be required for the TypeScript analyzer.

TODO: looking at the relationship between the JavaScript analyzer, `IEslintBridgeClient` and `EslintBridgeProcess`. Does the analyzer need to know about both of the other classes?

rita-sonarsource:
Fixes #2356

### DIFF
--- a/src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
+++ b/src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
@@ -57,7 +57,7 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
 
         [ImportingConstructor]
         public JavaScriptAnalyzer(IEslintBridgeClientFactory eslintBridgeClientFactory,
-            IEslintBridgeProcess eslintBridgeProcess,
+            IEslintBridgeProcessFactory eslintBridgeProcessFactory,
             IJavaScriptRuleDefinitionsProvider ruleDefinitionsProvider,
             IActiveJavaScriptRulesProvider activeRulesProvider,
             ITelemetryManager telemetryManager,
@@ -65,7 +65,7 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
             IActiveSolutionTracker activeSolutionTracker,
             IAnalysisConfigMonitor analysisConfigMonitor,
             ILogger logger)
-            : this(eslintBridgeClientFactory, eslintBridgeProcess, activeRulesProvider,
+            : this(eslintBridgeClientFactory, eslintBridgeProcessFactory, activeRulesProvider,
                 new EslintBridgeIssueConverter(ruleDefinitionsProvider.GetSonarRuleKey,
                     ruleDefinitionsProvider.GetDefinitions),
                 telemetryManager,
@@ -77,7 +77,7 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
         }
 
         internal JavaScriptAnalyzer(IEslintBridgeClientFactory eslintBridgeClientFactory,
-            IEslintBridgeProcess eslintBridgeProcess,
+            IEslintBridgeProcessFactory eslintBridgeProcessFactory,
             IActiveJavaScriptRulesProvider activeRulesProvider,
             IEslintBridgeIssueConverter issuesConverter,
             ITelemetryManager telemetryManager,
@@ -87,7 +87,6 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
             ILogger logger)
         {
             this.eslintBridgeClientFactory = eslintBridgeClientFactory;
-            this.eslintBridgeProcess = eslintBridgeProcess;
             this.activeRulesProvider = activeRulesProvider;
             this.issuesConverter = issuesConverter;
             this.telemetryManager = telemetryManager;
@@ -95,6 +94,8 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
             this.activeSolutionTracker = activeSolutionTracker;
             this.analysisConfigMonitor = analysisConfigMonitor;
             this.logger = logger;
+            
+            eslintBridgeProcess = eslintBridgeProcessFactory.Create();
 
             activeSolutionTracker.ActiveSolutionChanged += ActiveSolutionTracker_ActiveSolutionChanged;
             analysisConfigMonitor.ConfigChanged += AnalysisConfigMonitor_ConfigChanged;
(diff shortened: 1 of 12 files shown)


src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -41,31 +41,27 @@ internal sealed class JavaScriptAnalyzer : IAnalyzer, IDisposable
     {
         private readonly EventWaitHandle serverInitLocker = new EventWaitHandle(true, EventResetMode.AutoReset);
 
-        private readonly IEslintBridgeClientFactory eslintBridgeClientFactory;
```
rita-sonarsource, line 44: As we discussed, the IEslintBridgeProcess is now removed from the analyzer. The analyzer is only aware of the Client now.

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
```diff
@@ -31,44 +31,55 @@ namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
...
 
     internal sealed class EslintBridgeHttpWrapper : IEslintBridgeHttpWrapper
     {
         private readonly ILogger logger;
         private readonly HttpClient httpClient;
+        private readonly IEslintBridgeProcess eslintBridgeProcess;
 
-        public EslintBridgeHttpWrapper(ILogger logger)
-            : this(new HttpClientHandler(), logger)
+        public EslintBridgeHttpWrapper(IEslintBridgeProcessFactory eslintBridgeProcessFactory, ILogger logger)
```
rita-sonarsource, line 43: I decided to place the process handling here for now. I'm still not sure how the Client will behave once we add the typescript stuff (will we have 2 clients? probably). So this made it simpler to keep the changes in the client to a minimum. I will revisit this when working on TypescriptAnalyzer, unless you think it's better to tackle it now.

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -45,11 +44,8 @@ internal interface IEslintBridgeProcess : IDisposable
         void Stop();
     }
 
-    [Export(typeof(IEslintBridgeProcess))]
-    [PartCreationPolicy(CreationPolicy.Shared)]
     internal sealed class EslintBridgeProcess : IEslintBridgeProcess
```
rita-sonarsource, line 47: EslintBridgeProcess is no longer singleton, and it has a factory that is responsible for importing `EslintBridgeDirectoryMefContractName`

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -243,12 +238,12 @@ private string FindServerScriptFile()
 
         private static string GetQuotedScriptPath(string scriptPath)
         {
-            const string Quote = "\"";
+            const string quote = "\"";
```
rita-sonarsource, line 241: Cosmetics, my resharper was complaining that it's not the naming convention for local consts

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
```diff
@@ -31,44 +31,55 @@ namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
...
             this.logger = logger;
             httpClient = new HttpClient(httpHandler);
+
+            eslintBridgeProcess = eslintBridgeProcessFactory.Create();
         }
 
-        public async Task<string> PostAsync(Uri serverEndpoint, object request, CancellationToken cancellationToken)
+        public async Task<string> PostAsync(string serverEndpoint, object request, CancellationToken cancellationToken)
         {
+            var port = await eslintBridgeProcess.Start();
```
rita-sonarsource, line 60: As we discussed offline, I believe the init-linter logic should be here as well.
At the moment, if the process crashes, there is no place that does the previous logic  of `if port has changed then call init-linter`. So what will happen is:
* we will spin a new process 
* call analyze-js on it and fail because init-linter was not called
* the exception will be caught in JavascriptAnalyzer that will mark "shouldInitLinter=true"
* the next analysis will succeed

I think it's not ideal, so I do think the init-linter logic should be in the same place that is responsible for spinning up a new process.

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
duncanp-sonar: I assumed we'd have two clients. A client that manages two processes seems unnecessarily complicated - the caller would always to specify which process it wanted to talk to.

rita-sonarsource: "init-linter" method aside, all the methods are either JS or TS, so it's possible to have one client that manages both. (I think it's better to have 2 clients though)

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
duncanp-sonar: So far, this http wrapper class knows nothing about any of the available eslintbridge APIs, so it seems weird that it should have special logic to know about one of them i.e. `init-linter`.

Also, `init-linter` isn't required if the caller just wants to call `tsconfig-files` to resolve the config-to-source mappings e.g. if we spun up a separate process just to do that as we discussed this morning. Calling `init-linter` wouldn't be a major overhead in that case, but it's a design smell that the division of responsibilities of this class and its consumers isn't correct.

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -155,25 +150,16 @@ internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, Ca
...
-            var port = await eslintBridgeProcess.Start();
-
             try
             {
                 serverInitLocker.WaitOne();
 
-                if (port != javascriptServerPort)
-                {
-                    javascriptServerPort = port;
-                    shouldInitLinter = true;
```
duncanp-sonar, line 172: Why isn't `shouldInitLinter` `true` when the analyzer is created? It knows it needs to call it, and that it hasn't already called it.

duncanp-sonar (approved):

Issue #2190, opened 2021-02-24
Implement TypeScript analyzer


duncanp-sonar:
### Description

Create implemention `IAnalyzer` that uses `eslintbridge`

#### TODO
- [x] ~~decide whether to implement separate JS and TS analyzers / separate `eslintbridge` instances~~ --> separate
- [x] ~~remove legacy analyzer implementation and tests~~ --> done in #2251 
- [x] ~~implement `eslintbridge` start-up. Analysis needs to be queued until `eslintbridge` is running.~~ --> done in #2184 
- [ ] fetch active rules. TODO - if the rules change, do we need to reset `eslintbridge`? YES.
- [ ] fetch appropriate `tsconfig` file
- [ ] if necessary, notify `eslintbridge` that a new `tsconfig` file is being used
- [ ] request analysis
- [ ] convert returned data to SLVS issues (see separate ticket for implementation rule key conversion)
- [ ] Use IAnalysisStatusNotifier for status reports

PR #2371, opened 2021-04-28
Add initial TypeScriptAnalyzer

rita-sonarsource:
Part of #2190


src/Core/Analysis/AnalysisLanguage.cs
```diff
@@ -27,6 +27,7 @@ public enum AnalysisLanguage
     {
         Javascript,
         CFamily,
-        RoslynFamily
+        RoslynFamily,
+        TypeScript
```
rita-sonarsource, line 31: I'll change SonarLanguageRecognizer in the next PR to actually detect the right content type.

duncanp-sonar (approved):

PR #2374, opened 2021-04-29
Changed EslintBridgeIssueConverter to use IRulesProvider

duncanp-sonar:
... rather that IJavaScriptRulesDefinitionProvider

### DIFF
--- a/src/EmbeddedVsix.UnitTests/TypeScriptRuleDefinitions_SanityChecks.cs
+++ b/src/EmbeddedVsix.UnitTests/TypeScriptRuleDefinitions_SanityChecks.cs
@@ -45,12 +45,12 @@ namespace SonarLint.VisualStudio.AdditionalFiles.UnitTests
             File.Exists(filePath).Should().BeTrue("Test setup error: could not find rule metadata file. Expected path: " + filePath);
 
             // Sanity check that the json file is loadable and has rules
-            var rulesProvider = new RulesRepository(filePath);
+            var factory = new RulesProviderFactory(filePath);
 
-            var jsRules = ((IJavaScriptRuleDefinitionsProvider)rulesProvider).GetDefinitions();
+            var jsRules = factory.Create("javascript").GetDefinitions();
             CheckRules("JavaScript", jsRules);
 
-            var tsRules = ((ITypeScriptRuleDefinitionsProvider)rulesProvider).GetDefinitions();
+            var tsRules = factory.Create("typescript").GetDefinitions();
             CheckRules("TypeScript", tsRules);
         }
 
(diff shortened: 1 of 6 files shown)


src/EmbeddedVsix.UnitTests/TypeScriptRuleDefinitions_SanityChecks.cs
```diff
@@ -45,12 +45,12 @@ public void LoadRuleMetadataFromFile()
             File.Exists(filePath).Should().BeTrue("Test setup error: could not find rule metadata file. Expected path: " + filePath);
 
             // Sanity check that the json file is loadable and has rules
-            var rulesProvider = new RulesRepository(filePath);
+            var factory = new RulesProviderFactory(filePath);
```
duncanp-sonar, line 48: Miscellaneous: changed the sanity check test to use the new factory.

As part of this work I had deleted originally _RulesRepository_ and the associated tests, which forced me to fix this up. However, I can't completely remove _RulesRepository_ yet as the active rules stuff still references the old IJavaScriptRulesDefinitionProvider interface, so I re-instated it for the time being.

src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
```diff
@@ -32,25 +32,20 @@ internal interface IEslintBridgeIssueConverter
...
-    internal delegate IEnumerable<RuleDefinition> GetRuleDefinitions();
-
     internal class EslintBridgeIssueConverter : IEslintBridgeIssueConverter
     {
-        private readonly ConvertToSonarRuleKey keyMapper;
-        private readonly GetRuleDefinitions ruleDefinitionProvider;
+        private readonly IRulesProvider rulesProvider;
 
-        public EslintBridgeIssueConverter(ConvertToSonarRuleKey keyMapper, GetRuleDefinitions ruleDefinitionProvider)
+        public EslintBridgeIssueConverter(IRulesProvider rulesProvider)
```
duncanp-sonar, line 39: The converter now consumes the common interface rather than separate delegates.

I was thinking about the `GetSonarRuleKey` method: perhaps it should just be a method on this class, rather than being part of the interface. Still, it would be easy to move later if necessary.

src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
```diff
@@ -32,25 +32,20 @@ internal interface IEslintBridgeIssueConverter
...
+            this.rulesProvider = rulesProvider;
         }
 
         public IAnalysisIssue Convert(string filePath, Issue issue)
         {
-            var ruleDefinitions = ruleDefinitionProvider();
+            var ruleDefinitions = rulesProvider.GetDefinitions();
             var ruleDefinition = ruleDefinitions.SingleOrDefault(x => x.EslintKey == issue.RuleId);
-            var sonarRuleKey = keyMapper(issue.RuleId);
+            var sonarRuleKey = rulesProvider.GetSonarRuleKey(issue.RuleId);
```
duncanp-sonar, line 48: In fact, the `GetSonarRuleKey` method is redundant, because we've already fetched the matching `RuleDefinition`. I'll look at removing it separately.

rita-sonarsource (approved):

PR #2378, opened 2021-04-29
Drop redundant GetSonarRuleKey method

duncanp-sonar:
* it was only used by the issue converter, which had already fetched the appropriate rule definition

### DIFF
--- a/src/TypeScript.UnitTests/Analyzer/EslintBridgeIssueConverterTests.cs
+++ b/src/TypeScript.UnitTests/Analyzer/EslintBridgeIssueConverterTests.cs
@@ -40,7 +40,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
         {
             var eslintBridgeIssue = new Issue
             {
-                RuleId = "rule id",
+                RuleId = "eslint rule id",
                 Column = 1,
                 EndColumn = 2,
                 Line = 3,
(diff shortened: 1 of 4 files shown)


src/TypeScript.UnitTests/Analyzer/EslintBridgeIssueConverterTests.cs
```diff
@@ -73,6 +73,58 @@ public void Convert_IssueConverted()
             convertedIssue.Severity.Should().Be(AnalysisIssueSeverity.Major);
         }
 
+        [TestMethod]
+        [DataRow("aaa", "aaa")]
+        [DataRow("aaa", "AAA")]
+        [DataRow("AAA", "aaa")]
+        [DataRow("AAA", "AAA")]
+        public void Convert_KeyMatchingIsCaseInsensitive(string keyInIssue, string keyInDefinition)
```
duncanp-sonar, line 81: New test for case-insensitive matching.

The previous code was inconsistent: the converter was case-sensitive when looking for the matching rule definition, but GetSonarRuleKey was case-insensitive.

src/TypeScript.UnitTests/Analyzer/EslintBridgeIssueConverterTests.cs
```diff
@@ -73,6 +73,58 @@ public void Convert_IssueConverted()
...
+            };
+
+            var testSubject = CreateTestSubject(ruleDefinitions);
+            var convertedIssue = testSubject.Convert("some file", eslintBridgeIssue);
+
+            convertedIssue.RuleKey.Should().Be("sonar rule key");
+        }
+
+        [TestMethod]
+        public void Convert_UnrecognisedESLintKey_Throws()
```
duncanp-sonar, line 101: Test-as-documentation

This isn't a behaviour we'd thought about explicitly. We could code defensively, but I don't think it's worth it this point.

src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
```diff
@@ -44,8 +44,8 @@ public EslintBridgeIssueConverter(IRulesProvider rulesProvider)
         public IAnalysisIssue Convert(string filePath, Issue issue)
         {
             var ruleDefinitions = rulesProvider.GetDefinitions();
-            var ruleDefinition = ruleDefinitions.SingleOrDefault(x => x.EslintKey == issue.RuleId);
```
duncanp-sonar, line 47: The old null-handling behaviour was inconsistent - `SingleOrDefault` is used here, but then we'd have got null ref when accessing `ruleDefinition.Severity/Type` below.

rita-sonarsource (approved):

PR #2381, opened 2021-04-29
Added active rules retrieval to the RulesProvider

duncanp-sonar:


src/TypeScript.UnitTests/Rules/ActiveRulesCalculatorTests.cs
```diff
@@ -45,7 +45,7 @@ public void Get_NoUserOverrides_InactiveRulesAreNotReturned()
 
             var testSubject = CreateTestSubject(ruleDefns, EmptyUserSettingsProvider);
 
-            var result = testSubject.Get().ToArray();
+            var result = testSubject.Calculate().ToArray();
```
duncanp-sonar, line 48: Miscellaneous - renamed the method

duncanp-sonar:
This change extends the `RulesProvider` so it also returns the active rules.
The next PR will drop the old `IJavaScriptRulesDefinitionProvider` etc and switch all of the client code to use the new `RulesProvider`.

rita-sonarsource (approved):

src/TypeScript.UnitTests/Rules/RulesProviderFactoryTests.cs
```diff
@@ -138,6 +143,23 @@ public void Create_GetDefinitions_WithDefaultParams_ReturnsExpectedConfigs()
             result[3].DefaultParams.Should().HaveCount(2);
         }
 
+        [TestMethod]
```
rita-sonarsource, line 146: Minor: I'm not sure this test belongs here. It's part of the internal RuleProvider implementation, not the factory.

duncanp-sonar: Yes and no.  The detailed behaviour of the RuleProvider is tested elsewhere.
However, we have no way of testing that the factory is producing valid artefacts except to test those artefacts i.e. it's more like an integration test.

I could perhaps have added some assertions relating to the mocked users settings to check that the created RuleProvider was initialized as expected.

PR #2382, opened 2021-04-29
[2] Drop JavaScript- and TypeScript-specific interfaces

duncanp-sonar:


src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -55,33 +55,33 @@ internal sealed class JavaScriptAnalyzer : IAnalyzer, IDisposable
...
             IAnalysisConfigMonitor analysisConfigMonitor,
-            ILogger logger)
+            ILogger logger,
+            IEslintBridgeIssueConverter issuesConverter = null // settable for testing
+            )
         {
-            this.activeRulesProvider = activeRulesProvider;
-            this.issuesConverter = issuesConverter;
+            this.rulesProvider = rulesProvider;
+            this.issuesConverter = issuesConverter ?? new EslintBridgeIssueConverter(rulesProvider);
```
duncanp-sonar, line 84: I changed the constructor pattern slightly so that we can still inject an issue converter for testing if we want to, but if we don't we'll get a default one.

This allows us to only call the factory once and store the returned `IRuleProvider` instance as well as using it to create the constructor.

There are other ways this could be done, but they involve either calling the factory twice, or creating an additional private constructor.

PR #2383, opened 2021-04-29
Drop JavaScript- and TypeScript-specific interfaces

duncanp-sonar:


src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -161,7 +161,7 @@ private async Task EnsureEslintBridgeClientIsInitialized(CancellationToken cance
 
                 if (shouldInitLinter)
                 {
-                    await eslintBridgeClient.InitLinter(activeRulesProvider.Get(), cancellationToken);
+                    await eslintBridgeClient.InitLinter(rulesProvider.GetActiveRulesConfiguration(), cancellationToken);
```
duncanp-sonar, line 164: I changed the constructor pattern slightly so that we can still inject an issue converter for testing if we want to, but if we don't we'll get a default one.

This allows us to only call the factory once and store the returned IRuleProvider instance as well as using it to create the constructor.

There are other ways this could be done, but they involve either calling the factory twice, or creating an additional private constructor (have a look at the first commit in isolation if you want to see what that option looks like).

rita-sonarsource (approved):

PR #2384, opened 2021-04-29
Skip ruleDefinitions without eslintKey

rita-sonarsource:


src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
```diff
@@ -44,7 +44,7 @@ public EslintBridgeIssueConverter(IRulesProvider rulesProvider)
         public IAnalysisIssue Convert(string filePath, Issue issue)
         {
             var ruleDefinitions = rulesProvider.GetDefinitions();
-            var ruleDefinition = ruleDefinitions.Single(x => x.EslintKey.Equals(issue.RuleId, StringComparison.OrdinalIgnoreCase));
+            var ruleDefinition = ruleDefinitions.Single(x => x.EslintKey != null && x.EslintKey.Equals(issue.RuleId, StringComparison.OrdinalIgnoreCase));
```
rita-sonarsource, line 47: This got broken in https://github.com/SonarSource/sonarlint-visualstudio/pull/2369, the original code was comparing the given value to the `x.EslintKey`, not the other way around, so no exceptions were thrown for null `EslintKey`.
I decided to go with the explicit `x.EslintKey != null` so it will be clearer (and added the test).

duncanp-sonar (approved):

PR #2387, opened 2021-04-30
Locate tsconfig files in the current solution

### LINKED ISSUE #2187: [TSConfig] Implement discovery of tsconfig.json settings files
### Description

Note: this applies when analysing TypeScript files only. JavaScript files use a fixed set of config options.

To correctly analyse TypeScript files we need to provide the appropriate `tsconfig.json` files at runtime.
This means we need to detect the solt

#### TODO
- [ ] Detect `tsconfig.json` files in the current solution
- [ ] ??? add mechanism for the user to specify the configuration to use per-solution?

rita-sonarsource:
Fixes #2187


src/TypeScript/TsConfig/TsConfigsLocator.cs
```diff
@@ -0,0 +1,68 @@
...
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace SonarLint.VisualStudio.TypeScript.TsConfig
+{
+    internal interface ITsConfigsLocator
+    {
+        /// <summary>
+        /// Returns all the tsconfig files in the current solution.
+        /// </summary>
+        /// <returns></returns>
```
duncanp-sonar, line 36: Empty `returns`.

src/TypeScript/TsConfig/TsConfigsLocator.cs
```diff
@@ -0,0 +1,68 @@
...
+            var foundFiles = new string[MaxNumberOfFiles];
+
+            var hr = vsUiShellOpenDocument.SearchProjectsForRelativePath(
+                (uint) __VSRELPATHSEARCHFLAGS.RPS_UseAllSearchStrategies,
+                "tsconfig.json",
+                foundFiles);
+
+            return hr != VSConstants.S_OK
+                ? Enumerable.Empty<string>()
+                : foundFiles.Where(x => !string.IsNullOrEmpty(x));
```
duncanp-sonar, line 65: It would be better to return a fixed set of results rather than an lazily-evaluated LINQ expression.

src/TypeScript/TsConfig/TsConfigsLocator.cs
```diff
@@ -0,0 +1,68 @@
...
+            vsUiShellOpenDocument = serviceProvider.GetService(typeof(SVsUIShellOpenDocument)) as IVsUIShellOpenDocument;
+        }
+
+        public IEnumerable<string> Locate()
+        {
+            var foundFiles = new string[MaxNumberOfFiles];
+
+            var hr = vsUiShellOpenDocument.SearchProjectsForRelativePath(
+                (uint) __VSRELPATHSEARCHFLAGS.RPS_UseAllSearchStrategies,
+                "tsconfig.json",
```
duncanp-sonar, line 60: I'm not clear on (a) what behaviour we actually want, and (b) whether this achieves it ;-)

For the scenarios we want to support, do we know if the `tsconfig.json` files will always be referenced as project files?
If they don't need to be referenced as project files, will this method find them? (from looking at the docs the answer seems to be "it depends").

duncanp-sonar (approved):

src/TypeScript/TsConfig/TsConfigsLocator.cs
rita-sonarsource: I tested what VS does if the tsconfig exists on disk but not referenced in the project -- it is still taken into account.
In which case, I believe this method would not work.

So WDYT? Should we change to searching using the file system? I think that's not ideal either, because you could have projects/files that are scattered in the file system and not necessarily under a common root. Moreover, searching by a common root might be incorrect. We could investigate and map VS's behavior in all the cases but I'm not sure if it's worth it. How do you feel about starting with this method and gathering feedback?

duncanp-sonar: Yep, I think it's ok to start with this - we have to start somewhere. I don't think searching the file system would be any better; at least this method has some relationship to the solution, even if we're not sure exactly what it is.

The main point was that it made me realise that we don't really know either what we want or what this does so we don't know if the behaviour makes sense, which is not a particularly comfortable place to be. We should definitely focus on testing this with some real-world projects when testing/dogfooding so have a reasonable level of confidence before releasing.

duncanp-sonar (approved):

PR #2388, opened 2021-04-30
Update sonarjs version

rita-sonarsource:


src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -137,6 +134,17 @@
     </ItemGroup>
   </Target>
 
+  <PropertyGroup>
+    <VsixUtilPath>$(PkgMicrosoft_VSSDK_BuildTools)\tools\vssdk\bin\VsixUtil.exe</VsixUtilPath>
+    <CorFlagsPath>$(CORFLAGS_PATH)</CorFlagsPath>
```
rita-sonarsource, line 139: I ended up modifying the build pipeline and adding "locate" step like we do for the signing tool. Msbuild was not finding it otherwise.

src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -137,6 +134,17 @@
...
+    <CorFlagsPath Condition="'$(CorFlagsPath)' == ''">corflags.exe</CorFlagsPath>
+  </PropertyGroup>
+  
+  <Target Name="ModifyVsixUtilCorFlags">
+    <Message Importance="high" Text="Modifying VsixUtil.exe CorFlags" />
+    <Exec Command="&quot;$(CorFlagsPath)&quot; &quot;$(VsixUtilPath)&quot; /32BIT- /force" />
```
rita-sonarsource, line 145: On my PC the flag is actually "/32BITREQ-", but the CI's version of corflags is different and has "/32BIT-" instead. I suppose we could modify our local versions manually as a workaround, or add a step that detects the version based on file's properties perhaps. WDYT?

src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -87,16 +87,13 @@
     <PluginUrl>https://binaries.sonarsource.com/Distribution/sonar-javascript-plugin/sonar-javascript-plugin-$(EmbeddedSonarJSAnalyzerVersion).jar</PluginUrl>
     <TypeScriptFolderName>ts</TypeScriptFolderName>
 	
-	<!-- TODO - delete the following property once the new version of the SonarJS plugin has been released. See #2270 -->
```
rita-sonarsource, line 90: This is no longer required, the version was released.

src/EmbeddedVsix/EmbeddedVsix.csproj
duncanp-sonar: How big is corflags.exe?  I think it's standalone and not very big, so it might be easier to check in a copy under the `build` directory. That would be simpler and work on both CI and locally without changes.

duncanp-sonar (approved):

PR #2395, opened 2021-05-06
Implement TSConfig provider

### LINKED ISSUE #2360: [TSConfig] Implement TSConfig provider
### Description

Implement a component that returns the tsconfig file to use for a specified source file

#### Background
* a tsconfig file can contain inclusions and exclusions that together determine which source files the configuration applies to.
* a project could contain multiple tsconfig files at different levels (TODO - check; is this common? Might be more of an issue if the user is opening a folder rather than a project).
* a source file could be referred to by multiple tsconfig files, in which case the closest tsconfig file should be used.
* the tsconfig -> source files mapping is calculated by calling `eslintbridge:tsconfig-files`.

### Implementation
The component should maintain a mapping between tsconfig files and the source files to which they relate.

 #### Logical methods:

```
interface ITSConfigProvider
{
  // Notifies the provider that a configuration file has been discovered, added or updated
  void Update(string tsConfigFilePath);

  // Returns the path to the tsconfig file path to use for the specified source file.
  // Returns null (?) if a tsconfig file could not be found.
  string FindConfigFile(string sourceFilePath);
}

```

#### Logical implementation - Update
* remove any data for the specified config file
* call `eslintbridge:tsconfig-files` to resolve the list of source files
* update the source -> config mapping.

#### Logical implementation - FindConfigFile
* look up the source file in the mapping and return the closest match

#### source file referenced in multiple tsconfigs
Disk files:
-- ParentDir
  -- configA
  -- childDir
  -- -- configB
  -- -- fileA

Both configA and configB reference fileA.
* Update(configA) -> FindConfigFile should return configA
* Update(configB) -> FindConfigFile should return configB since closer
* configB is edited / deleted so it doesn't refer to fileA -> FindConfigFile should return configA

Unless we remember that configA also referenced fileA we'll return the wrong config after configB is edited / deleted.
Is this important enough to worry about?

rita-sonarsource:
Fixes #2360

### DIFF
--- a/src/TypeScript/TsConfig/TsConfigProvider.cs
+++ b/src/TypeScript/TsConfig/TsConfigProvider.cs
@@ -19,7 +19,6 @@
  */
 
 using System;
-using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


src/TypeScript/TsConfig/TsConfigProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public async Task<string> GetConfigForFile(string sourceFilePath, IEslintBridgeClient eslintBridgeClient, CancellationToken cancellationToken)
+        {
+            var allTsConfigsFilePaths = tsConfigsLocator.Locate();
+
+            foreach (var tsConfigsFilePath in allTsConfigsFilePaths)
+            {
+                var response = await eslintBridgeClient.TsConfigFiles(tsConfigsFilePath, cancellationToken);
+
+                if (response.Files != null && response.Files.Contains(sourceFilePath, StringComparer.OrdinalIgnoreCase))
```
rita-sonarsource, line 58: So the first commit I added was also checking the project references, however, I'm not sure if this is relevant in VS context.
We already gathered all the tsconfigs that are referenced in the solution and that should be our 'source of truth'.
Am I missing something here?

src/TypeScript/TsConfig/TsConfigProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+using SonarLint.VisualStudio.TypeScript.EslintBridgeClient;
+
+namespace SonarLint.VisualStudio.TypeScript.TsConfig
+{
+    internal interface ITsConfigProvider
+    {
+        /// <summary>
+        /// Returns the tsconfig file for the given source file, or null if an applicable tsconfig could not be found.
+        /// </summary>
+        Task<string> GetConfigForFile(string sourceFilePath, IEslintBridgeClient eslintBridgeClient, CancellationToken cancellationToken);
```
rita-sonarsource, line 35: As we discussed offline, `IEslintBridgeClient` is not MEF exportable and hence we need to accept it as a param.
I'm working on the TypescriptAnalyzer and I think I will create a MEF-exportable `ITypeScriptEslintBridgeClient` that this class could import. This is a workaround for now.

src/TypeScript/TsConfig/TsConfigProvider.cs
duncanp-sonar: No idea. The Java sensor code appears to be both steps i.e. locating files on disc then processing the project references too. We're not actually sure what a "project reference" is in the real world; presumably it has the same format as a tsconfig.json file, but it might be called something different.
We'll need to look at some real-world examples to work out what we should do - might be worth creating an investigate/check ticket to remind us to do it.

duncanp-sonar (approved):

Issue #2287, opened 2021-04-07
TaintSyncPackage does not load correctly in Visual Studio 2019


rlusian1:
### Description

We have multiple developers who are getting error messages about the TaintSyncPackage not being able to load:

---------------------------
Microsoft Visual Studio
---------------------------
The 'TaintSyncPackage' package did not load correctly.
The problem may have been caused by a configuration change or by the installation of another extension. You can get more information by examining the file 'C:\Users\itron\AppData\Roaming\Microsoft\VisualStudio\16.0_c7f9311d\ActivityLog.xml'.

This is an excerpt from an ActivityLog.xml file:

 <entry>
    <record>971</record>
    <time>2021/04/05 16:34:01.235</time>
    <type>Information</type>
    <source>VisualStudio</source>
    <description>Begin package load [TaintSyncPackage]</description>
    <guid>{EEEB81FA-D6C3-438D-B29E-9FAAC91471CC}</guid>
  </entry>
  <entry>
    <record>972</record>
    <time>2021/04/05 16:34:01.265</time>
    <type>Error</type>
    <source>VisualStudio</source>
[...]

PR #2407, opened 2021-05-07
Updated webclient to get fix for #2406

### LINKED ISSUE #2406: SonarQubeService.IsConnected should be false if a connection could not be established
### Description

It is expected that the call to SonarQubeService.ConnectAsync could fail e.g. because the server is not running or not reachable.
However, SonarQubeService.IsConnected should be false in that case.

### Repro steps

1. Connect a solution to a local SQ
2. Shutdown VS
3. Shutdown SQ
4. Open the solution in VS so one of the services calls "ConnectAsync".
The call to "ConnectAsync" will fail.

### Expected behavior

SonarQubeService.IsConnected should be false.

### Actual behavior

SonarQubeService.IsConnected is true.
This can cause failures in consuming components that check IsConnected before performing an operation e.g. #2287.

### Known workarounds

None.

### Related information

* SonarLint for VS Version: reported in v4.33 but definitely existed in earlier versions.
* SonarQube server version: n/a

duncanp-sonar:
Fixes #2406 and #2287

### DIFF
--- a/src/Integration/MefServices/ActiveSolutionBoundTracker.cs
+++ b/src/Integration/MefServices/ActiveSolutionBoundTracker.cs
@@ -127,9 +127,6 @@ namespace SonarLint.VisualStudio.Integration
                 await Core.WebServiceHelper.SafeServiceCallAsync(() => sonarQubeService.ConnectAsync(connectionInformation,
                     CancellationToken.None), this.logger);
             }
-
-            Debug.Assert((boundProject != null) == sonarQubeService.IsConnected,
-                $"Inconsistent connection state: Solution bound={boundProject != null}, service connected={sonarQubeService.IsConnected}");
         }
 
         private void OnBindingStateChanged(object sender, BindingStateEventArgs e)


src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -127,9 +127,6 @@ private async Task UpdateConnectionAsync()
                 await Core.WebServiceHelper.SafeServiceCallAsync(() => sonarQubeService.ConnectAsync(connectionInformation,
                     CancellationToken.None), this.logger);
             }
-
-            Debug.Assert((boundProject != null) == sonarQubeService.IsConnected,
-                $"Inconsistent connection state: Solution bound={boundProject != null}, service connected={sonarQubeService.IsConnected}");
```
duncanp-sonar, line 132: This assertion was incorrect; we do have binding information, but could not connect.

I've manually check the TaintPackage load bug doesn't repro, and didn't find any issues apart from this assertion.

The message logging for the other features (suppressions, taint issues etc) looks pretty good e.g. along the lines of "couldn't update suppressions - server is not connected".

rita-sonarsource (approved):

Issue #2409, opened 2021-05-10
init-linter is not called again when eslint-bridge process crashes


rita-sonarsource:
#### Description

Analysis will fail if eslint-bridge process crashes after the server was initialized successfully with init-linter.

Steps to reproduce:
1. Analyze a ts file --> eslint-bridge process starts, init-linter is called, analysis is successful
2. Open task manager and kill nodejs processes under Visual Studio
3. Re-analyze a ts file

Expected --> analysis is successful
Actual --> analysis fails with `GENERAL_ERROR Linter is undefined. Did you call /init-linter?`
Subsequent analysis also fails.

PR #2410, opened 2021-05-11
Return StartResult in EslintBridgeProcess.Start

rita-gorokhod:
Part of #2409


src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -85,9 +104,12 @@ public Task<int> Start()
 
                 if (!shouldSpawnNewProcess)
                 {
-                    return startTask.Task;
+                    return startTask.Task.IsCompleted
```
rita-sonarsource, line 107: If this method gets called 10 times before the server starts, all 10 threads will receive a response that this is a new process. The 11th thread will receive a response that this is not a process.

It will be up to the callers to handle locking the init for the first 10 calls.

PR #2414, opened 2021-05-12
Ensure InitLinter is called when new EslintBridgeProcess is spawned

### LINKED ISSUE #2409: init-linter is not called again when eslint-bridge process crashes

rita-sonarsource:
Fixes #2409

### DIFF
--- a/src/TypeScript/Analyzer/TypeScriptAnalyzer.cs
+++ b/src/TypeScript/Analyzer/TypeScriptAnalyzer.cs
@@ -150,6 +150,10 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
                     consumer.Accept(filePath, issues);
                 }
             }
+            catch (EslintBridgeClientNotInitializedException)
+            {
+                RequireLinterUpdate();
+            }
             catch (TaskCanceledException)
             {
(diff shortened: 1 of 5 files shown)


src/TypeScript/Analyzer/TypeScriptAnalyzer.cs
```diff
@@ -150,6 +150,10 @@ internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, Ca
                     consumer.Accept(filePath, issues);
                 }
             }
+            catch (EslintBridgeClientNotInitializedException)
```
rita-sonarsource, line 153: The analysis will fail but the next call will succeed.
I'll handle it better as part of #2190

PR #2422, opened 2021-05-12
Add TypeScriptAnalyzer

rita-sonarsource:
Changes:

* New class `EslintBridgeAnalyzer`, that `TypeScriptAnalyzer` and `JavaScriptAnalyzer` use.
* `EslintBridgeAnalyzer` receives the `IEslintBridgeClient` and calls its new "Analyze" method.
* `EslintBridgeClient` is now a base class. New classes `JavaScriptEslintBridgeClient` and `TypeScriptEslintBridgeClient`, both inherit from `EslintBridgeClient` and pass to it the name of the `analyze` endpoint (analyze-js or analyze-ts).

* `EslintBridgeProcess` now returns `StartResult`, that contains information if the process is new (and should be `init-linter`ed).
* `EslintBridgeClient` now holds `EslintBridgeProcess` and throws `EslintBridgeClientNotInitializedException` the process is new. The caller (`EslintBridgeAnalyzer`) catches this and calls InitLinter.


rita-sonarsource:
@duncanp-sonar please take a look at the description for the summary of the changes.
It might be easier to just check out the code and look at the final result of how everything works together, but I've kept each commit to a minimum so you could review each commit separately.

duncanp-sonar (approved):

PR #2426, opened 2021-05-13
Don't shutdown eslint-bridge process when settings are updated

### LINKED ISSUE #2425: Updating settings file in the middle of an analysis causes exceptions
#### Description

When the settings file is changed, eslint-bridge server is shut down. If this happens in the middle of a running analysis, an exception appears in the output window.

rita-sonarsource:
Fixes #2425

### DIFF
--- a/src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
+++ b/src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
@@ -135,9 +135,9 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
             await StopServer();
         }
 
-        private async void AnalysisConfigMonitor_ConfigChanged(object sender, EventArgs e)
+        private void AnalysisConfigMonitor_ConfigChanged(object sender, EventArgs e)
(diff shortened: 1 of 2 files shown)


src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
```diff
@@ -135,9 +135,9 @@ private async void ActiveSolutionTracker_ActiveSolutionChanged(object sender, Ac
             await StopServer();
         }
 
-        private async void AnalysisConfigMonitor_ConfigChanged(object sender, EventArgs e)
+        private void AnalysisConfigMonitor_ConfigChanged(object sender, EventArgs e)
         {
-            await StopServer();
+            RequireLinterUpdate();
```
rita-sonarsource, line 140: In terms of performance, there is no reason to shutdown the entire process when settings are changed, only when the solution is changed.
Restarting the server can take about 10 seconds, so this was very inefficient (and was also causing exceptions if the settings file was saved multiple times, see #2425 )

PR #2428, opened 2021-05-17
Fix calling init-linter when eslint-bridge process crashes

### LINKED ISSUE #2409: init-linter is not called again when eslint-bridge process crashes

rita-sonarsource:
Re-fix #2409

### DIFF
--- a/src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
+++ b/src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
@@ -137,7 +137,10 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
                 .ReturnsAsync(validResponse)
                 .ReturnsAsync(validResponse);
 
-            var testSubject = CreateTestSubject(client.Object);
+            var activeRules = new[] { new Rule { Key = "rule1" }, new Rule { Key = "rule2" } };
+            var activeRulesProvider = SetupActiveRulesProvider(activeRules);
+
+            var testSubject = CreateTestSubject(client.Object, rulesProvider: activeRulesProvider.Object);
 
             await testSubject.Analyze("some path", "some config", CancellationToken.None);
             client.Verify(x => x.InitLinter(It.IsAny<IEnumerable<Rule>>(), It.IsAny<CancellationToken>()), Times.Once);
@@ -150,6 +153,8 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
 
             await testSubject.Analyze("some path", "some config", CancellationToken.None);
             client.Verify(x => x.InitLinter(It.IsAny<IEnumerable<Rule>>(), It.IsAny<CancellationToken>()), Times.Exactly(2));
+
+            activeRulesProvider.Verify(x=> x.GetActiveRulesConfiguration(), Times.Exactly(2));
         }
 
         [TestMethod]
@@ -181,11 +186,11 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
         }
 
         [TestMethod]
-        public async Task Analyze_EslintBridgeClientNotInitializedResponse_FollowedByValidResponse_ConvertedIssuesReturned()
+        public async Task Analyze_EslintBridgeClientNotInitializedResponse_FollowedByValidResponse_ConvertedIssuesReturnedAndParsingErrorNotLogged()
         {
             var linterNotInitializedResponse = CreateLinterNotInitializedResponse();
             var validResponse = new AnalysisResponse {Issues = new List<Issue> {new Issue()}};
-
+            var logger = new TestLogger();
             var client = new Mock<IEslintBridgeClient>();
 
             client
@@ -197,10 +202,12 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
             var issueConverter = new Mock<IEslintBridgeIssueConverter>();
             SetupConvertedIssue(issueConverter, "some path", validResponse.Issues.First(), convertedIssues[0]);
 
-            var testSubject = CreateTestSubject(client.Object, issueConverter: issueConverter.Object);
+            var testSubject = CreateTestSubject(client.Object, issueConverter: issueConverter.Object, logger: logger);
             var result = await testSubject.Analyze("some path", "some config", CancellationToken.None);
 
             result.Should().BeEquivalentTo(convertedIssues);
+            logger.AssertNoOutputMessages();
+            client.Verify(x => x.Analyze("some path", "some config", CancellationToken.None), Times.Exactly(2));
         }
 
         [TestMethod]
--- a/src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
+++ b/src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
@@ -34,6 +34,7 @@ using SonarLint.VisualStudio.TypeScript.Analyzer;
 using SonarLint.VisualStudio.TypeScript.EslintBridgeClient;
 using SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract;
 using SonarLint.VisualStudio.TypeScript.Rules;
+using Resources = SonarLint.VisualStudio.TypeScript.Analyzer.Resources;
 
 namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
 {
@@ -118,9 +119,10 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
         }
 
         [TestMethod]
-        public async Task Analyze_EslintBridgeClientNotInitializedException_CallsInitLinter()
+        public async Task Analyze_EslintBridgeClientNotInitializedResponse_CallsInitLinter()
         {
             var validResponse = new AnalysisResponse { Issues = new List<Issue>() };
+            var linterNotInitializedResponse = new AnalysisResponse { ParsingError = new ParsingError{Message = Resources.ERR_ParsingError_LinterNotInitialized} };
             var client = new Mock<IEslintBridgeClient>();
 
             // First analysis: response is valid, InitLinter should be called once (first initialization)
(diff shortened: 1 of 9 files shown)


src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -116,15 +100,7 @@ public async Task<AnalysisResponse> Analyze(string filePath, string tsConfigFile
                 TSConfigFilePaths = tsConfigFilePaths
             };
 
-            var result = await eslintBridgeProcess.Start();
```
rita-sonarsource, line 119: I had to revert this change. My original reasoning for this change was that:
1. The process could die and start again on the same port, so the old `if port != currentPort` logic is flawed.
2. This class is the one that knows that a process was spawned, so why shouldn't it return this information back to the callers.

And for JS that worked, `EslintBridgeClient.Analyze` got back the response that this is a new process, threw an exception and init-linter was called.
But for TS, unfortunately TsConfigProvider was called first, which ignored the `IsNewProcess` flag completely. `EslintBridgeClient.Analyze` then received `IsNewProcess=false` and init-linter was not called. This bug happened only after the process died, since the init-linter was always called at least once.

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
```diff
@@ -70,28 +70,22 @@ internal sealed class EslintBridgeAnalyzer : IEslintBridgeAnalyzer
...
         {
             await EnsureEslintBridgeClientIsInitialized(rulesProvider.GetActiveRulesConfiguration(), cancellationToken);
-
             var analysisResponse = await eslintBridgeClient.Analyze(filePath, tsConfig, cancellationToken);
 
             if (analysisResponse.ParsingError != null)
             {
-                LogParsingError(filePath, analysisResponse.ParsingError);
-                return Array.Empty<IAnalysisIssue>();
+                if (analysisResponse.ParsingError.Message.Contains(Resources.ERR_ParsingError_LinterNotInitialized))
```
rita-sonarsource, line 79: This is now the bug fix: I decided to leave the decision up to eslint-bridge itself. If it returns an error that init-linter should be called, I call it.

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
```diff
@@ -70,28 +70,22 @@ internal sealed class EslintBridgeAnalyzer : IEslintBridgeAnalyzer
...
+                {
+                    await eslintBridgeClient.InitLinter(rulesProvider.GetActiveRulesConfiguration(), cancellationToken);
```
rita-sonarsource, line 81: I thought about using recursion and just re-invoking this method, but I was worried about somehow getting stuck in an infinite loop. So I'm just directly calling  `eslintBridgeClient.InitLinter` and `eslintBridgeClient.Analyze`.

src/TypeScript/Analyzer/Resources.resx
```diff
@@ -123,6 +123,9 @@
   <data name="ERR_ParsingError_General" xml:space="preserve">
     <value>Failed to parse file '{0}' at line {1}: {2} {3}</value>
   </data>
+  <data name="ERR_ParsingError_LinterNotInitialized" xml:space="preserve">
+    <value>Linter is undefined</value>
```
duncanp-sonar, line 127: Isn't this a constant value defined by an external component i.e. `eslintbridge`? Do we know if the returned string is localised? (if it is localised, we have a problem since we only handle a single default language)

Assuming it is an external, non-localised fixed value, it would be better to define it is a constant in the code. There's no reason for it to be a resource.

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
duncanp-sonar: Perhaps we should ask the JS bubble to return a specific error code so we don't have to parse the string.

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
duncanp-sonar: The flow isn't quite right - if the second call to `analyze` returns a parsing error then it won't be logged. Logically a parsing failure in the second "analyze" should be handled the same as if it occurred in the first "analyze" i.e.

```
response = await analyze;
if (linter is not initialized)
{
  await init-linter;
  response = await analyze;
}

if (parsing error in response)
{
 ...
}

```

duncanp-sonar (commented):
Reverting to the previous approach makes sense, but there are a couple of minor bits that don't look quite right.

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
rita-sonarsource: Yep, that could be great. I'll ask Tibor about the localization.

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
rita-sonarsource: Woops, you're right. Nice catch 👍

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
```diff
@@ -74,18 +77,17 @@ public async Task<IReadOnlyCollection<IAnalysisIssue>> Analyze(string filePath,
             await EnsureEslintBridgeClientIsInitialized(rulesProvider.GetActiveRulesConfiguration(), cancellationToken);
             var analysisResponse = await eslintBridgeClient.Analyze(filePath, tsConfig, cancellationToken);
 
+            if (analysisResponse.ParsingError != null && 
+                analysisResponse.ParsingError.Message.Contains(LinterIsNotInitializedError))
+            {
+                await eslintBridgeClient.InitLinter(rulesProvider.GetActiveRulesConfiguration(), cancellationToken);
```
duncanp-sonar, line 83: `rulesProvider.GetActiveRulesConfiguration()` was called above. Do we want to recalculate the value, or reuse the one from above?

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
```diff
@@ -74,18 +77,17 @@ public async Task<IReadOnlyCollection<IAnalysisIssue>> Analyze(string filePath,
             await EnsureEslintBridgeClientIsInitialized(rulesProvider.GetActiveRulesConfiguration(), cancellationToken);
             var analysisResponse = await eslintBridgeClient.Analyze(filePath, tsConfig, cancellationToken);
 
+            if (analysisResponse.ParsingError != null &&
```
duncanp-sonar, line 80: We've just called the method `EnsureEslintBridgeClientIsInitialized` above, so it's not obvious why we'd still need to explicitly handle "not initialized". I think it needs a comment here, and possibly in `EnsureEslintBridgeClientIsInitialized`.

src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
```diff
@@ -152,6 +152,75 @@ public async Task Analyze_EslintBridgeClientNotInitializedResponse_CallsInitLint
...
+                .ReturnsAsync(validResponse);
+
+            var convertedIssues = new[] { Mock.Of<IAnalysisIssue>() };
+            var issueConverter = new Mock<IEslintBridgeIssueConverter>();
+            SetupConvertedIssue(issueConverter, "some path", validResponse.Issues.First(), convertedIssues[0]);
+
+            var testSubject = CreateTestSubject(client.Object, issueConverter: issueConverter.Object);
+            var result = await testSubject.Analyze("some path", "some config", CancellationToken.None);
+
+            result.Should().BeEquivalentTo(convertedIssues);
```
duncanp-sonar, line 203: It wouldn't hurt to add some extra checks here e.g. that `analyze` is called twice, and that the "not initialized" message is not logged.

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
rita-sonarsource: Good question. From this class' point of view, it shouldn't matter, but since we know that it's re-reading the file behind the scene, it'd be better to re-use the values. I'll change.

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
rita-sonarsource: Originally I thought about calling 
```
RequireLinterUpdate();
await EnsureEslintBridgeClientIsInitialized();
```
But thought it would be more confusing. I'll change it to that since I see that this is more confusing :D

src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
```diff
@@ -71,22 +73,15 @@ internal sealed class EslintBridgeAnalyzer : IEslintBridgeAnalyzer
...
         {
-            try
-            {
-                return await GetFileIssues(filePath, tsConfig, cancellationToken);
-            }
-            catch (EslintBridgeClientNotInitializedException)
+            await EnsureEslintBridgeClientIsInitialized(cancellationToken);
+            var analysisResponse = await eslintBridgeClient.Analyze(filePath, tsConfig, cancellationToken);
+
+            if (LinterNotInitializedResponse(analysisResponse))
```
duncanp-sonar, line 79: I still don't think it's obvious _why_ we might get a "not initialised" error having just called `EnsureEslintBridgeClientIsInitialized`. I think it needs a comment e.g.

```
// The call to `EnsureEslintBridgeClientIsInitialized` above doesn't guarantee the client is correctly initialized (e.g. the external process might have crashed).
// So we still need to handle the "not initialised" case here.
```

PR #2433, opened 2021-05-18
Add VsHierarchyLocator

rita-sonarsource:
Prep for #2430


src/Integration/LocalServices/ProjectSystemHelper.cs
```diff
@@ -67,6 +67,13 @@ public ProjectSystemHelper(IServiceProvider serviceProvider)
             this.serviceProvider = serviceProvider;
         }
 
+        public IVsHierarchy GetFileVsHierarchy(string fileName)
```
rita-sonarsource, line 70: I debated if it's better to implement this in the Infrastructure.VS project, but:
1. I didn't want to have it reference the DTE
2. I needed `GetIVsHierarchy` method, which also relies on DTE

src/Infrastructure.VS/IVsHierarchyLocator.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace SonarLint.VisualStudio.Infrastructure.VS
+{
+    public interface IVsHierarchyLocator
+    {
+        /// <summary>
+        /// Returns the project in which the given <see cref="fileName"/> is located.
+        /// </summary>
+        /// <returns>Null if the file is not found in any project in the solution.</returns>
+        IVsHierarchy GetFileVsHierarchy(string fileName);
```
rita-sonarsource, line 31: Prep for the new tsConfigLocator logic.

duncanp-sonar, line 31: `GetContainingProject` is more descriptive of what the method does. Or `GetVsHierarchyForFile`.

PR #2434, opened 2021-05-18
Add FilePathsLocator

rita-sonarsource:
Prep for #2430


src/Infrastructure.VS/FilePathsLocator.cs
```diff
@@ -0,0 +1,118 @@
...
+            GetItemId(hierarchy, rootId, __VSHPROPID.VSHPROPID_FirstChild);
+
+        private static uint NextSibling(IVsHierarchy hierarchy, uint firstId) =>
+            GetItemId(hierarchy, firstId, __VSHPROPID.VSHPROPID_NextSibling);
+
+        private static uint GetItemId(IVsHierarchy hierarchy, uint itemId, __VSHPROPID propId)
+        {
+            hierarchy.GetProperty(itemId, (int)propId, out var idObj);
+
+            return idObj == null ? (uint)VSConstants.VSITEMID.Nil : (uint)(int)idObj;
```
duncanp-sonar, line 107: Why the double cast? `(uint)(int)idObj`

src/Infrastructure.VS/FilePathsLocator.cs
```diff
@@ -0,0 +1,118 @@
...
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace SonarLint.VisualStudio.Infrastructure.VS
+{
+    public interface IFilePathsLocator
+    {
+        /// <summary>
+        /// Returns the absolute file paths of <see cref="fileName"/> in the given <see cref="IVsHierarchy"/>
+        /// </summary>
+        IReadOnlyList<string> Locate(IVsHierarchy hierarchyToSearch, string fileName);
```
duncanp-sonar, line 36: We'd normally return an `IEnumerable`. Is there a reason to return a read-only list? (it implies that order is important)

duncanp-sonar (approved):

rita-sonarsource:
As we discussed offline, I'm closing this PR and changing the search to a file system based one.

PR #2435, opened 2021-05-18
Find all tsconfigs in analyzed file's project

### LINKED ISSUE #2430: TsConfigLocator does not find all the tsconfig.json files in the project/solution

rita-sonarsource:
Fixes #2430


src/TypeScript/TsConfig/TsConfigProvider.cs
```diff
@@ -55,7 +56,9 @@ public TsConfigProvider(ITsConfigsLocator tsConfigsLocator, ITypeScriptEslintBri
 
         public async Task<string> GetConfigForFile(string sourceFilePath, CancellationToken cancellationToken)
         {
-            var allTsConfigsFilePaths = tsConfigsLocator.Locate();
+            var allTsConfigsFilePaths = tsConfigsLocator.Locate(sourceFilePath);
+
+            logger.WriteLine(Resources.INFO_FoundTsConfigs, string.Join(Path.PathSeparator.ToString(), allTsConfigsFilePaths));
```
duncanp-sonar, line 61: Verbosity - do we want to print this list on every analysis?
Perhaps we should drop it to DEBUG. I think printing the tsconfig that is actually being used each time is ok, but maybe not the whole list.

src/TypeScript/TsConfig/TsConfigsLocator.cs
```diff
@@ -21,47 +21,109 @@
...
+            return foundTsConfigs.ToArray();
         }
 
-        public IReadOnlyList<string> Locate()
+        private string GetProjectDirectory(string sourceFilePath)
         {
-            var foundFiles = new string[MaxNumberOfFiles];
+            var hr = vsSolution.GetProperty(VSPROPID_IsInOpenFolderMode, out var isOpenAsFolder);
+
+            Debug.Assert(hr == VSConstants.S_OK || hr == VSConstants.E_NOTIMPL, "Failed to retrieve VSPROPID_IsInOpenFolderMode");
```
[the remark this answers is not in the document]
rita-sonarsource: It's not gonna fire, no, it returns `E_NOTIMPL` in 2015.

src/TypeScript/TsConfig/TsConfigsLocator.cs
```diff
@@ -21,47 +21,109 @@
...
+                .Where(x=> !x.Contains("\\node_modules\\"));
+
+            return foundTsConfigs.ToArray();
         }
 
-        public IReadOnlyList<string> Locate()
+        private string GetProjectDirectory(string sourceFilePath)
         {
-            var foundFiles = new string[MaxNumberOfFiles];
+            var hr = vsSolution.GetProperty(VSPROPID_IsInOpenFolderMode, out var isOpenAsFolder);
```
duncanp-sonar, line 97: As discussed offline, this needs a comment that VS2015 will return E_NOTIMPL since open-as-folder isn't supported.

src/TypeScript/TsConfig/TsConfigsLocator.cs
```diff
@@ -21,47 +21,109 @@
...
+
+            Debug.Assert(hr == VSConstants.S_OK || hr == VSConstants.E_NOTIMPL, "Failed to retrieve VSPROPID_IsInOpenFolderMode");
+
+            if (isOpenAsFolder != null && (bool) isOpenAsFolder)
+            {
+                hr = vsSolution.GetProperty((int) __VSPROPID.VSPROPID_SolutionDirectory, out var solutionDirectory);
```
duncanp-sonar, line 103: Worth commenting that in open-as-folder there isn't a project, so we're using the solution dir instead.

src/TypeScript/TsConfig/TsConfigsLocator.cs
```diff
@@ -21,47 +21,109 @@
...
+                    "Failed to retrieve VSPROPID_SolutionDirectory");
+
+                return (string) solutionDirectory;
+            }
+
+            var hierarchyToSearch = vsHierarchyLocator.GetVsHierarchyForFile(sourceFilePath);
+
+            if (hierarchyToSearch == null)
+            {
+                logger.WriteLine(Resources.ERR_NoVsHierarchy, sourceFilePath);
```
duncanp-sonar, line 115: The wording of the error message hasn't changed in this PR, but I think it would be better to say "project" than "IVSHierarchy" in the message. Almost no-one will know what an IVSHierarchy is.

src/TypeScript/TsConfig/TsConfigsLocator.cs
```diff
@@ -21,47 +21,109 @@
...
+        public IReadOnlyList<string> Locate(string sourceFilePath)
+        {
+            var projectDirectory = GetProjectDirectory(sourceFilePath);
+
+            if (projectDirectory == null)
+            {
+                return Array.Empty<string>();
+            }
+
+            var foundTsConfigs = fileSystem
```
duncanp-sonar, line 87: Worth a comment that this behaviour matches the observed VS behaviour i.e. only the file location matters, not whether they are referenced in the project.

src/TypeScript/packages.lock.json
```diff
@@ -84,6 +104,16 @@
         "resolved": "7.10.6070",
         "contentHash": "p7xNzxIoZWDqSGlCi84u1CsNSL9YL7HWBdurfBUW8TZOsfW/0DS8/2MQcicQZav2Csq1f7Pep+VRdZJZkmQKJA=="
       },
+      "Microsoft.VisualStudio.SDK.EmbedInteropTypes": {
```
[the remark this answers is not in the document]
duncanp-sonar: Hhmm. The TS class uses the VS service provider to fetch the hierarchy locator, so we'd have to do a bit more. Pity.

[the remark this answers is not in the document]
rita-sonarsource: Yes I believe we could, because we could place the interface in Core instead of Infrastructure.VS.
I'll open a task.

duncanp-sonar (approved):
Just some minor comments about comments and logging.

PR #2439, opened 2021-05-20
Don't stop eslintbridge when a solution is opened

### LINKED ISSUE #2438: Race condition when opening a solution with an open TypeScript file
### Description

When a solution with open files is opened (i.e. there were files that were open when the solution was last closed, so VS will reopen them automatically), several independent actions are triggered:
A) an analysis of the open files will be triggered. This will launch the eslintbridge if necessary.
B) the active solution tracker event "ActiveSolutionChanged" event will fire, causing the eslintbridge to be shut down.

(B) often happens after (A) i.e.
1. opening a file launches a new analysis which starts eslintbridge
2. (B) then occurs, terminating the process

If (A) hasn't completed, then no issues will be reported and there will be an ugly call stack in the output window.

### Fix
We only need to terminate the running eslintbridge process when a solution is closed, not when it is opened.
The `ActiveSolutionChangedEventArgs` provide enough information to determine this.

duncanp-sonar:
Fixes #2438

### DIFF
--- a/src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
+++ b/src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
@@ -272,7 +272,9 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
         }
 
         [TestMethod]
-        public async Task OnSolutionChanged_NextAnalysisCallsInitLinter()
+        [DataRow(true, false)]
+        [DataRow(false, true)]
+        public async Task OnSolutionChanged_NextAnalysisCallsInitLinter(bool isSolutionOpen, bool shouldReinitLinter)
         {
             var activeSolutionTracker = SetupActiveSolutionTracker();
             var client = SetupEslintBridgeClient(new AnalysisResponse { Issues = Enumerable.Empty<Issue>() });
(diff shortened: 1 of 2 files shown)


src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
```diff
@@ -272,7 +272,9 @@ public void Dispose_UnregisterFromConfigChangedEvent()
         }
 
         [TestMethod]
-        public async Task OnSolutionChanged_NextAnalysisCallsInitLinter()
+        [DataRow(true, false)]
+        [DataRow(false, true)]
+        public async Task OnSolutionChanged_NextAnalysisCallsInitLinter(bool isSolutionOpen, bool shouldReinitLinter)
```
duncanp-sonar, line 277: I could have just used one variable since the expected behaviour is the reverse of "isSolutionOpen", but I think the version is more declarative.

src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
```diff
@@ -284,24 +286,28 @@ public async Task OnSolutionChanged_NextAnalysisCallsInitLinter()
...
-            client.Verify(x => x.InitLinter(It.IsAny<IEnumerable<Rule>>(), It.IsAny<CancellationToken>()), Times.Exactly(2));
+            var expectedCallCount = shouldReinitLinter ? 2 : 1;
+            client.Verify(x => x.InitLinter(It.IsAny<IEnumerable<Rule>>(), It.IsAny<CancellationToken>()), Times.Exactly(expectedCallCount));
         }
 
         [TestMethod]
-        public void OnSolutionChanged_StopsEslintBridgeClient()
+        [DataRow(true, false)]
+        [DataRow(false, true)]
+        public void OnSolutionChanged_StopsEslintBridgeClient(bool isSolutionOpen, bool shouldCallStop)
```
rita-sonarsource, line 300: The name of the test is no longer correct.
I think for readability, it would be better to split it into 2 tests:
 OnSolutionChanged_NoOpenSolution_StopsEslintBridgeClient
OnSolutionChanged_HasOpenSolution_EslintBridgeClientIsNotStopped (and it's worth aligning it with OnConfigChanged_DoesNotStopEslintBridgeClient)

src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
rita-sonarsource: Similar to the test below -- the name of this test is now misleading. It says "NextAnalysisCallsInitLinter" but it's not true.
I think it's worth splitting it to 2 tests, despite the copy-paste.

rita-sonarsource (approved):
Just a comment about the tests

rita-sonarsource:
@duncanp-sonar - perhaps we should create a new event, ActiveSolutionClosed, and listen just to that.

src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
duncanp-sonar: I'll align the test names. I think splitting the tests makes it harder to see the connection between them i.e. that test B is the exact inverse of test A. Using a data driven test makes the connection unmissable.

src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
duncanp-sonar: Having renamed the tests, it's much more obvious what the expected behaviour should be from the value of the "solution open" property, so I've dropped the second parameter.

duncanp-sonar:
> @duncanp-sonar - perhaps we should create a new event, ActiveSolutionClosed, and listen just to that.

Could do, but then we'd have to refactor all of the other code that listens to the current event. I wouldn't do that as part of a bug fix.

rita-sonarsource:
> > @duncanp-sonar - perhaps we should create a new event, ActiveSolutionClosed, and listen just to that.
> 
> Could do, but then we'd have to refactor all of the other code that listens to the current event. I wouldn't do that as part of a bug fix.

I meant adding a new one, leaving the other places unchanged. But it doesn't matter much.

duncanp-sonar:
Then we'd have two events, one that was raised on open and close, and one that is only raised on close. I think that just adds unnecessary complexity and makes things more confusing.

PR #2440, opened 2021-05-20
Tidy up eslintbridge logging

duncanp-sonar:
- added process id to all EsLintBridgeProcess messages
- removed debug messages from resource files

### DIFF
--- a/src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
+++ b/src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
@@ -55,13 +55,13 @@ namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
             var serializedRequest = request == null
                 ? string.Empty
                 : JsonConvert.SerializeObject(request, Formatting.Indented);
-            logger.LogDebug(Resources.INFO_RequestDetails, serverEndpoint, serializedRequest);
+            logger.LogDebug("[eslint-bridge] Endpoint: {0}, request:{1}{2}", serverEndpoint, Environment.NewLine, serializedRequest);
 
             var content = new StringContent(serializedRequest, Encoding.UTF8, "application/json");
             var response = await httpClient.PostAsync(serverEndpoint, content, cancellationToken);
             var responseString = await response.Content.ReadAsStringAsync();
 
-            logger.LogDebug(Resources.INFO_ResponseDetails, serverEndpoint, responseString);
+            logger.LogDebug("[eslint-bridge] Endpoint: {0}, response:{1}{2}", serverEndpoint, Environment.NewLine, responseString);
 
             return responseString;
         }
--- a/src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
+++ b/src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
@@ -56,6 +56,9 @@ namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient
 
         internal Process Process;
 
+        private const int NO_PROCESS = 0;
+        int processId = NO_PROCESS;
+
         internal EslintBridgeProcess(string eslintBridgeStartupScriptPath,
             INodeLocator nodeLocator,
             ILogger logger)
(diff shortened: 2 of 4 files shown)


duncanp-sonar:
I was finding the eslintbridge log messages confusing because it wasn't possible to tell the difference between the JS and TS instances. I've tweaked the messages so they are in a standard format and all include the process id.

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
```diff
@@ -55,13 +55,13 @@ public async Task<string> PostAsync(Uri serverEndpoint, object request, Cancella
             var serializedRequest = request == null
                 ? string.Empty
                 : JsonConvert.SerializeObject(request, Formatting.Indented);
-            logger.LogDebug(Resources.INFO_RequestDetails, serverEndpoint, serializedRequest);
```
duncanp-sonar, line 58: There were a number of messages that were previously INFO that have been downgraded to DEBUG.
For consistency with the other debug messages, I've removed those downgraded messages from the resource file.

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -137,8 +140,11 @@ private void StartServer()
             Process.OutputDataReceived += OnOutputDataReceived;
 
             Process.Start();
-            logger.WriteLine(Resources.INFO_ServerProcessId, Process.Id);
-            logger.LogDebug(Resources.INFO_ServerHasExisted, Process.HasExited);
+            // Storing the process so we don't have to worry about checking whether Process is null when logging,
+            // and so that if any messages are logged after Process is set to null we know which process it related to.
+            processId = Process.Id;
+            logger.WriteLine(Resources.INFO_ServerProcessId, processId);
+            logger.LogDebug($"[eslint-bridge] [process id: {processId}] Server process HasExited: {1}.", processId, Process.HasExited);
```
duncanp-sonar, line 147: All of the messages start (debug and non-debug) now show the process id at the start of the message in this format:
```[prefix] [process id: {processId}] Message```

Feel free to suggest an alternative.

I considered omitting the `process id: ` label, but left it in so there was would be no confusion about whether it was logging the port number or the process id.

I also considered logging the port number, but we won't have it until some time after the process has started and we've received the output stream message from the node process.

File: src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs [hunk 3]
```diff
@@ -152,7 +158,8 @@ private void StartServer()
 
         private void Process_Exited(object sender, EventArgs e)
         {
-            Process = null;
+            logger.LogDebug($"[eslint-bridge] [process id: {processId}] Process exited.", processId);
```
duncanp-sonar, line 161: New debug message for completeness.

rita-sonarsource:
> I was finding the eslintbridge log messages confusing because it wasn't possible to tell the difference between the JS and TS instances. I've tweaked the messages so they are in a standard format and all include the process id.

Hmm, I think it would still be difficult to know which is which, but I don't know how implement the distinction in a pretty way. 
I was thinking about the prefixes that we add to our logging, i.e. [eslint-bridge]. Perhaps we should have a better logging factory, that would allow to define a source. Something to keep in mind when work on the logging. Just a general thought.

rita-sonarsource (approved):

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -152,7 +158,8 @@ private void StartServer()
...
+            ClearProcessData();
```
rita-sonarsource, line 162: I wonder if we have (and always had) a race condition here.
It's not protected in a lock, so it depends on when this event is called.
i.e.
1) Thread A: Process A has crashed, this event is raised.
2) Thread B: Calls EslintBridgeProcess.Start, gets to `Process.HasExited=true`
3) Thread B: Starts a new process and calls `Process = new Process { StartInfo = psi }`
4) Thread A: sets `Process = null`

duncanp-sonar: Yep, looks possible. I'm not sure what the consequences of that series of events would be :
* the external node process would still be running, so the the node process would still be up and would be able to reply to ThreadB.
* the _next_ request to analyze would then start another instance of the node server, since Process = null.
So we'd have an orphaned node process that is still running?

rita-sonarsource: Yep. I think that's what would happen. It could explain why I saw two "started" logs sometimes, even though I didn't open a JS file, only TS.

PR #2445, opened 2021-05-21
Process projectReferences in tsconfig.json files

### LINKED ISSUE #2441: Process projectReferences in tsconfig.json files
#### Description

`tsconfig.json` can extend another config using the `extends` keyword, e.g.:

tsconfig.json A:
```
{
 "references": [{ 
     "path": "tsconfig.app.json"    
  }]
}
```
tsconfig.json B:
```
 "compilerOptions": {
    "module": "commonjs",
    "target": "ES2020",
    "lib": [ "ES2020" ],
    "sourceMap": true,
    "allowUnreachableCode": true
  },
  "exclude": [
    "node_modules",
  ],
  "include": []
}
```

`eslint-bridge` returns the extended tsconfigs via the `projectReferences` property.

TODO:
- [x] take `projectReferences` into account

rita-sonarsource:
Fixes #2441

### DIFF
--- a/src/TypeScript.UnitTests/TsConfig/TsConfigProviderTests.cs
+++ b/src/TypeScript.UnitTests/TsConfig/TsConfigProviderTests.cs
@@ -181,12 +181,13 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.TsConfig
             var tsConfigsLocator = SetupTsConfigsLocator(testedFileName, tsConfigsInSolution);
 
             var eslintBridgeClient = new Mock<ITypeScriptEslintBridgeClient>();
+            var parsingError = new ParsingError { Code = ParsingErrorCode.UNSUPPORTED_TYPESCRIPT, Message = "some message", Line = 5555 };
             SetupEslintBridgeResponse(eslintBridgeClient, new Dictionary<string, TSConfigResponse>
             {
                 {
                     "config1", new TSConfigResponse
                     {
-                        ParsingError = new ParsingError{Code = ParsingErrorCode.UNSUPPORTED_TYPESCRIPT, Message = "some message", Line = 55},
+                        ParsingError = parsingError,
                         Files = new[] {testedFileName} // should be ignored
                     }
                 },
@@ -202,8 +203,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.TsConfig
             eslintBridgeClient.Verify(x => x.TsConfigFiles("config2", CancellationToken.None), Times.Once);
             eslintBridgeClient.VerifyNoOtherCalls();
 
-            logger.AssertPartialOutputStringExists("some message");
-            logger.AssertPartialOutputStringExists(ParsingErrorCode.UNSUPPORTED_TYPESCRIPT.ToString());
+            logger.AssertPartialOutputStringExists(parsingError.Message, parsingError.Code.ToString(), parsingError.Line.ToString());
         }
 
         [TestMethod]
--- a/src/TypeScript.UnitTests/TsConfig/TsConfigProviderTests.cs
+++ b/src/TypeScript.UnitTests/TsConfig/TsConfigProviderTests.cs
@@ -141,6 +141,133 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.TsConfig
             eslintBridgeClient.VerifyNoOtherCalls();
         }
 
+        [TestMethod]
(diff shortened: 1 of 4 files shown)


src/TypeScript/TsConfig/TsConfigProvider.cs
```diff
@@ -61,16 +62,69 @@ public async Task<string> GetConfigForFile(string sourceFilePath, CancellationTo
...
+
+                if (response.ProjectReferences == null || !response.ProjectReferences.Any())
+                {
+                    continue;
+                }
+
+                logger.LogDebug(Resources.INFO_CheckingReferencedTsConfigs, tsConfigFilePath, string.Join(Path.DirectorySeparatorChar.ToString(), response.ProjectReferences));
+
+                var result = await GetConfigForFile(sourceFilePath,
+                    response.ProjectReferences,
```
rita-sonarsource, line 121: On the java side, they first gather all the tsconfigs and their references, and only then iterate over them to find one that contains the analyzed file.
Implementing it this way would help with #2446, but it would also hurt the performance. In this PR everything is evaluated lazily. I'm a bit reluctant to take the java approach as I've seen eslint-bridge take 4+ seconds to return a result for a single tsconfig.
For #2446, perhaps we can start by sorting and giving priority to the closest tsconfig found in the project's directory, ignoring the project-references. Perhaps the tsconfig found in the project directory be closer than the tsconfigs mentioned in projectReferences, but I don't know what are the real-world scenarios. WDYT?

duncanp-sonar: The Java code is in the analyzer sensor, which is optimised for the batch processing scanning performed by the scanner. The other SL flavours don't have any opportunity to change the approach, since they are emulating the scanner API.
It doesn't make sense for us to use the same approach for single file analysis.

I don't know enough about the real-world scenarios either.

Scenario:
* tsconfig **A** is closest to the file
* **A** contains settings but doesn't contain any includes
* **A** referencs tsconfig **B**
* **B** includes the file.

Which tsconfig should we return? **A**, since it's closer and indirectly references the file, or **B** which directly references the file?

rita-sonarsource: I honestly don't know. I'd say that this is a very weird project structure 😆

rita-sonarsource:

src/TypeScript/TsConfig/TsConfigProvider.cs
```diff
@@ -61,16 +62,67 @@ public async Task<string> GetConfigForFile(string sourceFilePath, CancellationTo
...
+                allTsConfigsFilePaths,
+                checkedTsConfigs: new List<string>(),
+                cancellationToken);
+
+            return tsConfigFile;
+        }
+
+        private async Task<string> GetConfigForFile(string sourceFilePath,
+            IEnumerable<string> candidateTsConfigs,
+            ICollection<string> checkedTsConfigs,
```
duncanp-sonar, line 75: `visited` or `alreadyChecked` might be slightly clearer than `checkedTsConfigs`

src/TypeScript.UnitTests/TsConfig/TsConfigProviderTests.cs
```diff
@@ -141,6 +141,163 @@ public async Task GetConfigForFile_TsConfigContainsFileWithIllegalCharacters_Fil
...
+            var testSubject = CreateTestSubject(tsConfigsLocator.Object, eslintBridgeClient.Object, logger);
+            var result = await testSubject.GetConfigForFile(testedFileName, CancellationToken.None);
+            result.Should().BeNull();
+
+            eslintBridgeClient.Verify(x => x.TsConfigFiles("config1", CancellationToken.None), Times.Once);
+            eslintBridgeClient.Verify(x => x.TsConfigFiles("config2", CancellationToken.None), Times.Once);
+            eslintBridgeClient.VerifyNoOtherCalls();
+
+            logger.AssertPartialOutputStringExists("some message");
+            logger.AssertPartialOutputStringExists(ParsingErrorCode.UNSUPPORTED_TYPESCRIPT.ToString());
```
duncanp-sonar, line 206: `AssertPartialOutputStringExists` takes a paramarray of strings, so you could check the message, code and line are all logged in a single assertion.
That would also check that they occur in the _same_ message.

duncanp-sonar (approved):

PR #2447, opened 2021-05-21
Fix for potential race condition terminating eslintbridge

duncanp-sonar:


src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -144,7 +144,7 @@ private void StartServer()
             // and so that if any messages are logged after Process is set to null we know which process it related to.
             processId = Process.Id;
             logger.WriteLine(Resources.INFO_ServerProcessId, processId);
-            logger.LogDebug($"[eslint-bridge] [process id: {processId}] Server process HasExited: {1}.", processId, Process.HasExited);
```
duncanp-sonar, line 147: Miscellaneous: I messed up converting from the resource string format to the interpolated string format. The method was logging "HasExited: 1".

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -158,8 +158,18 @@ private void StartServer()
 
         private void Process_Exited(object sender, EventArgs e)
         {
-            logger.LogDebug($"[eslint-bridge] [process id: {processId}] Process exited.", processId);
-            ClearProcessData();
+            // Guard against a potential race condition i.e. this event being received
```
duncanp-sonar, line 161: Simple fix for the race condition you identified. There may still be other issues, but this is an improvement.

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -158,8 +158,18 @@ private void StartServer()
...
+            // after an new analysis request has already created a new Process
+            if(sender is Process terminatedProcess)
+            {
+                logger.LogDebug($"[eslint-bridge] [process id: {terminatedProcess.Id}] Process exited.");
+                terminatedProcess.Exited -= Process_Exited;
```
duncanp-sonar, line 166: This is the only place we unregister the event handler. In theory that should be enough.
The other code paths that call `ClearProcessData` will have disposed the old process so the event registration will have be removed.

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -158,8 +158,18 @@ private void StartServer()
...
+
+                if (ReferenceEquals(terminatedProcess, Process))
+                {
+                    ClearProcessData();
```
rita-sonarsource, line 170: Don't we still have a race condition here? Between referenceEquals and ClearProcessData?

I was thinking about it more -- do we really care about calling ClearProcessData in this case? The "Start" function checks `Process.HasExited`, which should evaluate to true. So perhaps we should just log a message here and do nothing. WDYT?

duncanp-sonar: Yes, there is still a race condition, but it's less likely to occur.

Perhaps we can get away with doing nothing. The `portId` won't be reset, but then that shouldn't matter since we shouldn't be logging anything else until a new process has started.

I'll remove the call to `ClearProcessData`.

rita-sonarsource (approved):

PR #2458, opened 2021-05-25
Support directory path in project references

### LINKED ISSUE #2456: Support specifying directory path in projectReferences
#### Description

According to [project references documentation](https://www.typescriptlang.org/docs/handbook/project-references.html), it's possible to specify a directory path under project references ("The path property of each reference can point to a directory containing a tsconfig.json file").

TODO:
- [ ] Support directory path responses from eslint-bridge

rita-sonarsource:
Fixes #2456

### DIFF
--- a/src/TypeScript.UnitTests/TsConfig/TsConfigProviderTests.cs
+++ b/src/TypeScript.UnitTests/TsConfig/TsConfigProviderTests.cs
@@ -20,6 +20,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO.Abstractions;
 using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
@@ -66,24 +67,24 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.TsConfig
         public async Task GetConfigForFile_SourceFileIsNotInAnyTsConfig_Null()
         {
             const string testedFileName = "some file";
-            var tsConfigsInSolution = new[] { "config1.json", "config2.json", "config3.json" };
+            var tsConfigsInSolution = new[] { "config1", "config2", "config3" };
             var tsConfigsLocator = SetupTsConfigsLocator(testedFileName, tsConfigsInSolution);
 
             var eslintBridgeClient = new Mock<ITypeScriptEslintBridgeClient>();
             SetupEslintBridgeResponse(eslintBridgeClient, new Dictionary<string, TSConfigResponse>
             {
-                {"config1.json", new TSConfigResponse()},
-                {"config2.json", new TSConfigResponse{Files = new List<string>{"some other file"}}},
-                {"config3.json", new TSConfigResponse{Files = new List<string>{"some file2"}}},
+                {"config1", new TSConfigResponse()},
+                {"config2", new TSConfigResponse{Files = new List<string>{"some other file"}}},
+                {"config3", new TSConfigResponse{Files = new List<string>{"some file2"}}},
             });
 
             var testSubject = CreateTestSubject(tsConfigsLocator.Object, eslintBridgeClient.Object);
             var result = await testSubject.GetConfigForFile(testedFileName, CancellationToken.None);
             result.Should().BeNull();
 
-            eslintBridgeClient.Verify(x => x.TsConfigFiles("config1.json", CancellationToken.None), Times.Once);
-            eslintBridgeClient.Verify(x => x.TsConfigFiles("config2.json", CancellationToken.None), Times.Once);
-            eslintBridgeClient.Verify(x => x.TsConfigFiles("config3.json", CancellationToken.None), Times.Once);
+            eslintBridgeClient.Verify(x => x.TsConfigFiles("config1", CancellationToken.None), Times.Once);
+            eslintBridgeClient.Verify(x => x.TsConfigFiles("config2", CancellationToken.None), Times.Once);
+            eslintBridgeClient.Verify(x => x.TsConfigFiles("config3", CancellationToken.None), Times.Once);
             eslintBridgeClient.VerifyNoOtherCalls();
         }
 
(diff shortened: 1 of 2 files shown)


src/TypeScript/TsConfig/TsConfigProvider.cs
```diff
@@ -77,25 +77,33 @@ public async Task<string> GetConfigForFile(string sourceFilePath, CancellationTo
         {
             foreach (var tsConfigFilePath in candidateTsConfigs)
             {
-                if (visited.Contains(tsConfigFilePath))
+                var tsConfigToCheck = tsConfigFilePath;
+                var isDirectory = !Path.HasExtension(tsConfigToCheck);
+
+                if (isDirectory)
+                {
+                    tsConfigToCheck = Path.GetFullPath(Path.Combine(tsConfigToCheck, "tsconfig.json"));
```
rita-sonarsource, line 85: The documentation was unclear about whether the tsconfig has to be under the specified directory or if it can be in a sub folder -- but VS fails if it's not directly under, so that's what I went with as well.

src/TypeScript.UnitTests/TsConfig/TsConfigProviderTests.cs
```diff
@@ -66,24 +66,24 @@ public async Task GetConfigForFile_NoTsConfigsInSolution_Null()
         public async Task GetConfigForFile_SourceFileIsNotInAnyTsConfig_Null()
         {
             const string testedFileName = "some file";
-            var tsConfigsInSolution = new[] { "config1", "config2", "config3" };
+            var tsConfigsInSolution = new[] { "config1.json", "config2.json", "config3.json" };
```
rita-sonarsource, line 69: I had to add an extension to all the existing tests, otherwise they were considered as a folder due to the new logic.

src/TypeScript.UnitTests/TsConfig/TsConfigProviderTests.cs
```diff
@@ -135,39 +135,36 @@ public async Task GetConfigForFile_TsConfigContainsFileWithIllegalCharacters_Fil
 
             var testSubject = CreateTestSubject(tsConfigsLocator.Object, eslintBridgeClient.Object);
             var result = await testSubject.GetConfigForFile(testedFileName, CancellationToken.None);
-            result.Should().Be("config");
-
-            eslintBridgeClient.Verify(x => x.TsConfigFiles("config", CancellationToken.None), Times.Once);
```
rita-sonarsource, line 140: Unrelated: this assertion was true but irrelevant so I removed it to make the test clearer what it was testing.

src/TypeScript.UnitTests/TsConfig/TsConfigProviderTests.cs
```diff
@@ -210,94 +207,121 @@ public async Task GetConfigForFile_FailedToProcessTsConfig_ParsingError_TsConfig
...
-            eslintBridgeClient.Verify(x => x.TsConfigFiles("config2", CancellationToken.None), Times.Once);
-            eslintBridgeClient.Verify(x => x.TsConfigFiles("config3", CancellationToken.None), Times.Once);
+            eslintBridgeClient.Verify(x => x.TsConfigFiles("config1.json", CancellationToken.None), Times.Once);
+            eslintBridgeClient.Verify(x => x.TsConfigFiles("config2.json", CancellationToken.None), Times.Once);
+            eslintBridgeClient.Verify(x => x.TsConfigFiles("config3.json", CancellationToken.None), Times.Once);
             eslintBridgeClient.VerifyNoOtherCalls();
         }
 
+        [TestMethod]
+        public async Task GetConfigForFile_TsConfigHasFolderProjectReference_ReferenceProcessedWithDefaultTsConfigName()
```
rita-sonarsource, line 299: New test for the folder check

src/TypeScript/TsConfig/TsConfigProvider.cs
```diff
@@ -77,25 +77,33 @@ public async Task<string> GetConfigForFile(string sourceFilePath, CancellationTo
         {
             foreach (var tsConfigFilePath in candidateTsConfigs)
             {
-                if (visited.Contains(tsConfigFilePath))
+                var tsConfigToCheck = tsConfigFilePath;
+                var isDirectory = !Path.HasExtension(tsConfigToCheck);
```
duncanp-sonar, line 81: Paths can contain full stops, in which case `Path.HasExtension` will return the wrong result.

`Directory.Exists` returns the correct result.

src/TypeScript/TsConfig/TsConfigProvider.cs
```diff
@@ -77,25 +77,33 @@ public async Task<string> GetConfigForFile(string sourceFilePath, CancellationTo
...
+
+                if (isDirectory)
+                {
+                    tsConfigToCheck = Path.GetFullPath(Path.Combine(tsConfigToCheck, "tsconfig.json"));
+                }
+
+                if (visited.Contains(tsConfigToCheck))
```
duncanp-sonar, line 88: `Contains` does a case-sensitive comparison.
It would be better to use a hashset created with an appropriate comparer.

duncanp-sonar (commented):
A couple of minor changes are needed I think.

src/TypeScript/TsConfig/TsConfigProvider.cs
rita-sonarsource: Ah, you're right. I should've used `Path.HasExtension(Path.GetFullPath(tsConfigToCheck))`, because then a directory would've always ended with "\\". 
`Directory.Exists` -- it doesn't have to exist, e.g. if you input a wrong directory name in the external tsconfig.json. The ts compiler will complain, but it doesn't stop us from running the analysis and failing with:
`Failed to process 'C:/Users/Rita/source/repos/NodejsConsoleApp3/NewFolder1/a': Cannot read file 'C:/Users/Rita/source/repos/NodejsConsoleApp3/NewFolder1/a'.`
Overall I think it doesn't matter, I'll change it to directory.

rita-sonarsource:
> A couple of minor changes are needed I think.

Fixed

src/TypeScript/TsConfig/TsConfigProvider.cs
duncanp-sonar: Hhmm - what about relative paths? Using Path.HasExtension with a canonical path might work better than Directory.Exists in that case.

rita-sonarsource: Eslint-bridge returns us full file paths, even if the tsconfig contains a relative path. So we're ok with that.

rita-sonarsource: Overall I think if the user input something wrong (i.e. non existing directory, or a directory that doesn't have a tsconfig.json), the analysis will likely fail either way, and so would the build itself. So as long as we correctly support directories in general and don't fail with an ugly exception, I think it should be ok.

duncanp-sonar: It's probably worth adding an assertion that the paths returned from eslintbridge are rooted i.e. to document and validate our assumptions about the behaviour of an external system.

duncanp-sonar (approved):
I think it's worth adding an assertion about the behaviour of eslintbridge; otherwise looks good.

PR #2480, opened 2021-07-07
Don't start EslintBridgeServer when VS is shutting down

### LINKED ISSUE #2469: Closing down VS starts two node processes which are then orphaned and left running
### Description

New node.js processes for TypeScript and JavaScript are started when VS is shut down. They are then orphaned (since VS has closed), so they remain running until the ended by the user / logoff / machine restart.

### Repro steps

1. Open a VS solution

2. Close VS

Note: it is not necessary to open/analyse any JS/TS files.

Related community thread: https://community.sonarsource.com/t/sonarlint-for-visual-studio-leaves-behind-two-orphaned-node-js-processes-on-exit/45384

The problem repros intermittently for me locally, and reliably for the reporting user (?timing issue?).

### Expected behaviour

Node processes are shut down.

### Actual behaviour

Orphaned processes are left running.

### Known workarounds

None

### Additional information

The code that handles solution close events causes the EslintBridgeServer to send a command to the server telling it to stop. However, if the server isn't currently running, it will be started first, then told to stop immediately.

The call stack and output window below show what happens when a solution is closed (**not** what happens when VS is closed):

##### Call stack when closing a solution (no TS/JS file had been opened -> no EslintBridge processes running)
image

##### Output window
image

Closing the solution will start a new process if one is not running, and then close it immediately.
It may be the case that in the event of VS shutdown, the node processes are started, but VS terminates before they are told to shut down again.

### Related information

* SonarLint for VS Version: v4,35, possibly also v.4.34
* Languages in your solution: n/a

rita-sonarsource:
Fixes #2469

### DIFF
--- a/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
@@ -136,6 +136,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
         }
 
         [TestMethod]
+        [Description("Regression test for #2469")]
         public void Dispose_EslintBridgeProcessIsNotRunning_DisposesHttpWrapper()
         {
             var eslintBridgeProcess = SetupServerProcess(isRunning: false);
--- a/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
@@ -136,11 +136,25 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
         }
 
         [TestMethod]
-        public void Dispose_DisposesHttpWrapper()
+        public void Dispose_EslintBridgeProcessIsNotRunning_DisposesHttpWrapper()
         {
+            var eslintBridgeProcess = SetupServerProcess(isRunning: false);
             var httpWrapper = new Mock<IEslintBridgeHttpWrapper>();
 
-            var testSubject = CreateTestSubject(httpWrapper.Object);
+            var testSubject = CreateTestSubject(httpWrapper.Object, eslintBridgeProcess: eslintBridgeProcess.Object);
+            testSubject.Dispose();
+
+            httpWrapper.Verify(x => x.Dispose(), Times.Once);
+            httpWrapper.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public void Dispose_EslintBridgeProcessIsRunning_ClosesAndDisposesHttpWrapper()
+        {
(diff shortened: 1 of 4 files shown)


src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -114,13 +114,19 @@ public async Task Close()
         {
             try
             {
-                await MakeCall("close", null, CancellationToken.None);
+                if (eslintBridgeProcess.IsRunning)
```
rita-sonarsource, line 117: This is not wrapped in a lock, but I can't think of a scenario in which thread A will be in the process of of calling "close" and thread B will already get to "eslintBridgeProcess.Stop()" -- there is only one instance of this class per analyzer, and the analyzer calls this method only when solution closes / vs shuts down. Both operations are very slow so I don't think a lock is needed. Let me know if you can think of a use case.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
```diff
@@ -136,11 +136,25 @@ public void Analyze_CriticalException_ExceptionNotCaught()
         }
 
         [TestMethod]
-        public void Dispose_DisposesHttpWrapper()
+        public void Dispose_EslintBridgeProcessIsNotRunning_DisposesHttpWrapper()
```
duncanp-sonar, line 139: Could add a comment saying this is a regression test for https://github.com/SonarSource/sonarlint-visualstudio/issues/2469

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
```diff
@@ -252,6 +251,96 @@ public async Task Dispose_ProcessAlreadyKilled_NoException()
...
+
+            testSubject.IsRunning.Should().BeTrue();
+
+            testSubject.Dispose();
+
+            testSubject.IsRunning.Should().BeFalse();
+        }
+
+        [TestMethod]
+        public async Task IsRunning_ServerWasRestarted_True()
```
duncanp-sonar, line 314: This test checks everything that is tested in the two previous tests, so the other tests don't add much value. It might be clearer to have one test appropriately named and commented.

duncanp-sonar (approved):

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
rita-sonarsource: I removed IsRunning_ServerWasStartedThenDisposed_False and IsRunning_ServerWasStarted_True.
The other two (IsRunning_ServerWasNeverStarted_False and IsRunning_ServerWasNeverStartedAndWasDisposed_False) test a bit different things.

PR #2481, opened 2021-07-08
Raise ActiveDocumentChanged event when transitioning from toolwindow to document

### LINKED ISSUE #2242: Preamble not refreshing on change of focus
To reproduce: change the focus from any C++ file to the output windows. Then change the focus from the output window back to the file.
  
Tested working version: 4.28
The preamble is refreshed. The message is printed in the SonarLint output windows.
Output: 
image

Tested buggy version 4.34
Doing the same steps doesn't lead to any message in the SonarLint output windows. The preamble is no refreshed.

rita-sonarsource:
Fixes #2242


rita-sonarsource:
@duncanp-sonar as we discussed offline: the first commit is removing the TestInitialize from the tests.

duncanp-sonar (approved):

PR #2483, opened 2021-07-10
Remove noisy debug log in EslintBridgeProcess

rita-sonarsource:
Caused by #2480


src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -211,15 +211,13 @@ private void TerminateRunningProcess()
             {
                 if (Process == null || Process.HasExited)
                 {
-                    logger.LogDebug($"[eslint-bridge] [process id: {processId}] Cannot terminate eslint-bridge process: no process is running");
```
rita-sonarsource, line 214: TerminateRunningProcess is called when the class is Disposed, which happens whenever a solution is closed. 
If no JS/TS file was analyzed, this would still be printed. It's redundant so I've removed it.

PR #2486, opened 2021-07-10
Fix LF line hash calculations

### LINKED ISSUE #2276: CFamily analysis fails when file has mixed line endings (CRLF and LF)
### Description

When a CFamily file (C, C++, .h) contains mixed line endings (CRLF and LF), the analysis fails:

```
Failed to analyze C:\Users\michal\source\repos\Project3\Project3\Source1.cpp, exception details: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.VisualStudio.Text.SnapshotSpan.TranslateTo(ITextSnapshot targetSnapshot, SpanTrackingMode spanTrackingMode)
   at SonarLint.VisualStudio.Integration.Vsix.TextBufferIssueTracker.<>c__DisplayClass22_0.<TranslateSpans>b__1(IAnalysisIssueVisualization x)
   at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at SonarLint.VisualStudio.Integration.Vsix.TextBufferIssueTracker.TranslateSpans(IEnumerable`1 issues, ITextSnapshot activeSnapshot)
   at SonarLint.VisualStudio.Integration.Vsix.TextBufferIssueTracker.HandleNewIssues(IEnumerable`1 issues)
   at SonarLint.VisualStudio.Integration.Vsix.AccumulatingIssueConsumer.Accept(String path, IEnumerable`1 issues)
   at SonarLint.VisualStudio.Integration.Vsix.CFamily.CLangAnalyzer.HandleMessage(Message message, Request request, IIssueConsumer consumer, Int32& issueCount)
   at SonarLint.VisualStudio.Integration.Vsix.CFamily.CLangAnalyzer.<>c__DisplayClass14_0.<TriggerAnalysisAsync>b__1(Message message)
   at SonarLint.VisualStudio.Integration.Vsix.CFamily.Protocol.Read(BinaryReader reader, Action`1 handleIssue, String issueFilePath)
   at SonarLint.VisualStudio.Integration.Vsix.CFamily.CFamilyHelper.<>c__DisplayClass23_0.<CallClangAnalyzer>b__1(StreamReader reader)
   at SonarLint.VisualStudio.Integration.Vsix.CFamily.ProcessRunner.Execute(ProcessRunnerArguments runnerArgs)
   at SonarLint.VisualStudio.Integration.Vsix.CFamily.CFamilyHelper.CallClangAnalyzer(Action`1 handleMessage, Request request, IProcessRunner runner, ILogger logger, CancellationToken cancellationToken)
   at SonarLint.VisualStudio.Integration.Vsix.CFamily.CLangAnalyzer.CallSubProcess(Action`1 handleMessage, Request request, ISonarLintSettings settings, ILogger logger, CancellationToken cancellationToken)
   at SonarLint.VisualStudio.Integration.Vsix.CFamily.CLangAnalyzer.<TriggerAnalysisAsync>d__14.MoveNext()
ERROR: LLVM ERROR: IO failure on output stream: invalid argument
```

#### Workaround
Fixing the line endings to CRLF fixes the problem.

#### Repro project
[Project3 (1).zip](https://github.com/SonarSource/sonarlint-visualstudio/files/6237077/Project3.1.zip)

#### Notes
There are 2 separate problems that cause this issue:
1. [LineHashCalculator](https://github.com/SonarSource/sonarlint-visualstudio/blob/6b9a8a7623906def168229d9089bf6e17a9c00a4/src/IssueViz/Editor/LineHashCalculator.cs#L76) splits lines according to CRLF only, which causes the hash to be incorrect if the hashed line is followed by a LF end line. [IIssueSpanCalculator](https://github.com/SonarSource/sonarlint-visualstudio/blob/6b9a8a7623906def168229d9089bf6e17a9c00a4/src/IssueViz/Editor/IIssueSpanCalculator.cs#L68) then returns an empty span due to the hashes being different (as VS is able to split the lines correctly).
2. [TextBufferIssueTracker](https://github.com/SonarSource/sonarlint-visualstudio/blob/f1a2d46fd631512efa19797139872e821f05226e/src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs#L122) doesn't take into account that the returned span could be empty, which causes the reported exception.

TODO:
- [ ] Fix TextBufferIssueTracker to ignore empty spans
- [ ] Fix LineHashCalculator to split the lines accordingly

rita-sonarsource:
Fixes #2276

### DIFF
--- a/src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
@@ -105,7 +105,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
 
             return filePaths.ToDictionary(x => x,
                 path => fileSystem.File.Exists(path)
-                    ? textDocumentFactoryService.CreateAndLoadTextDocument(path, filesContentType)
+                    ? textDocumentFactoryService.CreateAndLoadTextDocument(path, filesContentType) // the document is being loaded from disc, so it should match the version that was analyzed by the subprocess
                     : null);
         }
 
@@ -143,9 +143,14 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
                                       cFamilyIssueLocation.EndColumn == 0 &&
                                       cFamilyIssueLocation.EndLine == 0;
 
-            return isFileLevelLocation
-                ? null
-                : lineHashCalculator.Calculate(fileContents[cFamilyIssueLocation.Filename], cFamilyIssueLocation.Line);
+            if (isFileLevelLocation)
+            {
+                return null;
+            }
+
+            var textSnapshot = fileContents[cFamilyIssueLocation.Filename]?.TextBuffer?.CurrentSnapshot;
+
+            return textSnapshot == null ? null : lineHashCalculator.Calculate(textSnapshot, cFamilyIssueLocation.Line);
         }
 
         private AnalysisIssueLocation ToAnalysisIssueLocation(MessagePart cFamilyIssueLocation, IDictionary<string, ITextDocument> fileContents)
--- a/src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
@@ -25,6 +25,8 @@ using System.Diagnostics;
 using System.IO;
 using System.IO.Abstractions;
 using System.Linq;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Utilities;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Analysis;
 using SonarLint.VisualStudio.Core.CFamily;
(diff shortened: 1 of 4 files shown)


rita-sonarsource:
@duncanp-sonar please review. First commit is getting rid of TestInitialize in one of the test files.

src/IssueViz/Editor/LineHashCalculator.cs
```diff
@@ -52,27 +48,25 @@ internal LineHashCalculator(IChecksumCalculator checksumCalculator)
...
         {
-            if (string.IsNullOrEmpty(documentText) || oneBasedLineNumber < 1)
+            if (oneBasedLineNumber < 1 || documentText?.TextBuffer?.CurrentSnapshot == null)
             {
                 return null;
             }
 
-            var lines = GetLines(documentText);
+            // SonarLint issues line numbers are 1-based, span lines are 0-based
+            var lineToHash = documentText.TextBuffer.CurrentSnapshot.GetLineFromLineNumber(oneBasedLineNumber - 1);
```
duncanp-sonar, line 59: Might be worth commenting that we're letting the text document find the line as it is aware of different types of line break.

src/IssueViz/Editor/LineHashCalculator.cs
```diff
@@ -32,16 +31,13 @@ public interface ILineHashCalculator
         /// <param name="documentText">The text from which to extract the line</param>
         /// <param name="oneBasedLineNumber">1-based line to hash</param>
         /// <returns>hash of line</returns>
-        string Calculate(string documentText, int oneBasedLineNumber);
+        string Calculate(ITextDocument documentText, int oneBasedLineNumber);
```
duncanp-sonar, line 34: The line hash calculator shouldn't be responsible for choosing which snapshot to use. It would be better if the snapshot was supplied by the caller.

src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
```diff
@@ -77,32 +86,35 @@ public IAnalysisIssue Convert(Message cFamilyIssue, string sqLanguage, ICFamilyR
...
+        private IDictionary<string, ITextDocument> LoadFileContentsOfReportedFiles(Message cFamilyIssue)
         {
             var filePaths = cFamilyIssue.Parts
                 .Select(x => x.Filename)
                 .Union(new[] {cFamilyIssue.Filename})
                 .Distinct();
 
             return filePaths.ToDictionary(x => x,
-                path => fileSystem.File.Exists(path) ? fileSystem.File.ReadAllText(path) : null);
+                path => fileSystem.File.Exists(path)
```
duncanp-sonar, line 107: It doesn't seem right that we are checking both the file system and the in-memory representation of the file. If we're switching to look at the in-memory representation then we should look at that for everything.

duncanp-sonar (commented):
There are a couple of bits I think are worth changing now, the main one being removing the dependence on the file system since we're now relying on the in-memory representation of the file.

There are a few more things wrong with the way we handle the problem of translating between the file representation and the in-memory representation which are out of scope for this bug fix.

For example, using `TextBuffer.CurrentSnapshot` when translating from file-to-snapshot at this low level isn't correct as current snapshot could have changed.
Instead, we should store the current snapshot at the point the analysis is triggered (the "analysis snapshot"), since that should corresponding to the version that is being analysed. The issue converter should use this analysis snapshot to translate the issue locations, and then we should translate the locations from the analysis snapshot to the current snapshot when creating the tags.
The `TextBufferIssueTracker` and `AccumulatingIssueConsumer` do handle some of this. Unfortunately, the "analysis snapshot" is not passed to the analyzer. Fixing this won't be straightforward.

src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
duncanp-sonar: Ok, that helps. I'll continue reviewing then. Ignore the other long comment about file->in-memory translations.

src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
```diff
@@ -77,32 +86,35 @@ public IAnalysisIssue Convert(Message cFamilyIssue, string sqLanguage, ICFamilyR
...
+                    ? textDocumentFactoryService.CreateAndLoadTextDocument(path, filesContentType)
+                    : null);
```
duncanp-sonar, line 109: Worth calling out that the document is being loaded from disc, so it should match the version that was analysed by the subprocess.

duncanp-sonar (approved):

PR #2487, opened 2021-07-12
Fix hotspot help link in the Editor

### LINKED ISSUE #2411: [Hotspots] Documentation rule link is broken for some rules
#### Description

Opening a hotspot for some rules (i.e. [S4823](https://rules.sonarsource.com/csharp/RSPEC-4823)) leads to a broken link (`https://rules.sonarsource.com/csharpsquid/RSPEC-4823` instead of `https://rules.sonarsource.com/csharp/RSPEC-4823`).

TODO:
- [x] Fix rules coming from csharpsquid repository
- [x] Check vb.net hotspots --> no problem for VB.NET as the repo's name is identical to the category on rules.sonarsource.com
- [x] Check if there are hotspots in other repositories --> yes, for JS and TS, but JS/TS doesn't support connected mode yet. See repo names on the java side: https://github.com/SonarSource/sonarlint-core/blob/bb9077fbfaaaf2bb51803ceee5c3bb89adf9c607/core/src/main/java/org/sonarsource/sonarlint/core/serverapi/issue/IssueApi.java#L52
- [x] Check if taint vulnerabilities are impacted --> no, they are in the `roslyn.sonaranalyzer.security.cs` repository.

rita-sonarsource:
Fixes #2411

### DIFF
--- a/src/Core/RuleHelpLinkProvider.cs
+++ b/src/Core/RuleHelpLinkProvider.cs
@@ -32,7 +32,7 @@ namespace SonarLint.VisualStudio.Core
         private static readonly IDictionary<string, string> repoKeyToFolderNameMap = new Dictionary<string, string>
         {
             { SonarRuleRepoKeys.CSharpSecurityRules, "csharp" },
-            { SonarRuleRepoKeys.CSharpHotspotSecurityRules, "csharp" }
+            { SonarRuleRepoKeys.CSharpRules, "csharp" }
         };
 
         public string GetHelpLink(string ruleKey)
--- a/src/Core/RuleHelpLinkProvider.cs
+++ b/src/Core/RuleHelpLinkProvider.cs
@@ -31,7 +31,8 @@ namespace SonarLint.VisualStudio.Core
     {
         private static readonly IDictionary<string, string> repoKeyToFolderNameMap = new Dictionary<string, string>
         {
-            { SonarRuleRepoKeys.CSharpSecurityRules, "csharp" }
+            { SonarRuleRepoKeys.CSharpSecurityRules, "csharp" },
+            { SonarRuleRepoKeys.CSharpHotspotSecurityRules, "csharp" }
         };
 
         public string GetHelpLink(string ruleKey)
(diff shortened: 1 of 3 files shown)


src/Core/RuleHelpLinkProvider.cs
```diff
@@ -31,7 +31,8 @@ public class RuleHelpLinkProvider : IRuleHelpLinkProvider
     {
         private static readonly IDictionary<string, string> repoKeyToFolderNameMap = new Dictionary<string, string>
         {
-            { SonarRuleRepoKeys.CSharpSecurityRules, "csharp" }
+            { SonarRuleRepoKeys.CSharpSecurityRules, "csharp" },
+            { SonarRuleRepoKeys.CSharpHotspotSecurityRules, "csharp" }
```
rita-sonarsource, line 35: I took the time to see if there are other hotspot/taint repositories. There are for JS/TS: https://github.com/SonarSource/sonarlint-core/blob/bb9077fbfaaaf2bb51803ceee5c3bb89adf9c607/core/src/main/java/org/sonarsource/sonarlint/core/serverapi/issue/IssueApi.java#L52

However we don't have connected mode for JS/TS, so hotspots/taint are not supported anyway. I've added this info to the connected mode ticket so we won't forget.

src/Core/SonarArtefactKeys.cs
```diff
@@ -67,6 +67,7 @@ public static class SonarRuleRepoKeys
         public const string Cpp = "cpp";
 
         public const string CSharpSecurityRules = "roslyn.sonaranalyzer.security.cs";
+        public const string CSharpHotspotSecurityRules = "csharpsquid";
```
rita-sonarsource, line 70: Is it right to call it "HotspotSecurity"? I believe this is just the normal csharp repository, no? (that the analyzer strips out the repository for the editor, and we should do the same for hotspots, so a better name would be simply CSharpRules)

duncanp-sonar: It is just the normal repo, so `CSharpRules` would do.

FYI the diagnostic descriptors for "normal" C# Roslyn rules just use the simple `Sxxx` values as the rule ID, so that is was is reported in the IDE i.e. there is nothing to strip out. The only reason we have a problem with security rules and hotspot rules is that we are fetching the data from the server, which uses the composite IDs.

duncanp-sonar (approved):

duncanp-sonar:
Oops, might have approved too soon. What about VB.NET? (we have hotspot rules for VB: https://rules.sonarsource.com/vbnet/type/Security%20Hotspot)

rita-sonarsource:
> Oops, might have approved too soon. What about VB.NET? (we have hotspot rules for VB: https://rules.sonarsource.com/vbnet/type/Security%20Hotspot)

I will take a look, good catch. I think last time we wrote this there were no VS rules so I forgot all about this.

PR #2495, opened 2021-07-14
Added debug logging for binding

duncanp-sonar:


src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -39,19 +40,28 @@ internal class CSharpVBProjectBinder : IProjectBinder
         private readonly ICSharpVBAdditionalFileReferenceChecker additionalFileReferenceChecker;
         private readonly IRuleSetSerializer ruleSetSerializer;
         private readonly CreateBindingOperationFunc createBindingOperationFunc;
+        private readonly ILogger logger;
 
         public CSharpVBProjectBinder(IServiceProvider serviceProvider, IFileSystem fileSystem, ILogger logger)
-            : this(serviceProvider, fileSystem, new RuleSetReferenceChecker(serviceProvider, logger), new CSharpVBAdditionalFileReferenceChecker(serviceProvider), GetCreateBindingOperationFunc(serviceProvider, logger))
+            : this(serviceProvider, fileSystem, logger,
```
duncanp-sonar, line 46: Added the logger as an externally-provided (i.e. non-test) parameter.
Moved the creation of the injectable parameters onto separate lines for readability.

src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -87,19 +97,27 @@ public bool IsBindingRequired(BindingConfiguration binding, Project project)
 
         private bool IsFullyBoundProject(BindingConfiguration binding, Project project, Language language)
         {
-            if (!IsSolutionBound(binding, language, out var solutionRuleSetFilePath, out var additionalFilePath))
+            logger.LogDebug($"[Binding check] Checking binding for project '{project.Name}', language '{language}'");
```
duncanp-sonar, line 100: All of the message are debug-only, so I haven't made them localisable.

src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -87,19 +97,27 @@ public bool IsBindingRequired(BindingConfiguration binding, Project project)
 
         private bool IsFullyBoundProject(BindingConfiguration binding, Project project, Language language)
         {
-            if (!IsSolutionBound(binding, language, out var solutionRuleSetFilePath, out var additionalFilePath))
```
duncanp-sonar, line 90: The behaviour should be exactly the same as before. However, I had to refactor slightly to capture the value(s) to log. The checks are now all in the form:

```
var someCondition = CheckCondition();
logger.LogDebug( log the result);

if (!someCondition)
{
  return;
}

```

src/Integration/LocalServices/UnboundProjectFinder.cs
```diff
@@ -64,14 +72,20 @@ private IEnumerable<Project> GetUnboundProjects(BindingConfiguration binding)
             var projectSystem = serviceProvider.GetService<IProjectSystemHelper>();
             projectSystem.AssertLocalServiceIsNotNull();
 
-            var filteredSolutionProjects = projectSystem.GetFilteredSolutionProjects();
+            var filteredSolutionProjects = projectSystem.GetFilteredSolutionProjects().ToArray();
```
duncanp-sonar, line 75: This logging is in a project-type-agnostic class so it will log messages for all bound project types i.e. including C++.

src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -87,19 +97,27 @@ public bool IsBindingRequired(BindingConfiguration binding, Project project)
...
+            logger.LogDebug($"[Binding check] Checking binding for project '{project.Name}', language '{language}'");
+
+            var isSolutionBound = IsSolutionBound(binding, language, out var solutionRuleSetFilePath, out var additionalFilePath);
+            logger.LogDebug($"[Binding check] Is solution bound: {isSolutionBound}");
+
+            if (!isSolutionBound)
             {
+                logger.LogDebug("[Binding check] Solution is not bound. Skipping project-level checks.");
```
duncanp-sonar, line 107: The C#/VB checks look at multiple conditions, so I've logged data about each check.

src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -108,15 +126,20 @@ private bool IsSolutionBound(BindingConfiguration binding, Language language, ou
...
             solutionRuleSetFilePath = solutionRuleSet?.FilePath;
 
-            return solutionRuleSet != null;
+            var rulesetExists = solutionRuleSet != null;
+            logger.LogDebug($"[Binding check] Solution-level ruleset exists: {rulesetExists} (file path: '{solutionRuleSetFilePath}')");
+
+            return rulesetExists;
         }
 
         private RuleSet GetSolutionRuleSet(BindingConfiguration binding, Language language)
```
rita-sonarsource, line 145: It might be interesting to add logging in `ruleSetSerializer.LoadRuleSet` -- at the moment I see that it catches exceptions and writes them into `Trace.WriteLine`.

src/Integration/LocalServices/UnboundProjectFinder.cs
```diff
@@ -64,14 +72,20 @@ private IEnumerable<Project> GetUnboundProjects(BindingConfiguration binding)
...
+            logger.LogDebug($"[Binding check] Number of bindable projects: {filteredSolutionProjects.Length}");
```
rita-sonarsource, line 76: Would it perhaps be more useful to log the names of the projects?

src/Integration/LocalServices/UnboundProjectFinder.cs
```diff
@@ -33,27 +34,34 @@ internal class UnboundProjectFinder : IUnboundProjectFinder
...
             var configProvider = serviceProvider.GetService<IConfigurationProviderService>();
             configProvider.AssertLocalServiceIsNotNull();
 
             // Only applicable in connected mode (legacy or new)
             var bindingConfig = configProvider.GetConfiguration();
+            logger.LogDebug($"[Binding check] Binding mode: {bindingConfig.Mode}");
 
-            return bindingConfig.Mode.IsInAConnectedMode() ? GetUnboundProjects(bindingConfig) : Enumerable.Empty<Project>();
+            var unbound = bindingConfig.Mode.IsInAConnectedMode() ? GetUnboundProjects(bindingConfig) : Enumerable.Empty<Project>();
+            logger.LogDebug($"[Binding check] Number of unbound projects: {unbound}");
```
rita-sonarsource, line 63: I think the should be `unbound.Count()`

src/Integration/LocalServices/UnboundProjectFinder.cs
```diff
@@ -64,14 +72,20 @@ private IEnumerable<Project> GetUnboundProjects(BindingConfiguration binding)
...
 
             return filteredSolutionProjects
                 .Where(project =>
                 {
                     var configProjectBinder = projectBinderFactory.Get(project);
 
-                    return configProjectBinder.IsBindingRequired(binding, project);
+                    logger.LogDebug($"[Binding check] Checking binding for project '{project.Name}'. Binder type: {configProjectBinder.GetType().Name}");
+                    var required = configProjectBinder.IsBindingRequired(binding, project);
+                    logger.LogDebug($"[Binding check] Is binding required: {required} (project: {project.Name})");
```
rita-sonarsource, line 86: Would project.Name work for open as folder projects? I can't remember. I remember some specific properties returning empty strings.

rita-sonarsource (approved):

src/Integration/LocalServices/UnboundProjectFinder.cs
duncanp-sonar: Good point. I'm not sure what we could return instead in that case.
I'll try it and see.

duncanp-sonar: Actually, "open as folder" projects should have been filtered out by this point since we shouldn't recognise them as being bind-able. I'll check, though.

src/Integration/Binding/CSharpVBProjectBinder.cs
duncanp-sonar: Ok. I'll drop the Trace statement and pass a logger to the serializer.

src/Integration/LocalServices/UnboundProjectFinder.cs
duncanp-sonar: Update: you can't mix "real" projects and "open as folder" projects in a solution.
"Open as folder" projects are not bindable in the first place, so we'd never get to this point.

rita-sonarsource: What would happen in CMake's case? Would we not support connected mode for CMake?

duncanp-sonar: At the moment we don't detect it as a bindable project type. We'd have to tackle that as part of the sprint.

PR #2497, opened 2021-07-15
Don't show empty lines in the Error List when issues are non-navigable

### LINKED ISSUE #2351: Error List displays a empty line if the span of the selected issue is invalidated

rita-sonarsource:
Fixes #2351

### DIFF
--- a/src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotFactoryTests.cs
+++ b/src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotFactoryTests.cs
@@ -77,7 +77,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.ErrorList
             testSubject.CurrentSnapshot.Should().Be(snapshot.Object);
             location.CurrentFilePath.Should().Be("some file1");
 
-            snapshot.Verify(x=> x.IncrementVersion(), Times.Never);
+            snapshot.Verify(x=> x.GetUpdatedSnapshot(), Times.Never);
             snapshot.Verify(x=> x.CreateUpdatedSnapshot(It.IsAny<string>()), Times.Never);
         }
 
@@ -106,8 +106,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.ErrorList
             location1.CurrentFilePath.Should().Be("file1");
             location2.CurrentFilePath.Should().Be("file2");
 
-            snapshot.Verify(x => x.IncrementVersion(), Times.Never);
-            updatedSnapshot.Verify(x => x.IncrementVersion(), Times.Never);
+            snapshot.Verify(x => x.GetUpdatedSnapshot(), Times.Never);
+            updatedSnapshot.Verify(x => x.GetUpdatedSnapshot(), Times.Never);
         }
 
         [TestMethod]
@@ -136,8 +136,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.ErrorList
             location1.CurrentFilePath.Should().Be("new file1");
             location2.CurrentFilePath.Should().Be("file2");
 
-            snapshot.Verify(x => x.IncrementVersion(), Times.Never);
-            updatedSnapshot.Verify(x => x.IncrementVersion(), Times.Never);
+            snapshot.Verify(x => x.GetUpdatedSnapshot(), Times.Never);
+            updatedSnapshot.Verify(x => x.GetUpdatedSnapshot(), Times.Never);
         }
 
         [TestMethod]
@@ -165,8 +165,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.ErrorList
             location1.CurrentFilePath.Should().Be("new file1");
             location2.CurrentFilePath.Should().Be("file2");
 
-            snapshot.Verify(x => x.IncrementVersion(), Times.Never);
-            updatedSnapshot.Verify(x => x.IncrementVersion(), Times.Never);
+            snapshot.Verify(x => x.GetUpdatedSnapshot(), Times.Never);
+            updatedSnapshot.Verify(x => x.GetUpdatedSnapshot(), Times.Never);
         }
 
         [TestMethod]
--- a/src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotFactoryTests.cs
+++ b/src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotFactoryTests.cs
@@ -169,6 +169,20 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.ErrorList
             updatedSnapshot.Verify(x => x.IncrementVersion(), Times.Never);
         }
 
+        [TestMethod]
+        public void UpdateSnapshot_SnapshotIsReplaced()
+        {
+            var oldSnapshot = Mock.Of<IIssuesSnapshot>();
+            var testSubject = new IssuesSnapshotFactory(oldSnapshot);
+
+            testSubject.CurrentSnapshot.Should().Be(oldSnapshot);
+
+            var newSnapshot = Mock.Of<IIssuesSnapshot>();
+            testSubject.UpdateSnapshot(newSnapshot);
+
+            testSubject.CurrentSnapshot.Should().Be(newSnapshot);
+        }
+
         private static Mock<IIssuesSnapshot> CreateIssuesSnapshot(string analyzedFilePath, params IAnalysisIssueLocationVisualization[] locations)
         {
             var snapshotMock = new Mock<IIssuesSnapshot>();
--- a/src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
+++ b/src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
@@ -304,16 +304,17 @@ namespace SonarLint.VisualStudio.Integration.Vsix
                 currentIndex >= 0 && currentIndex < issues.Count) // defensive - shouldn't happen unless VS passes an invalid index
             {
                 var issueInOldSnapshot = issues[currentIndex];
+                var issueInNewSnapshot = newIssuesSnapshot.issues.ElementAtOrDefault(currentIndex);
 
-                if (issues.Count == newIssuesSnapshot.Count)
+                // perf optimization: attempt to find the issue in the same index before doing a full search
+                if (ReferenceEquals(issueInOldSnapshot, issueInNewSnapshot))
                 {
-                    // defensive: if the issue becomes non-navigable,
-                    // SonarErrorListDataSource should create a new snapshot without it
-                    return ShouldHideIssue(issueInOldSnapshot) ? -1 : currentIndex;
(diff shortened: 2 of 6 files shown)


src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -212,10 +212,18 @@ public void Refresh(IEnumerable<string> affectedFilePaths)
...
-                    
-                    if (snapshot.FilesInSnapshot.Any(snapshotPath => affectedFilePaths.Any(affected => PathHelper.IsMatchingPath(snapshotPath, affected))))
+                    var isSnapshotAffected = snapshot.FilesInSnapshot.Any(snapshotPath => affectedFilePaths.Any(affected => PathHelper.IsMatchingPath(snapshotPath, affected)));
+
+                    if (isSnapshotAffected)
                     {
-                        snapshot.IncrementVersion();
+                        if (snapshot.HasNonNavigableIssues)
+                        {
+                            factory.UpdateSnapshot(factory.CurrentSnapshot.CreateUpdatedSnapshot());
```
rita-sonarsource, line 221: After extensive testing, I believe this is the "best" we could do with.
I tested 2 other approaches:
1. Removing the non navigable issues when IncrementVersion is called.
2. Managing 2 lists in the snapshot, Issues (for a history of all issues) and VisibleIssues (for GetValue, Count, etc).

Both approaches failed in IndexOf, returning garbage results in various scenarios, completely throwing off secondary locations. It was way worse in some cases than master, i.e. selecting a navigable issue and invalidating a non navigable issue would throw the selection, even though the user selected an issue that was not impacted.
You can read the reasoning behind this here: https://github.com/SonarSource/sonarlint-visualstudio/blob/67dd4681e3ed08911b074de1883379e8371f3c9a/src/Integration.Vsix/ErrorList/IssuesSnapshot.cs#L287

So in the end I opted for creating a new snapshot. The reasoning is that the issues _are_ changed: it's not the same exact set of issues anymore, so it makes sense to create a new one. This allows IndexOf to behave properly, as I can reliably find the old issue in the new snapshot.

src/Integration.Vsix/ErrorList/IssuesSnapshotFactory.cs
```diff
@@ -32,6 +32,8 @@ internal interface IIssuesSnapshotFactory : ITableEntriesSnapshotFactory
         /// Returns true/false if the factory was updated due to the renames
         /// </summary>
         bool HandleFileRenames(IReadOnlyDictionary<string, string> oldNewFilePaths);
+
```
rita-sonarsource, line 35: Method already existed and was used by TBIT. I just added it to the interface.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -270,22 +284,35 @@ public override bool TryCreateDetailsStringContent(int index, out string content
...
+                var issueInOldSnapshot = issues[currentIndex];
+
+                if (issues.Count == newIssuesSnapshot.Count)
+                {
+                    // defensive: if the issue becomes non-navigable,
+                    // SonarErrorListDataSource should create a new snapshot without it
+                    return ShouldHideIssue(issueInOldSnapshot) ? -1 : currentIndex;
+                }
+
+                return newIssuesSnapshot.issues.IndexOf(issueInOldSnapshot);
```
duncanp-sonar, line 315: Offline discussion: 

* fetch both by index and compare items by ref as a optimisation
* otherwise search
* if found something, then do defensive check if navigable test

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -212,10 +212,18 @@ public void Refresh(IEnumerable<string> affectedFilePaths)
...
+                        }
+                        else
+                        {
+                            snapshot.IncrementVersion();
+                        }
```
duncanp-sonar, line 226: Add comment below about lines 230+ re clearing selection based on previous contents
Rename snapshot to oldSnapshot

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -212,10 +212,18 @@ public void Refresh(IEnumerable<string> affectedFilePaths)
...
                 {
                     var snapshot = factory.CurrentSnapshot;
-                    
-                    if (snapshot.FilesInSnapshot.Any(snapshotPath => affectedFilePaths.Any(affected => PathHelper.IsMatchingPath(snapshotPath, affected))))
+                    var isSnapshotAffected = snapshot.FilesInSnapshot.Any(snapshotPath => affectedFilePaths.Any(affected => PathHelper.IsMatchingPath(snapshotPath, affected)));
+
+                    if (isSnapshotAffected)
                     {
-                        snapshot.IncrementVersion();
+                        if (snapshot.HasNonNavigableIssues)
```
duncanp-sonar, line 219: Offline discussion:
* create new "GetUpdatedSnapshot" method on snapshot that handles increment/remove non-navigable items as appropriate
* always call factory.UpdateSnapshot(...)
* ?remove IncrementSnapshot?
* HasNonNavigableIssues doesn't need to be public

src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotFactoryTests.cs
```diff
@@ -77,7 +77,7 @@ public void HandleFileRenames_NoReferencesToToRenamedFile_NoChanges()
             testSubject.CurrentSnapshot.Should().Be(snapshot.Object);
             location.CurrentFilePath.Should().Be("some file1");
 
-            snapshot.Verify(x=> x.IncrementVersion(), Times.Never);
+            snapshot.Verify(x=> x.GetUpdatedSnapshot(), Times.Never);
```
rita-sonarsource, line 80: These tests highlight that there is a problem between GetUpdatedSnapshot() and CreateUpdatedSnapshot(string) that is used for renaming. But that's a problem for another day...

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
rita-sonarsource: I decided to remove the second defensive check (about navigability). If the new snapshot contains the issue, I will just return that index and hope for VS to call it again after SonarErrorListDataSource creates a new snapshot without the non-navigable issues.

rita-sonarsource:
@duncanp-sonar please re-review the changes

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -270,23 +272,37 @@ public override bool TryCreateDetailsStringContent(int index, out string content
...
-                currentIndex >= 0 && currentIndex < issues.Count && // defensive - shouldn't happen unless VS passes an invalid index
-                !ShouldHideIssue(issues[currentIndex]) // don't map hidden issues: see #2351
-                )
+                currentIndex >= 0 && currentIndex < issues.Count) // defensive - shouldn't happen unless VS passes an invalid index
             {
-                return currentIndex;
+                var issueInOldSnapshot = issues[currentIndex];
+                var issueInNewSnapshot = newIssuesSnapshot.issues.ElementAtOrDefault(currentIndex);
+
+                // perf optimization: attempt to find the issue in the same index before doing a full search
```
duncanp-sonar, line 297: Strictly speaking setting `issueInNewSnapshot` in line 295 is part of this optimisation.
It might be better to co-locate all of the optimisation-related code i.e. move line 295 below the comment.

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -60,17 +60,19 @@ internal interface IIssuesSnapshot : ITableEntriesSnapshot
...
         /// </summary>
         IIssuesSnapshot CreateUpdatedSnapshot(string analyzedFilePath);
+
+        /// <summary>
+        /// Notifies the snapshot that some part of the contained data has changed.
+        /// The snapshot will then decide whether:
+        /// 1. Return the same snapshot instance with an incremented version so the Error List recognises it as having changed
+        /// 2. Return a new snapshot instance, excluding non-navigable issues
+        /// </summary>
+        IIssuesSnapshot GetUpdatedSnapshot();
```
duncanp-sonar, line 75: A couple of points:
* I think the detail about exactly what changes is too implementation-specific for the interface. It's important, but should be in the class instead.
* we should capture the point we discussed about object identities

I was going to try to add a GitHub `suggestion`, but it doesn't look like they can span multiple lines so I've just typed it below:

```
/// <Summary>
/// Notifies the snapshot that some part of the contained data has changed.
/// An updated snapshot will be returned that the Error List will recognise as having changed.
/// Any non-navigable items will have been removed.
/// </summary>
/// <remarks>
/// Note about object identities:
/// * the caller *cannot* rely on the identity of the snapshot object being preserved
/// * the caller *can* rely on the identity of individual issues being preserved
/// </remarks>
```

src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotFactoryTests.cs
duncanp-sonar: What's the problem?

duncanp-sonar (approved):
Looks good - just some comments about comments.
The product code is definitely simpler to understand.

src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotFactoryTests.cs
rita-sonarsource: Just the readability issue. We're testing that we call CreateUpdatedSnapshot and not GetUpdatedSnapshot -- it's clumsy and confusing, especially when you see them side-by-side in the same test.

PR #2499, opened 2021-07-15
Fix local test failures caused by wrong version of NewtonSoft being used

duncanp-sonar:


src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -55,24 +55,27 @@
...
   <Import Project="..\EmbeddedSonarAnalyzer.props" />
   <ItemGroup Label="Embedded SonarLint analyzers">
     <!-- Reference the packages we want to import, but don't include them as analyzers for this project -->
     <PackageReference Include="SonarAnalyzer.CSharp" Version="$(EmbeddedSonarAnalyzerVersion)" ExcludeAssets="all" PrivateAssets="all" />
     <PackageReference Include="SonarAnalyzer.VisualBasic" Version="$(EmbeddedSonarAnalyzerVersion)" ExcludeAssets="all" PrivateAssets="all" />
 
-    <!-- Select all of the analyzer binaries (we assume that the package restore has happened so the files are already on disc) -->
-    <SonarAnalyzerAssemblies Include="$(NuGetPackageRoot)SonarAnalyzer.CSharp\$(EmbeddedSonarAnalyzerVersion)\analyzers\*.dll" />
-    <SonarAnalyzerAssemblies Include="$(NuGetPackageRoot)SonarAnalyzer.VisualBasic\$(EmbeddedSonarAnalyzerVersion)\analyzers\*.dll" />
+    <!-- Include all of the analyzer binaries including third-party binaries in the VSIX -->
```
duncanp-sonar, line 65: Explicitly split out (A) the set of files we want to include in the VSIX and (B) the set of files we want to include the AsmRef report.

We don't need to copy set (A) to the output directory; it's enough to add them to the `VSIXSourceItem` group.

This now does what we want:
* the correct versions are included in the VSIX
* the expected files are included in the AsmRef check
* the correct NewtonSoft version is used when running the tests locally

We do need to copy set (B) to the output directory; the AsmRef report is based on the files in that directory.

The reason the tests were failing locally is because the test project was copying the v13 version of the NuGet assembly from the VSIX bin folder. I don't know why it wasn't also failing on the CI machine.

duncanp-sonar:
@rita-sonarsource could you also check that this works on your machine please?

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -55,24 +55,27 @@
...
+    <AllSonarAnalyzerAssemblies Include="$(NuGetPackageRoot)SonarAnalyzer.CSharp\$(EmbeddedSonarAnalyzerVersion)\analyzers\*.dll" />
+    <AllSonarAnalyzerAssemblies Include="$(NuGetPackageRoot)SonarAnalyzer.VisualBasic\$(EmbeddedSonarAnalyzerVersion)\analyzers\*.dll" />
+    <VSIXSourceItem Include="@(AllSonarAnalyzerAssemblies)" />
 
-    <!-- We need to reference the core analyzer assembly since we use one of the types. -->
+    <!-- Make sure the Sonar.* analyzer assemblies we want to include in the AsmRef check to the output directory. -->
+    <!-- We don't want to copy third-party assemblies from the analyzers in case there are version conflicts e.g. NewtonSoft.Json.dll -->
```
rita-sonarsource, line 71: If we are only copying the Sonar* assemblies for the AsmRef, then the Newtonsoft.Json.dll appears in the asmref report because it's used by Sonar*?

rita-sonarsource (approved):
The tests pass on my machine 🎉

src/Integration.Vsix/Integration.Vsix.csproj
duncanp-sonar: Not quite. SLVS used it anyway so it was always in the report. Now it is listed multiple times, but with different versions:

```
---
Assembly: 'SonarAnalyzer.CSharp, Version=8.26.0.0, Culture=neutral, PublicKeyToken=c5b62af9de6d7244'
Relative path: 'SonarAnalyzer.CSharp.dll'

Referenced assemblies:
...
- 'Newtonsoft.Json, Version=13.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed'
...


Assembly: 'SonarLint.2019, Version=4.37.0.0, Culture=neutral, PublicKeyToken=null'
Relative path: 'SonarLint.2019.dll'

Referenced assemblies:
...
- 'Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed'
...
```

PR #2501, opened 2021-07-16
Catch exceptions on UIThread in ErrorListInfoBarController

### LINKED ISSUE #2494: Update-to-date binding check fails with COMException
### Description

The check that projects are up to date can fail with a COMException.

Reported in this community thread: https://community.sonarsource.com/t/sonarlint-extension-when-enabled-visual-studio-2019-crash/43114

### Repro steps
Note: not repro-ed locally.

1. Binding a solution.
2. Close and re-open the solution

### Expected behavior

No exception.

### Actual behavior

Exception is thrown,  crashing VS

### Known workarounds

None

### Related information

* SonarLint for VS Version: v4.35
* Visual Studio Version (e.g. Professional v15.4.1) : VS2109
* Languages in your solution: C#, VB.Net

duncanp-sonar:
Fixes #2494

### DIFF
--- a/src/Integration.UnitTests/LocalServices/ErrorListInfoBarControllerTests.cs
+++ b/src/Integration.UnitTests/LocalServices/ErrorListInfoBarControllerTests.cs
@@ -22,6 +22,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition.Primitives;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Windows.Threading;
 using FluentAssertions;
 using Microsoft.VisualStudio;
@@ -29,6 +30,7 @@ using Microsoft.VisualStudio.ComponentModelHost;
 using Microsoft.VisualStudio.Imaging;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Core.InfoBar;
@@ -53,6 +55,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         private ConfigurableVsOutputWindowPane outputWindowPane;
         private ConfigurableConfigurationProvider configProvider;
         private ConfigurableStateManager stateManager;
+        private ILogger logger;
 
         #region Test plumbing
 
@@ -85,6 +88,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
 
             this.host = new ConfigurableHost(this.serviceProvider, Dispatcher.CurrentDispatcher);
             this.stateManager = (ConfigurableStateManager)this.host.VisualStateManager;
+            this.logger = new TestLogger();
         }
 
         #endregion Test plumbing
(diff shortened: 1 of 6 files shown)


src/Integration.UnitTests/LocalServices/ErrorListInfoBarControllerTests.cs
```diff
@@ -94,29 +98,30 @@ public void TestInit()
...
+            Action act = () => new ErrorListInfoBarController(this.host, this.unboundProjectFinder, null);
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("logger");
+        }
+
         [TestMethod]
         public void ErrorListInfoBarController_Refresh_ActiveSolutionBoundAndFullyLoaded_HasNoUnboundProjects()
         {
             // Arrange
             this.SetBindingMode(SonarLintMode.LegacyConnected);
-            var testSubject = new ErrorListInfoBarController(this.host, this.unboundProjectFinder);
```
duncanp-sonar, line 119: The controller now needs the logger as a constructor parameter.
I changed all of the existing tests to call a helper method to create the test subject rather than add the new parameter inline.

src/Integration.UnitTests/LocalServices/ErrorListInfoBarControllerTests.cs
```diff
@@ -270,12 +275,49 @@ public void ErrorListInfoBarController_CurrentBackgroundProcessorCancellation()
             this.infoBarManager.AssertHasNoAttachedInfoBar(ErrorListInfoBarController.ErrorListToolWindowGuid);
         }
 
+        [TestMethod]
+        public void ErrorListInfoBarController_Refresh_NonCriticalExceptionIsSuppressed()
```
duncanp-sonar, line 279: The new tests don't require anything from the existing `TestInitialize` method.
It was simpler just to use mocks and create everything inline than to try to change shared initialization code.

src/Integration/Helpers/Constants.cs
```diff
@@ -60,7 +60,8 @@ public static class Constants
         /// <summary>
         /// SonarLint issues home page
         /// </summary>
-        public const string SonarLintIssuesWebUrl = "https://groups.google.com/forum/#!forum/sonarlint";
```
duncanp-sonar, line 63: I didn't realise we had this generic error message with an out-of-date link. However, it was only used in one place.

I've updated the link and added some line breaks in the message to make it easier to read.

src/Integration/LocalServices/ErrorListInfoBarController.cs
```diff
@@ -169,26 +162,36 @@ private void CancelQualityProfileProcessing()
 
         internal /*for testing purposes*/ void ProcessSolutionBinding()
         {
-            // No need to do anything if by the time got here the solution was closed (or unbound)
-            var mode = this.configProvider.GetConfiguration().Mode;
-            if (!mode.IsInAConnectedMode())
+            // We could be on a UI thread so any unhandled exceptions will crash VS.
```
duncanp-sonar, line 165: I haven't checked that the other entry points in this are "safe" yet.
I'll create a separate PR if they are not.

src/Integration/LocalServices/ErrorListInfoBarController.cs
```diff
@@ -169,26 +162,36 @@ private void CancelQualityProfileProcessing()
...
+            // We can't catch at a higher level as this method queues a recursive callback
+            // if the solution is not fully loaded.
+            try
```
duncanp-sonar, line 168: No changes to the existing code - just added the try-catch.

src/Integration/Helpers/Constants.cs
rita-sonarsource: I had no idea we had this...

rita-sonarsource (approved):

PR #2503, opened 2021-07-16
Guard additional UI thread paths in ErrorListInfoBarController

duncanp-sonar:
Related to #2494

### DIFF
--- a/src/Integration.UnitTests/LocalServices/ErrorListInfoBarControllerTests.cs
+++ b/src/Integration.UnitTests/LocalServices/ErrorListInfoBarControllerTests.cs
@@ -279,14 +279,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         public void ErrorListInfoBarController_Refresh_NonCriticalExceptionIsSuppressed()
         {
             var logger = new TestLogger();
-
-            var configProvider = new Mock<IConfigurationProviderService>();
-            configProvider.Setup(x => x.GetConfiguration()).Throws(new COMException("thrown by test code"));
-
-            var host = new Mock<IHost>();
-            host.As<IServiceProvider>().Setup(x => x.GetService(typeof(IConfigurationProviderService))).Returns(configProvider.Object);
-
-            var testSubject = new ErrorListInfoBarController(host.Object, Mock.Of<IUnboundProjectFinder>(), logger);
+            var host = CreateHostWithThrowingConfigProvider(new COMException("thrown by test code"));
+            var testSubject = new ErrorListInfoBarController(host, Mock.Of<IUnboundProjectFinder>(), logger);
 
             // Act - should not throw
             testSubject.ProcessSolutionBinding();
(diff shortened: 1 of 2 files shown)


src/Integration.UnitTests/LocalServices/ErrorListInfoBarControllerTests.cs
```diff
@@ -279,14 +279,8 @@ public void ErrorListInfoBarController_CurrentBackgroundProcessorCancellation()
...
             var logger = new TestLogger();
-
-            var configProvider = new Mock<IConfigurationProviderService>();
-            configProvider.Setup(x => x.GetConfiguration()).Throws(new COMException("thrown by test code"));
-
-            var host = new Mock<IHost>();
-            host.As<IServiceProvider>().Setup(x => x.GetService(typeof(IConfigurationProviderService))).Returns(configProvider.Object);
-
-            var testSubject = new ErrorListInfoBarController(host.Object, Mock.Of<IUnboundProjectFinder>(), logger);
+            var host = CreateHostWithThrowingConfigProvider(new COMException("thrown by test code"));
```
duncanp-sonar, line 282: Refactored the two existing tests to use a helper method

src/Integration.UnitTests/LocalServices/ErrorListInfoBarControllerTests.cs
```diff
@@ -796,6 +783,30 @@ public void ErrorListInfoBarController_InfoBar_ClickButton_MoreThanOnce()
             bindCommandExecuted.Should().Be(1, "Expecting the command to be executed only once");
         }
 
+        [TestMethod]
```
duncanp-sonar, line 786: Added critical/non-critical exception tests for the one event handler that could be easily tested.

src/Integration/LocalServices/ErrorListInfoBarController.cs
```diff
@@ -255,38 +256,58 @@ private void UpdateRequired(string customInfoBarMessage = null)
 
         private void CurrentErrorWindowInfoBar_Closed(object sender, EventArgs e)
         {
-            this.ClearCurrentInfoBar();
+            // Called on the UI thread -> must handle exceptions
```
duncanp-sonar, line 259: There are two other event handlers for which uncaught exceptions (will cause VS to crash - clicking on the button in the infobar, and closing the infobar.

I haven't changed the existing code, just wrapped it in a try-catch block.

I couldn't sensibly add a test for this click handler. The handler only calls one injectable object, but that object is only created after a series of stateful transitions (i.e. checking that an update is required, then creating and wiring up an infobar).

The test would either be extremely complex, or I'd have to do quite a bit of refactoring of the product code. Writing the test didn't seem like a good use of time, and changing the product code is too risky.

I've manually tested both code paths by debugging and changing the code to throw an exception at an appropriate point.

duncanp-sonar:
The failing check is due to insufficient coverage - seven out of ten lines are covered. The other lines are not easily testable.

rita-sonarsource (approved):

PR #2510, opened 2021-07-20
Add CFamily assembly

### LINKED ISSUE #2465: Create new CFamily assembly
#### Description

Create new assembly for the CFamily analyzer code, similar to TypeScript.

Name: `SonarLint.VisualStudio.CFamily`

rita-sonarsource:
Fixes #2465


src/CFamily/CFamily.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net46</TargetFramework>
+    <ProjectGuid>{58619C0F-0F3D-4E8C-B204-A19B332D45E5}</ProjectGuid>
+    <RootNamespace>SonarLint.VisualStudio.CFamily</RootNamespace>
+    <AssemblyName>SonarLint.VisualStudio.CFamily</AssemblyName>
+    <RequiresSigning>true</RequiresSigning>
+    <RequiresNewtonsoft>true</RequiresNewtonsoft>
+    <CopyBuildOutputToOutputDirectory>true</CopyBuildOutputToOutputDirectory>
```
duncanp-sonar, line 10: Why do we need this property? Most projects don't have it. 
Is the assembly still included in the VSIX without it?

duncanp-sonar (approved):

src/CFamily/CFamily.csproj
rita-sonarsource: Yes indeed it's not needed. I copied it originally from Typescript --> which was copied from IssueViz --> and there I believe it's needed because it's a package. So I will remove it.

Issue #2195, opened 2021-02-25
Locate the compilation database file for the current file


rita-sonarsource:
#### Description

CMake projects can take advantage of a compilation database, which will contain build instructions for cpp files in each project.

Notes:
* The compilation database file (`compile_commands.json`) can be generated for the entire "solution" (by placing `set(CMAKE_EXPORT_COMPILE_COMMANDS ON)` on the root `CMakeLists.txt`), or only for specific projects (by placing the command on specific `CMakeLists.txt` files). 
However, there will still be only one compilation database file, and it will contain entries for projects that participated in the generation.
* The analyzed file might not have an entry in the compilation database file, if its project did not turn on the flag. 
* The analyzed file might have multiple entries in the compilation database file, if it was referenced by more than one project.
* Header files are not listed in the compilation database file.
* The build configuration is stored under the global `CMakeSettings.json`, which is an optional file. If the file exists, the `buildRoot` property can be overriden, which will change the location of the compilation database file.
* It is possible to build the project via command-line, in which case the output directory will be specified by the user and unknown to us.

TODO:
- [ ] ~~Investigate how to collect the current build configuration (VS API?)~~ --> will be done in #2508
- [ ] ~~Investigate how to find the output directory (if the `CMakeSettings.json` file exists, can we use `buildRoot` directory? what if the project was built via command line and a different output directory was specified? should we just attempt to auto-detect as a first step and document that in the wiki?)~~ --> will not be supported in the first phase, we will use only what is configured in the `CMakeSettings.json`
- [ ] ~~Investigate which entry to use in the compilation database file if there is more than 1 entry for the analyzed file (i.e. if multiple projects reference it).~~ --> will be done in #2509
- [ ] Implement a search logic, that identifies the current build configuration and locates the compilation database file for the given build configuration.
- [ ] Support default mode - when there is no CMakeSettings.json
- [ ] Provide meaningful logging + documentation link if the compilation database file does not exist

PR #2512, opened 2021-07-21
Add ProjectDirectoryProvider

rita-sonarsource:
Prep for #2195

### DIFF
--- a/src/Infrastructure.VS/ProjectDirectoryProvider.cs
+++ b/src/Infrastructure.VS/ProjectDirectoryProvider.cs
@@ -1,0 +1,94 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.ComponentModel.Composition;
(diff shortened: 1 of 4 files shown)


src/TypeScript/TsConfig/TsConfigsLocator.cs
```diff
@@ -93,41 +73,5 @@ public IReadOnlyList<string> Locate(string sourceFilePath)
...
-                    "Failed to retrieve VSPROPID_SolutionDirectory");
-
-                return (string) solutionDirectory;
-            }
-
-            var vsProject = vsHierarchyLocator.GetVsHierarchyForFile(sourceFilePath);
-
-            if (vsProject == null)
-            {
-                logger.WriteLine(Resources.ERR_NoVsProject, sourceFilePath);
```
rita-sonarsource, line 119: extracted as-is minus this log output

src/Infrastructure.VS/ProjectDirectoryProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+        public ProjectDirectoryProvider([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider,
+            IVsHierarchyLocator vsHierarchyLocator)
+        {
+            vsSolution = serviceProvider.GetService(typeof(SVsSolution)) as IVsSolution;
+            this.vsHierarchyLocator = vsHierarchyLocator;
+        }
+
+        public string GetProjectDirectory(string sourceFilePath)
+        {
+            // "Open as Folder" was introduced in VS2017, so in VS2015 the hr result will be `E_NOTIMPL` and `isOpenAsFolder` will be null.
```
duncanp-sonar, line 61: Looking at this again, we're making a fairly big assumption, namely that in the "open as folder" case the supplied file path is under the folder root. That's true for the way we call it currently, but it looks odd now the interface has been separated out.

Effectively, the behaviour in the "open as folder" case is that the caller can pass in any garbage and the function will return the root folder path.
I think we should either document this behaviour on the interface, or add a check to the code.

duncanp-sonar (commented):
The changes look good.

However, I'm not clear how we'll use this in CMake scenario. Which file name are we going to pass in to `GetProjectDirectory`?

src/Infrastructure.VS/ProjectDirectoryProvider.cs
rita-sonarsource: This is actually exactly the design problem I'm 'struggling' with right now.
Eventually the higher caller would call something like
`CompilationOptions Get(string analyzedFile)` --> and that in turn would get the compilation database, get the entry and parse it.
So in a way we do always start with a file, but strictly speaking, I'm writing a class now `CompilationDatabase Locate()` that absolutely does not need a file path.

PR #2514, opened 2021-07-22
Locate Compilation Database file

### LINKED ISSUE #2195: Locate the compilation database file for the current file

rita-sonarsource:
Fixes #2195


src/CFamily/CompilationDatabase/CMakeSettings.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <summary>
+    /// Schema based on https://docs.microsoft.com/en-us/cpp/build/cmakesettings-reference?view=msvc-160
+    /// </summary>
+    internal class CMakeSettings
+    {
+        [JsonProperty("configurations")]
+        public CMakeBuildConfiguration[] Configurations { get; set; }
+    }
+
+    internal class CMakeBuildConfiguration
```
rita-sonarsource, line 34: I didn't include property that we don't use (yet) just to keep it simple for now.

src/CFamily/CompilationDatabase/CompilationDatabaseLocator.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// <summary>
+        /// Returns absolute path to the compilation database file of the currently active build configuration.
+        /// Returns null if the file was not found.
+        /// </summary>
+        string Locate();
+    }
+
+    [Export(typeof(ICompilationDatabaseLocator))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class CompilationDatabaseLocator : ICompilationDatabaseLocator
```
duncanp-sonar, line 42: This logic is specific to VS-CMake, isn't it? It would be better if the class name/namespace reflected that (the `CMakeSettings` class is also VS-CMake-specific, so the same point about the namespace would apply).

? `SonarLint.VisualStudio.CFamily.CMake` ?

src/CFamily/CompilationDatabase/CompilationDatabaseLocator.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        internal const string CompilationDatabaseFileName = "compile_commands.json";
+        internal const string CMakeSettingsFileName = "CMakeSettings.json";
+        internal const string DefaultConfiguration = "x64-Debug";
```
duncanp-sonar, line 46: Maybe `VSDefaultConfiguration` to make it clearer that the default is being copied from VS, rather than being one we've created?

src/CFamily/CompilationDatabase/CompilationDatabaseLocator.cs
```diff
@@ -0,0 +1,130 @@
...
+
+                logger.WriteLine(Resources.NoCMakeSettings, CMakeSettingsFileName, rootDirectory, defaultLocation);
+
+                return defaultLocation;
+            }
+
+            logger.WriteLine(Resources.ReadingCMakeSettings, cmakeSettingsFullPath);
+
+            var settingsString = fileSystem.File.ReadAllText(cmakeSettingsFullPath);
+            var settings = JsonConvert.DeserializeObject<CMakeSettings>(settingsString);
```
duncanp-sonar, line 114: This will throw if the json is invalid, won't it?

It's not clear how we want to handle this yet (i.e. catch here or higher up). It's worth adding a note to the board to test invalid json to make sure we don't crash VS.

src/CFamily/CompilationDatabase/Resources.resx
```diff
@@ -0,0 +1,138 @@
...
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="FoundCompilationDatabaseFile" xml:space="preserve">
+    <value>[CompilationDatabaseLocator] Reading compilation database from '{0}</value>
+  </data>
+  <data name="NoBuildConfigInCMakeSettings" xml:space="preserve">
+    <value>[CompilationDatabaseLocator] The specified build configuration '{0}' could not be found in the {1}</value>
```
duncanp-sonar, line 124: I don't think the "the" is needed in the message.

src/CFamily.UnitTests/CompilationDatabase/CompilationDatabaseLocatorTests.cs
```diff
@@ -0,0 +1,204 @@
...
+            if (configuredPathExists)
+            {
+                result.Should().Be(compilationDatabaseFullLocation);
+            }
+            else
+            {
+                result.Should().BeNull();
+            }
+
+            fileSystem.Verify(x => x.File.Exists(compilationDatabaseFullLocation), Times.Once);
```
duncanp-sonar, line 136: In general, it would be worth asserting that some of the expected data is logged e.g. the missing build config name, the missing file path etc.

duncanp-sonar (commented):
Looks good - just a few comments.

src/CFamily/CompilationDatabase/Resources.resx
rita-sonarsource: Which "the"? first/second/both?

PR #2516, opened 2021-07-23
Add CMake IActiveConfigProvider

### LINKED ISSUE #2508: Discover active build configuration in CMake projects
#### Description

`CMakeSettings.json` is a VS-specific file, containing all the build configurations for a given CMake "solution". In order to support #2195, we need to find a way to get the active build configuration. 

image

TODO:
- [x] Find a way to retrieve the active build configuration for CMake projects.

##### Standard VS APIs
The current configuration setting does not appear to be accessible from the standard VS APIs e.g.
* `dte.Solution.FindProjectItem(path).ConfigurationManager` --> returns null
* `projectItem.ContainingProject.ConfigurationManager` --> returns null
* `IVsBuildManagerAccessor.GetSolutionConfiguration` for both `IVsSolution` as `IVsHierarchy`, and converting
* `projectItem.ContainingProject` into  `IVsHierarchy`.
* `IVsSolutionBuildManager.FindActiveProjectCfg` for both `IVsSolution` as `IVsHierarchy`, and converting
* `projectItem.ContainingProject` into `IVsHierarchy`.
* Converting `IVsSolution` into `IVsBuildPropertyStorage` --> null

#### VS CMake implementation notes

[MS docs
](https://docs.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=msvc-160)
[MS blog post](https://devblogs.microsoft.com/cppblog/cmake-support-in-visual-studio/)

VS has a set of classes to provide general-purpose "open as folder" functionality, with extension points to allow for folder/project-type-specific plugins to be added (see [IVsSolutionWorkspaceService](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.workspace.vsintegration.ivssolutionworkspaceservice?view=visualstudiosdk-2017) as a starting point).

The VS CMake functionality is built on top of the generic "open as folder" experience. It has public interfaces and MEF exports of custom CMake-specific types. Unfortunately, these are not available via NuGet packages, so they can't easily be referenced or used.

The current configuration setting does not appear to be available through the generic IVsSolutionWorkspaceService APIs.

##### ProjectSettings.json
The current configuration is saved in `.vs\ProjectSettings.json` e.g.
```
{
  "CurrentProjectSetting": "x64-Release"
}
```
The file is updated as soon as the user selects another value in the dropdown in the UI.
This is not a public/external extension point so the behaviour could change at any time. However, it appears to be stable across VS2017 and VS2019.

Note that VS does not read changes to this file in real time e.g. if the same CMake project is open in two instances of VS at the same time, changing the selected value in VS (1) will update the file immediately, but VS (2) won't pick up the change. However, this won't matter in practice.

duncanp-sonar:
Fixes #2508


src/CFamily/CMake/ActiveConfigProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// <returns>The active config, or the default config if the setting could not be found/read</returns>
+        public string GetActiveConfig(string rootDirectory)
+        {
+            if (rootDirectory == null)
+            {
+                throw new ArgumentNullException(nameof(rootDirectory));
+            }
+
+            var fullPath = Path.Combine(rootDirectory, RelativeSettingsPath);
+            LogDebug("Internal VS CMake settings file: " + fullPath);
```
duncanp-sonar, line 70: I made all of the log message debug since they will be written for every analysis.
I see the compilation db locator ones are not debug. They should probably be consistent. What level do you think they should be?

src/CFamily/CMake/ActiveConfigProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+using System;
+using System.IO;
+using System.IO.Abstractions;
+using Newtonsoft.Json.Linq;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.Helpers;
+
+namespace SonarLint.VisualStudio.CFamily.CMake
+{
+    internal interface IActiveConfigProvider
```
rita-sonarsource, line 30: `Config` can be a bit ambiguous in this context, maybe `ActiveBuildConfig` ?

src/CFamily/CMake/ActiveConfigProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// Returns the name of the current configuration
+        /// </summary>
+        /// <param name="rootDirectory">The root directory for the solution</param>
+        string GetActiveConfig(string rootDirectory);
+    }
+
+    internal class ActiveConfigProvider : IActiveConfigProvider
+    {
+        private const string VSDefaultConfiguration = "x64-Debug";
+        private const string RelativeSettingsPath = ".vs\\ProjectSettings.json";
```
rita-sonarsource, line 42: Just wondering, is that configurable in any way?

src/CFamily/CMake/ActiveConfigProvider.cs
rita-sonarsource: Oof, that's a good question. I haven't thought about the fact that it will run for every single analysis.
I'll change them to debug too.

src/CFamily.UnitTests/CMake/ActiveConfigProviderTests.cs
```diff
@@ -0,0 +1,120 @@
...
+using SonarLint.VisualStudio.CFamily.CMake;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.UnitTests;
+
+namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
+{
+    [TestClass]
+    public class ActiveConfigProviderTests
+    {
+        private const string ExpectedDefaultConfig = "x64-Debug";
```
rita-sonarsource, line 36: Maybe make the const in the class Internal and re-use it?

src/CFamily.UnitTests/CMake/ActiveConfigProviderTests.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        private static void AssertSettingsFileRead(Mock<IFileSystem> fileSystem, string rootDirectory)
+        {
+            var fullPath = CalcFullSettingsPath(rootDirectory);
+            fileSystem.Verify(x => x.File.ReadAllText(rootDirectory), Times.Once);
+        }
+
+        private static string CalcFullSettingsPath(string rootDirectory) =>
+            Path.Combine(rootDirectory, ".vs", "ProjectSettings.json");
```
rita-sonarsource, line 118: I think it's worth making RelativeSettingsPath internal in the product code and re-using it.

src/CFamily.UnitTests/CMake/ActiveConfigProviderTests.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.CFamily.CMake;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.UnitTests;
+
+namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
+{
+    [TestClass]
+    public class ActiveConfigProviderTests
```
rita-sonarsource, line 34: worth adding a test for failure to read the file

src/CFamily.UnitTests/CMake/ActiveConfigProviderTests.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        private static void AssertSettingsFileExistenceChecked(Mock<IFileSystem> fileSystem, string fullSettingsPath)
+        {
+            fileSystem.Verify(x => x.File.Exists(fullSettingsPath), Times.Once);
+        }
+
+        private static void AssertSettingsFileRead(Mock<IFileSystem> fileSystem, string rootDirectory)
+        {
+            var fullPath = CalcFullSettingsPath(rootDirectory);
```
rita-sonarsource, line 113: `fullPath` is not used and `rootDirectory` is in fact `fullPath` (passed by the caller), I'm guessing it's leftovers from refactor.

rita-sonarsource (approved):
Looks good, just some comments about catching ReadAllText exception

src/CFamily/CMake/ActiveConfigProvider.cs
duncanp-sonar: I doubt it; the contents of the '.vs' folder are private, so there would no reason to make the file name/location configurable.

src/CFamily.UnitTests/CMake/ActiveConfigProviderTests.cs
duncanp-sonar: I prefer to have unit tests checking explicitly for the expected value. If the test is checking a constant on the product code then there is nothing checking that the constant has the correct value - it could contain anything, and the test would still pass.

src/CFamily/CMake/ActiveConfigProvider.cs
duncanp-sonar: I went with `IBuildConfigProvider` since the method name is `GetActiveConfig` (didn't seem to be much point in repeating "active" in the interface name).

src/CFamily.UnitTests/CMake/ActiveConfigProviderTests.cs
duncanp-sonar: (same comment as for the `ExpectedDefaultConfig` constant above)

PR #2517, opened 2021-08-02
Locate CompilationDatabaseEntry for the analyzed file

rita-sonarsource:


src/CFamily/CMake/CompilationDatabaseLocator.cs
```diff
@@ -52,68 +55,68 @@ internal class CompilationDatabaseLocator : ICompilationDatabaseLocator
...
+            ILogger logger)
         {
             this.folderWorkspaceService = folderWorkspaceService;
             this.fileSystem = fileSystem;
-            this.buildConfigProvider = activeConfigProvider;
+            this.buildConfigProvider = buildConfigProvider;
             this.logger = logger;
         }
 
         public string Locate()
```
rita-sonarsource, line 76: Previously the code would call GetCompilationDatabaseFolderPath and then check if there is a file under that folder. 
But in the end it would return only the file path, as the deserialization is done in the new `CompilationConfigProvider`.
So it looked awkward to test for file existence here.
I removed it and now essentially the logic is only the previous GetCompilationDatabaseFolderPath.

src/CFamily/CMake/CompilationDatabaseLocator.cs
```diff
@@ -52,68 +55,68 @@ internal class CompilationDatabaseLocator : ICompilationDatabaseLocator
...
         {
-            var databaseFolderPath = GetCompilationDatabaseFolderPath();
+            var rootDirectory = folderWorkspaceService.FindRootDirectory();
 
-            if (string.IsNullOrEmpty(databaseFolderPath))
+            if (string.IsNullOrEmpty(rootDirectory))
             {
+                logger.LogDebug(Resources.NoRootDirectory);
```
rita-sonarsource, line 82: Changed some strings to debug as we discussed

src/CFamily/CMake/CompilationDatabaseLocator.cs
```diff
@@ -52,68 +55,68 @@ internal class CompilationDatabaseLocator : ICompilationDatabaseLocator
...
 
-        public string GetCompilationDatabaseFolderPath()
+        private string GetConfiguredLocation(string cmakeSettingsFullPath, string activeConfiguration, string rootDirectory)
         {
-            var rootDirectory = folderWorkspaceService.FindRootDirectory();
+            logger.LogDebug(Resources.ReadingCMakeSettings, cmakeSettingsFullPath);
+            CMakeSettings settings;
 
-            if (string.IsNullOrEmpty(rootDirectory))
+            try
```
rita-sonarsource, line 109: The try-catch was missing before. I decided to catch it here, instead of in the caller, and added new tests.

src/CFamily/CMake/Resources.resx
```diff
@@ -117,17 +117,26 @@
...
   </data>
   <data name="NoBuildConfigInCMakeSettings" xml:space="preserve">
     <value>[CompilationDatabaseLocator] Specified build configuration '{0}' could not be found at '{1}'</value>
   </data>
   <data name="NoCMakeSettings" xml:space="preserve">
-    <value>[CompilationDatabaseLocator] No {0} was found under {1}, searching for compilation database under default location: {2}</value>
+    <value>[CompilationDatabaseLocator] No {0} was found under {1}, returning default location: {2}</value>
+  </data>
+  <data name="NoCompilationDatabaseEntry" xml:space="preserve">
+    <value>[CompilationConfigProvider] Could not find file '{0}' in the compilation database.</value>
```
duncanp-sonar, line 136: File paths are often longer than the width of the output window, so putting the full file name in the middle of a message means the user has to scroll to understand the message.

If the variable part of the content can make the message harder to read, I generally move it to the end e.g.

`[CompilationConfigProvider] Could not find file in the compilation database. File: {file name}`

src/CFamily/CMake/Resources.resx
```diff
@@ -117,17 +117,26 @@
...
   </data>
   <data name="NoCompilationDatabaseFile" xml:space="preserve">
-    <value>[CompilationDatabaseLocator] Could locate compilation database at '{0}'. Make sure that your project is configured correctly. See https://github.com/SonarSource/sonarlint-visualstudio/wiki for more information.</value>
+    <value>[CompilationConfigProvider] Could locate compilation database at '{0}'. Make sure that your project is configured correctly. See https://github.com/SonarSource/sonarlint-visualstudio/wiki for more information.</value>
```
duncanp-sonar, line 139: I'd put a line break before the second sentence and indent it slightly to make the "where to get help" text easy to see.

src/CFamily/CMake/Resources.resx
```diff
@@ -117,17 +117,26 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="BadCMakeSettings" xml:space="preserve">
+    <value>[CompilationDatabaseLocator] Could not parse CMakeSettings.json: {0}</value>
```
duncanp-sonar, line 121: Are these message debug-only, or are some of them info?

We should agree on whether we are going to localise all debug messages or not.

src/CFamily/CMake/CompilationDatabaseEntry.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    /// <summary>
+    /// Schema based on https://clang.llvm.org/docs/JSONCompilationDatabase.html
+    /// </summary>
+    internal class CompilationDatabaseEntry
+    {
+        [JsonProperty("directory")]
+        public string Directory { get; set; }
+
+        [JsonProperty("command")]
```
duncanp-sonar, line 33: The CFamily doc says "command or arguments" should be passed.
Is support for arguments going to be added separately?

src/CFamily/CMake/Resources.resx
duncanp-sonar: Since this is a non-debug message, I don't think it should include the `CompilationConfigProvider` prefix - it's an implementation detail, and isn't going to be meaningful to the end user.

I think it's worth having a user-friendly prefix, though e.g. `[CMake]`

src/CFamily/CMake/CompilationConfigProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace SonarLint.VisualStudio.CFamily.CMake
+{
+    internal interface ICompilationConfigProvider
+    {
+        /// <summary>
+        /// Returns the compilation configuration for the given file,
+        /// as specified in the compilation database file for the currently active build configuration.
+        /// Returns null if there is no compilation database or if the file does not exist in the compilation database.
+        /// </summary>
+        CompilationDatabaseEntry GetConfig(string analyzedFile);
```
duncanp-sonar, line 41: Minor: `filePath` might be better than `analyzedFile`
* makes it obvious that a full path is expected
* `analyzed` is leaking information about how a particular caller is using this service into the naming. This service shouldn't know or care what the caller is doing.

src/CFamily/CMake/Resources.resx
```diff
@@ -117,17 +117,26 @@
...
+  </data>
+  <data name="BadCompilationDatabaseFile" xml:space="preserve">
+    <value>[CompilationConfigProvider] Could not parse compilation database: {0}</value>
+  </data>
   <data name="FoundCompilationDatabaseFile" xml:space="preserve">
-    <value>[CompilationDatabaseLocator] Reading compilation database from '{0}</value>
+    <value>[CompilationConfigProvider] Reading compilation database from '{0}</value>
   </data>
   <data name="NoBuildConfigInCMakeSettings" xml:space="preserve">
     <value>[CompilationDatabaseLocator] Specified build configuration '{0}' could not be found at '{1}'</value>
```
duncanp-sonar, line 130: Not changed in this PR, but this is a non-debug message, so the `[CompilationDatabaseLocator]` prefix won't be meaningful to the user (same comment for new line #139).

src/CFamily.UnitTests/CMake/CompilationConfigProviderTests.cs
```diff
@@ -0,0 +1,236 @@
...
+            var logger = new TestLogger();
+
+            var testSubject = CreateTestSubject(compilationDatabaseLocator.Object, fileSystem.Object, logger);
+
+            var result = testSubject.GetConfig("c:\\some file.cpp");
+            result.Should().BeEquivalentTo(compilationDatabaseContent[0]);
+        }
+
+        [TestMethod]
+        public void GetConfig_MultipleEntriesForFile_ReturnsFirstOne()
```
duncanp-sonar, line 179: Is this a valid scenario, or is the produce code just being defensive?

If the product code is just being defensive then it would be worth adding a comment to make it clear e.g. 
`Defensive: a valid compilation database should only have at most one entry per file`.

If a compilation database can validly have multiple entries for the same file then we probably need to do more work on our side.

duncanp-sonar (commented):
Looks good. Mainly comments on logging.

src/CFamily/CMake/CompilationDatabaseEntry.cs
rita-sonarsource: Good catch, I missed it in the schema example

src/CFamily/CMake/Resources.resx
rita-sonarsource: Changed to [CMake] for all the info messages

src/CFamily.UnitTests/CMake/CompilationConfigProviderTests.cs
rita-sonarsource: It can be included multiple times, yes, for instance if it's referenced by two different projects.
I don't know how we would go about getting the right compilation in this case, I'll ask Abbas

rita-sonarsource: FYI according to Abbas it's possible but unlikely, so any entry will do.

PR #2526, opened 2021-08-06
Add new protocol writer interface

rita-sonarsource:


src/Core/CFamily/IProtocolWriterFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace SonarLint.VisualStudio.Core.CFamily
+{
+    public interface IProtocolWriterFactory
+    {
+        /// <summary>
+        /// Creates <see cref="IProtocolWriter"/> that can serialize the given <see cref="IRequest"/>.
+        /// Will return null if there is no supported protocol for this request.
+        /// </summary>
+        IProtocolWriter TryGet(IRequest request);
```
rita-sonarsource, line 29: Technically we only need `string analyzedFilePath` at the moment, but since the protocol is responsible for writing the request, it's a bit odd not to pass the request in.

src/Core/CFamily/IRequestBuilder.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace SonarLint.VisualStudio.Core.CFamily
+{
+    public interface IRequestBuilder
+    {
+        /// <summary>
+        /// Creates <see cref="IRequest"/> for the given <see cref="analyzedFilePath"/>.
+        /// Returns null if request could not be created.
+        /// </summary>
+        IRequest TryBuild(string analyzedFilePath, CFamilyAnalyzerOptions analyzerOptions);
```
rita-sonarsource, line 29: CLangAnalyzer will start by calling this interface, but it's unclear yet what will be the implementation inside: 
- will we add `IRequest.ProtocolWriter` property and set it?
- or will we have this create only the request object, and then CLangAnalyzer will call IProtocolWriterFactory? (effectively then passing down the line IProtocolWriter instead of IRequest).

src/Core/CFamily/IProtocolWriterFactory.cs
duncanp-sonar: Hhmm, so the request will have to have a settable "writer" property?

The comment is a bit misleading as it implies that there are different types of request, whereas now there will only be a single type of request. Also, it suggests factory is just returning a serializer, whereas it is also responsible for fetching additional request-specific data.

src/Core/CFamily/IRequestBuilder.cs
duncanp-sonar: I don't think the CLangAnalyzer should know that the protocol writer factory exists.
The RequestBuilder should be responsible for doing everything that is necessary to create valid request, and the CLangAnalyzer should just work with IRequest.

The writer could either be exposed as a property, or we could leave IRequest as it is and just make the Request.WriteX methods delegate the work to the selected protocol writer i.e. we could hide the existence of the protocol writer as an implementation detail inside the request code. I don't think it matters too much.

src/Core/CFamily/IProtocolWriterFactory.cs
rita-sonarsource: > settable "writer" property
If we don't want CLangAnalyzer to know about protocol writer, I don't see another way. 

how about: 
```
Creates <see cref="IProtocolWriter"/> that can write the given <see cref="IRequest"/> in subprocess' expected format.
Will return null if there is no protocol that can support the arguments of this request.
```

duncanp-sonar: That still suggests that the request already contains all of the necessary data. How about:

```
<summary>
Creates a <see cref="IProtocolWriter"/> that can write the given <see cref="IRequest"/> in one of the formats supported by the subprocess.
Will return null if an appropriate protocol cannot be determined.
</summary>
<remarks>
All of the subprocess protocols require additional data that is not in the <see cref="IRequest"/>, so a writer will only be created if that additional data can be located.
</remarks>
```

rita-sonarsource: Sounds good, changed

duncanp-sonar (approved):

PR #2527, opened 2021-08-06
Moved last CFamily rules-related class from Integration.csproj

duncanp-sonar:
* "Integration" still has CFamily binding-related code


src/CFamily/CFamily.csproj
```diff
@@ -25,13 +25,22 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>Resources.resx</DependentUpon>
     </Compile>
+    <Compile Update="Rules\Resources.Designer.cs">
```
duncanp-sonar, line 28: Added a new resource file for rules-specific strings

src/CFamily/Rules/CFamilyRulesConfigProvider.cs
```diff
@@ -24,11 +24,9 @@
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Integration;
 
-// Note: currently this class needs to be in the the Integration assembly because it
-//   references IHost and other internal interfaces.
```
duncanp-sonar, line 29: We did some refactorings in April/May 2020 which mean "host" is no longer required.

I tried to move the class to Core, but it uses MEF so I moved it to CFamily instead.

duncanp-sonar:
Summary of changes:
* moved the `CFamilyRuleConfigProvider` class to `CFamily`
* moved the tests
* added a new resource file for the rule strings
* removed the strings from the Integration resource file.

rita-sonarsource (approved):

PR #2530, opened 2021-08-06
Extracted RequestContext from IRequest

duncanp-sonar:
* fix-up to CFamilyHelper / Request is hacky and needs more work

### DIFF
--- a/src/Integration.Vsix/CFamily/CFamilyHelper.cs
+++ b/src/Integration.Vsix/CFamily/CFamilyHelper.cs
@@ -74,13 +74,13 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
 
         private static Request CreateRequest(FileConfig fileConfig, string absoluteFilePath, ICFamilyRulesConfigProvider cFamilyRulesConfigProvider, IAnalyzerOptions analyzerOptions)
         {
-            var request = ToRequest(fileConfig, absoluteFilePath, out string cFamilyLanguage);
-            if (request?.File == null || cFamilyLanguage == null)
+            var request = ToRequest(fileConfig, absoluteFilePath);
+            if (request?.File == null || request?.CFamilyLanguage == null)
             {
                 return null;
             }
 
-            // TODO - remove File and PchFile from Request (both on RequestContext)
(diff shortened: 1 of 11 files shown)


src/Integration.Vsix/CFamily/PortedFromJava/Analyzer.cs
```diff
@@ -62,6 +62,9 @@ internal partial class Request
         public string Predefines { get; set; } = "";
         public string[] Macros { get; set; } = Array.Empty<string>();
         public string TargetTriple { get; set; } = "x86_64-unknown-unknown";
+
+        // File and PchFile are also now part of the C# RequestContext class
+        // TODO: refactor to remove them from this class
```
duncanp-sonar, line 67: The changes to the `CLangAnalyzer` are clean because it just used the interface directly.

The changes to the VCX proj classes are messy. Removing the duplication and fixing the tests involved a lot of changes. I decided instead just to patch up the existing code and tests to get the new interface/class checked in so we can continue working in parallel.

src/Integration.Vsix/CFamily/CFamilyHelper.cs
```diff
@@ -102,14 +102,20 @@ private static Request CreateRequest(FileConfig fileConfig, string absoluteFileP
...
-                request.Options = GetKeyValueOptionsList(request.RulesConfiguration);
+                rulesConfig = cFamilyRulesConfigProvider.GetRulesConfiguration(request.CFamilyLanguage);
+                Debug.Assert(rulesConfig != null, "RulesConfiguration should be have been retrieved");
+                request.Options = GetKeyValueOptionsList(rulesConfig);
             }
 
+            var context = new RequestContext(request.CFamilyLanguage, rulesConfig, request.File, SubProcessFilePaths.PchFilePath,
+                analyzerOptions as CFamilyAnalyzerOptions);
+
+            request.Context = context;
```
duncanp-sonar, line 117: For now, the `CFamilyHelper` just creates a new context object at that duplicates some of the information in the `Request/FileConfig`.

This would be cleaned up as part of moving to the factory model.

rita-sonarsource (approved):

PR #2532, opened 2021-08-09
Implement IRequest for compilation database entry protocol

### LINKED ISSUE #2528: Implement IRequest for compilation database mode
### Description

Implement protocol writer for the new compilation database mode.

TODO

- [x] write binary data to subprocess
- [x] write diagnostic data
- [x] tests

duncanp-sonar:
Fixes #2528


src/CFamily/CMake/CompilationDatabaseEntry.cs
```diff
@@ -25,7 +25,7 @@ namespace SonarLint.VisualStudio.CFamily.CMake
     /// <summary>
     /// Schema based on https://clang.llvm.org/docs/JSONCompilationDatabase.html
     /// </summary>
-    internal class CompilationDatabaseEntry
+    public class CompilationDatabaseEntry
```
duncanp-sonar, line 28: Needs to be accessible from the VSIX project

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        private readonly CompilationDatabaseEntry databaseEntry;
+
+        public CompilationDatabaseRequest(CompilationDatabaseEntry databaseEntry, RequestContext context)
+        {
+            this.databaseEntry = databaseEntry ?? throw new ArgumentNullException(nameof(databaseEntry));
+            Context = context ?? throw new ArgumentNullException(nameof(context));
+
+            // Must have a Command or Arguments but not both
+            var hasArgs = !string.IsNullOrEmpty(databaseEntry.Arguments);
```
duncanp-sonar, line 42: The spec says we should only have one or the other. I guess it depends on what the third-party tools actually produce, and whether they have any bugs. Either way, we might as well validate that we have valid inputs.

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+            WriteQualityProfile(writer);
+
+            // Optional inputs
+            WriteSetting(writer, "PreambleFile", Context.PchFile);
+            WriteSetting(writer, "CreateReproducer", Context?.AnalyzerOptions?.CreateReproducer ?? false ? "true" : "false");
+            WriteSetting(writer, "BuildPreamble", Context?.AnalyzerOptions?.CreatePreCompiledHeaders ?? false ? "true" : "false");
+
+            // TODO - only supplied if analysing a header file
```
duncanp-sonar, line 77: TODO next sprint

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
```diff
@@ -0,0 +1,120 @@
...
+//            WriteSetting(writer, "HeaderFilerLanguage", Context.CFamilyLanguage);
+            WriteRuleSettings(writer);
+
+            Protocol.WriteUTF(writer, "SL-END");
+        }
+
+        public void WriteRequestDiagnostics(TextWriter writer)
+        {
+            var data = JsonConvert.SerializeObject(databaseEntry, Formatting.Indented);
```
duncanp-sonar, line 86: I decided to indent the json to make it more readable. We could do the same for the other reproducer class.

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
```diff
@@ -0,0 +1,120 @@
...
+            }
+        }
+
+        public RequestContext Context { get; }
+
+        public void WriteRequest(BinaryWriter writer)
+        {
+            Protocol.WriteUTF(writer, "SL-IN");
+
+            // Required inputs
```
duncanp-sonar, line 57: Interestingly, we don't specify the language as one of the inputs, except in the case of header files.
So presumably the subprocess must correctly work it out.

However, we still have to pass in the correct set of active rule ids, so we'll need to work it out too. I'll create another ticket for that.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -71,7 +71,7 @@
     <!-- We don't want to copy third-party assemblies from the analyzers in case there are version conflicts e.g. NewtonSoft.Json.dll -->
     <SonarAuthoredAnalyzerAssemblies Include="$(NuGetPackageRoot)SonarAnalyzer.CSharp\$(EmbeddedSonarAnalyzerVersion)\analyzers\Sonar*.dll" />
     <SonarAuthoredAnalyzerAssemblies Include="$(NuGetPackageRoot)SonarAnalyzer.VisualBasic\$(EmbeddedSonarAnalyzerVersion)\analyzers\Sonar*.dll" />
-    <None Include="@(SonarAuthoredAnalyzerAssemblies)" CopyToOutputDirectory="PreserveNewest"/>
+    <None Include="@(SonarAuthoredAnalyzerAssemblies)" CopyToOutputDirectory="PreserveNewest" />
```
duncanp-sonar, line 74: Auto-formatting by VS

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -113,6 +113,7 @@ Referenced assemblies:
 - 'PresentationCore, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
 - 'PresentationFramework, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
 - 'SonarAnalyzer, Version=8.26.0.0, Culture=neutral, PublicKeyToken=c5b62af9de6d7244'
+- 'SonarLint.VisualStudio.CFamily, Version=4.38.0.0, Culture=neutral, PublicKeyToken=c5b62af9de6d7244'
```
duncanp-sonar, line 116: There was a project reference to the CFamily project before, but I guess the code is now actually using something from the assembly.

src/Integration.Vsix.UnitTests/CFamily/CMake/CompilationDatabaseRequestTests.cs
```diff
@@ -0,0 +1,327 @@
...
+            var analyzerOptions = new CFamilyAnalyzerOptions { CreatePreCompiledHeaders = createPch };
+            var context = CreateContext(analyzerOptions: analyzerOptions);
+
+            var tokens = WriteRequest(ValidDbEntry, context);
+
+            CheckExpectedSetting(tokens, "BuildPreamble", expectedValue);
+        }
+
+        [TestMethod]
+        public void WriteReqest_QualityProfile_ExpectedSettingWritten()
```
rita-sonarsource, line 158: There is a [code](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/CFamily/CFamilyHelper.cs#L116) in the vcx that rules can be null when it's a PCH build.
We have a choice here we would we want to put that logic, if in the caller then the Request would need to be able to handle null rules.

rita-sonarsource (approved):

src/Integration.Vsix.UnitTests/CFamily/CMake/CompilationDatabaseRequestTests.cs
duncanp-sonar: Good point. I'll do the same here so we don't write the setting - every millisecond counts...

PR #2535, opened 2021-08-10
Add RequestFactory

### LINKED ISSUE #2194: Refactor CreateRequest to support different project configurations (wrap FileConfig with vcx request creator)
#### Description

`CFamilyHelper.CreateRequest` is written for vcx projects and as such it makes assumptions on the ways to collect the needed configuration:
* That the file is part of a project in a solution
* That the configuration can be collected via VCProjectEngine framework.
* The created `Request` object represents the needed analysis configuration in detailed properties.

This does not work for CMake projects, where:
* The file is not part of a project/solution
* The file cannot be converted to VCProjectEngine framework objects
* The created `Request` object should return only the location of the compilation database file (? TBD)

 TODO:
- [ ] Refactor `CreateRequest` to support different configurations
- [ ] Check if global analysis logic, such as PCH and CreateReproducer, should apply to all configuration types, or only vcx projects.

rita-sonarsource:
Fixes #2194


src/Integration.Vsix/CFamily/VcxRequestFactory.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            var projectItem = dte?.Solution?.FindProjectItem(analyzedFilePath);
+
+            if (projectItem == null)
+            {
+                return null;
+            }
+
+            // todo: extract the code from CFamilyHelper into this class
+            return CFamilyHelper.CreateRequest(logger,
```
rita-sonarsource, line 57: This bit is not tested yet -- I added the todo to extract code from CFamilyHelper, along with the tests

duncanp-sonar (approved):

PR #2536, opened 2021-08-10
Added CompilationDatabaseRequestFactory

### LINKED ISSUE #2534: Implement CMake request factory
### Description

Implement factory code for CMake requests

TODO
- [ ] implement new factory
- [ ] tests

duncanp-sonar:
Fixes #2534

### DIFF
--- a/src/Core/CFamily/CFamilyShared.cs
+++ b/src/Core/CFamily/CFamilyShared.cs
@@ -31,5 +31,28 @@ namespace SonarLint.VisualStudio.Core.CFamily
 
         public static readonly StringComparer RuleKeyComparer = SonarRuleRepoKeys.RepoKeyComparer;
         public static readonly StringComparison RuleKeyComparison = StringComparison.Ordinal;
+
+        /// <summary>
+        /// Attempts to detect whether the file is C or C++ based on the file extension.
+        /// Returns null if the extension is not recognised.
+        /// </summary>
+         public static string FindLanguageFromExtension(string analyzedFilePath)
+        {
+            string cfamilyLanguage = null;
+
+            // Compile files with extensions ".cpp", ".cxx" and ".cc" as Cpp and files with extension ".c" as C
+            if (analyzedFilePath.EndsWith(".cpp", StringComparison.OrdinalIgnoreCase) ||
+                analyzedFilePath.EndsWith(".cxx", StringComparison.OrdinalIgnoreCase) ||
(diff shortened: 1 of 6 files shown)


src/Core/CFamily/CFamilyShared.cs
```diff
@@ -31,5 +31,28 @@ public static class CFamilyShared
 
         public static readonly StringComparer RuleKeyComparer = SonarRuleRepoKeys.RepoKeyComparer;
         public static readonly StringComparison RuleKeyComparison = StringComparison.Ordinal;
+
+        /// <summary>
+        /// Attempts to detect whether the file is C or C++ based on the file extension.
+        /// Returns null if the extension is not recognised.
+        /// </summary>
+         public static string FindLanguageFromExtension(string analyzedFilePath)
```
duncanp-sonar, line 39: I extracted this code from `CFamilyHelper.Capture` class. The only change is to use an `OrdinalIgnoreCase` string comparer, rather than forcing every string to lower case.

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequestFactory.cs
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.CFamily.CMake;
+using SonarLint.VisualStudio.Core.CFamily;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.CMake
+{
+    [Export(typeof(IRequestFactory))]
+    internal class CompilationDatabaseRequestFactory : IRequestFactory
```
rita-sonarsource, line 28: Does this need to be in the vsix proj? I don't see that it references something from Vsix

rita-sonarsource (approved):

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequestFactory.cs
duncanp-sonar: It doesn't, but the request type it creates does (depend on `Protocol`). It seemed simpler to leave the request and factory together.

rita-sonarsource: Ah right, thanks

PR #2538, opened 2021-08-11
Extract code from CFamilyHelper into FileConfigProvider

rita-sonarsource:
Part of #2537

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/FileConfigProviderTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/FileConfigProviderTests.cs
@@ -1,0 +1,147 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.VCProjectEngine;
+using Moq;
+using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+using static SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests.CFamilyTestUtility;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
+{
+    [TestClass]
+    public class FileConfigProviderTests
+    {
+        [TestMethod]
+        public void Get_FileIsNotInSolution_ReturnsNull()
+        {
+            var projectItemMock = CreateMockProjectItem("c:\\foo\\SingleFileISense\\xxx.vcxproj");
+           
+            var testSubject = CreateTestSubject();
+            var result = testSubject.Get(projectItemMock.Object, "c:\\dummy", new CFamilyAnalyzerOptions());
+
+            result.Should().BeNull();
+            projectItemMock.Verify(x => x.ContainingProject, Times.Once);
+        }
+
+        [TestMethod]
+        public void Get_FailureToCheckIfFileIsInSolution_NonCriticalException_ExceptionCaughtAndNullReturned()
+        {
+            var projectItemMock = CreateMockProjectItem("c:\\foo\\SingleFileISense\\xxx.vcxproj");
+            projectItemMock.Setup(x => x.ContainingProject).Throws<NotImplementedException>();
+
+            var testSubject = CreateTestSubject();
+            var result = testSubject.Get(projectItemMock.Object, "c:\\dummy", new CFamilyAnalyzerOptions());
+
+            result.Should().BeNull();
+            projectItemMock.Verify(x => x.ContainingProject, Times.Once);
+        }
+
+        [TestMethod]
+        public void Get_FailureToCheckIfFileIsInSolution_CriticalException_ExceptionThrown()
+        {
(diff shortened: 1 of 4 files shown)


src/Integration.Vsix/CFamily/FileConfigProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+
+            try
+            {
+                // Note: if the C++ tools are not installed then it's likely an exception will be thrown when
+                // the framework tries to JIT-compile the TryGet method (since it won't be able to find the MS.VS.VCProjectEngine
+                // types).
+                return CFamilyHelper.FileConfig.TryGet(analysisLogger, projectItem, analyzedFilePath);
+            }
+            catch (Exception ex) when (!Microsoft.VisualStudio.ErrorHandler.IsCriticalException(ex))
```
rita-sonarsource, line 61: The [original](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix/CFamily/CFamilyHelper.cs#L228) code didn't handle critical exceptions but I've decided to add it here, like we do for all try-catch cases

src/Integration.Vsix/CFamily/FileConfigProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+                       projectItem.ConfigurationManager.ActiveConfiguration != null;
+            }
+            catch (Exception ex) when (!Microsoft.VisualStudio.ErrorHandler.IsCriticalException(ex))
+            {
+                // Suppress non-critical exceptions
+            }
+            return false;
+        }
+
+        private class NoOpLogger : ILogger
```
rita-sonarsource, line 105: CFamilyHelper actually used a logger (and the fake one) only for the file config retrieval, so it's handy that we can now hide this hacky bit.

src/Integration.Vsix/CFamily/FileConfigProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+
+        public CFamilyHelper.FileConfig Get(ProjectItem projectItem, string analyzedFilePath, CFamilyAnalyzerOptions analyzerOptions)
+        {
+            var analysisLogger = GetAnalysisLogger(analyzerOptions);
+
+            if (!IsFileInSolution(projectItem))
+            {
+                analysisLogger.LogDebug($"[CFamilyHelper] Unable to retrieve the configuration for file '{analyzedFilePath}'. Check the file is part of a project in the current solution.", analyzerOptions);
```
rita-sonarsource, line 50: Unfortunately it's a LogDebug so I can't verify it in unit tests on the CI. Instead I check that the relevant mocks were invoked.

src/Integration.Vsix.UnitTests/CFamily/FileConfigProviderTests.cs
```diff
@@ -0,0 +1,147 @@
...
+using Microsoft.VisualStudio.VCProjectEngine;
+using Moq;
+using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+using static SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests.CFamilyTestUtility;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
+{
+    [TestClass]
+    public class FileConfigProviderTests
```
rita-sonarsource, line 33: There weren't a lot of tests in CFamilyHelper so I just wrote new ones.
I took [this one](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelperTests.cs#L73) and wrote similar new ones.

src/Integration.Vsix.UnitTests/CFamily/FileConfigProviderTests.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        [TestMethod]
+        public void Get_SuccessfulConfig_ConfigReturned()
+        {
+            var projectItemMock = CreateMockProjectItem("c:\\foo\\xxx.vcxproj");
+            
+            var testSubject = CreateTestSubject();
+            var result = testSubject.Get(projectItemMock.Object, "c:\\dummy", new CFamilyAnalyzerOptions());
+
+            result.Should().NotBeNull();
```
rita-sonarsource, line 137: [FileConfigTests](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Integration.Vsix.UnitTests/CFamily/FileConfigTests.cs) has more dedicated tests

src/Integration.Vsix/CFamily/FileConfigProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+using System;
+using EnvDTE;
+using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Integration.Helpers;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
+{
+    internal interface IFileConfigProvider
```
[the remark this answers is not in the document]
duncanp-sonar: Currently `FileConfig` is nested inside `CFamilyHelper`. Are you going to nest it inside this class instead?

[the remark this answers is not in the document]
rita-sonarsource: I'm not sure yet. I thought about removing the nesting altogether, but nesting here is an option too.

src/Integration.Vsix/CFamily/FileConfigProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+                return null;
+            }
+        }
+
+        private ILogger GetAnalysisLogger(CFamilyAnalyzerOptions analyzerOptions)
+        {
+            if (analyzerOptions is CFamilyAnalyzerOptions cFamilyAnalyzerOptions &&
+                cFamilyAnalyzerOptions.CreatePreCompiledHeaders)
+            {
+                // In case the requeset is coming from PCH generation, we don't log failures.
```
duncanp-sonar, line 73: Pre-existing typo

src/Integration.Vsix/CFamily/FileConfigProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+                // This is to avoid redundant messages while navigating unsupported files.
```
duncanp-sonar, line 74: We don't currently do this for CMake. However, I think we'll only log debug messages saying the compilation database can't be located, so that's probably ok.

src/Integration.Vsix/CFamily/FileConfigProvider.cs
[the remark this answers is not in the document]
duncanp-sonar: The nesting was partly to hide a (messy) implementation detail, and also to make the intended scope of usage of the type clearer.

An alternative would be to create a new `VcxProject` namespace and put the request and supporting classes in it so they don't pollute the more general `CFamily` namespace.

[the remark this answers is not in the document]
rita-sonarsource: I like the namespace approach, will do that 👍

PR #2543, opened 2021-08-11
Fix debug test failures

duncanp-sonar:
Mea culpa - I didn't run all of the tests in debug after fixing the assertion to use `PathHelper`. Two of the tests failed in debug due to invalid paths, but of course this wasn't picked up by the CI.

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
@@ -99,7 +99,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         [TestMethod]
         public void TriggerAnalysisAsync_StreamsIssuesFromSubProcessToConsumer()
         {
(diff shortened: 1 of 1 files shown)


src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -99,7 +99,7 @@ public void ExecuteAnalysis_RequestCanBeCreated_AnalysisIsTriggered()
         [TestMethod]
         public void TriggerAnalysisAsync_StreamsIssuesFromSubProcessToConsumer()
         {
-            const string fileName = "c:\\data\\aaa\\bbb\file.txt";
+            const string fileName = "c:\\data\\aaa\\bbb\\file.txt";
```
duncanp-sonar, line 102: Incorrect escaping in the path - missing \.

It didn't matter before when the Debug.Assert was comparing strings.
However, `PathHelper` uses `Path.GetCanonicalPath`, which validates the characters -> exception.

Issue #2539, opened 2021-08-11
Handle protocol-specific environment variables


duncanp-sonar:
### Description

The new subprocess protocol expects certain environment variables to be set, so the request-handling infrastructure needs to be updated to allow a request to specify environment variables.

TODO
- [x] add EnvironmentVariables to IRequest
- [x] modify CLangAnalyzer to pass the environment variables to the process runner
- [x] unit tests

PR #2544, opened 2021-08-11
Added EnvironmentVariables to IRequest

duncanp-sonar:
Part of #2539

### DIFF
--- a/src/Core/CFamily/IRequest.cs
+++ b/src/Core/CFamily/IRequest.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.Collections.Generic;
 using System.IO;
 
 namespace SonarLint.VisualStudio.Core.CFamily
@@ -34,6 +35,11 @@ namespace SonarLint.VisualStudio.Core.CFamily
         /// Serializes the request in the form required by the subprocess
         /// </summary>
         void WriteRequest(BinaryWriter writer);
+
+        /// <summary>
+        /// Any environment variables that need to be passed to the subprocess. Can be null.
+        /// </summary>
+        IDictionary<string, string> EnvironmentVariables { get; }
 
         /// <summary>
         /// Serializes the request for diagnostic purposes
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamily_CallAnalyzerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamily_CallAnalyzerTests.cs
@@ -49,7 +49,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
             // Assert
             dummyProcessRunner.ExecuteCalled.Should().BeTrue();
 
-            response.Count().Should().Be(1);
+            response.Count.Should().Be(1);
             response[0].Filename.Should().Be("file.cpp");
         }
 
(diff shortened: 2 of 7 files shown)


duncanp-sonar:
This is a less hacky way of handling the new protocol requirement that the INCLUDE setting should be passed as an environment variable.

It doesn't necessary completely fix the issue, since the INCLUDE value is hard-coded to empty. However, it contains the necessary API changes to `IRequest` and fixes up `CLangAnalyzer`.

src/Core/CFamily/IRequest.cs
```diff
@@ -35,6 +36,11 @@ public interface IRequest
         /// </summary>
         void WriteRequest(BinaryWriter writer);
 
+        /// <summary>
+        /// Any environment variables that need to be passed to the subprocess. Can be null.
+        /// </summary>
+        IDictionary<string, string> EnvironmentVariables { get; }
```
duncanp-sonar, line 42: The new protocol requires some inputs to be passed via environment variables, so I've added a dictionary to the request interface.

src/Integration.Vsix.UnitTests/CFamily/CFamily_CallAnalyzerTests.cs
```diff
@@ -49,7 +49,7 @@ public void CallAnalyzer_Succeeds_ReturnsMessages()
             // Assert
             dummyProcessRunner.ExecuteCalled.Should().BeTrue();
 
-            response.Count().Should().Be(1);
+            response.Count.Should().Be(1);
```
duncanp-sonar, line 52: Miscellaneous: no need to use LINQ since there is a `Count` property

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -222,6 +222,7 @@ private void HandleMessage(Message message, IRequest request, IIssueConsumer con
                 CmdLineArgs = new[] { communicateViaStreaming },
                 CancellationToken = cancellationToken,
                 WorkingDirectory = SubProcessFilePaths.WorkingDirectory,
+                EnvironmentVariables = request.EnvironmentVariables,
```
duncanp-sonar, line 225: Fortunately, the `ProcessRunnerArguments` already supported environment variables, even though we weren't using them (we copied the runner and arguments from the Scanner for MSBuild, which does pass environment variables, and fortunately we didn't remove the property).

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
```diff
@@ -50,6 +51,18 @@ public CompilationDatabaseRequest(CompilationDatabaseEntry databaseEntry, Reques
...
         public RequestContext Context { get; }
 
+        public IDictionary<string, string> EnvironmentVariables
+        {
+            get
+            {
+                return new Dictionary<string, string>
+                {
+                    // TODO - check whether a value is required. #2539
+                    { "INCLUDE", string.Empty }
```
duncanp-sonar, line 61: We'll need to work out what we need to pass, if anything, and how to calculate it.

The INCLUDE environment variable is set from the VS command line, but it is not set in VS. That either means it isn't needed for CMake builds in the IDE, or that the data is being passed in some other way. I think we'll need to discuss it with the CFamily team.

rita-sonarsource (approved):

PR #2545, opened 2021-08-12
Add RulesConfigProtocolFormatter

rita-sonarsource:
Part of #2537

### DIFF
--- a/src/CFamily/Analysis/RulesConfigProtocolFormatter.cs
+++ b/src/CFamily/Analysis/RulesConfigProtocolFormatter.cs
@@ -67,7 +67,7 @@ namespace SonarLint.VisualStudio.CFamily.Analysis
         {
             var ruleParameters = new Dictionary<string, string>();
             
-            foreach (var ruleKey in rulesConfiguration.AllPartialRuleKeys)
+            foreach (var ruleKey in rulesConfiguration.ActivePartialRuleKeys)
             {
                 if (rulesConfiguration.RulesParameters.TryGetValue(ruleKey, out var ruleParams))
                 {
--- a/src/CFamily/Analysis/RulesConfigProtocolFormatter.cs
+++ b/src/CFamily/Analysis/RulesConfigProtocolFormatter.cs
@@ -1,0 +1,85 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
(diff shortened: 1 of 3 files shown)


src/CFamily/Analysis/RulesConfigProtocolFormatter.cs
```diff
@@ -0,0 +1,85 @@
...
+        public Dictionary<string, string> RuleParameters { get; }
+
+        public RuleConfigProtocolFormat(string qualityProfile, Dictionary<string, string> ruleParameters)
+        {
+            QualityProfile = qualityProfile;
+            RuleParameters = ruleParameters;
+        }
+    }
+
+    public class RulesConfigProtocolFormatter : IRulesConfigProtocolFormatter
```
rita-sonarsource, line 51: The code is shared between cmake and vcx, so I created a helper class.

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
```diff
@@ -89,11 +99,29 @@ public void WriteRequest(BinaryWriter writer)
...
+            if (!Context.AnalyzerOptions?.CreatePreCompiledHeaders ?? true)
+            {
+                var rulesProtocolFormat = rulesConfigProtocolFormatter.Format(Context.RulesConfiguration);
+
+                WriteQualityProfile(writer, rulesProtocolFormat.QualityProfile);
+                WriteRuleSettings(writer, rulesProtocolFormat.RuleParameters);
+            }
+            else
+            {
+                WriteQualityProfile(writer, "");
```
rita-sonarsource, line 121: I believe we don't need this, I tested the pch generation and it seems fine without it. Will ask the cfamily.

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
```diff
@@ -33,10 +34,19 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.CMake
     internal class CompilationDatabaseRequest : IRequest
```
rita-sonarsource, line 34: I haven't changed the tests yet, so they're using the inner class as an integration.
I will be done with the vcx request's tests and then adopt CompilationDatabaseRequestTests.

src/CFamily/Analysis/RulesConfigProtocolFormatter.cs
```diff
@@ -0,0 +1,85 @@
...
+            var ruleParameters = GetRuleParameters(rulesConfig);
+
+            return new RuleConfigProtocolFormat(qualityProfile, ruleParameters);
+        }
+
+        private static Dictionary<string, string> GetRuleParameters(ICFamilyRulesConfig rulesConfiguration)
+        {
+            var ruleParameters = new Dictionary<string, string>();
+            
+            foreach (var ruleKey in rulesConfiguration.AllPartialRuleKeys)
```
duncanp-sonar, line 70: If I remember correctly, the VCX version used `AllPartialRuleKeys`, and in the CMake version I used `AllActiveRuleKeys`. There didn't seem to be much point in passing parameters for rules that won't be executed.

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
duncanp-sonar: It's listed as an essential field in the CMake integration spec, and their code suggests they will error without it:
```
llvm::Optional<DriverError> MSVCDriver::setCompilationIndependentOpt(AnalyzerInput& analyzerInput) const {
...
  if (auto errorOrOpt = getOptionOrError("QualityProfile")) {
...
}
```

It's possible they are returning an error message and we are ignoring it.

duncanp-sonar (approved):

src/CFamily/Analysis/RulesConfigProtocolFormatter.cs
rita-sonarsource: Good point, I'll align that.

PR #2549, opened 2021-08-12
Extract CFamilyHelper code into VcxRequestFactory

### LINKED ISSUE #2537: Refactor and cleanup vcx Request

rita-sonarsource:
Fixes #2537


src/Integration.Vsix/CFamily/VcxProject/IFileConfig.cs
```diff
@@ -0,0 +1,56 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject
+{
+    internal interface IFileConfig
```
rita-sonarsource, line 23: Created an interface for ease of mocking in VcxRequestFactoryTests; otherwise it was difficult to check that certain properties (namely AbsoluteFilePath) were indeed checked.

src/Integration.Vsix/CFamily/VcxProject/VcxRequestFactory.cs
```diff
@@ -0,0 +1,161 @@
...
+            this.rulesConfigProtocolFormatter = rulesConfigProtocolFormatter;
+            this.fileConfigProvider = fileConfigProvider;
+            this.logger = logger;
+        }
+
+        public IRequest TryGet(string analyzedFilePath, CFamilyAnalyzerOptions analyzerOptions)
+        {
+            try
+            {
+                var projectItem = dte?.Solution?.FindProjectItem(analyzedFilePath);
```
rita-sonarsource, line 71: I copied the code as-is from CFamilyHelper but with a few changes:
1. Using FileConfigProvider and RulesConfigProtocolFormatter
2. Wrapping it with a try-catch, so the top-level aggregate factory wouldn't fail when iterating over the factories
3. I removed the ` if (request?.File == null)` condition -- it's not possible for this to happen without an exception and it wasn't tested in the original CFamilyHelperTests.
4. We no longer save the file config here, as it's now part of the caller CLangAnalyzer
5. Technically the try-catch will use the real logger, not the fake NoOp -- but this is ok, because there was no try-catch in the CFamilyHelper to begin with, so any exceptions would've propagated to CLangAnalyzer and written with a real logger.

src/Integration.Vsix.UnitTests/CFamily/VcxRequestFactoryTests.cs
```diff
@@ -76,5 +317,66 @@ private static Mock<IServiceProvider> CreateServiceProviderReturningProjectItem(
...
+                rulesConfigProtocolFormatter.Object);
+        
+            return testSubject.TryGet(fileToAnalyze, analyzerOptions) as Request;
+        }
+
+        private Mock<IFileConfig> CreateDummyFileConfig(string filePath)
+        {
+            var fileConfig = new Mock<IFileConfig>();
+
+            fileConfig.SetupGet(x => x.PlatformName).Returns("Win32");
```
rita-sonarsource, line 360: All of these are unfortunately "mandatory" properties in the Captures logic. Empty strings are acceptable but nulls are not...

src/Integration.Vsix.UnitTests/CFamily/VcxRequestFactoryTests.cs
```diff
@@ -19,19 +19,27 @@
...
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.CFamily;
 using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+using SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject;
+using static SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests.CFamilyTestUtility;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
 {
     [TestClass]
     public class VcxRequestFactoryTests
```
rita-sonarsource, line 38: I copied some of the tests from CFamilyHelperTests, including the method GetSuccessfulRequest.
However, I had to modify things so I think it's better to treat all of them as "new".

src/Integration.Vsix.UnitTests/CFamily/VcxRequestFactoryTests.cs
```diff
@@ -53,14 +61,247 @@ public void TryGet_NoProjectItem_Null()
...
+        }
+
+        [TestMethod]
+        public void TryGet_HeaderFile_IsSupported()
+        {
+            var projectItemConfig = new ProjectItemConfig { ItemType = "ClInclude" };
+            var projectItemMock = CreateMockProjectItem("c:\\foo\\xxx.vcxproj", projectItemConfig);
+
+            var fileConfig = CreateDummyFileConfig("c:\\dummy\\file.h");
+            fileConfig.Setup(x => x.CompileAs).Returns("CompileAsCpp");
```
rita-sonarsource, line 188: This was hidden in the original tests -- if you don't specify CompileAsCpp (or as C), we won't detect a language and would not analyze the header file. I guess it's ok because you need to do it for header files anyway. At least now it's explicit in the test.

src/Integration.Vsix/CFamily/VcxProject/FileConfig.cs
```diff
@@ -0,0 +1,283 @@
+﻿/*
```
rita-sonarsource, line 1: SC is complaining about a whitespace difference -- I fixed it and will push along with PR fixes.

src/Integration.Vsix/CFamily/CFamilyStrings.resx
```diff
@@ -195,4 +195,7 @@
   <data name="ERROR_InvalidCompilationEntry" xml:space="preserve">
     <value>Invalid compilation database entry. Either Command or Arguments must be supplied, but not both.</value>
   </data>
+  <data name="ERROR_CreatingVcxRequest" xml:space="preserve">
```
rita-sonarsource, line 198: There is a missing dot here; I fixed it and will push along with the rest of the review comments.

rita-sonarsource:
@duncanp-sonar please review, it might help to look at each commit separately. I apologize for the overall number of changes, I didn't expect the final outcome to span so many files.

src/Integration.Vsix/CFamily/VcxProject/IFileConfig.cs
duncanp-sonar: Let's start with them separate and see what happens if/when a new field gets added.

src/Integration.Vsix/CFamily/VcxProject/VcxRequestFactory.cs
duncanp-sonar: (4) was done already, wasn't it? Has anything else changed about saving the config in this PR?


```
  5. Technically the try-catch will use the real logger, not the fake NoOp -- but this is ok, because there was no try-catch in the CFamilyHelper to begin with, so any exceptions would've propagated to CLangAnalyzer and written with a real logger.
```

Both `CFamilyHelper.IsFileInSolution` and `CFamilyHelper.TryGetConfig` had exception handling before which would have used the no-op logger. 

I thought that the no-op logger was added because the logged output was very verbose for unsupported project types, and lots of messages got logged for PCH generation (although with the addition of CMake support this should happen less often).

I guess that is the real question - what now gets logged for PCH file generation for unsupported files?

rita-sonarsource: IsFileInSolution and TryGetConfig are now both part of FileConfigProvider, which has the NoOp logger. So that behavior remains the same -- no verbose output for PCH.

rita-sonarsource: (4) no, this was just to clarify that the code existed in CFamilyHelper and was not ported.

duncanp-sonar: Ok. I found the comment confusing because it's about something that was done previously, not something that changed in this PR.

duncanp-sonar (approved):

PR #2554, opened 2021-08-12
Fix PCH verbose output log

### LINKED ISSUE #2553: Handle verbose PCH output in CLangAnalyzer
#### Description

CLangAnalyzer is using the new AggregateRequestFactory. When the factory returns a null request, we write an error to the output log:
```
Unable to retrieve the configuration for file '{0}'.
    Check the file is part of a supported project type in the current solution.
```

However, this message is logged for PCH as well and it's noisy.

TODO:
- [x] Don't log the message for PCH.

rita-sonarsource:
Fixes #2553

### DIFF
--- a/src/Integration.Vsix/CFamily/CLangAnalyzer.cs
+++ b/src/Integration.Vsix/CFamily/CLangAnalyzer.cs
@@ -110,11 +110,15 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
         {
             Debug.Assert(IsAnalysisSupported(detectedLanguages));
 
-            var request = requestFactory.TryGet(path, analyzerOptions as CFamilyAnalyzerOptions);
+            var cFamilyAnalyzerOptions = analyzerOptions as CFamilyAnalyzerOptions;
+            var request = requestFactory.TryGet(path, cFamilyAnalyzerOptions);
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -110,11 +110,15 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
             Debug.Assert(IsAnalysisSupported(detectedLanguages));
 
-            var request = requestFactory.TryGet(path, analyzerOptions as CFamilyAnalyzerOptions);
+            var cFamilyAnalyzerOptions = analyzerOptions as CFamilyAnalyzerOptions;
+            var request = requestFactory.TryGet(path, cFamilyAnalyzerOptions);
 
             if (request == null)
             {
-                logger.WriteLine(CFamilyStrings.MSG_UnableToCreateConfig, path);
+                if (cFamilyAnalyzerOptions == null || !cFamilyAnalyzerOptions.CreatePreCompiledHeaders)
```
duncanp-sonar, line 118: I think a comment would help e.g. `Logging for PCH is too noisy: #2553`

PR #2559, opened 2021-08-16
Support CMake header files analysis

### LINKED ISSUE #2550: Implement header file db entry location algorithm
### Description

Implement header file db entry location algorithm

### TODO
- [x] implement matching algorithm 
- [x] add debug output to show how long the search takes
- [x] unit tests

### File matching rules
The proposed algorithm is as follows:

(A) If code file:
- 1) first matching [path]\[filename with extension]
No fallback if not found

(B) If header file:
- 1) first file matching [path]\[filename without extension], else
- 2) first file matching [path], else
- 3) first item

Currently all of the matching logic is in `CompilationConfigProvider`.
The search logic for (A) has already been implemented.

### Out of scope
The initial implementation will not including caching.
The time taken to execute the matching process will debug-logged to help assess the impact.

rita-sonarsource:
Fixes #2550

### DIFF
--- a/src/CFamily/CMake/CompilationConfigProvider.cs
+++ b/src/CFamily/CMake/CompilationConfigProvider.cs
@@ -21,10 +21,13 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
+using System.Diagnostics;
+using System.IO;
 using System.IO.Abstractions;
 using System.Linq;
 using Microsoft.VisualStudio;
 using Newtonsoft.Json;
+using SonarLint.VisualStudio.Core.CFamily;
 using SonarLint.VisualStudio.Core.Helpers;
 using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.Integration.Helpers;
@@ -93,12 +96,19 @@ namespace SonarLint.VisualStudio.CFamily.CMake
                 var compilationDatabaseString = fileSystem.File.ReadAllText(compilationDatabaseLocation);
                 var compilationDatabaseEntries = JsonConvert.DeserializeObject<IEnumerable<CompilationDatabaseEntry>>(compilationDatabaseString);
 
-                var entry = compilationDatabaseEntries?.FirstOrDefault(x => !string.IsNullOrEmpty(x.File) && PathHelper.IsMatchingPath(filePath, x.File));
-
-                if (entry == null)
+                if (compilationDatabaseEntries == null || !compilationDatabaseEntries.Any())
                 {
-                    logger.WriteLine(Resources.NoCompilationDatabaseEntry, filePath);
+                    logger.WriteLine(Resources.EmptyCompilationDatabaseFile, compilationDatabaseLocation);
+                    return null;
                 }
+
+                var stopwatch = Stopwatch.StartNew();
+
+                var entry = IsHeaderFile(filePath)
+                    ? LocateMatchingCodeEntry(filePath, compilationDatabaseEntries)
+                    : LocateExactCodeEntry(filePath, compilationDatabaseEntries);
+
+                logger.WriteLine("***** [CompilationConfigProvider] time (ms) to locate entry: " + stopwatch.ElapsedMilliseconds);
 
                 return entry;
             }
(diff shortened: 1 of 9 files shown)


src/Integration.Vsix.UnitTests/CFamily/CMake/CompilationDatabaseRequestFactoryTests.cs
```diff
@@ -57,11 +58,33 @@ public void TryGet_NoConfig_ReturnsNull()
...
+            // However, the compilation entry is a ".c" file, so we expect the code to calculate the rules based on the entry.
+            rulesConfigProvider.Verify(x=> x.GetRulesConfiguration(SonarLanguageKeys.C), Times.Once());
+        }
+
         [TestMethod]
         public void TryGet_UnrecognizedLanguage_ReturnsNull()
         {
             const string fileName = "c:\\file.txt";
-            var compilationConfigProvider = CreateCompilationProvider(fileName, new CompilationDatabaseEntry());
+
```
rita-sonarsource, line 85: Changed existing tests to create a valid CreateCompilationDatabaseEntry (with a file name)

src/CFamily/CMake/CompilationConfigProvider.cs
```diff
@@ -109,5 +120,97 @@ public CompilationDatabaseEntry GetConfig(string filePath)
...
+
+            if (entry == null)
+            {
+                logger.WriteLine(Resources.NoCompilationDatabaseEntry, filePath);
+            }
+
+            return entry;
+        }
+
+        private CompilationDatabaseEntry LocateMatchingCodeEntry(string headerFilePath, IEnumerable<CompilationDatabaseEntry> compilationDatabaseEntries)
```
rita-sonarsource, line 141: I can't think of a nicer way to structure this block 🙈

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequestFactory.cs
```diff
@@ -47,7 +47,7 @@ public IRequest TryGet(string analyzedFilePath, CFamilyAnalyzerOptions analyzerO
             }
 
             // TODO - handle user specifying the language via a command / argument #2533
-            var languageKey = CFamilyShared.FindLanguageFromExtension(analyzedFilePath);
+            var languageKey = CFamilyShared.FindLanguageFromExtension(dbEntry.File);
```
rita-sonarsource, line 50: We need to take the extension from the code entry, not the header file. Added an explicit test for this.

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
```diff
@@ -78,7 +78,7 @@ public void WriteRequest(BinaryWriter writer)
             WriteHeader(writer);
 
             // Required inputs
-            WriteSetting(writer, "File", databaseEntry.File);
+            WriteSetting(writer, "File", Context.File);
```
rita-sonarsource, line 81: We need to write the analyzed header file, not the code entry -- the code entry is used only for the build parameters. I added a test for this.

src/Integration.Vsix.UnitTests/CFamily/CMake/CompilationDatabaseRequestTests.cs
```diff
@@ -118,7 +130,7 @@ public void WriteReqest_WithCommand_ExpectedSettingWritten()
         }
 
         [TestMethod]
-        public void WriteReqest_WithArguments_ExpectedSettingWritten()
+        public void WriteRequest_WithArguments_ExpectedSettingWritten()
```
rita-sonarsource, line 133: Cosmetics, renaming Reqest to Request

src/CFamily/CMake/CompilationConfigProvider.cs
duncanp-sonar: One option would be to push more logic into each "Locate" method, so they handle the loop and the "not found" logging, returning null if they didn't find anything.

That would simplify the call to something like:
```
  var match = LocateCodeEntry(codeFile, compilationDatabaseEntries) ??
    LocateFirstCodeEntryWithSameName(codeFile, compilationDatabaseEntries) ??
    LocateFirstCodeEntryUnderRoot(codeFile, compilationDatabaseEntries) ??
    LocateFirstCodeEntry(compilationDatabaseEntries);
```

You'd probably also want to pass `codeFilesWithSameNameAndSamePath` into the "Locate" methods so they didn't have to recalculate it each time.

I'm not sure it's a big improvement, though.

duncanp-sonar (approved):

PR #2561, opened 2021-08-17
Added service to return info about the VS installation

duncanp-sonar:
Prep for ##2560

### DIFF
--- a/src/Infrastructure.VS/VsInfoService.cs
+++ b/src/Infrastructure.VS/VsInfoService.cs
@@ -1,0 +1,71 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.Infrastructure.VS
+{
+    /// <summary>
+    /// Returns information about the current VS installation
+    /// </summary>
+    public interface IVsInfoService
+    {
+        /// <summary>
+        /// Returns the root directory for executing VS installation
+        /// </summary>
+        string InstallRootDir { get; }
+    }
+
+    [Export(typeof(IVsInfoService))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class VsInfoService : IVsInfoService
+    {
+        private readonly ILogger logger;
+
+        [ImportingConstructor]
+        public VsInfoService([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider, ILogger logger)
+        {
+            this.logger = logger;
+            InstallRootDir = GetInstallRootDir(serviceProvider);
+        }
+
+        public string InstallRootDir { get; }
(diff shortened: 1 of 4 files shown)


src/Infrastructure.VS/VsInfoService.cs
```diff
@@ -0,0 +1,71 @@
...
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.Infrastructure.VS
+{
+    /// <summary>
+    /// Returns information about the current VS installation
+    /// </summary>
+    public interface IVsInfoService
```
duncanp-sonar, line 33: We fetch info about the VS installation in a number of places. I've created this interface to provide an abstraction so the consumers don't need to know about `IServiceProvider` or `IVsShell`.

We could refactor some of the existing code to use this service at some point. I'll create a separate ticket for that.

src/Infrastructure.VS/VsInfoService.cs
```diff
@@ -0,0 +1,71 @@
...
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class VsInfoService : IVsInfoService
+    {
+        private readonly ILogger logger;
+
+        [ImportingConstructor]
+        public VsInfoService([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider, ILogger logger)
+        {
+            this.logger = logger;
+            InstallRootDir = GetInstallRootDir(serviceProvider);
```
duncanp-sonar, line 51: This follows the same pattern as the `VsVersionProvider` i.e. fetching data on construction.

src/Infrastructure.VS/VsInfoService.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public string InstallRootDir { get; }
+
+        private string GetInstallRootDir(IServiceProvider serviceProvider)
+        {
+            IVsShell shell = serviceProvider.GetService(typeof(SVsShell)) as IVsShell;
```
duncanp-sonar, line 58: I didn't bother handling the case where we can't fetch `IVsShell` - there would have to be something seriously wrong with VS for that to happen in practice, and it isn't something we could really recover from.

src/Infrastructure.VS/VsInfoService.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            object value;
+            int hr = shell.GetProperty((int)__VSSPROPID2.VSSPROPID_InstallRootDir, out value);
+            if (ErrorHandler.Succeeded(hr))
+            {
+                return value as string;
+            }
+
+            logger.WriteLine(Resources.FailedToFetchVsInstallRootDir, hr);
```
duncanp-sonar, line 67: I'm not sure there's much point in logging failure here. Still, it's consistent with what we do elsewhere.

src/Infrastructure.VS/VsInfoService.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public string InstallRootDir { get; }
+
+        private string GetInstallRootDir(IServiceProvider serviceProvider)
+        {
+            IVsShell shell = serviceProvider.GetService(typeof(SVsShell)) as IVsShell;
+
+            object value;
+            int hr = shell.GetProperty((int)__VSSPROPID2.VSSPROPID_InstallRootDir, out value);
```
rita-sonarsource, line 61: In VsVersionProvider we use `__VSSPROPID.VSSPROPID_InstallDirectory`.
They seem to be the same though
[VSSPROPID_InstallDirectory](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.__vsspropid?view=visualstudiosdk-2019): `BSTR. Directory where visual studio executable was installed.`
[VSSPROPID_InstallRootDir](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.__vsspropid2?view=visualstudiosdk-2019): `BSTR: full path of root location of installation (for example, drive>:\Program Files\Microsoft Visual Studio <version>).`
I wonder if there is any difference between different VS versions? The doc suggests that no, so we could revisit it in the future when we use it in VsVersionProvider. Just something to remember to test.

src/Infrastructure.VS/VsInfoService.cs
rita-sonarsource: We could consider throwing here. The reason that we catch in VsVersionProvider is because that whole block fails in 2015, and we needed the version for telemetry. So in 2015 we just don't send the version.
So here we could simplify and just not catch anything.

rita-sonarsource (approved):

src/Infrastructure.VS/VsInfoService.cs
duncanp-sonar: I'll change it to
```
            ErrorHandler.ThrowOnFailure(shell.GetProperty((int)__VSSPROPID2.VSSPROPID_InstallRootDir, out value));
```

... and get rid of the logging.

PR #2563, opened 2021-08-18
Enable Bind command on Open as Folder projects

### LINKED ISSUE #2562: Enable Bind command on Open as Folder projects
#### Description

Open as folder projects are not recognizable as bindable due to `KnownUIContexts.SolutionExistsAndFullyLoadedContext` and
`KnownUIContexts.SolutionExistsAndNotBuildingAndNotDebuggingContext` evaluating to `false` when in Open as Folder mode. See also #2555.

TODO:
- [x] Support showing the bind command when in Open as Folder mode

rita-sonarsource:
Fixes #2562

### DIFF
--- a/src/Integration.UnitTests/Binding/BindingControllerTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingControllerTests.cs
@@ -31,6 +31,7 @@ using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Infrastructure.VS;
 using SonarLint.VisualStudio.Integration.Binding;
 using SonarLint.VisualStudio.Integration.NewConnectedMode;
 using SonarLint.VisualStudio.Integration.ProfileConflicts;
@@ -58,6 +59,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
         private ConfigurableConfigurationProvider configProvider;
         private ConfigurableVsOutputWindowPane outputWindowPane;
         private ConfigurableSolutionRuleSetsInformationProvider ruleSetsInformationProvider;
+        private Mock<IFolderWorkspaceService> folderWorkspaceServiceMock;
 
         private readonly BoundSonarQubeProject ValidProject = new BoundSonarQubeProject(new Uri("http://any"), "projectKey", "Project Name");
         private readonly BindCommandArgs ValidBindingArgs = new BindCommandArgs("any key", "any name", new ConnectionInformation(new Uri("http://anyXXX")));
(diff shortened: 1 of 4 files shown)


src/Integration/Binding/BindingController.cs
```diff
@@ -95,12 +99,13 @@ protected override int OnQueryStatus(ref Guid pguidCmdGroup, uint cCmds, OLECMD[
...
-                && VsShellUtils.IsSolutionExistsAndFullyLoaded()
-                && VsShellUtils.IsSolutionExistsAndNotBuildingAndNotDebugging()
-                && (this.projectSystemHelper.GetSolutionProjects()?.Any() ?? false);
+                   && args.ProjectKey != null
+                   && this.host.VisualStateManager.IsConnected
+                   && !this.host.VisualStateManager.IsBusy
+                   && (folderWorkspaceService.IsFolderWorkspace()
+                       || (VsShellUtils.IsSolutionExistsAndFullyLoaded()
+                           && VsShellUtils.IsSolutionExistsAndNotBuildingAndNotDebugging()
+                           && (this.projectSystemHelper.GetSolutionProjects()?.Any() ?? false)));
```
duncanp-sonar, line 108: `GetSolutionProjects` currently calls through to `ProjectToLanguageMapper.GetLanguageForProject`, so this condition will only be true if there is a bindable project.

We probably still want to check that condition for CMake projects, since it is a more strict check than just that the current solution is an "open as folder" one i.e.

```
...
&& (folderWorkspaceService.IsFolderWorkspace()
                       || (VsShellUtils.IsSolutionExistsAndFullyLoaded()
                           && VsShellUtils.IsSolutionExistsAndNotBuildingAndNotDebugging())
&& (this.projectSystemHelper.GetSolutionProjects()?.Any() ?? false)
```

src/Integration.UnitTests/Binding/BindingControllerTests.cs
```diff
@@ -436,27 +486,33 @@ private static BindCommandArgs CreateBindingArguments(string key, string name, s
             return new BindCommandArgs(key, name, new ConnectionInformation(new Uri(serverUri)));
         }
 
-        private static ServerViewModel CreateServerViewModel()
```
duncanp-sonar, line 439: Presumably this just wasn't called?

src/Integration.UnitTests/Binding/BindingControllerTests.cs
```diff
@@ -66,28 +68,32 @@ public class BindingControllerTests
         public void TestInitialize()
         {
             KnownUIContextsAccessor.Reset();
-            this.sonarQubeService = new Mock<ISonarQubeService>();
```
duncanp-sonar, line 69: There are lot of stylistic changes which make it difficult to focus on the real changes. Reviewing would be easier if they were done in separate commit. Also, something as straightforward as removing `this` everywhere doesn't really need to be reviewed, so it doesn't have to form part of the review.

duncanp-sonar (commented):
Just a question about whether we should still check `projectSystemHelper.GetSolutionProjects` for "open as folder" projects.

src/Integration.UnitTests/Binding/BindingControllerTests.cs
rita-sonarsource: You're right, my apologies.

src/Integration/Binding/BindingController.cs
rita-sonarsource: Yes you're right, this would allow us to easy disable it for JS 👍

PR #2565, opened 2021-08-18
Support running TestProjectIndicators on Open as Folder projects

### LINKED ISSUE #2564: Support running TestProjectIndicators on Open as Folder projects
#### Description

`BuildPropertyTestProjectIndicator` assumes that the given project has to implement IVsBuildPropertyStorage. This is not the case for Open as Folder projects and the assertion is causing the binding to fail. See also #2555.

TODO:
- [x] Fix TestProjectIndicators to return `null` (= status "don't know") if they cannot retrieve the needed information in order to calculate if it's a test project or not.

rita-sonarsource:
Fixes #2564

### DIFF
--- a/src/Integration/LocalServices/ProjectSystemHelper.cs
+++ b/src/Integration/LocalServices/ProjectSystemHelper.cs
@@ -355,10 +355,6 @@ namespace SonarLint.VisualStudio.Integration
                 Debug.Assert(!ErrorHandler.Succeeded(hr) || hr != E_XML_ATTRIBUTE_NOT_FOUND,
                     $"Failed to get the property '{propertyName}' for project '{dteProject.Name}'.");
             }
-            else
-            {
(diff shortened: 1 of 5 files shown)


src/Integration/LocalServices/ProjectSystemHelper.cs
```diff
@@ -355,10 +355,6 @@ public string GetProjectProperty(Project dteProject, string propertyName, string
                 Debug.Assert(!ErrorHandler.Succeeded(hr) || hr != E_XML_ATTRIBUTE_NOT_FOUND,
                     $"Failed to get the property '{propertyName}' for project '{dteProject.Name}'.");
             }
-            else
-            {
-                Debug.Fail("Could not get IVsBuildPropertyStorage for EnvDTE.Project");
```
rita-sonarsource, line 360: I removed this debug failure. The alternative was to check if it's open as folder in the caller and avoid calling this method to begin with.

duncanp-sonar (approved):

PR #2566, opened 2021-08-18
Remove noisy log message from ServiceGuidTestProjectIndicator

rita-sonarsource:


src/Integration/LocalServices/TestProjectIndicators/ServiceGuidTestProjectIndicator.cs
```diff
@@ -58,8 +55,6 @@ internal ServiceGuidTestProjectIndicator(ILogger logger, IFileSystem fileSystem)
             }
             catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
             {
-                logger.WriteLine(Strings.FailedToCheckIfTestProject, project.UniqueName, ex.ToString());
```
rita-sonarsource, line 61: Noticed this now when looking at the output for projectMapper fix

duncanp-sonar (approved):

PR #2567, opened 2021-08-18
Add service to fetch VSDevCmd settings

### LINKED ISSUE #2560: Add service to fetch VSDevCmd.bat environment variables
### Description

Locate `VSDevCmd.bat`, execute it, and return the environment variables that have been set by the script.

### Notes

* VS run `vsDevCmd.bat` once or twice per session for the first CMake solution that is opened (it doesn't appear to be run for `vcxproj` projects). It is not re-run if the project is closed and re-opened, or if another CMake project is opened.
VS assembly: `C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\Microsoft.VisualStudio.VC.dll`
Method called: `Microsoft.VisualStudio.VC.dll!Microsoft.VisualStudio.VC.Workspace.VsDevCmdPropertyEvaluator.GetVsDevCmdEnvironment(string scriptParams)`

duncanp-sonar:
Fixes #2560

### DIFF
--- a/src/CFamily.UnitTests/CMake/VsDevCmdEnvironmentProviderTests.cs
+++ b/src/CFamily.UnitTests/CMake/VsDevCmdEnvironmentProviderTests.cs
@@ -36,7 +36,7 @@ using SonarLint.VisualStudio.Integration.UnitTests;
 namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
 {
     [TestClass]
-    public class _VsDevCmdEnvironmentProviderTests
+    public class VsDevCmdEnvironmentProviderTests
     {
         public TestContext TestContext { get; set; }
 
--- a/src/CFamily.UnitTests/CMake/VsDevCmdEnvironmentProviderTests.cs
+++ b/src/CFamily.UnitTests/CMake/VsDevCmdEnvironmentProviderTests.cs
@@ -21,7 +21,10 @@
 
 using System;
 using System.Diagnostics;
+using System.IO;
+using System.IO.Abstractions;
 using System.Linq;
+using System.Threading;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
@@ -37,69 +40,65 @@ namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
     {
         public TestContext TestContext { get; set; }
 
-        private readonly IVsInfoService ValidVsInfoService = CreateVsInfoService("c:\\any");
-
         [TestMethod]
         public void Get_MissingFile_ReturnsEmptySettings()
         {
+            var context = new ProcessContext(installRootDir: "c:\\aaa\\bbb", batchFileExists: false);
 
+            context.TestSubject.Get("any").Should().BeNull();
+
+            context.FileSystem.Verify(x => x.File.Exists("c:\\aaa\\bbb\\Common7\\Tools\\VsDevCmd.bat"), Times.Once);
         }
 
         [TestMethod]
-        public void Get_FileExists_ReturnsExpectedSettings()
+        [DataRow("input-script-params")]
+        [DataRow(null)]
+        public void Get_ExpectedCommandsPassed(string scriptParams)
         {
-        }
-
-        [TestMethod]
-        public void Get_ExpectedProcessArgumentsPassed()
-        {
-            var infoService = CreateVsInfoService("c:\\any");
-
-            ProcessStartInfo actualProcessStartInfo = null;
-
-            var process = new Mock<IProcess>();
-            var processFactory = new Mock<IProcessFactory>();
-            processFactory.Setup(x => x.Start(It.IsAny<ProcessStartInfo>()))
-                .Callback<ProcessStartInfo>(x => actualProcessStartInfo = x)
-                .Returns(process.Object);
-
-            var testSubject = new VsDevCmdEnvironmentProvider(infoService, new TestLogger(), processFactory.Object);
+            var context = new ProcessContext(installRootDir: "d:\\myinstalldir\\");
 
             // Act
-            testSubject.Get("input-script-arguments");
+            context.TestSubject.Get(scriptParams);
 
-            actualProcessStartInfo.Should().NotBeNull();
-            actualProcessStartInfo.FileName.Should().Be(Environment.GetEnvironmentVariable("COMSPEC"));
+            context.ActualProcessStartInfo.Should().NotBeNull();
+            context.ActualProcessStartInfo.FileName.Should().Be(Environment.GetEnvironmentVariable("COMSPEC"));
 
-            actualProcessStartInfo.Arguments.Should().NotBeNull();
+            context.ActualProcessStartInfo.Arguments.Should().NotBeNull();
+            context.TestSubject.UniqueId.Should().NotBeNullOrEmpty();
 
-            // Split and clean up the command string to make the individual arguments easier to check
-            var args = actualProcessStartInfo.Arguments.Split(new string[] { " && " }, StringSplitOptions.None)
+            // Split and clean up the command string to make the individual commands easier to check
+            var args = context.ActualProcessStartInfo.Arguments.Split(new string[] { " && " }, StringSplitOptions.None)
                 .Select(x => x.Trim())
                 .ToArray();
 
             args[0].Should().Be("/U /K set VSCMD_SKIP_SENDTELEMETRY=1"); // args to cmd.exe and first command in the sequence
-            args[1].Should().Be(@"""c:\any\Common7\Tools\VsDevCmd.bat"" input-script-arguments"); // calculated full path to VsDevCmd.bat
-            args[2].Should().StartWith("echo SONARLINT_BEGIN_CAPTURE");
+            args[1].Should().Be($@"""d:\myinstalldir\Common7\Tools\VsDevCmd.bat"" {scriptParams}".TrimEnd()); // calculated full path to VsDevCmd.bat with additional params
+            args[2].Should().Be("echo SONARLINT_BEGIN_CAPTURE " + context.TestSubject.UniqueId);
             args[3].Should().Be("set");
-            args[4].Should().StartWith("echo SONARLINT_END_CAPTURE");
+            args[4].Should().StartWith("echo SONARLINT_END_CAPTURE " + context.TestSubject.UniqueId);
         }
 
         [TestMethod]
-        public void Get_Lifecycle_NoOutput_RunsToCompletion()
+        public void Get_ExpectedTimeoutDurationUsed()
         {
-            var process = new Mock<IProcess>();
-            var processFactory = new Mock<IProcessFactory>();
-            processFactory.Setup(x => x.Start(It.IsAny<ProcessStartInfo>()))
-                .Returns(process.Object);
-
-            var testSubject = new VsDevCmdEnvironmentProvider(ValidVsInfoService, new TestLogger(), processFactory.Object);
+            var context = new ProcessContext();
 
             // Act
-            testSubject.Get("any");
+            context.TestSubject.Get("any");
 
-            // Just checking the invociation order
-            var invokedMembers = process.Invocations.Select(x => x.Method.Name).ToArray();
+            context.Process.Verify(x => x.WaitForExit(4000), Times.Once);
+        }
+
+        [TestMethod]
+        public void Get_Lifecycle_RunsToCompletion()
+        {
+            var context = new ProcessContext(hasExited: true);
+
+            // Act
+            context.TestSubject.Get("any");
+
+            // Just checking the invocation order
+            var invokedMembers = context.Process.Invocations.Select(x => x.Method.Name).ToArray();
             invokedMembers.Should().ContainInOrder(
                 // Initialize and run
                 "set_HandleOutputDataReceived",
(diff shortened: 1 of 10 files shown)


src/CFamily/SystemAbstractions/IProcess.cs
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+using System;
+using System.Diagnostics;
+
+/// <summary>
+/// Wrappers around System.Diagnostics.Process for testing
+/// </summary>
+
+namespace SonarLint.VisualStudio.CFamily.SystemAbstractions
```
duncanp-sonar, line 28: As discussed offline, I'll rename the existing `Core...IProcess` class and move this file to `Core` in a separate commit.

src/CFamily/SystemAbstractions/IProcess.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    internal interface IProcessFactory
+    {
+        IProcess Start(ProcessStartInfo startInfo);
+    }
+
+    internal interface IProcess : IDisposable
```
duncanp-sonar, line 35: Limited abstraction around `System.Diagnostics.Process` - just those members required by the CMake VsDevCmd runner code.

src/TestInfrastructure/Helpers/ProcessCleanupHelper.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    /// <summary>
+    /// Tries to ensure the process is terminated when a test finishes.
+    /// Usage:
+    ///   using var cleanHelper = new ProcessCleanupHelper(process);
+    /// </summary>
+    /// <remarks>Disposing a process does not terminate it, so we can't rely on
+    /// "using var process = ..." to clean up properly.
+    /// Instead, we'll try to fetch the process by id then "Kill" it.
+    /// </remarks>
```
duncanp-sonar, line 18: Originally, I was using the 'using [a Process instance]`, assuming that the `Process.Dispose()` would terminate the process, but it doesn't.

However, `using var cleanupHelper ...` will.

src/CFamily.UnitTests/SystemAbstractions/ProcessWrapperTests.cs
```diff
@@ -0,0 +1,209 @@
...
+
+namespace SonarLint.VisualStudio.CFamily.UnitTests.SystemAbstractions
+{
+    [TestClass]
+    public class ProcessWrapperTests
+    {
+        public TestContext TestContext { get; set; }
+
+        [TestMethod]
+        public void ProcessFactory_Start_ReturnsWrappedProcess()
```
duncanp-sonar, line 39: Strictly speaking this first test is for the `ProcessFactory` class. However, it didn't seem worth putting it in a separate class, particularly since it uses some of the same helper methods.

src/CFamily/SystemAbstractions/IProcess.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        int Id { get; }
+        bool HasExited { get; }
+        int ExitCode { get; }
+        ProcessStartInfo StartInfo { get; }
```
duncanp-sonar, line 40: The `StartInfo` property isn't used by the `VSDevCmd` runner. It's there mainly to help with testing the factory class i.e. to make sure it has passed through the expected instance.

duncanp-sonar:
There's more code here than I'd normally have in a PR, so apologies for that. I suggest reviewing the supporting `ProcessFactory` / `ProcessWrapper` types and tests first, and then review the VsDevCmd code.

src/CFamily/SystemAbstractions/IProcess.cs
```diff
@@ -0,0 +1,107 @@
...
+        public void BeginOutputReadLine() => wrapped.BeginOutputReadLine();
+
+        private Action<string> handleOutputData;
+
+        public Action<string> HandleOutputDataReceived
+        {
+            get { return handleOutputData; }
+            set
+            {
+                handleOutputData = value;
```
duncanp-sonar, line 88: This is the only complicated bit - turning the event into a callback.

src/CFamily.UnitTests/SystemAbstractions/ProcessWrapperTests.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        [TestMethod]
+        public void Execute_PropertiesReturnExpectedValue()
+        {
+            var exeName = WriteBatchFileForTest(TestContext,
+@"@echo Hello world
+exit -2
+");
+
+            using var processScope = StartProcess(exeName);
```
duncanp-sonar, line 71: There's an explanation in the new `ProcessScope` class that explains why we need the `using var processScope = StartProcess(...)` pattern.

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+        internal /* for testing */ VsDevCmdEnvironmentProvider(IVsInfoService vsInfoService, ILogger logger,
+            IProcessFactory processFactory, IFileSystem fileSystem)
+        {
+            this.vsInfoService = vsInfoService;
+            this.logger = logger;
+            this.processFactory = processFactory;
+            this.fileSystem = fileSystem;
+        }
+
+        public IReadOnlyDictionary<string, string> Get(string scriptParams)
```
duncanp-sonar, line 78: Overall the implementation is similar to the VS version, but will more logging and interfaces to make it testable.

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+        {
+            try
+            {
+                string filePath = Path.Combine(vsInfoService.InstallRootDir, RelativePathToBatchFile);
+                LogDebug($"VsDevCmd location: {filePath}");
+
+                if (!fileSystem.File.Exists(filePath))
+                {
+                    logger.WriteLine(Resources.VsDevCmd_FileNotFound, filePath);
```
duncanp-sonar, line 87: Any failures are logged as normal priority so they will always be visible.

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+        {
+            UniqueId = Guid.NewGuid().ToString();
+            var beginToken = "SONARLINT_BEGIN_CAPTURE " + UniqueId;
+            var endToken = "SONARLINT_END_CAPTURE " + UniqueId;
+
+            ProcessStartInfo startInfo = new ProcessStartInfo()
+            {
+                FileName = Environment.GetEnvironmentVariable("COMSPEC"),
+                Arguments = "/U " +                             // Unicode input / output
+                "/K " +                                         // Execute the command and remain active. Need to do this or we can't fetch the env vars
```
duncanp-sonar, line 123: `/K` means the process won't terminate, which is why there is code that explicitly kills it, either in the event of a timeout, or after the "end capture" marker has been read.

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+
+            foreach (var item in capturedOutput)
+            {
+                var equalsIndex = item.IndexOf("=", StringComparison.OrdinalIgnoreCase);
+                if (equalsIndex > -1)
+                {
+                    var key = item.Substring(0, equalsIndex);
+                    var newValue = item.Substring(equalsIndex + 1);
+
+                    settings.Add(key, newValue);
```
duncanp-sonar, line 203: The VS version filters out environment variables that weren't changed by the script.
I don't think we need to do that - I'll check.

src/CFamily.UnitTests/CMake/VsDevCmdEnvironmentProviderTests.cs
```diff
@@ -0,0 +1,357 @@
...
+            context.FileSystem.Reset();
+            context.FileSystem.Setup(x => x.File.Exists(It.IsAny<string>()))
+                .Throws(new StackOverflowException("thrown from test"));
+
+            Action act = () => context.TestSubject.Get("any");
+
+            act.Should().ThrowExactly<StackOverflowException>().And.Message.Should().Be("thrown from test");
+        }
+
+        private class ProcessContext
```
duncanp-sonar, line 232: The tests are relatively straightforward, because the setup complexity is all in this class.

The class falls into two main parts:
* a set of Mocks that are always created for the process factory etc.
* a more complicated "simulate process execution" part that is only configured if the test provides a callback to simulate the process doing something.

src/CFamily.UnitTests/CMake/VsDevCmdEnvironmentProviderTests.cs
```diff
@@ -0,0 +1,358 @@
...
+using Moq;
+using SonarLint.VisualStudio.CFamily.CMake;
+using SonarLint.VisualStudio.CFamily.SystemAbstractions;
+using SonarLint.VisualStudio.Infrastructure.VS;
+using SonarLint.VisualStudio.Integration.UnitTests;
+
+namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
+{
+    [TestClass]
+    public class _VsDevCmdEnvironmentProviderTests
```
duncanp-sonar, line 39: Not sure why this class has a "_" at the start. I'll remove it.

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+    internal interface IVsDevCmdEnvironmentProvider
+    {
+        /// <summary>
+        /// Returns the environment settings set by VsDevCmd.bat
+        /// </summary>
+        /// <param name="scriptParams">Any additional parameters to pass to the batch file. Can be null.</param>
+        IReadOnlyDictionary<string, string> Get(string scriptParams);
+    }
+
+    internal class VsDevCmdEnvironmentProvider : IVsDevCmdEnvironmentProvider
```
duncanp-sonar, line 44: I'll add an `Export` if necessary in a later PR when I wire this class to the `CompilationDatabaseRequestFactory`.

src/TestInfrastructure/Helpers/ProcessScope.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        public void Dispose()
+        {
+            try
+            {
+                var runningProcess = Process.GetProcessById(id);
+                Log("Process is running.  Id: " + id);
+
+                // Make sure we don't accidentally kill a newly-created process with the same id
```
rita-sonarsource, line 41: Regarding this comment: if our code is the one creating the process, wouldn't it always have the same module name and process name?

src/TestInfrastructure/Helpers/ProcessScope.cs
```diff
@@ -0,0 +1,97 @@
...
+            id = process.Id;
+
+            (mainModuleFilName, processName) = TryGetProcessDetails(process);
+        }
+
+        public void Dispose()
+        {
+            try
+            {
+                var runningProcess = Process.GetProcessById(id);
```
rita-sonarsource, line 38: Why not dispose the existing Process property?

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+using System.Text;
+using Microsoft.VisualStudio;
+using SonarLint.VisualStudio.CFamily.SystemAbstractions;
+using SonarLint.VisualStudio.Infrastructure.VS;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.Helpers;
+
+namespace SonarLint.VisualStudio.CFamily.CMake
+{
+    internal interface IVsDevCmdEnvironmentProvider
```
rita-sonarsource, line 35: Maybe IVsDevCmdEnvironmentVarsProvider ? or just IDevCmdEnvironmentVarsProvider.
It's a bit longer but just Environment is less accurate

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        // Interfaces for testing
+        private readonly IProcessFactory processFactory;
+        private readonly IFileSystem fileSystem;
+
+
+        public VsDevCmdEnvironmentProvider(IVsInfoService vsInfoService, ILogger logger)
+            : this(vsInfoService, logger, new ProcessFactory(), new FileSystem())
+        {
+            this.vsInfoService = vsInfoService;
```
rita-sonarsource, line 65: no need for there initializations

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+        {
+            try
+            {
+                string filePath = Path.Combine(vsInfoService.InstallRootDir, RelativePathToBatchFile);
+                LogDebug($"VsDevCmd location: {filePath}");
+
+                if (!fileSystem.File.Exists(filePath))
+                {
+                    logger.WriteLine(Resources.VsDevCmd_FileNotFound, filePath);
+                    return null;
```
rita-sonarsource, line 88: just a question of usage, returning empty vs null.

src/TestInfrastructure/Helpers/ProcessScope.cs
duncanp-sonar: That's the point; we don't know if it is the same process, just that it has the same id as the original process.

Checking the other properties give some additional level of certainty that we are killing the correct process, and not a newly-created process that Windows happens to have allocated the same Id.

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+            }
+            catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+            {
+                logger.WriteLine(Resources.VsDevCmd_ErrorFetchingSettings, ex);
+            }
+
+            return null;
+        }
+
+        internal /* for testing */ string UniqueId { get; private set;}
```
rita-sonarsource, line 111: I'm debating if we should allow to run this more than once. The variables are cached on VS's side, so is there any benefit in supporting multiple calls? It would simplify things to fetch it in the CTOR and cache them. WDYT?

src/TestInfrastructure/Helpers/ProcessScope.cs
duncanp-sonar: Disposing won't help, we need to call Kill.

Could do, but the call to Kill would fail if the process has been disposed (as would calls to any other members). So we'd have to make that Kill call inside a separate try-catch block, then fall back GetProcessById if that didn't work.

The GetProcessById version should work regardless of whether the original Process has been disposed or not.

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
```diff
@@ -0,0 +1,231 @@
...
+            if (capturedOutput == null)
+            {
+                return null;
+            }
+
+            var settings = new Dictionary<string, string>();
+
+            foreach (var item in capturedOutput)
+            {
+                var equalsIndex = item.IndexOf("=", StringComparison.OrdinalIgnoreCase);
```
rita-sonarsource, line 197: why not split on "="?

duncanp-sonar: It's valid for an environment variable value to contain "=".  We want to split on the first "=" only.

rita-sonarsource: Right, good point

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
duncanp-sonar: My assumption is that we might need to call this multiple times with different 'scriptParams', so potentially we'd call this multiple times with different inputs.

We certainly won't want to run this on every call, though - it takes just under a second to run on my local machine. 
There is also the issue of some kind of synchronization mechanism to make sure we've fetched the data before the analyzers run, and this issue that the user might be able to override some of the settings in the json/cmake settings file.

However, I think all of those things need to be done outside this class. This class is just responsible for fetching the VS settings when asked.

src/CFamily/CMake/VsDevCmdEnvironmentProvider.cs
duncanp-sonar: Difficult to say which is more convenient at this point. We can always change it later when we have a real consumer.

src/CFamily.UnitTests/SystemAbstractions/ProcessWrapperTests.cs
```diff
@@ -0,0 +1,212 @@
...
+                wrappedProcess = testSubject.Start(processStartInfo);
+
+                wrappedProcess.Should().NotBeNull();
+                wrappedProcess.StartInfo.Should().BeSameAs(processStartInfo);
+            }
+            finally
+            {
+                if (!wrappedProcess.HasExited)
+                {
+                    wrappedProcess?.Kill();
```
rita-sonarsource, line 58: This can throw. We do a try-catch in [similar tests](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs#L69) but maybe the wrapper can do that instead.

rita-sonarsource (approved):
Looks good. There is one failing test ATM, `Execute_OutputIsForwarded` -- I'm not sure why that would fail. Is it possible that the CI machine was busy and hasn't really launched the process?

src/CFamily.UnitTests/SystemAbstractions/ProcessWrapperTests.cs
duncanp-sonar: I'll add a try-catch here and keep the wrapper as a simple passthrough for now.

duncanp-sonar:
> Is it possible that the CI machine was busy and hasn't really launched the process?

No - the test successfully called `Process.ExitCode`, which would have thrown if the process hadn't ended. The output messages are processed asynchronously, so it's possible they hadn't arrived at the point the test checked for them.

PR #2571, opened 2021-08-19
Move/rename IProcess and tests

duncanp-sonar:


duncanp-sonar:
No code changes: just file moves, namespace changes, and scope changes (moved classes and interfaces made public so they are accessible from the CFamily project).

PR #2572, opened 2021-08-19
Added logging to flaky test

duncanp-sonar:
Tweaked test in an attempt to reduce the flakiness.
Added test logging to help diagnose failures.

### DIFF
--- a/src/CFamily.UnitTests/SystemAbstractions/ProcessWrapperTests.cs
+++ b/src/CFamily.UnitTests/SystemAbstractions/ProcessWrapperTests.cs
@@ -24,6 +24,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Text;
 using FluentAssertions;
+using Microsoft.Internal.VisualStudio.PlatformUI;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarLint.VisualStudio.CFamily.SystemAbstractions;
 using SonarLint.VisualStudio.Integration.UnitTests.Helpers;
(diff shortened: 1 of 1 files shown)


src/CFamily.UnitTests/SystemAbstractions/ProcessWrapperTests.cs
```diff
@@ -125,20 +126,22 @@ public void Execute_ProcessTimesOut_CanBeKilled()
...
 set /p arg=
+exit -999
 ");
 
             using var processScope = StartProcess(exeName);
             var testSubject = new ProcessWrapper(processScope.Process);
 
             // Should timeout be because the batch file is waiting for input...
-            testSubject.WaitForExit(2000);
+            testSubject.WaitForExit(1000);
```
duncanp-sonar, line 136: Minor tweak to the test - shouldn't need to wait two seconds, since we know the process shouldn't exit.

rita-sonarsource (approved):

src/CFamily.UnitTests/SystemAbstractions/ProcessWrapperTests.cs
```diff
@@ -125,20 +126,22 @@ public void Execute_ProcessTimesOut_CanBeKilled()
...
             testSubject.HasExited.Should().Be(false);
             CheckProcessIsRunning(testSubject.Id);
 
             // Now kill the process
             testSubject.Kill();
             testSubject.HasExited.Should().Be(true);
             testSubject.ExitCode.Should().Be(-1);
+            System.Threading.Thread.Sleep(500);
```
duncanp-sonar, line 144: Added a short delay before checking the process has terminated (in case the problem was that the managed framework _Process_ class has finished it's work, but the OS hasn't finished clearing up the underlying native process).

If these tests continue to be flaky I'll [Ignore] them and create a ticket to investigate them later.

Issue #2570, opened 2021-08-19
Integrate CMake request factory and VS env vars provider


duncanp-sonar:
### Description

The CMake request factory needs to get the VS env vars.

However, we don't want to fetch them on every analysis (takes about one second to fetch the data), so some kind of cache will be required.

It's ok if the env vars are only fetched once per VS session.

PR #2573, opened 2021-08-19
Pass env vars to CompilationDatabaseRequest from factory

duncanp-sonar:
* the factory is responsible for getting the values to pass

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CMake/CompilationDatabaseRequestTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CMake/CompilationDatabaseRequestTests.cs
@@ -38,14 +38,15 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily.CMake
     {
         private readonly RequestContext ValidContext = new RequestContext("cpp", Mock.Of<ICFamilyRulesConfig>(), "file.txt", "pchFile.txt", null);
         private readonly CompilationDatabaseEntry ValidDbEntry = new CompilationDatabaseEntry { File = "file.txt", Directory = "c:\\", Command = "a command" };
+        private readonly IReadOnlyDictionary<string, string> ValidEnvVars = new Dictionary<string, string> { { "key1", "value1" } };
 
         [TestMethod]
         public void Ctor_NullArguments_Throws()
         {
-            Action act = () => new CompilationDatabaseRequest(null, ValidContext);
+            Action act = () => new CompilationDatabaseRequest(null, ValidContext, ValidEnvVars);
             act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("databaseEntry");
 
-            act = () => new CompilationDatabaseRequest(ValidDbEntry, null);
+            act = () => new CompilationDatabaseRequest(ValidDbEntry, null, ValidEnvVars);
             act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("context");
         }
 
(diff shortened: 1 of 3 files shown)


duncanp-sonar:
Minor refactoring in preparation for #2570 (integrating the VsDevCmd helper).
The request should be passed the environment variables to use by the factory.

src/Integration.Vsix.UnitTests/CFamily/CMake/CompilationDatabaseRequestTests.cs
```diff
@@ -38,23 +38,31 @@ public class CompilationDatabaseRequestTests
     {
         private readonly RequestContext ValidContext = new RequestContext("cpp", Mock.Of<ICFamilyRulesConfig>(), "file.txt", "pchFile.txt", null);
         private readonly CompilationDatabaseEntry ValidDbEntry = new CompilationDatabaseEntry { File = "file.txt", Directory = "c:\\", Command = "a command" };
+        private readonly IReadOnlyDictionary<string, string> ValidEnvVars = new Dictionary<string, string> { { "key1", "value1" } };
 
         [TestMethod]
-        public void Ctor_NullArguments_Throws()
+        public void Ctor_InvalidArguments_Throws()
```
duncanp-sonar, line 44: Renamed - EnvVars can be null.

src/Integration.Vsix.UnitTests/CFamily/CMake/CompilationDatabaseRequestTests.cs
```diff
@@ -38,23 +38,31 @@ public class CompilationDatabaseRequestTests
...
+            testSubject.EnvironmentVariables.Should().BeNull();
+        }
+
         [TestMethod]
         [DataRow(null, "args")]
         [DataRow("", "args")]
         [DataRow("cmd", null)]
         [DataRow("cmd", "")]
-        public void Ctor_ValidDBEntry_ShouldNotThrow(string command, string args)
+        public void Ctor_ValidCommandArgsCombination_ShouldNotThrow(string command, string args)
```
duncanp-sonar, line 65: In retrospect, I didn't need to rename this test or the next one. However, the new names are probably more descriptive of what the test is doing.

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
```diff
@@ -36,16 +36,17 @@ internal class CompilationDatabaseRequest : IRequest
         private readonly CompilationDatabaseEntry databaseEntry;
         private readonly IRulesConfigProtocolFormatter rulesConfigProtocolFormatter;
 
-        public CompilationDatabaseRequest(CompilationDatabaseEntry databaseEntry, RequestContext context)
-            : this(databaseEntry, context, new RulesConfigProtocolFormatter())
+        public CompilationDatabaseRequest(CompilationDatabaseEntry databaseEntry, RequestContext context, IReadOnlyDictionary<string, string> environmentVariables)
```
rita-sonarsource, line 39: I wonder if this is something that should be on the context.
We discussed that the subprocess already supports env vars, so logically it makes sense for it to be for all request types.

src/Integration.Vsix.UnitTests/CFamily/CMake/CompilationDatabaseRequestTests.cs
rita-sonarsource: If they can be null, maybe it's worth to be explicit here and pass null, otherwise it looks like a forgotten case.

rita-sonarsource (approved):

src/Integration.Vsix/CFamily/CMake/CompilationDatabaseRequest.cs
duncanp-sonar: It could be, I guess. It would be easy enough to move.

duncanp-sonar: On second thoughts, I don't think it should be. The environment variables are calculated by the factory, and form part of the request protocol itself (env vars + binary input message).

Looking at the `RequestContext` again, I don't think it's responsibilities or role are clear - it seems to be doing multiple things.

PR #2575, opened 2021-08-19
Add support for CMake in ProjectToLanguageMapper

### LINKED ISSUE #2576: Support detecting the Analysis Language of CMake projects
### Description

`ProjectToLanguageMapper` attempts to resolve the analysis language based on the project's guid. For Open as Folder projects such as CMake, the guid is always "{66a2671d-8fb5-11d2-aa7e-00c04f688dde}" (see also #2555).

TODO:
- [x] Support identifying language for CMake projects.

rita-sonarsource:
Fixes #2576

### DIFF
--- a/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
@@ -34,8 +34,8 @@ using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using NuGet;
-using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Infrastructure.VS;
 using SonarLint.VisualStudio.Integration.Binding;
 using SonarLint.VisualStudio.Integration.NewConnectedMode;
 using SonarLint.VisualStudio.Integration.Resources;
@@ -77,7 +77,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             this.serviceProvider.RegisterService(typeof(IProjectSystemHelper), this.projectSystemHelper);
             this.serviceProvider.RegisterService(typeof(ISolutionRuleSetsInformationProvider), this.ruleSetsInformationProvider);
 
-            var projectToLanguageMapper = new ProjectToLanguageMapper(Mock.Of<IAbsoluteFilePathLocator>());
+            var projectToLanguageMapper = new ProjectToLanguageMapper(Mock.Of<IFolderWorkspaceService>());
             var mefHost = ConfigurableComponentModel.CreateWithExports(MefTestHelpers.CreateExport<IProjectToLanguageMapper>(projectToLanguageMapper));
             serviceProvider.RegisterService(typeof(SComponentModel), mefHost);
 
--- a/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
@@ -29,10 +29,12 @@ using System.Windows.Threading;
 using EnvDTE;
 using FluentAssertions;
 using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.ComponentModelHost;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using NuGet;
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Integration.Binding;
 using SonarLint.VisualStudio.Integration.NewConnectedMode;
@@ -74,6 +76,10 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             this.serviceProvider.RegisterService(typeof(IRuleSetSerializer), ruleSerializer);
             this.serviceProvider.RegisterService(typeof(IProjectSystemHelper), this.projectSystemHelper);
             this.serviceProvider.RegisterService(typeof(ISolutionRuleSetsInformationProvider), this.ruleSetsInformationProvider);
+
+            var projectToLanguageMapper = new ProjectToLanguageMapper(Mock.Of<IAbsoluteFilePathLocator>());
+            var mefHost = ConfigurableComponentModel.CreateWithExports(MefTestHelpers.CreateExport<IProjectToLanguageMapper>(projectToLanguageMapper));
+            serviceProvider.RegisterService(typeof(SComponentModel), mefHost);
 
             this.host = new ConfigurableHost(this.serviceProvider, Dispatcher.CurrentDispatcher);
         }
--- a/src/Integration.UnitTests/Binding/ProjectBinderFactoryTests.cs
+++ b/src/Integration.UnitTests/Binding/ProjectBinderFactoryTests.cs
@@ -21,8 +21,10 @@
 using System;
 using System.IO.Abstractions;
 using FluentAssertions;
+using Microsoft.VisualStudio.ComponentModelHost;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration.Binding;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
@@ -30,12 +32,22 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
     [TestClass]
     public class ProjectBinderFactoryTests
     {
+        private Mock<IProjectToLanguageMapper> projectToLanguageMapper;
         private ProjectBinderFactory testSubject;
 
         [TestInitialize]
         public void TestInitialize()
         {
-            testSubject = new ProjectBinderFactory(Mock.Of<IServiceProvider>(), Mock.Of<ILogger>(), Mock.Of<IFileSystem>());
+            projectToLanguageMapper = new Mock<IProjectToLanguageMapper>();
+
+            var serviceProvider = new ConfigurableServiceProvider();
+            var mefHost = ConfigurableComponentModel.CreateWithExports(MefTestHelpers.CreateExport<IProjectToLanguageMapper>(projectToLanguageMapper.Object));
+            serviceProvider.RegisterService(typeof(SComponentModel), mefHost);
+            serviceProvider.RegisterService(typeof(ISolutionRuleSetsInformationProvider), Mock.Of<ISolutionRuleSetsInformationProvider>());
+            serviceProvider.RegisterService(typeof(IRuleSetSerializer), Mock.Of<IRuleSetSerializer>());
+            serviceProvider.RegisterService(typeof(IProjectSystemHelper), Mock.Of<IProjectSystemHelper>());
+
+            testSubject = new ProjectBinderFactory(serviceProvider, Mock.Of<ILogger>(), Mock.Of<IFileSystem>());
         }
 
         [TestMethod]
@@ -65,8 +77,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
         [TestMethod]
         public void Get_CSharpProject_CSharpVBProjectBinderReturned()
         {
-            var project = new ProjectMock("c:\\foo.proj");
-            project.SetCSProjectKind();
+            var project = new ProjectMock("c:\\foo.xxxx");
+            projectToLanguageMapper.Setup(x => x.GetAllBindingLanguagesForProject(project))
+                .Returns(new[] {Language.CSharp});
 
             using (new AssertIgnoreScope())
             {
@@ -78,8 +91,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
         [TestMethod]
         public void Get_VbNetProject_CSharpVBProjectBinderReturned()
         {
-            var project = new ProjectMock("c:\\foo.proj");
-            project.SetVBProjectKind();
+            var project = new ProjectMock("c:\\foo.xxxx");
+            projectToLanguageMapper.Setup(x => x.GetAllBindingLanguagesForProject(project))
+                .Returns(new[] { Language.VBNET });
 
             using (new AssertIgnoreScope())
             {
@@ -91,8 +105,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
         [TestMethod]
         public void Get_CppProject_CFamilyProjectBinderReturned()
         {
-            var project = new ProjectMock("c:\\foo.proj");
-            project.ProjectKind = ProjectSystemHelper.CppProjectKind;
+            var project = new ProjectMock("c:\\foo.xxxx");
+            projectToLanguageMapper.Setup(x => x.GetAllBindingLanguagesForProject(project))
+                .Returns(new[] { Language.Cpp });
 
             using (new AssertIgnoreScope())
             {
@@ -102,10 +117,26 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
         }
 
         [TestMethod]
-        public void Get_NonRoslynProject_CFamilyProjectBinderReturned()
+        public void Get_CProject_CFamilyProjectBinderReturned()
         {
-            var project = new ProjectMock("c:\\foo.proj");
-            project.ProjectKind = "{" + Guid.NewGuid() + "}";
+            var project = new ProjectMock("c:\\foo.xxxx");
+            projectToLanguageMapper.Setup(x => x.GetAllBindingLanguagesForProject(project))
+                .Returns(new[] { Language.C });
+
+            using (new AssertIgnoreScope())
+            {
+                var configProjectBinder = testSubject.Get(project);
+                configProjectBinder.Should().BeOfType<CFamilyProjectBinder>();
+            }
+        }
+
+        [TestMethod]
+        public void Get_UnknownProject_CFamilyProjectBinderReturned()
+        {
+            // todo: this is wrong and should be null
+            var project = new ProjectMock("c:\\foo.xxxx");
+            projectToLanguageMapper.Setup(x => x.GetAllBindingLanguagesForProject(project))
+                .Returns(new[] { Language.Unknown });
 
             using (new AssertIgnoreScope())
             {
--- a/src/Integration.UnitTests/Binding/ProjectBinderFactoryTests.cs
+++ b/src/Integration.UnitTests/Binding/ProjectBinderFactoryTests.cs
@@ -131,9 +131,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
         }
 
         [TestMethod]
-        public void Get_UnknownProject_CFamilyProjectBinderReturned()
+        public void Get_UnknownProject_Null()
         {
-            // todo: this is wrong and should be null
             var project = new ProjectMock("c:\\foo.xxxx");
             projectToLanguageMapper.Setup(x => x.GetAllBindingLanguagesForProject(project))
                 .Returns(new[] { Language.Unknown });
@@ -141,7 +140,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
             using (new AssertIgnoreScope())
             {
                 var configProjectBinder = testSubject.Get(project);
-                configProjectBinder.Should().BeOfType<CFamilyProjectBinder>();
+                configProjectBinder.Should().BeNull();
             }
         }
     }
--- a/src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
+++ b/src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
@@ -31,7 +31,6 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Infrastructure.VS;
 using IVsHierarchy = Microsoft.VisualStudio.Shell.Interop.IVsHierarchy;
-using Language = SonarLint.VisualStudio.Core.Language;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests
 {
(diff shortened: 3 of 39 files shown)


src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -75,6 +77,10 @@ public void TestInitialize()
             this.serviceProvider.RegisterService(typeof(IProjectSystemHelper), this.projectSystemHelper);
             this.serviceProvider.RegisterService(typeof(ISolutionRuleSetsInformationProvider), this.ruleSetsInformationProvider);
 
+            var projectToLanguageMapper = new ProjectToLanguageMapper(Mock.Of<IAbsoluteFilePathLocator>());
```
rita-sonarsource, line 80: The tests were complicated to switch to mocks so I'm using the real instance.

src/Integration/ProjectToLanguageMapper.cs
```diff
@@ -46,37 +71,43 @@ public static class ProjectToLanguageMapper
...
-            Language language;
-            if (KnownProjectTypes.TryGetValue(projectKind, out language))
+            if (KnownProjectTypes.TryGetValue(projectKind, out var language))
             {
                 return language;
             }
+
+            if (projectKind == OpenAsFolderProject)
+            {
+                var cmakeProjectFile = absoluteFilePathLocator.Locate("CMakeLists.txt");
```
rita-sonarsource, line 93: As we discussed offline, this is the real product code change.

src/Integration/ProfileConflicts/ConflictsManager.cs
```diff
@@ -138,7 +140,7 @@ private RuleSetInformation[] CheckSlnLevelConfigExistsAndReturnAllProjectRuleSet
             {
                 // Solution-level checks (done here because the expected solution-level config
                 // depends on the languages supported by the project that exist)
-                string baselineRuleSet = bindingConfiguration.BuildPathUnderConfigDirectory(ProjectToLanguageMapper.GetLanguageForProject(project).FileSuffixAndExtension);
+                string baselineRuleSet = bindingConfiguration.BuildPathUnderConfigDirectory(projectToLanguageMapper.GetAllBindingLanguagesForProject(project).FirstOrDefault().FileSuffixAndExtension);
```
rita-sonarsource, line 143: `GetAllBindingLanguagesForProject` can never return an empty list, as it uses the inner `GetLanguageForProject` that would return Unknown -- so it would be a list of 1 with "Unknown" in it.

In hindsight I think using `First` would be better than `FirstOrDefault`, LMK what you think and I'll update here and in the other places.

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -70,8 +71,20 @@ internal sealed class VsSessionHost : IHost, IProgressStepRunnerWrapper, IDispos
...
         [ImportingConstructor]
         public VsSessionHost([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider,
-            ISonarQubeService sonarQubeService, IActiveSolutionTracker solutionTacker, ICredentialStoreService credentialStoreService, ILogger logger)
-            : this(serviceProvider, null, null, sonarQubeService, solutionTacker, credentialStoreService, logger, Dispatcher.CurrentDispatcher)
+            ISonarQubeService sonarQubeService, 
+            IActiveSolutionTracker solutionTacker, 
+            ICredentialStoreService credentialStoreService,
+            IProjectToLanguageMapper projectToLanguageMapper,
+            ILogger logger)
+            : this(serviceProvider,
```
rita-sonarsource, line 79: cosmetics + adding projectToLanguageMapper

src/Integration/LocalServices/ProjectSystemHelper.cs
```diff
@@ -85,7 +88,9 @@ public IEnumerable<Project> GetSolutionProjects()
             foreach (var hierarchy in EnumerateProjects(solution))
             {
                 Project Project = GetProject(hierarchy);
-                if (Project != null && !ProjectToLanguageMapper.GetLanguageForProject(Project).Equals(Language.Unknown))
+
+                if (Project != null && 
+                    !projectToLanguageMapper.GetAllBindingLanguagesForProject(Project).Any(x=> x.Equals(Language.Unknown)))
```
rita-sonarsource, line 93: Updated to support the new multiple signature, instead of the old one that was marked Obsolete.

src/Integration/LocalServices/ProjectSystemFilter.cs
```diff
@@ -62,10 +66,11 @@ public bool IsAccepted(DteProject project)
 
         #region Helpers
 
-        private static bool IsNotSupportedProject(DteProject project)
+        private bool IsNotSupportedProject(DteProject project)
         {
-            var language = ProjectToLanguageMapper.GetLanguageForProject(project);
-            return (language == null || !language.IsSupported);
+            var languages = projectToLanguageMapper.GetAllBindingLanguagesForProject(project);
```
rita-sonarsource, line 71: language couldn't have been null, it would've been Unknown.

src/Integration/Binding/NuGetBindingOperation.cs
```diff
@@ -39,18 +39,8 @@ internal class NuGetBindingOperation : INuGetBindingOperation
...
-            }
-
-            if (logger == null)
-            {
-                throw new ArgumentNullException(nameof(logger));
-            }
-
-            this.serviceProvider = serviceProvider;
-            this.logger = logger;
+            this.serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
```
rita-sonarsource, line 42: cosmetics

src/Integration/Binding/NuGetBindingOperation.cs
```diff
@@ -88,7 +78,9 @@ public bool InstallPackages(ISet<Project> projectsToBind, IProgress<FixedStepsPr
             var projectNugets = projectsToBind
                 .SelectMany(bindingProject =>
                 {
-                    var projectLanguage = ProjectToLanguageMapper.GetLanguageForProject(bindingProject);
+                    var projectToLanguageMapper = serviceProvider.GetMefService<IProjectToLanguageMapper>();
+                    var projectLanguages = projectToLanguageMapper.GetAllBindingLanguagesForProject(bindingProject);
+                    var projectLanguage = projectLanguages.FirstOrDefault();
```
rita-sonarsource, line 83: As in the other places, I'm only taking the first one if the code was using the Obsolete signature. 
I thought about changing it to multiple but I didn't want to touch the product code.

src/Integration/Binding/CSharpVBProjectBinder.cs
```diff
@@ -59,8 +63,10 @@ public CSharpVBProjectBinder(IServiceProvider serviceProvider, IFileSystem fileS
             {
                 throw new ArgumentNullException(nameof(serviceProvider));
             }
+
+            this.projectToLanguageMapper = projectToLanguageMapper ?? throw new ArgumentNullException(nameof(projectToLanguageMapper));
             this.fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
-            this.logger = logger;
+            this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
```
rita-sonarsource, line 69: Added for consistency with the other params

src/Integration/Binding/CFamilyProjectBinder.cs
```diff
@@ -33,16 +33,13 @@ namespace SonarLint.VisualStudio.Integration.Binding
 {
     internal class CFamilyProjectBinder : IProjectBinder
     {
+        private readonly IProjectToLanguageMapper projectToLanguageMapper;
         private readonly ILogger logger;
         private readonly IFileSystem fileSystem;
 
-        public CFamilyProjectBinder(IServiceProvider serviceProvider, ILogger logger, IFileSystem fileSystem)
+        public CFamilyProjectBinder(IProjectToLanguageMapper projectToLanguageMapper, ILogger logger, IFileSystem fileSystem)
```
rita-sonarsource, line 40: This class actually didn't use serviceProvider so I just removed it

src/Integration.UnitTests/ProjectToLanguageMapperTests.cs
```diff
@@ -29,90 +31,93 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
...
-#pragma warning restore CS0618 // Type or member is obsolete
+            MefTestHelpers.CheckTypeCanBeImported<ProjectToLanguageMapper, IProjectToLanguageMapper>(null, new[]
+            {
+                MefTestHelpers.CreateExport<IAbsoluteFilePathLocator>(Mock.Of<IAbsoluteFilePathLocator>())
+            });
         }
 
         [TestMethod]
-        public void Mapper_GetLanguage_UnknownLanguage_ReturnsUnknown()
+        public void GetAllBindingLanguagesForProject_NullProject_ArgumentNullException()
```
rita-sonarsource, line 43: This was previously `Mapper_GetLanguage_KnownLanguage_ArgChecks`

src/Integration.UnitTests/ProjectToLanguageMapperTests.cs
```diff
@@ -29,90 +31,93 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
...
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("dteProject");
         }
 
         [TestMethod]
-        public void Mapper_GetLanguage_KnownLanguage_ReturnsCorrectLanguage_VB()
+        [DataRow("")]
+        [DataRow("{F63A7EA2-4179-46BF-B9AE-E758BE4EC87C}")]
+        [DataRow("wibble")]
+        [DataRow("wibble;bibble")]
+        public void GetAllBindingLanguagesForProject_UnknownProjectType_UnknownProject(string projectTypeGuid)
```
rita-sonarsource, line 57: Converted to testCases

src/Integration.UnitTests/ProjectToLanguageMapperTests.cs
```diff
@@ -29,90 +31,93 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
...
         [TestMethod]
-        public void Mapper_AllBindingLanguages_CppProject()
+        public void GetAllBindingLanguagesForProject_SupportedProjectType_CPP_ReturnsCorrectLanguage()
         {
             CheckGetAllBindingsLanguages(ProjectSystemHelper.CppProjectKind, Language.Cpp, Language.C);
         }
 
         [TestMethod]
-        public void Mapper_AllBindingLanguages_UnknownProject()
+        public void GetAllBindingLanguagesForProject_OpenAsFolderProject_HasCmakeFiles_CFamilyLanguages()
```
rita-sonarsource, line 90: New test

src/Integration.UnitTests/MefServices/VsSessionHostTests.cs
```diff
@@ -62,46 +62,13 @@ public void TestInitialize()
...
             var mefExport2 = MefTestHelpers.CreateExport<ILogger>(new SonarLintOutputLogger(serviceProvider));
-            var mefModel = ConfigurableComponentModel.CreateWithExports(mefExport1, mefExport2);
+            var mefExport3 = MefTestHelpers.CreateExport<IProjectToLanguageMapper>(Mock.Of<IProjectToLanguageMapper>());
+            var mefModel = ConfigurableComponentModel.CreateWithExports(mefExport1, mefExport2, mefExport3);
             this.serviceProvider.RegisterService(typeof(SComponentModel), mefModel);
         }
 
         #region Tests
 
-        [TestMethod]
```
rita-sonarsource, line 71: This was not relevant anyway as it's a MEF CTOR.

src/Integration.UnitTests/LocalServices/ProjectSystemHelperTests.cs
```diff
@@ -79,16 +82,6 @@ public void MefCtor_CheckExports()
 
         #region Tests
 
-        [TestMethod]
-        public void ProjectSystemHelper_ArgCheck()
```
rita-sonarsource, line 83: MEF ctor and hence not relevant

src/Integration.UnitTests/Binding/ProjectBinderFactoryTests.cs
```diff
@@ -21,21 +21,33 @@
...
+        private Mock<IProjectToLanguageMapper> projectToLanguageMapper;
         private ProjectBinderFactory testSubject;
 
         [TestInitialize]
         public void TestInitialize()
         {
-            testSubject = new ProjectBinderFactory(Mock.Of<IServiceProvider>(), Mock.Of<ILogger>(), Mock.Of<IFileSystem>());
+            projectToLanguageMapper = new Mock<IProjectToLanguageMapper>();
+
+            var serviceProvider = new ConfigurableServiceProvider();
```
rita-sonarsource, line 43: Previously we would pass a Mock serviceProvider and all the GetService would simply return null.
To support GetMefService<> I had to use ConfigurableComponentModel and ConfigurableServiceProvider, as GetMefService uses extension methods. This unfortunately then lead to a bunch of errors that all these dependencies could not resolved, even though they're not really used.
Hence I'm registering them with mock values.

rita-sonarsource:
@duncanp-sonar please review.
The product has not changed, besides using the existing `GetAllBindingLanguagesForProject` instead of the obsolete `GetLanguageForProject`. It didn't feel right to add an obsolete method to a new interface, and by the time I've made all the changes I was committed to seeing it through...
There is a question of using FirstOrDefault vs First -- GetAllBindingLanguagesForProject will never return an empty list, or a null, it will always return a list of 1 with Unknown in it. Hence we can use First(). Let me know what you think.

The tests were horrible. I tried to use mocks wherever possible but in some cases it was just simpler to use the real class and keep them integration-like as they were before. Otherwise the changes would've been even larger.

On the bright side, we can bind cmake now... 🎉

duncanp-sonar:
Did you look at the bug that is causing the SonarCloud analysis to fail?

src/Integration.UnitTests/Binding/ProjectBinderFactoryTests.cs
```diff
@@ -102,10 +131,12 @@ public void Get_CppProject_CFamilyProjectBinderReturned()
         }
 
         [TestMethod]
-        public void Get_NonRoslynProject_CFamilyProjectBinderReturned()
+        public void Get_UnknownProject_CFamilyProjectBinderReturned()
         {
-            var project = new ProjectMock("c:\\foo.proj");
-            project.ProjectKind = "{" + Guid.NewGuid() + "}";
+            // todo: this is wrong and should be null
```
duncanp-sonar, line 136: This sounds like a bug. Is there a ticket to track it?

rita-sonarsource: I think somewhere along the path it works out, because we filter supported projects before calling this code (based on language). And since we support only roslyn and cpp ATM, if we got to this point and it's not roslyn, then it "must be" cpp. So this fallback was technically working.
That said, it's obviously wrong and if we support JS it would be broken. But at the moment "it works".

src/Integration/LocalServices/ProjectSystemFilter.cs
```diff
@@ -62,10 +66,11 @@ public bool IsAccepted(DteProject project)
...
+
+            return languages.Any(x => !x.IsSupported);
```
duncanp-sonar, line 73: Is this logic correct? It's saying _"the project is unsupported if any of the languages are not supported"._

Shouldn't it be _"the project is supported if any of the languages is supported"_ (i.e. it's only unsupported if all languages are unsupported)?

If the logic is wrong, then it suggests there's a missing test case.

src/Integration/LocalServices/ProjectSystemHelper.cs
duncanp-sonar: The logic looks suspect here too - the `Any` implies there could be multiple languages, one of which is `Unknown`. In practice that shouldn't happen - I guess we're assuming that if `Unknown` occurs it is the only entry.

 It feels as if there is a missing `HasSupportedLanguage` method.

src/Integration/ProfileConflicts/ConflictsManager.cs
duncanp-sonar: In general, I think `First` would be better; if the list is empty there is something wrong in our code, so it would be better to fail so we can find the error and fix it.

Here, the projects have already been filtered, so we shouldn't get this far if the language is Unknown, should we?

Separately, both the old and new logic look suspect for C/C++ - the project supports two languages, but we are only ever checking one of them here. However, this doesn't seem to be a problem in practice - I've manually tested that the current code correctly reports when either C/C++ ruleset file is missing, and it detects both cases. I'm guess this is done later by the language-specific binding code.

src/Integration/ProjectToLanguageMapper.cs
```diff
@@ -35,6 +47,19 @@ public static class ProjectToLanguageMapper
...
 
+        internal static Guid OpenAsFolderProject = new Guid("{66A2671D-8FB5-11D2-AA7E-00C04F688DDE}");
+
+        [ImportingConstructor]
+        public ProjectToLanguageMapper(IAbsoluteFilePathLocator absoluteFilePathLocator)
+        {
+            this.absoluteFilePathLocator = absoluteFilePathLocator;
+        }
+
+        IEnumerable<Language> IProjectToLanguageMapper.GetAllBindingLanguagesForProject(Project dteProject)
```
duncanp-sonar, line 58: What was the reason for implementing this interface explicitly?

src/Integration/ProjectToLanguageMapper.cs
```diff
@@ -46,37 +71,43 @@ public static class ProjectToLanguageMapper
...
+                {
+                    return Language.Cpp;
+                }
+            }
+
             return Language.Unknown;
         }
 
         /// <summary>
         /// Returns all of the supported Sonar languages for the sepcified project or Unknown
```
duncanp-sonar, line 105: Miscellaneous: pre-existing typo "sepcified".

This doc comment now belongs on the interface method.

src/Integration.Vsix/Commands/ProjectTestPropertySetCommand.cs
```diff
@@ -85,7 +82,7 @@ protected override void QueryStatusInternal(OleMenuCommand command)
                                           .GetSelectedProjects()
                                           .ToList();
 
-            if (projects.Any() && projects.All(x => ProjectToLanguageMapper.GetLanguageForProject(x).IsSupported))
+            if (projects.Any() && projects.All(x => projectToLanguageMapper.GetAllBindingLanguagesForProject(x).All(l => l.IsSupported)))
```
duncanp-sonar, line 85: Another logic question (here, and in line 64 above) - should this be `All` or `Any`?
As it happens, both currently work because of the way `GetAllBindingLanguagesForProject` currently works, but the callers shouldn't be relying on that.

A helper method would avoid the problem of having to get the logic right in every caller.

src/Integration.Vsix/Commands/ProjectSonarLintMenuCommand.cs
```diff
@@ -60,7 +57,7 @@ protected override void QueryStatusInternal(OleMenuCommand command)
                                           .GetSelectedProjects()
                                           .ToList();
 
-            if (projects.Any() && projects.All(x => ProjectToLanguageMapper.GetLanguageForProject(x).IsSupported))
+            if (projects.Any() && projects.All(x => projectToLanguageMapper.GetAllBindingLanguagesForProject(x).All(l => l.IsSupported)))
```
duncanp-sonar, line 60: `Any`?

src/Integration.Vsix/Commands/ProjectExcludePropertyToggleCommand.cs
```diff
@@ -104,7 +101,7 @@ protected override void QueryStatusInternal(OleMenuCommand command)
                                           .GetSelectedProjects()
                                           .ToList();
 
-            if (projects.Any() && projects.All(x => ProjectToLanguageMapper.GetLanguageForProject(x).IsSupported))
+            if (projects.Any() && projects.All(x => projectToLanguageMapper.GetAllBindingLanguagesForProject(x).All(l => l.IsSupported)))
```
duncanp-sonar, line 104: `Any`?

duncanp-sonar (commented):
Nice.

Two things;
* a question about the use of `All` vs `Any` for working out whether a project is supported which affects multiple locations, and
* a new TODO that looks like it's describing a bug

src/Integration.UnitTests/Binding/ProjectBinderFactoryTests.cs
duncanp-sonar: Ok. This isn't a new test, but it's clearly the wrong behaviour. If the language is unknown / not Roslyn/CFamily the factory should return null.

Having a test that checks the behaviour is wrong isn't very useful or discoverable.
It would be better to have an [Ignored] test that checks for the desired behaviour and fails, ideally with a ticket to fix it.

src/Integration/ProjectToLanguageMapper.cs
```diff
@@ -93,11 +100,14 @@ private Language GetLanguageForProject(Project dteProject)
...
             }
 
-            if (projectKind == OpenAsFolderProject)
+            var isOpenAsFolder = folderWorkspaceService.IsFolderWorkspace();
+
+            if (isOpenAsFolder)
             {
-                var cmakeProjectFile = absoluteFilePathLocator.Locate("CMakeLists.txt");
+                var rootDirectory = folderWorkspaceService.FindRootDirectory();
+                var isCMake = fileSystem.Directory.EnumerateFiles(rootDirectory, "CMakeLists.txt", SearchOption.AllDirectories).Any();
```
rita-sonarsource, line 108: @duncanp-sonar I decided to go with fileSystem search. I opened sonar-clang multiple times and absoluteFilePathLocator kept returning null, even after waiting for a long time for VS to finish loading. Then once it did work. And then it stopped working again.

All in all, the behavior was inconsistent so I switched to using the file system. We had a similar problem with TsConfigLocator, we started by using IAbsoluteFilePathLocator and switched to file system.
https://github.com/SonarSource/sonarlint-visualstudio/pull/2387

I think we should revisit using IAbsoluteFilePathLocator at all, we use it in hotspots and it's highly inconsistent.

PR #2579, opened 2021-08-19
Support analyzing additional header extensions

### LINKED ISSUE #2577: [CMake] Support analyzing additional header file extensions (.hpp, .hh, .hxx)

rita-sonarsource:
Fixes #2577

### DIFF
--- a/src/CFamily.UnitTests/CMake/CompilationConfigProviderTests.cs
+++ b/src/CFamily.UnitTests/CMake/CompilationConfigProviderTests.cs
@@ -258,7 +258,7 @@ namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
         [TestMethod]
         public void GetConfig_HeaderFile_CodeWithSameNameExactPath_GivenPriorityOverDifferentPath()
         {
-            const string headerFilePath = "c:\\test\\a.h";
+            const string headerFilePath = "c:\\test\\a.hxx";
 
             var compilationDatabaseContent = new[]
             {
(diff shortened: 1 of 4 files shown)


src/CFamily.UnitTests/CMake/CompilationConfigProviderTests.cs
```diff
@@ -258,7 +258,7 @@ public void GetConfig_HeaderFile_CodeWithSameNameDifferentPath_ReturnsCodeFileEn
         [TestMethod]
         public void GetConfig_HeaderFile_CodeWithSameNameExactPath_GivenPriorityOverDifferentPath()
         {
-            const string headerFilePath = "c:\\test\\a.h";
+            const string headerFilePath = "c:\\test\\a.hxx";
```
rita-sonarsource, line 261: To simplify things, I've just varied the existing tests to use the different header file extensions.
There is also an unchanged test that uses .h.

src/CFamily/CMake/CompilationConfigProvider.cs
```diff
@@ -121,8 +121,13 @@ public CompilationDatabaseEntry GetConfig(string filePath)
             }
         }
 
-        private bool IsHeaderFile(string filePath) => 
-            Path.GetExtension(filePath).Equals(".h", StringComparison.OrdinalIgnoreCase);
+        private bool IsHeaderFile(string filePath)
```
duncanp-sonar, line 124: Why not add this method to the `CFamilyShared` class?
It would also be easier to test all of the valid extensions, casing etc.

duncanp-sonar (commented):
Just a comment about putting the new method in `CFamilyShared`.

duncanp-sonar (approved):

PR #2581, opened 2021-08-21
Fix CMake suppressions

### LINKED ISSUE #2557: Fix suppressions mechanism
### Description

`SonarQubeIssuesProvider.GetSuppressedIssues` asserts because CMake file paths have forward-slashes.

The assertion indicates that there is a problem, but it's not clear what the correct fix is e.g. change the suppression matching, or change the CMake file path.

We should check what we get back from SonarQube/SonarCloud for analysed CMake projects.

rita-sonarsource:
Fixes #2557

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueToAnalysisIssueConverterTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueToAnalysisIssueConverterTests.cs
@@ -436,6 +436,24 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
         }
 
         [TestMethod]
+        [Description("Regression test for https://github.com/SonarSource/sonarlint-visualstudio/issues/2557")]
+        [DataRow("", "")] // empty should not throw
+        [DataRow("a.txt", "a.txt")] // not-rooted should stay the same
+        [DataRow("c:\\a.txt", "c:\\a.txt")]
+        [DataRow("c:/a.txt", "c:\\a.txt")]
+        [DataRow("c:/a/b/c.txt", "c:\\a\\b\\c.txt")]
+        [DataRow("c:/a\\b/c.txt", "c:\\a\\b\\c.txt")]
+        public void Convert_FilePath_QualifiedFilePath(string originalPath, string expectedPath)
+        {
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueToAnalysisIssueConverterTests.cs
```diff
@@ -435,6 +435,24 @@ public void Convert_HasMessageParts_QualifiedFilePath(string originalPath, strin
...
 
+        [TestMethod]
+        [Description("Regression test for https://github.com/SonarSource/sonarlint-visualstudio/issues/2557")]
+        [DataRow("", "")] // empty should not throw
+        [DataRow("a.txt", "a.txt")] // not-rooted should stay the same
+        [DataRow("c:\\a.txt", "c:\\a.txt")]
+        [DataRow("c:/a.txt", "c:\\a.txt")]
+        [DataRow("c:/a/b/c.txt", "c:\\a\\b\\c.txt")]
+        [DataRow("c:/a\\b/c.txt", "c:\\a\\b\\c.txt")]
+        public void Convert_FilePath_QualifiedFilePath(string originalPath, string expectedPath)
```
rita-sonarsource, line 446: We actually do the same thing for the secondary locations, test cases are copy-pasted from `Convert_HasMessageParts_QualifiedFilePath` above

src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
```diff
@@ -122,7 +122,7 @@ public IAnalysisIssue Convert(Message cFamilyIssue, string sqLanguage, ICFamilyR
                 severity: Convert(defaultSeverity),
                 type: Convert(defaultType),
 
-                filePath: cFamilyIssue.Filename,
+                filePath: Path.IsPathRooted(cFamilyIssue.Filename) ? Path.GetFullPath(cFamilyIssue.Filename) : cFamilyIssue.Filename,
```
rita-sonarsource, line 125: Similar to what we do in [secondary locations](https://github.com/SonarSource/sonarlint-visualstudio/blob/e1f9b8308698fd1133a405835d784304bd0ddfba/src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs#L160)

Issue #2585, opened 2021-08-24
Refactor request creation to be async


duncanp-sonar:
### Description

Fetching the required data to create a request can be time-consuming, so should async.

PR #2586, opened 2021-08-24
CLangAnalyzer: moved request-creation to background thread

duncanp-sonar:
Part of #2585

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
@@ -57,13 +57,14 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         public void ExecuteAnalysis_RequestCannotBeCreated_NoAnalysis()
         {
             var analysisOptions = new CFamilyAnalyzerOptions();
-            var requestFactory = CreateRequestFactory("path", analysisOptions, null);
+            var requestFactoryContext = new RequestFactoryContext("path", analysisOptions, null);
 
-            var testSubject = CreateTestableAnalyzer(requestFactory: requestFactory.Object);
-            ExecuteAndWaitForCompletion(testSubject, "path", "charset", new[] { AnalysisLanguage.CFamily }, Mock.Of<IIssueConsumer>(), analysisOptions, CancellationToken.None);
+            var testSubject = CreateTestableAnalyzer(requestFactory: requestFactoryContext.Factory.Object);
+            testSubject.ExecuteAnalysis("path", "charset", new[] { AnalysisLanguage.CFamily }, Mock.Of<IIssueConsumer>(), analysisOptions, CancellationToken.None);
 
+            requestFactoryContext.WaitForFactoryCallToComplete();
+            requestFactoryContext.Factory.Verify(x => x.TryGet("path", analysisOptions), Times.Once);
             testSubject.SubProcessExecutedCount.Should().Be(0);
-            requestFactory.Verify(x => x.TryGet("path", analysisOptions), Times.Once);
         }
 
         [TestMethod]
@@ -72,15 +73,16 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         public void ExecuteAnalysis_RequestCannotBeCreated_NotPCH_LogOutput(bool isNullOptions)
         {
             var analysisOptions = isNullOptions ? null : new CFamilyAnalyzerOptions { CreatePreCompiledHeaders = false };
-            var requestFactory = CreateRequestFactory("path", analysisOptions, null);
+            var requestFactoryContext = new RequestFactoryContext("path", analysisOptions, null);
             var testLogger = new TestLogger();
 
             var testSubject = CreateTestableAnalyzer(
-                requestFactory: requestFactory.Object,
+                requestFactory: requestFactoryContext.Factory.Object,
                 logger: testLogger);
 
-            ExecuteAndWaitForCompletion(testSubject, "path", "charset", new[] { AnalysisLanguage.CFamily }, Mock.Of<IIssueConsumer>(), analysisOptions, CancellationToken.None);
+            testSubject.ExecuteAnalysis("path", "charset", new[] { AnalysisLanguage.CFamily }, Mock.Of<IIssueConsumer>(), analysisOptions, CancellationToken.None);
 
+            requestFactoryContext.WaitForFactoryCallToComplete();
             testLogger.AssertOutputStringExists(string.Format(CFamilyStrings.MSG_UnableToCreateConfig, "path"));
         }
 
@@ -88,15 +90,16 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         public void ExecuteAnalysis_RequestCannotBeCreated_PCH_NoLogOutput()
         {
             var analysisOptions = new CFamilyAnalyzerOptions {CreatePreCompiledHeaders = true};
-            var requestFactory = CreateRequestFactory("path", analysisOptions, null);
+            var requestFactoryContext = new RequestFactoryContext("path", analysisOptions, null);
             var testLogger = new TestLogger();
 
             var testSubject = CreateTestableAnalyzer(
-                requestFactory: requestFactory.Object,
+                requestFactory: requestFactoryContext.Factory.Object,
                 logger: testLogger);
 
-            ExecuteAndWaitForCompletion(testSubject, "path", "charset", new[] { AnalysisLanguage.CFamily }, Mock.Of<IIssueConsumer>(), analysisOptions, CancellationToken.None);
+            testSubject.ExecuteAnalysis("path", "charset", new[] { AnalysisLanguage.CFamily }, Mock.Of<IIssueConsumer>(), analysisOptions, CancellationToken.None);
 
+            requestFactoryContext.WaitForFactoryCallToComplete();
             testLogger.AssertNoOutputMessages();
         }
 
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -24,6 +24,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
+using Microsoft.CodeAnalysis;
```
duncanp-sonar, line 27: Pesky VS trying to be helpful.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -120,22 +140,18 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
-            TriggerAnalysisAsync(request, consumer, statusNotifier, cancellationToken)
-                .Forget(); // fire and forget
-
         protected /* for testing */ virtual void CallSubProcess(Action<Message> handleMessage, IRequest request, ISonarLintSettings settings, ILogger logger, CancellationToken cancellationToken)
         {
             ExecuteSubProcess(handleMessage, request, new ProcessRunner(settings, logger), logger, cancellationToken, fileSystem);
         }
 
-        internal /* for testing */ async Task TriggerAnalysisAsync(IRequest request, IIssueConsumer consumer, IAnalysisStatusNotifier statusNotifier, CancellationToken cancellationToken)
+        private async Task RunAnalysisAsync(IRequest request, IIssueConsumer consumer, IAnalysisStatusNotifier statusNotifier, CancellationToken cancellationToken)
```
duncanp-sonar, line 154: I modified the existing tests so they call Execute instead.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -102,6 +102,15 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
         }
 
         public void ExecuteAnalysis(string path,
+            IEnumerable<AnalysisLanguage> detectedLanguages,
+            IIssueConsumer consumer,
+            IAnalyzerOptions analyzerOptions,
+            IAnalysisStatusNotifier statusNotifier,
+            CancellationToken cancellationToken) =>
+            TriggerAnalysisAsync(path, detectedLanguages, consumer, analyzerOptions, statusNotifier, cancellationToken)
```
duncanp-sonar, line 110: The product changes are quite simple:
* previously, the request creation was done before the analysis was triggered on a separate thread.
* now, Execute immediately triggers a separate thread to do all of the work, including creating the request.

Unfortunately, changing the tests wasn't as straightforward.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -102,6 +102,15 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
+                .Forget(); // fire and forget
+
+        internal /* for testing */ async Task TriggerAnalysisAsync(string path,
```
duncanp-sonar, line 113: The old code had an "internal testing" method with the same name but a different signature (a lower-level one that accepted a request).
I changed the existing tests to call this higher-level method instead (I wanted to make this method private, but there wasn't an easy way to fix the tests if I did that).

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -110,6 +119,17 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
 
+            var request = await TryCreateRequestAsync(path, analyzerOptions);
+
+            if (request != null)
+            {
+                await RunAnalysisAsync(request, consumer, statusNotifier, cancellationToken);
+            }
+        }
+
+        protected virtual /* for testing */ async Task<IRequest> TryCreateRequestAsync(string path, IAnalyzerOptions analyzerOptions)
```
duncanp-sonar, line 130: This new method is virtual so the testable subclass can override it and provide a synchronisation method so tests running on the "main" thread can wait until this method (on another thread) has finished.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -407,7 +458,10 @@ public void CallSubProcess(Action<Message> handleMessage, IRequest request, ISon
             }
 
             public void WaitUntilSubProcessCalledByAnalyzer()
-                => callbackFromCLangReceived.WaitOne(20000);
+            {
+                var result = callbackFromCLangReceived.WaitOne(20000);
+                result.Should().BeTrue("Timed out waiting for subprocess to be called");
```
duncanp-sonar, line 463: Previously, the test would still have passed if the callback was never received.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -382,6 +415,24 @@ protected override void TriggerAnalysis(IRequest request, IIssueConsumer consume
                     onCallSubProcess(handleMessage, request, settings, logger, cancellationToken);
                 }
             }
+
+            protected override Task<IRequest> TryCreateRequestAsync(string path, IAnalyzerOptions analyzerOptions)
```
duncanp-sonar, line 419: Overridden to provide a synchronisation mechanism so that tests can reliably wait until the factory code has completed.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -348,31 +370,42 @@ private static Mock<IRequestFactoryAggregate> CreateRequestFactory(string filePa
...
                 IAnalysisStatusNotifier analysisStatusNotifier, ILogger logger, 
                 ICFamilyIssueToAnalysisIssueConverter cFamilyIssueConverter, IRequestFactoryAggregate requestFactory)
                 : base(telemetryManager, settings, analysisStatusNotifier, cFamilyIssueConverter, requestFactory, logger)
             {}
 
-            protected override void TriggerAnalysis(IRequest request, IIssueConsumer consumer, IAnalysisStatusNotifier analysisStatusNotifier, CancellationToken cancellationToken)
-            {
-                TriggerAnalysisCallCount++;
-            }
-
```
duncanp-sonar, line 372: Previously this method did nothing except count the call i.e. the real code wasn't called.

Now more of the "real" code is executed by the tests.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -96,21 +101,25 @@ public void ExecuteAnalysis_RequestCannotBeCreated_PCH_NoLogOutput()
...
+
             testLogger.AssertNoOutputMessages();
         }
 
         [TestMethod]
         public void ExecuteAnalysis_RequestCanBeCreated_AnalysisIsTriggered()
         {
             var analysisOptions = new CFamilyAnalyzerOptions();
-            var requestFactory = CreateRequestFactory("path", analysisOptions, Mock.Of<IRequest>());
+            var request = CreateRequest("path");
```
duncanp-sonar, line 113: The request now needs to be valid i.e. have a non-null `RequestContext` property.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -96,21 +101,25 @@ public void ExecuteAnalysis_RequestCannotBeCreated_PCH_NoLogOutput()
...
+            var requestFactory = CreateRequestFactory("path", analysisOptions, request);
+
 
             var testSubject = CreateTestableAnalyzer(
                 requestFactory: requestFactory.Object);
 
-            testSubject.ExecuteAnalysis("path", "charset", new[] { AnalysisLanguage.CFamily }, Mock.Of<IIssueConsumer>(), analysisOptions, CancellationToken.None);
```
duncanp-sonar, line 111: Previously this method was testing work that was done on the "main" thread before the background thread was started.
Now everything is done on the background thread, so we have to do some extra work to make sure the background thread has finished before we check the results.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -61,8 +61,13 @@ public void ExecuteAnalysis_RequestCannotBeCreated_NoAnalysis()
             var testSubject = CreateTestableAnalyzer(requestFactory: requestFactory.Object);
             testSubject.ExecuteAnalysis("path", "charset", new[] { AnalysisLanguage.CFamily }, Mock.Of<IIssueConsumer>(), analysisOptions, CancellationToken.None);
 
-            testSubject.TriggerAnalysisCallCount.Should().Be(0);
+            testSubject.WaitForFactoryCallToComplete();
+
             requestFactory.Verify(x => x.TryGet("path", analysisOptions), Times.Once);
+
+            // TODO - modify check to be more reliable
+            Thread.Sleep(400); // delay in case the background thread has gone on to call the subprocess
```
duncanp-sonar, line 69: I couldn't think of simple way of making this reliable without more extensive changes.
I'll revisit it when I do the changes to make the request factories asynchronous.

duncanp-sonar:
This is an interim commit; more work will be done on the tests and the CLangAnalyzer in the next commit as the request factories are made asynchronous.

I've managed to remove some of the complexity/overrides/internals-for-testing from the CLangAnalyzer, but I expect further changes will be required. Any suggestions gratefully received; I'll either include them in this PR or the next one.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
rita-sonarsource: Perhaps it's worth having the tests call this method instead so that they could await on it.
This would remove the need for `WaitForFactoryCallToComplete` and simplify the tests.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -61,8 +61,13 @@ public void ExecuteAnalysis_RequestCannotBeCreated_NoAnalysis()
             var testSubject = CreateTestableAnalyzer(requestFactory: requestFactory.Object);
             testSubject.ExecuteAnalysis("path", "charset", new[] { AnalysisLanguage.CFamily }, Mock.Of<IIssueConsumer>(), analysisOptions, CancellationToken.None);
 
-            testSubject.TriggerAnalysisCallCount.Should().Be(0);
+            testSubject.WaitForFactoryCallToComplete();
```
rita-sonarsource, line 64: I think it would be easier to read the tests if we were to call the internal async method.

rita-sonarsource (approved):

rita-sonarsource:
I think we should start taking bits out of CLangAnalyzer, even the product code is getting a bit difficult to follow.
`ExecuteSubProcess` is a method that could be taken out as-is, as it's already a static and receives everything in the method signature. Something like `SubProcessCaller` perhaps.

Issue #2198, opened 2021-02-25
[CMake] Collect analysis telemetry


rita-g-sonarsource:
#### Description

TBD format + metrics.

TODO:
- [ ] Number of daily opened cmake projects vs vcxproj? (just vcx doesn't mean that the file is analyzable, i.e. unreal engine projects are also vcx)

PR #2589, opened 2021-08-25
Move compilation database file existence check

rita-sonarsource:
Part of #2198

### DIFF
--- a/src/CFamily/CMake/CompilationDatabaseLocator.cs
+++ b/src/CFamily/CMake/CompilationDatabaseLocator.cs
@@ -31,7 +31,7 @@ using SonarLint.VisualStudio.Integration.Helpers;
 
 namespace SonarLint.VisualStudio.CFamily.CMake
 {
-    internal interface ICompilationDatabaseLocator
+    public interface ICompilationDatabaseLocator
     {
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/CFamily/CMake/CompilationDatabaseLocator.cs
```diff
@@ -86,9 +86,18 @@ public string Locate()
             var cmakeSettingsFullPath = Path.GetFullPath(Path.Combine(rootDirectory, CMakeSettingsFileName));
             var activeConfiguration = buildConfigProvider.GetActiveConfig(rootDirectory);
 
-            return fileSystem.File.Exists(cmakeSettingsFullPath)
+            var compilationDatabaseLocation = fileSystem.File.Exists(cmakeSettingsFullPath)
                 ? GetConfiguredLocation(cmakeSettingsFullPath, activeConfiguration, rootDirectory)
                 : GetDefaultLocation(rootDirectory, activeConfiguration);
+
+            if (fileSystem.File.Exists(compilationDatabaseLocation))
```
rita-sonarsource, line 93: I didn't want to repeat this logic in telemetry, so it felt more sense to return null if the file does not exist.

duncanp-sonar (approved):

PR #2590, opened 2021-08-25
Extract CMake project type indication logic

rita-sonarsource:
Part of #2198

### DIFF
--- a/src/Core/CFamily/ICFamilyProjectTypeIndicator.cs
+++ b/src/Core/CFamily/ICFamilyProjectTypeIndicator.cs
@@ -20,8 +20,16 @@
 
 namespace SonarLint.VisualStudio.Core.CFamily
 {
+    /// <summary>
+    /// Returns information if the currently open solution is one of the supported cfamily project types.
+    /// Methods return false if there is no open solution.
+    /// </summary>
     public interface ICFamilyProjectTypeIndicator
     {
+        /// <summary>
+        /// Returns true if the currently open solution is a CMake project.
+        /// Returns false if the solution is not CMake, or if there is no open solution.
+        /// </summary>
         bool IsCMake();
     }
 }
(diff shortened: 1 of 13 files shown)


src/Core/CFamily/ICFamilyProjectTypeIndicator.cs
```diff
@@ -0,0 +1,27 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Core.CFamily
+{
+    public interface ICFamilyProjectTypeIndicator
+    {
+        bool IsCMake();
```
rita-sonarsource, line 25: Not sure yet about the API, but for now this is what I need for telemetry as well.
I will probably add another "IsVcx()" method for telemetry.

duncanp-sonar: Missing doc comments for the interface and method.
What's the expected behaviour if there isn't an open solution - false?

src/Integration.UnitTests/ProjectToLanguageMapperTests.cs
```diff
@@ -90,36 +87,27 @@ public void GetAllBindingLanguagesForProject_SupportedProjectType_CPP_ReturnsCor
...
-        [DataRow("c:\\some directory\\sub\\CMakeLists.txt")]
-        [DataRow("c:\\some directory\\sub\\folder\\CMakeLists.txt")]
-        public void GetAllBindingLanguagesForProject_OpenAsFolderProject_HasCmakeFiles_CFamilyLanguages(string cmakeListsLocation)
+        public void GetAllBindingLanguagesForProject_CMakeProject_CFamilyLanguages()
         {
-            var folderWorkspaceService = SetupOpenAsFolder("c:\\some directory");
-            var fileSystem = new MockFileSystem();
-            fileSystem.AddDirectory("c:\\some directory");
-            fileSystem.AddFile(cmakeListsLocation, new MockFileData(""));
+            var folderWorkspaceService = SetupCMakeProject(true);
```
duncanp-sonar, line 92: The variable name 'folderWorkspaceService' needs to be updated in all of the tests in this class.

duncanp-sonar (approved):
Looks good, just comments about comments and out-of-date variable names.

PR #2591, opened 2021-08-25
Add CFamilyProjectTypes to telemetry payload

rita-sonarsource:
Part of #2198

### DIFF
--- a/src/Integration/Telemetry/Payload/CFamilyProjectTypes.cs
+++ b/src/Integration/Telemetry/Payload/CFamilyProjectTypes.cs
@@ -1,0 +1,33 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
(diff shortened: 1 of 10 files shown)


src/Integration/Telemetry/Payload/CFamilyProjectTypes.cs
```diff
@@ -0,0 +1,33 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Newtonsoft.Json;
+
+namespace SonarLint.VisualStudio.Integration.Telemetry.Payload
+{
+    public sealed class CFamilyProjectTypes
```
rita-sonarsource, line 25: After discussion with Marco yesterday, CMake is the priority. Vcx was not in the MMF so it is nice to have, but he said that if it's low effort he'd want that too.
I'll add that after we're done with cmake if there is nothing else to do.

duncanp-sonar (approved):

PR #2593, opened 2021-08-25
Add CFamilyTelemetryManager

### LINKED ISSUE #2198: [CMake] Collect analysis telemetry

rita-sonarsource:
Fixes #2198


src/CFamily/CMake/CompilationDatabaseLocator.cs
```diff
@@ -25,21 +25,13 @@
...
 using Microsoft.VisualStudio;
 using Newtonsoft.Json;
+using SonarLint.VisualStudio.Core.CFamily;
 using SonarLint.VisualStudio.Infrastructure.VS;
 using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.Integration.Helpers;
 
 namespace SonarLint.VisualStudio.CFamily.CMake
 {
-    public interface ICompilationDatabaseLocator
```
rita-sonarsource, line 34: Moved to Core so I could use it in Integration

src/Integration.UnitTests/Telemetry/CFamilyTelemetryManagerTests.cs
```diff
@@ -0,0 +1,278 @@
...
+using Moq;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Integration.Telemetry;
+using SonarLint.VisualStudio.Integration.Telemetry.Payload;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.Telemetry
+{
+    [TestClass]
+    public class CFamilyTelemetryManagerTests
```
rita-sonarsource, line 33: I wrote the tests with ActiveSolutionChanged event before knowing it doesn't work. I didn't change them to raise BeforeClosed since the code still technically listens to both.

src/Integration.Vsix/SonarLintTelemetryPackage.cs
```diff
@@ -68,6 +69,8 @@ private async System.Threading.Tasks.Task InitOnUIThreadAsync()
                 telemetryManager = await this.GetMefServiceAsync<ITelemetryManager>();
                 logger.WriteLine(Resources.Strings.Telemetry_InitializationComplete);
 
+                await this.GetMefServiceAsync<ICFamilyTelemetryManager>();
```
rita-sonarsource, line 72: Initializing the event listeners

src/Integration/Telemetry/CFamilyTelemetryManager.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System;
```
rita-sonarsource, line 1: Forgot to add the header, I'll add it before committing

src/Integration.Vsix/SonarLintTelemetryPackage.cs
duncanp-sonar: This deserves a comment; it's not obvious why the class is importing a type it isn't using - 
 the new component is completely independent, but we need it to be created somehow, so we're importing it here.

duncanp-sonar (approved):

PR #2613, opened 2021-08-30
Add vcx telemetry

### LINKED ISSUE #2599: Collect telemetry for Vcx projects
### Description

Following #2198, collect telemetry about opened vcx projects.

```
{
   "cfamily_project_types" : {
       "vcx_analyzable" : true/false, // was an analyzable vcx project opened since last sending a ping
       "vcx_non_analyzable" : true/false // was a non-analyzable vcx project opened since last sending a ping
   }
}
```

Notes:
- Vcx project can be non-analyzable, for example #2127

rita-sonarsource:
Fixes #2599

### DIFF
--- a/src/Integration.Vsix/CFamily/VcxProject/VcxProjectTypeIndicator.cs
+++ b/src/Integration.Vsix/CFamily/VcxProject/VcxProjectTypeIndicator.cs
@@ -1,0 +1,76 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.VCProjectEngine;
+using SonarLint.VisualStudio.Core.CFamily;
+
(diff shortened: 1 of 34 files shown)


src/Integration.Vsix/CFamily/VcxProject/VcxProjectTypeIndicator.cs
```diff
@@ -0,0 +1,76 @@
...
+
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.VCProjectEngine;
+using SonarLint.VisualStudio.Core.CFamily;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject
+{
+    [Export(typeof(IVcxProjectTypeIndicator))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class VcxProjectTypeIndicator : IVcxProjectTypeIndicator
```
rita-sonarsource, line 29: Perhaps this class should be inside the telemetry module, as I don't see how it can be used by anything else; the api is too specific, but it's because I didn't want to expose methods that work with dte objects.

src/Integration/Telemetry/CFamilyTelemetryManager.cs
```diff
@@ -101,6 +96,41 @@ private void UpdateTelemetry()
             }
         }
 
+        private void CollectCMakeUsage()
```
rita-sonarsource, line 99: Extracted as-is to function

src/Integration/Telemetry/CFamilyTelemetryManager.cs
```diff
@@ -101,6 +96,41 @@ private void UpdateTelemetry()
...
+            }
+            else
+            {
+                telemetryDataRepository.Data.CFamilyProjectTypes.IsCMakeNonAnalyzable = true;
+            }
+
+            telemetryDataRepository.Save();
+        }
+
+        private void CollectVcxUsage()
```
rita-sonarsource, line 118: new code

src/Integration/Telemetry/CFamilyTelemetryManager.cs
```diff
@@ -101,6 +96,41 @@ private void UpdateTelemetry()
...
+        {
+            var vcxProjectTypes = vcxProjectTypeIndicator.GetProjectTypes();
+
+            if (vcxProjectTypes.HasAnalyzableVcxProjects)
+            {
+                telemetryDataRepository.Data.CFamilyProjectTypes.IsVcxAnalyzable = true;
+            }
+            if (vcxProjectTypes.HasNonAnalyzableVcxProjects)
```
rita-sonarsource, line 126: In theory a solution can contain both analyzable and non analyzable, although it might not be a real use case in practice. Still, it's worth capturing.

duncanp-sonar (approved):

PR #2615, opened 2021-08-30
Don't cache null envVars

### LINKED ISSUE #2614: [CMake] Fetching environment variables fails intermittently
#### Description

On a slow VM machine, fetching environment variables sometimes fails with a timeout. When that happens, any subsequent analysis fail as well, as we cache the null result and don't attempt to retrieve it again.

Log:
```
DEBUG: [CMake:EnvVars] [Thread id: 12] Cache miss. Script params: ""
DEBUG: [CMake:EnvVars] [Thread id: 12] 	Waiting to acquire lock...
DEBUG: [CMake:EnvVars] [Thread id: 12] 	Acquired lock.
DEBUG: [CMake:VsDevCmd] [Thread id: 12] VsDevCmd location: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\Tools\VsDevCmd.bat
DEBUG: [CMake:VsDevCmd] [Thread id: 12] Started process. Id: 5704
[CMake] Failed to fetch VsDevCmd settings: timed out.
[CMake] Failed to fetch VsDevCmd settings: no settings found.
Unable to retrieve the configuration for file ... Check the file is part of a supported project type in the current solution.
```

rita-sonarsource:
Fixes #2614

### DIFF
--- a/src/CFamily/CMake/EnvironmentVarsProvider.cs
+++ b/src/CFamily/CMake/EnvironmentVarsProvider.cs
@@ -91,7 +91,11 @@ namespace SonarLint.VisualStudio.CFamily.CMake
                 }
 
                 var newSettings = await vsDevCmdProvider.GetAsync(vsDevCmdScriptParams);
-                cachedVsDevCmdSettings[vsDevCmdScriptParams] = newSettings;
+
+                if (newSettings != null)
+                {
+                    cachedVsDevCmdSettings[vsDevCmdScriptParams] = newSettings;
+                }
 
                 return newSettings;
             }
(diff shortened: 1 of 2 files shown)


src/CFamily/CMake/EnvironmentVarsProvider.cs
```diff
@@ -91,7 +91,11 @@ public EnvironmentVarsProvider(IVsDevCmdEnvironmentProvider vsDevCmdProvider, IL
                 }
 
                 var newSettings = await vsDevCmdProvider.GetAsync(vsDevCmdScriptParams);
-                cachedVsDevCmdSettings[vsDevCmdScriptParams] = newSettings;
+
+                if (newSettings != null)
```
rita-sonarsource, line 95: Let's discuss this fix tomorrow.
We could also potentially increase the timeout of the process, but I think this would be a better approach.

PR #2617, opened 2021-08-31
Reduce verbosity of invalid CMake settings message

### LINKED ISSUE #2547: Log output is too verbose when CMake settings/compilation DB are invalid
### Description

The new protocol handles invalid 'CMakeSettings.json' and 'compilation_commands.json' files and dumps the full error stack to the output window.
The call stack isn't very useful in this case. Also, the "create request" code is called frequently (e.g. to update the pre-compiled header when a new document is selected) so the output window quickly contains a lot of output.

#### Current - invalid CMakeSettings.json
image

#### Current - invalid compilation_commands.json
image

#### Proposed output
Just log the message:
image

#### Visual Studio settings editor
For comparison, the VS editor just shows the message:
image

duncanp-sonar:
Fixes #2547

### DIFF
--- a/src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
+++ b/src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
@@ -117,7 +117,7 @@ namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
         [TestMethod]
         public void Locate_FailedToParseCMakeSettings_Null()
         {
-            const string invalidJon = "invalid json";
+            const string invalidJson = "invalid json";
             var expectedMessage = GetExpectedDeserializationMessage();
             var configProvider = CreateConfigProvider("my config");
             var cmakeSettingsLocation = GetCmakeSettingsLocation(RootDirectory);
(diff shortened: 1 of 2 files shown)


src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
```diff
@@ -117,22 +117,37 @@ public void Locate_FailedToReadCMakeSettings_NonCriticalException_Null()
         [TestMethod]
         public void Locate_FailedToParseCMakeSettings_Null()
         {
+            const string invalidJon = "invalid json";
+            var expectedMessage = GetExpectedDeserializationMessage();
```
duncanp-sonar, line 121: I didn't want to search for a hard-coded string, so the easiest thing to do seemed to be to generate the expected exception message by deserializing the same invalid json and catching the exception.

src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
```diff
@@ -117,22 +117,37 @@ public void Locate_FailedToReadCMakeSettings_NonCriticalException_Null()
...
             var configProvider = CreateConfigProvider("my config");
             var cmakeSettingsLocation = GetCmakeSettingsLocation(RootDirectory);
 
             var fileSystem = new Mock<IFileSystem>();
             fileSystem.Setup(x => x.File.Exists(cmakeSettingsLocation)).Returns(true);
             fileSystem
                 .Setup(x => x.File.ReadAllText(cmakeSettingsLocation))
-                .Returns("invalid json");
+                .Returns(invalidJon);
```
rita-sonarsource, line 129: Typo in "jon"

rita-sonarsource (approved):

PR #2621, opened 2021-09-01
Added extension methods for IFileSystem mocks

duncanp-sonar:
* no product changes

### DIFF
--- a/src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
+++ b/src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
@@ -30,6 +30,7 @@ using SonarLint.VisualStudio.Core.CFamily;
 using SonarLint.VisualStudio.Infrastructure.VS;
 using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.Integration.UnitTests;
+using static SonarLint.VisualStudio.Integration.UnitTests.Extensions.FileSystemExtensions;
 
 namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
 {
@@ -71,13 +72,11 @@ namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
             var defaultLocation = GetDefaultDatabaseFileLocation(activeConfiguration);
             var cmakeSettingsLocation = GetCmakeSettingsLocation(RootDirectory);
 
-            var fileSystem = new Mock<IFileSystem>();
-            fileSystem.Setup(x => x.File.Exists(cmakeSettingsLocation)).Returns(false);
+            var fileSystem = new Mock<IFileSystem>()
+                .FileDoesNotExist(cmakeSettingsLocation)
+                .SetFileExists(defaultLocation, fileExists);
 
-            fileSystem.Setup(x => x.File.Exists(defaultLocation)).Returns(fileExists);
-
-            var logger = new TestLogger();
-            var testSubject = CreateTestSubject(RootDirectory, configProvider, fileSystem.Object, logger);
+            var testSubject = CreateTestSubject(RootDirectory, configProvider, fileSystem.Object);
 
             var result = testSubject.Locate();
 
@@ -90,7 +89,7 @@ namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
                 result.Should().BeNull();
             }
 
-            fileSystem.Verify(x=> x.File.Exists(cmakeSettingsLocation), Times.Once);
+            fileSystem.VerifyFileExistsCalledOnce(cmakeSettingsLocation);
         }
 
         [TestMethod]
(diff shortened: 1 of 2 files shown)


duncanp-sonar:
I was adding some new test extension methods as part of the property evaluation work to help write the new tests, and began fixing up some of the existing tests in the compilation db locator tests.
The volume of changes to the existing changes was starting to obscure the new tests/code, so I've broken out the test factoring into this separate PR.

src/TestInfrastructure/Extensions/FileSystemExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+using Moq;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.Extensions
+{
+    /// <summary>
+    /// Extension methods to simplify working with IFileSystem mocks
+    /// </summary>
+    public static class FileSystemExtensions
+    {
+        public static Mock<IFileSystem> FileDoesNotExist(this Mock<IFileSystem> fileSystem, string fullPath) =>
```
duncanp-sonar, line 31: We don't strictly need three "FileExists"-related methods - just the last one would do. However, I'm experimenting with whether the first two are easier to read/more declarative. If not, we can ditch them later.

src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
```diff
@@ -156,16 +153,16 @@ public void Locate_FailedToReadCMakeSettings_CriticalException_ExceptionThrown()
...
-            var fileSystem = new Mock<IFileSystem>();
-            fileSystem.Setup(x => x.File.Exists(cmakeSettingsLocation)).Returns(true);
+            var fileSystem = new Mock<IFileSystem>()
+                .FileExists(cmakeSettingsLocation);
+
             fileSystem
                 .Setup(x => x.File.ReadAllText(cmakeSettingsLocation))
                 .Throws(new StackOverflowException());
 
-            var logger = new TestLogger();
```
duncanp-sonar, line 165: Miscellaneous: logger was not used.

src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
```diff
@@ -294,14 +291,9 @@ public void Locate_HasCMakeSettingsFile_ConfiguredPathHasParameters_ParametersRe
                     activeBuildConfiguration),
                 CompilationDatabaseLocator.CompilationDatabaseFileName));
 
-        private void SetupCMakeSettingsFileExists(Mock<IFileSystem> fileSystem, string cmakeSettingsLocation, CMakeSettings cmakeSettings)
```
duncanp-sonar, line 297: Changed to an extension method below.

src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
```diff
@@ -313,11 +305,17 @@ private void SetupCMakeSettingsFileExists(Mock<IFileSystem> fileSystem, string c
...
-        private IBuildConfigProvider CreateConfigProvider(string activeConfiguration)
+        private static IBuildConfigProvider CreateConfigProvider(string activeConfiguration)
         {
             var provider = new Mock<IBuildConfigProvider>();
             provider.Setup(x => x.GetActiveConfig(It.IsAny<string>())).Returns(activeConfiguration);
             return provider.Object;
         }
     }
+ 
+    internal static class CompilationDatabaseLocationTestsExtensions
```
duncanp-sonar, line 316: Extension methods specific to this class.
It only has one method currently, but the new tests for the property evaluator service add more.

Originally, the IFileSystem extensions were in this class, then I realised they were general enough to be worth moving to the test infrastructure  project.

src/TestInfrastructure/Extensions/FileSystemExtensions.cs
rita-sonarsource: I think it's definitely easier to read, but I'd add "Set" prefix to them, otherwise it's a bit confusing if it's a verify or a setup.

src/TestInfrastructure/Extensions/FileSystemExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.IO.Abstractions;
+using Moq;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.Extensions
```
rita-sonarsource, line 24: Is that the right namespace? I never noticed what's the default namespace in the tests infra project but this looks off.

src/TestInfrastructure/Extensions/FileSystemExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+        public static Mock<IFileSystem> FileExists(this Mock<IFileSystem> fileSystem, string fullPath) =>
+            fileSystem.SetFileExists(fullPath, true);
+
+        public static Mock<IFileSystem> SetFileExists(this Mock<IFileSystem> fileSystem, string fullPath, bool result)
+        {
+            fileSystem.Setup(x => x.File.Exists(fullPath)).Returns(result);
+            return fileSystem;
+        }
+
+        public static Mock<IFileSystem> SetFileReadAllText(this Mock<IFileSystem> fileSystem, string fullPath, string result)
```
rita-sonarsource, line 43: This is awesome 👍

rita-sonarsource (approved):

src/TestInfrastructure/Extensions/FileSystemExtensions.cs
duncanp-sonar: It does look weird. I'll leave it because it's the default. We can bulk-change the namespace later (it'll change a lot of product files too).

duncanp-sonar:
FYI I moved the test-specific extension class into it's own namespace to stop it from being visible in other tests.

PR #2624, opened 2021-09-02
Return CMakeSettingsSearchResult

rita-sonarsource:


src/CFamily/CMake/CMakeSettingsProvider.cs
```diff
@@ -34,7 +34,23 @@ internal interface ICMakeSettingsProvider
...
+        public CMakeSettingsSearchResult(CMakeSettings parsedSettings, 
+            string cMakeSettingsFilePath, 
+            string rootCMakeListsFilePath)
+        {
+            ParsedSettings = parsedSettings;
+            CMakeSettingsFilePath = cMakeSettingsFilePath;
+            RootCMakeListsFilePath = rootCMakeListsFilePath;
+        }
+
+        public CMakeSettings ParsedSettings { get; }
```
duncanp-sonar, line 51: Minor: I don't think having `Parsed` in the property name adds any useful information - it could just be `Settings` / `AllSettings`.

duncanp-sonar (approved):

Issue #2619, opened 2021-09-01
Handle parsing additional macros in CMakeSettings.json


rita-sonarsource:
### Description

In `CMakeSettings.json` it is possible to use macros for the `buildRoot` argument (=the location of the compilation database). At the moment, we support only `projectDir` and `name` macros. However, there are additional macros that can be used, such as `workspaceHash`, `workspaceRoot`, as well as environment variables `env.VALUE`.

https://docs.microsoft.com/en-us/cpp/build/cmakesettings-reference?view=msvc-160#macros

![image](https://user-images.githubusercontent.com/60586879/131631806-078f75ff-5e4e-42ce-aa8c-cc5a5079ef91.png)

TODO:
- [x] Support parsing additional macros

PR #2627, opened 2021-09-02
Added macro detection and processing

duncanp-sonar:
Part of #2619

### DIFF
--- a/src/CFamily/CMake/MacroEvaluationService.cs
+++ b/src/CFamily/CMake/MacroEvaluationService.cs
@@ -18,6 +18,12 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.Helpers;
+
 namespace SonarLint.VisualStudio.CFamily.CMake
 {
     internal interface IMacroEvaluationService
@@ -35,17 +41,62 @@ namespace SonarLint.VisualStudio.CFamily.CMake
 
     internal class MacroEvaluationService : IMacroEvaluationService
     {
+        private readonly ILogger logger;
+        private readonly IMacroEvaluator macroEvaluator;
+
+        // Looks for patterns like "${name}" and "${prefix.name}" and captures the [prefix, name] tuple.
+        private static readonly Regex CMakeSettingsMacroRegex = new Regex(
+            "\\${" +                 // Match opening literals "${". $ is a special Regex symbol that needs be escaped,
+                                     // and "\" needs to be escaped in the C# string.
+            "((?<prefix>\\w+)\\.)" + // One or more word chars followed by ".". Like "$" above, the "." needs to be escaped with "\\".
+                                     // The word chars are captured in a group called "prefix".
+            "?" +                    // Match the previous group 0 or 1 times i.e. the prefix is optional.
+            "(?<name>\\w+)" +        // Capture one or more word consecutive word chars in a group called "name".
+            "}");                    // Match the closing literal.
+
+        public MacroEvaluationService(ILogger logger)
+            :this(logger, new MacroEvaluator()) {}
+
+        internal MacroEvaluationService(ILogger logger, IMacroEvaluator macroEvaluator)
+        {
+            this.logger = logger;
+            this.macroEvaluator = macroEvaluator;
+        }
+
         public string Evaluate(string input, string activeConfiguration, string workspaceRootDir)
         {
             if (input == null)
             {
                 return null;
             }
+
+            var context = new EvaluationContext(activeConfiguration, workspaceRootDir);
+            var sb = new StringBuilder(input);
+
+            foreach(Match match in CMakeSettingsMacroRegex.Matches(input))
+            {
+                var prefix= match.Groups["prefix"].Value; // will be String.Empty if not found
+                var name = match.Groups["name"].Value;
+
+                var evaluatedProperty = macroEvaluator.TryEvaluate(prefix, name, context);
+
+                if (evaluatedProperty == null)
+                {
+                    // Give up if we failed to evaluate any property
+                    LogDebug($"Failed to evaluate macro: {match.Value}");
+                    return null;
+                }
+
+                LogDebug($"{prefix}.{name} = {evaluatedProperty}");
+                sb.Replace(match.Value, evaluatedProperty);
+            }
             
-            var output = input
-                .Replace("${projectDir}", workspaceRootDir)
-                .Replace("${name}", activeConfiguration);
-            return output;
+            return sb.ToString();
+        }
+
+        private void LogDebug(string message)
+        {
+            logger.LogDebug($"[CMake:MacroEval] [Thread id: {Thread.CurrentThread.ManagedThreadId}] {message}");
         }
     }
 }
--- a/src/CFamily/CMake/MacroEvaluationService.cs
+++ b/src/CFamily/CMake/MacroEvaluationService.cs
@@ -83,7 +83,7 @@ namespace SonarLint.VisualStudio.CFamily.CMake
                 if (evaluatedProperty == null)
                 {
                     // Give up if we failed to evaluate any property
-                    LogDebug($"Failed to evaluate macro: {match.Value}");
+                    logger.WriteLine(Resources.MacroEval_FailedToEvaluateMacro, match.Value);
                     return null;
                 }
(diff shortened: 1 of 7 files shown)


src/CFamily/CMake/MacroEvaluationService.cs
```diff
@@ -34,17 +40,62 @@ internal interface IMacroEvaluationService
 
     internal class MacroEvaluationService : IMacroEvaluationService
     {
+        private readonly ILogger logger;
+        private readonly IMacroEvaluator macroEvaluator;
+
+        // Looks for patterns like "${name}" and "${prefix.name}" and captures the [prefix, name] tuple.
+        private static readonly Regex CMakeSettingsMacroRegex = new Regex(
+            "\\${" +                 // Match opening literals "${". $ is a special Regex symbol that needs be escaped,
```
duncanp-sonar, line 48: This is the same regular expression as the VS code, with additional comments.

src/CFamily/CMake/MacroEvaluationService.cs
```diff
@@ -34,17 +40,62 @@ internal interface IMacroEvaluationService
...
+            "((?<prefix>\\w+)\\.)" + // One or more word chars followed by ".". Like "$" above, the "." needs to be escaped with "\\".
+                                     // The word chars are captured in a group called "prefix".
+            "?" +                    // Match the previous group 0 or 1 times i.e. the prefix is optional.
+            "(?<name>\\w+)" +        // Capture one or more word consecutive word chars in a group called "name".
+            "}");                    // Match the closing literal.
+
+        public MacroEvaluationService(ILogger logger)
+            :this(logger, new MacroEvaluator()) {}
+
+        internal MacroEvaluationService(ILogger logger, IMacroEvaluator macroEvaluator)
```
duncanp-sonar, line 59: The job of the service is to parse the input for macros, asking the 'macroEvaluator' to calculate the correct value.

src/CFamily/CMake/MacroEvaluator.cs
```diff
@@ -0,0 +1,69 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.CFamily.CMake
+{
+    internal class EvaluationContext
+    {
+        public EvaluationContext(string activeDirectory, string rootDirectory)
```
duncanp-sonar, line 25: Data class containing the information needed to evaluate macros.

It's an internal implementation detail shared between the macro evaluation service and macro evaluator.
It might disappear / change format when I add the real macro evaluation.

src/CFamily/CMake/MacroEvaluator.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public string ActiveDirectory { get; }
+    }
+
+    /// <summary>
+    /// Returns the evaluated value for the supplied macro, or null if it could not be evaluated
+    /// </summary>
+    internal interface IMacroEvaluator
+    {
+        string TryEvaluate(string macroPrefix, string macroName, EvaluationContext context);
```
duncanp-sonar, line 41: The processing of individual macros will be done in this class.

src/CFamily/CMake/MacroEvaluator.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                return evaluationContext.RootDirectory;
+            }
+
+            if (macroName == "name")
+            {
+                return evaluationContext.ActiveDirectory;
+            }
+
+            return null;
```
duncanp-sonar, line 66: I moved the existing "known" macro processing to this class.

src/CFamily/CMake/MacroEvaluationService.cs
```diff
@@ -34,17 +40,62 @@ internal interface IMacroEvaluationService
...
+            this.macroEvaluator = macroEvaluator;
+        }
+
         public string Evaluate(string input, string activeConfiguration, string workspaceRootDir)
         {
             if (input == null)
             {
                 return null;
             }
+
```
duncanp-sonar, line 71: The service now orchestrates the process of finding the macros, asking for the values, and replacing them in the input.

The actual calculation of individual values is delegated to the new `IMacroEvaluator`.

src/CFamily/CMake/MacroEvaluationService.cs
```diff
@@ -34,17 +40,62 @@ internal interface IMacroEvaluationService
...
+                if (evaluatedProperty == null)
+                {
+                    // Give up if we failed to evaluate any property
+                    LogDebug($"Failed to evaluate macro: {match.Value}");
+                    return null;
+                }
+
+                LogDebug($"{prefix}.{name} = {evaluatedProperty}");
+                sb.Replace(match.Value, evaluatedProperty);
```
duncanp-sonar, line 90: `match.Value` contains all of the text that was matched and needs to be replaced

e.g. 
match.Value = `${aaa.bbb}` 
`match.Groups["prefix"]` = `aaa`
`match.Groups["name"]` = `bbb`

src/CFamily/CMake/MacroEvaluator.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            if (macroName == "projectDir")
+            {
+                return context.RootDirectory;
+            }
+
+            if (macroName == "name")
+            {
+                return context.ActiveDirectory;
```
rita-sonarsource, line 63: I believe "name" is the activate configuration, not directory.

src/CFamily/CMake/MacroEvaluator.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        public EvaluationContext(string activeDirectory, string rootDirectory)
+        {
+            ActiveDirectory = activeDirectory;
+            RootDirectory = rootDirectory;
+        }
+
+        public string RootDirectory { get; }
+
+        public string ActiveDirectory { get; }
```
rita-sonarsource, line 33: Should be ActiveConfiguration

src/CFamily/CMake/MacroEvaluationService.cs
```diff
@@ -34,17 +40,62 @@ internal interface IMacroEvaluationService
...
+        public MacroEvaluationService(ILogger logger)
+            :this(logger, new MacroEvaluator()) {}
+
+        internal MacroEvaluationService(ILogger logger, IMacroEvaluator macroEvaluator)
+        {
+            this.logger = logger;
+            this.macroEvaluator = macroEvaluator;
+        }
+
         public string Evaluate(string input, string activeConfiguration, string workspaceRootDir)
```
rita-sonarsource, line 65: We will need more information from CMakeBuildConfiguration, such as the generator, so we might need to adopt this interface method when more macros are added.

src/CFamily/CMake/MacroEvaluationService.cs
```diff
@@ -34,17 +40,62 @@ internal interface IMacroEvaluationService
...
+            {
+                var prefix= match.Groups["prefix"].Value; // will be String.Empty if not found
+                var name = match.Groups["name"].Value;
+
+                var evaluatedProperty = macroEvaluator.TryEvaluate(prefix, name, context);
+
+                if (evaluatedProperty == null)
+                {
+                    // Give up if we failed to evaluate any property
+                    LogDebug($"Failed to evaluate macro: {match.Value}");
```
rita-sonarsource, line 85: VS shows this information to the user, so maybe it should be an info message instead of debug.

rita-sonarsource (approved):
Nice!

src/CFamily/CMake/MacroEvaluationService.cs
duncanp-sonar: Yep. We might promote the 'EvaluationContext' class and use that here instead of passing multiple arguments. Or we might be able to use one of the existing CMake settings/config types.

If the required data naturally fits into one of the CMake settings/config types then I'd add it there.
If there isn't a natural home for it then I'd add it to the `EvaluationContext` and use that in this interface too.

src/CFamily/CMake/MacroEvaluationService.cs
duncanp-sonar: I wondered about that. At the moment, we'll display a message at a higher level saying we can't evaluate the build root property, which includes the pre-evaluation version of the string i.e. with the macros included, but it won't say which macro we couldn't process.
I was worried that displaying a lower-level message would be too noisy, but it will make it easier for the user to fix any issues. I'll change it and update the test.

PR #2631, opened 2021-09-03
Evaluate env variables macros

### LINKED ISSUE #2628: [CMake] Handle environment variables macros
#### Description

Support `env.value` macros in the buildRoot.

TODO:
- [x] Check case sensitivity for "env" and the variable name
- [x] What VS does when the variable is not found? Replaced with empty string?
- [x] Environment variable wrapper for testing

rita-sonarsource:
Fixes #2628

### DIFF
--- a/src/Core.UnitTests/EnvironmentVariableProviderTests.cs
+++ b/src/Core.UnitTests/EnvironmentVariableProviderTests.cs
@@ -1,0 +1,92 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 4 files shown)


src/Core.UnitTests/EnvironmentVariableProviderTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("variableName");
+        }
+
+        [TestMethod]
+        public void TryGet_VariableDoesNotExist_Null()
+        {
+            var testSubject = new EnvironmentVariableProvider();
+
+            var result = testSubject.TryGet("test");
```
duncanp-sonar, line 48: It might be worth testing with a more-likely-to-be-unique environment variable name.

duncanp-sonar (approved):

Issue #2630, opened 2021-09-03
Calculate other simple macro values


duncanp-sonar:
### Description

Add support for other non-environment variable macros.
See https://docs.microsoft.com/en-us/cpp/build/cmakesettings-reference?view=msvc-160#macros

PR #2633, opened 2021-09-03
Pass evaluation context + Parse the generator from CMakeSettings

rita-sonarsource:
Prep for #2630

### DIFF
--- a/src/CFamily/CMake/CompilationDatabaseLocator.cs
+++ b/src/CFamily/CMake/CompilationDatabaseLocator.cs
@@ -124,7 +124,9 @@ namespace SonarLint.VisualStudio.CFamily.CMake
                 return null;
             }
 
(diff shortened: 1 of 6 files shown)


src/CFamily/CMake/CompilationDatabaseLocator.cs
```diff
@@ -124,7 +124,9 @@ private string GetDefaultLocation(string rootDirectory, string activeConfigurati
                 return null;
             }
 
-            var evaluatedBuildRoot = macroEvaluationService.Evaluate(buildConfiguration.BuildRoot, buildConfiguration.Name, rootDirectory);
+            var evaluationContext = new EvaluationContext(buildConfiguration.Name, rootDirectory);
```
rita-sonarsource, line 127: I need to pass more information so I'll extend the context to include additional fields

rita-sonarsource:
@duncanp-sonar this is a prep work PR, in the next one I will add the additional macros.

duncanp-sonar (approved):

PR #2635, opened 2021-09-03
Support additional simple macros

### LINKED ISSUE #2630: Calculate other simple macro values

rita-sonarsource:
Fixes #2630

### DIFF
--- a/src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
+++ b/src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
@@ -298,7 +298,7 @@ namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
                 MacroEvalService = new Mock<IMacroEvaluationService>();
                 MacroEvalService.Setup(x =>
                     x.Evaluate(unevaluatedBuildRoot,
-                            It.Is((EvaluationContext context) =>
+                            It.Is((IEvaluationContext context) =>
                                 context != null &&
                                 context.ActiveConfiguration == activeConfig &&
                                 context.RootDirectory == workspaceRootDir &&
--- a/src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
+++ b/src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
@@ -202,7 +202,9 @@ namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
             context.MacroEvalService.Invocations.Count.Should().Be(1);
         }
 
-        private static CMakeSettings CreateCMakeSettings(string activeConfigurationName, string buildRoot) =>
+        private static CMakeSettings CreateCMakeSettings(string activeConfigurationName, 
+            string buildRoot, 
+            string generator = "generator") =>
             new()
             {
                 Configurations = new[]
@@ -210,7 +212,8 @@ namespace SonarLint.VisualStudio.CFamily.UnitTests.CMake
                     new CMakeBuildConfiguration
                     {
                         Name = activeConfigurationName,
-                        BuildRoot = buildRoot
+                        BuildRoot = buildRoot,
+                        Generator = generator
                     }
                 }
             };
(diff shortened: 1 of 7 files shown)


src/CFamily.UnitTests/CMake/CompilationDatabaseLocatorTests.cs
```diff
@@ -294,13 +300,16 @@ public MacroEvalContext(string macroServiceReturnValue, string unevaluatedBuildR
                     x.Evaluate(unevaluatedBuildRoot,
                             It.Is((EvaluationContext context) =>
                                 context != null &&
-                                context.ActiveConfiguration == ActiveConfig &&
-                                context.RootDirectory == WorkspaceRootDir)))
+                                context.ActiveConfiguration == activeConfig &&
```
rita-sonarsource, line 303: Verifying that the context is initialized with the right arguments

src/CFamily/CMake/EvaluationContext.cs
```diff
@@ -0,0 +1,72 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.CFamily.CMake
+{
+    internal interface IEvaluationContext
```
rita-sonarsource, line 23: Extracted to a new file just because it started to be long.
I added the interface so I could do a simple mock in MacroEvaluationServiceTests; otherwise a change were needed each time a new field was introduced.

src/CFamily/CMake/MacroEvaluator.cs
```diff
@@ -19,29 +19,17 @@
  */
 
 using System;
+using System.IO;
 using SonarLint.VisualStudio.Core;
 
 namespace SonarLint.VisualStudio.CFamily.CMake
 {
-    internal class EvaluationContext
```
rita-sonarsource, line 26: Extracted to another file

src/CFamily/CMake/MacroEvaluator.cs
```diff
@@ -58,33 +46,44 @@ internal MacroEvaluator(IEnvironmentVariableProvider environmentVariableProvider
...
+                case "projectDir":
+                    return Path.GetDirectoryName(evaluationContext.RootCMakeListsFilePath);
+                case "projectDirName":
+                {
+                    var directoryName = Path.GetDirectoryName(evaluationContext.RootCMakeListsFilePath);
+                    return string.IsNullOrEmpty(directoryName) ? null : new DirectoryInfo(directoryName).Name;
+                }
+                case "thisFile":
+                    return evaluationContext.CMakeSettingsFilePath;
+                case "thisFileDir":
```
rita-sonarsource, line 78: `thisFileDir` is not listed on msdn but it does appear in the IDE's tooltip + VS parses it, so I added it.

src/CFamily/CMake/MacroEvaluator.cs
```diff
@@ -58,33 +46,44 @@ internal MacroEvaluator(IEnvironmentVariableProvider environmentVariableProvider
...
+                case "projectFile":
+                    return evaluationContext.RootCMakeListsFilePath;
+                case "projectDir":
+                    return Path.GetDirectoryName(evaluationContext.RootCMakeListsFilePath);
+                case "projectDirName":
+                {
+                    var directoryName = Path.GetDirectoryName(evaluationContext.RootCMakeListsFilePath);
+                    return string.IsNullOrEmpty(directoryName) ? null : new DirectoryInfo(directoryName).Name;
+                }
+                case "thisFile":
```
rita-sonarsource, line 76: Technically all the file arguments don't make sense (`thisFile`, `projectFile`) in the buildRoot context. I don't know if it would be better to leave them and let it fail. VS would fail as well.

src/CFamily.UnitTests/CMake/MacroEvaluatorTests.cs
```diff
@@ -100,17 +208,11 @@ public void Evaluate_EnvPrefix_NoMacroName_Null(string macroName)
             envVariableProvider.Invocations.Count.Should().Be(0);
         }
 
-        [TestMethod]
-        [DataRow(null)]
-        [DataRow("")]
-        public void Evaluate_NoMacroName_Null(string macroName)
```
rita-sonarsource, line 106: Unchanged, just moved up

src/CFamily/CMake/MacroEvaluator.cs
duncanp-sonar: Might as well leave them.

duncanp-sonar (approved):

PR #2642, opened 2021-09-14
Handle CMake internal messages

### LINKED ISSUE #2540: Handle CMake-specific error messages from the subprocess
### Description

The new subprocess mode adds three new message to return errors:

* `internal.InvalidInput` : Subprocess is called the wrong way from SL side
* `internal.UnsupportedConfig` : user config is not supported. For example “Command” contains an unsupported compiler or an unsupported compiler option.
* `internal.UnexpectedFailure` : subprocess failed in an unexpected way.

The message should be logged to the output window to help the user understand why the analysis was not performed.

duncanp-sonar:
Fixes #2540

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
@@ -226,7 +226,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
         }
 
         [TestMethod]
-        public async Task TriggerAnalysisAsync_AnalysisFails_NotifiesOfFailure()
+        public async Task TriggerAnalysisAsync_AnalysisFailsDueToException_NotifiesOfFailure()
         {
             void MockSubProcessCall(Action<Message> message, IRequest request, ISonarLintSettings settings, ILogger logger, CancellationToken token)
             {
@@ -246,6 +246,37 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
 
             statusNotifier.Verify(x=> x.AnalysisStarted(filePath), Times.Once);
             statusNotifier.Verify(x=> x.AnalysisFailed(filePath, It.Is<NullReferenceException>(e => e.Message == "test")), Times.Once);
+            statusNotifier.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public async Task TriggerAnalysisAsync_AnalysisFailsDueToInternalMessage_NotifiesOfFailure()
+        {
+            const string fileName = "c:\\data\\aaa\\bbb\\file.txt";
+            var request = CreateRequest(fileName);
+            var requestFactory = CreateRequestFactory(fileName, ValidAnalyzerOptions, request);
+
+            var internalErrorMessage = new Message("internal.UnexpectedFailure", "", 1, 1, 1, 1, "XXX Error in subprocess XXX", false, Array.Empty<MessagePart>());
+
+            var issueConverter = new Mock<ICFamilyIssueToAnalysisIssueConverter>();
+            var mockConsumer = new Mock<IIssueConsumer>();
+            var statusNotifier = new Mock<IAnalysisStatusNotifier>();
+
+            var testSubject = CreateTestableAnalyzer(issueConverter: issueConverter.Object,
+                requestFactory: requestFactory.Object);
+
+            TestableCLangAnalyzer.HandleCallSubProcess subProcessOp = (handleMessage, _, _, _, _) =>
+            {
+                handleMessage(internalErrorMessage);
+            };
+            testSubject.SetCallSubProcessBehaviour(subProcessOp);
+
+            await testSubject.TriggerAnalysisAsync(fileName, ValidDetectedLanguages, mockConsumer.Object, ValidAnalyzerOptions, statusNotifier.Object, CancellationToken.None);
+
+            testSubject.SubProcessCompleted.Should().BeTrue();
+
+            statusNotifier.Verify(x => x.AnalysisStarted(fileName), Times.Once);
+            statusNotifier.Verify(x => x.AnalysisFailed(fileName, CFamilyStrings.MSG_GenericAnalysisFailed), Times.Once);
             statusNotifier.VerifyNoOtherCalls();
         }
 
@@ -292,15 +323,8 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.UnitTests
 
             private HandleCallSubProcess onCallSubProcess;
 
-            public void SetCallSubProcessBehaviour(HandleCallSubProcess onCallSubProcess)
-            {
+            public void SetCallSubProcessBehaviour(HandleCallSubProcess onCallSubProcess) =>
                 this.onCallSubProcess = onCallSubProcess;
-
-                // The sub process is executed on a separate thread, so any exceptions might be
-                // squashed by the product code. So, we'll set a flag to indicate whether it
-                // ran to completion.
-                SubProcessCompleted = true;
-            }
 
             public bool SubProcessCompleted { get; private set; }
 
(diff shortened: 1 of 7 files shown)


src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -249,6 +249,37 @@ void MockSubProcessCall(Action<Message> message, IRequest request, ISonarLintSet
             statusNotifier.VerifyNoOtherCalls();
         }
 
+        [TestMethod]
+        public async Task TriggerAnalysisAsync_AnalysisFailsDueToInternalMessage_NotifiesOfFailure()
+        {
```
duncanp-sonar, line 254: New test - if we get an internal rule error message the analysis should be reported as having failed.

src/Integration.Vsix.UnitTests/CFamily/CLangAnalyzerTests.cs
```diff
@@ -292,16 +323,9 @@ private class TestableCLangAnalyzer : CLangAnalyzer
...
-            {
+            public void SetCallSubProcessBehaviour(HandleCallSubProcess onCallSubProcess) =>
                 this.onCallSubProcess = onCallSubProcess;
 
-                // The sub process is executed on a separate thread, so any exceptions might be
-                // squashed by the product code. So, we'll set a flag to indicate whether it
-                // ran to completion.
-                SubProcessCompleted = true;
-            }
-
```
duncanp-sonar, line 304: Miscellaneous: fixed an error in the test. The "action completed" flag was being set immediately; it should have been set only after the action has been executed.

src/Integration.Vsix.UnitTests/CFamily/MessageHandlerTests.cs
```diff
@@ -95,6 +106,7 @@ public void HandleMessage_IssueProcessing_InactiveRulesAreIgnored_ActiveRulesAre
             // Now stream an active rule message
             testSubject.HandleMessage(activeRuleMessage);
 
+            testSubject.AnalysisSucceeded.Should().BeTrue();
```
duncanp-sonar, line 109: Updated the existing message handler tests: they should now set the `AnalysisSucceeded` property appropriately.

src/Integration.Vsix.UnitTests/CFamily/MessageHandlerTests.cs
```diff
@@ -186,7 +250,53 @@ private static IRequest CreateRequest(string file = null, string language = null
             return request.Object;
         }
 
-        private static Message CreateMessage(string ruleId, string fileName = "any file") =>
-            new Message(ruleId, fileName, -1, -1, -1, -1, "any text", false, null);
+        private static Message CreateMessage(string ruleId, string fileName = "any file", string text = "any text") =>
+            new Message(ruleId, fileName, -1, -1, -1, -1, text, false, null);
+
+
+        private class MessageHandlerTestContext
```
duncanp-sonar, line 257: New helper class to simplify setting up the tests. I'll simplify the existing tests to use the helper class separately.

src/Integration.Vsix/CFamily/CFamilyStrings.resx
```diff
@@ -198,4 +198,17 @@
...
+    <value>[CFamily Analysis] Internal error calling the analysis subprocess: {0}</value>
+  </data>
+  <data name="MsgHandler_ReportUnexpectedFailure" xml:space="preserve">
+    <value>[CFamily Analysis] Internal error in the analysis subprocess: {0}</value>
+  </data>
+  <data name="MsgHandler_ReportUnsupportedConfiguration" xml:space="preserve">
+    <value>[CFamily Analysis] Unsupported configuration: {0}</value>
+  </data>
+  <data name="MSG_GenericAnalysisFailed" xml:space="preserve">
+    <value>Check the output window for more information.</value>
```
duncanp-sonar, line 211: This text is passed to the status notifier which displays it both in the status bar and the output window. The wording is ok for the status bar but it's a bit clumsy for the output window if that is where the user reads it. Hopefully clear enough. Feel free to suggest alternatives.

src/Integration.Vsix/CFamily/CFamilyStrings.resx
```diff
@@ -198,4 +198,17 @@
   <data name="ERROR_CreatingVcxRequest" xml:space="preserve">
     <value>[CFamily Analysis] Error creating VCX request. File: {0}, Error: {1}</value>
   </data>
+  <data name="MsgHandler_ReportInvalidInput" xml:space="preserve">
+    <value>[CFamily Analysis] Internal error calling the analysis subprocess: {0}</value>
```
duncanp-sonar, line 202: The "internal error" messages don't log the full file path. I decided not to log it because  is logged in the "analysis failed" message that comes immediately afterwards. The file path can be quite long, so omitting it makes it more likely the user can read the specific failure message without scrolling.

src/Integration.Vsix/CFamily/CLangAnalyzer.cs
```diff
@@ -176,8 +176,15 @@ private void RunAnalysis(IRequest request, IIssueConsumer consumer, IAnalysisSta
                 }
                 else
                 {
-                    var analysisTime = DateTime.Now - analysisStartTime;
-                    statusNotifier?.AnalysisFinished(request.Context.File, messageHandler.IssueCount, analysisTime);
+                    if (messageHandler.AnalysisSucceeded)
```
duncanp-sonar, line 179: The CLangAnalyzer needed to be updated to report "analysis failed" for any internal error.

src/Integration.Vsix/CFamily/MessageHandler.cs
```diff
@@ -55,17 +68,50 @@ internal class MessageHandler : IMessageHandler
...
         {
             this.request = request;
             this.issueConsumer = issueConsumer;
             this.issueConverter = issueConverter;
+            this.logger = logger;
         }
 
         public void HandleMessage(Message message)
+        {
+            // Handle known internal rule keys - used to return info/warnings
```
duncanp-sonar, line 87: The changes to the message handler itself are straightforward.

src/Integration.Vsix/CFamily/CFamilyStrings.resx
rita-sonarsource: The status notifier only needs the file name, the failure message that we pass is the one that is shown in the output window.
https://github.com/SonarSource/sonarlint-visualstudio/blob/2fcabc8baab30c3dde67dc359290d602932cb65d/src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs#L74

So this message will always be shown only in the output window. We can change it to something generic, "internal error" / "analysis error".

src/Integration.Vsix/CFamily/MessageHandler.cs
```diff
@@ -55,17 +68,50 @@ internal class MessageHandler : IMessageHandler
...
+
+                case "internal.UnexpectedFailure": // unexpected failure in the subprocess
+                    AnalysisSucceeded = false;
+                    logger.WriteLine(CFamilyStrings.MsgHandler_ReportUnexpectedFailure, message.Text);
+                    break;
+
+                case "internal.fileDependency": // not currently handle. See https://github.com/SonarSource/sonarlint-visualstudio/issues/2611
+                    break;
+
+                default: // assume anything else is an analysis sissue
```
rita-sonarsource, line 108: minor typo, "sissue"

src/Integration.Vsix/CFamily/MessageHandler.cs
```diff
@@ -55,17 +68,50 @@ internal class MessageHandler : IMessageHandler
...
+                    AnalysisSucceeded = false;
+                    logger.WriteLine(CFamilyStrings.MsgHandler_ReportInvalidInput, message.Text);
+                    break;
+
+                case "internal.UnexpectedFailure": // unexpected failure in the subprocess
+                    AnalysisSucceeded = false;
+                    logger.WriteLine(CFamilyStrings.MsgHandler_ReportUnexpectedFailure, message.Text);
+                    break;
+
+                case "internal.fileDependency": // not currently handle. See https://github.com/SonarSource/sonarlint-visualstudio/issues/2611
```
rita-sonarsource, line 105: minor typo, "handle"

rita-sonarsource (approved):

src/Integration.Vsix/CFamily/CFamilyStrings.resx
duncanp-sonar: You're right: this is what appears in the status bar:

image

The wording is very similar to what I had in an earlier message, so I thought the status bar was displaying my "new" message.

I'll change it as discussed offline.

PR #2650, opened 2021-09-16
Fix code smells + signing

rita-sonarsource:


build/Signing.targets
```diff
@@ -41,7 +41,7 @@
     <Error Condition="@(VSIXesToSign) == '' " Text="The list of VSIXes to sign is empty." />
     
     <!-- For info on timestamping see https://www.digicert.com/kb/code-signing/signcode-signtool-command-line.htm -->
-    <Exec Command="$(VsixSignToolPath) sign /f &quot;$(pfxCertificatePath)&quot; /p $(pfxPassword) /sha1 $(pfxSha1) /tr http://timestamp.digicert.com?alg=sha256 /v &quot;%(VSIXesToSign.Identity)&quot;" />
+    <Exec Command="$(VsixSignToolPath) sign /f &quot;$(pfxCertificatePath)&quot; /p $(pfxPassword) /sha1 $(pfxSha1) /tr http://timestamp.digicert.com?alg=sha256 /td SHA1 /v &quot;%(VSIXesToSign.Identity)&quot;" />
```
rita-sonarsource, line 44: @duncanp-sonar it seems that something changed on the CI agents, the builds suddenly started to fail:
```
"C:\Program Files (x86)\Windows Kits\10\bin\10.0.22000.0\x86\signtool.exe" sign /fd SHA256 /sha1 *** /f "D:\a\_temp\sonarsource-2019-2021.pfx" /p *** /tr http://timestamp.digicert.com?alg=sha256 "D:\a\1\s\src\EmbeddedVsix\obj\Release\SonarLint.VisualStudio.AdditionalFiles.dll"
##[error]EXEC(0,0): Error : No file timestamp algorithm specified. Please specify the timestamp algorithm with the /td flag. Using /td SHA256 is recommended and more secure than SHA1. Calling signtool with /td sha1 is equivalent to the previous behavior.
    24>EXEC : SignTool error : No file timestamp algorithm specified. Please specify the timestamp algorithm with the /td flag. Using /td SHA256 is recommended and more secure than SHA1. Calling signtool with /td sha1 is equivalent to the previous behavior. [D:\a\1\s\src\EmbeddedVsix\EmbeddedVsix.csproj]
    ```
    
According to the [documentation](https://docs.microsoft.com/en-us/dotnet/framework/tools/signtool-exe) this was a warning in the past and "In later versions of SignTool, the warning will become an error". 
This is a build that passed yesterday: 
https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=36814&view=logs&j=ae5f67bc-b508-52d0-18c7-3e0886a35c88&t=395a6205-8b2d-59b4-e3ec-324edbb35b0e&l=12

And it points to signtool being at `C:\Program Files (x86)\Windows Kits\10\bin\10.0.20348.0\x86\signtool.exe`. Now it's `C:\Program Files (x86)\Windows Kits\10\bin\10.0.22000.0\x86\signtool.exe`, so they _are_ different, but it seems that it's an older version now and not newer.

So I'm not sure what to do with this and how to verify that the signing is still correct (and whether it should be `/td SHA1` -- the message says that that is the default, so that's what should've been executed before).

File: src/CFamily/CMakeProjectTypeIndicator.cs [hunk 2]
```diff
@@ -58,11 +57,6 @@ public bool IsCMake()
 
             var rootDirectory = folderWorkspaceService.FindRootDirectory();
 
-            if (string.IsNullOrEmpty(rootDirectory))
```
duncanp-sonar, line 61: This is actually a change of behaviour, but it probably doesn't make any difference in practice - an different exception will still be thrown when we try to use the null value.

build/Signing.targets
duncanp-sonar: `10.0.20348.0` < `10.0.22000.0`, so it is a slightly newer version.

The documentation recommends using SHA256, so I think we should do that. It shouldn't have any effect on users, but even if it does I think we should still do it since it's a security-related improvement.

build/Signing.targets
```diff
@@ -51,7 +51,7 @@
     <Error Condition=" $(SIGNTOOL_PATH) == '' " Text="The location of the signtool.exe has not been set ('SIGNTOOL_PATH')" />
     <Error Condition=" !Exists($(SIGNTOOL_PATH)) " Text="Signing tool exe does not exist at the specified location: $(SIGNTOOL_PATH)" />
 
-    <Exec Command="&quot;$(SIGNTOOL_PATH)&quot; sign /fd SHA256 /sha1 $(pfxSha1) /f &quot;$(pfxCertificatePath)&quot; /p $(pfxPassword) /tr http://timestamp.digicert.com?alg=sha256 &quot;%(AssembliesToSign.FullPath)&quot;" />
+    <Exec Command="&quot;$(SIGNTOOL_PATH)&quot; sign /fd SHA256 /sha1 $(pfxSha1) /f &quot;$(pfxCertificatePath)&quot; /p $(pfxPassword) /tr http://timestamp.digicert.com?alg=sha256 /td SHA1 &quot;%(AssembliesToSign.FullPath)&quot;" />
```
duncanp-sonar, line 54: Going forwards, I think it would be better to have separate PRs for changes like these i.e. separate the signing changes from the cleanup changes. They're unrelated, and signing is important and should have a material impact on the build, while fixing the code smells should not change the behaviour.

src/CFamily/CMakeProjectTypeIndicator.cs
rita-sonarsource: It will still be an ArgumentNullException, coming from the directory enumerate. I left the test that verifies it.

build/Signing.targets
rita-sonarsource: > `10.0.20348.0` < `10.0.22000.0`, so it is a slightly newer version.

I can't read 🤦

duncanp-sonar:
FYI I've manually verified the signing of the VSIX and individual assemblies produced by the CI build.

PR #2663, opened 2021-09-27
Improve performance of local builds

### LINKED ISSUE #2568: Improve performance of local builds
### Description

Local builds seem to involve a lot of rebuilding each time.

TODO:
- [ ] Check if the MSBuild configuration is correct for EmbeddedVsix / Integration.Vsix and if we're not doing unnecessary rebuilds.

rita-sonarsource:
Fixes #2568

### DIFF
--- a/src/EmbeddedVsix/EmbeddedVsix.csproj
+++ b/src/EmbeddedVsix/EmbeddedVsix.csproj
@@ -3,8 +3,34 @@
 
   <PropertyGroup>
     <RequiresSigning>true</RequiresSigning>
+  </PropertyGroup>
+
+  <!-- *********************************************************-->
+  <!-- Hack to work round the VSIX being created every time -->
+  <!-- *********************************************************-->
+  <PropertyGroup>
     <CreateVsixContainer>false</CreateVsixContainer>
   </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Prevent the vsix from being cleaned during incremental builds -->
+    <FileWritesShareable Include="$(OutputPath)$(AssemblyName).vsix"/>
+  </ItemGroup>
+
+  <!-- Will be called if "GetTypeScriptFiles" target is executed.
+       When building the project, even if nothing has changed in the Typescript files, "Microsoft.VsSDK.targets" always generates a VSIX.
+       This has a big performance penalty, especially on slower machines.
+       To overcome this, we always set the property "CreateVsixContainer" to "false" at the start of the run, and set it to "true" only if GetTypeScriptFiles target 
+       was executed (= if GetTypeScriptFiles target was executed, it means that the Typescript files were changed and we should generate a new VSIX).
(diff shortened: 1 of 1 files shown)


duncanp-sonar (approved):

src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -102,6 +103,9 @@
     <!-- Specifies the target inputs and outputs so MSBuild can skip fetching the files if the analyzer version has not changed. -->
     <AdditionalFilesInput Include="..\EmbeddedSonarAnalyzer.props" />
     <AdditionalFilesOutput Include="$(OutputPath)$(AssemblyName).vsix" />
+    
+    <!-- Prevent the vsix from being cleaned during incremental builds -->
+    <FileWritesShareable Include="$(OutputPath)$(AssemblyName).vsix"/>
```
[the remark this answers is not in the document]
duncanp-sonar: We're trying to work round a bug in the SDK targets, so whatever we do will be a bit hacky. The best we can do is make it obvious what the bug is and what we have in our target files to work round it (FYI I checked and the same problem exists in a "clean" new VSIX that doesn't contain all of our extra targets and tweaks - one of the files is always regenerated, which in turn means the VSIX is always re-created).

We could make our hacks more obvious by co-locating them in the file as much as possible e.g.

```
<!-- *********************************************************-->
<!-- Hack to work round the VSIX being created every time -->
<!-- *********************************************************-->
<PropertyGroup>
  <CreateVsixContainer>false</CreateVsixContainer>
</PropertyGroup>

<ItemGroup>
    <!-- Prevent the vsix from being cleaned during incremental builds -->
    <FileWritesShareable Include="$(OutputPath)$(AssemblyName).vsix"/>
</ItempGroup>

<Target Name="SonarLint_ForceCreateVsix" ... />
<!-- *********************************************************-->

```

duncanp-sonar (approved):
I think we should try to locate the changes together as much as possible. Other than that, looks good.

src/EmbeddedVsix/EmbeddedVsix.csproj
```diff
@@ -102,6 +129,9 @@
     <!-- Specifies the target inputs and outputs so MSBuild can skip fetching the files if the analyzer version has not changed. -->
     <AdditionalFilesInput Include="..\EmbeddedSonarAnalyzer.props" />
     <AdditionalFilesOutput Include="$(OutputPath)$(AssemblyName).vsix" />
+    
+    <!-- Prevent the vsix from being cleaned during incremental builds -->
```
duncanp-sonar, line 133: Shouldn't this property-set have been deleted? (now done in line 17)

rita-sonarsource: You're right, it's a conflict, I'll fix and force push

PR #2664, opened 2021-09-27
Remove duplicate property

rita-sonarsource:
Part of #2568


rita-sonarsource:
New PR since I can't force-push to master

PR #2686, opened 2021-09-30
Reference SonarLint.Secrets nuget package

rita-sonarsource:


src/CloudSecrets/Analyzer.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Threading;
+using SonarLint.Secrets.DotNet;
+using SonarLint.VisualStudio.Core.Analysis;
+
+namespace SonarLint.VisualStudio.CloudSecrets
+{
+    [Export(typeof(IAnalyzer))]
+    internal class Analyzer : IAnalyzer
+    {
+        private readonly DummySecretDetector dummySecretDetector;
```
rita-sonarsource, line 32: SC complaints about this class but it's not relevant, it will change anyway.

duncanp-sonar (approved):

Issue #2666, opened 2021-09-28
Implement secrets detection analyzer


rita-sonarsource:
### Description

Implement `IAnalyzer` that is responsible for analyzing files for secrets.
All text files should be supported (.txt, .yml, .json, etc), including code files for languages that SLVS does not support (i.e. .py).

TODO:
- [x] Implement IAnalyzer
- [ ] ~~Port analysis logic from the java code (TBD regex)~~ --> will be done in separate tickets
- [x] Convert secret's location to VS line position
- [x] Create IAnalysisIssue for the detected issues, with Warning severity
- [ ] ~~Rule link: redirect to rules.sonarsource.com as normal (not all rules are there currently, but they will be at some point)~~ --> done separately
- [ ] ~~Populate the error list~~ --> will be done in #2671

Issue #2679, opened 2021-09-28
Support enabling/disabling secret rules


rita-sonarsource:
### Description

Each provider/check is a separate rule i.e. we can just treat the "secret types" as if they were independent rules that users can disable/enable.

TODO:
- [ ] Support enabling/disabling secret rules via settings.json.
- [ ] In connected mode, the settings.json should be taken into account as well.

PR #2691, opened 2021-10-05
Add SecretsAnalyzer

rita-sonarsource:
Part of #2666


src/CloudSecrets/SecretsAnalyzer.cs
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+
+        private IAnalysisIssue CreateIssue(ISecret secret, string filePath)
+        {
+            var ruleKey = secret.RuleKey;
+            var severity = AnalysisIssueSeverity.Major;
+            var message = "this is a secret";
+
+            var startLine = 0;
```
rita-sonarsource, line 100: I'll open a new PR to convert the text range into VS line

src/CloudSecrets/SecretsAnalyzer.cs
```diff
@@ -0,0 +1,119 @@
...
+            analysisStatusNotifier.AnalysisStarted(filePath);
+
+            try
+            {
+                var stopwatch = Stopwatch.StartNew();
+
+                var fileContent = fileSystem.File.ReadAllText(filePath);
+
+                // todo: pass rules configuration if should run the detector? or should each detector expose metadata?
+                // todo: pass cancellation token that should be checked before running each detector? or should check the cancellation here?
```
rita-sonarsource, line 77: This is something we should decide and I'll fix it

src/CloudSecrets/SecretsAnalyzer.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            analysisStatusNotifier.AnalysisStarted(filePath);
+
+            try
+            {
+                var stopwatch = Stopwatch.StartNew();
+
+                var fileContent = fileSystem.File.ReadAllText(filePath);
+
+                // todo: pass rules configuration if should run the detector? or should each detector expose metadata?
```
rita-sonarsource, line 76: will be handled in #2679

src/Integration.Vsix.UnitTests/packages.lock.json
```diff
@@ -588,8 +588,8 @@
       },
       "SonarLint.Secrets.DotNet": {
         "type": "Transitive",
-        "resolved": "0.0.1",
-        "contentHash": "+duxMJOFgXLvgGJKhLJeuRIDa0U+BZ0qeXWSEcHv18Cxx74Jfw9kWL6oqk6c47bPBKt6Gz0KtEKpMMP2C1sLxQ=="
+        "resolved": "0.0.2",
```
rita-sonarsource, line 591: update for the latest API

src/CloudSecrets/SecretsAnalyzer.cs
duncanp-sonar: Let's discuss this offline. FYI my current thinking is as follows:

I was thinking that the detectors would expose a rule key, and we'd do the check here.

The considerations are the similar for the question about cancellation below - we could make the detectors as simple as possible, and have more of the logic here, or we could push more of the responsibility down to the individual detectors.

Currently, I think we should make the detectors as simple as possible.
* if we want to pass the rule configuration to the detectors then we need a shared API that the secrets assembly and SLVS can both reference, which we don't currently have.
* the detector rules will be direct ports of the equivalent Java code, and. I think keeping the code as minimal as possible would make it simpler to maintain i.e. don't add responsibilities in the .NET code that aren't in the Java code.

We could pass the cancellation token if we think the code in each detector will take long enough to run  and executes in multiple steps i.e. so it is worth checking for cancellation between each step.
At the moment it looks like the detectors will only have one main step, so it's probably not worth passing the token.

duncanp-sonar (approved):

rita-sonarsource:
@duncanp-sonar please re-review, finally finished

PR #2733, opened 2021-11-01
Changed BindingController to use IKnownUIContexts

duncanp-sonar:


src/Integration/Helpers/IKnownUIContexts.cs
```diff
@@ -23,11 +23,19 @@
...
+
     public interface IKnownUIContexts
     {
         event EventHandler<UIContextChangedEventArgs> SolutionBuildingContextChanged;
         event EventHandler<UIContextChangedEventArgs> SolutionExistsAndFullyLoadedContextChanged;
         event EventHandler<UIContextChangedEventArgs> CSharpProjectContextChanged;
         event EventHandler<UIContextChangedEventArgs> VBProjectContextChanged;
+
+        IUIContext SolutionExistsAndFullyLoadedContext { get; }
+        IUIContext SolutionExistsAndNotBuildingAndNotDebuggingContext { get; }
```
duncanp-sonar, line 39: Previously the wrapper just handled events. Now it needs to return UI contexts that can be queried.

src/Integration/Helpers/VsShellUtils.cs
```diff
@@ -85,26 +85,6 @@ public static void WriteToSonarLintOutputPane(IServiceProvider serviceProvider,
...
-        public static bool IsSolutionExistsAndFullyLoaded()
-        {
-            Debug.Assert(KnownUIContexts.SolutionExistsAndFullyLoadedContext != null, "KnownUIContexts.SolutionExistsAndFullyLoadedContext is null");
-            if (KnownUIContexts.SolutionExistsAndFullyLoadedContext != null)
-            {
-                return KnownUIContexts.SolutionExistsAndFullyLoadedContext.IsActive;
-            }
-            return false;
-        }
-
```
duncanp-sonar, line 107: The binding controller used these two static helper methods that in turn accessed `KnownUIContexts`. I've dropped these methods and changed the binding controller to reference `IKnownUIContexts` directly.

src/Integration/LocalServices/ErrorListInfoBarController.cs
```diff
@@ -261,7 +261,7 @@ private void CurrentErrorWindowInfoBar_Closed(object sender, EventArgs e)
             {
                 ClearCurrentInfoBar();
             }
-            catch(Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+            catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
```
duncanp-sonar, line 264: Unintended - reformatting by VS.

src/Integration/Helpers/KnownUIContextsWrapper.cs
```diff
@@ -74,5 +84,9 @@ public class KnownUIContextsWrapper : IKnownUIContexts
                 KnownUIContexts.VBProjectContext.UIContextChanged -= value;
             }
         }
+
+        public IUIContext SolutionExistsAndFullyLoadedContext => new UIContextWrapper(KnownUIContexts.SolutionExistsAndFullyLoadedContext);
+
+        public IUIContext SolutionExistsAndNotBuildingAndNotDebuggingContext => new UIContextWrapper(KnownUIContexts.SolutionBuildingContext);
```
duncanp-sonar, line 90: I've added the contexts needed by the binding controller. The source controlled file system and error list controller will need other contexts. I'll handle those separately.

src/Integration.UnitTests/Binding/BindingControllerTests.cs
```diff
@@ -156,32 +153,51 @@ public void BindingController_BindCommand_Status()
         }
 
         [TestMethod]
-        public void BindingController_BindCommand_Status_VsState()
```
duncanp-sonar, line 159: I've split this test into three separate tests:
* one to test the UIContext flags (cases 1 and 2),
* one to test the non-managed project kind (case 3)
* one to test the "no projects" scenario (case 4)

rita-sonarsource (approved):

PR #2734, opened 2021-11-01
Changed SourceControlledFileSystem to use IKnownUIContexts

duncanp-sonar:


src/Integration/Helpers/IKnownUIContexts.cs
```diff
@@ -35,7 +35,9 @@ public interface IKnownUIContexts
         event EventHandler<UIContextChangedEventArgs> CSharpProjectContextChanged;
         event EventHandler<UIContextChangedEventArgs> VBProjectContextChanged;
 
+        IUIContext SolutionBuildingContext { get; }
```
duncanp-sonar, line 38: Two extra new properties for contexts used by the SCC file system.

rita-sonarsource (approved):

PR #2735, opened 2021-11-01
Changed ErrorListInfoBarController to use IKnownUIContext

duncanp-sonar:


src/Integration.UnitTests/LocalServices/ErrorListInfoBarControllerTests.cs
```diff
@@ -57,14 +55,16 @@ public class ErrorListInfoBarControllerTests
...
+        private Mock<IKnownUIContexts> knownUIContexts;
         private ILogger logger;
 
         #region Test plumbing
 
         [TestInitialize]
         public void TestInit()
         {
-            KnownUIContextsAccessor.Reset();
+            ThreadHelper.SetCurrentThreadAsUIThread();
```
duncanp-sonar, line 66: The product code started asserting that it needed to be called on the UI thread (previously it was being invoked indirectly by the VS UI context on a different thread).

rita-sonarsource (approved):

PR #2750, opened 2021-11-03
Exclude secrets assemblies from VSIX

duncanp-sonar:


src/CloudSecrets.UnitTests/SecretsAnalyzerTests.cs
```diff
@@ -40,7 +40,7 @@ public class SecretsAnalyzerTests
         private const string ValidFilePath = "c:\\test";
         private const string ValidFileContent = "this is a content";
 
-        [TestMethod, Ignore("Secrets not yet supported")]
+        [TestMethod, Ignore("Secrets not yet supported")] // __secrets
```
duncanp-sonar, line 43: I haven't deleted anything; instead, I've comment bits out, and tagged them all with `__secrets` (including the MEF-changes you made earlier today).

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -260,10 +259,10 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
-    <Content Include="secrets\LICENSE.txt">
+    <!--<Content Include="secrets\LICENSE.txt">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <IncludeInVSIX>true</IncludeInVSIX>
-    </Content>
+    </Content>-->
```
duncanp-sonar, line 265: There is another section in this file that adds the secrets assemblies is included in the correct place in the VSIX:

```
  <!-- *************************************************************************** -->
  <!-- Sonar Secrets -->
  <!-- *************************************************************************** -->
  <!-- In the VSIX, the secrets binary(ies) should be in the secrets subfolder next to the specific license file.
       We don't need to copy the file(s). It's enough to specify the VSIXSubPath we want.
  -->
  <Target Name="SecretsBinariesToVSIX" AfterTargets="GeneratePkgDef" BeforeTargets="GetVsixSourceItems">
    <ItemGroup>
      <SonarSecretsFilesToAdd Include="$(OutDir)SonarLint.Secrets.*.dll" />
    </ItemGroup>
    
    <Message Importance="high" Text="Adding the SonarSecrets files to the VSIX:@(SonarSecretsFilesToAdd)" />
    
    <ItemGroup>
      <VSIXSourceItem Include="@(SonarSecretsFilesToAdd)">
        <VSIXSubPath>secrets</VSIXSubPath>
      </VSIXSourceItem>
    </ItemGroup>
  </Target>
```

I didn't bother to comment it out because it wasn't necessary - no files in the directory match `SonarLint.Secrets.*.dll`, so nothing gets copied.

duncanp-sonar:
It looks like a lot of changes, but most of them are to the package lock files.

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -28,7 +28,7 @@
 using Microsoft.VisualStudio.Shell.TableManager;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
-using SonarLint.VisualStudio.CloudSecrets;
+// __secrets using SonarLint.VisualStudio.CloudSecrets;
```
rita-sonarsource, line 31: the namespace is not in use so it can be removed

src/Integration/Telemetry/TelemetryManager.cs
```diff
@@ -23,7 +23,7 @@
 using System.Diagnostics;
 using System.Linq;
 using Microsoft.VisualStudio.Shell;
-using SonarLint.VisualStudio.CloudSecrets;
+//using SonarLint.VisualStudio.CloudSecrets; __secrets
```
rita-sonarsource, line 26: same comment, not really used but can be handled later

rita-sonarsource (approved):

PR #2758, opened 2021-11-05
Build in 2022 by default

rita-sonarsource:


rita-sonarsource:
@duncanp-sonar FYI I had to upload 2 nuget packages: 
microsoft.diagnostics.tracing.eventsource.redist.1.1.28
microsoft.netcore.targets.1.1.3

These were the versions used when testing with LockedMode on.

duncanp-sonar (approved):

PR #2775, opened 2021-11-16
Fix some SC warnings

### LINKED ISSUE #2768: Investigate SonarCloud test issues
### Description

The C# scanner is now reporting issues on test code, and SLVS has a number of new issues. We need to triage the newly-raised test issues.

Note:
The VSSDK meta-package includes the VSSDK analyzers, which are now also running on test code. They are producing lots of false positives which we can ignore (if we can't find some way of stopping them being reported in the first place).

rita-sonarsource:
Fixes #2768


duncanp-sonar (approved):
Looks good.

I suggest that when we do cleanup like this in the future we do test and product cleanup separately i.e. different commits. There's always a slight chance that a seemingly innocuous change will have an unexpected side-effect. From a risk-management point of view, if a commit has only test changes then it's no-risk from a product point of view. Commits containing any product change contain some risk.

PR #2782, opened 2021-11-21
`platformName` should be compared in a case insensitive way

### LINKED ISSUE #2781: platformName should be compared in a case insensitive way
### Description

Currently, we support `Win32` & `x64`. These are the default when a platform is set from visual studio. When the project is generated from CMake a user might set it to `win32` or `X64` instead. Or even custom `Win64` or `x86`

Related community thread: https://community.sonarsource.com/t/win32-platform-is-not-supportted-for-c-c-code-analysis-using-sonarlint-in-visual-studio/53439/2

### Repro steps

change the platform name to win32 and try to run the analysis
`<Platform>win32</Platform>`

### Expected behavior

`win32` should work like `Win32`

### Actual behavior

> Analysis error: System.ArgumentException: Unsupported PlatformName: win32

### Known workarounds

change platform to match the platform name case.

abbas-sabra-sonarsource:
Fixes #2781

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelper_CaptureTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyHelper_CaptureTests.cs
@@ -152,7 +152,12 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
         public void PlatformName()
         {
             CFamilyHelper.Capture.IsPlatformX64("Win32").Should().Be(false);
+            CFamilyHelper.Capture.IsPlatformX64("win32").Should().Be(false);
+            CFamilyHelper.Capture.IsPlatformX64("X86").Should().Be(false);
+            CFamilyHelper.Capture.IsPlatformX64("x86").Should().Be(false);
             CFamilyHelper.Capture.IsPlatformX64("x64").Should().Be(true);
+            CFamilyHelper.Capture.IsPlatformX64("X64").Should().Be(true);
+            CFamilyHelper.Capture.IsPlatformX64("WiN64").Should().Be(true);
 
             Action action = () => CFamilyHelper.Capture.IsPlatformX64("foo");
             action.Should().ThrowExactly<ArgumentException>().And.Message.Should()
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix.UnitTests/CFamily/CFamilyHelper_CaptureTests.cs
```diff
@@ -149,12 +149,23 @@ public void ToCaptures_Test2()
...
         [TestMethod]
-        public void PlatformName()
+        [DataRow("Win32", false)]
+        [DataRow("win32", false)]
+        [DataRow("X86", false)]
+        [DataRow("x86", false)]
+        [DataRow("x64", true)]
+        [DataRow("X64", true)]
+        [DataRow("WiN64", true)]
+        public void PlatformName_StringCaseIsIgnored(string platformName, bool isX64)
```
rita-sonarsource, line 159: @abbas-sabra-sonarsource FYI, TestCases make diagnosing test failures simpler

src/Integration.Vsix/CFamily/CFamilyHelper_Capture.cs
```diff
@@ -108,15 +108,18 @@ public static Capture[] ToCaptures(IFileConfig fileConfig, string path, out stri
 
             internal /* for testing */ static bool IsPlatformX64(string platformName)
             {
-                switch (platformName)
+                if (platformName.Equals("win32", StringComparison.OrdinalIgnoreCase) ||
```
rita-sonarsource, line 111: @abbas-sabra-sonarsource FYI for comparisons it's [a good practice](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#common-string-comparison-methods-in-net) to use Equals with a comparer instead of ToLower

rita-sonarsource:
@duncanp-sonar made some changes to the PR, feel free to review

src/Integration.Vsix/CFamily/CFamilyHelper_Capture.cs
abbas-sabra-sonarsource: @rita-sonarsource Thanks for the link. I find the switch more readable and extendable. I don't think the problem described there can happen in real life for the platformName. Yet if it can, I feel that the solution should be using `String.ToLowerInvariant ` instead of dropping the switch for ifs.
(not asking to revert the code, just making sure that I understood the reason correctly)

PR #2785, opened 2021-11-22
Run only necessary parts of VcxRequestFactory on the UI thread

rita-sonarsource:


src/Integration.Vsix/CFamily/VcxProject/VcxRequestFactory.cs
```diff
@@ -75,36 +75,21 @@ internal class VcxRequestFactory : IRequestFactory
 
         public async Task<IRequest> TryCreateAsync(string analyzedFilePath, CFamilyAnalyzerOptions analyzerOptions)
         {
-            IRequest request = null;
-            await threadHandling.RunOnUIThread(() => request = TryCreateSync(analyzedFilePath, analyzerOptions));
```
rita-sonarsource, line 79: Previously everything related to VcxRequest creation was running on the UI thread, including the effective rules calculation, hence why the UI was locking up (#2784).
Only the FileConfig creation needs to be done on the UI thread as it uses the DTE object, so I've moved everything else to the background thread.
I think in general we should take the "closest to source" approach and put UI switching next to the code that really needs it, similar to what the VS analyzers suggest to do. Otherwise it can lead to situations like this one, where we introduce a UI lockup due to code that should've been running on the background thread.

rita-sonarsource:
@duncanp-sonar tested with vcx / cmake / unsupported. Looks good.

src/Integration.Vsix/CFamily/VcxProject/VcxRequestFactory.cs
```diff
@@ -75,36 +75,28 @@ internal class VcxRequestFactory : IRequestFactory
...
-            return request;
-        }
-
-        private IRequest TryCreateSync(string analyzedFilePath, CFamilyAnalyzerOptions analyzerOptions)
-        {
-            LogDebug("Trying to create request for " + analyzedFilePath);
+            threadHandling.ThrowIfOnUIThread();
```
duncanp-sonar, line 78: I'd forgotten we'd added this thread-handling abstraction. Handy.

duncanp-sonar (approved):

src/Integration.Vsix/CFamily/VcxProject/VcxRequestFactory.cs
rita-sonarsource: Definitely!

PR #2787, opened 2021-11-24
Taint package load - handle exceptions when server is unavailable

### LINKED ISSUE #2788: Taint vulnerabilities package fails to load when the bound solution's server is inaccessable
### Description

Taint vulnerabilities package calls `TaintIssuesSynchronizer.SynchronizeWithServer`, which in turn checks if the solution is in connected mode and if the server's version supports taint vulnerabilities.

This code is not protected in a try-catch and fails when the solution is bound but the server is not available.

https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs#L86

rita-sonarsource:
Fixes #2788

### DIFF
--- a/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
+++ b/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
@@ -81,16 +81,16 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
 
         public async Task SynchronizeWithServer()
         {
-            var bindingConfiguration = configurationProvider.GetConfiguration();
-
-            if (IsStandalone(bindingConfiguration) || !IsConnected() || !IsFeatureSupported())
-            {
-                HandleNoTaintIssues();
-                return;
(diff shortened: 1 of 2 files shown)


src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -81,16 +81,16 @@ internal sealed class TaintIssuesSynchronizer : ITaintIssuesSynchronizer
...
-            {
-                HandleNoTaintIssues();
-                return;
-            }
-
             try
             {
+                var bindingConfiguration = configurationProvider.GetConfiguration();
+
+                if (IsStandalone(bindingConfiguration) || !IsConnected() || !IsFeatureSupported())
```
rita-sonarsource, line 88: When I saw the exception, the call to `IsFeatureSupported` was the one that failed.

duncanp-sonar (approved):

PR #2790, opened 2021-11-25
Issue template - add note to about using Preview tab

duncanp-sonar:


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,6 +1,8 @@
 Please use the community forum instead of creating issues in this repo.
 The community forum already has answers to many common questions/problems and is actively monitored by many SonarSourcers.
 
+(click on the "Preview" tab above to make the links below clickable)
```
rita-sonarsource, line 4: Perhaps instead of that, we can just list the links.
I.e. 
```... click [here] or visit https://....```

Comment (duncanp-sonar) [reply .github/ISSUE_TEMPLATE.md:4]: It would need to be e.g.

```
 * for support issues click [here](https://community.sonarsource.com/c/help/sl/11) or visit https://community.sonarsource.com/c/help/sl/11
```

... which is pretty unreadable unless you know to click on Preview.


We could say e.g.

```
 * for support issues visit https://community.sonarsource.com/c/help/sl/11
 etc
```

I still think the hint about switching to the Preview tab is useful in that case.

rita-sonarsource: Sure, that works 👍 (links + preview note)

rita-sonarsource (approved):

PR #2801, opened 2021-12-13
Fix CMake analysis

rita-sonarsource:


src/CFamily/CMake/VsDevCmdEnvironmentVarsProvider.cs
```diff
@@ -66,8 +66,8 @@ internal class VsDevCmdEnvironmentVarsProvider : IVsDevCmdEnvironmentProvider
         private readonly IFileSystem fileSystem;
 
         [ImportingConstructor]
-        public VsDevCmdEnvironmentVarsProvider(IVsInfoService vsInfoService, IThreadHandling threadHandling, ILogger logger)
-            : this(vsInfoService, threadHandling, logger, new ProcessFactory(), new FileSystem())
+        public VsDevCmdEnvironmentVarsProvider(IVsInfoService vsInfoService, ILogger logger)
```
rita-sonarsource, line 69: This caused the CMake's `CompilationDatabaseRequestFactory` not to load, so only Vcx factory was loaded, which caused an error when analyzing cmake files: 
```
Unable to retrieve the configuration for file 'C:\Users\Rita\source\repos\CMakeProject1\CMakeProject1\a.cpp'.
    Check the file is part of a supported project type in the current solution.
    ```

PR #2811, opened 2021-12-15
Make CFamily converter stateful

### LINKED ISSUE #2807: [PERF] Optimise processing CPP issues
### Description

Issues are individually streamed from the subprocess and processed individually. Processing requires the `ITextDocument` for the file(s) reported issue.

Currently, the converter is a MEF-imported shared instance i.e. a singleton so it can't cache the `ITextDocument` instances, and the "load text document" method is being called each time. However, it is expected that most of the reported issues will relate to the same file or small set of related files i.e. the actual file being analyzed, and its header files.

An analysis with hundreds of issues will cause the "load" method for those small set of files to the be loaded hundreds of times.

### Proposed optimisation
Make the converter stateful, with a new one being created for each analysis. The converter will cache the loaded `ITextDocument` instances, so they need only be loaded once per file.

Changes required:
- [x] add ETW events to allow measuring the impact
- [x] add a MEF-importable converter factory
- [x] change the converter to be stateful, and cache the loaded `ITextDocument` instances
- [x] measure the change in performance
- [x] fix up existing tests and add tests for the factory

#### Before: ETW trace without caching

File is loaded every time.
Average time to convert an issue: 5ms

Before

FileLoaded_Before

#### After: ETW trace with caching

File is loaded once.
Time for first conversion (involving load): around 5ms
Average time to convert an issue: < 0.1ms

FileLoaded_After1

duncanp-sonar:
Fixes #2807

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueConverterFactoryTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueConverterFactoryTests.cs
@@ -1,0 +1,62 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Utilities;
+using Moq;
+using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
+{
+    [TestClass]
+    public class CFamilyIssueConverterFactoryTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<CFamilyIssueConverterFactory, ICFamilyIssueConverterFactory>(null, new[]
+            {
+                MefTestHelpers.CreateExport<ITextDocumentFactoryService>(Mock.Of<ITextDocumentFactoryService>()),
+                MefTestHelpers.CreateExport<IContentTypeRegistryService>(Mock.Of<IContentTypeRegistryService>())
+            });
+        }
+
+        [TestMethod]
+        public void Create_ReturnsNewInstance()
+        {
+            var textDocumentFactory = Mock.Of<ITextDocumentFactoryService>();
+            var contentTypeRegistry = Mock.Of<IContentTypeRegistryService>();
+
+            var testSubject = new CFamilyIssueConverterFactory(textDocumentFactory, contentTypeRegistry);
+
+            // Create first item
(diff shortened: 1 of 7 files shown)


src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueConverterFactoryTests.cs
```diff
@@ -0,0 +1,62 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Utilities;
+using Moq;
+using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
+{
+    [TestClass]
+    public class CFamilyIssueConverterFactoryTests
```
duncanp-sonar, line 31: Tests that the new factory is exportable and creates unique instances.

src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueToAnalysisIssueConverterTests.cs
```diff
@@ -390,14 +380,55 @@ public void Convert_HasMessageParts_EachFileIsLoadedOnlyOnce()
 
             var message = new Message("rule2", "file2.cpp", 40, 3, 2, 1, "this is a test", false, messageParts.ToArray());
 
-            var fileSystemMock = CreateFileSystemMock();
-            var testSubject = CreateTestSubject(fileSystem: fileSystemMock.Object);
+            var fileSystemMock = CreateFileSystemMock(fileExists: true);
+            var textDocFactory = new Mock<ITextDocumentFactoryService>();
+
+            var testSubject = CreateTestSubject(fileSystem: fileSystemMock.Object, textDocFactory: textDocFactory.Object);
```
duncanp-sonar, line 386: Tweaked the `CreateTestSubject` so we can test which documents are loaded.

src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
```diff
@@ -63,10 +102,14 @@ public CFamilyIssueToAnalysisIssueConverter(ITextDocumentFactoryService textDocu
             this.lineHashCalculator = lineHashCalculator;
             this.fileSystem = fileSystem;
             filesContentType = contentTypeRegistryService.UnknownContentType;
+
+            pathToTextDocMap = new Dictionary<string, ITextDocument>(StringComparer.OrdinalIgnoreCase);
```
duncanp-sonar, line 106: The subprocess should use a consistent case for the file paths it returns. However, we tolerate inconsistent casing.

src/Integration/ETW/CodeMarkers.cs
```diff
@@ -87,6 +89,40 @@ public static class Keywords
 
         #endregion
 
+        #region Analysis: 2000-2999
```
rita-sonarsource, line 92: From readability point of view, it might help to rename this to "General analysis" and the other to "CFamily analysis"

src/Integration.Vsix/CFamily/MessageHandler.cs
```diff
@@ -109,6 +112,8 @@ public void HandleMessage(Message message)
                     HandleAnalysisIssue(message);
                     break;
             }
+
+            CodeMarkers.Instance.HandleMessageStop();
```
rita-sonarsource, line 116: how does ETW know to correlate between the start and end? In my POC I passed a guid to both stop and start.

src/Integration.Vsix/CFamily/CFamilyIssueToAnalysisIssueConverter.cs
rita-sonarsource: It's good because I believe the subprocess is actually inconsistent between the main message and the secondary locations. I vaguely remember a bug handling that.

rita-sonarsource (approved):

src/Integration.Vsix/CFamily/MessageHandler.cs
duncanp-sonar: No idea. It may not - it might be logic in PerfView that is doing the calculation. It's open source, so we should be able to find out.

src/Integration/ETW/CodeMarkers.cs
duncanp-sonar: `IMessageHandler` is actually CFamily-specific (I originally thought it was generic). So we don't have any generic analysis events at the moment.

I'll move the HandleMessage events to the `CFamily` section (which will leave `Analysis` range empty, so I'll remove it).

PR #2822, opened 2022-01-18
Fix2694: don't show "hide issue visualisation" menu for issues without secondary locations

ugras-ergun-sonarsource:
2694 Added existence of secondary location control for selected primary items

### DIFF
--- a/src/IssueViz.UnitTests/Editor/QuickActions/IssueLocationActionsSourceTests.cs
+++ b/src/IssueViz.UnitTests/Editor/QuickActions/IssueLocationActionsSourceTests.cs
@@ -24,11 +24,13 @@ using System.Linq;
 using System.Threading;
 using FluentAssertions;
 using Microsoft.VisualStudio.Language.Intellisense;
+using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Text.Editor;
 using Microsoft.VisualStudio.Text.Tagging;
+using Microsoft.VisualStudio.Threading;
 using Moq;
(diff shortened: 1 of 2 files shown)


src/IssueViz.UnitTests/Editor/QuickActions/IssueLocationActionsSourceTests.cs
```diff
@@ -326,7 +353,9 @@ private static IAnalysisIssueVisualization CreateIssueViz(params IAnalysisIssueF
             var testSubject = CreateTestSubject(selectedIssueLocationsTagAggregator.Object, issueLocationsTagAggregator.Object, selectionService.Object);
             var actualActionsSet = testSubject.GetSuggestedActions(null, mockSpan, CancellationToken.None);
 
-            return actualActionsSet.ToList();
+            var hasActions = testSubject.HasSuggestedActionsAsync(null, mockSpan, CancellationToken.None).Result;
```
duncanp-sonar, line 356: This looks weird: each unit test is now calling two separate public methods on the test subject. At the very least, it makes all of the test names misleading since the names suggest that only `GetSuggestedActions` is being called.

ugras-ergun-sonarsource: Yes but results of both these methods are tied together and it is not expected them to be act independently. Maybe changing the test names would makes sense but separating the tests would only lead to code duplication.

duncanp-sonar: Ok, makes sense. Please change the method names for this method and calling tests, and add a comment saying why we're calling two public APIs in the this case.

ugras-ergun-sonarsource: I have made the changes

duncanp-sonar (approved):

PR #2824, opened 2022-01-19
Drop VS2015-specific DeprecationManager

### LINKED ISSUE #2786: Drop DeprecationManager
### Description

The `DeprecationManager` class displays a warning about VS2015 < Update 3 not being supported.
It is now redundant since support for VS2015 has been dropped.

TODO
- [x] drop the class
- [x] clean up any methods only called by the class

duncanp-sonar:
Fixes #2786 

The DeprecationManager warned users of VS < 2015 Update 3 that support for that version would be dropped.
We no longer ship SLVS for VS2015 so this code is redundant.


src/Integration.UnitTests/Connection/ConnectionWorkflowTests.cs
```diff
@@ -595,28 +594,6 @@ public async Task ConnectionWorkflow_DownloadServiceParameters_Cancelled_AbortsW
             AssertServiceDisconnectCalled();
         }
 
-        [TestMethod]
-        public void IsSonarQubePluginSupported_WhenUnsupportedVersionOfVSAndCSharpPluginEquals70_ReturnsFalseAndWriteExpectedText()
-        {
-            TestPluginSupport(
-                expectedResult: false,
-                vsVersion: "14.0",
```
duncanp-sonar, line 603: The deleted tests were all testing what happened in V2015 RTM, so they are no longer needed.

rita-sonarsource (approved):
Woohoo! 🚀

ugras-ergun-sonarsource (approved):

PR #2825, opened 2022-01-24
Ue/2506 show visualization pane

ugras-ergun-sonarsource:


duncanp-sonar:
@ugras-ergun-sonarsource the PR is targeting the wrong branch.

src/IssueViz.Security/IssueViz.Security.csproj
```diff
@@ -32,6 +32,17 @@
 
   </PropertyGroup>
 
+  <ItemGroup>
```
duncanp-sonar, line 35: Are all of these changes necessary?

It looks like you've added a copy of the `sonarlint.png` file. If so, is it necessary to change the references to the existing file?

src/IssueViz.Security/IssueViz.Security.csproj
```diff
@@ -64,6 +75,7 @@
       <Generator>XamlIntelliSenseFileGenerator</Generator>
       <CopyToOutputDirectory>Never</CopyToOutputDirectory>
     </Page>
+    <Resource Include="Taint\TaintList\sonarlint.png" />
```
duncanp-sonar, line 78: Project files are easier to understand if related changes are grouped together e.g. group removing this from the `None` group with adding it to the `Resource` group.

Sometimes it makes sense to use a separate `ItemGroup` with an explanatory label e.g.

```
<ItemGroup Label="XAML resources">
  <None Remove="Taint\TaintList\sonarlint.png" />
  <Resource Include="Taint\TaintList\sonarlint.png" />
</ItemGroup>
```

src/IssueViz.Security/IssueViz.Security.csproj
ugras-ergun-sonarsource: those lines were added by vs automatically. When i tried to remove those in cleanup Logo was not visible on the menu.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -232,6 +233,69 @@ public void Dispose_UnregisterFromSelectionChangedEvent()
...
+
+            menuCommandService.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public void ShowVisualizationPaneCommand_Execute_LocationNavigated()
+        {
+            var menuCommandService = new Mock<IMenuCommandService>();
+
+            var viewModel = new Mock<ITaintIssueViewModel>();
```
duncanp-sonar, line 274: FYI if you are creating a mock just to get a valid object (i.e. if the default values are ok and you don't need to verify the mock afterwards) you can use `Mock.Of<ITaintIssueViewModel>`.  It's slightly shorter, but more importantly it makes it clear that the object won't been checked later.

duncanp-sonar: Ah; I see you've used `Mock.Of<>` above. It's good to be consistent.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -232,6 +233,69 @@ public void Dispose_UnregisterFromSelectionChangedEvent()
...
+
+            var testSubject = CreateTestSubject(menuCommandService: menuCommandService.Object);
+            testSubject.ShowVisualizationPaneCommand.Execute(viewModel.Object);
+
+            menuCommandService.Verify(x => x.GlobalInvoke(It.IsAny<CommandID>()), Times.Once);
```
duncanp-sonar, line 279: Is there are reason not to check the expected command is being called?

src/IssueViz.Security/IssueViz.Security.csproj
```diff
@@ -32,6 +32,17 @@
...
+  <ItemGroup Label="Taint Xaml Sources">
+    <None Remove="Taint\TaintList\sonarlint.png" />
+    <Resource Include="Taint\TaintList\sonarlint.png" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="Commands\Resources\sonarlint.png">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
```
duncanp-sonar, line 44: You've reverted part of the change to `Commands\Resources\sonarlint.png` in the project file but not this bit.
This file isn't referenced by any of your code changes, is it? Is it needed?

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
ugras-ergun-sonarsource: Because it's a reference type i didn't think it would work. But after the comment I've looked at the de-compiled code and saw equals method was overridden

duncanp-sonar: An alternative approach in that case would have been to capture the value that was passed and assert against the individual fields instead.

duncanp-sonar (approved):

PR #2855, opened 2022-02-09
Apply a quick-fix only if the text was unchanged

rita-sonarsource:


src/IssueViz/Models/AnalysisIssueVisualizationConverter.cs
```diff
@@ -123,8 +123,9 @@ private IReadOnlyList<IQuickFixVisualization> GetQuickFixVisualizations(IAnalysi
                 var editVisualizations = fix.Edits.Select(edit =>
                 {
                     var editSpan = issueSpanCalculator.CalculateSpan(edit, textSnapshot);
+                    var originalText = textSnapshot.GetText(editSpan);
```
rita-sonarsource, line 126: This is the first computation: when we calculate the original span, we also store a copy of the text as it was during the calculation. This code will be called immediately upon getting the analysis results and (hopefully) will be identical to what was analyzed.
It is still possible that this would be wrong: for example, if we trigger an analysis and it takes 1 minute to return back the issue from the subprocess. During this time, the user might have edited the code, and both the span and the text would be incorrect.
I don't think we have a way to handle that, not unless the subprocess also returns us the text that they used to compute the fix.

src/IssueViz/Models/QuickFixVisualization.cs
```diff
@@ -29,38 +32,69 @@ public interface IQuickFixVisualization
...
+
+        public bool IsApplicable(ITextSnapshot currentSnapshot) =>
+            EditVisualizations.All(x => IsApplicable(x, currentSnapshot));
+
+        private bool IsApplicable(IQuickFixEditVisualization editViz, ITextSnapshot currentSnapshot)
+        {
+            var updatedSpan = spanTranslator.TranslateTo(editViz.Span, currentSnapshot, SpanTrackingMode.EdgeExclusive);
+            var currentText = currentSnapshot.GetText(updatedSpan);
+
+            return string.Equals(currentText, editViz.OriginalText, StringComparison.InvariantCultureIgnoreCase);
```
rita-sonarsource, line 71: simple text comparison for now

src/IssueViz/Models/QuickFixVisualization.cs
```diff
@@ -29,38 +32,69 @@ public interface IQuickFixVisualization
...
+            this.spanTranslator = spanTranslator;
             Fix = fix;
             EditVisualizations = editVisualizations;
         }
 
         public IQuickFix Fix { get; }
 
         public IReadOnlyList<IQuickFixEditVisualization> EditVisualizations { get; }
+
+        public bool IsApplicable(ITextSnapshot currentSnapshot) =>
```
rita-sonarsource, line 63: I decided to put it on the Visualization class itself since it made the mocking very easy in the rest of the code.
A quick fix now exposes IsApplicable method and the underlying implementation checks if all the edits still have the same text.

src/IssueViz/Editor/QuickActions/QuickFixes/QuickFixActionsSource.cs
```diff
@@ -71,11 +70,13 @@ internal sealed class QuickFixActionsSource : ISuggestedActionsSource
         {
             var allActions = new List<ISuggestedAction>();
 
-            if (IsOnIssueWithQuickFixes(range, out var issuesWithFixes))
+            if (IsOnIssueWithApplicableQuickFixes(range, out var issuesWithFixes))
```
rita-sonarsource, line 73: IsOnIssueWithApplicableQuickFixes returns true if the quick fix has at least 1 applicable fix: this would be used by VS to know if it should even render the lightbulb. Then `applicableFixes` are filtered to create an action only for the applicable fixes.

src/IssueViz/Editor/QuickActions/QuickFixes/QuickFixActionsSource.cs
```diff
@@ -88,19 +89,20 @@ public async Task<bool> HasSuggestedActionsAsync(ISuggestedActionCategorySet req
...
+        private bool IsOnIssueWithApplicableQuickFixes(SnapshotSpan range, out IEnumerable<IAnalysisIssueVisualization> issuesWithFixes)
         {
             var tagSpans = issueLocationsTagAggregator.GetTags(range);
 
             issuesWithFixes = tagSpans
                 .Select(x => x.Tag.Location)
                 .OfType<IAnalysisIssueVisualization>()
-                .Where(x => x.QuickFixes.Any());
+                .Where(x =>
+                    x.QuickFixes.Any(fix => fix.IsApplicable(textView.TextSnapshot)));
```
rita-sonarsource, line 105: There is a potential performance hit here.
For every quick fix, we will check 3 times if it's still applicable:
1. Once by VS to know if it should render the light bulb
2. Another time when the user opens the lightbulb to populate the list of available actions
3.  A third time when the quick fix is applied

So 3 times we will do a span translation and string comparison of all the edits, of all the fixes that an issue has.

I don't know if it's expensive or not. I didn't notice a performance hit but I was testing on a small class that returned a small amount of issues, with very few fixes. Just something to keep in mind that we might need to optimize this.

src/IssueViz/Editor/QuickActions/QuickFixes/QuickFixSuggestedAction.cs
```diff
@@ -52,7 +52,8 @@ internal class QuickFixSuggestedAction : BaseSuggestedAction
 
         public override void Invoke(CancellationToken cancellationToken)
         {
-            if (cancellationToken.IsCancellationRequested)
+            if (cancellationToken.IsCancellationRequested ||
+                !quickFixVisualization.IsApplicable(textBuffer.CurrentSnapshot))
```
ugras-ergun-sonarsource, line 56: Is not doing anything here a good solution? I think we should somehow inform the user that quick fix is not applicable anymore.

src/IssueViz/Models/QuickFixVisualization.cs
ugras-ergun-sonarsource: Why are we ignoring the case? If the case sensitive change in a class name it could invalidate the span.

rita-sonarsource: You're right 👍 I'll fix

src/IssueViz/Editor/QuickActions/QuickFixes/QuickFixSuggestedAction.cs
rita-sonarsource: I'll add logging to the output window so we would at least know it happened. I think a message box might be too intrusive, I'm not sure what would be the best UX. There is also a status bar in VS but no one ever notices it.

src/IssueViz/Models/AnalysisIssueVisualizationConverter.cs
duncanp-sonar: > It is still possible that this would be wrong: for example, if we trigger an analysis and it takes 1 minute to return back the issue from the subprocess.

We already try to minimise this risk as much as possible. The `AccumulatingIssueConsumer` has a reference to the snapshot at the point the analysis was triggered. That is the snapshot that is passed to the issue converter, so any subsequent edits by the user won't be visible.

There is a window in which the user could make edits that would mess things up, but that window is between when we trigger the analysis and the subprocess reads in the file.

There's a comment about this in the `AccumulatingIssueConsumer`. Maybe we need a short comment in the converter too e.g. that the caller is responsible for providing the snapshot that matches the raised issues.


https://github.com/SonarSource/sonarlint-visualstudio/blob/63a9400ef82f8e62298b8ab2eb433723dc17a8f8/src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs#L48

src/IssueViz/Models/QuickFixVisualization.cs
```diff
@@ -29,38 +32,69 @@ public interface IQuickFixVisualization
...
+        }
     }
 
     public interface IQuickFixEditVisualization
     {
         IEdit Edit { get; }
 
         SnapshotSpan Span { get; }
+
+        string OriginalText { get; }
```
duncanp-sonar, line 81: `SnapshotSpan` has a `GetText` method - do we need to store the text separately ourselves too? (might depend on whether we ever update the span we store).

duncanp-sonar (commented):
Just a question - do we need a separate "OriginalText" property?

ugras-ergun-sonarsource (approved):
LGTM

PR #2856, opened 2022-02-10
Ue/2839 remove applied fixes

ugras-ergun-sonarsource:


src/IssueViz/Editor/QuickActions/QuickFixes/QuickFixSuggestedAction.cs
```diff
@@ -64,6 +66,7 @@ public override void Invoke(CancellationToken cancellationToken)
                 textEdit.Replace(updatedSpan, edit.Edit.Text);
             }
 
+            issueViz.InvalidateSpan();
```
ugras-ergun-sonarsource, line 69: Can't validate this method call in tests, because it's an extension method.

rita-sonarsource: You can test the expected behavior.
InvalidateSpan sets the span to empty, so you can create an IssueViz with a non-empty span and assert that it becomes empty afterwards. You can look at a similar behavior in `SetCurrentFilePath_FilePathIsNull_SpanIsInvalidated`

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTaggerTests.cs
```diff
@@ -364,6 +364,41 @@ public void GetTags_DifferentSnapshots_SnapshotsAreTranslated()
             testSubject.TagSpans[1].Tag.Location.Span.Value.Snapshot.Should().Be(snapshotMock2);
         }
 
+
+        [TestMethod]
+        public void GetTags_InvalidatedSpans_ShouldBeFiltered()
```
rita-sonarsource, line 369: A lot of the code in this test is made to assert span translations.
I think it can be simplified, as we don't care about translation in this case, we just want the tag to be removed completely.
`GetTags_HasIntersectingSpans_ReturnsExpectedTags` is a shorter version, what do you think about basing the test on that?

src/IssueViz.UnitTests/Editor/LocationTagging/LocationTaggerTests.cs
```diff
@@ -364,6 +364,41 @@ public void GetTags_DifferentSnapshots_SnapshotsAreTranslated()
...
+            testSubject.TagSpans[0].Span.Snapshot.Should().Be(snapshotMock1);
+            testSubject.TagSpans[1].Span.Snapshot.Should().Be(snapshotMock1);
+            testSubject.TagSpans[0].Tag.Location.Span.Value.Snapshot.Should().Be(snapshotMock1);
+            testSubject.TagSpans[1].Tag.Location.Span.Value.Snapshot.Should().Be(snapshotMock1);
+
+            testSubject.TagSpans[1].Tag.Location.Span = new SnapshotSpan();
+
+            testSubject.GetTags(inputSpans).ToArray(); // GetTags is lazy so we need to reify the result to force evaluation
+
+            testSubject.TagSpans.Count().Should().Be(1);
```
rita-sonarsource, line 399: Worth also asserting that the tag that remains is really the expected one

src/IssueViz.UnitTests/Editor/ErrorTagging/ErrorTaggerTests.cs
```diff
@@ -65,6 +65,38 @@ public void GetTags_FilterIsApplied_ExpectedTagsCreated()
...
 
+        [TestMethod]
+        public void GetTags_EmptySpansShouldBeFiltered()
+        {
+            var snapshot = CreateSnapshot(length: 50);
+
+            var inputSpans = CreateSpanCollectionSpanningWholeSnapshot(snapshot);
+
+            var primary1 = CreateTagSpanWithPrimaryLocation(snapshot, new Span(1, 5), message: "error message 1", ruleKey: "cpp:S5350");
+            var primary2 = CreateTagSpanWithPrimaryLocation(snapshot, new Span(10, 0), message: "error message 2", ruleKey: "cpp:emptyCompoundStatement");
```
rita-sonarsource, line 76: This is a very subtle change, I had to do diff compare with the test above to know what actually makes the location invalid.
I think it would help to extract it to a variable with an explicit name (i.e. `invalidSpan`) and assign it.

Also, the secondary locations seem irrelevant here.

rita-sonarsource (changes requested):
@ugras-ergun-sonarsource some comments on the tests

Issue #2841, opened 2022-01-27
[Quick fixes] Add telemetry


rita-sonarsource:
### Description

Add telemetry for the applied quick fixes. Format should match the telemetry sent in the other SL flavors.

https://jira.sonarsource.com/browse/SLLS-41

PR #2868, opened 2022-02-15
Add QuickFixes to telemetry payload

rita-sonarsource:
Fixes [#2841](https://github.com/SonarSource/sonarlint-visualstudio/issues/2841)

### DIFF
--- a/src/Core/Telemetry/IQuickFixesTelemetryManager.cs
+++ b/src/Core/Telemetry/IQuickFixesTelemetryManager.cs
@@ -18,7 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-namespace SonarLint.VisualStudio.IssueVisualization.Editor.QuickActions.QuickFixes
+namespace SonarLint.VisualStudio.Core.Telemetry
 {
     public interface IQuickFixesTelemetryManager
     {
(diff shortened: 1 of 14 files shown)


src/Core/Telemetry/IQuickFixesTelemetryManager.cs
```diff
@@ -18,7 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-namespace SonarLint.VisualStudio.IssueVisualization.Editor.QuickActions.QuickFixes
+namespace SonarLint.VisualStudio.Core.Telemetry
```
rita-sonarsource, line 21: Changed from the base branch: I had to move the manager to Core so it could be used in issueviz and implemented by telemetrymanager in integration.

src/Integration.UnitTests/Telemetry/TelemetryManagerTests.cs
```diff
@@ -731,6 +733,103 @@ public void SendPayload_ListIsOrdered()
             telemetryRepositoryMock.Verify(x => x.Save(), Times.Once);
         }
 
+        [TestMethod]
+        public void QuickFixApplied_NoPreviousIssues_TelemetryUpdated()
```
rita-sonarsource, line 737: This is very blatant copy-paste of the secrets tests. The product code and the tests are identical except the list is different, but I can't think of a neater way to do this.

duncanp-sonar (approved):

PR #2869, opened 2022-02-16
Embed SonarC# 8.35 and SonarVB 8.35

ugras-ergun-sonarsource:


rita-sonarsource (approved):
@ugras-ergun-sonarsource LGTM, please also run sanity (i.e. open a C# file and see that you can still analyze it), just to be sure nothing is broken in runtime

PR #2870, opened 2022-02-17
Fix multithreading issue in TelemetryManager

### LINKED ISSUE #2867: Error recording analysis telemetry
### Description

The telemetry XML file can be corrupted, which then causes SLVS to throw when trying to record subsequent analyses.

**Root cause:  **
The invalid XML file below from my machine, but I haven't been able to reproduce the steps that cause it.

Prior to the exception I had opened an C++ project in a VS instance that does not have the C++ workload installed, so I assumed the bug was related to #2853. However, I haven't managed to repro it subsequently.

**Impact: **
The exceptions are caught and handled internally i.e. no immediately visible user impact.
However, the bug prevents subsequent telemetry about language analysis from being recorded correctly. It is not clear whether analysis data is lost only until the next "reset" of the telemetry (i.e. when the file is reset after the data is uploaded), or permanently.

#### Example of invalid XML

[telemetry - bad.xml.txt](https://github.com/SonarSource/sonarlint-visualstudio/files/8069402/telemetry.-.bad.xml.txt)

```
  <Analyses>
    <Analysis xsi:nil="true" />
    <Analysis>
      <Language>cpp</Language>
    </Analysis>
  </Analyses>
```

#### Subsequent exception
 
image

### Affected versions
Initially observed in VS2019. The file is in the per-user roaming profile, so it will affect all of instances of VS for that user on any machine they use in the same domain.

rita-sonarsource:
Fixes #2867

### DIFF
--- a/src/Integration/Telemetry/TelemetryManager.cs
+++ b/src/Integration/Telemetry/TelemetryManager.cs
@@ -213,10 +213,13 @@ namespace SonarLint.VisualStudio.Integration
             Debug.Assert(!string.IsNullOrEmpty(languageKey), "Supplied languageKey should not be null/empty");
             Debug.Assert(telemetryRepository.Data != null);
 
-            if (!telemetryRepository.Data.Analyses.Any(x => string.Equals(x.Language, languageKey, StringComparison.OrdinalIgnoreCase)))
+            lock (Lock)
             {
-                telemetryRepository.Data.Analyses.Add(new Analysis { Language = languageKey });
-                telemetryRepository.Save();
+                if (!telemetryRepository.Data.Analyses.Any(x => string.Equals(x.Language, languageKey, StringComparison.OrdinalIgnoreCase)))
+                {
+                    telemetryRepository.Data.Analyses.Add(new Analysis { Language = languageKey });
+                    telemetryRepository.Save();
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Integration/Telemetry/TelemetryManager.cs
```diff
@@ -213,10 +213,13 @@ public void LanguageAnalyzed(string languageKey)
             Debug.Assert(!string.IsNullOrEmpty(languageKey), "Supplied languageKey should not be null/empty");
             Debug.Assert(telemetryRepository.Data != null);
 
-            if (!telemetryRepository.Data.Analyses.Any(x => string.Equals(x.Language, languageKey, StringComparison.OrdinalIgnoreCase)))
+            lock (Lock)
```
rita-sonarsource, line 216: As we discussed offline, lack of Lock resulted in non thread-safe access to the list: it was possible for List.Any to execute while was another thread was calling List.Add.

src/Integration/Telemetry/TelemetryManager.cs
```diff
@@ -248,16 +251,13 @@ public void SecretDetected(string ruleId)
         {
             var rulesUsage = telemetryRepository.Data.RulesUsage;
 
-            if (!rulesUsage.RulesThatRaisedIssues.Contains(ruleId))
+            lock (Lock)
```
rita-sonarsource, line 254: Similar to the issue above, List.Contains shouldn't be called outside of the lock if another thread can call List.Add in the meanwhile. For readability, it is easier to just wrap everything in a Lock, as that also removes the need for Distinct.

rita-sonarsource:
@duncanp-sonar @ugras-ergun-sonarsource please review. I didn't add the test for it as the test itself was flaky due to the nature of this bug.

duncanp-sonar (approved):

ugras-ergun-sonarsource (approved):

PR #2871, opened 2022-02-18
Ue/2860 first line empty

### LINKED ISSUE #2860: Fix false negatives on file's issues when the first line is empty
### Description

Issues reported with `endline==0 `is meant to be issues on the entire file. In SLVS these issues are reported on the entire first line. That is okay if the first line is not empty. But if the first line is empty this issue will not be shown and the rule will not trigger.

### Reproduction steps
1- enable a file level rule like `cpp:CommentMixedStyles`:
>    "cpp:CommentMixedStyles": {
>       "level": "On"
>     },

2-  analyzed this file. CommentMixedStyles is raised on the first line:

```
// the rule is raised on this line
typedef int x1;

/*      */
//
```

3- Make the first line empty, and the rule will no longer be raised:

```

// the issue is not raised on this file
typedef int x1;

/*      */
//
```

### Expected results
The issue appears in the Error List and there are squigglies under the opening comment in the editor.

### Actual results
The issue does not appear in the Error List. There are no  squigglies in the editor.

### Affected versions:
?All SLVS versions to date that support C++ analysis.
Affected languages: C/C++

ugras-ergun-sonarsource:
fixes #2860


src/Integration.Vsix.UnitTests/CFamily/CFamilyIssueToAnalysisIssueConverterTests.cs
```diff
@@ -238,15 +288,26 @@ public void Convert_HasMessageParts_LineHashCalculatedForIssueAndLocations()
...
+            const string fileLevelLocationFilePath = "file1.cpp";
+            const string nonFileLevelLocationFilePath = "file2.cpp";
+            const int nonFileLevelLocationLine = 20;
+
+            var textSnapshot = CreateTextSnapShot(Array.Empty<ITextSnapshotLine>());
+            var textBuffer = CreateTextBuffer(textSnapshot);
+            var textDocument = CreateTextDocument(textBuffer);
+
             var fileSystemMock = CreateFileSystemMock();
+            fileSystemMock.Setup(x => x.File.Exists(fileLevelLocationFilePath)).Returns(true);
```
ugras-ergun-sonarsource, line 304: For file level issues we need to get the snapshot so we can check if there are any leading empty lines.

Issue #2661, opened 2021-09-23
Use subprocess SL mode for the .vcxproj


abbas-sabra-sonarsource:
**Why**
- Reduce the maintenance cost of`.vcxproj` projects.
- Propagate new CFamily compiler-related features automatically to`.vcxproj` projects.

Details:
We introduced a new subprocess SL mode. This mode moves most of the CFamily compilers/driver’s details to the CFamily side. This mode is used for CMake projects.
For `.vcxproj` we still use the old subprocess mode. This means:
- A duplicate version of the driver is maintained on the SonarLint side: `MsvcDriver.cs`
- The scanner protocol data is needlessly maintained on the SonarLint side: `Protocol.cs`. The data sent in this protocol, unlike with the subprocess SL mode, need to be continuously updated with fields irrelevant to SL.
- Maintaining mapping of the compilation options to the command line equivalence: `CFamilyHelper_Capture.cs`
- Supporting a new compiler or a compiler feature requires duplicating the work on the SonarLint side.

**What**
Use the subprocess SL mode for the `.vcxproj`.
Drop the no longer needed duplicated and hard-to-maintain CFamily specific code.
**How**

1. Load all the information required by the subprocess SL mode from the `.vcxproj` project(pretty straightforward):
     - Compilation Database: 
       - File: absolute path of the file to be analyzed
       - Directory: VS project Directory
       - Command: 
               vcFileSettings.GetEvaluatedPropertyValue("ClCompilerPath")
                                                                  +
                                                "All Options" field found here:
![image](https://user-images.githubusercontent.com/49131500/134556500-4b97f523-9dd2-41e0-aa47-40b825888ecf.png)
                                                                  + 
                                         Additional options (already loaded - no change)
                                                                  +
                                                          File full Path

   - Include env variables: vcFileSettings.GetEvaluatedPropertyValue("AdditionalIncludeDirectories"),
   - Preamble/Reproducer/Header file identification/Header language detection remain the same.

2. Use the Procol of the subprocess SL mode.
3. Drop the `PortedFromJava` Directory.
4. Drop most of the file config loading. Since we no longer have to load each option separately we just load "All Options": `FileConfig.cs`
5. Drop the mapping from VS option to command line: `CFamilyHelper_Capture.cs`

PR #2874, opened 2022-02-20
Update embedded CFamily analyzer from 6.29.0.41127 to 6.31.0.44497

### LINKED ISSUE #2873: Update embedded CFamily analyzer from 6.29.0.41127 to 6.31.0.44497
###  v6.30.0.42324:
Release Notes: https://jira.sonarsource.com/secure/ReleaseNote.jspa?projectId=10090&version=17126
Depends on #2661 as the 6.30 version changed the MSVC driver and the Analyzer Request content.
After #2661 the changes should be transparent to SLVS.
This version improves the handling of many MSVC compiler options that are listed in the release notes.
### 6.31.0.44497:
Release Notes: https://jira.sonarsource.com/secure/ReleaseNote.jspa?projectId=10090&version=17229
This version added 15 new quick fixes and improved MSVC configuration support

abbas-sabra-sonarsource:
fixes #2873 depends on #2661 (PR #2872)


duncanp-sonar:
Hi @abbas-sabra-sonarsource. It looks like v6.32 may contain breaking changes around the legacy key changes so we'll need to do some more investigation before we merge it (see #2956).

Can you think of anything that might cause a breaking change in v6.31? I can't see anything that looks problematic in the [release notes](https://jira.sonarsource.com/secure/ReleaseNote.jspa?projectId=10090&version=17229).

Assuming v6.31 doesn't require any more work on our side it's likely that we'll release that first and v6.32 later.

abbas-sabra-sonarsource:
Hello @duncanp-sonar, I confirm there are no breaking changes in `6.31`. It makes sense to release `6.31` first and take the necessary time to do the investigation around legacy keys before `6.32`

abbas-sabra-sonarsource:
@duncanp-sonar I changed the ticket and branch to  target `6.31` and created a separate issue(#2958) for 6.32

src/Integration.Vsix/CFamily/VcxProject/FileConfig.cs
```diff
@@ -37,76 +35,45 @@ public static FileConfig TryGet(ILogger logger, ProjectItem dteProjectItem, stri
...
-            if (vcFile.ItemType == "ClInclude")
+            var compilerPath = vcConfig.GetEvaluatedPropertyValue("ClCompilerPath");
+            if (string.IsNullOrEmpty(compilerPath))
             {
-                // If the project language is not specified. Headers are compiled as CPP
-                compileAs = (compileAs == "Default") ? "CompileAsCpp" : compileAs;
-                // Use PCH in header files if project PCH is configured and not enforced through force include
-                // In normal code, it should be self to assume this since PCH should be used on every CPP file
-                if (string.IsNullOrEmpty(forcedIncludeFiles) && precompiledHeader == "Use")
+                // in case ClCompilerPath is not available on VS2017
```
rita-sonarsource, line 50: @abbas-sabra-sonarsource we've dropped support for VS 2017, so this code would only run in VS 2019+. Does this have any implications on the code below? Can it be simplified?

src/Integration.Vsix/CFamily/MessageHandler.cs
```diff
@@ -105,7 +105,10 @@ public void HandleMessage(Message message)
                     logger.WriteLine(CFamilyStrings.MsgHandler_ReportUnexpectedFailure, message.Text);
                     break;
 
-                case "internal.fileDependency": // not currently handled. See https://github.com/SonarSource/sonarlint-visualstudio/issues/2611
+                // Rules that start with internal shouldn't be treated as issues.
+                // Some of them should be handled like `internal.fileDependency`. See: https://github.com/SonarSource/sonarlint-visualstudio/issues/2611
+                // Others should can simply ignored like `internal.z3RefutationRate`, which is used to log in the scanner how many issues are rejected by the Z3 solver
```
rita-sonarsource, line 110: is this a bug fix? is this something that we already accidently report to the user?

src/Integration.Vsix/CFamily/VcxProject/FileConfig.cs
abbas-sabra-sonarsource: Even though we dropped support to VS 2017, its toolchain can still be used with newer vs. So better to keep it. maybe we can improve the comment to say `VS2017 toolchains`

rita-sonarsource: I'll update the comment, toolchains helps to understand the problem much better 👍

src/Integration.Vsix/CFamily/VcxProject/CmdBuilder.cs
```diff
@@ -0,0 +1,394 @@
...
+
+using System;
+using System.Text;
+using System.Text.RegularExpressions;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.VCProjectEngine;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject
+{
+    internal class CmdBuilder
```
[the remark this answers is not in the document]
abbas-sabra-sonarsource: No. This file is loading options from the UI and is specific to VS. the idea was that CFamily java side change on every release. It is really hard to synchronize especially after a lot of refactoring on the java side that has no impact on SonarLint.  We had a lot of bugs because it is hard to synchronize. Now we have only have this code that is specific to the CFamily compilation options. It is localized and isolated. The idea is to be able to update the CFamily analyzer without any changes on the SonarLint side. Like the update to 6.30, 6.31..
Later, we can even drop the entire file if we find a way to load All option UI (as mentioned in #2661)

Of course `CmdBuilder.cs` need to be maintained to handle two things: discovering bugs and VS adding new compiler options in their UI(which doesn't happen that often). I'm happy to maintain `CmdBuilder.cs` which should be low effort. It should be the only place that contains CFamily specific code that supposes to evolve with new.

src/Integration.Vsix/CFamily/MessageHandler.cs
abbas-sabra-sonarsource: It can be after the update but before it wasn't since most of the internal messages has invalid loc (line number 0..) which is ignored.
I added this here because it is the right behavior and it impacted the integration test after fixing the bug about Z3 (#2875) in the "Use the subprocess SonarLint mode" branch.

By the way, you are reviewing the #2661 code in the update of analyzer PR. even though the dependents on each others might be better to separate the review.

rita-sonarsource: Thanks for pointing that out, this branch contained everything, hence I missed the other PR. 
I've changed the base of this branch to point to the other branch, and it makes it much simpler now.

PR #2894, opened 2022-02-28
Add SuppressedIssuesFileSynchronizer

### LINKED ISSUE #2889: Trigger suppressions file update whenever suppressions are fetched
### Description

Whenever suppressions are changed, call the service that is responsible for saving the suppressions to disk.

Relevant classes:
- `ISuppressedIssuesMonitor.SuppressionsUpdateRequested`
- `ISonarQubeIssuesProvider`

rita-sonarsource:
Fixes #2889


src/Integration/Suppression/SuppressedIssuesFileSynchronizer.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// Responsible for listening to <see cref="ISuppressedIssuesMonitor.SuppressionsUpdateRequested"/> and calling
+    /// <see cref="ISuppressedIssuesFileStorage.Update"/> with the new suppressions.
+    /// </summary>
+    public interface ISuppressedIssuesFileSynchronizer : IDisposable
+    {
+    }
+
+    [Export(typeof(ISuppressedIssuesFileStorage))]
+    // todo: temp, will be replaced with real implementation
+    internal class DummySuppressedIssuesFileStorage : ISuppressedIssuesFileStorage
```
rita-sonarsource, line 41: Dummy implementation so the MEF import won't fail, this would be replaced by @ugras-ergun-sonarsource 's class

src/Integration/Suppression/SuppressedIssuesFileSynchronizer.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        public IEnumerable<SonarQubeIssue> Get(string sonarProjectKey)
+        {
+            throw new NotImplementedException();
+        }
+    }
+
+    [Export(typeof(ISuppressedIssuesFileSynchronizer))]
+    internal class SuppressedIssuesFileSynchronizer : ISuppressedIssuesFileSynchronizer
```
rita-sonarsource, line 54: will make sealed to fix the SC issue after the PR comments

src/Integration/Suppression/SuppressedIssuesFileSynchronizer.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    /// <summary>
+    /// Responsible for listening to <see cref="ISuppressedIssuesMonitor.SuppressionsUpdateRequested"/> and calling
+    /// <see cref="ISuppressedIssuesFileStorage.Update"/> with the new suppressions.
+    /// </summary>
+    public interface ISuppressedIssuesFileSynchronizer : IDisposable
+    {
+    }
+
+    [Export(typeof(ISuppressedIssuesFileStorage))]
```
duncanp-sonar, line 39: It's not part of this change, but weren't we going to call this something more general like `IRoslynSettingsFileStorage`?

src/Integration/Suppression/SuppressedIssuesFileSynchronizer.cs
```diff
@@ -0,0 +1,91 @@
...
+    [Export(typeof(ISuppressedIssuesFileSynchronizer))]
+    internal class SuppressedIssuesFileSynchronizer : ISuppressedIssuesFileSynchronizer
+    {
+        private readonly ISuppressedIssuesMonitor suppressedIssuesMonitor;
+        private readonly ISonarQubeIssuesProvider suppressedIssuesProvider;
+        private readonly ISuppressedIssuesFileStorage suppressedIssuesFileStorage;
+        private readonly IActiveSolutionBoundTracker activeSolutionBoundTracker;
+
+        [ImportingConstructor]
+        public SuppressedIssuesFileSynchronizer(ISuppressedIssuesMonitor suppressedIssuesMonitor,
```
duncanp-sonar, line 62: Does this type need to be in `Integration`? It's specific to Roslyn suppressions, and all of the types it imports seem to be in `Core`. Could it be in the new suppressions assembly instead?

rita-sonarsource: The new assembly doesn't reference MEF so it can't be there. Additionally it's only used on the VS side, and that's where all the implementations of these interfaces are.

src/Integration.UnitTests/Suppression/SuppressedIssuesFileSynchronizerTests.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        private SonarQubeIssue CreateSonarQubeIssue()
+        {
+            return new SonarQubeIssue(Guid.NewGuid().ToString(), Guid.NewGuid().ToString(), Guid.NewGuid().ToString(),
+                Guid.NewGuid().ToString(), Guid.NewGuid().ToString(), Guid.NewGuid().ToString(), false,
+                SonarQubeIssueSeverity.Blocker, new DateTimeOffset(), new DateTimeOffset(), null, null);
+        }
+
+        private static BindingConfiguration CreateBindingConfiguration(string projectKey)
```
duncanp-sonar, line 133: Might be better to call this `CreateBoundConfiguration`/`CreateConnectedConfiguration` since it's hard-coded to return use connected mode.

duncanp-sonar (commented):
Looks good, just a question about whether this can in the suppressions assembly instead.

duncanp-sonar (approved):

PR #2897, opened 2022-03-01
Write suppression files to a temp location

### LINKED ISSUE #2890: Write suppression files to a temp location
[ ] Write suppressions file to temp location
[ ] Name convention: `SonarLint.RoslynSettings.<sonar project key>.json`
[ ] Perf: Write only C#/VB suppressions?

ugras-ergun-sonarsource:
fixes #2890


src/Integration.UnitTests/Suppression/SuppressedIssuesFileStorageTests.cs
```diff
@@ -0,0 +1,372 @@
...
+
+        private static Mock<IFileSystem> CreateFileSystem(Mock<IDirectory> directory, Mock<IFile> file)
+        {
+            var fileSystem = new Mock<IFileSystem>();
+            fileSystem.SetupGet(fs => fs.File).Returns(file.Object);
+            fileSystem.SetupGet(fs => fs.Directory).Returns(directory.Object);
+            return fileSystem;
+        }
+
+        private static SonarQubeIssue CreateIssue(string issueKey)
```
ugras-ergun-sonarsource, line 346: I used only one field to differentiate between records. Adding more fields would not serve any purpose other than testing the Serializer.

src/Integration/Suppression/SuppressedIssuesFileStorage.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        internal SuppressedIssuesFileStorage(ILogger logger, IFileSystem fileSystem)
+        {
+            this.fileSystem = fileSystem;
+            this.logger = logger;
+
+            fileDirectory = Path.Combine(Path.GetTempPath(), "SLVS", "Roslyn");
+        }
+
+        public IEnumerable<SonarQubeIssue> Get(string sonarProjectKey)
```
duncanp-sonar, line 51: The information in the file needs to be passed to the diagnostic suppressor, so doesn't this class need to be in the new assembly?

src/Integration/Suppression/SuppressedIssuesFileStorage.cs
```diff
@@ -0,0 +1,113 @@
...
+                    return Enumerable.Empty<SonarQubeIssue>();
+                }
+
+                var filePath = GetFilePath(sonarProjectKey);
+                var fileContent = fileSystem.File.ReadAllText(filePath);
+                return JsonConvert.DeserializeObject<IEnumerable<SonarQubeIssue>>(fileContent);
+            }
+            catch (Exception ex)
+            {
+                logger.WriteLine($"SuppressedIssuesFileStorage Get Error: {ex.Message}");
```
duncanp-sonar, line 66: In general all logged messages except "debug"-level messages should be localised i.e. the text should be in a resource file.

src/Integration/Suppression/SuppressedIssuesFileStorage.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+            catch (Exception ex)
+            {
+                logger.WriteLine($"SuppressedIssuesFileStorage Update Error: {ex.Message}");
+            }
+        }
+
+        private bool ValidateFileName(string sonarProjectKey)
+        {
+            if (string.IsNullOrWhiteSpace(sonarProjectKey))
```
duncanp-sonar, line 94: When would it be valid to call this method with a null/empty string? Wouldn't it be better to throw/assert here?

src/Integration/Suppression/SuppressedIssuesFileStorage.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                logger.WriteLine("SuppressedIssuesFileStorage: File name should not be empty");
+                return false;
+            }
+            if (sonarProjectKey.IndexOfAny(Path.GetInvalidFileNameChars()) >= 0)
```
duncanp-sonar, line 99: This is an interesting case. Could we support keys with invalid characters by replacing the invalid characters with valid ones?

src/Integration.UnitTests/Suppression/SuppressedIssuesFileStorageTests.cs
```diff
@@ -0,0 +1,372 @@
...
+    public class SuppressedIssuesFileStorageTests
+    {
+       
+        [TestMethod]
+        public void Update_Should_WriteSingleIssueToFilePath()
+        {
+            SonarQubeIssue issue = CreateIssue("issueKey");
+            var issues = new List<SonarQubeIssue> { issue };
+
+            var logger = new Mock<ILogger>();
```
duncanp-sonar, line 44: FYI we have a custom `TestLogger` class in the `TestInfrastructure` project that has specific assertion methods for checking for logged output. It's easier to use to use that using mocking `ILogger`.

src/Integration.UnitTests/Suppression/SuppressedIssuesFileStorageTests.cs
```diff
@@ -0,0 +1,372 @@
...
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
+{
+    [TestClass]
+    public class SuppressedIssuesFileStorageTests
+    {
+       
+        [TestMethod]
+        public void Update_Should_WriteSingleIssueToFilePath()
```
duncanp-sonar, line 39: Is there much value in checking reading/writing a single issue as well as multiple issues?

Also, there is a lot of duplication between write one/many tests. If you think it's worth having both, please factor out the duplication into a helper method.

src/Integration.UnitTests/Suppression/SuppressedIssuesFileStorageTests.cs
```diff
@@ -0,0 +1,372 @@
...
+            file.Verify(f => f.ReadAllText(GetFilePath("projectKey")), Times.Once);
+            logger.Verify(l => l.WriteLine(It.IsAny<string>(), It.IsAny<object[]>()), Times.Never);
+
+            Assert.AreEqual(2, issuesGotten?.Count);
+            Assert.AreEqual(issue1.IssueKey, issuesGotten[0].IssueKey);
+            Assert.AreEqual(issue2.IssueKey, issuesGotten[1].IssueKey);
+        }
+
+        [TestMethod]
+        public void Update_Should_LogWhenProjectKeyIsNull()
```
duncanp-sonar, line 137: There's a lot of duplication between this test and the "project key is empty" test.
It would be better as a data-driven test (although I think we should be throwing if the project key is invalid).

duncanp-sonar (changes requested):
See comments inline

src/Integration/Suppression/SuppressedIssuesFileStorage.cs
ugras-ergun-sonarsource: All the other interface implementations for same namespace was here so I've added this also here to be consistent.

src/Integration/Suppression/SuppressedIssuesFileStorage.cs
ugras-ergun-sonarsource: It's because we have talked about not showing the exceptions to users.

src/Integration/Suppression/SuppressedIssuesFileStorage.cs
ugras-ergun-sonarsource: We can replace them but could it cause key duplication?

src/Integration/Suppression/SuppressedIssuesFileStorage.cs
duncanp-sonar: How is the diagnostic suppressor going to use this type?

src/Integration/Suppression/SuppressedIssuesFileStorage.cs
duncanp-sonar: If we call this method with a null/empty, then it means there is a problem with the our calling code.
We shouldn't silently handle it in that case - we want the failure to be obvious to us during development so we can fix it.

We need to handle runtime errors so the process doesn't crash e.g. some find of I/O or locking failure.
However, we shouldn't be coding defensively to hide logic errors in our code.

src/Integration/Suppression/SuppressedIssuesFileStorage.cs
duncanp-sonar: It could but it's pretty unlikely to happen in practice, and it is still better than failing to suppress if the key contains a single invalid character.

There are other places we use the project key in a file name. I didn't think we handled invalid characters properly, but it looks like we do - there is a `PathHelper.EscapeFileName` method in `Core`.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/Roslyn.Suppressions.UnitTests.csproj
```diff
@@ -8,6 +8,8 @@
     <PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
     <PackageReference Include="Moq" Version="4.14.5" />
     <PackageReference Include="FluentAssertions" Version="5.9.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
```
rita-sonarsource, line 11: This was already added above (I'm assuming it's auto merge)

src/Roslyn.Suppressions/Roslyn.Suppressions/Roslyn.Suppressions.csproj
```diff
@@ -17,7 +17,32 @@
...
     <AssemblyName>SonarLint.VisualStudio.Roslyn.Suppressions</AssemblyName>
+    <RequiresNewtonsoft>true</RequiresNewtonsoft>
   </PropertyGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="System.IO.Abstractions" Version="9.0.4" />
+  </ItemGroup>
+  
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\sonarqube-webclient\SonarQube.Client\SonarQube.Client.csproj" />
```
rita-sonarsource, line 28: Auto-merge duplication with the code below (there was already a reference to Core project). Also I think it's worth putting the package dependencies next to each other.

src/Roslyn.Suppressions/Roslyn.Suppressions/Resources/Strings.resx
```diff
@@ -0,0 +1,126 @@
...
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="SuppressedIssuesFileStorageGetError" xml:space="preserve">
+    <value>SuppressedIssuesFileStorage Get Error: {0}</value>
+  </data>
+  <data name="SuppressedIssuesFileStorageUpdateError" xml:space="preserve">
+    <value>SuppressedIssuesFileStorage Update Error: {0}</value>
```
rita-sonarsource, line 124: If we are using Logger.Writeline, on the VS side this would appear to the user in the output window. I think we should have a more descriptive message for the user, or alternatively maybe change it to logger.LogDebug.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressedIssuesFileStorageTests.cs
```diff
@@ -0,0 +1,258 @@
...
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
+{
+    [TestClass]
+    public class SuppressedIssuesFileStorageTests
+    {
+
+        [TestMethod]
+        public void Update_Should_WriteIssuesToFilePath()
```
rita-sonarsource, line 39: We haven't discussed our unit testing naming convention. We're following `MethodName_StateUnderTest_ExpectedBehavior`, so this would be  something like `Update_HasIssues_IssuesWrittenToFile`.

We can decide together as a team on a [different](https://dzone.com/articles/7-popular-unit-test-naming) one.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressedIssuesFileStorageTests.cs
```diff
@@ -0,0 +1,258 @@
...
+        public void Update_Should_ThrowWhenProjectKeyIsNull(string projectKey)
+        {
+            SonarQubeIssue issue = CreateIssue("issueKey");
+            var issues = new List<SonarQubeIssue> { issue };
+
+            var logger = new TestLogger();
+
+            var directory = new Mock<IDirectory>();
+            var file = new Mock<IFile>();
+            Mock<IFileSystem> fileSystem = CreateFileSystem(directory, file);
```
rita-sonarsource, line 100: Since you're not expecting IDirectory and IFile to be called, we have a pattern of having nullable arguments. 
So you could do something like `var fileSystem = CreateFileSystem();` and the method implementation would receive optional params and set them to Mock.Of<> if they're null. 
This is just to make the tests a bit shorter and easier to read as it removes mocks that won't be tested.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressedIssuesFileStorageTests.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            var fileStorage = new SuppressedIssuesFileStorage(logger, fileSystem.Object);
```
rita-sonarsource, line 102: Similar to the comment above, it's worth have a `CreateTestSubject` method that receives optional parameters and assigns dummy values to things that weren't passed. I.e. the logger is irrelevant in this test scenario.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressedIssuesFileStorageTests.cs
```diff
@@ -0,0 +1,258 @@
...
+            var ex = Assert.ThrowsException<ArgumentException>(() => fileStorage.Update(projectKey, issues));
```
rita-sonarsource, line 103: Another minor team convention: we use FluentAssertions so it would be written instead 
```
Action act = () => fileStorage.Update(projectKey, issues);
act.Should().ThrowExactly<ArgumentException>().And.ParamName.Should().Be....
```

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressedIssuesFileStorageTests.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            Assert.AreEqual("Argument must have at least one non white space character\r\nParameter name: sonarProjectKey", ex.Message);
+        }
+
+        [TestMethod]
+        public void Update_Should_LogWhenProjectKeyHasInvalidChars()
+        {
```
rita-sonarsource, line 110: This test's name doesn't match what's being tested.

rita-sonarsource (changes requested):
@ugras-ergun-sonarsource just some comments, it's mostly around our team conventions.

src/Roslyn.Suppressions/Roslyn.Suppressions/Resources/Strings.resx
duncanp-sonar: (I added a comment about the suggested formatting of the message in the previous version of this PR).

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/SuppressedIssuesFileStorage.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        public void Update(string sonarProjectKey, IEnumerable<SonarQubeIssue> allSuppressedIssues)
+        {
+            ValidateSonarProjectKey(sonarProjectKey);
+            try
+            {
+                var escapedName = PathHelper.EscapeFileName(sonarProjectKey);
+                var filePath = GetFilePath(escapedName);
+                var fileContent = JsonConvert.SerializeObject(allSuppressedIssues);
+                fileSystem.Directory.CreateDirectory(fileDirectory);
```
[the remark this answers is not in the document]
ugras-ergun-sonarsource: Isn't the method will be called only once?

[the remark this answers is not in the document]
rita-sonarsource: It should be called each time the file changes on the VS side, so either every 10 min or whenever the user triggers an update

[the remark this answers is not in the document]
duncanp-sonar: Also, it's a parameterised public method. The implementation of the class shouldn't be making that kind of assumption about how it is going to be used.

src/Roslyn.Suppressions/Roslyn.Suppressions/Resources/Strings.resx
```diff
@@ -0,0 +1,129 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="SuppressedIssuesFileStorageFileNotFound" xml:space="preserve">
+    <value>Settings File was not found</value>
```
duncanp-sonar, line 121: Missing `[Roslyn Suppressions]` prefix. It would help if the message said where it was expecting to find the file.

ugras-ergun-sonarsource: This will not be shown alone but will be used instead of error message in "SuppressedIssuesFileStorageGetError"

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressedIssuesFileStorageTests.cs
```diff
@@ -0,0 +1,298 @@
...
+
+            var file = new Mock<IFile>();
+            file.Setup(f => f.WriteAllText(It.IsAny<string>(), It.IsAny<string>())).Throws(new Exception("Test Exception"));
+            Mock<IFileSystem> fileSystem = CreateFileSystem(file);
+
+            var fileStorage = CreateTestSubject(logger, fileSystem.Object);
+
+            fileStorage.Update("projectKey", issues);
+
+            logger.AssertOutputStrings("[Roslyn Suppressions] Error writing settings for project projectKey. Issues suppressed on the server may not be suppressed in the IDE. Error: Test Exception");
```
duncanp-sonar, line 167: FYI if a logged message has parameters, we don't tend to check the fully-constructed strings - we would normally just check the expected parameters are present using one of the `TestLogger` assertion methods like `AssertPartialOutputStrings`.
There's no particular problem with checking for the full string, it's just not something we tend to do.

duncanp-sonar (approved):
Just a couple of minor comments about strings.

rita-sonarsource (approved):

PR #2899, opened 2022-03-03
Force update suppressions file on package load

rita-sonarsource:
Fixes [#2900](https://github.com/SonarSource/sonarlint-visualstudio/issues/2900)


src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -136,6 +136,13 @@ public IEnumerable<SonarQubeIssue> GetSuppressedIssues(string projectGuid, strin
 
         public IEnumerable<SonarQubeIssue> GetAllSuppressedIssues()
         {
+            // Block the call while the cache is being built.
+            // If the task has already completed then this will return immediately
+            // (e.g. on subsequent calls)
+            // If we time out waiting for the initial fetch then we won't suppress any issues.
+            // We'll try to fetch the issues again when the timer elapses.
+            this.initialFetch?.Wait(MillisecondsToWaitForInitialFetch);
```
rita-sonarsource, line 144: Previously this was always called AFTER the list was already fetched. Now it's possible that it's called while the first fetch is being executed, in which case we need to block.
The other method has a lot of tests that check that the call is blocked but it wrong to just copy-paste them. I think this class could use some refactoring but that's obviously out of scope for now.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/SuppressedIssuesFileSynchronizer.cs
```diff
@@ -53,34 +57,59 @@ public IEnumerable<SonarQubeIssue> Get(string sonarProjectKey)
...
         }
 
         private void SuppressedIssuesMonitor_SuppressionsUpdateRequested(object sender, EventArgs e)
         {
-            var sonarProjectKey = activeSolutionBoundTracker.CurrentConfiguration.Project?.ProjectKey;
+            UpdateFileStorage();
+        }
 
-            if (!string.IsNullOrEmpty(sonarProjectKey))
+        public void UpdateFileStorage()
```
rita-sonarsource, line 99: Initially this code was being executed in the CTOR, however that caused the MEF export test to fail. ThreadHandling is not a MEF dependency so I could not inject a dummy mock to replace it and it was failing in the real implementation. So I moved the logic to a method and I'm explicitly calling this method in SonarLintIntegrationPackage.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/SuppressedIssuesFileSynchronizer.cs
```diff
@@ -53,34 +57,59 @@ public IEnumerable<SonarQubeIssue> Get(string sonarProjectKey)
...
-            }
+                var sonarProjectKey = activeSolutionBoundTracker.CurrentConfiguration.Project?.ProjectKey;
+
+                if (!string.IsNullOrEmpty(sonarProjectKey))
+                {
+                    var allSuppressedIssues = suppressedIssuesProvider.GetAllSuppressedIssues();
+                    suppressedIssuesFileStorage.Update(sonarProjectKey, allSuppressedIssues);
+                }
+
+                return Task.FromResult(true);
```
rita-sonarsource, line 111: A little bit ugly but the current Run method expects a Task<T> and doesn't support void. I wanted to avoid further changes so I'm just returning a bool.

src/Roslyn.Suppressions/Roslyn.Suppressions/Roslyn.Suppressions.csproj
```diff
@@ -24,6 +24,17 @@
   
   <ItemGroup>
     <ProjectReference Include="..\..\Core\Core.csproj" />
+    <ProjectReference Include="..\..\Infrastructure.VS\Infrastructure.VS.csproj" />
+  </ItemGroup>
+  
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
```
rita-sonarsource, line 31: For some reason this was added automatically by VS, perhaps it has to be on the csproj directly and not in the targets file?

src/Roslyn.Suppressions/Roslyn.Suppressions/SuppressorCodeGen.targets
```diff
@@ -13,10 +13,6 @@
                reference a couple of Roslyn NuGet packages here to stop NuGet complaining about version conflicts. 
     -->
     <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="3.3.1" PrivateAssets="all" GeneratePathProperty="true" />
-    <!-- Note: System.Collections.Immutable is a transitive dependency of MS.CA.Common, so strictly speaking we don't need to
-         reference it directly. However, we have an explicit reference so we use the property path in the .tt file.
-         When updating MS.CA.Common, make sure to update the Sytem.Collections.Immutable to the version it uses. -->
-    <PackageReference Include="System.Collections.Immutable" Version="1.5" PrivateAssets="all" GeneratePathProperty="true" />
```
rita-sonarsource, line 19: As we discussed offline, removed due to package conflict with Infrastructure.VS

src/Integration.Vsix/SonarLintIntegrationPackage.cs
```diff
@@ -99,6 +99,7 @@ private async System.Threading.Tasks.Task InitOnUIThreadAsync()
                     serviceProvider.GetMefService<IProjectToLanguageMapper>());
 
                 this.suppressedIssuesFileSynchronizer = await this.GetMefServiceAsync<ISuppressedIssuesFileSynchronizer>();
+                suppressedIssuesFileSynchronizer.UpdateFileStorage();
```
rita-sonarsource, line 102: See comment in the SuppressedIssuesFileSynchronizer file

src/Integration/Suppression/SonarQubeIssuesProvider.cs
duncanp-sonar: It looks like it would be really easy to turn this into an `async` method...
Still, now isn't the time to do it. Let's keep it in mind as we add new services.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/InProcess/SuppressedIssuesFileSynchronizerTests.cs
```diff
@@ -124,6 +126,43 @@ public void OnSuppressionsUpdateRequested_ConnectedMode_StorageUpdated(bool hasI
...
+        public async Task UpdateFileStorage_FileStorageIsUpdatedOnBackgroundThread()
+        {
+            Func<Task<bool>> fileUpdateTask = null;
+            var threadHandling = new Mock<IThreadHandling>();
+            threadHandling
+                .Setup(x => x.Run(It.IsAny<Func<Task<bool>>>()))
+                .Callback((Func<Task<bool>> task) => fileUpdateTask = task);
+
+            var configuration = CreateConnectedConfiguration("some project key");
+            var activeSolutionBoundTracker = new Mock<IActiveSolutionBoundTracker>();
```
duncanp-sonar, line 139: Minor: all of the tests would be a little simpler if there were helper methods to create the `activeSolutionBoundTracker` and `suppressedIssuesProvider`.

src/Roslyn.Suppressions/Roslyn.Suppressions/Roslyn.Suppressions.csproj
duncanp-sonar: It may just be VS being stupid. If we remove it, VS will just keep adding it back.
Probably the best we can do is make it clearer why they are there e.g.

```
  <ItemGroup Label="Automatically added by VS">
    <!-- Note: these settings relate to files in the SuppressorCodeGen.targets -->
    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
    <Compile Update="SupportedSuppressionBuilder.g.cs">
      <DesignTime>True</DesignTime>
    </Compile>
  </ItemGroup>
```

duncanp-sonar (approved):
Just a couple of minor comments about tests and the targets.

ugras-ergun-sonarsource (approved):

PR #2903, opened 2022-03-04
Add helper class for RoslynSettings file location

rita-sonarsource:


src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressedIssuesFileStorageTests.cs
```diff
@@ -20,17 +20,18 @@
...
 using Moq;
 using Newtonsoft.Json;
-using SonarLint.VisualStudio.Integration.Roslyn.Suppression.SettingsFile;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.Roslyn.Suppressions.SettingsFile;
 using SonarQube.Client.Models;
 
-namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
+namespace SonarLint.VisualStudio.Roslyn.Suppressions.UnitTests
```
rita-sonarsource, line 34: Was in the wrong namespace

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/SuppressedIssuesFileStorage.cs
```diff
@@ -20,20 +20,19 @@
...
 using System.Linq;
 using Newtonsoft.Json;
 using SonarLint.VisualStudio.Core.Helpers;
 using SonarLint.VisualStudio.Core.Suppressions;
+using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.Roslyn.Suppressions.Resources;
 using SonarQube.Client.Models;
 
-namespace SonarLint.VisualStudio.Integration.Roslyn.Suppression.SettingsFile
+namespace SonarLint.VisualStudio.Roslyn.Suppressions.SettingsFile
```
rita-sonarsource, line 32: Was in the wrong namespace

src/Roslyn.Suppressions/Roslyn.Suppressions/RoslynSettingsFileInfo.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System.IO;
+
+namespace SonarLint.VisualStudio.Roslyn.Suppressions
+{
+    internal static class RoslynSettingsFileInfo
+    {
+        public static readonly string Directory = Path.Combine(Path.GetTempPath(), "SLVS", "Roslyn");
+
+        public static string GetFilePathFromEscapedProjectKey(string sonarProjectKey)
```
ugras-ergun-sonarsource, line 29: maybe name the variable as escaped project key so it would be obvious for later.

ugras-ergun-sonarsource (approved):
Only one comment about variable name

src/Roslyn.Suppressions/Roslyn.Suppressions/RoslynSettingsFileInfo.cs
duncanp-sonar: Why not move the escaping into this class too? That way the caller(s) don't have to know how to escape it.

duncanp-sonar (approved):
Only one comment about the escaping.

src/Roslyn.Suppressions/Roslyn.Suppressions/RoslynSettingsFileInfo.cs
rita-sonarsource: We can't reverse it back so one method would be `GetFilePathFromProjectKey` and the other would be `GetEscapedProjectKeyFromFilePath`, which is a bit weird.

duncanp-sonar: It is a bit odd, but it doesn't seem right that the caller needs to know how to do part of the name-mangling calculation.

Yep, perhaps just changing the naming would be enough e.g. `projectMoniker`, `projectToken`, `projectId`, `settingsKey`

```
// Returns the full file path for the given project key
string GetSettingsFilePath(string sonarProjectKey)


// Returns an identifier for the project settings to which the settings file relates.
// <remarks>The identifier is *not* the actual project key since we can't recover it accurately from the file name </remarks> 
string GetSettingsKey(string settingsFilePath)

```

duncanp-sonar: Nice - the updated version looks good.

Issue #2901, opened 2022-03-04
Add FileSystemWatcher for the RoslynSettings file


rita-sonarsource:
### Description

Listen to file changes and refresh the cache when the file changes.

PR #2905, opened 2022-03-04
Add SuppressedIssuesFileWatcher

rita-sonarsource:
Fixes [#2901](https://github.com/SonarSource/sonarlint-visualstudio/issues/2901)


src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/SuppressedIssuesFileWatcher.cs
```diff
@@ -0,0 +1,71 @@
...
+        public SuppressedIssuesFileWatcher(ISuppressedIssuesCache suppressedIssuesCache)
+            : this(suppressedIssuesCache, new FileSystem())
+        {
+        }
+
+        internal SuppressedIssuesFileWatcher(ISuppressedIssuesCache suppressedIssuesCache, IFileSystem fileSystem)
+        {
+            this.suppressedIssuesCache = suppressedIssuesCache;
+
+            fileSystem.Directory.CreateDirectory(RoslynSettingsFileInfo.Directory);
```
rita-sonarsource, line 46: From manual testing, `FileSystemWatcher.FromPath` will throw argument exception if the directory does not exist. In practice it _shouldn't_ happen, but maybe it's better to safeguard just in case.

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/SuppressedIssuesFileWatcher.cs
```diff
@@ -0,0 +1,71 @@
...
+            fileSystemWatcher = fileSystem.FileSystemWatcher.FromPath(RoslynSettingsFileInfo.Directory);
+
+            fileSystemWatcher.Created += InvalidateCache;
+            fileSystemWatcher.Changed += InvalidateCache;
+            fileSystemWatcher.Deleted += InvalidateCache;
```
rita-sonarsource, line 51: Technically there are more events, notably `Renamed`, but they shouldn't be raised since we're only calling WriteAllText on the VS side.

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/RoslynSettingsFileInfo.cs
```diff
@@ -21,7 +21,7 @@
 using System.IO;
 using SonarLint.VisualStudio.Core.Helpers;
 
-namespace SonarLint.VisualStudio.Roslyn.Suppressions
+namespace SonarLint.VisualStudio.Roslyn.Suppressions.SettingsFile
```
rita-sonarsource, line 24: Moved into the SettingsFile directory, along with the file watcher.

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/SuppressedIssuesFileWatcher.cs
```diff
@@ -0,0 +1,71 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.IO;
+using System.IO.Abstractions;
+
+namespace SonarLint.VisualStudio.Roslyn.Suppressions.SettingsFile
+{
+    internal interface ISuppressedIssuesCache
```
rita-sonarsource, line 27: Just a dummy interface for now

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/SuppressedIssuesFileWatcher.cs
duncanp-sonar: Yep, better to be defensive.
There are a couple of scenarios in which the directory might not exist:
* first use; the external process might have started running before SLVS has had a chance to create the directory for the first time;
* Windows clears up the temp folder

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/SuppressedIssuesFileWatcher.cs
```diff
@@ -0,0 +1,79 @@
...
+            fileSystemWatcher.Changed += InvalidateCache;
+            fileSystemWatcher.Deleted += InvalidateCache;
+
+            fileSystemWatcher.EnableRaisingEvents = true;
+        }
+
+        private void InvalidateCache(object sender, FileSystemEventArgs e)
+        {
+            var fileName = e.Name;
+            var settingsKey = RoslynSettingsFileInfo.GetSettingsKey(fileName);
```
duncanp-sonar, line 67: Again, it might be better to be defensive and ignore cases where the we can't retrieve the settings key e.g. in case something else like a virus checker ever writes a temporary file in the directory.

duncanp-sonar (commented):
Couple of minor comments about being defensive.

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/SuppressedIssuesFileWatcher.cs
```diff
@@ -0,0 +1,79 @@
...
+            suppressedIssuesCache.Invalidate(settingsKey);
+        }
+
+        public void Dispose()
+        {
+            fileSystemWatcher.Created -= InvalidateCache;
+            fileSystemWatcher.Changed -= InvalidateCache;
+            fileSystemWatcher.Deleted -= InvalidateCache;
+            fileSystemWatcher?.Dispose();
```
ugras-ergun-sonarsource, line 76: Null check here seems redundant. If the fileSystemWatcher is null it will fail on previous lines. Maybe it would be better to have a null check at the beginning of the method.

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/SuppressedIssuesFileWatcher.cs
rita-sonarsource: Hm, that's a good question. If we try-catch the exception, it will leave VS' external process in an inconsistent state where our stuff are partially initialized (i.e. suppressions code is there but the file watcher was never created, so the cache would never update). Perhaps it would be better to fail, throw, and let the caller handle it, so that next time we would try to re-create it.

PR #2906, opened 2022-03-07
Add cache for roslyn settings #2902 

### LINKED ISSUE #2902: Add Singleton cache for roslyn settings
### Description

Add a singleton cache for Roslyn settings file so that suppressor would not go to the disk every time.

ugras-ergun-sonarsource:
fixes #2902


src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/Settings.Cache/SettingsCacheTest.cs
```diff
@@ -0,0 +1,123 @@
+﻿using System.Collections;
```
duncanp-sonar, line 1: Missing copyright headers

src/Roslyn.Suppressions/Roslyn.Suppressions/Settings.Cache/ISettingsCache.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.Roslyn.Suppressions.Settings.Cache
+{
+    public interface ISettingsCache
+    {
+        IEnumerable<SonarQubeIssue> GetSettings(string projectKey);
```
duncanp-sonar, line 12: We're calling the key `settingsKey` rather than `projectKey` since it isn't really the project key.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressedIssuesFileStorageTests.cs
```diff
@@ -40,8 +40,8 @@ public class SuppressedIssuesFileStorageTests
         [TestMethod]
         public void Update_HasIssues_IssuesWrittenToFile()
         {
-            SonarQubeIssue issue1 = CreateIssue("issueKey1");
-            SonarQubeIssue issue2 = CreateIssue("issueKey2");
+            SonarQubeIssue issue1 = TestHelper.CreateIssue("issueKey1");
```
duncanp-sonar, line 43: If you make the helper class static you can add a `static using` so there is no need to include the class name. See `using static SonarQube.Client.Tests.Infra.MocksHelper;` for an example.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/TestHelper.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.Roslyn.Suppressions.UnitTests
+{
+    internal class TestHelper
```
duncanp-sonar, line 10: Should be static.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/TestHelper.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        internal static SonarQubeIssue CreateIssue(string issueKey)
```
duncanp-sonar, line 12: By convention, we use `public` for methods in `internal` classes that we expect to be called by other classes i.e. we use `internal class - public method` rather than `internal class - internal method`.

It doesn't make any difference to the compiler, but gives a clearer indication to people reading the code that the method is part of its "public" interface i.e. usable by anyone.

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/SuppressedIssuesFileStorage.cs
```diff
@@ -51,6 +53,7 @@ public IEnumerable<SonarQubeIssue> Get(string sonarProjectKey)
         {
             ValidateSonarProjectKey(sonarProjectKey);
 
+            ValidateSonarProjectKey(sonarProjectKey);
```
duncanp-sonar, line 56: Merge error? Duplicate of line 54.

src/Roslyn.Suppressions/Roslyn.Suppressions/Settings.Cache/SettingsCache.cs
```diff
@@ -0,0 +1,43 @@
...
+        private readonly ISuppressedIssuesFileStorage fileStorage;
+        private readonly ConcurrentDictionary<string, IEnumerable<SonarQubeIssue>> settingsCollection;
+
+
+        public SettingsCache(ILogger logger) : this(new SuppressedIssuesFileStorage(logger), new SettingsSingleton())
+        {
+
+        }
+
+        internal SettingsCache(ISuppressedIssuesFileStorage fileStorage, ISettingsSingleton singleton)
```
duncanp-sonar, line 21: I don't understand what the purpose of the `ISettingsSingleton` interface is. We aren't going to be creating multiple instances of `SettingsCache`.

If you need to get hold of the internal dictionary for testing purposes, you could change the `internal` constructor so the tests can directly pass in the dictionary to use:

`public SettingsCache(ILogger logger);
internal SettingsCache(ILogger logger, ConcurrentDictionary<...> settingsMap)`

Or add an internal accessor that exposes the dictionary to the tests e.g. 
`internal /* for testing */ IReadOnlyDictionary<...> SettingsMap { get { returns _settingsMap; } }`

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/Settings.Cache/SettingsCacheTest.cs
```diff
@@ -0,0 +1,123 @@
...
+            var issues = CreateIssues();
+
+            var cacheObject = CreateCacheObject(issues);
+
+            var fileStorage = new Mock<ISuppressedIssuesFileStorage>();
+
+            var settingsSingleton = CreateSettingsSingleton(cacheObject);
+
+            var testSubject = CreateTestSubject(fileStorage, settingsSingleton);
+            var settings = testSubject.GetSettings("projectKey");
```
duncanp-sonar, line 50: There's a hidden coupling between the `CreateCacheObject` helper method and the tests i.e. that if the helper method returns a pre-populated cache it will contain a single entry with the key `projectKey`. 

One way to make the coupling explicit would be to have explicit helper methods e.g.
```
CreateEmptyCache();
CreatePrePopulatedCache(string key, IEnumerable<SonarQubeIssue> issues);
```

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/Settings.Cache/SettingsCacheTest.cs
```diff
@@ -0,0 +1,123 @@
...
+using SonarLint.VisualStudio.Roslyn.Suppressions.Settings.Cache;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.Roslyn.Suppressions.UnitTests.Settings.Cache
+{
+    [TestClass]
+    public class SettingsCacheTest
+    {
+        [TestMethod]
+        public void GetSettings_SettingNotInCache_SettingsReadFromFile()
```
duncanp-sonar, line 18: As discussed offline, I think there's a missing test:
* cache pre-populated with X
* request Y -> empty list returned.

duncanp-sonar (changes requested):
See comments inline

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/Settings.Cache/SettingsCacheTest.cs
ugras-ergun-sonarsource: After realising CreateEmptyCache() is basicaly one line code It didn't make sense to make it a seperate method.

src/Roslyn.Suppressions/Roslyn.Suppressions/Settings.Cache/SettingsCache.cs
ugras-ergun-sonarsource: As we discussed I have removed the Singleton.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/Settings.Cache/SettingsCacheTest.cs
duncanp-sonar: It depends what you are trying to achieve. It doesn't reduce the amount of code, but it does make the test more declarative and easier to understand anyone reading it.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/Settings.Cache/SettingsCacheTest.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            var testSubject = CreateTestSubject(fileStorage, cacheObject);
+            var settings = testSubject.GetSettings("settingsKey");
+
+            fileStorage.Verify(fs => fs.Get(It.IsAny<string>()), Times.Never);
+            settings.Should().BeSameAs(issues);
+        }
+
+        [TestMethod]
+        public void GetSettings_SettingNotInCacheAndFile_SettingsEmpty()
```
duncanp-sonar, line 75: Minor: "_SettingsNotInCacheOrFile_" ("or" rather than "and")

duncanp-sonar (approved):

Issue #2927, opened 2022-03-15
Remove old Roslyn suppression code


duncanp-sonar:
### Description

The existing Roslyn suppression handling code can be removed, along with the code that handles the SonarAnalyzer NuGet packages being installed in the MSBuild project (see #2925).

TODO
- [ ] delete the existing Roslyn suppression handing code
- [ ] check CFamily suppressions are not affected
- [ ] if possible, remove the reference to the SonarAnalyzer assembly from the SLVS code and MSBuild projects

PR #2928, opened 2022-03-15
Disable old Roslyn analyzer suppression code

duncanp-sonar:
Part of #2927


src/Integration.Vsix/SonarLintIntegrationPackage.cs
```diff
@@ -91,7 +90,6 @@ private async System.Threading.Tasks.Task InitOnUIThreadAsync()
                 var workspace = await this.GetMefServiceAsync<VisualStudioWorkspace>();
 
                 var vsSolution = serviceProvider.GetService<SVsSolution, IVsSolution>();
-                this.sonarAnalyzerManager = new SonarAnalyzerManager(activeSolutionBoundTracker, workspace, vsSolution, logger, sonarQubeIssuesProvider);
```
duncanp-sonar, line 94: This is a quick change to disable the old Roslyn suppression behaviour and make it easier to debug/test the new version.

Completely removing the old version is more complicated: there's quite a lot of old code to remove, and changes to some of the project files (we should be able to get rid out our references to the SonarAnalyzer assembly). Also, I need to make sure I don't break CFamily suppressions.

PR #2929, opened 2022-03-15
Made GetAllSuppressedIssues method async

### LINKED ISSUE #2922: Bug: no suppressed issues after refresh
Suppressed issues are not fetched on manual refresh

### LINKED ISSUE #2931: Make GetAllSuppressedIssues method async
### Description

This method calls asynchronous methods and itself is a synchronous method. To make asynchronous work consistent we should also mark the method as async.

TODO:
* Mark GetAllSuppressedIssues method as async.
* Wait the method wherever it is called.
* Fix the tests

ugras-ergun-sonarsource:
fixes #2931

### DIFF
--- a/src/Integration/Suppression/SonarQubeIssuesProvider.cs
+++ b/src/Integration/Suppression/SonarQubeIssuesProvider.cs
@@ -83,8 +83,7 @@ namespace SonarLint.VisualStudio.Integration.Suppression
             refreshTimer.Elapsed += OnRefreshTimerElapsed;
 
             initialFetchCancellationTokenSource = new CancellationTokenSource();
-            var hede = Task.Factory.StartNew(DoInitialFetchAsync, initialFetchCancellationTokenSource.Token);
-            this.initialFetch = hede.Unwrap();
+            this.initialFetch = Task.Factory.StartNew(DoInitialFetchAsync, initialFetchCancellationTokenSource.Token).Unwrap();
             refreshTimer.Start();
         }
 
--- a/src/Integration/Suppression/SonarQubeIssuesProvider.cs
+++ b/src/Integration/Suppression/SonarQubeIssuesProvider.cs
@@ -83,7 +83,8 @@ namespace SonarLint.VisualStudio.Integration.Suppression
             refreshTimer.Elapsed += OnRefreshTimerElapsed;
 
             initialFetchCancellationTokenSource = new CancellationTokenSource();
-            this.initialFetch = Task.Factory.StartNew(DoInitialFetchAsync, initialFetchCancellationTokenSource.Token);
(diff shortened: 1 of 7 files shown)


src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -83,7 +83,7 @@ public sealed class SonarQubeIssuesProvider : ISonarQubeIssuesProvider
             refreshTimer.Elapsed += OnRefreshTimerElapsed;
 
             initialFetchCancellationTokenSource = new CancellationTokenSource();
-            this.initialFetch = Task.Factory.StartNew(DoInitialFetchAsync, initialFetchCancellationTokenSource.Token);
+            this.initialFetch = Task.Factory.StartNew(DoInitialFetchAsync, initialFetchCancellationTokenSource.Token).Unwrap();
```
duncanp-sonar, line 86: Ah. I suspect that the "Unwrap" here explains the problem you were seeing before. `DoInitialFetchAsync` is returning a Task, so `Task.Factory.StartNew(DoInitialFetchAsync)` is returning a Task that returns a Task. Can we simplify this?

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/SuppressedIssuesFileSynchronizer.cs
```diff
@@ -82,13 +82,13 @@ private void SuppressedIssuesMonitor_SuppressionsUpdateRequested(object sender,
 
         public void UpdateFileStorage()
         {
-            threadHandling.Run(() =>
+            threadHandling.Run(async () =>
```
duncanp-sonar, line 85: This doesn't look right - we shouldn't be using the blocking `Run` here when we're calling something that is `awaitable`.
Or is this part of the "issues are fetched on the UI" problem that is going to be fixed separately?

ugras-ergun-sonarsource: Yes this should be part of the UI thread issue.

duncanp-sonar: It might be better if this method were `async` too. However, that moves the problem to the call in `SuppressedIssuesMonitor_SuppressionsUpdateRequested`, which is a `void` event handler that needs to fire-and-forget the `async UpdateFileStorage` method on another thread.
There are few places in the code where we use the VS `.Forget()` method from `void` methods. Would that approach work?

duncanp-sonar: Ok.

duncanp-sonar (approved):

PR #2930, opened 2022-03-15
Remove old Roslyn suppression code

### LINKED ISSUE #2927: Remove old Roslyn suppression code

duncanp-sonar:
Fixes #2927


src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -94,8 +94,6 @@ Assembly: 'SonarLint.2022, Version=6.0.0.0, Culture=neutral, PublicKeyToken=c5b6
 Relative path: 'SonarLint.2022.dll'
 
 Referenced assemblies:
-- 'Microsoft.CodeAnalysis, Version=1.3.1.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
```
duncanp-sonar, line 97: We no longer have code that directly references the SonarAnalyzer callbacks, which means we don't need to reference the MS CodeAnalysis assemblies either.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -125,11 +122,6 @@
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
 
-  <PropertyGroup>
-    <!-- Suppress NuGet warning about the explicit downgrade to System.Collections.Immutable v1.1.37 -->
-    <NoWarn>NU1605</NoWarn>
-  </PropertyGroup>
-
```
duncanp-sonar, line 132: We don't reference the MS.CodeAnalysis NuGet packages, which means we don't need to reference specific versions of the `System.Collections.Immutable` or `System.Reflection.Metadata` assemblies either.

That in turn means we no longer need to suppress this NuGet warning.

duncanp-sonar:
There are two commits:
* the first just deletes the old code.
* the second updates the project file, package lock files and asmrefs.

I've checked that the contents of the VSIX haven't changed - the set of files we ship is still the same, even though we no longer reference the SonarAnalyzer assembly directly.

I also manually tested that C++ suppressions still work (or at least, work as well as they do currently).

rita-sonarsource (approved):
Woohoo!

duncanp-sonar:
FYI 

**Before (LOC, #Types)**
image

**After (LOC, #Types)**
image


= -561 lines of product code (6.0% reduction)
= -1492 lines of test code (9.3% reduction)

Issue #2932, opened 2022-03-16
Bug: Suppressions are being fetched on the UI thread


rita-sonarsource:
### Description

When `SonarLintIntegrationPackage` loads, it calls `ISuppressedIssuesFileSynchronizer.UpdateFileStorage`. The call is being made on the UI thread and `UpdateFileStorage` is a blocking call.

TODO:
- [x] Make `UpdateFileStorage` a non-blocking call

duncanp-sonar:
The code coverage is low because I can't work out how to test an exception handler - see the comment inline.

PR #2933, opened 2022-03-16
Add RunAsync method to ThreadHandling

rita-sonarsource:
Prep for [#2932](https://github.com/SonarSource/sonarlint-visualstudio/issues/2932)

### DIFF
--- a/src/Infrastructure.VS/ThreadHandling.cs
+++ b/src/Infrastructure.VS/ThreadHandling.cs
@@ -42,11 +42,7 @@ namespace SonarLint.VisualStudio.Infrastructure.VS
 
         public T Run<T>(Func<Task<T>> asyncMethod) => ThreadHelper.JoinableTaskFactory.Run(asyncMethod);
 
-        public Task RunAsync<T>(Func<Task<T>> asyncMethod) => ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
-        {
-            await Task.Yield(); // get off the caller's callstack.
-            await asyncMethod();
-        }).Task;
+        public async Task RunAsync<T>(Func<Task<T>> asyncMethod) => await ThreadHelper.JoinableTaskFactory.RunAsync(asyncMethod);
 
         public IAwaitableWrapper SwitchToBackgroundThread() => new TaskSchedulerAwaitableWrapper(new AwaitExtensions.TaskSchedulerAwaitable(TaskScheduler.Default));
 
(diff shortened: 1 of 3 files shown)


src/Infrastructure.VS/ThreadHandling.cs
```diff
@@ -42,6 +42,12 @@ public class ThreadHandling : IThreadHandling
 
         public T Run<T>(Func<Task<T>> asyncMethod) => ThreadHelper.JoinableTaskFactory.Run(asyncMethod);
 
+        public Task RunAsync<T>(Func<Task<T>> asyncMethod) => ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+        {
+            await Task.Yield(); // get off the caller's callstack.
```
rita-sonarsource, line 47: This is based on the cookbook although I'm not entirely sure we need it.
https://github.com/microsoft/vs-threading/blob/main/doc/cookbook_vs.md

src/Infrastructure.VS/ThreadHandling.cs
```diff
@@ -42,6 +42,12 @@ public class ThreadHandling : IThreadHandling
...
+            await asyncMethod();
+        }).Task;
```
rita-sonarsource, line 49: By default the method returns JoinableTask which is in a VS assembly, but this interface is in Core so I'm returning only the Task.

src/Infrastructure.VS/ThreadHandling.cs
duncanp-sonar: If we include the `Task.Yield()` call then we're not really wrapping the `JoinableTaskFactory` method, we're providing additional functionality. I think it would be better to provide either a simple wrapper that just does a pass-through, or come up with our own method with the functionality we want.
Given that we're not sure that we need the `Yield()` here, it might be better just to have the simple wrapper method.

src/Infrastructure.VS/ThreadHandling.cs
duncanp-sonar: I don't really understand what the `JoinableTask` is doing, but I'm worried that if we consume the inner `Task` directly we might mess up how it works. Could we do something like this instead, so we only wrapping the function?

```
        public async Task RunAsync(Func<Task> asyncMethod) => await ThreadHelper.JoinableTaskFactory.RunAsync(asyncMethod);
```

duncanp-sonar (commented):
See inline

duncanp-sonar (approved):

duncanp-sonar (approved):

PR #2944, opened 2022-03-22
Store only required issue fields required for suppression matching

### LINKED ISSUE #2938: Store only required suppression fields
- storing all fields even though not required
- ?? is it possible to have / in the file path? - normalize. UPDATE: not necessary - already normalized by the webclient assembly.

TODO
- [x] create new wrapper object (instead of returning a simple list)
- [x] create new minimal issue object
- [x] store only C#/VB.NET issues

duncanp-sonar:
Fixes #2938

### DIFF
--- a/src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/InProcess/RoslynSettingsFileSynchronizerTests.cs
+++ b/src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/InProcess/RoslynSettingsFileSynchronizerTests.cs
@@ -33,6 +33,8 @@ using SonarLint.VisualStudio.Roslyn.Suppressions.InProcess;
 using SonarLint.VisualStudio.Roslyn.Suppressions.SettingsFile;
 using SonarQube.Client.Models;
 using EventHandler = System.EventHandler;
+using static SonarLint.VisualStudio.Roslyn.Suppressions.UnitTests.TestHelper;
+using System.Linq;
 
 namespace SonarLint.VisualStudio.Roslyn.Suppressions.UnitTests.InProcess
 {
@@ -154,6 +156,47 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions.UnitTests.InProcess
             roslynSettingsFileStorage.Verify(x => x.Update(It.IsAny<RoslynSettings>()), Times.Once);
         }
 
+        [TestMethod]
+        public async Task UpdateFileStorage_IssuesAreConvertedAndFiltered()
+        {
+            var configuration = CreateConnectedConfiguration("some project key");
+            var activeSolutionBoundTracker = CreateActiveSolutionBoundTracker(configuration);
+
+            var sonarIssues = new []
+            {
+                CreateSonarQubeIssue(ruleId: "csharpsquid:S111"), // C# issue
+                CreateSonarQubeIssue(ruleId: "vbnet:S222"),// VB issue
+                CreateSonarQubeIssue(ruleId: "cpp:S333"),// C++ issue - ignored
+                CreateSonarQubeIssue(ruleId: "xxx:S444"),// unrecognised repo - ignored
+                CreateSonarQubeIssue(ruleId: "xxxS555"),// invalid repo key - ignored
+                CreateSonarQubeIssue(ruleId: "xxx:"),// invalid repo key (no rule id) - ignored
+            };
+            var suppressedIssuesProvider = CreateSuppressedIssuesProvider(sonarIssues);
+
+            RoslynSettings actualSettings = null;
+            var roslynSettingsFileStorage = new Mock<IRoslynSettingsFileStorage>();
+            roslynSettingsFileStorage.Setup(x => x.Update(It.IsAny<RoslynSettings>()))
+                .Callback<RoslynSettings>(x => actualSettings = x);
+
+            var testSubject = CreateTestSubject(
+                suppressedIssuesProvider: suppressedIssuesProvider.Object,
+                roslynSettingsFileStorage: roslynSettingsFileStorage.Object,
+                activeSolutionBoundTracker: activeSolutionBoundTracker.Object);
+
+            await testSubject.UpdateFileStorageAsync();
+
+            actualSettings.Should().NotBeNull();
+            actualSettings.SonarProjectKey.Should().Be("some project key");
+            actualSettings.Suppressions.Should().NotBeNull();
+
+            var actualSuppressions = actualSettings.Suppressions.ToList();
+            actualSuppressions.Count.Should().Be(2);
+            actualSuppressions[0].RoslynLanguage.Should().Be(RoslynLanguage.CSharp);
+            actualSuppressions[0].RoslynRuleId.Should().Be("S111");
+            actualSuppressions[1].RoslynLanguage.Should().Be(RoslynLanguage.VB);
+            actualSuppressions[1].RoslynRuleId.Should().Be("S222");
+        }
+
         private static Mock<IActiveSolutionBoundTracker> CreateActiveSolutionBoundTracker(BindingConfiguration configuration)
         {
             var activeSolutionBoundTracker = new Mock<IActiveSolutionBoundTracker>();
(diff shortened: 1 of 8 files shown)


src/Roslyn.Suppressions/Roslyn.Suppressions/RoslynSettings.cs
```diff
@@ -31,7 +30,7 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions
     /// </summary>
     public class RoslynSettings
     {
-        public static readonly RoslynSettings Empty = new RoslynSettings { Suppressions = Enumerable.Empty<SonarQubeIssue>() };
+        public static readonly RoslynSettings Empty = new RoslynSettings { Suppressions = Enumerable.Empty<SuppressedIssue>() };
```
duncanp-sonar, line 33: This is the root of the changes - switching to a new type that contains fewer fields.

src/Roslyn.Suppressions/Roslyn.Suppressions/RoslynSettings.cs
```diff
@@ -42,45 +41,60 @@ public class RoslynSettings
...
-    //        RoslynLanguage = roslynLanguage;
-    //        RoslynRuleId = roslynRuleId;
-    //        TextRange = textRange;
-    //    }
+    public enum RoslynLanguage
+    {
+        Unknown = 0,
+        CSharp = 1,
+        VB = 2
+    }
```
duncanp-sonar, line 52: This enum is only used for in this assembly.
Currently, the language field is serialized as "1" or "2". I think we could tell NewtonSoft to serialize the name instead.
I'll have a quick look - if it's easy I'll do it.

src/Roslyn.Suppressions/Roslyn.Suppressions/RoslynSettings.cs
```diff
@@ -42,45 +41,60 @@ public class RoslynSettings
...
+        /// <remarks>This Sonar rule key without repository key.</remarks>
+        [JsonProperty("rule")]
+        public string RoslynRuleId { get; }
+
+        /// <summary>
+        /// The 0-based line for the issue. Will be null for file-level issues
+        /// </summary>
+        /// <remarks>Roslyn issues are 0-based - Sonar issues are 1-based.</remarks>
+        [JsonProperty("line")]
+        public int? RoslynIssueLine { get; }
```
duncanp-sonar, line 98: This needs to be nullable to indicate a file-level issue.
We can't use zero since that is a valid Roslyn line number.

src/Roslyn.Suppressions/Roslyn.Suppressions/RoslynSettings.cs
```diff
@@ -42,45 +41,60 @@ public class RoslynSettings
...
+
+        [JsonProperty("lang")]
+        public RoslynLanguage RoslynLanguage { get; }
+
+        /// <summary>
+        /// The rule ID reported by the Roslyn analyzer e.g. S123
+        /// </summary>
+        /// <remarks>This Sonar rule key without repository key.</remarks>
+        [JsonProperty("rule")]
+        public string RoslynRuleId { get; }
```
duncanp-sonar, line 91: I included "Roslyn" in the field name to make it clear we're not talking about Sonar rule keys.

src/Roslyn.Suppressions/Roslyn.Suppressions/SuppressionChecker.cs
```diff
@@ -117,23 +116,22 @@ public bool IsSuppressed(Diagnostic reportedDiagnostic, string settingsKey)
...
             ?? false;
 
-        private static bool IsSameLine(FileLinePositionSpan roslynLineSpan, IssueTextRange sonarTextRange)
+        private static bool IsSameLine(FileLinePositionSpan roslynLineSpan, SuppressedIssue suppressedIssue)
         {
             // Special case: file-level issues
             var roslynFileLevelIssue = IsRoslynFileLevelIssue(roslynLineSpan);
-            if (sonarTextRange == null || roslynFileLevelIssue)
+            var sonarFileLevelIssue = !suppressedIssue.RoslynIssueLine.HasValue;
+            if (sonarFileLevelIssue || roslynFileLevelIssue)
```
duncanp-sonar, line 128: Minor tweak to the code since we are now checking an `int?` rather than a nullable `IssueTextRange` reference type. The logic is the same, though.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -130,5 +136,51 @@ public void Dispose()
         {
             suppressedIssuesMonitor.SuppressionsUpdateRequested -= SuppressedIssuesMonitor_SuppressionsUpdateRequested;
         }
+
+        // Converts SonarQube issues to SuppressedIssues that can be compared more easily with Roslyn issues
+        internal static class IssueConverter
```
duncanp-sonar, line 141: I made this a nested class since this logic is only used by this class, so it doesn't need to pollute the namespace. I didn't think it was worth introducing an interface for it.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressionCheckerTests.cs
```diff
@@ -187,20 +187,19 @@ public void IsMatch_RuleIdsAreDifferent_ReturnsFalse()
         private const string DiagFileText = "0\n1\n2\n3 text \n\n";
         private const string DiagSelectedText = "text";
         private const int DiagRoslynLineNumber = 3; // the 0-base line in which the word "text" appears
-        private const int MatchingSonarLineNumber = DiagRoslynLineNumber + 1;
 
         [TestMethod]
-        [DataRow(DiagFileName, DiagRuleId, MatchingSonarLineNumber, DiagHash, true)]
-        [DataRow("c:\\wrong file name.cs", DiagRuleId, MatchingSonarLineNumber, DiagHash, false)]
-        [DataRow(DiagFileName, "wrong rule id", MatchingSonarLineNumber, DiagHash, false)]
+        [DataRow(DiagFileName, DiagRuleId, DiagRoslynLineNumber, DiagHash, true)]
```
duncanp-sonar, line 192: The suppression checker is now matching Roslyn `Diagnostic` instances against the new `SuppressedIssue` type.
The line numbers have already been translated from 1-based to 0-based so the comparisons are simpler.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/InProcess/RoslynSettingsFileSynchronizerTests.cs
```diff
@@ -170,12 +213,6 @@ private static Mock<ISonarQubeIssuesProvider> CreateSuppressedIssuesProvider(IEn
             return suppressedIssuesProvider;
         }
 
-        private SonarQubeIssue CreateSonarQubeIssue()
```
duncanp-sonar, line 173: Now using the common helper method.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -130,5 +136,51 @@ public void Dispose()
...
+            {
+                (string repoKey, string ruleKey) = GetRepoAndRuleKey(issue.RuleId);
+                var language = GetRoslynLanguage(repoKey);
+
+                int? line = issue.TextRange == null ? (int?)null : issue.TextRange.StartLine - 1;
+                return new SuppressedIssue(issue.FilePath,
+                    issue.Hash, language, ruleKey, line);
+            }
+
+            private static (string, string) GetRepoAndRuleKey(string sonarRuleId)
```
ugras-ergun-sonarsource, line 153: adding names to return tuple would make it easier to read I think

ugras-ergun-sonarsource (approved):
One small note about readability, the rest looks good.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
rita-sonarsource: I think it would've been cleaner to have an interface that we can mock in `UpdateFileStorageAsync` tests and then test the conversion separately. We can leave it like that for now.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/InProcess/RoslynSettingsFileSynchronizerTests.cs
```diff
@@ -154,6 +156,47 @@ public async Task UpdateFileStorage_FileStorageIsUpdatedOnBackgroundThread()
             roslynSettingsFileStorage.Verify(x => x.Update(It.IsAny<RoslynSettings>()), Times.Once);
         }
 
+        [TestMethod]
+        public async Task UpdateFileStorage_IssuesAreConvertedAndFiltered()
+        {
+            var configuration = CreateConnectedConfiguration("some project key");
+            var activeSolutionBoundTracker = CreateActiveSolutionBoundTracker(configuration);
+
+            var sonarIssues = new []
```
rita-sonarsource, line 165: Side comment: I think it would've been a bit cleaner and easier to read if we would just mock this, instead of repeating the details of the actual logic. But we can take a look at it next time.

rita-sonarsource (approved):

PR #2945, opened 2022-03-22
Add ETW tracing 

### LINKED ISSUE #2936: Add ETW tracing
### Description

See the existing CodeMarkers class

TODO
- [ ] decide what to record
- [ ] record it

? start/end: reloading settings from disc
? start/end: issue matching
? start/end: create execution context
? start/end: initialising the container
? fetching suppressions from the server

ugras-ergun-sonarsource:
fixes #2936

### DIFF
--- a/src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
+++ b/src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
@@ -20,6 +20,7 @@
 
 using System;
 using System.ComponentModel.Composition;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Threading;
 using SonarLint.VisualStudio.Core;
@@ -30,6 +31,7 @@ using SonarLint.VisualStudio.Infrastructure.VS;
 using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.Integration.Helpers;
 using SonarLint.VisualStudio.Roslyn.Suppressions.SettingsFile;
+using SonarQube.Client.Models;
 
 namespace SonarLint.VisualStudio.Roslyn.Suppressions.InProcess
 {
@@ -117,10 +119,14 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions.InProcess
                 if (!string.IsNullOrEmpty(sonarProjectKey))
                 {
                     var allSuppressedIssues = await suppressedIssuesProvider.GetAllSuppressedIssuesAsync();
+
                     var settings = new RoslynSettings
                     {
                         SonarProjectKey = sonarProjectKey,
-                        Suppressions = allSuppressedIssues,
+                        Suppressions = allSuppressedIssues
+                                            .Select(x => IssueConverter.Convert(x))
+                                            .Where(x => x.RoslynLanguage != RoslynLanguage.Unknown && !string.IsNullOrEmpty(x.RoslynRuleId))
+                                            .ToArray(),
                     };
                     roslynSettingsFileStorage.Update(settings);
                 }
@@ -134,6 +140,52 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions.InProcess
         public void Dispose()
         {
             suppressedIssuesMonitor.SuppressionsUpdateRequested -= SuppressedIssuesMonitor_SuppressionsUpdateRequested;
+        }
+
+        // Converts SonarQube issues to SuppressedIssues that can be compared more easily with Roslyn issues
+        internal static class IssueConverter
+        {
+            public static SuppressedIssue Convert(SonarQubeIssue issue)
+            {
+                (string repoKey, string ruleKey) = GetRepoAndRuleKey(issue.RuleId);
+                var language = GetRoslynLanguage(repoKey);
+
+                int? line = issue.TextRange == null ? (int?)null : issue.TextRange.StartLine - 1;
+                return new SuppressedIssue(issue.FilePath,
+                    issue.Hash, language, ruleKey, line);
+            }
+
+            private static (string repoKey, string ruleKey) GetRepoAndRuleKey(string sonarRuleId)
+            {
+                // Sonar rule ids are in the form "[repo key]:[rule key]"
+                var separatorPos = sonarRuleId.IndexOf(":", StringComparison.OrdinalIgnoreCase);
+                if (separatorPos > -1)
+                {
+                    var repoKey = sonarRuleId.Substring(0, separatorPos);
+                    var ruleKey = sonarRuleId.Substring(separatorPos + 1);
+
+                    return (repoKey, ruleKey);
+                }
+
+                return (null, null); // invalid rule key -> ignore
+            }
+
+            private static RoslynLanguage GetRoslynLanguage(string repoKey)
+            {
+                // Currently the only Sonar repos which contain Roslyn analysis rules are 
+                // csharpsquid and vbnet. These include "normal" and "hotspot" rules.
+                // The taint rules are in a different repo, and the part that is implemented
+                // as a Roslyn analyzer won't raise issues anyway.
+                switch (repoKey)
+                {
+                    case "csharpsquid": // i.e. the rules in SonarAnalyzer.CSharp
+                        return RoslynLanguage.CSharp;
+                    case "vbnet":       // i.e. SonarAnalyzer.VisualBasic
+                        return RoslynLanguage.VB;
+                    default:
+                        return RoslynLanguage.Unknown;
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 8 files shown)


src/Integration/ETW/CodeMarkers.cs
```diff
@@ -34,6 +34,7 @@ public static class Keywords
             public const EventKeywords Binding = (EventKeywords)2;
             public const EventKeywords Analysis = (EventKeywords)4;
             public const EventKeywords CFamily = (EventKeywords)8;
+            public const EventKeywords Suppression = (EventKeywords)2;
```
duncanp-sonar, line 37: We're only recording Roslyn-related events for the moment, but we do handle suppressions for other languages. For now, I think making the key `RoslynSuppression` would be better as it will remind us in the future that it's Roslyn-specific.
The alternative is to add another keyword for `Roslyn`, but I don't think that is worth doing at the moment.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -90,13 +91,18 @@ private void SuppressedIssuesMonitor_SuppressionsUpdateRequested(object sender,
             // does all of its work on a background thread.
             try
             {
+                CodeMarkers.Instance.FileSynchronizerUpdateRequesteStart();
```
duncanp-sonar, line 94: This will tell us how often the update is requested. The actual update is happening on a different thread, and this method isn't waiting for the result so we don't know how long the update will take.

I'm not sure if it's worth recording Start/Stop if all of the work is being done on another thread, but I guess it doesn't hurt.

However, do we want to measure how long the actual update takes?

src/Roslyn.Suppressions/Roslyn.Suppressions/SonarDiagnosticSuppressor.cs
```diff
@@ -70,7 +73,7 @@ public override void ReportSuppressions(SuppressionAnalysisContext context)
             {
                 return Enumerable.Empty<Suppression>();
             }
-
+            CodeMarkers.Instance.GetSuppressionsStart();
```
duncanp-sonar, line 76: Currently, this is going report as follows:
In standalone mode:
```
ReportSuppressionStart
ReportSuppressionStop
```

i.e. we have to infer that we are in standalone mode by the absence of the other events.

In connected mode:
```
ReportSuppressionStart
GetSuppressionsStart
GetSuppressionsStop(n)
ReportSuppressionStop
```
The two extra events are adding much from a timing point of view since they are called very close to the `ReportSuppressionStart/Stop`.

It would be simpler and more useful if we always reported something like:
```
ReportSuppressionStart
ReportSuppressionStop (mode, suppressionCount)
```

src/Integration/Integration.csproj
```diff
@@ -76,6 +76,10 @@
     <ProjectReference Include="..\ProgressVS\ProgressVS.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="ETW\" />
```
rita-sonarsource, line 80: Why is this needed?

src/Roslyn.Suppressions/Roslyn.Suppressions/SuppressionExecutionContext.cs
```diff
@@ -55,5 +55,7 @@ private void GetProjectKey(AnalyzerOptions analyzerOptions)
         public bool IsInConnectedMode => SettingsKey != null;
 
         public string SettingsKey { get; private set; } = null;
+
+        public string Mode => SettingsKey != null ? "Connected" : "Standalone";
```
ugras-ergun-sonarsource, line 59: Even it adds more complexity, I think having the mode name logic here makes the code more readable

src/Integration/Integration.csproj
ugras-ergun-sonarsource: it's not. A leftover from moving file. Removed it.

src/Roslyn.Suppressions/Roslyn.Suppressions/SuppressionExecutionContext.cs
rita-sonarsource, line 59: Can we maybe do this conversion in the ETW method and pass along the boolean IsInConnectedMode? It feels odd to add this extra field to the class when it's used only for logging.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -122,8 +122,7 @@ public async Task UpdateFileStorageAsync()
                 };
                 roslynSettingsFileStorage.Update(settings);
             }
-
-            logger.LogDebugExtended("End");
+            CodeMarkers.Instance.FileSynchronizerUpdateStop();
```
rita-sonarsource, line 125: Just to point out that this won't be logged if we have an exception in the middle. I don't know if it's relevant from ETW's perspective to have the Stop when there are errors.

src/Core/ETW/CodeMarkers.cs
```diff
@@ -34,6 +34,7 @@ public static class Keywords
             public const EventKeywords Binding = (EventKeywords)2;
             public const EventKeywords Analysis = (EventKeywords)4;
             public const EventKeywords CFamily = (EventKeywords)8;
+            public const EventKeywords RoslynSuppression = (EventKeywords)2;
```
rita-sonarsource, line 37: Isn't this supposed to be 16? why 2?

src/Core/ETW/CodeMarkers.cs
```diff
@@ -118,6 +119,59 @@ public static class Keywords
...
+        private const int FileStorageUpdateStopId = 3007;
+
+        private const int FileWatcherInvalidateStartId = 3007;
+        private const int FileWatcherInvalidateStopId = 3008;
+
+        private const int FileSynchronizerUpdateStartId = 3009;
+        private const int FileSynchronizerUpdateStopId = 3010;
+
+
+        [Event(ReportSuppressionsStartId, Level = EventLevel.Informational, Keywords = Keywords.RoslynSuppression)]
```
rita-sonarsource, line 143: Is this still used?

src/Core/ETW/CodeMarkers.cs
```diff
@@ -118,6 +119,59 @@ public static class Keywords
 
         #endregion
 
+        #region Roslyn Suppressions 3000-3999
+
+        private const int ReportSuppressionsStartId = 3000;
+        private const int ReportSuppressionsStopId = 3001;
+
+        private const int GetSuppressionsStartId = 3002;
```
rita-sonarsource, line 127: Some of these are not used anymore

src/Roslyn.Suppressions/Roslyn.Suppressions/SuppressionExecutionContext.cs
ugras-ergun-sonarsource: I thought of that, but I think this would make more sense from readability point and also we would not have context logic in ETW method.

rita-sonarsource: Then just to make it readable maybe reuse `IsInConnectedMode` instead of checking the settings key again.

rita-sonarsource (changes requested):
Please see my comments

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
duncanp-sonar: I don't think ETW cares - it just logs whatever events it is given. I'm guessing that the matching of Start/Stop event pairs is done by PerfView.

We have try-finally blocks for the other event pairs, although for those the try-catch already existed. However, we may as well be consistent and have a try-finally so we always have matching pairs of events - it will make the logs easier to understand.

src/Roslyn.Suppressions/Roslyn.Suppressions/SonarDiagnosticSuppressor.cs
```diff
@@ -52,6 +53,7 @@ internal SonarDiagnosticSuppressor(Func<IContainer> getContainer)
 
         public override void ReportSuppressions(SuppressionAnalysisContext context)
         {
+            CodeMarkers.Instance.ReportSuppressionsStart();
```
duncanp-sonar, line 56: This is now the only place we have ETW events that isn't in a try-finally block. However, I'd be inclined to leave it as it is.
I checked, and VS silently catches and squashes any exceptions thrown here (I thought it would report an AD0001 in the Error List, but it doesn't). 
If we see that we are losing "Stop" events we could add a try-finally at that point.

duncanp-sonar (approved):

PR #2949, opened 2022-03-24
Cache is not invalidated due to Case Mismatch on read and write 

### LINKED ISSUE #2947: Cache is not invalidated due to Case Mismatch on read and write
### Description

While invalidating the cache cache key we're trying to invalidate is in lower case and because of this it does not match to existing key if there's a upper case character.

ugras-ergun-sonarsource:
fixes #2947


src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/TestHelpers.cs
```diff
@@ -56,4 +59,30 @@ internal static class TestHelper
                 null
                 );
     }
+
+    //This is to make sure normalising the keys done correctly with culture invariant
+    //Lower case of SETTINGSKEY in Turkish is not settingskey but settıngskey
+    //https://en.wikipedia.org/wiki/Dotted_and_dotless_I 
+    public class TemporaryCultureSwitch : IDisposable
```
ugras-ergun-sonarsource, line 66: instead of adding a new file this made more sense. To remove the confusion I've changed the filename to "TestHelpers" instead of "TestHelper"

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/RoslynSettingsFileInfoTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        [DataRow("III", "iii")]//Testing upper case with invariant culture
+        [DataRow("project:key", "project_key")]//Testing illegal characters
+        [TestMethod]
+        public void GetSettingsFilePath_ReturnsFilePathCorrectly(string projectKey, string expectedFileName)
+        {
+            var expectedPath = Path.Combine(RoslynSettingsFileInfo.Directory, expectedFileName + ".json");
+
+            //This is to make sure normalising the keys done correctly with invariant culture
+            //https://en.wikipedia.org/wiki/Dotted_and_dotless_I 
+            using (new TemporaryCultureSwitch(new CultureInfo("tr-TR")))
```
duncanp-sonar, line 44: You could use the version of `using` without braces e.g.

`using var tempCulture = new TemporaryCultureSwitch(new CultureInfo("tr-TR"));`

I don't think it makes much difference in very short methods like this one, but it does match methods with more content inside the scope more readable.

src/Roslyn.Suppressions/Roslyn.Suppressions/Settings.Cache/SettingsCache.cs
```diff
@@ -42,17 +43,18 @@ internal SettingsCache(IRoslynSettingsFileStorage fileStorage, ConcurrentDiction
 
         public RoslynSettings GetSettings(string settingsKey)
         {
-            if(!settingsCollection.ContainsKey(settingsKey))
+            var normalisedKey = NormalizeKey(settingsKey);
```
duncanp-sonar, line 46: A better approach is to create the dictionary with type of key comparer we want it to use i.e.

```
new ConcurrentDictionary<string, RoslynSettings>(StringComparer.OrdinalIgnoreCase)
```

It's more reliable than remember to manually convert the key every time (and more efficient, since it doesn't involve converting and creating a new string.

Also, the MS Docs [recommend](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#choosing-a-stringcomparison-member-for-your-method-call) using `OrdinalIgnoreCase` when comparing file paths.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/TestHelpers.cs
duncanp-sonar: I'm not convinced it does make more sense; if we end up twenty test helpers then we probably wouldn't want them in all in the same class.
However, I don't feel strongly about it when there are only two.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/TestHelpers.cs
```diff
@@ -56,4 +59,30 @@ internal static class TestHelper
...
+        public TemporaryCultureSwitch(CultureInfo cultureInfo)
+        {
+            _originalCulture = Thread.CurrentThread.CurrentCulture;
+            _originalUICulture = Thread.CurrentThread.CurrentUICulture;
+
+            Thread.CurrentThread.CurrentCulture = cultureInfo;
+            Thread.CurrentThread.CurrentUICulture = cultureInfo;
+        }
+
+        public TemporaryCultureSwitch(string cultureName) : this(new CultureInfo(cultureName)) { }
```
duncanp-sonar, line 80: It looks like this class is always created with code like the following:

```
using (new TemporaryCultureSwitch(new CultureInfo("tr-TR")))
```

i.e. this constructor isn't being used, even though it is simpler.

In practice, we're very unlikely to use any culture other than Turkish in testing, so the code could be even simpler

e.g. a factory method `public static TemporaryCultureScope CreateTurkishScope()`

or even simpler

e.g. the class could be called `TurkishCultureScope` and hard-coded to use Turkish

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/RoslynSettingsFileInfo.cs
```diff
@@ -32,9 +32,9 @@ internal static class RoslynSettingsFileInfo
         /// </summary>
         public static string GetSettingsFilePath(string sonarProjectKey)
         {
-            var escapedName = PathHelper.EscapeFileName(sonarProjectKey);
+            var escapedName = PathHelper.EscapeFileName(NormalizeKey(sonarProjectKey));
```
duncanp-sonar, line 35: I'm not sure this is actually required currently if the dictionary is configured with the correct string comparer on constructor.

It won't hurt to force to lower case, but the `NormalizeKey` method doesn't need to be public.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/TestHelpers.cs
```diff
@@ -56,4 +59,30 @@ internal static class TestHelper
...
+
+    //This is to make sure normalising the keys done correctly with culture invariant
+    //Lower case of SETTINGSKEY in Turkish is not settingskey but settıngskey
+    //https://en.wikipedia.org/wiki/Dotted_and_dotless_I 
+    public class TemporaryCultureSwitch : IDisposable
+    {
+        private readonly CultureInfo _originalCulture;
+        private readonly CultureInfo _originalUICulture;
+
+        public TemporaryCultureSwitch(CultureInfo cultureInfo)
```
duncanp-sonar, line 71: FYI we have other test-only disposable types that make a temporary change. By convention they're named `Scope` rather than `Switch` e.g. `AssertIgnoreScope`.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/TestHelpers.cs
duncanp-sonar, line 66: This a general-purpose test class, so it might be better to put it in the shared `TestInfrastructure` project - we're more likely to remember it when writing other tests.

duncanp-sonar (changes requested):
See inline comments

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/RoslynSettingsFileInfo.cs
ugras-ergun-sonarsource: I think it's better to have it lowered and be consistent and use the same key everywhere.

src/TestInfrastructure/Helpers/TurkishCultureScope.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
```
duncanp-sonar, line 1: Missing copyright

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/TestHelper.cs
```diff
@@ -18,6 +18,9 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
+using System.Globalization;
+using System.Threading;
```
duncanp-sonar, line 23: Are these using still required?

duncanp-sonar (approved):
Just a couple of minor issues

PR #2953, opened 2022-03-24
Fix ETW ArrayTypeMismatch exception

### LINKED ISSUE #2952: ArrayTypeMismatch exception in ETW
```
ThreadID="15,728"
ProcessorNumber="1" 
message="EventSourceException while processing event "ReportSuppressionsStop": System.ArrayTypeMismatchException:Attempted to access an element as a type incompatible with the array." 
FormattedMessage="EventSourceException while processing event "ReportSuppressionsStop": System.ArrayTypeMismatchException:Attempted to access an element as a type incompatible with the array."
ActivityID="00018849-0001-0000-ec82-0000ffdcd7b5" 
```

image

duncanp-sonar:
Fixes #2952

### DIFF
--- a/src/Core/ETW/CodeMarkers.cs
+++ b/src/Core/ETW/CodeMarkers.cs
@@ -141,7 +141,7 @@ namespace SonarLint.VisualStudio.Core.ETW
         public void ReportSuppressionsStart() => Write(ReportSuppressionsStartId);
         
         [Event(ReportSuppressionsStopId, Level = EventLevel.Informational, Keywords = Keywords.RoslynSuppression)]
-        public void ReportSuppressionsStop(string mode ,int suppressionCount) => Write(ReportSuppressionsStopId, mode, suppressionCount.ToString());
+        public void ReportSuppressionsStop(string mode, int suppressionCount) => Write(ReportSuppressionsStopId, mode, suppressionCount.ToString());
 
         [Event(FileStorageGetStartId, Level = EventLevel.Informational, Keywords = Keywords.RoslynSuppression)]
         public void FileStorageGetStart() => Write(FileStorageGetStartId);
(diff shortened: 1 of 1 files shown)


duncanp-sonar:
With this fix the new events seem to be recorded correctly, although the number of suppressions looks suspiciously high and needs to be investigated (VS is configured just to analyse the current file, and there are only three suppressions listed in the file):

image

src/Core/ETW/CodeMarkers.cs
```diff
@@ -179,14 +179,21 @@ private void Write(int id)
         }
 
         [NonEvent]
-        private void Write(int id, params string[] texts)
+        private void Write(int id, string text)
         {
             if (IsEnabled())
             {
-                WriteEvent(id, texts);
+                WriteEvent(id, text);
```
rita-sonarsource, line 186: Why didn't it work before? Is it because the params array got passed as an array as well? So it ended up being array of array?

I'm guessing that this is because the base method accepts `params object[]`. What if we were to cast our array to object before sending?
I.e. `WriteEvent(texts.Cast<Object>().ToArray())`. If it works then we can keep our params method.

rita-sonarsource (commented):
Please see my comment, maybe there is another workaround. If not then LGTM.

rita-sonarsource:
> With this fix the new events seem to be recorded correctly, although the number of suppressions looks suspiciously high and needs to be investigated (VS is configured just to analyse the current file, and there are only three suppressions listed in the file):
> 
> image
> 

This explains why I didn't see all the events. I suspected that something was throwing but I thought it was the supression matching. Thanks for fixing!

src/Core/ETW/CodeMarkers.cs
duncanp-sonar: Yes, that's probably what was happening. We don't need to cast the array, we just need to change our overload to accept `params object[]`.

duncanp-sonar:
Passing the suppression count as an `int` to the `params object[]` array fixes the problem with the number being recorded:
image

The numbers 48, 49, 50 that were being shown in PerfView before are the ASCII codes for 0, 1, 2.

PR #2964, opened 2022-04-03
Allow C&C++ users to customize how and when to run the analysis 

### LINKED ISSUE #1348: CFamily: allow user to customize how to run the analysis
**Is your feature request related to a problem? Please describe.**
If SLVS is enabled you cannot save a file without running the analysis. Sometimes we want to save files without rerunning the analysis:
1- Running analysis while building your project makes VS unresponsive.
2- Batch saving triggers a large number of subprocesses at the same time which makes the machine unresponsive. This can happen when doing "find and replace all" or refactoring commands like renaming a variable in a project. 

**Describe the solution you'd like**
- Option to disable analysis while building.
- Option to disable analysis on saving and provide another shortcut to do it
- Limit the number of possible subprocesses running at the same time.

abbas-sabra-sonarsource:
fixes #1348


abbas-sabra-sonarsource:
Outdated and no longer useful to keep

Issue #2971, opened 2022-04-06
Refactor IAnalysisIssueLocation to have an ITextRange


duncanp-sonar:
... i.e. has-a instead of is-a.

PR #2974, opened 2022-04-07
Change IEdit to have ITextRange

rita-sonarsource:
Prep for [#2971](https://github.com/SonarSource/sonarlint-visualstudio/issues/2971)

### DIFF
--- a/src/Core/Analysis/AnalysisIssue.cs
+++ b/src/Core/Analysis/AnalysisIssue.cs
@@ -118,4 +118,22 @@ namespace SonarLint.VisualStudio.Core.Analysis
 
         public string LineHash { get; }
     }
+
+    public class TextRange : ITextRange
+    {
+        public TextRange(int startLine, int endLine, int startLineOffset, int endLineOffset, string lineHash)
+        {
+            StartLine = startLine;
+            EndLine = endLine;
+            StartLineOffset = startLineOffset;
+            EndLineOffset = endLineOffset;
+            LineHash = lineHash;
+        }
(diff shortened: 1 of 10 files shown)


src/Core/Analysis/AnalysisIssue.cs
```diff
@@ -118,4 +118,22 @@ public class AnalysisIssueLocation : IAnalysisIssueLocation
 
         public string LineHash { get; }
     }
+
+    public class TextRange : ITextRange
```
rita-sonarsource, line 122: Created a new class so it could be used in #2971

ugras-ergun-sonarsource (approved):

src/Core/Analysis/AnalysisIssue.cs
duncanp-sonar: Is this a VS text range or a Sonar text range? Or do  we use it for both?

src/Core/Analysis/IAnalysisIssue.cs
```diff
@@ -55,14 +55,29 @@ public interface IAnalysisIssueLocation : ITextRange
...
+        /// </summary>
         int StartLine { get; }
 
+        /// <summary>
+        /// 1-based line
+        /// </summary>
         int EndLine { get; }
 
+        /// <summary>
+        /// 0-based column
```
rita-sonarsource, line 74: We should be consistent and have everything either 1 based or 0 based. I've documented the existing behavior.

PR #2975, opened 2022-04-07
Notify EslintBridge that the caller is SonarLint 

### LINKED ISSUE #2364: [EslintBridge] Notify EslintBridge that the caller is SonarLint
### Description

The new version of `eslintbridge` supports an additional boolean command line argument to indicate that the caller is SonarLint.
This allows `eslintbridge` to make some performance optimisations e.g. skipping calculating metrics.

From [eslint-bridge\bin\server](https://github.com/SonarSource/SonarJS/blob/0e83c667fc6bc1687111db9343de73f725c992ca/eslint-bridge/bin/server#L4):
```
/*
* This script expects following arguments
*
* port - port number on which server should listen
* host - host address on which server should listen
* workDir - working directory from SonarQube API
* shouldUseTypeScriptParserForJS - whether TypeScript parser should be used for JS code (default true, can be set to false in case of perf issues)
* sonarlint - when running in SonarLint (used to not compute metrics, highlighting, etc)
* additionalRules - ; or : delimited paths to additional rule bundles
*/
```

### Notes
Currently, we don't pass any arguments when launching so default values are used. The launched process finds a free port and logs to the output stream port used. We parse it from the output stream.

Defaults: [see here](https://github.com/SonarSource/SonarJS/blob/0e83c667fc6bc1687111db9343de73f725c992ca/eslint-bridge/src/server.ts#L41).
* port: 0
* host: '127.0.0.1'. NOTE: this won't work if the machine only supports IPv6. Should really be the actual loopback address for the machine.
* working directory: directory for temporary files. Currently only used by security rules to store UCFGs, but it would be better to pass a real value. TODO - need to think about clearing up the directory afterwards (don't really want each analysis to add another directory that gets left behind).

ugras-ergun-sonarsource:
fixes #2364

### DIFF
--- a/src/Core/Helpers/PathHelper.cs
+++ b/src/Core/Helpers/PathHelper.cs
@@ -157,32 +157,23 @@ namespace SonarLint.VisualStudio.Core.Helpers
         /// Gets Temp Directory under SLVS folder under %Temp%
         /// </summary>
         /// <param name="perVSInstance">Determines if the path will be per VS Instance</param>
-        /// <param name="paths">Creates sub folder structure under SLVS folder</param>
-        public static string GetTempDirForTask(bool perVSInstance, params string[] paths)
+        /// <param name="folders">Creates sub folder structure under SLVS folder</param>
+        public static string GetTempDirForTask(bool perVSInstance, params string[] folders)
         {
             var SLVSTempFolder = Path.Combine(Path.GetTempPath(), "SLVS");
 
-            var taskPaths = new List<string> { SLVSTempFolder };
+            var taskFolders = new List<string> { SLVSTempFolder };
 
-            if (paths != null)
-            {
-                foreach (var path in paths)
-                {
-                    if (!string.IsNullOrWhiteSpace(path))
-                    {
-                        taskPaths.Add(path);
-                    }
-                }
-            }          
+            taskFolders.AddRange(folders);  
             
-            var taskFolder = Path.Combine(taskPaths.ToArray());
+            var taskPath = Path.Combine(taskFolders.ToArray());
 
             if(perVSInstance)
             {
-                return Path.Combine(taskFolder, perVSInstanceFolderName.ToString());
+                return Path.Combine(taskPath, perVSInstanceFolderName.ToString());
             }
 
-            return taskFolder;            
+            return taskPath;            
         }
     }
 }
--- a/src/Core/Helpers/PathHelper.cs
+++ b/src/Core/Helpers/PathHelper.cs
@@ -29,6 +29,8 @@ namespace SonarLint.VisualStudio.Core.Helpers
 {
     public static class PathHelper
     {
+        private static readonly Guid perVSInstanceFolderName = Guid.NewGuid();
+
         /// <summary>
         /// Replace all invalid file path characters with the underscore ("_").
         /// </summary>
@@ -148,6 +150,39 @@ namespace SonarLint.VisualStudio.Core.Helpers
         {
             string escapedPath = uri.OriginalString.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
             return Uri.UnescapeDataString(escapedPath);
+        }
+
+
+        /// <summary>
+        /// Gets Temp Directory under SLVS folder under %Temp%
+        /// </summary>
+        /// <param name="perVSInstance">Determines if the path will be per VS Instance</param>
+        /// <param name="paths">Creates sub folder structure under SLVS folder</param>
+        public static string GetTempDirForTask(bool perVSInstance, params string[] paths)
+        {
+            var SLVSTempFolder = Path.Combine(Path.GetTempPath(), "SLVS");
+
+            var taskPaths = new List<string> { SLVSTempFolder };
+
+            if (paths != null)
+            {
+                foreach (var path in paths)
+                {
+                    if (!string.IsNullOrWhiteSpace(path))
+                    {
+                        taskPaths.Add(path);
+                    }
+                }
+            }          
+            
+            var taskFolder = Path.Combine(taskPaths.ToArray());
+
+            if(perVSInstance)
+            {
+                return Path.Combine(taskFolder, perVSInstanceFolderName.ToString());
+            }
+
+            return taskFolder;            
         }
     }
 }
--- a/src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/RoslynSettingsFileInfo.cs
+++ b/src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/RoslynSettingsFileInfo.cs
@@ -25,7 +25,7 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions.SettingsFile
 {
     internal static class RoslynSettingsFileInfo
     {
-        public static readonly string Directory = Path.Combine(Path.GetTempPath(), "SLVS", "Roslyn");
+        public static readonly string Directory = PathHelper.GetTempDirForTask(false, "Roslyn");
 
         /// <summary>
         /// Returns the full file path for the given project key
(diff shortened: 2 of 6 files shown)


src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
```diff
@@ -354,5 +354,14 @@ private static void SafeKillProcess(Process process)
                 // do nothing
             }
         }
+
+        private static string GetDefaultParameters()
+        {
+            var workDir = Path.Combine(Path.GetTempPath(), "SLVS", "ESLintBridge", "workdir");
```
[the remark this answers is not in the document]
duncanp-sonar: This working directory will be the same every instance of Visual Studio. Two related questions:

1) what happens if there are multiple instances of VS running?

2) what was the previous behaviour? i.e. if we passed nothing did the server generate a per-instance directory, or did it always use the same one?

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -265,9 +266,22 @@ private static string GetQuotedScriptPath(string scriptPath)
...
+
             Debug.Assert(!scriptPath.Contains(quote), "Not expecting the imported script path to be quoted");
 
-            return quote + scriptPath + quote;
+            return quote + scriptPath + quote + defaultParameters;
+        }
+
+        private static string GetDefaultParameters()
+        {
+            var workDir = Path.Combine(Path.GetTempPath(), "SLVS", "ESLintBridge", "workdir");
```
duncanp-sonar, line 280: Perhaps we need a shared helper function for returning temporary working paths e.g. 
`static GetPerVSInstanceTempDir(string featureName)`

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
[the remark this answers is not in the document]
ugras-ergun-sonarsource: According to my talk with Tibor; 
1) It should be unique per project, so if two instances are open with different projects, it could create a problem; not sure if it's two instances of the same project. But I did not question it further because I've decided to go with a per-instance approach. 
2) This directory is used for only "security rules" at the moment, which are not available in SonarLint. So what we passed did not matter and does not matter now. But I've decided to implement a per-instance-based approach so if anything changes, it won't break.

src/Core/Helpers/PathHelper.cs
```diff
@@ -149,5 +151,38 @@ private static string ToFilePathString(Uri uri)
             string escapedPath = uri.OriginalString.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
             return Uri.UnescapeDataString(escapedPath);
         }
+
+
+        /// <summary>
+        /// Gets Temp Directory under SLVS folder under %Temp%
+        /// </summary>
+        /// <param name="perVSInstance">Determines if the path will be per VS Instance</param>
+        /// <param name="paths">Creates sub folder structure under SLVS folder</param>
```
duncanp-sonar, line 160: Naming: calling them `paths` is a bit misleading, they're really folder names.

src/Core/Helpers/PathHelper.cs
```diff
@@ -149,5 +151,38 @@ private static string ToFilePathString(Uri uri)
...
+        {
+            var SLVSTempFolder = Path.Combine(Path.GetTempPath(), "SLVS");
+
+            var taskPaths = new List<string> { SLVSTempFolder };
+
+            if (paths != null)
+            {
+                foreach (var path in paths)
+                {
+                    if (!string.IsNullOrWhiteSpace(path))
```
duncanp-sonar, line 171: We don't need to code defensively here; we're the only people who use this, so it we're passing in bad arguments then it's a problem we should fix.

It would be enough to assert e.g.
```
Debug.Assert(!folders.Any(x => string.IsNullOrWhiteSpace(x)), "Supplied folder name  should not be null/empty");
```

src/Core/Helpers/PathHelper.cs
```diff
@@ -149,5 +151,38 @@ private static string ToFilePathString(Uri uri)
...
+        /// </summary>
+        /// <param name="perVSInstance">Determines if the path will be per VS Instance</param>
+        /// <param name="paths">Creates sub folder structure under SLVS folder</param>
+        public static string GetTempDirForTask(bool perVSInstance, params string[] paths)
+        {
+            var SLVSTempFolder = Path.Combine(Path.GetTempPath(), "SLVS");
+
+            var taskPaths = new List<string> { SLVSTempFolder };
+
+            if (paths != null)
```
duncanp-sonar, line 167: `params` arguments are never null. If you don't specify any values the function is passed a zero-length array.

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -265,9 +267,23 @@ private static string GetQuotedScriptPath(string scriptPath)
...
+        }
+
+        private static string GetDefaultParameters()
+        {
+            var workDir = PathHelper.GetTempDirForTask(true , "ESLintBridge", "workdir");
+
+            
+            //To pass the sonarlint parameter we have to pass all the parameters before 
+            //Commandline interface for eslintbridge is not accepting named parameters  
+            return $" \"0\" \"127.0.0.1\" \"{workDir}\" \"true\" \"true\"";
```
duncanp-sonar, line 286: It's worth commenting what the parameters are.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeProcessTests.cs
```diff
@@ -354,5 +355,14 @@ private static void SafeKillProcess(Process process)
                 // do nothing
             }
         }
+
+        private static string GetDefaultParameters()
```
duncanp-sonar, line 359: Having a unit test that just contains a cut-and-paste from the product code seems a bit weak. I'm not sure how we'd improve it, though.

duncanp-sonar (changes requested):
See inline comments

src/Core/Helpers/PathHelper.cs
ugras-ergun-sonarsource: If we don't do that "Path.Combine" will throw an error. We can add an assert here, but I'd rather not to throw an exception.

src/Core/Helpers/PathHelper.cs
ugras-ergun-sonarsource: It can be passed as null explicitly, however unlikely it will be. We can also add an assert for that scenario, but I think we should keep the check.

src/Core/Helpers/PathHelper.cs
duncanp-sonar: If we were using values supplied by the end user or from a dynamic source that could fail then we would need to be defensive i.e. if there are potential runtime issues to guard against.

Here, the paths are constructed from static data. If a null is passed in then it's because there is a bug in the our calling code. We'd rather know about it sooner than later, so it's appropriate for an exception to be thrown.

We should code defensively against runtime issues. We shouldn't code defensively against our own code being badly written.

src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/RoslynSettingsFileInfo.cs
```diff
@@ -30,6 +30,7 @@ internal static class RoslynSettingsFileInfo
         /// <summary>
         /// Returns the full file path for the given project key
         /// </summary>
+        /// <remarks> File will be on a shared and would be able to accessed by multiple instances of VS simulateneously</remarks>
```
duncanp-sonar, line 33: Typo: `File will be in a shared location and ...`

PR #2982, opened 2022-04-14
Upgrade analyzer

rita-sonarsource:


src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -86,7 +86,12 @@ public Task InitLinter(IEnumerable<Rule> rules, CancellationToken cancellationTo
                 Environments = analysisConfiguration.GetEnvironments()
             };
 
-            return MakeCall("init-linter", initLinterRequest, cancellationToken);
+            var responseString = await MakeCall("init-linter", initLinterRequest, cancellationToken);
```
rita-sonarsource, line 89: Small error handling. The call doesn't throw if there is a problem, it just returns a string with the problem...

src/TypeScript/Rules/ActiveRulesCalculator.cs
```diff
@@ -89,7 +89,9 @@ private static Rule Convert(RuleDefinition ruleDefinition)
                 Key = ruleDefinition.EslintKey,
 
                 // TODO: handle parameterised rules #2284
-                Configurations = ruleDefinition.DefaultParams
+                Configurations = ruleDefinition.DefaultParams,
+
+                FileTypeTarget = new []{"MAIN"}
```
rita-sonarsource, line 94: new mandatory field. Options are "MAIN" and "TEST", I'm not sure where we are supposed to take them from.

src/TypeScript/EslintBridgeClient/Contract/AnalysisResponse.cs
```diff
@@ -81,6 +82,9 @@ internal class Issue
 
         [JsonProperty("cost")]
         public int? Cost { get; set; }
+
+        [JsonProperty("quickFixes")]
+        public QuickFix[] QuickFixes { get; set; }
```
rita-sonarsource, line 87: Support for quick fixes. For now I re-defined the types that are in core so that I could put JsonPropertyAttribute on them. We should think how we want to handle it. I don't think it makes sense for the types in core to have attributes with values that are specific to eslint-bridge.

src/TypeScript.UnitTests/Analyzer/EslintBridgeIssueConverterTests.cs
```diff
@@ -195,13 +196,13 @@ public void Convert_HasSecondaryLocations_IssueWithSingleFlowAndLocations()
 
             var expectedLocations = new List<AnalysisIssueLocation>
             {
-                new AnalysisIssueLocation("message1", "some file", new TextRange(3, 4, 1, 2, null)),
-                new AnalysisIssueLocation("message2", "some file", new TextRange(7, 8, 5, 6, null)),
+                new("message1", "some file", new TextRange(3, 4, 1, 2, null)),
```
rita-sonarsource, line 199: Had a namespace conflict here

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
```diff
@@ -257,7 +257,7 @@ public async Task Close_FailsToCallClose_EslintBridgeProcessStillStopped()
             return new TypeScript.EslintBridgeClient.EslintBridgeClient(AnalyzeEndpoint, eslintBridgeProcess, httpWrapper, analysisConfiguration);
         }
 
-        private static Mock<IEslintBridgeHttpWrapper> SetupHttpWrapper(string endpoint, string response = null, Action<object> assertReceivedRequest = null)
+        private static Mock<IEslintBridgeHttpWrapper> SetupHttpWrapper(string endpoint, string response = "OK!", Action<object> assertReceivedRequest = null)
```
ugras-ergun-sonarsource, line 260: Because of the bug that Rita fixed, we weren't throwing when there was an error. Now we have fixed the issue, the tests started to fail because of a null response.

duncanp-sonar:
Superceded by #2987

PR #2983, opened 2022-04-19
[TypeScript] File-level are not displayed in the Error List

### LINKED ISSUE #2443: [TypeScript] File-level are not displayed in the Error List
### Description

TypeScript analysis can return issues with a start line number of 0, indicating a file-level issue.
These issues do not appear in the Error List.
They should appear, but with empty `line` and `column` columns.

TODO:
* check if this occurs for other languages. Update: yes ,CFamily - see #2963.
* test that having issues with start line = 0 doesn't cause an issue elsewhere e.g. in the taggers

### Repro steps

1. Analyse a TS file that produces a file-level issue

e.g. analysing https://github.com/SonarSource/www.sonarlint.org/blob/master/src/pages/atom.tsx triggers _typescript:S3317_ : _Rename this file to "IndexPage"_.

### Expected behavior

The issue appears in the Error List, but not in the editor.

### Actual behavior

The issue does not appear in the Error List.

An assertion occurs in debug builds:
image

### Known workarounds

None

### Related information

`AccumulatingIssueConsumer.IsIssueInAnalysisSnapshot` filters out issues with a start line of < 1.

* SonarLint for VS Version: 4.34
* Languages in your solution: TypeScript

ugras-ergun-sonarsource:
fixes #2443

### DIFF
--- a/src/Core.UnitTests/Analysis/AnalysisIssueTests.cs
+++ b/src/Core.UnitTests/Analysis/AnalysisIssueTests.cs
@@ -45,9 +45,9 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Analysis
             analysisIssue.IsFileLevel().Should().BeTrue();
         }
 
-        private IAnalysisIssue CreateTestSubject(bool primaryLocationHasTextRange)
+        private IAnalysisIssueBase CreateTestSubject(bool primaryLocationHasTextRange)
         {
-            var analysisIssue = new Mock<IAnalysisIssue>();
+            var analysisIssue = new Mock<IAnalysisIssueBase>();
             var primaryLocation = new Mock<IAnalysisIssueLocation>();
 
             if(primaryLocationHasTextRange)
--- a/src/Core.UnitTests/Analysis/AnalysisIssueTests.cs
+++ b/src/Core.UnitTests/Analysis/AnalysisIssueTests.cs
@@ -1,8 +1,23 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
@@ -15,7 +30,7 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Analysis
     public class AnalysisIssueTests
     {
         [TestMethod]
-        public void IsFileLevel_PrimaryLocationHasTextRange_True()
+        public void IsFileLevel_PrimaryLocationHasTextRange_ReturnsFalse()
         {
             var analysisIssue = CreateTestSubject(true);
 
@@ -23,7 +38,7 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Analysis
         }
 
         [TestMethod]
-        public void IsFileLevel_PrimaryLocationHasNoTextRange_False()
+        public void IsFileLevel_PrimaryLocationHasNoTextRange_ReturnsTrue()
         {
             var analysisIssue = CreateTestSubject(false);
 
--- a/src/Core.UnitTests/Analysis/AnalysisIssueTests.cs
+++ b/src/Core.UnitTests/Analysis/AnalysisIssueTests.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Integration.UnitTests;
+
+namespace SonarLint.VisualStudio.Core.UnitTests.Analysis
+{
+    [TestClass]
+    public class AnalysisIssueTests
+    {
+        [TestMethod]
+        public void IsFileLevel_PrimaryLocationHasTextRange_True()
+        {
+            var analysisIssue = CreateTestSubject(true);
+
+            analysisIssue.IsFileLevel().Should().BeFalse();
+        }
+
+        [TestMethod]
+        public void IsFileLevel_PrimaryLocationHasNoTextRange_False()
+        {
+            var analysisIssue = CreateTestSubject(false);
+
+            analysisIssue.IsFileLevel().Should().BeTrue();
+        }
+
+        private IAnalysisIssue CreateTestSubject(bool primaryLocationHasTextRange)
+        {
+            var analysisIssue = new Mock<IAnalysisIssue>();
+            var primaryLocation = new Mock<IAnalysisIssueLocation>();
+
+            if(primaryLocationHasTextRange)
+            {
+                primaryLocation.SetupGet(p => p.TextRange).Returns(new DummyTextRange());
+            }
+            analysisIssue.SetupGet(a => a.PrimaryLocation).Returns(primaryLocation.Object);
+            return analysisIssue.Object;
+        }
+    }
+}
--- a/src/Core/Analysis/IAnalysisIssue.cs
+++ b/src/Core/Analysis/IAnalysisIssue.cs
@@ -100,4 +100,12 @@ namespace SonarLint.VisualStudio.Core.Analysis
         Bug,
         Vulnerability
     }
+
+    public static class IAnalysisIssueExtensions
+    {
+        public static bool IsFileLevel(this IAnalysisIssue issue)
+        {
+            return issue.PrimaryLocation.TextRange == null;
+        }
+    }
 }
--- a/src/Core/Analysis/IAnalysisIssue.cs
+++ b/src/Core/Analysis/IAnalysisIssue.cs
@@ -103,7 +103,7 @@ namespace SonarLint.VisualStudio.Core.Analysis
 
     public static class IAnalysisIssueExtensions
     {
-        public static bool IsFileLevel(this IAnalysisIssue issue)
+        public static bool IsFileLevel(this IAnalysisIssueBase issue)
         {
             return issue.PrimaryLocation.TextRange == null;
         }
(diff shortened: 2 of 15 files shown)


src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests _GetValue.cs
```diff
@@ -236,7 +238,7 @@ private object GetValue(string columnName)
             return content;
         }
 
-        private void AssertGetValueReturnsNull(int index = 0, string columnName = StandardTableKeyNames.Line)
+        private void AssertGetValueReturnsFalse(int index = 0, string columnName = StandardTableKeyNames.Line)
```
ugras-ergun-sonarsource, line 241: changed name because null content does not mean fail anymore.

src/Core.UnitTests/Analysis/AnalysisIssueTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Integration.UnitTests;
+
+namespace SonarLint.VisualStudio.Core.UnitTests.Analysis
+{
+    [TestClass]
+    public class AnalysisIssueTests
+    {
+        [TestMethod]
+        public void IsFileLevel_PrimaryLocationHasTextRange_True()
```
duncanp-sonar, line 18: The test names are the wrong way round: the "_True" test is checking for a `false` result, and vice versa.
Naming: we'd normally name the tests `xxx_ReturnsTrue/False`

src/Core.UnitTests/Analysis/AnalysisIssueTests.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
```
duncanp-sonar, line 1: Copyright

src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests _GetValue.cs
```diff
@@ -94,31 +94,44 @@ public void TryCreateDetailsStringContent_ReturnsIssueMessage()
...
         [DataRow(999)]
         public void GetValue_IndexOutOfRange_Null(int index)
         {
-            AssertGetValueReturnsNull(index);
+            AssertGetValueReturnsFalse(index);
         }
 
         [TestMethod]
-        public void GetValue_IssueHasNoSpan_Null()
+        public void GetValue_IssueHasNoSpanKeyIsLine_Null()
```
duncanp-sonar, line 109: The test name isn't easy to read. An extra `_` would help i.e. `GetValue_IssueHasNoSpan_KeyIsLine_Null`.
Same for the next test.

duncanp-sonar: ... and the two tests could be combined into a single test with two DataRows, couldn't it?

src/IssueViz/Editor/IIssueSpanCalculator.cs
```diff
@@ -32,7 +32,7 @@ public interface IIssueSpanCalculator
         /// Returns the text span corresponding to the supplied analysis issue location.
         /// Returns null if the location line hash is different from the snapshot line hash
         /// </summary>
-        SnapshotSpan CalculateSpan(ITextRange range, ITextSnapshot currentSnapshot);
+        SnapshotSpan? CalculateSpan(ITextRange range, ITextSnapshot currentSnapshot);
```
duncanp-sonar, line 35: The summary comment above needs to be updated (and it looks like it was wrong before because it looks like an EmptySpan was returned if the line hashes didn't match).

It looks like the new behaviour is to return the corresponding span, or `null` if there isn't a `TextRange`, or `EmptySpan` if the `TextRange` can't be found in the document/the line hashes don't match.

src/Integration.Vsix/SonarLintTagger/AccumulatingIssueConsumer.cs
```diff
@@ -105,9 +105,12 @@ public void Accept(string path, IEnumerable<IAnalysisIssue> issues)
         /// <summary>
         /// Checks that the analysis issue can be mapped to location in the text snapshot.
         /// </summary>
-        private bool IsIssueInAnalysisSnapshot(IAnalysisIssue issue) =>
-            // Sonar issues line numbers are 1-based; 0 = file-level issue
+        private bool IsIssueFileLevelOrInAnalysisSnapshot(IAnalysisIssue issue)
```
duncanp-sonar, line 108: The summary comment is now out of date. I'm not sure it's adding much value.

src/Core/Analysis/IAnalysisIssue.cs
```diff
@@ -100,4 +100,12 @@ public enum AnalysisIssueType
         Bug,
         Vulnerability
     }
+
+    public static class IAnalysisIssueExtensions
+    {
+        public static bool IsFileLevel(this IAnalysisIssue issue)
```
duncanp-sonar, line 106: Shouldn't this use `IAnalysisIssueBase` since that is the interface that has `PrimaryLocation`?

src/Integration.Vsix/ErrorList/IssuesSnapshot.cs
```diff
@@ -181,14 +181,22 @@ public override bool TryGetValue(int index, string keyName, out object content)
                     // Note: the line and column numbers are taken from the SnapshotSpan, not the Issue.
                     // The SnapshotSpan represents the live document, so the text positions could have
                     // changed from those reported from the Issue.
-                    content = issueViz.Span.Value.Start.GetContainingLine().LineNumber;
+                    // If Span is null it means file level issue and would not have a line.
```
duncanp-sonar, line 184: Wouldn't it be better to check the underlying issue and use the new extension method you've created? e.g.

` if (issueViz.Issue.IsFileLevel()) ...`

Here, we're checking a calculated value rather than the original, stable value that is the true indicator of whether an issue is file-level or not.

Or are there other circumstances in which we expect the span to be null?

duncanp-sonar: Or perhaps create another `IsFileLevel` extension method that accepts an issue visualization.

src/TypeScript/Analyzer/EslintBridgeIssueConverter.cs
```diff
@@ -46,6 +46,17 @@ public IAnalysisIssue Convert(string filePath, Issue issue)
             var ruleDefinitions = rulesProvider.GetDefinitions();
             var ruleDefinition = ruleDefinitions.Single(x => x.EslintKey != null && x.EslintKey.Equals(issue.RuleId, StringComparison.OrdinalIgnoreCase));
             var sonarRuleKey = ruleDefinition.RuleKey;
+            ITextRange textRange = null;
+
+            if(issue.Line != 0)
```
duncanp-sonar, line 51: This deserves a comment that for the JS/TS analyzer `Line = 0` means the issue is a file-level issue.

src/Integration.Vsix.UnitTests/SonarLintTagger/AccumulatingIssueConsumerTests.cs
```diff
@@ -184,6 +201,15 @@ private static ITextSnapshot CreateSnapshot(int lineCount)
                 }
             };
 
+        private static IAnalysisIssue CreateFileLevelIssue()
+        {
+          return new DummyAnalysisIssue
+            {
+                PrimaryLocation = new DummyAnalysisIssueLocation()
```
duncanp-sonar, line 208: Minor: this would be clearer and more robust if it explicitly set the `TextRange` to `null` (it would be a strong hint that `TextRange` being `null` is what makes it a file-level issue, and the test won't break if we later added a default constructor to `DummyAnalysisIssueLocation` that changed its behaviour).

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -331,6 +333,30 @@ public void WhenNewIssuesAreFound_AndFilterRemovesAllIssues_ListenersAreUpdated(
...
+            testSubject.HandleNewIssues(issues);
+
+            // Assert
+            // Check the expected issues were passed to the filter
+            capturedFilterInput.Count.Should().Be(1);
+            capturedFilterInput.Should().BeEquivalentTo(issues);
+
+            CheckErrorListRefreshWasRequestedOnce(testSubject.Factory);
+
+            // Check there are no issues
```
duncanp-sonar, line 356: Comment is incorrect.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -165,6 +165,8 @@ public void Dispose_CleansUpEventsAndRegistrations()
             mockedJavascriptDocumentFooJs.VerifyRemove(x => x.FileActionOccurred -= It.IsAny<EventHandler<TextDocumentFileActionEventArgs>>(), Times.Once);
         }
 
+        //public void TranslateSpans_FileLevelIssue
```
duncanp-sonar, line 168: Left-behind WIP

duncanp-sonar:
Is this change going to work correctly with secondary locations?

Now, there are various places in the code where `Span == null` is taken to mean that the issue as a file-level issue.

However, when we looked at this before, @rita-sonarsource pointed out that when navigating to secondary locations, `Span == null` is taken to mean that we now need to calculate the span - see [here](https://github.com/SonarSource/sonarlint-visualstudio/blob/63a9400ef82f8e62298b8ab2eb433723dc17a8f8/src/IssueViz/Editor/LocationNavigator.cs#L82).

I think we need to separate those two things; we shouldn't be overloading `Span == null` to mean different things in different places. We might be able to do this by always going back to the original issue and checking `TextRange == null` to determine whether an issue is file-level or not.

duncanp-sonar (changes requested):
See inline. `Span == null` now means different things in different places. Might be easiest if we had a chat tomorrow about how to resolve this.

src/IssueViz/Models/AnalysisIssueVisualizationConverter.cs
```diff
@@ -122,7 +122,7 @@ private IReadOnlyList<IQuickFixVisualization> GetQuickFixVisualizations(IAnalysi
             {
                 var editVisualizations = fix.Edits.Select(edit =>
                 {
-                    var editSpan = issueSpanCalculator.CalculateSpan(edit.RangeToReplace, textSnapshot);
+                    var editSpan = issueSpanCalculator.CalculateSpan(edit.RangeToReplace, textSnapshot) ?? new SnapshotSpan();
```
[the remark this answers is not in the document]
ugras-ergun-sonarsource: I've made it so we wouldn't change how quickfixes behave. We could discuss it to make this better but I don't think it should be scope of this task.

src/Core/Analysis/IAnalysisIssue.cs
```diff
@@ -100,4 +100,12 @@ public enum AnalysisIssueType
         Bug,
         Vulnerability
     }
+
+    public static class IAnalysisIssueExtensions
+    {
+        public static bool IsFileLevel(this IAnalysisIssueBase issue)
```
ugras-ergun-sonarsource, line 106: We discussed About making this for IAnalysisIssue but to be able to use this in issuevisualisation we need to use base interface.

duncanp-sonar:
@ugras-ergun-sonarsource the changes look ok, but some of the tests are failing on the CI machine.
Are they failing on your machine?

duncanp-sonar (approved):
Looks good.

Did you check whether navigation works for file-level issues?
Did you decide that it wasn't necessary to add a file-level check to the navigator [here](https://github.com/SonarSource/sonarlint-visualstudio/blob/63a9400ef82f8e62298b8ab2eb433723dc17a8f8/src/IssueViz/Editor/LocationNavigator.cs#L82)?

ugras-ergun-sonarsource:
> 

We don't hit try navigate from file-level issues. (or at least I was not able to do it). And no other places were broken (secondary locations...etc) So I have decided adding a control there would be unnecessary.

duncanp-sonar:
@ugras-ergun-sonarsource ok. Please remember to add comments to the review about material decisions like this. The onus shouldn't be on reviewers to ask.

PR #2987, opened 2022-04-22
Update embedded JS/TS analyzer to 8.9.0.17411 

### LINKED ISSUE #2829: Update embedded JS/TS analyzer to 8.9.0.17411
* Deprecating Node.js 12 support for JS/TS analyses
* Fewer False Positive

https://github.com/SonarSource/SonarJS/releases/tag/8.8.0.17228

Note: SonarJS v9.0 contains a breaking change for VS2019 users. Separate ticket created: #2985

ugras-ergun-sonarsource:
fixes #2829

### DIFF
--- a/src/TypeScript.UnitTests/Rules/ActiveRulesCalculatorTests.cs
+++ b/src/TypeScript.UnitTests/Rules/ActiveRulesCalculatorTests.cs
@@ -35,6 +35,23 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Rules
         private static readonly IUserSettingsProvider EmptyUserSettingsProvider = new UserSettingsBuilder();
 
         [TestMethod]
+        public void Get_ReturnsRulesWithCorrectFileTargetType()
+        {
+            var ruleDefns = new RuleDefinitionsBuilder()
+                .AddRule("active 1", activeByDefault: true)
+                .AddRule("inactive AAA", activeByDefault: false)
+                .AddRule("active 2", activeByDefault: true);
+
+            var testSubject = CreateTestSubject(ruleDefns, EmptyUserSettingsProvider);
+
+            var result = testSubject.Calculate().ToArray();
+
+            result.Length.Should().Be(2);
+            result[0].FileTypeTarget.Should().BeEquivalentTo("MAIN");
+            result[1].FileTypeTarget.Should().BeEquivalentTo("MAIN");
+        }
+
+        [TestMethod]
         public void Get_NoUserOverrides_InactiveRulesAreNotReturned()
         {
             var ruleDefns = new RuleDefinitionsBuilder()
--- a/src/TypeScript/EslintBridgeClient/Contract/InitLinterRequest.cs
+++ b/src/TypeScript/EslintBridgeClient/Contract/InitLinterRequest.cs
@@ -38,7 +38,11 @@ namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract
     {
         [JsonProperty("key")]
         public string Key { get; set; }
+
         [JsonProperty("configurations")]
         public object[] Configurations { get; set; }
+
+        [JsonProperty("fileTypeTarget")]
+        public string[] FileTypeTarget { get; set; }
     }
 }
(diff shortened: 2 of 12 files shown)


src/TypeScript.UnitTests/Rules/ActiveRulesCalculatorTests.cs
```diff
@@ -34,6 +34,23 @@ public class ActiveRulesCalculatorTests
     {
         private static readonly IUserSettingsProvider EmptyUserSettingsProvider = new UserSettingsBuilder();
 
+        [TestMethod]
+        public void Get_ReturnsRulesWithCorrectFileTargetType()
```
rita-sonarsource, line 38: @ugras-ergun-sonarsource FYI added this test

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -86,7 +86,12 @@ public Task InitLinter(IEnumerable<Rule> rules, CancellationToken cancellationTo
                 Environments = analysisConfiguration.GetEnvironments()
             };
 
-            return MakeCall("init-linter", initLinterRequest, cancellationToken);
+            var responseString = await MakeCall("init-linter", initLinterRequest, cancellationToken);
+
+            if (!"OK!".Equals(responseString))
+            {
+                throw new InvalidOperationException(Resources.ERR_InvalidResponse);
```
duncanp-sonar, line 93: Would it help troubleshooting if the error included the `responseString`?

src/TypeScript/EslintBridgeClient/Contract/InitLinterRequest.cs
```diff
@@ -38,7 +38,11 @@ internal class Rule
     {
         [JsonProperty("key")]
         public string Key { get; set; }
+
         [JsonProperty("configurations")]
         public object[] Configurations { get; set; }
+
+        [JsonProperty("fileTypeTarget")]
+        public string[] FileTypeTarget { get; set; }
```
duncanp-sonar, line 46: The parameter has been added to the rule metadata to describe which type of files it applies to.

There is a corresponding change on the "analyze file" method that we have not implemented i.e. we should be passing in the type of the file being analyzed.
See here: https://github.com/SonarSource/SonarJS/blob/master/eslint-bridge/src/analyzer.ts

The fact that analysis seems to work without us passing this parameter is probably just lucky. We don't currently have a way to decide whether a file is a test file or not, so all we could do is pass a hard-coded value. However, it would probably be safer to pass it.

src/TypeScript/Rules/ActiveRulesCalculator.cs
```diff
@@ -89,7 +89,9 @@ private static Rule Convert(RuleDefinition ruleDefinition)
                 Key = ruleDefinition.EslintKey,
 
                 // TODO: handle parameterised rules #2284
-                Configurations = ruleDefinition.DefaultParams
+                Configurations = ruleDefinition.DefaultParams,
+
+                FileTypeTarget = new []{"MAIN"}
```
duncanp-sonar, line 94: Categorising the dozen `TEST` rules as `MAIN` rules have presumably means they will be run against all files. If they do report any issues then they will be false positives.

We should check whether those rules will report issues against `MAIN` files, If they do, we should consider excluding them from the rules config (e.g. by having a hard-coded list of rules we don't want to run ever).

I'll add a ticket to the board to investigate whether this is necessary.

duncanp-sonar:
FYI [eslintbridge](https://github.com/SonarSource/SonarJS/blob/master/eslint-bridge/src/analyzer.ts) now handles [Stylelint](https://stylelint.io/) rules configuration too. However, it looks like this is CSS-only, so I think we can ignore it for now.

duncanp-sonar (commented):
A couple of comments inline. I think we should pass the file category when analysing a file, although that change could be part of a separate PR.

src/TypeScript/EslintBridgeClient/Contract/InitLinterRequest.cs
ugras-ergun-sonarsource: I've added another ticket to project for this.

duncanp-sonar (approved):

PR #2996, opened 2022-04-26
Introduce NodeVersionInfoProvider

rita-sonarsource:
Two commits can be looked at separately, the first one is cosmetic renaming.
Prep for [#2997](https://github.com/SonarSource/sonarlint-visualstudio/issues/2997)

### DIFF
--- a/src/TypeScript.UnitTests/NodeJSLocator/CompatibleNodeLocatorTests.cs
+++ b/src/TypeScript.UnitTests/NodeJSLocator/CompatibleNodeLocatorTests.cs
@@ -20,10 +20,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel.Composition;
-using System.Diagnostics;
-using System.IO.Abstractions;
-using System.Reflection;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
@@ -41,13 +37,13 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.NodeJSLocator
         {
             MefTestHelpers.CheckTypeCanBeImported<CompatibleNodeLocator, ICompatibleNodeLocator>(null, new[]
             {
-                MefTestHelpers.CreateExport<INodeLocationsProvider>(Mock.Of<INodeLocationsProvider>()),
+                MefTestHelpers.CreateExport<INodeVersionInfoProvider>(Mock.Of<INodeVersionInfoProvider>()),
                 MefTestHelpers.CreateExport<ILogger>(Mock.Of<ILogger>())
             });
         }
 
         [TestMethod]
-        public void Locate_NoCandidateLocations_Null()
+        public void Locate_NoDetectedVersions_Null()
         {
             var testSubject = CreateTestSubject();
             var result = testSubject.Locate();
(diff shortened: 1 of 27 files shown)


src/TypeScript/NodeJSLocator/NodeVersionInfoProvider.cs
```diff
@@ -19,48 +19,42 @@
...
             Func<string, Version> getNodeExeVersion)
         {
             this.nodeLocationsProvider = nodeLocationsProvider;
-            this.logger = logger;
             this.fileSystem = fileSystem;
             this.getNodeExeVersion = getNodeExeVersion;
         }
 
-        public string Locate()
+        public IEnumerable<NodeVersionInfo> GetAllNodeVersions()
```
rita-sonarsource, line 57: This class is now responsible for lazily locating all the node installations and calculating their versions.
Telemetry will call this.

src/TypeScript/EslintBridgeClient/EslintBridgeProcess.cs
```diff
@@ -65,7 +65,7 @@ internal sealed class EslintBridgeProcess : IEslintBridgeProcess
         private static readonly object Lock = new object();
 
         private readonly string eslintBridgeStartupScriptPath;
-        private readonly INodeLocator nodeLocator;
+        private readonly ICompatibleNodeLocator compatibleNodeLocator;
```
rita-sonarsource, line 68: Cosmetics. Can be look at separately in the first commit.

src/TypeScript/NodeJSLocator/NodeVersionInfoProvider.cs
```diff
@@ -71,23 +65,8 @@ public string Locate()
...
 
-                if (!IsCompatibleVersion(nodeVersion))
-                {
-                    logger.WriteLine(Resources.ERR_IncompatibleVersion, nodeVersion, nodeExePath);
-                    continue;
-                }
-
-                logger.WriteLine(Resources.INFO_FoundCompatibleVersion, nodeVersion, nodeExePath);
-                return nodeExePath;
+                yield return new NodeVersionInfo(nodeExePath, nodeVersion);
```
rita-sonarsource, line 68: yield return so it doesn't degrade the performance of eslint-bridge initialization. We will still stop as soon as we find the first compatible.

src/TypeScript.UnitTests/NodeJSLocator/CompatibleNodeLocatorTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        public void Locate_NoDetectedVersions_Null()
+        {
+            var testSubject = CreateTestSubject();
+            var result = testSubject.Locate();
+
+            result.Should().BeNull();
+        }
+
+        [TestMethod]
+        public void Locate_ReturnsFirstCompatiblePath()
```
rita-sonarsource, line 55: Tests are now simplified

src/TypeScript/NodeJSLocator/NodeVersionInfoProvider.cs
duncanp-sonar: It's a non-obvious perf optimisation, so it's worth commenting in the code.

duncanp-sonar (approved):
My former boss was fond of saying "there's no problem that can't be solved by adding a layer of abstraction, except having too many layers of abstraction."

Having `ICompatibleNodeLocator`, `INodeVersionInfoProvider` and `INodeLocationsProvider` and all of the associated tests is starting to feel over-complicated. Still, it's subjective whether the alternatives are any better, so I wouldn't change it.

ugras-ergun-sonarsource (approved):

PR #3017, opened 2022-05-03
Re-enable secrets analysis

duncanp-sonar:


src/Integration.Vsix/SonarLintIntegrationPackage.cs
```diff
@@ -38,6 +38,10 @@ namespace SonarLint.VisualStudio.Integration.Vsix
     [Guid(CommonGuids.Package)]
     [ProvideMenuResource("Menus.ctmenu", 1)]
     [ProvideBindingPath]
+    // The secrets assemblies are in a sub-folder so we need to add it to the probing path
+    // to make sure it can be loaded reliably.
+    // See https://github.com/SonarSource/sonarlint-visualstudio/issues/3008
+    [ProvideBindingPath(SubPath = "secrets")]
```
duncanp-sonar, line 44: This is the project that contains the manifest which specifies that the assembly should be installed into the `secrets` subfolder, so it makes sense for the binding path attribute to be in this project too.

duncanp-sonar:
This PR contains the same code changes as the previously-approved-but-now-closed PR #2998.
There are a couple of minor differences:
* I created a new branch based on master to fix all of the package merge conflicts
* it doesn't contain the C# analyzer upgrade, since that is already in master, and
* this PR also contains the `ProvidBindingAttribute` change

src/Integration.Vsix/SonarLintIntegrationPackage.cs
duncanp-sonar: Also, I managed to reproduce the failure on a new VM.
On that VM (`slvs-vs2019latest`), the issue occurs if I install a vsix that does not have this attribute (e.g. [47369](https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=47369&view=artifacts&pathAsName=false&type=publishedArtifacts)).
It does not occur with a vsix that does contain this attribute (e.g. [47431](https://dev.azure.com/sonarsource/DotNetTeam%20Project/_build/results?buildId=47431&view=artifacts&pathAsName=false&type=publishedArtifacts)).

ugras-ergun-sonarsource (approved):

rita-sonarsource (approved):

PR #3032, opened 2022-05-21
Move subprocess runner code from Integration.Vsix to CFamily

duncanp-sonar:
* no code changes


src/CFamily/CFamily.csproj
```diff
@@ -29,6 +29,11 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>Resources.resx</DependentUpon>
     </Compile>
+    <Compile Update="Subprocess\CFamilyStrings.Designer.cs">
```
duncanp-sonar, line 32: To reduce the number of file edits, I created a new resources file under the new "Subprocess" folder with the same name as the existing Integration.Vsix resources file. I can rename it later, along with the namespaces.

The new file contains just the subprocess runner-specific strings.

src/CFamily.UnitTests/Subprocess/ProcessRunnerTests.cs
```diff
@@ -26,6 +26,7 @@
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.CFamily.Subprocess;
```
duncanp-sonar, line 29: I haven't changed the namespaces of the test and product files yet, so they don't match the namespace of the generated resources file -> this `using` is required so the resources strings are accessible.

src/Integration.Vsix/CFamily/CFamilyStrings.resx
```diff
@@ -117,48 +117,12 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="ERROR_ProcessRunner_ExeNotFound" xml:space="preserve">
```
duncanp-sonar, line 120: Moved runner-specific strings.

duncanp-sonar:
Summary of changes:

* created a new `Subprocess` folder in `CFamily` and moved the subprocess runner code to it
* moved `ISonarLintSettings` to `Core` so it can be referenced by the VSIX and the sub-process runner

PR #3036, opened 2022-05-24
Move PCH from Integration.VSIX to CFamily

duncanp-sonar:
* no code changes
* supporting types moved too
* scope of some types changed since they are now used in more than one assembly


src/CFamily/PreCompiledHeaders/PreCompiledHeadersEventListener.cs
```diff
@@ -29,7 +29,7 @@
 
 namespace SonarLint.VisualStudio.Integration.Vsix.CFamily
 {
-    interface IPreCompiledHeadersEventListener : IDisposable
+    public interface IPreCompiledHeadersEventListener : IDisposable
```
duncanp-sonar, line 32: This is instantiated by the SonarLintDaemon package so it needs to be public now.

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -128,7 +128,9 @@ Assembly: 'SonarLint.VisualStudio.CFamily, Version=6.5.0.0, Culture=neutral, Pub
 Relative path: 'SonarLint.VisualStudio.CFamily.dll'
 
 Referenced assemblies:
+- 'Microsoft.VisualStudio.CoreUtility, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
 - 'Microsoft.VisualStudio.Shell.Framework, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
+- 'Microsoft.VisualStudio.Text.Data, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```
duncanp-sonar, line 133: Types from these two VS interfaces are exposed by the `ISonarLangaugeRecognizer` interface so the `CFamily` assembly now depends on them too.

rita-sonarsource (approved):

PR #3039, opened 2022-05-25
Move CFamily types from Core to CFamily

duncanp-sonar:
* no code changes


src/CFamily.UnitTests/Helpers/DummyExeHelper.cs
```diff
@@ -18,9 +18,6 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-// TODO: 21/0/5/2022 - this class can be moved to the CFamily.UnitTests once CLangAnalyzer has
-// been moved from Integration.Vsix to CFamily.
-
```
duncanp-sonar, line 23: Done, so the comment is no longer relevant.

duncanp-sonar:
With this change, all of the easy-to-move CFamily-specific types have been moved out of `Integration.VSIX` and `Core` into `CFamily`. There are still a few types in `Core` relating to binding and telemetry, which are used in `Integration`. They will require more refactoring to move so I'll leave them for now.

The next step will be to fix up the namespaces and limit the scopes of types where possible. This will be very low risk but could be quite noisy, so I probably won't send it for review unless I come across something controversial/worth discussing.

rita-sonarsource (approved):

PR #3046, opened 2022-05-31
Implement FileStorage for saving and reading Files 

### LINKED ISSUE #3057: Implement FileStorage for saving and reading Files

ugras-ergun-sonarsource:
fixes #3057


duncanp-sonar:
The naming and location of the settings file is quite important from a UX point of view - what's the thinking behind the name/location you've chosen?

* the other binding settings are in the solution-level `.sonarlint` folder. What's the reason for this file being saved elsewhere? 
* do we expect this file to be committed to source control?
* if so, is it better to add a new file, or does the data naturally fit in one of the existing files?
* the directory/file name `Extensions` is quite specific. Would something more general be better e.g. `AnalysisSettings` / `ServerSettings` ?

rita-sonarsource:
It's an interesting debate. `.sonarlint` folder contains files that we needed to mark as AdditionalFiles for the roslyn analyzer. The users _don't_ like this and in Rider we chose not to do it. Saving it to disk under SCC would've been more consistent, but since we know our users hate this approach, perhaps it's time that we stop doing that.

duncanp-sonar:
> Saving it to disk under SCC would've been more consistent, but since we know our users hate this approach, perhaps it's time that we stop doing that.

It's tempting. However, all of the other rule settings and config files are already in the `.sonarlint` folder and it would be confusing if we had to explain why some settings are under `.sonarlint` and others are not.
There is a task in the roadmap about changing connected mode and how the configuration is stored so it might be better to remain consistent for now, and leave making any changes in how/where the configuration is stored until we address that roadmap item.

src/Integration.UnitTests/Exclusions/ExclusionSettingsFileStorageTests.cs
```diff
@@ -0,0 +1,188 @@
...
+            logger.AssertOutputStrings(1);
+            logger.AssertPartialOutputStringExists("Error writing settings for project");
+        }
+
+        private ExclusionSettingsFileStorage CreateTestSubject(IFileSystem fileSystem, ILogger logger)
+        {
+            return new ExclusionSettingsFileStorage(logger, fileSystem);
+        }
+
+        private IFileSystem CreateFileSystem(IFile file)
```
rita-sonarsource, line 166: Minor clean-up suggestion: the file system is never really checked, so maybe CreateTestSubject can receive IFile. Would make the tests slightly shorter and easier to read.

src/Integration.UnitTests/Exclusions/ExclusionSettingsFileStorageTests.cs
```diff
@@ -0,0 +1,188 @@
...
+            settings.Should().NotBeNull();
+            settings.Inclusions.Count().Should().Be(2);
+            settings.Exclusions.Count().Should().Be(1);
+            settings.GlobalInclusions.Count().Should().Be(1);
+            settings.GlobalExclusions.Count().Should().Be(1);
+            logger.AssertOutputStrings(0);
+        }
+
+        [TestMethod]
+        public void GetSettings_HaveNotSettings_ReturnsNull()
```
rita-sonarsource, line 67: Cosmetics: "HaveNotSettings" can also sound as if the file exists but empty. Maybe "SettingsFileDoesNotExist" is better

rita-sonarsource (commented):
@ugras-ergun-sonarsource Added some comments for the tests that are unrelated to where we would save the file.

src/Core/Exclusions/IExclusionSettingsFileStorage.cs
```diff
@@ -0,0 +1,30 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.Core.Exclusions
```
rita-sonarsource, line 23: FYI, the interface I added is under `SonarLint.VisualStudio.Core.Analysis`. I think Analysis is a better folder for this, wdyt? either way one of them needs to go

src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
```diff
@@ -17,23 +17,84 @@
...
+using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Core.Exclusions;
+using SonarLint.VisualStudio.Core.Helpers;
+using SonarLint.VisualStudio.Integration.NewConnectedMode;
+using SonarLint.VisualStudio.Integration.Resources;
 using SonarQube.Client.Models;
 
 namespace SonarLint.VisualStudio.Integration.Exclusions
 {
-    [Export(typeof(IExclusionSettingsFileStorage))]
```
rita-sonarsource, line 27: If it's no longer MEF-exportable, then AnalyzableFileIndicator is going to fail at runtime

src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
```diff
@@ -17,23 +17,84 @@
...
     internal class ExclusionSettingsFileStorage : IExclusionSettingsFileStorage
     {
-        public void SaveSettings(ServerExclusions settings)
+        private readonly BindingConfiguration bindingConfiguration;
+        private readonly ILogger logger;
+        private readonly IFileSystem fileSystem;
+
+        public ExclusionSettingsFileStorage(ILogger logger, IFileSystem fileSystem, BindingConfiguration bindingConfiguration)
```
rita-sonarsource, line 39: How is this class going to be instantiated by AnalyzableFileIndicator?
I think that maybe we should split between the read and the write. The read takes the bindingConfiguration only to write logging:
1) is that really a message that should be printed to the user at every save/file open?
2) the analysis indicator has no way to get this info without getting extra classes in the CTOR

src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
```diff
@@ -17,23 +17,84 @@
...
+            }
+
             return null;
         }
+
+        public void SaveSettings(ServerExclusions settings)
+        {
+            try
+            {
+                if (bindingConfiguration.Mode == SonarLintMode.Standalone)
```
rita-sonarsource, line 77: Is this being part of the connected mode binding process? If so, how would this be standalone when we reach this point?

src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
ugras-ergun-sonarsource: `The read takes the bindingConfiguration only to write logging`
What do you mean by this? We need this to both finding the mode and finding the file path. File path is calculated based on bindingconfiguration.

src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
duncanp-sonar: I think coding defensively is the wrong approach here; it means there is something wrong elsewhere in our logic.
An assertion would be better.

ugras-ergun-sonarsource: I understand but in this case, we would have the possibility to write a file to disk unnecessarily. Do we really want that?

duncanp-sonar: Yes. We should not be coding round/hiding errors in our code, particularly new code. We should be making the failures obvious so we find them during development/dogfooding and fix them.

PR #3048, opened 2022-05-31
Fixup CFamily legacy rule keys

### LINKED ISSUE #3047: Add component to map old-new keys
### Description

Translate legacy rule keys to the new format.

Rule key mapping: https://github.com/SonarSource/sonar-cpp/blob/c51c7ccb23e32f587a543a2e4b08f10e92daf2a7/sonar-cfamily-plugin/src/main/java/com/sonar/cpp/plugin/AbstractRulesDefinition.java#L35

duncanp-sonar:
Fixes #3047

### DIFF
--- a/src/CFamily/Rules/RulesConfigFixup.cs
+++ b/src/CFamily/Rules/RulesConfigFixup.cs
@@ -145,27 +145,54 @@ namespace SonarLint.VisualStudio.CFamily.Rules
         /// </summary>
         public RulesSettings Apply(RulesSettings input)
         {
-            foreach (var inputKey in input.Rules.Keys.ToArray())
+            /*
+
+            We're making a shallow copy of the list of rules. If we modify the original list, any exclusions we 
+            add could end up be saved in the user settings.json file(if that is where the custom rules
+            came from).
+            
+            Modifying the saved user settings.json would be a problem in the following scenario:
+                * the file has a legacy key settings e.g.cpp:C99CommentUsage
+                * the user has multiple VS instances with "old" and "new" SonarLint instances installed
+            e.g.they still have an instance oF VS2015 / 2017 they need to use
+                
+            In that case, we don't want to update the legacy keys in the settings file since it would
+            re - enable the rules in the "old" version.
+
+            However, _not_ updating the legacy keys in the file has a different issue: the file could
+            contain both old and new keys e.g.
+                * settings file has legacy rule key e.g.cpp:C99CommentUsage(set to "On")
+                * user disables the corresponding "new" rule S787.
+            In that case, we'll warn in the output window that the legacy setting is being ignored.
+
+             */
+            
+            var modifiedSettings = new RulesSettings
+            {
+                Rules = new Dictionary<string, RuleConfig>(input.Rules, input.Rules.Comparer)
+            };
+
+            foreach (var inputKey in modifiedSettings.Rules.Keys.ToArray())
             {
                 if (fullLegacyToNewKeyMap.TryGetValue(inputKey, out var newKey))
                 {
-                    var inputConfig = input.Rules[inputKey];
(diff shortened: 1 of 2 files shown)


src/CFamily.UnitTests/Rules/DynamicCFamilyRulesConfigTests.cs
```diff
@@ -68,8 +69,7 @@ public void NullOrEmptyRulesSettings_DefaultsUsed()
             // Act
             using (new AssertIgnoreScope())
             {
-                var testSubject =
-                    new DynamicCFamilyRulesConfig(defaultConfig, settings, new TestLogger(), NoExcludedRules);
+                var testSubject = CreateTestSubject(defaultConfig, settings);
```
duncanp-sonar, line 72: Miscellaneous: changed the existing tests to use the new `CreateTestSubject` method with default parameters.

duncanp-sonar:
These changes mean we will never pass legacy rule keys to the subprocess. However, the version of the subprocess we are currently embedding doesn't know about the new rule keys, so these changes will only work properly once the subprocess has been upgraded too.

src/CFamily/Rules/RulesConfigFixup.cs
```diff
@@ -0,0 +1,198 @@
...
+        public RulesConfigFixup(ILogger logger) => this.logger = logger ?? throw new System.ArgumentNullException(nameof(logger));
+
+        /// <summary>
+        /// Translates any legacy rule keys in the input to new Sxxx rule keys
+        /// </summary>
+        public RulesSettings Apply(RulesSettings input)
+        {
+            /*
+
+            We're making a shallow copy of the list of rules. If we modify the original list, any exclusions we
```
duncanp-sonar, line 150: The first part of this comment giving the reasoning is taken from here: https://github.com/SonarSource/sonarlint-visualstudio/blob/63a9400ef82f8e62298b8ab2eb433723dc17a8f8/src/Core/CFamily/DynamicCFamilyRulesConfig.cs#L201

I've added more details about the specific legacy key scenarios below.

src/CFamily/Rules/Resources.resx
```diff
@@ -117,12 +117,16 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="CFamily_DuplicateLegacyAndNewRuleKey" xml:space="preserve">
+    <value>[CFamily] Rule settings contain entries for both legacy rule key and new rule keys. The legacy value will be ignored. Legacy key: {0}, new key: {1}</value>
```
rita-sonarsource, line 121: I'm not sure how useful this would be to the end user. If they're using connected mode, then this might confuse them ever further as there is nothing they can do besides upgrading the SQ instance.
I wonder if we should have a wiki page for this or if it's an overkill. Do we have some metrics from cfamily how many users would be impacted by this change?

rita-sonarsource (approved):
LGTM

src/CFamily/Rules/Resources.resx
duncanp-sonar: Connected mode: true, it might be confusing if they see the output message. A more noticeable source of confusion is likely to be the Error List, since users connected to "old" SonarQube instances won't see e.g. `cpp:C99CommentUsage` as the rule key, even though that is what they will see in their Quality Profile. We could add a wiki page later if we get community questions about it.

I thought it was better to output something rather than nothing, even though I don't think many users will be affected, and almost none of those who are will notice this in the output window. It was more to help us in answering forum questions if users don't understand what is happening.

Metrics: I haven't checked.

PR #3049, opened 2022-05-31
Dp/cfamily6 4

duncanp-sonar:


src/Integration.Vsix.UnitTests/CFamily/CFamilyEmbeddedSonarWayRulesTests.cs
```diff
@@ -73,15 +76,15 @@ public void Read_Active_Rules()
         [TestMethod]
         public void Read_Rules_Params()
         {
-            rulesMetadataCache.GetRulesConfiguration("cpp").RulesParameters.TryGetValue("ClassComplexity", out var parameters);
```
duncanp-sonar, line 76: Previously this test and the following one happened to use a rule with a legacy rule key. Happily, these tests failed when the analyzer was updated. They pass with the new rule key.

ugras-ergun-sonarsource (approved):

PR #3060, opened 2022-06-13
Add inclusions/exclusions logic

rita-sonarsource:
Fixes [#3061](https://github.com/SonarSource/sonarlint-visualstudio/issues/3061)

Code coverage is low due to an empty class that would be implemented in a separate task.

### DIFF
--- a/src/Core/Analysis/IExclusionSettingsFileStorage.cs
+++ b/src/Core/Analysis/IExclusionSettingsFileStorage.cs
@@ -1,0 +1,30 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.Core.Analysis
+{
+    public interface IExclusionSettingsFileStorage
+    {
+        void SaveSettings(ServerExclusions settings);
+        ServerExclusions GetSettings();
+    }
+}
--- a/src/Integration.UnitTests/Exclusions/AnalyzableFileIndicatorTests.cs
+++ b/src/Integration.UnitTests/Exclusions/AnalyzableFileIndicatorTests.cs
@@ -56,24 +56,33 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Exclusions
         }
 
         [TestMethod]
-        [DataRow(true, false)]
-        [DataRow(false, true)]
-        public void ShouldAnalyze_NoInclusions_HasExclusions_ReturnsIfExcluded(bool exclusionsApply, bool expectedResult)
+        [DataRow(true, true, false)]
+        [DataRow(true, false, false)]
+        [DataRow(false, true, false)]
+        [DataRow(false, false, true)]
+        public void ShouldAnalyze_NoInclusions_HasExclusions_ReturnsIfExcluded(
+            bool projectExclusionsApply,
+            bool globalExclusionsApply,
+            bool expectedResult)
         {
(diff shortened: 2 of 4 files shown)


src/Core/Analysis/IExclusionSettingsFileStorage.cs
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.Core.Analysis
+{
+    public interface IExclusionSettingsFileStorage
```
rita-sonarsource, line 25: Based on Ugras' PR

src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.Integration.Exclusions
+{
+    [Export(typeof(IExclusionSettingsFileStorage))]
+    internal class ExclusionSettingsFileStorage : IExclusionSettingsFileStorage
```
rita-sonarsource, line 28: Just dummy implementation for now

src/Integration/Exclusions/AnalyzableFileIndicator.cs
```diff
@@ -27,9 +29,49 @@ namespace SonarLint.VisualStudio.Integration.Exclusions
...
+
+            bool IsExcludedInProjectOrGlobally() =>
+                IsExcluded(serverExclusions.Exclusions) ||
+                IsExcluded(serverExclusions.GlobalExclusions);
+
+            bool IsExcluded(string[] exclusions) =>
+                exclusions != null &&
+                exclusions.Any(IsMatch);
+
+            bool IsMatch(string pattern) => globPatternMatcher.IsMatch(pattern, filePath);
```
rita-sonarsource, line 74: At the moment I'm comparing the pattern as-is to the file path, disregarding the root directory from which we should start. I added a separate task to investigate this.

src/Integration/Exclusions/AnalyzableFileIndicator.cs
```diff
@@ -27,9 +29,49 @@ namespace SonarLint.VisualStudio.Integration.Exclusions
...
+                return true;
+            }
+
+            var shouldAnalyze = IsIncluded() && !IsExcludedInProjectOrGlobally();
+
+            return shouldAnalyze;
+
+            bool IsIncluded() =>
+                serverExclusions.Inclusions == null ||
+                serverExclusions.Inclusions.Length == 0 ||
```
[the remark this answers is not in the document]
rita-sonarsource: For inclusions, if the array is empty, the item is considered included. Whereas if the array is not empty, then the item might be excluded.
In exclusions, an empty array and a non-empty array behave the same. Hence why there is no need to check for empty.

src/Integration/Exclusions/AnalyzableFileIndicator.cs
```diff
@@ -27,9 +29,49 @@ namespace SonarLint.VisualStudio.Integration.Exclusions
...
+                serverExclusions.Inclusions.Any(IsMatch);
+
+            bool IsExcludedInProjectOrGlobally() =>
```
duncanp-sonar, line 66: Do the nested functions improve the readability?

rita-sonarsource: You tell me :D

src/Integration.UnitTests/Exclusions/AnalyzableFileIndicatorTests.cs
```diff
@@ -18,19 +18,198 @@
...
+
+            var testSubject = CreateTestSubject(exclusionConfig, patternMatcher.Object);
+
+            var result = testSubject.ShouldAnalyze(TestedFilePath);
+
+            result.Should().Be(expectedResult);
+        }
+
+        [TestMethod]
+
```
duncanp-sonar, line 147: FYI extra blank line in after the attribute in this test and the next one.

src/Integration/Exclusions/AnalyzableFileIndicator.cs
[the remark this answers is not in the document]
duncanp-sonar: Ok, makes sense, but it's not obvious. An explicit comment would help.

src/Core/Analysis/IExclusionSettingsFileStorage.cs
duncanp-sonar: @ugras-ergun-sonarsource do you mean you'll push the change to 'branch-exclusions`, or you'll create a PR?

ugras-ergun-sonarsource: I have a pr already. I had made some changes to signatures locally but I've reverted it back and in the confusion, I have created noise you can disregard my comment.

PR #3064, opened 2022-06-14
Fetch the server settings during binding and save to disk 

### LINKED ISSUE #3034: Fetch the server settings during binding and save to disk

ugras-ergun-sonarsource:
fixes #3034


src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -59,17 +62,24 @@ internal class BindingProcessImpl : IBindingProcess
             IBindingConfigProvider bindingConfigProvider,
             SonarLintMode bindingMode,
             bool isFirstBinding = false,
-            IThreadHandling threadHandling = null /* overrideable for testing */)
+            IThreadHandling threadHandling = null /* overrideable for testing */,
+            IFileSystem fileSystem = null,
+            IExclusionSettingsStorage exclusionSettingsStorage = null
```
rita-sonarsource, line 67: If we are always receiving it from the caller, it shouldn't be null. The params above are nullable because they have a default value passed in the other CTOR

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -59,17 +62,24 @@ internal class BindingProcessImpl : IBindingProcess
...
             this.bindingArgs = bindingArgs ?? throw new ArgumentNullException(nameof(bindingArgs));
             this.solutionBindingOperation = solutionBindingOperation ?? throw new ArgumentNullException(nameof(solutionBindingOperation));
             this.NuGetBindingOperation = nugetBindingOperation ?? throw new ArgumentNullException(nameof(nugetBindingOperation));
             this.unboundProjectFinder = unboundProjectFinder ?? throw new ArgumentNullException(nameof(unboundProjectFinder));
             this.bindingConfigProvider = bindingConfigProvider ?? throw new ArgumentNullException(nameof(bindingConfigProvider));
+            this.exclusionSettingsStorage = exclusionSettingsStorage ?? throw new ArgumentNullException(nameof(exclusionSettingsStorage));
             this.bindingMode = bindingMode;
             projectToLanguageMapper = host.GetMefService<IProjectToLanguageMapper>();
             this.threadHandling = threadHandling ?? new ThreadHandling();
+            this.fileSystem = fileSystem ?? new FileSystem();
```
rita-sonarsource, line 80: Was this a bug?

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -196,6 +206,27 @@ public void PrepareToInstallPackages()
             this.NuGetBindingOperation.PrepareOnUIThread();
         }
 
+        public async Task<bool> SaveServerExclusionsAsync(CancellationToken cancellationToken)
+        {
+            try
+            {
+                var configurationProviderService = host.GetService<IConfigurationProviderService>();
```
rita-sonarsource, line 213: why not use `bindingArgs.ProjectKey` ?

src/Integration/Binding/BindingWorkflow.cs
```diff
@@ -135,6 +136,9 @@ private ProgressStepDefinition[] CreateWorkflowSteps(IProgressController control
                 new ProgressStepDefinition(null, StepAttributes.Hidden | StepAttributes.Indeterminate,
                         (token, notifications) => this.FinishSolutionBindingOnUIThread(controller, token)),
 
+                new ProgressStepDefinition(Strings.BindingProjectsDisplayMessage, StepAttributes.BackgroundThread | StepAttributes.Indeterminate,
```
rita-sonarsource, line 139: What does Indeterminate flag do?

src/Integration/Binding/BindingWorkflow.cs
```diff
@@ -135,6 +136,9 @@ private ProgressStepDefinition[] CreateWorkflowSteps(IProgressController control
...
+                        (token, notifications) => this.SaveServerExclusionsAsync(controller, notifications, token).GetAwaiter().GetResult()),
```
rita-sonarsource, line 140: should we block this?

src/Integration/Resources/Strings.resx
```diff
@@ -772,4 +772,16 @@ This is the general format to use when writing errors to output window or when t
   <data name="ExclusionGetError" xml:space="preserve">
     <value>Error loading server settings. Analysis settings from the server (e.g. inclusions/exclusions) will not be applied in the IDE. Error: {0}</value>
   </data>
+  <data name="ExclusionSaveError" xml:space="preserve">
+    <value>Error writing settings for project. File exclusions on the server may not be excluded in the IDE. Error: {0}</value>
```
rita-sonarsource, line 776: "may not be excluded in the IDE" -- is this information accurate? We are aborting the whole binding process, so for it for sure would not be applied.

src/Integration/Resources/Strings.resx
```diff
@@ -772,4 +772,16 @@ This is the general format to use when writing errors to output window or when t
...
+  </data>
+  <data name="ExclusionOnStandaloneNotSupported" xml:space="preserve">
+    <value>File exclusions are not supported in Standalone mode. To use file exclusions you need connected mode. </value>
```
rita-sonarsource, line 779: Where is this used?

src/Integration/Resources/Strings.resx
```diff
@@ -772,4 +772,16 @@ This is the general format to use when writing errors to output window or when t
...
+  </data>
+  <data name="SaveExclusionsFailed" xml:space="preserve">
+    <value>Failed to save exclusion file to the disk. Server Side exclusions may fail. Error: {0}</value>
+  </data>
+  <data name="SaveServerExclusionsMessage" xml:space="preserve">
```
rita-sonarsource, line 784: Where is this used?

src/Integration.UnitTests/Binding/BindingWorkflowTests.cs
```diff
@@ -102,6 +102,46 @@ private void SetDownloadQPResult(bool result)
...
+            var notifications = new ConfigurableProgressStepExecutionEvents();
+
+            SetSaveServerExclusionsResult(true);
+
+            // Act
+            await testSubject.SaveServerExclusionsAsync(controller, notifications, CancellationToken.None)
+                .ConfigureAwait(false);
+
+            // Assert
+            controller.NumberOfAbortRequests.Should().Be(0);
```
rita-sonarsource, line 119: Perhaps it's also worth adding Verify that `SaveServerExclusionsAsync` was indeed called? same in the other test

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -55,6 +56,8 @@ public class BindingProcessImplTests
         private ConfigurableSolutionRuleSetsInformationProvider ruleSetsInformationProvider;
         private ConfigurableHost host;
 
+        private const string objectJson = "{\"Exclusions\":[\"exclusion\"],\"GlobalExclusions\":[\"globalExclusion\"],\"Inclusions\":[\"inclusion1\",\"inclusion2\"],\"GlobalInclusions\":[\"globalInclusion\"]}";
```
rita-sonarsource, line 59: Where is this used?

rita-sonarsource (commented):
@ugras-ergun-sonarsource I think there are some leftovers from merge conflicts. Some things were not in use.

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -196,6 +206,27 @@ public void PrepareToInstallPackages()
...
+                var sonarQubeService = host.SonarQubeService;
+                var projectKey = configurationProviderService.GetConfiguration().Project.ProjectKey;
+
+                var exclusions = await sonarQubeService.GetServerExclusions(projectKey, cancellationToken);
+
+                exclusionSettingsStorage.SaveSettings(exclusions);
+            }
+            catch (Exception ex)
```
duncanp-sonar, line 221: Don't catch critical exceptions.

src/Integration/Resources/Strings.resx
```diff
@@ -772,4 +772,16 @@ This is the general format to use when writing errors to output window or when t
   <data name="ExclusionGetError" xml:space="preserve">
     <value>Error loading server settings. Analysis settings from the server (e.g. inclusions/exclusions) will not be applied in the IDE. Error: {0}</value>
   </data>
+  <data name="ExclusionSaveError" xml:space="preserve">
+    <value>Error writing settings for project. File exclusions on the server may not be excluded in the IDE. Error: {0}</value>
+  </data>
+  <data name="ExclusionOnStandaloneNotSupported" xml:space="preserve">
```
duncanp-sonar, line 778: Where is this string used? (the PR is changing the binding code, and I can't see a reason why the binding code would ever need such a string).

src/Integration/Binding/BindingWorkflow.cs
```diff
@@ -185,6 +189,21 @@ private ProgressStepDefinition[] CreateWorkflowSteps(IProgressController control
...
         }
 
+        internal /*for testing purposes*/ async Task SaveServerExclusionsAsync(
+            IProgressController controller, IProgressStepExecutionEvents notificationEvents,
+            CancellationToken cancellationToken)
+        {
+            Debug.Assert(controller != null);
+            Debug.Assert(notificationEvents != null);
+
+            notificationEvents.ProgressChanged(Strings.SaveServerExclusionsMessage);
```
duncanp-sonar, line 199: The message being logged is 
>No file was found under &apos;{0}&apos;.

... but it looks like the message is being logged unconditionally i.e. in the case of the binding being updated, isn't it possible that the file will already exist?

src/Integration/Resources/Strings.resx
```diff
@@ -772,4 +772,16 @@ This is the general format to use when writing errors to output window or when t
   <data name="ExclusionGetError" xml:space="preserve">
     <value>Error loading server settings. Analysis settings from the server (e.g. inclusions/exclusions) will not be applied in the IDE. Error: {0}</value>
   </data>
+  <data name="ExclusionSaveError" xml:space="preserve">
```
duncanp-sonar, line 775: Is this string used?

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -196,6 +206,27 @@ public void PrepareToInstallPackages()
...
+            {
+                host.Logger.WriteLine(string.Format(Strings.SaveExclusionsFailed, ex.Message));
```
duncanp-sonar, line 223: The message being logged is
> Failed to save exclusion file to the disk. Server Side exclusions may fail. Error: {0}.

... and it will be logged if the webservice call failed, which would be confusing.

Doesn't the exclusion settings class already log an exception detailing why the save failed? Either way, this message should perhaps be more generic e.g. "Failed to retrieve/process server settings. Some analysis settings will not be applied (e.g. file inclusions/exclusion)."

src/Integration/Resources/Strings.resx
duncanp-sonar, line 784: Is this used?

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
duncanp-sonar, line 59: Is this string used?

src/Integration/Binding/BindingWorkflow.cs
duncanp-sonar: It's means the progress is indeterminate i.e. can't be reported as a percentage/step x of y. So the UI will show something like a spinner to indicate work is happening, but no indication of how much there is left to do.

duncanp-sonar (commented):
See comments on log messages.

src/Integration/Binding/BindingWorkflow.cs
ugras-ergun-sonarsource: if we don't we can't fail the workflow process.

src/Integration/Binding/BindingProcessImpl.cs
ugras-ergun-sonarsource: If we remove the catch this comment should be removed as well there are no scenarios in our code that will fail without an exception.

src/Integration/Resources/Strings.resx
ugras-ergun-sonarsource: This was used in the Get which is a valid scenario. While merging with Rita's changes it became orphan.

src/Integration/Binding/BindingProcessImpl.cs
duncanp-sonar: I'm not suggesting removing this catch; it's at the outer level, and there are other things that could fail e.g. the webservice call.

I'm asking whether the output message we are logging overall will make sense to the user.

ugras-ergun-sonarsource: As I understand you suggested removing the catch in the previous comment.

duncanp-sonar: That wasn't what I was suggesting.

src/Integration/Resources/Strings.resx
ugras-ergun-sonarsource: It was supposed to be called from  BindingWorkflow.

src/Integration/Binding/BindingWorkflow.cs
rita-sonarsource: But other steps, like `PrepareSolutionBinding` don't do this. 
Inside SaveServerExclusionsAsync we have a call to AbortWorkflow, so it should be able to handle exceptions.

What would happen if you were to remove this?

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -90,35 +91,123 @@ public void TestInitialize()
...
+        public async Task SaveServerExclusionsAsync_HasError_ReturnsFalse()
+        {
+            var bindingArgs = CreateBindCommandArgs(projectKey: "projectKey");
+
+            CreateConfigurationProviderService("C:\\SolutionPath");
+
+            var sonarQubeService = new Mock<ISonarQubeService>();
+            sonarQubeService.Setup(s => s.GetServerExclusions("projectKey", It.IsAny<CancellationToken>())).Throws(new Exception("Expected Error"));
+
+            var exclusionSettingsStorage = new Mock<IExclusionSettingsStorage>();
```
rita-sonarsource, line 162: exclusionSettingsStorage is not being verified so it can be removed to make the test shorter. Same for the test below.

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -90,35 +91,123 @@ public void TestInitialize()
...
+            await testSubject.SaveServerExclusionsAsync(CancellationToken.None);
+
+            exclusionSettingsStorage.Verify(fs => fs.SaveSettings(settings), Times.Once);
+            this.outputWindowPane.AssertOutputStrings(0);
+        }
+
+        [TestMethod]
+        public async Task SaveServerExclusionsAsync_HasError_ReturnsFalse()
+        {
+            var bindingArgs = CreateBindCommandArgs(projectKey: "projectKey");
```
rita-sonarsource, line 155: Can this be a default param in CreateTestSubject?

ugras-ergun-sonarsource: It's a default parameter used in other tests already. I don't want to change it because I wouldn't like it to affect existing tests. So I need to create it in new tests.

PR #3065, opened 2022-06-15
Read exclusions from disk

rita-sonarsource:
Part of [#3057](https://github.com/SonarSource/sonarlint-visualstudio/issues/3057)

Example json file (`sonar.settings.json`):
```
{
	"Exclusions" : [],
	"GlobalExclusions" : [],
	"Inclusions" : ["**/*"]
}
```

User messages:
image
image

### DIFF
--- a/src/Integration.UnitTests/Exclusions/ExclusionSettingsFileStorageTests.cs
+++ b/src/Integration.UnitTests/Exclusions/ExclusionSettingsFileStorageTests.cs
@@ -128,6 +128,20 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Exclusions
             file.VerifyNoOtherCalls();
         }
 
+        [TestMethod]
+        public void GetSettings_CriticalException_NotSuppressed()
+        {
+            var file = new Mock<IFile>();
+            file.Setup(f => f.Exists(ExpectedExclusionsFilePath)).Returns(true);
+            file.Setup(f => f.ReadAllText(ExpectedExclusionsFilePath)).Throws(new StackOverflowException());
+
+            var testSubject = CreateTestSubject(file.Object);
+
+            Action act = () => testSubject.GetSettings();
+
+            act.Should().Throw<StackOverflowException>();
+        }
+
         private ExclusionSettingsFileStorage CreateTestSubject(IFile file, BindingConfiguration bindingConfiguration = null)
         {
             var fileSystem = CreateFileSystem(file);
--- a/src/Integration.UnitTests/Exclusions/ExclusionSettingsFileStorageTests.cs
+++ b/src/Integration.UnitTests/Exclusions/ExclusionSettingsFileStorageTests.cs
@@ -1,0 +1,157 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.IO.Abstractions;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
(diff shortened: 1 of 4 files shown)


src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
```diff
@@ -18,22 +18,80 @@
...
         {
         }
 
         public ServerExclusions GetSettings()
         {
+            try
+            {
+                var bindingConfiguration = bindingConfigProvider.GetConfiguration();
+
+                if (bindingConfiguration.Mode == SonarLintMode.Standalone)
```
rita-sonarsource, line 67: I ended up leaving the standalone check. The AnalyzableIndicator shouldn't do this check, so it's ok that it's here.

src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
```diff
@@ -18,22 +18,80 @@
...
+using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Integration.Helpers;
+using SonarLint.VisualStudio.Integration.Resources;
 using SonarQube.Client.Models;
+using static System.String;
 
 namespace SonarLint.VisualStudio.Integration.Exclusions
 {
     [Export(typeof(IExclusionSettingsFileStorage))]
     internal class ExclusionSettingsFileStorage : IExclusionSettingsFileStorage
```
rita-sonarsource, line 36: I will rename the file in a separate PR

src/Integration.UnitTests/Exclusions/ExclusionSettingsFileStorageTests.cs
```diff
@@ -0,0 +1,157 @@
...
+using Moq;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Integration.Exclusions;
+using SonarLint.VisualStudio.Integration.NewConnectedMode;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.Exclusions
+{
+    [TestClass]
+    public class ExclusionSettingsFileStorageTests
```
rita-sonarsource, line 34: Based on Ugras' PR with slight changes

src/Integration/Resources/Strings.resx
```diff
@@ -766,4 +766,10 @@ This is the general format to use when writing errors to output window or when t
     <value>Ruleset file does not exist: {0}</value>
     <comment>Output window message for a ruleset that does not exist on disc</comment>
   </data>
+  <data name="ExclusionFileNotFound" xml:space="preserve">
+    <value>No file was not found under '{0}'</value>
+  </data>
+  <data name="ExclusionGetError" xml:space="preserve">
+    <value>Error loading exclusion settings. File exclusions on the server will not be excluded in the IDE. Error: {0}</value>
```
duncanp-sonar, line 773: The error message only mentions exclusions. Also, the file name for generic `settings`, not just inclusion\exclusions. Perhaps the message should be something more general e.g.

> Error loading server settings. Analysis settings from the server (e.g. inclusions/exclusions) will not be applied in the IDE. Error: {0}

src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
duncanp-sonar: I still think a `Debug.Fail` would be useful in this case so that we have a chance of detecting the issue during development.

src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
```diff
@@ -18,22 +18,80 @@
...
+                {
+                    logger.WriteLine(Strings.ExclusionGetError, Format(Strings.ExclusionFileNotFound, exclusionsFilePath));
+                    return null;
+                }
+
+                var fileContent = fileSystem.File.ReadAllText(exclusionsFilePath);
+                
+                return JsonConvert.DeserializeObject<ServerExclusions>(fileContent);
+            }
+            catch (Exception ex)
```
duncanp-sonar, line 85: We don't normally suppress critical exceptions.

We have quite a few "critical exceptions are not caught" tests, but I'm not convinced those tests add enough to justify writing them.

duncanp-sonar (commented):
Looks good. The main thing is the comment about the user-facing message.

src/Integration/Resources/Strings.resx
```diff
@@ -766,4 +766,10 @@ This is the general format to use when writing errors to output window or when t
     <value>Ruleset file does not exist: {0}</value>
     <comment>Output window message for a ruleset that does not exist on disc</comment>
   </data>
+  <data name="ExclusionFileNotFound" xml:space="preserve">
+    <value>No file was not found under '{0}'</value>
```
duncanp-sonar, line 770: "**No** file was **not** found" ! :-)

src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
rita-sonarsource: During Read it won't be a mistake, it's a real use case.
AnalysisController calls AnalyzableIndicator, that calls this class. None of them know anything about connected mode.  So they would call this even when they're in standalone.

PR #3067, opened 2022-06-15
Save exclusions to disk

rita-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3057


src/Integration/Exclusions/ExclusionSettingsFileStorage.cs
```diff
@@ -56,6 +56,18 @@ public ExclusionSettingsFileStorage(IConfigurationProvider bindingConfigProvider
 
         public void SaveSettings(ServerExclusions settings)
         {
+            var bindingConfiguration = bindingConfigProvider.GetConfiguration();
```
rita-sonarsource, line 59: Based on Ugra's PR with one difference: exceptions are thrown to the caller. This would allow the binding process to abort.

PR #3072, opened 2022-06-17
Use IUnboundSolutionChecker in BindingChecker

rita-sonarsource:
Part of [#3069](https://github.com/SonarSource/sonarlint-visualstudio/issues/3069)

### DIFF
--- a/src/Integration.UnitTests/LocalServices/BindingCheckerTests.cs
+++ b/src/Integration.UnitTests/LocalServices/BindingCheckerTests.cs
@@ -31,22 +31,26 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.LocalServices
         [TestMethod]
         public void IsBindingUpdateRequired_SolutionIsUnbound_True()
         {
-            var testSubject = CreateTestSubject(isSolutionBound: false);
+            var unboundSolutionChecker = CreateUnboundSolutionChecker(isSolutionBound: false);
+            var testSubject = CreateTestSubject(unboundSolutionChecker.Object);
 
             var result = testSubject.IsBindingUpdateRequired();
 
             result.Should().BeTrue();
+            unboundSolutionChecker.VerifyAll();
         }
 
         [TestMethod]
         public void IsBindingUpdateRequired_SolutionIsUnbound_ProjectBindingIsNotChecked()
         {
             var projectBinder = new Mock<IUnboundProjectFinder>();
-            var testSubject = CreateTestSubject(isSolutionBound: false, projectBinder.Object);
+            var unboundSolutionChecker = CreateUnboundSolutionChecker(isSolutionBound: false);
+            var testSubject = CreateTestSubject(unboundSolutionChecker.Object, projectBinder.Object);
 
             testSubject.IsBindingUpdateRequired();
 
             projectBinder.Invocations.Should().BeEmpty();
+            unboundSolutionChecker.VerifyAll();
         }
 
         [TestMethod]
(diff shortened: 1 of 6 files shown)


src/Integration/LocalServices/IUnboundSolutionChecker.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    internal interface IUnboundSolutionChecker : IBindingChecker
+    {
+    }
+
+    internal class UnboundSolutionChecker : IUnboundSolutionChecker
+    {
+        public bool IsBindingUpdateRequired()
+        {
+            return false;
```
rita-sonarsource, line 31: Implementation in next PR

src/Integration/LocalServices/IUnboundSolutionChecker.cs
```diff
@@ -0,0 +1,35 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Integration
+{
+    internal interface IUnboundSolutionChecker : IBindingChecker
```
rita-sonarsource, line 23: Decided to reuse the interface for no particular reason other than it's exactly the method I wanted to add 🤷‍♀️

duncanp-sonar: Hhmm, that's confusing. It might be taken to suggest that the `ErrorListInfoBarController` could be validly be handed an `UnboundSolutionChecker`, which isn't really what we want.

I think it would be simpler to keep the interfaces completely separate for now, so that the compiler won't let us pass the "wrong" type to the `ErrorListInfoBarController`.

Having identical interfaces seems like a design smell that perhaps the responsibilities aren't quite right yet.

rita-sonarsource: Good point 👍

src/Integration/Resources/Strings.resx
```diff
@@ -772,4 +772,8 @@ This is the general format to use when writing errors to output window or when t
   <data name="ExclusionGetError" xml:space="preserve">
     <value>Error loading server settings. Analysis settings from the server (e.g. inclusions/exclusions) will not be applied in the IDE. Error: {0}</value>
   </data>
+  <data name="SonarLintFoundUnboundSolution" xml:space="preserve">
+    <value>Some binding configuration is missing. Please rebind the solution.</value>
```
duncanp-sonar, line 776: Would it help to have a consistent prefix for binding message? (and are we using a prefix for binding anywhere else?)

rita-sonarsource: No prefix appears for the other messages that are used by the ErrorListInfoBarController :/

src/Integration.UnitTests/LocalServices/BindingCheckerTests.cs
```diff
@@ -29,7 +29,28 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.LocalServices
...
     {
         [TestMethod]
-        public void IsBindingUpdateRequired_NoUnboundProjects_False()
+        public void IsBindingUpdateRequired_SolutionIsUnbound_True()
+        {
+            var testSubject = CreateTestSubject(isSolutionBound: false);
+
+            var result = testSubject.IsBindingUpdateRequired();
+
+            result.Should().BeTrue();
```
duncanp-sonar, line 38: Perhaps check that the unbound solution checker is called in this test and the next one?

duncanp-sonar (commented):
A couple of minor comments.

PR #3073, opened 2022-06-17
Add bool SettingsExist

rita-sonarsource:
Part of https://github.com/SonarSource/sonarlint-visualstudio/issues/3069

### DIFF
--- a/src/Integration.UnitTests/Exclusions/ExclusionSettingsStorageTests.cs
+++ b/src/Integration.UnitTests/Exclusions/ExclusionSettingsStorageTests.cs
@@ -202,18 +202,16 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Exclusions
         }
 
         [TestMethod]
-        public void SettingsExist_StandAloneMode_ThrowsInvalidOperationException()
+        public void SettingsExist_StandAloneMode_False()
         {
             var file = new Mock<IFile>();
             var bindingConfiguration = BindingConfiguration.Standalone;
 
             var testSubject = CreateTestSubject(file.Object, bindingConfiguration);
 
-            using var scope = new AssertIgnoreScope();
+            var result = testSubject.SettingsExist();
 
-            Action act = () => testSubject.SettingsExist();
-
-            act.Should().Throw<InvalidOperationException>();
+            result.Should().BeFalse();
             file.Invocations.Count.Should().Be(0);
         }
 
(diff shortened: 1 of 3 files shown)


src/Integration.UnitTests/Exclusions/ExclusionSettingsStorageTests.cs
```diff
@@ -201,6 +201,48 @@ public void SaveSettings_NoError_SavesSettings()
...
+            var file = new Mock<IFile>();
+            file.Setup(f => f.Exists(ExpectedExclusionsFilePath)).Returns(true);
+
+            var testSubject = CreateTestSubject(file.Object);
+
+            var result = testSubject.SettingsExist();
+
+            result.Should().BeTrue();
+        }
+
```
rita-sonarsource, line 245: I'm debating if it's worth adding 3 additional tests:
1. Failed to access the file (non critical ex)
2. Failed to access (critical)
3. File exists but content inside is invalid

I think the right approach for all 3 would be to delete the existing file from disk (if we can) and return that the settings don't exist. But I don't think that's what we're doing in other cases (i.e. cfamily). Returning that the file doesn't exist when in fact we couldn't access the file feels wrong to me. So I think it's better not to handle anything here and let the caller (UnboundSolutionChecker) handle the exceptions.

src/Integration/Exclusions/ExclusionSettingsStorage.cs
```diff
@@ -105,7 +101,27 @@ public ServerExclusions GetSettings()
...
+
         private static string GetFilePath(BindingConfiguration bindingConfiguration) => 
             Path.Combine(bindingConfiguration.BindingConfigDirectory, "sonar.settings.json");
+
+        private static void VerifyConnectedMode(BindingConfiguration bindingConfiguration)
+        {
+            if (bindingConfiguration.Mode == SonarLintMode.Standalone)
+            {
+                Debug.Fail("Not expecting to be called in Standalone mode");
+                throw new InvalidOperationException("Cannot save exclusions in Standalone mode.");
```
duncanp-sonar, line 123: This will also be thrown if `SettingsExists` is called in standalone mode. 
Is that is what is intended? If so, it might be worth tweaking the message to be more generic.

It seems slightly inconsistent that `LoadSettings` is safe to call in standalone mode, but `SettingsExist` isn't.

src/Integration.UnitTests/Exclusions/ExclusionSettingsStorageTests.cs
duncanp-sonar: We haven't had problems using this approach for the CFamily, so far I'd start simple for now.

duncanp-sonar (commented):
Couple of comments inline.

src/Integration/Exclusions/ExclusionSettingsStorage.cs
rita-sonarsource: I ended up changing it to behave same as Get: it will return false in standalone.

PR #3074, opened 2022-06-17
Implement UnboundSolutionChecker

rita-sonarsource:
Fixes [#3069](https://github.com/SonarSource/sonarlint-visualstudio/issues/3069)

### DIFF
--- a/src/Integration/Resources/Strings.resx
+++ b/src/Integration/Resources/Strings.resx
@@ -777,6 +777,6 @@ This is the general format to use when writing errors to output window or when t
     <comment>Output window message</comment>
   </data>
   <data name="BindingUpdateFailedToCheckSettings" xml:space="preserve">
-    <value>[Binding check] Failed to check bound project settings file: {0}</value>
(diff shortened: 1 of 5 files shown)


rita-sonarsource:
This message is not exactly right for this scenario, I'm not sure if it's worth changing:

image

And the message in the output window:
image

src/Integration/Resources/Strings.resx
```diff
@@ -776,4 +776,7 @@ This is the general format to use when writing errors to output window or when t
     <value>Some binding configuration is missing. Please rebind the solution.</value>
     <comment>Output window message</comment>
   </data>
+  <data name="BindingUpdateFailedToCheckSettings" xml:space="preserve">
+    <value>[Binding check] Failed to check bound project settings file: {0}</value>
```
duncanp-sonar, line 780: Perhaps `Error checking bound project settings file: ...` rather than `Failed to check...`

ugras-ergun-sonarsource (approved):

PR #3079, opened 2022-06-20
Windows paths does not work correctly on Exclusions 

ugras-ergun-sonarsource:


src/Integration/Exclusions/AnalyzableFileIndicator.cs
```diff
@@ -87,6 +87,10 @@ public bool ShouldAnalyze(string filePath)
             exclusions != null &&
             exclusions.Any(x => IsMatch(x, filePath));
 
-        private bool IsMatch(string pattern, string filePath) => globPatternMatcher.IsMatch(pattern, filePath);
+        private bool IsMatch(string pattern, string filePath)
+        {
```
rita-sonarsource, line 91: IsMatch would be called multiple times, why do it each time? Not to mention it would be incorrect for the (newly added) debug logs.

I think it should be done at the start of the ShouldAnalyze method.

src/Integration.UnitTests/Exclusions/AnalyzableFileIndicatorTests.cs
```diff
@@ -142,6 +142,28 @@ public void ShouldAnalyze_HasInclusions_NoExclusions_ReturnsIfIncluded(bool incl
             result.Should().Be(expectedResult);
         }
 
+        [TestMethod]
+        public void ShouldAnalyze_HasBackSlash_ReplacesWithForwardSlash()
```
duncanp-sonar, line 146: It would help if the test name made it clear what had the backslash (i.e. the file name, as opposed to the pattern).

rita-sonarsource (approved):
Only the comment about running it multiple times.

src/Integration.UnitTests/Exclusions/AnalyzableFileIndicatorTests.cs
```diff
@@ -142,6 +142,28 @@ public void ShouldAnalyze_HasInclusions_NoExclusions_ReturnsIfIncluded(bool incl
             result.Should().Be(expectedResult);
         }
 
+        [TestMethod]
```
duncanp-sonar, line 145: If a test is directly related to a specific bug we often put a comment to help with traceability e.g. 

> // Regression test for https://github.com/SonarSource/sonarlint-visualstudio/issues/3075

src/Integration/Exclusions/AnalyzableFileIndicator.cs
```diff
@@ -87,6 +87,10 @@ public bool ShouldAnalyze(string filePath)
...
+            filePath = filePath.Replace("\\", "/");
```
duncanp-sonar, line 92: Worth explaining in a comment that SonarQube/SonarCloud always uses / as the directory separator in patterns regardless of OS.

duncanp-sonar (approved):
Just a couple of comments about comments.

PR #3081, opened 2022-06-21
Added Mend/Whitesource configuration file

duncanp-sonar:


duncanp-sonar:
This PR is more for information for now.

My plan is:
* enable WhiteSource/Mend scanning
* look at the reported issues
* send more PRs for any required package upgrades (e.g. there is a reported vulnerability for the SharpLib package we use at build-time; it doesn't affect us, but the simplest solution is to upgrade the package to make the warning go away).
* finally, merge the changes to `master`.

There are still a number of questions I need to address e.g. do we need to do separate scans for VS2019 and VS2022, since we conditionally reference difference packages? (I'm expecting the answer "yes").

PR #3082, opened 2022-06-21
Show goldbar if server exclusions are different

rita-sonarsource:
Fixes [#3083](https://github.com/SonarSource/sonarlint-visualstudio/issues/3083)

Perhaps worth reviewing in separate commits: first commit changes the call to async, and second commit pulls exclusions from the server and compares.

### DIFF
--- a/src/Integration/LocalServices/IUnboundSolutionChecker.cs
+++ b/src/Integration/LocalServices/IUnboundSolutionChecker.cs
@@ -23,8 +23,10 @@ using System.Threading;
 using System.Threading.Tasks;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Integration.Helpers;
 using SonarLint.VisualStudio.Integration.Resources;
+using SonarQube.Client;
 
 namespace SonarLint.VisualStudio.Integration
 {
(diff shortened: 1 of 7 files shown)


src/Integration/LocalServices/IUnboundSolutionChecker.cs
```diff
@@ -31,25 +35,42 @@ internal interface IUnboundSolutionChecker
...
-                return !exclusionSettingsStorage.SettingsExist();
+                var savedExclusions = exclusionSettingsStorage.GetSettings();
+
+                if (savedExclusions == null)
+                {
+                    return true;
+                }
+
+                var bindingConfiguration = bindingConfigProvider.GetConfiguration();
+                var serverExclusions = await sonarQubeService.GetServerExclusions(bindingConfiguration.Project.ProjectKey, token);
```
rita-sonarsource, line 71: If we are in standalone mode, then GetSettings would return null and we will never get here. So Project must be not null here.

rita-sonarsource:
I will update the submodule to branch-exclusions before committing

src/Integration/LocalServices/IUnboundSolutionChecker.cs
duncanp-sonar: I think it's worth adding an assertion to document this assumption.

duncanp-sonar (approved):

ugras-ergun-sonarsource (approved):

PR #3104, opened 2022-07-14
Rename CFamily binding config to generic non-roslyn

rita-sonarsource:
Prep for [#3105](https://github.com/SonarSource/sonarlint-visualstudio/issues/3105)

### DIFF
--- a/src/Core.UnitTests/Binding/NonRoslynBindingConfigFileTests.cs
+++ b/src/Core.UnitTests/Binding/NonRoslynBindingConfigFileTests.cs
@@ -25,26 +25,26 @@ using System.Linq;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
-using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Core.Binding;
 
-namespace SonarLint.VisualStudio.Core.UnitTests.CFamily
+namespace SonarLint.VisualStudio.Core.UnitTests.Binding
 {
     [TestClass]
-    public class CFamilyBindingConfigTests
+    public class NonRoslynBindingConfigFileTests
     {
         [TestMethod]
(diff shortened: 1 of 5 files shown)


src/Core.UnitTests/Binding/NonRoslynBindingConfigFileTests.cs
```diff
@@ -25,34 +25,34 @@
...
 using Moq;
-using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Core.Binding;
 
-namespace SonarLint.VisualStudio.Core.UnitTests.CFamily
+namespace SonarLint.VisualStudio.Core.UnitTests.Binding
 {
     [TestClass]
-    public class CFamilyBindingConfigTests
+    public class NonRoslynBindingConfigFileTests
```
rita-sonarsource, line 33: There was nothing cfamily specific about these classes. They're simply fetching the QP and saving it to disk. The code for js/ts was identical so I renamed them to NonRoslynXXX. 
Unfortunately the Roslyn class is similar but more complicated. I think we can unite parts there as well but I've left it for now.

src/Core.UnitTests/Binding/NonRoslynBindingConfigProviderTests.cs
```diff
@@ -27,29 +27,38 @@
...
+namespace SonarLint.VisualStudio.Core.UnitTests.Binding
 {
     [TestClass]
-    public class CFamilyBindingConfigProviderTests
+    public class NonRoslynBindingConfigProviderTests
     {
+        private static readonly Language[] SupportedLanguages = { Language.C, Language.Cpp };
+
+        [TestMethod]
+        public void Ctor_NullSupportedLanguages_ArgumentNullException()
```
rita-sonarsource, line 43: Added this test for consistency with the others although they add little value.

duncanp-sonar (approved):

Issue #3105, opened 2022-07-14
Fetch JS/TS QP from the server and save to disk


rita-sonarsource:
### Description

Support fetching JS/TS quality profile from the server

PR #3106, opened 2022-07-14
Fetch JS/TS QP from the server and save to disk

rita-sonarsource:
Fixes [#3105](https://github.com/SonarSource/sonarlint-visualstudio/issues/3105)
It might help to look at each commit separately; the changes done in each are only to fix a specific part of the binding.

### DIFF
--- a/src/Core/Language.cs
+++ b/src/Core/Language.cs
@@ -47,6 +47,8 @@ namespace SonarLint.VisualStudio.Core
         public readonly static Language VBNET = new Language("VB", CoreStrings.VBNetLanguageName, "vb.ruleset", SonarQubeLanguage.VbNet);
         public readonly static Language Cpp = new Language("C++", CoreStrings.CppLanguageName, "_cpp_settings.json", SonarQubeLanguage.Cpp);
         public readonly static Language C = new Language("C", "C", "_c_settings.json", SonarQubeLanguage.C);
+        public readonly static Language Js = new Language("Js", "JavaScript", "_js_settings.json", SonarQubeLanguage.Js);
+        public readonly static Language Ts = new Language("Ts", "TypeScript", "_ts_settings.json", SonarQubeLanguage.Ts);
 
         /// <summary>
         /// Returns the language for the specified language key, or null if it does not match a known language
--- a/src/Core/Language.cs
+++ b/src/Core/Language.cs
@@ -83,18 +83,7 @@ namespace SonarLint.VisualStudio.Core
         /// <summary>
         /// Returns whether or not this language is a supported project language.
         /// </summary>
-        public bool IsSupported => SupportedLanguages.Contains(this);
-
-        /// <summary>
-        /// All languages which are supported for project binding.
-        /// </summary>
-        public static IEnumerable<Language> SupportedLanguages
-        {
-            get
-            {
-                return new[] { CSharp, VBNET, Cpp, C };
-            }
-        }
+        public bool IsSupported => KnownLanguages.Contains(this);
 
         /// <summary>
         /// All known languages.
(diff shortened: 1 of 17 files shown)


src/Core/Language.cs
```diff
@@ -81,18 +83,7 @@ public sealed class Language : IEquatable<Language>
         /// <summary>
         /// Returns whether or not this language is a supported project language.
         /// </summary>
-        public bool IsSupported => SupportedLanguages.Contains(this);
-
-        /// <summary>
-        /// All languages which are supported for project binding.
-        /// </summary>
-        public static IEnumerable<Language> SupportedLanguages
```
rita-sonarsource, line 89: This was an exact duplication of KnownLanguages and not used anywhere else.

src/Core/SonarArtefactKeys.cs
```diff
@@ -57,6 +57,8 @@ public static class SonarPluginKeys
         public const string SonarCSharp = "csharp";
         public const string SonarVBNet = "vbnet";
         public const string SonarCFamily = "cpp";
+        public const string SonarJs = "javascript";
+        public const string SonarTs = "typescript";
```
rita-sonarsource: SonarTs is a mistake and was removed in a subsequent commit

src/Integration/Connection/MinimumSupportedSonarQubePlugin.cs
```diff
@@ -35,7 +35,10 @@ internal class MinimumSupportedSonarQubePlugin
         // i.e. the analysis rules implementations should be similar so the issues shown in the IDE should be
         // similar to those reported on the server.
         public static readonly MinimumSupportedSonarQubePlugin CFamily = new MinimumSupportedSonarQubePlugin(SonarPluginKeys.SonarCFamily, "SonarCFamily", "6.0", Language.Cpp, Language.C);
-        public static readonly IEnumerable<MinimumSupportedSonarQubePlugin> All = new[] { CSharp, VbNet, CFamily };
+
+        public static readonly MinimumSupportedSonarQubePlugin JsTs = new MinimumSupportedSonarQubePlugin(SonarPluginKeys.SonarJs, "SonarJS", "8.6", Language.Js, Language.Ts);
```
rita-sonarsource, line 39: I put version 8.6 as "minimal" because that's what we embed in SLVS in this release.
Perhaps it would be better to pick an older version, in cause the user wants to connect to a SQ version that has an older version. SC is at 9.4 btw.

src/Integration/ProjectToLanguageMapper.cs
```diff
@@ -47,34 +48,25 @@ public interface IProjectToLanguageMapper
...
-        /// </summary>
-        /// <returns>
-        /// Previously the code assumed a one-to-one mapping between project types and languages.
-        /// The worked when the only supported languages were C# and VB. It doesn't work now that
-        /// connected mode is supported for C++ projects (which can have both C++ and C files).
-        /// New code should call <see cref="GetAllBindingLanguagesForProject(EnvDTE.Project)"/> instead
-        /// and handle the fact that there could be multiple supported languages.
-        /// </returns>
-        private Language GetLanguageForProject(Project dteProject)
+        public IEnumerable<Language> GetAllBindingLanguagesForProject(Project dteProject)
```
rita-sonarsource, line 69: This method needed a small clean up since a project can be both CSharp AND have js.

src/Integration/ProjectToLanguageMapper.cs
```diff
@@ -83,32 +75,34 @@ private Language GetLanguageForProject(Project dteProject)
...
-            return Language.Unknown;
-        }
-
-        public IEnumerable<Language> GetAllBindingLanguagesForProject(Project dteProject)
-        {
-            var language = GetLanguageForProject(dteProject);
+            if (jsTsProjectTypeIndicator.IsJsTs())
+            {
+                languages.Add(Language.Js);
+                languages.Add(Language.Ts);
```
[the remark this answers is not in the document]
rita-sonarsource: I think for simplicity's sake it might be easier to keep it like that. We can revisit it in the future.

Issue #3109, opened 2022-07-19
Apply connected mode settings when running analysis


rita-sonarsource:
### Description

Apply JS/TS connected mode settings when running analysis

PR #3112, opened 2022-07-20
Reuse IRuleSettingsProvider in CFamilyRuleConfigProvider

rita-sonarsource:
Cleanup following #[3109](https://github.com/SonarSource/sonarlint-visualstudio/issues/3109)

### DIFF
--- a/src/CFamily.UnitTests/Rules/CFamilyRuleConfigProviderTests.cs
+++ b/src/CFamily.UnitTests/Rules/CFamilyRuleConfigProviderTests.cs
@@ -18,16 +18,14 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.Collections.Generic;
-using System.IO.Abstractions;
 using FluentAssertions;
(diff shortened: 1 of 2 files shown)


src/CFamily.UnitTests/Rules/CFamilyRuleConfigProviderTests.cs
```diff
@@ -18,233 +18,86 @@
...
 using SonarLint.VisualStudio.CFamily.Helpers.UnitTests;
 using SonarLint.VisualStudio.Core;
-using SonarLint.VisualStudio.Core.Binding;
-using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.Integration;
 
 namespace SonarLint.VisualStudio.CFamily.Rules.UnitTests
 {
     [TestClass]
     public class CFamilyRuleConfigProviderTests
```
rita-sonarsource, line 33: IRuleSettingsProviderFactory hides now the details between connected mode and standalone mode so there is no need for the tests anymore. I've kept a single integration test that in fact tests the EffectiveRulesConfigCalculator.

Issue #2315, opened 2021-04-13
Refactor CFamily rules definition code


duncanp-sonar:
### Description

Currently we are embedded hundreds of json files in the VSIX that contain the rules definitions for the C and C++ rules.
These files are then loaded at run time to give the default configuration, before any user settings or QP are applied.

There are two problems:
* unnecessary overhead at runtime - the information could be calculated at compile time
* it creates packaging issues. The MS toolchain that creates the VSIX file falls over if there are too many files in the VSIX. We could embed one file in the VSIX rather than hundreds.

TODO
- [ ] generate a single consolidated settings files are build time, using the same format as for JS/TS
- [ ] move the JS/TS rule definition data classes to a shared project e.g. Core.
- [ ] embed the consolidate file in the VSIX
- [ ] modify the CFamily rule configuration provider to process the new file

PR #3114, opened 2022-07-20
[Infra] Move CFamily jar processing from MSBuild task to exe

duncanp-sonar:
* added a new console app that does the equivalent of the current `DownloadSonarPlugin` targets/tasks for the CFamily.
* changed the VSIX project to use the new console app

At the moment, the processing is the same as before i.e. the app will copy all of the required files to the target `lib` folder in the VSIX project.

Part of #2315


build/tools/CFamilyJarPreProcessor/CFamilyJarPreProcessor.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
duncanp-sonar, line 1: Separate solution file for the tool. Not strictly required, but makes working in VS easier.

build/tools/CFamilyJarPreProcessor/Common.cs
```diff
@@ -0,0 +1,251 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Net.Http;
+using SharpCompress.Compressors.Xz;
+
+namespace CFamilyJarPreProcessor
+{
+    internal static class Common
```
duncanp-sonar, line 33: Copy of the existing file from `DownloadSonarPlugin`, with an modified namespace and copyright date.

build/tools/CFamilyJarPreProcessor/ILogger.cs
```diff
@@ -0,0 +1,36 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+using System;
+
+namespace CFamilyJarPreProcessor
```
duncanp-sonar, line 22: The `DownloadSonarPlugin` project created MSBuild Tasks, which use the MSBuild task logger. This is a minimal replacement.

build/tools/CFamilyJarPreProcessor/Preprocessor.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            var files = new List<string>();
+
+            files.AddRange(Common.FindSingleFiles(searchRoot, SingleFilePatterns, logger));
+            files.AddRange(Common.FindMultipleFiles(searchRoot, MultipleFilesPatterns, logger));
+
+            return files;
+        }
+
+        private void CopyFilesToOutputDirectory(IList<string> files, string destinationDir)
```
duncanp-sonar, line 97: This method and the one below are new.

Logically, the net effect of calling the task is still the same - the jar is downloaded and the files are copied to the destination folder.

However, in the MSBuild Task version, the task return the list of files to the MSBuild target which then did the copy.

Now, the new exe does the copy part too i.e. this method and the one below.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -364,20 +364,25 @@
     <!-- Hook the target that fetches the CFamily files into to the build process. 
          This needs to be set after the SDK targets have been imported.
     -->
-    <PrepareForBuildDependsOn>$(PrepareForBuildDependsOn);GetCFamilyFiles</PrepareForBuildDependsOn>
+    <CoreCompileDependsOn>$(CoreCompileDependsOn);GetCFamilyFiles</CoreCompileDependsOn>
```
duncanp-sonar, line 367: We used to be running this target very early in the build process. I changed it to be a bit later, but it probably doesn't make any difference.

The reason I tried changing it was that previously the target was being executed by VS at design time i.e. when VS triggered a design-time build in the background. I wanted to see if moving it later would stop VS from triggering it repeatedly (it doesn't, but see below).

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -364,20 +364,25 @@
...
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+      <IncludeOutputGroupsInVSIX>GetItemsToIncludeInVSIX_NoOp</IncludeOutputGroupsInVSIX>
+    </ProjectReference>
+  </ItemGroup>
 
-    <!-- Note: building the helper project explicitly using MSBuild rather than referencing it via a ProjectReference.
-         This makes is easier for us to control when the build is performed, so we can trigger it early enough
-         to include the downloaded files as normal Content files in this project.
-         Also, we are using the "Exec" task to invoke MSBuild.exe rather the "MSBuild" task when building
-         DownloadSonarPlugin.csproj to avoid getting file-locking errors when rebuilding.
```
duncanp-sonar, line 377: Hopefully this new very will actually fix the file-locking issues.
It should, because we are now referencing the tool via a `ProjectReference`, so MSBuild will be able to correctly handle the build dependencies, and will only build the tool project once.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -364,20 +364,25 @@
...
-    -->
-    <Exec Command="&quot;$(MSBuildBinPath)\msbuild.exe&quot; ..\..\build\DownloadSonarPlugin\DownloadSonarPlugin.csproj /p:PluginUrl=$(PluginUrl) /p:TargetDirectory=&quot;$(CFamilyTargetDirectory)&quot; /p:UseSharedCompilation=false /t:Restore;Build;EmbedCFamilyFiles /p:VsTargetVersion=$(VsTargetVersion) /nr:false /p:BuildInParallel=false" LogStandardErrorAsError="true" IgnoreExitCode="false" />
```
duncanp-sonar, line 379: The command is simpler now - just run the 'exe`, passing the URL and target folder.

I'm not sure why we were passing the `VSTargetVersion` before - it wasn't used.

SonarLint.VisualStudio.Integration.sln
```diff
@@ -4,6 +4,9 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 VisualStudioVersion = 17.0.31815.197
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Integration.Vsix", "src\Integration.Vsix\Integration.Vsix.csproj", "{FF2AD819-28F4-493A-8E9B-1D3F16BD4689}"
+	ProjectSection(ProjectDependencies) = postProject
```
duncanp-sonar, line 7: This tells MSBuild that the VSIX project depends on the new tool project.

It was set from the "Project Build Order" dialogue:

image

build/tools/CFamilyJarPreProcessor/CFamilyJarPreProcessor.sln
duncanp-sonar, line 1: New solution file referring to just the new tool project.
It's not strictly required, but it makes development and testing much quicker since I don't have to build the whole SLVS sln.

src/Integration.Vsix/Integration.Vsix.csproj
duncanp-sonar: Comment is no longer relevant - I reverted this change.

rita-sonarsource (approved):
As we discussed offline, LGTM

Issue #3113, opened 2022-07-20
Connected mode is not enabled when JS project is opened as folder


rita-sonarsource:
### Description

When a folder containing JS code is opened "as folder", connected mode fails with the message "This solution contains no supported project language".

Notes:
* This is because IProjectSystemHelper.GetSolutionProjects returns nothing, so we think it's the equivalent of an empty solution.
* In case of CMake: opening a folder with CMake leads to VS asking if the user would like to treat it as a CMake project. If the user answers yes, `GetSolutionProjects` correctly returns "miscellaneous folder" project. If the user answers no and opens the CMake as a regular folder, the same behavior occurs as with JS.

TODO:
- [ ] Can we use IFolderWorkspaceService.IsFolderWorkspace when calling IProjectSystemHelper.GetSolutionProjects ?

PR #3115, opened 2022-07-21
Support connected mode for folders

rita-sonarsource:
Fixes [#3113](https://github.com/SonarSource/sonarlint-visualstudio/issues/3113)

### DIFF
--- a/src/Integration.UnitTests/Binding/BindingControllerTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingControllerTests.cs
@@ -485,14 +485,14 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
         }
 
         [TestMethod]
-        public void CanExecute_NoProjects_FolderWorkspace_False()
+        public void CanExecute_NoProjects_FolderWorkspace_True()
         {
             SetupSolutionState(isSolutionLoaded: false, isSolutionNotBuilding: true, isOpenAsFolder: true, hasProjects: false);
 
             var testSubject = CreateBindingController();
 
             var canExecute = testSubject.BindCommand.CanExecute(CreateBindingArguments("project1", "name1", "http://localhost"));
-            canExecute.Should().BeFalse();
+            canExecute.Should().BeTrue();
         }
 
         [TestMethod]
--- a/src/Integration/Binding/BindingController.cs
+++ b/src/Integration/Binding/BindingController.cs
@@ -107,8 +107,8 @@ namespace SonarLint.VisualStudio.Integration.Binding
                    && !host.VisualStateManager.IsBusy
                    && (folderWorkspaceService.IsFolderWorkspace()
                        || (knownUIContexts.SolutionExistsAndFullyLoadedContext.IsActive
-                           && knownUIContexts.SolutionExistsAndNotBuildingAndNotDebuggingContext.IsActive))
-                   && (projectSystemHelper.GetSolutionProjects()?.Any() ?? false);
+                           && knownUIContexts.SolutionExistsAndNotBuildingAndNotDebuggingContext.IsActive
+                           && (projectSystemHelper.GetSolutionProjects()?.Any() ?? false)));
         }
 
         private void OnBind(BindCommandArgs args)
--- a/src/Integration/Binding/BindingProcessImpl.cs
+++ b/src/Integration/Binding/BindingProcessImpl.cs
@@ -21,7 +21,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.IO.Abstractions;
 using System.Linq;
 using System.Text;
 using System.Threading;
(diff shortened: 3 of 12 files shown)


src/Integration/Binding/BindingController.cs
```diff
@@ -107,8 +107,8 @@ private bool OnBindStatus(BindCommandArgs args)
                    && !host.VisualStateManager.IsBusy
                    && (folderWorkspaceService.IsFolderWorkspace()
                        || (knownUIContexts.SolutionExistsAndFullyLoadedContext.IsActive
-                           && knownUIContexts.SolutionExistsAndNotBuildingAndNotDebuggingContext.IsActive))
-                   && (projectSystemHelper.GetSolutionProjects()?.Any() ?? false);
+                           && knownUIContexts.SolutionExistsAndNotBuildingAndNotDebuggingContext.IsActive
+                           && (projectSystemHelper.GetSolutionProjects()?.Any() ?? false)));
```
rita-sonarsource, line 111: `projectSystemHelper.GetSolutionProjects()` returns empty when it's a folder that is not CMake. The previous condition worked only for folder that is CMake.

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -127,6 +134,11 @@ public async Task<bool> DownloadQualityProfileAsync(IProgress<FixedStepsProgress
             {
                 var serverLanguage = language.ServerLanguage;
 
+                if (serverLanguage == null)
```
rita-sonarsource, line 137: For some reason, ProjectToLanguageMapper returns a list with `Language.Unknown` if no language was detected. I don't know why this is the behavior, and I'm not sure what will break if I change it. So I'm safe-guarding for it here.

src/Integration/ProjectToLanguageMapper.cs
```diff
@@ -112,6 +119,11 @@ public bool HasSupportedLanguage(Project project)
             return languages.Any(x => x.IsSupported);
         }
 
+        public IEnumerable<Language> GetAllBindingLanguagesInSolution()
+        {
+            return GetLanguagesInProject(Guid.Empty);
```
rita-sonarsource, line 124: essentially executes just IsCMake and IsJsTs

src/Integration.UnitTests/Binding/BindingControllerTests.cs
```diff
@@ -485,14 +485,14 @@ public void CanExecute_NoProjects_NotFolderWorkspace_False()
         }
 
         [TestMethod]
-        public void CanExecute_NoProjects_FolderWorkspace_False()
+        public void CanExecute_NoProjects_FolderWorkspace_True()
         {
-            SetupSolutionState(isSolutionLoaded: false, isSolutionNotBuilding: true, isOpenAsFolder: true, hasProjects: false);
+            SetupSolutionState(isSolutionLoaded: true, isSolutionNotBuilding: true, isOpenAsFolder: true, hasProjects: false);
```
rita-sonarsource, line 490: I changed "isSolutionLoaded" to true as it was in other tests.

src/Integration/ProjectToLanguageMapper.cs
```diff
@@ -112,6 +119,11 @@ public bool HasSupportedLanguage(Project project)
             return languages.Any(x => x.IsSupported);
         }
 
+        public IEnumerable<Language> GetAllBindingLanguagesInSolution()
```
rita-sonarsource, line 122: Not tested because it will be changed by @ugras-ergun-sonarsource

rita-sonarsource (commented):
.

src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -271,6 +272,13 @@ private static bool IsTestProjectPatternProperty(SonarQubeProperty property)
         private async Task<bool> AreSolutionProjectsAndSonarQubePluginsCompatibleAsync(IProgressController controller,
             IProgressStepExecutionEvents notifications, CancellationToken cancellationToken)
         {
+            var folderWorkspaceService = host.GetMefService<IFolderWorkspaceService>();
+
+            if (folderWorkspaceService.IsFolderWorkspace())
+            {
+                return true;
```
duncanp-sonar, line 279: This isn't strictly correct. If I remember correctly, the purpose of this method was supposed to be to only allow binding if there are bindable projects with the corresponding plugins that could analyse those project types e.g. if you had a C++ project but don't have the C++ plugin because you are using Community edition, you couldn't bind.

Here, we don't know which languages are being used in the solution, so we shouldn't just return true.

However, it's a corner case, so perhaps just create a ticket for it to do "later".

duncanp-sonar (approved):

ugras-ergun-sonarsource (approved):

src/Integration/Connection/ConnectionWorkflow.cs
rita-sonarsource: You're right, the logic here should also check if there are bindable languages. I'll add a ticket.

PR #3124, opened 2022-07-27
moved result of GetContentTypeForExtension to a variable

ugras-ergun-sonarsource:
Rita suggested this change after the merge of the branch. With using of GetContentTypeForExtension the result will not be different for each language anymore so no need to run the method multiple times.

### DIFF
--- a/src/Infrastructure.VS/LanguageDetection/SonarLanguageRecognizer.cs
+++ b/src/Infrastructure.VS/LanguageDetection/SonarLanguageRecognizer.cs
@@ -124,12 +124,13 @@ namespace SonarLint.VisualStudio.IssueVisualization.Editor.LanguageDetection
         {
             var extension = GetNormalizedExtention(fileName);
 
+            var contentTypeName = fileExtensionRegistryService.GetContentTypeForExtension(extension).TypeName;
+
             // ContentType for "js" is typescript we do manual check to be consistent with Detect method
             if (JavascriptSupportedExtensions.Contains(extension)) { return AnalysisLanguage.Javascript; }
-            if (fileExtensionRegistryService.GetContentTypeForExtension(extension).TypeName == TypeScriptTypeName) { return AnalysisLanguage.TypeScript; }
-            if (fileExtensionRegistryService.GetContentTypeForExtension(extension).TypeName == CFamilyTypeName) { return AnalysisLanguage.CFamily; }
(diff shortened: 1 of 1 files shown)


src/Infrastructure.VS/LanguageDetection/SonarLanguageRecognizer.cs
```diff
@@ -124,12 +124,13 @@ private static bool IsTypeScriptDocument(IEnumerable<IContentType> contentTypes)
...
 
+            var contentTypeName = fileExtensionRegistryService.GetContentTypeForExtension(extension).TypeName;
+
             // ContentType for "js" is typescript we do manual check to be consistent with Detect method
             if (JavascriptSupportedExtensions.Contains(extension)) { return AnalysisLanguage.Javascript; }
-            if (fileExtensionRegistryService.GetContentTypeForExtension(extension).TypeName == TypeScriptTypeName) { return AnalysisLanguage.TypeScript; }
-            if (fileExtensionRegistryService.GetContentTypeForExtension(extension).TypeName == CFamilyTypeName) { return AnalysisLanguage.CFamily; }
-            if (fileExtensionRegistryService.GetContentTypeForExtension(extension).TypeName == CSharpTypeName 
-                || fileExtensionRegistryService.GetContentTypeForExtension(extension).TypeName == BasicTypeName) { return AnalysisLanguage.RoslynFamily; }
+            if (contentTypeName == TypeScriptTypeName) { return AnalysisLanguage.TypeScript; }
```
rita-sonarsource, line 131: I'd move the call to GetContentTypeForExtension after the javascript call, just for potential performance. Also a switch case might make this more readable, with default returning null.

src/Infrastructure.VS/LanguageDetection/SonarLanguageRecognizer.cs
```diff
@@ -123,15 +123,23 @@ private static bool IsTypeScriptDocument(IEnumerable<IContentType> contentTypes)
         public AnalysisLanguage? GetAnalysisLanguageFromExtension(string fileName)
         {
             var extension = GetNormalizedExtention(fileName);
-
+            
             // ContentType for "js" is typescript we do manual check to be consistent with Detect method
             if (JavascriptSupportedExtensions.Contains(extension)) { return AnalysisLanguage.Javascript; }
```
rita-sonarsource, line 128: Final nitpicking: drop a line please, no "reason" anymore to have the return on the same line

rita-sonarsource (approved):
See final cosmetics comment

PR #3132, opened 2022-07-28
Support Taint vulnerabilities in folder workspaces

rita-sonarsource:
Fixes [#3133](https://github.com/SonarSource/sonarlint-visualstudio/issues/3133)

### DIFF
--- a/src/Core/IFolderWorkspaceInitializedEvent.cs
+++ b/src/Core/IFolderWorkspaceInitializedEvent.cs
@@ -1,0 +1,30 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+
+using System;
+
+namespace SonarLint.VisualStudio.Core
+{
+    public interface IFolderWorkspaceInitializedEvent
+    {
+        event EventHandler FolderWorkspaceInitialized;
+    }
+}
--- a/src/Integration.UnitTests/MefServices/ActiveSolutionTrackerTests.cs
+++ b/src/Integration.UnitTests/MefServices/ActiveSolutionTrackerTests.cs
@@ -125,10 +125,10 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             counter.Should().Be(isFolderWorkspace ? 1 : 0);
         }
 
-        private ActiveSolutionTracker CreateTestSubject(bool? isFolderWorkspace = null)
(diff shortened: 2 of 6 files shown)


src/Integration/MefServices/ActiveSolutionTracker.cs
```diff
@@ -55,6 +59,10 @@ public ActiveSolutionTracker([Import(typeof(SVsServiceProvider))] IServiceProvid
         #region IVsSolutionEvents
         int IVsSolutionEvents.OnAfterOpenProject(IVsHierarchy pHierarchy, int fAdded)
         {
+            if (folderWorkspaceService.IsFolderWorkspace())
```
rita-sonarsource, line 62: Technically we know from the given pHierarchy that it's a folder, but I thought it would be more consistent to use the existing `folderWorkspaceService`.

src/Integration.UnitTests/MefServices/ActiveSolutionTrackerTests.cs
```diff
@@ -105,5 +107,31 @@ public void ActiveSolutionTracker_RaiseEventOnFolderOpen()
...
+            this.solutionMock.SimulateProjectOpen(null);
+
+            // Assert
+            counter.Should().Be(isFolderWorkspace ? 1 : 0);
+        }
+
+        private ActiveSolutionTracker CreateTestSubject(bool? isFolderWorkspace = null)
+        {
+            var folderWorkspaceFolder = new Mock<IFolderWorkspaceService>();
+            folderWorkspaceFolder.Setup(x => x.IsFolderWorkspace()).Returns(isFolderWorkspace ?? false);
```
ugras-ergun-sonarsource, line 131: instead of a null-check it would make more sense to me to have false as the default value.

src/Core/IFolderWorkspaceInitializedEvent.cs
```diff
@@ -0,0 +1,30 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+
+using System;
+
+namespace SonarLint.VisualStudio.Core
+{
+    public interface IFolderWorkspaceInitializedEvent
```
duncanp-sonar, line 26: Naming: having `Event` in the name looks a bit odd. Something like `IFolderWorkspaceMonitor` would be more consistent with `IAnalysisConfigMonitor` and `ISuppressedIssuesMonitor` (both of which just raise a single event).

src/Integration/MefServices/ActiveSolutionTracker.cs
duncanp-sonar: Probably worth adding a comment as to why we need to raise this event for folder workspaces, as it's not obvious.

ugras-ergun-sonarsource (approved):
one small change in tests otherwise looks good.

src/IssueViz.Security.UnitTests/Taint/TaintIssuesBindingMonitorTests.cs
```diff
@@ -59,23 +62,40 @@ public void Ctor_SubscribeToSolutionBindingChanged()
...
+        {
+            var synchronizer = new Mock<ITaintIssuesSynchronizer>();
+            var activeSolutionBoundTracker = new Mock<IActiveSolutionBoundTracker>();
+            var folderWorkspaceInitialized = new Mock<IFolderWorkspaceInitializedEvent>();
+
+            new TaintIssuesBindingMonitor(activeSolutionBoundTracker.Object, folderWorkspaceInitialized.Object, synchronizer.Object);
+
+            folderWorkspaceInitialized.Raise(x => x.FolderWorkspaceInitialized += null, EventArgs.Empty);
+
+            synchronizer.Verify(x => x.SynchronizeWithServer(), Times.Once);
```
duncanp-sonar, line 85: Maybe add a line to clear the synchronizer (`synchronizer.Invocations.Clear()`?) immediately before raising the event. That way the test guarantees the synchronizer is called from the event handler rather than the constructor.
Same for the other test.

duncanp-sonar (commented):
Looks good, just a couple of minor comments.

duncanp-sonar (approved):

PR #3140, opened 2022-08-01
Added decorator to wrap server calls

duncanp-sonar:
- adds logging for timing and thread info


duncanp-sonar:
No longer required.
I used this build to check if we're making web calls on the UI thread, and we are.
This version adds a decorator which wraps all of the calls with logging. It worked fine for that, but it doesn't really help us fix the problem.

A newer version of the potential fix is #3141, which adds ETW tracing to the web client, and also makes `InvokeRequestAsync` `protected virtual`. That means:
1) we don't need this wrapper to do the logging, and 
2) the MEF subclass could handling switching to the background thread if required.

Issue #3139, opened 2022-07-29
Execute web calls on the background thread


duncanp-sonar:
### Description

Related to #3137. That ticket is one instance we know about. There many be others.

There are various things we could do:
* find and check each web call separately
* add assertions to catch such issues earlier
* update the `MefSonarQubeService` wrapper to always run on the background thread

Issue #3144, opened 2022-08-02
[Tests] Fix/remove tests that use HackedVSThreadingScope


duncanp-sonar:
### Description

We have a test helper with some very ugly hack reflection-based code to help test threading wrappers.
They fail on VS2022.

We should fix them or remove them.

pavel-mikula-sonarsource:
Moved to [SLVS-2237](https://sonarsource.atlassian.net/browse/SLVS-2237)

[SLVS-2237]: https://sonarsource.atlassian.net/browse/SLVS-2237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3146, opened 2022-08-02
Added RunOnBackgroundThread to IThreadHandling

duncanp-sonar:
Prep for #3139

### DIFF
--- a/src/Infrastructure.VS.UnitTests/ThreadHandlingTests.cs
+++ b/src/Infrastructure.VS.UnitTests/ThreadHandlingTests.cs
@@ -23,6 +23,7 @@ using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarLint.VisualStudio.Integration.UnitTests;
+using VSThreadHelper = Microsoft.VisualStudio.Shell.ThreadHelper;
 
 namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
 {
@@ -44,7 +45,7 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
         }
 
         [TestMethod]
-        public async Task CheckAccess_NotUIThread_ReturnsFlase()
+        public async Task CheckAccess_NotUIThread_ReturnsFalse()
         {
             bool actual = false;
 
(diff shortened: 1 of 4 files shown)


src/Infrastructure.VS.UnitTests/ThreadHandlingTests.cs
```diff
@@ -44,7 +45,7 @@ public async Task CheckAccess_IsUIThread_ReturnsTrue()
         }
 
         [TestMethod]
-        public async Task CheckAccess_NotUIThread_ReturnsFlase()
+        public async Task CheckAccess_NotUIThread_ReturnsFalse()
```
duncanp-sonar, line 48: Miscellaneous: fixing typo

src/Infrastructure.VS.UnitTests/ThreadHandlingTests.cs
```diff
@@ -113,6 +114,36 @@ public async Task ThrowIfNotOnUIThread_NotUIThread_Throws()
...
+
+                // ... and finish on a background thread
+                VSThreadHelper.CheckAccess().Should().BeFalse();
+                System.Threading.Thread.CurrentThread.ManagedThreadId.Should().Be(startThreadId);
+            });
+        }
+
+        [TestMethod]
+        [Ignore] // Can't test this case because it requires mocking SwitchToMainThreadAsync which
+                 // is broken. See https://github.com/SonarSource/sonarlint-visualstudio/issues/3144
```
duncanp-sonar, line 141: We have a helper class `HackedVSThreadingScope` that messes about with VS internals so that calls to `ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync` won't fail.

Unfortunately, it doesn't work in VS2022 (which incidentally means some of the `RunOnUIThread` tests are being ignored. I created #3144 to fix it.

The hack is replacing a field in a VS type using reflection, and the owning type and field name have changed. I know what the new names are, but the RunUIThread tests still fail when you run them all together (they pass individually). At that point I gave up trying to fix it immediately.

src/Infrastructure.VS/ThreadHandling.cs
```diff
@@ -40,6 +41,21 @@ public class ThreadHandling : IThreadHandling
...
+        public async Task<T> RunOnBackgroundThread<T>(Func<T> asyncMethod)
+        {
+            if (!ThreadHelper.CheckAccess())
+            {
+                return asyncMethod();
+            }
+
+            await SwitchToBackgroundThread();
+            var result = asyncMethod();
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
duncanp-sonar, line 53: Line 51-53 are not currently tested.

ugras-ergun-sonarsource (approved):

PR #3147, opened 2022-08-03
Add try-catch for WPF commands

rita-sonarsource:
Fixes [#3142](https://github.com/SonarSource/sonarlint-visualstudio/issues/3142)


src/Integration/WPF/RelayCommand.cs
```diff
@@ -49,12 +50,25 @@ public RelayCommand(Action execute, Func<bool> canExecute)
 
         public bool CanExecute()
         {
-            return this.canExecute?.Invoke() ?? true;
+            try
```
duncanp-sonar, line 53: Probably worth commenting that these methods can be called directly from XAML on the UI thread so we need to guard against unhandled exception.

src/Integration/WPF/RelayCommand.cs
```diff
@@ -49,12 +50,25 @@ public RelayCommand(Action execute, Func<bool> canExecute)
...
         public void Execute()
         {
-            this.execute();
+            try
+            {
+                this.execute();
+            }
+            catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+            {
+            }
```
duncanp-sonar, line 71: Could add a comment here and the other empty `catch` to remove the code smells
e.g.
`// Just squash the exception`

duncanp-sonar (approved):

PR #3149, opened 2022-08-03
Changing method signature of RunOnBackgroundThread

duncanp-sonar:
* accepting a Func<T> doesn't fit nicely with using the method from MefSonarQubeService, which already has a task
* changing the argument to  Task<T> falls foul of VSTHRD003

https://github.com/Microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD003.md

### DIFF
--- a/src/Infrastructure.VS.UnitTests/ThreadHandlingTests.cs
+++ b/src/Infrastructure.VS.UnitTests/ThreadHandlingTests.cs
@@ -115,25 +115,11 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
         }
 
         [TestMethod]
+        [Ignore] // Can't test this case because it requires mocking ThreadHelper.JoinableTaskFactory.RunAsync which
+                 // is broken. See https://github.com/SonarSource/sonarlint-visualstudio/issues/3144
         public async Task RunOnBackgroundThread_StartsOnBackgroundThread_NoSwitch()
         {
-            await RunOnBackgroundThread(async () =>
-            {
-                var testSubject = new ThreadHandling();
-
-                var startThreadId = System.Threading.Thread.CurrentThread.ManagedThreadId;
-                // Should start on a background thread...
-                VSThreadHelper.CheckAccess().Should().BeFalse();
-
-                var result = await testSubject.RunOnBackgroundThread(() => VSThreadHelper.CheckAccess());
-
-                // Work should be done on the background thread
-                result.Should().BeFalse();
-
-                // ... and finish on a background thread
-                VSThreadHelper.CheckAccess().Should().BeFalse();
-                System.Threading.Thread.CurrentThread.ManagedThreadId.Should().Be(startThreadId);
-            });
+            // TODO
         }
 
         [TestMethod]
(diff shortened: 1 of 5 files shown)


src/Infrastructure.VS.UnitTests/ThreadHandlingTests.cs
```diff
@@ -115,25 +115,11 @@ public async Task ThrowIfNotOnUIThread_NotUIThread_Throws()
...
-                var result = await testSubject.RunOnBackgroundThread(() => VSThreadHelper.CheckAccess());
-
-                // Work should be done on the background thread
-                result.Should().BeFalse();
-
-                // ... and finish on a background thread
-                VSThreadHelper.CheckAccess().Should().BeFalse();
-                System.Threading.Thread.CurrentThread.ManagedThreadId.Should().Be(startThreadId);
-            });
+            // TODO
```
duncanp-sonar, line 122: The old test actually passed, but it wasn't working correctly as it was throwing on a background thread when calling `ThreadHelper.JoinableTaskFactory.RunAync`.

src/Infrastructure.VS/ThreadHandling.cs
```diff
@@ -41,19 +40,26 @@ public class ThreadHandling : IThreadHandling
...
-            var result = asyncMethod();
-            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                    if (startedOnUIThread)
+                    {
+                        await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                    }
 
-            Debug.Assert(ThreadHelper.CheckAccess(), "Expecting to returning on the UI thread");
-            return result;
+                    return result;
```
duncanp-sonar, line 61: This method is now compliant with [VSTHRD003](https://github.com/Microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD003.md). It can't accept a Task directly as that violates the rule.

Instead, it has to accept a factory method that returns the task, which can then be passed to `ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync`.

Note: there are currently no tests for this method.

src/Infrastructure.VS/ThreadHandling.cs
```diff
@@ -41,19 +40,26 @@ public class ThreadHandling : IThreadHandling
...
+            var startedOnUIThread = ThreadHelper.CheckAccess();
+
+            return await ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+                {
+                    if (startedOnUIThread)
+                    {
+                        await SwitchToBackgroundThread();
+                    }
+
+                    var result = await asyncMethod();
```
rita-sonarsource, line 54: I would just add a thread.sleep here for testing purposes, to make sure that `JoinableTaskFactory.RunAsync` doesn't block the UI thread

rita-sonarsource (approved):

src/Infrastructure.VS/ThreadHandling.cs
duncanp-sonar: I've manually tested with a ten-second sleep immediately before the execution of `asyncMethod` and the UI remains responsive.

image

This screenshot is from the slvs-community test VM. The UI thread id is 4,292.

PR #3150, opened 2022-08-03
Added ETW events to the ThreadHandling class

duncanp-sonar:


src/Infrastructure.VS/ThreadHandling.cs
```diff
@@ -42,22 +43,28 @@ public class ThreadHandling : IThreadHandling
...
+                        CodeMarkers.Instance.Threading_RunOnBackgroundThread_SwitchingToBackground();
                         await SwitchToBackgroundThread();
                     }
 
+                    CodeMarkers.Instance.Threading_RunOnBackgroundThread_ExecutingCall();
                     var result = await asyncMethod();
 
                     if (startedOnUIThread)
                     {
+                        CodeMarkers.Instance.Threading_RunOnBackgroundThread_SwitchingToUI();
```
rita-sonarsource, line 63: Question about Switching ToBackground/ToMain: they will be logged on the other thread; is that a problem for us when investigating?
i.e. we will see "switching to main thread" but the thread id will be background, and vice-versa.

rita-sonarsource (approved):

ugras-ergun-sonarsource (approved):

src/Infrastructure.VS/ThreadHandling.cs
duncanp-sonar: I think we'll have enough information because the execute is logged and we'll see what thread that is running on I that will tell us that it was successful.

If not, I'll change it.

Issue #3152, opened 2022-08-03
Access to ServerInfo is not thread-safe


rita-sonarsource:
### Description

Relates to #696 and #3142. SonarQubeService can be disconnected at any time, hence ServerInfo should be stored in a variable when accessing it multiple times.

PR #3153, opened 2022-08-03
Re-use the same ServerInfo

rita-sonarsource:
Fixes [#3152](https://github.com/SonarSource/sonarlint-visualstudio/issues/3152)

### DIFF
--- a/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
+++ b/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
@@ -85,7 +85,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
             {
                 var bindingConfiguration = configurationProvider.GetConfiguration();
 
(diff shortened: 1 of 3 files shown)


src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -85,7 +85,7 @@ public async Task SynchronizeWithServer()
             {
                 var bindingConfiguration = configurationProvider.GetConfiguration();
 
-                if (IsStandalone(bindingConfiguration) || !IsConnected() || !IsFeatureSupported())
+                if (IsStandalone(bindingConfiguration) || !IsConnected(out var serverInfo) || !IsFeatureSupported(serverInfo))
```
rita-sonarsource, line 88: This is a bit weak and in fact the problem can still occur: for example, if we pass the serverInfo checks and then fail when we actually try to call the server. 
But I think this is still better and safer, instead of the ugly null exception.

duncanp-sonar (approved):

PR #3156, opened 2022-08-04
Ignore user settings change when in connected mode

rita-sonarsource:
Fixes [#3157](https://github.com/SonarSource/sonarlint-visualstudio/issues/3157)

### DIFF
--- a/src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
+++ b/src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
@@ -62,25 +62,28 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis
 
         private void OnUserSettingsChanged(object sender, EventArgs e)
         {
-            // NB assumes exception handling is done by the AnalysisRequester
-            logger.WriteLine(AnalysisStrings.ConfigMonitor_UserSettingsChanged);
-            
-            // TODO: once connected mode is supported for all languages this event could
-            // be ignored in connected mode. See #771.
-            if (activeSolutionBoundTracker.CurrentConfiguration.Mode != SonarLintMode.Standalone)
+            if (activeSolutionBoundTracker.CurrentConfiguration.Mode == SonarLintMode.Standalone)
+            {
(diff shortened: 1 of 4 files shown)


src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -62,25 +62,28 @@ internal sealed class AnalysisConfigMonitor : IAnalysisConfigMonitor, IDisposabl
...
         {
-            // NB assumes exception handling is done by the AnalysisRequester
-            logger.WriteLine(AnalysisStrings.ConfigMonitor_UserSettingsChanged);
-            
-            // TODO: once connected mode is supported for all languages this event could
-            // be ignored in connected mode. See #771.
-            if (activeSolutionBoundTracker.CurrentConfiguration.Mode != SonarLintMode.Standalone)
+            if (activeSolutionBoundTracker.CurrentConfiguration.Mode == SonarLintMode.Standalone)
+            {
+                logger.WriteLine(AnalysisStrings.ConfigMonitor_UserSettingsChanged);
```
rita-sonarsource, line 67: Previously, this log message was written always, and it says "open documents will be re-analyzed". Since that's no longer true, I've changed it and now we're either writing the Standalone message or the Connected message.

src/Integration.Vsix/Analysis/AnalysisStrings.resx
```diff
@@ -121,7 +121,7 @@
     <value>Suppressions have been updated. Open documents will be re-analysed.</value>
   </data>
   <data name="ConfigMonitor_UserSettingsIgnoredForConnectedModeLanguages" xml:space="preserve">
-    <value>Note: user settings are ignored for languages that support Connected Mode i.e. C and C++</value>
+    <value>User settings change is ignored: current workspace is in Connected Mode.</value>
```
rita-sonarsource, line 124: This will be shown only when we're in connected mode.
The message below will be shown only when we're in standalone mode.

ugras-ergun-sonarsource (approved):

duncanp-sonar (approved):

PR #3158, opened 2022-08-04
Raise analysis config changed on the background thread

### LINKED ISSUE #3160: Raise analysis config changed on the background thread
### Description

An event is raised when the user settings file changes. This can result in a lot of processing.

The event should be raised on a background thread so it doesn't make the UI unresponsive.

It is the responsibility of the listeners to switch to the UI thread if/when required.

### Notes
Changing the event to a background thread could potentially cause issues if any of the listeners expect to be on the UI thread.

However, this is not expected to cause an issue here.
- the `AnalysisConfigMonitor` class currently raised the event the UI some cases and a background thread in others, depending on the underlying cause of the changes -> any listeners are already being called on the background thread.
- the config changes causes all open files to be reanalysed. The JS, TS and CFamily analyzers all handle being called on the UI or background threads (in fact, they all switch a background thread as soon as they are called).

### Known issues
Raising the event on the UI thread causes at least one known gold bar - #3138.

Updating the eslintbridge analysis config takes some time and involves a lock.
If the event is raised on the UI thread then the UI thread can be blocked for multiple seconds waiting to acquire the lock.

See the log output below for more information.

[AnalysisConfig] [OnSuppressionsUpdated] [Thread: 1, 06:09:26.899]  Raising analysis config changed event...
[EslintBridgeAnalyzer] [AnalysisConfigMonitor_ConfigChanged] [Thread: 1, 06:09:26.899]  Handling analysis config changed event...
[EslintBridgeAnalyzer] [RequireLinterUpdate] [Thread: 1, 06:09:26.899]  Waiting to acquire lock...
…
[EslintBridgeAnalyzer] [RequireLinterUpdate] [Thread: 1, 06:09:32.142]  Lock acquired. Running total: 5242ms

ugras-ergun-sonarsource:
fixes #3160

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
@@ -19,6 +19,7 @@
  */
 
 using System;
+using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
@@ -27,6 +28,7 @@ using SonarLint.VisualStudio.Core.Analysis;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Core.Suppression;
 using SonarLint.VisualStudio.Integration.UnitTests;
+using static SonarLint.VisualStudio.Integration.UnitTests.NoOpThreadHandler;
 
 namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
 {
@@ -70,8 +72,10 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
             builder.TestSubject.ConfigChanged += eventHandler.Object;
 
             builder.SimulateUserSettingsChanged();
+            
 
             eventHandler.Verify(x=> x(builder.TestSubject, EventArgs.Empty), Times.Once);
+            builder.AssertSwitchedToBackgroundThread();
         }
 
         [TestMethod]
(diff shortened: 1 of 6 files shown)


src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
```diff
@@ -42,6 +44,7 @@ public void WhenUserSettingsChange_StandaloneMode_AnalysisIsRequested()
 
             // Should re-analyse
             builder.AssertAnalysisIsRequested();
+            builder.AssertSwitchedToBackgroundThread();
```
ugras-ergun-sonarsource, line 47: I didn't think adding another test for just checking on background task made sense. So I've added the checks in scenarios where it should run on background.

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -48,13 +52,22 @@ internal sealed class AnalysisConfigMonitor : IAnalysisConfigMonitor, IDisposabl
             IUserSettingsProvider userSettingsProvider, // reports changes to user settings.json
             IActiveSolutionBoundTracker activeSolutionBoundTracker, // reports changes to connected mode
             ISuppressedIssuesMonitor suppressedIssuesMonitor,
-            ILogger logger)
+            ILogger logger) : this(analysisRequester, userSettingsProvider, activeSolutionBoundTracker, suppressedIssuesMonitor, logger, new ThreadHandling())
```
ugras-ergun-sonarsource, line 55: Because I have minimal change test coverage fails because of these lines.

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -48,13 +52,22 @@ internal sealed class AnalysisConfigMonitor : IAnalysisConfigMonitor, IDisposabl
             IUserSettingsProvider userSettingsProvider, // reports changes to user settings.json
```
rita-sonarsource, line 52: Cosmetics: no point in repeating these comments in both ctors

rita-sonarsource (approved):

PR #3162, opened 2022-08-05
Run ActiveSolutionTracker_ActiveSolutionChanged on the background thread

### LINKED ISSUE #3161: UI freeze when closing a folder/solution after a JS/TS analysis was done
### Description

steps to reproduce:
1. Open a solution/folder that has a js/ts file
2. Analyze the file --> verify that eslint bridge has started
3. Close the solution/folder

Expected --> no lockup
Actual --> UI freezes for couple of seconds 

Notes:
* This is especially visible when opening a folder with a js file already opened (so the analysis kicks in immediately) and closing the folder right away -- the analysis takes 4 seconds to initialize the eslint-bridge, and the UI waits for it to finish before killing the process.

rita-sonarsource:
Fixes  #3161

### DIFF
--- a/src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
+++ b/src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
@@ -25,6 +25,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Infrastructure.VS;
 using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.TypeScript.EslintBridgeClient;
 using SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract;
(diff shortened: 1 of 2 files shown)


src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
```diff
@@ -150,12 +165,15 @@ private void AnalysisConfigMonitor_ConfigChanged(object sender, EventArgs e)
 
         private async Task StopServer()
         {
+            await threadHandling.SwitchToBackgroundThread();
```
rita-sonarsource, line 168: I can add a test that verifies that the event switches to a background thread, but I'm not sure how much value it will have.

duncanp-sonar: I think it's worth adding, particularly if we can check the order of invocations i.e. switch to background thread is called before the close.

duncanp-sonar: Actually, let's commit without the test so we can get a final product version to build.

duncanp-sonar (approved):

PR #3163, opened 2022-08-05
Added regression test for #3161

duncanp-sonar:


src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
```diff
@@ -309,6 +309,60 @@ public void OnSolutionChanged_EslintBridgeClientStopped_OnlyIfSolutionClosed(boo
             client.Verify(x => x.Close(), Times.Exactly(expectedCallCount));
         }
 
+        [TestMethod]
```
duncanp-sonar, line 312: I've added two tests with differing implementations. We only need one of them. Which one do we want to keep?

The first checks the order of execution "manually", by recording the invocations in a list and checking the result.

The second uses `MockSequence`.

Which one should we keep?

rita-sonarsource: Personally I think the callOrder list is simple and easy to understand. WDYT?

rita-sonarsource (approved):

src/TypeScript.UnitTests/Analyzer/EslintBridgeAnalyzerTests.cs
duncanp-sonar: That's the one I prefer.

The `MockSequence` one was harder to write, requires us to check things we don't care about (e.g. the `ThrowIfOnUIThread` call), and is more fragile (for example, if you forget the `Strict` behaviour the test might still pass even if methods are called in a different order). In fact, it seems like creator the `MockSequence` [isn't happy](https://stakx.github.io/dissecting-moq4/2020/12/26/mock-sequence.html) with how they work.

PR #3177, opened 2022-08-12
Embed SonarC# 8.43 and SonarVB 8.43

### LINKED ISSUE #3148: Embed SonarC# 8.43 and SonarVB 8.43
[8.43] Hardening
[Release notes](https://github.com/SonarSource/sonar-dotnet/releases/tag/8.43.0.51858)

[8.42] Support tuple deconstruction; Fix 2 FNs and one FP; Other improvements
[Release notes](https://github.com/SonarSource/sonar-dotnet/releases/tag/8.42.0.51121)

[8.41] Azure function rules
[Release notes](https://github.com/SonarSource/sonar-dotnet/releases/tag/8.41.0.50478)

duncanp-sonar:
Fixes #3148


duncanp-sonar:
FYI I added a couple of files to the [standalone sample project](https://github.com/SonarSource/sonarlint-visualstudio-sampleprojects/commit/1640338a1bc2cdcb7ad2716e2d8307a3498287fb) to check one of the new VB.NET rules is firing.

This took longer than expected because the rule wasn't being reported. It turns out that both new rules are off by default, so need to be turned on in a ruleset or `.editorconfig` file.

How can we tell if a rule should be off by default or not? It appears that you can work it out from the rules specification page: if the `defaultQualityProfiles` element is empty then it's off by default. Otherwise, it will be in the `Sonar way` profile e.g. [S4060](https://sonarsource.github.io/rspec/#/rspec/S4/vbnet).

rita-sonarsource (approved):

rita-sonarsource:
> It turns out that both new rules are off by default

I thought that all new rules have to be enabled by default... thanks for checking 👍

PR #3178, opened 2022-08-15
Add notification service interfaces 

### LINKED ISSUE #3179: Add notification service interfaces
Add interfaces for the notification service

ugras-ergun-sonarsource:
fixes #3179


ugras-ergun-sonarsource:
It fails because of coverage. I don't think there are any meaningful tests to write, and I don't think it's worth writing just to make SC happy.

src/TypeScript/Notifications/INodeExpiredNotificationService.cs
```diff
@@ -0,0 +1,27 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.TypeScript.Notifications
+{
+    public interface INodeExpiredNotificationService
```
duncanp-sonar, line 23: Node hasn't "expired" - for whatever reason we haven't found a sufficiently up to date version.
`IUnsupportedNodeVersionNotificationService` is longer but more accurate.

Thinking ahead, in a few months time the JS/TS bubble are going to raise the minimum required version again. This means we might need show another notification to the same user. That suggests we'd need a separate notification id, since we'd still want to show the new notification even if the user has said "don't show me again" for the previous notification.

However, we don't need to decide how to do that now.

src/Core/SonarLintNotifications/NotificationAction.cs
```diff
@@ -0,0 +1,40 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+using System;
+
+namespace SonarLint.VisualStudio.Core.SonarLintNotifications
+{
+    public interface INotificationAction
+    {
+        string Message { get; }
```
duncanp-sonar, line 26: `Message` doesn't seem quite right, and it's confusing that it's the same as the field on `INotification`.

Something like `CommandText` might be better since it's describing what appears on the shortcut/button.

src/Core/SonarLintNotifications/NotificationAction.cs
```diff
@@ -0,0 +1,40 @@
...
+        Action Action { get; }
+    }
+    public class NotificationAction : INotificationAction
```
duncanp-sonar, line 29: Very minor - missing space between the types.

rita-sonarsource (approved):
LGTM, I agree with the comment regarding the node class name.

src/Core/SonarLintNotifications/NotificationAction.cs
```diff
@@ -0,0 +1,40 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+using System;
+
+namespace SonarLint.VisualStudio.Core.SonarLintNotifications
+{
+    public interface INotificationAction
+    {
+        string Message { get; }
+        Action Action { get; }
```
duncanp-sonar, line 27: The gold bar can show buttons and hyperlinks. This API doesn't give a way for the caller to specify which.

Do we think it needs to? Stylistically, MS-VS gold bars seem to use a mix of buttons and links, but my impression is that the newer gold bars just use links.

duncanp-sonar (commented):
See inline - minor naming changes.

duncanp-sonar (approved):

src/Core/SonarLintNotifications/NotificationAction.cs
ugras-ergun-sonarsource: It might make sense to show a button for main actions. For instance for nodejs we can show something like "install node" and making that button makes sense while "more info" and "don't show this again" would be better as links.

PR #3188, opened 2022-08-16
Support multiple buttons in the infobar

rita-sonarsource:
Fixes [#3189](https://github.com/SonarSource/sonarlint-visualstudio/issues/3189)

### DIFF
--- a/src/Core/InfoBar/IInfoBar.cs
+++ b/src/Core/InfoBar/IInfoBar.cs
@@ -30,7 +30,7 @@ namespace SonarLint.VisualStudio.Core.InfoBar
         /// <summary>
         /// Click event
         /// </summary>
-        event EventHandler ButtonClick;
+        event EventHandler<InfoBarButtonClickedEventArgs> ButtonClick;
 
         /// <summary>
         /// Closed event
@@ -41,5 +41,15 @@ namespace SonarLint.VisualStudio.Core.InfoBar
         /// Close the info bar
         /// </summary>
         void Close();
+    }
+
+    public class InfoBarButtonClickedEventArgs : EventArgs
+    {
+        public string ClickedButtonText { get; }
+
+        public InfoBarButtonClickedEventArgs(string clickedButtonText)
+        {
+            ClickedButtonText = clickedButtonText;
+        }
     }
 }
--- a/src/Core/InfoBar/IInfoBarManager.cs
+++ b/src/Core/InfoBar/IInfoBarManager.cs
@@ -52,7 +52,7 @@ namespace SonarLint.VisualStudio.Core.InfoBar
         /// <remarks>
         /// The created info bar will have multiple buttons with a "hyperlink" style
         /// </remarks>
-        IInfoBar AttachInfoBarWithButtons(Guid toolWindowGuid, string message, IReadOnlyList<string> buttonTexts, SonarLintImageMoniker imageMoniker);
+        IInfoBar AttachInfoBarWithButtons(Guid toolWindowGuid, string message, IEnumerable<string> buttonTexts, SonarLintImageMoniker imageMoniker);
 
         /// <summary>
         /// Detaches an <see cref="IInfoBar"/> from its tool window
--- a/src/Core/InfoBar/IInfoBarManager.cs
+++ b/src/Core/InfoBar/IInfoBarManager.cs
@@ -19,6 +19,7 @@
  */
 
 using System;
+using System.Collections.Generic;
 
 namespace SonarLint.VisualStudio.Core.InfoBar
 {
@@ -44,9 +45,20 @@ namespace SonarLint.VisualStudio.Core.InfoBar
         IInfoBar AttachInfoBarWithButton(Guid toolWindowGuid, string message, string buttonText, SonarLintImageMoniker imageMoniker);
 
         /// <summary>
+        /// Attach an info bar with a message and multiple buttons to an existing tool window
+        /// </summary>
+        IInfoBar AttachInfoBarWithButtons(Guid toolWindowGuid, string message, IReadOnlyList<string> buttonTexts, ButtonStyle buttonStyle, SonarLintImageMoniker imageMoniker);
+
+        /// <summary>
         /// Detaches an <see cref="IInfoBar"/> from its tool window
         /// </summary>
         /// <param name="currentInfoBar">Instance of <see cref="IInfoBar"/> created by <see cref="AttachInfoBar(Guid, string, SonarLintImageMoniker)"/></param>
         void DetachInfoBar(IInfoBar currentInfoBar);
+    }
+
+    public enum ButtonStyle
+    {
+        Hyperlink,
+        Button
     }
 }
--- a/src/Core/InfoBar/IInfoBarManager.cs
+++ b/src/Core/InfoBar/IInfoBarManager.cs
@@ -37,28 +37,27 @@ namespace SonarLint.VisualStudio.Core.InfoBar
         /// <summary>
         /// Attach an info bar with a message and a button to an existing tool window
         /// </summary>
+        /// <remarks>
+        /// The created info bar will have one button with a "button" style
+        /// </remarks>
         /// <param name="toolWindowGuid">Tool window Guid</param>
         /// <param name="message">Message to show on the info bar</param>
         /// <param name="buttonText">The button text</param>
         /// <param name="imageMoniker">Image</param>
-        /// <returns><see cref="IInfoBar"/></returns>
         IInfoBar AttachInfoBarWithButton(Guid toolWindowGuid, string message, string buttonText, SonarLintImageMoniker imageMoniker);
 
         /// <summary>
         /// Attach an info bar with a message and multiple buttons to an existing tool window
         /// </summary>
-        IInfoBar AttachInfoBarWithButtons(Guid toolWindowGuid, string message, IReadOnlyList<string> buttonTexts, ButtonStyle buttonStyle, SonarLintImageMoniker imageMoniker);
+        /// <remarks>
+        /// The created info bar will have multiple buttons with a "hyperlink" style
+        /// </remarks>
+        IInfoBar AttachInfoBarWithButtons(Guid toolWindowGuid, string message, IReadOnlyList<string> buttonTexts, SonarLintImageMoniker imageMoniker);
 
         /// <summary>
         /// Detaches an <see cref="IInfoBar"/> from its tool window
         /// </summary>
         /// <param name="currentInfoBar">Instance of <see cref="IInfoBar"/> created by <see cref="AttachInfoBar(Guid, string, SonarLintImageMoniker)"/></param>
         void DetachInfoBar(IInfoBar currentInfoBar);
-    }
-
-    public enum ButtonStyle
-    {
-        Hyperlink,
-        Button
     }
 }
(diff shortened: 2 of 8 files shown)


src/Core/InfoBar/IInfoBar.cs
```diff
@@ -42,4 +42,14 @@ public interface IInfoBar
         /// </summary>
         void Close();
     }
+
+    public class InfoBarButtonClickedEventArgs : EventArgs
+    {
+        public string ClickedButtonText { get; }
```
rita-sonarsource, line 48: From the original event we have access to IVsInfoBarActionItem but:
1) it's a VS object and this is Core
2) It doesn't actually have anything useful in it besides the text

https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsinfobaractionitem?view=visualstudiosdk-2022

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -224,7 +241,7 @@ public void Close()
             #region IVsInfoBarUIEvents
             void IVsInfoBarUIEvents.OnActionItemClicked(IVsInfoBarUIElement infoBarUIElement, IVsInfoBarActionItem actionItem)
             {
-                this.ButtonClick?.Invoke(this, EventArgs.Empty);
+                this.ButtonClick?.Invoke(this, new InfoBarButtonClickedEventArgs(actionItem.Text));
```
rita-sonarsource, line 244: The only useful information is the button text
https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsinfobaractionitem?view=visualstudiosdk-2022

src/TestInfrastructure/Framework/ConfigurableVsInfoBarUIElement.cs
```diff
@@ -116,7 +117,7 @@ public bool IsClosed
 
         public void SimulateClickEvent(IVsInfoBarActionItem item = null)
         {
-            this.sinks.Values.ToList().ForEach(v => v.OnActionItemClicked(this, item));
+            this.sinks.Values.ToList().ForEach(v => v.OnActionItemClicked(this, item ?? Mock.Of<IVsInfoBarActionItem>()));
```
rita-sonarsource, line 120: Avoiding the null reference in the product code that does item.Text

src/Core/InfoBar/IInfoBarManager.cs
```diff
@@ -43,10 +44,21 @@ public interface IInfoBarManager
         /// <returns><see cref="IInfoBar"/></returns>
         IInfoBar AttachInfoBarWithButton(Guid toolWindowGuid, string message, string buttonText, SonarLintImageMoniker imageMoniker);
 
+        /// <summary>
+        /// Attach an info bar with a message and multiple buttons to an existing tool window
+        /// </summary>
+        IInfoBar AttachInfoBarWithButtons(Guid toolWindowGuid, string message, IReadOnlyList<string> buttonTexts, ButtonStyle buttonStyle, SonarLintImageMoniker imageMoniker);
```
duncanp-sonar, line 50: Is it worth allowing a `ButtonStyle` here? It's just something extra to test, and based on this morning's discussion we'll only use hyperlinks.

src/Core/InfoBar/IInfoBar.cs
duncanp-sonar: The ActionContext might be useful as a way of passing arbitrary context data, but we don't need it now.

src/Core/InfoBar/IInfoBarManager.cs
rita-sonarsource: Regarding different styles for buttons: that would look horrible :D
I'll drop the param and default to hyperlink.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -114,6 +117,7 @@ public void InfoBarManager_AttachInfoBar()
             infoBarUI.Model.TextSpans.GetSpan(0).Text.Should().Be("Hello");
             infoBarUI.Model.ActionItems.Count.Should().Be(1);
             infoBarUI.Model.ActionItems.GetItem(0).Text.Should().Be("world");
+            infoBarUI.Model.ActionItems.GetItem(0).IsButton.Should().BeTrue();
```
rita-sonarsource, line 120: The old method should create infobars with button style

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -202,10 +206,118 @@ public void InfoBarManager_DetachInfoBar()
...
+            infoBarUI.Model.TextSpans.Count.Should().Be(1);
+            infoBarUI.Model.TextSpans.GetSpan(0).Text.Should().Be("message");
+
+            infoBarUI.Model.ActionItems.Count.Should().Be(3);
+
+            infoBarUI.Model.ActionItems.GetItem(0).Text.Should().Be("button1");
+            infoBarUI.Model.ActionItems.GetItem(1).Text.Should().Be("button2");
+            infoBarUI.Model.ActionItems.GetItem(2).Text.Should().Be("button3");
+
+            infoBarUI.Model.ActionItems.GetItem(0).IsButton.Should().BeFalse();
```
rita-sonarsource, line 264: The new method should create infobars with hyperlink style

src/TestInfrastructure/Framework/ConfigurableVsInfoBarUIElement.cs
```diff
@@ -132,4 +133,19 @@ public IVsInfoBar Model
 
         #endregion Test helpers
     }
-}
\ No newline at end of file
+
+    public class ConfigurableVsInfoBarActionItem : IVsInfoBarActionItem
```
rita-sonarsource, line 137: For some reason I'm getting failures on 2019 when I try to mock IVsInfoBarActionItem:
```
Test method SonarLint.VisualStudio.Integration.UnitTests.InfoBarManagerTests.InfoBarManager_AttachInfoBar threw exception:
System.MissingMethodException: Error: Missing method 'instance string [SonarLint.2019] Microsoft.VisualStudio.Shell.Interop.IVsInfoBarActionItem::get_Text()' from class 'Castle.Proxies.IVsInfoBarActionItemProxy'.
```

ugras-ergun-sonarsource (approved):

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
duncanp-sonar: Perhaps call this out in a comment in the manager class? It's not obvious that it's being done to preserve the legacy behaviour.

src/Core/InfoBar/IInfoBarManager.cs
```diff
@@ -36,13 +37,23 @@ public interface IInfoBarManager
...
-        /// <returns><see cref="IInfoBar"/></returns>
         IInfoBar AttachInfoBarWithButton(Guid toolWindowGuid, string message, string buttonText, SonarLintImageMoniker imageMoniker);
 
+        /// <summary>
+        /// Attach an info bar with a message and multiple buttons to an existing tool window
+        /// </summary>
+        /// <remarks>
+        /// The created info bar will have multiple buttons with a "hyperlink" style
+        /// </remarks>
+        IInfoBar AttachInfoBarWithButtons(Guid toolWindowGuid, string message, IReadOnlyList<string> buttonTexts, SonarLintImageMoniker imageMoniker);
```
duncanp-sonar, line 55: `IEnumerable` would do instead of `IReadOnlyList`, wouldn't it?
`IEnumerable` is read-only and gives in-order access, which is all we need.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -57,7 +59,22 @@ public IInfoBar AttachInfoBarWithButton(Guid toolWindowGuid, string message, str
...
+        }
+
+        public IInfoBar AttachInfoBarWithButtons(Guid toolWindowGuid, string message, IReadOnlyList<string> buttonTexts, SonarLintImageMoniker imageMoniker)
+        {
+            if (string.IsNullOrWhiteSpace(message))
+            {
+                throw new ArgumentNullException(nameof(message));
+            }
+
+            if (buttonTexts == null)
```
duncanp-sonar, line 72: This checks `null` but not `empty`; and empty list will crash in line 77.

Does it make sense to have a gold bar with no additional links, just dismiss?
I don't think it does at the moment (i.e. I can't think of a good example), so perhaps we should fail for an empty list too.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -144,12 +161,17 @@ private static bool TryCreateInfoBarUI(IVsInfoBarUIFactory infoBarUIFactory, IVs
...
-            var vsImageMoniker = new ImageMoniker {Guid = imageMoniker.Guid, Id = imageMoniker.Id};
+            var vsImageMoniker = new ImageMoniker { Guid = imageMoniker.Guid, Id = imageMoniker.Id };
+
+            var actionItems = buttonTexts.Select(x =>
+                buttonStyle == ButtonStyle.Button
+                    ? (IVsInfoBarActionItem) new InfoBarButton(x)
+                    : new InfoBarHyperlink(x));
 
             return new InfoBarModel(message,
-                actionItems: buttonText != null ? new[] { new InfoBarButton(buttonText) } : new IVsInfoBarActionItem[0],
```
duncanp-sonar, line 152: I'm guessing we didn't previously have any callers that supplied `null`?

src/Integration.Vsix/InfoBar/InfoBarManager.cs
ugras-ergun-sonarsource: Why would it crash?

duncanp-sonar: You're right, it won't.

The question about whether it makes sense to allow an empty list remains.
If we're allowing it there should be a test for it.

ugras-ergun-sonarsource: I think it's reasonable to think that at least "do not show this again" should be there all the time.

src/TestInfrastructure/Framework/ConfigurableVsInfoBarUIElement.cs
duncanp-sonar: That's weird; `Castles.Proxies` is part of the Mock framework isn't it? Why is the Mock framework involved?

src/Integration.Vsix/InfoBar/InfoBarManager.cs
rita-sonarsource: Technically if you called `AttachInfoBar` instead of `AttachInfoBarWithButton` then this could would be called with null.
But no one calls `AttachInfoBar` in the code, so no :)

src/TestInfrastructure/Framework/ConfigurableVsInfoBarUIElement.cs
rita-sonarsource: I don't know and I still get failures on 2019 :(

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
rita-sonarsource: The comment is on the interface itself. WDYT?

src/Integration.Vsix/InfoBar/InfoBarManager.cs
rita-sonarsource: The method `AttachInfoBar` uses this implementation as well, and asserts that it runs without buttons.

duncanp-sonar (approved):

PR #3190, opened 2022-08-16
Add Notification Storage Interfaces 

ugras-ergun-sonarsource:


src/Core/SonarLintNotifications/INotificationSettingStorage.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
```
duncanp-sonar, line 1: Missing copyright headers.
(I would have expected a code smell for this, but for some reason there isn't one)

src/Core/SonarLintNotifications/NotificationSetting.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SonarLint.VisualStudio.Core.SonarLintNotifications
+{
+    public interface INotificationSetting
+    {
+        string Key { get; }
+        bool DoNotShow { get; }
```
duncanp-sonar, line 12: Is the notification setting likely to have more properties that just `DoNotShow`?
I guess it might, but I can't think of any examples right now.

src/Core/SonarLintNotifications/INotificationSettingStorage.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        User,
+        VSInstance,
+        Solution
+    }
+
+    public interface INotificationSettingStorage
+    {
+        void SaveNotificationSetting(INotificationSetting setting, StorageLocation storageLocation);
+        void GetNotificationSetting(string notificationSettingKey, StorageLocation storageLocation);
```
duncanp-sonar, line 19: The `get` doesn't return anything.

src/Core/SonarLintNotifications/INotificationSettingStorage.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
duncanp-sonar, line 5: Are all of these `usings` required?

src/Core/SonarLintNotifications/NotificationSetting.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SonarLint.VisualStudio.Core.SonarLintNotifications
+{
+    public interface INotificationSetting
```
rita-sonarsource, line 9: Wouldn't it make sense to have a file that only contains "do not show" notifications?
What is the benefit of storing notifications with DoNotShow=false?
Also, double-negative is hard to understand.

src/Core/SonarLintNotifications/NotificationSetting.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        string Key { get; }
```
rita-sonarsource, line 11: Is the Key different from NotificationId in the notification object?

src/Core/SonarLintNotifications/INotificationSettingStorage.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace SonarLint.VisualStudio.Core.SonarLintNotifications
+{
+    public enum StorageLocation
```
rita-sonarsource, line 9: Do we plan to implement all of these strategies already?

src/Core/SonarLintNotifications/INotificationSettingStorage.cs
duncanp-sonar: The caller has to know where the data is stored or they won't find anything. I guess it doesn't matter here, since in the expected usage there will only be one caller using each notification key, and they will be calling `get` then `set`.

src/Core/SonarLintNotifications/INotificationSettingStorage.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        User,
+        VSInstance,
+        Solution
+    }
+
+    public interface INotificationSettingStorage
+    {
+        void SaveNotificationSetting(INotificationSetting setting, StorageLocation storageLocation);
```
rita-sonarsource, line 18: I'm not sure if Setting is the right word here. It sounds like a configuration.

How about a simpler API, only for dismissed notifications?

`void SaveBlockedNotification(string notificationId)`
`IEnumerable<string> GetBlockedNotifications()`

duncanp-sonar, line 18: Naming: `Save` and `Load`, or `Set` and `Get` (combining `Save` and `Get` looks a bit odd).

src/Core/SonarLintNotifications/NotificationSetting.cs
ugras-ergun-sonarsource: I also thought about it but couldn't really find a good example.

src/Core/SonarLintNotifications/NotificationSetting.cs
ugras-ergun-sonarsource: Yeah, it would make sense to have them same

src/Core/SonarLintNotifications/NotificationSetting.cs
ugras-ergun-sonarsource: I think there could be other info that we could store. I think it's better to save info than not saving. And for the name I couldn't find a better name. Maybe "DoNotShowSelected" to avoid double negative?

src/Core/SonarLintNotifications/INotificationSettingStorage.cs
```diff
@@ -0,0 +1,30 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarLint.VisualStudio.Core.SonarLintNotifications
+{
+    public interface IDisabledNotificationsStorage
+    {
+        void DisableNotification(string notificationId);
```
duncanp-sonar, line 27: The parameter names could just be `id` to match the other interfaces.

duncanp-sonar (approved):

PR #3192, opened 2022-08-18
Add null arg checks

rita-sonarsource:


src/Core/Notifications/NotificationAction.cs
```diff
@@ -32,8 +32,8 @@ public class NotificationAction : INotificationAction
     {
         public NotificationAction(string commandText, Action action)
```
rita-sonarsource, line 33: Just some sanity checks, encountered this during my tests for the notification service by mistake.

duncanp-sonar (approved):

PR #3193, opened 2022-08-18
Implement NotificationService

rita-sonarsource:
Fixes [#3194](https://github.com/SonarSource/sonarlint-visualstudio/issues/3194).

Supporting "do not show again" is a sizable chunk so I've created a new task for it and will do it in a separate PR.

### DIFF
--- a/src/Core/Notifications/INotificationService.cs
+++ b/src/Core/Notifications/INotificationService.cs
@@ -21,6 +21,7 @@
 using System;
 using System.Linq;
 using SonarLint.VisualStudio.Core.InfoBar;
+using SonarLint.VisualStudio.Integration;
 
 namespace SonarLint.VisualStudio.Core.Notifications
 {
@@ -38,13 +39,15 @@ namespace SonarLint.VisualStudio.Core.Notifications
 
         private readonly IInfoBarManager infoBarManager;
         private readonly IThreadHandling threadHandling;
+        private readonly ILogger logger;
 
         private Tuple<IInfoBar, INotification> activeNotification;
 
-        public NotificationService(IInfoBarManager infoBarManager, IThreadHandling threadHandling)
+        public NotificationService(IInfoBarManager infoBarManager, IThreadHandling threadHandling, ILogger logger)
         {
             this.infoBarManager = infoBarManager;
             this.threadHandling = threadHandling;
+            this.logger = logger;
         }
 
         public void ShowNotification(INotification notification)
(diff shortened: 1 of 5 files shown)


src/Core/SonarLintImageMoniker.cs
```diff
@@ -27,6 +27,11 @@ namespace SonarLint.VisualStudio.Core
     /// </summary>
     public readonly struct SonarLintImageMoniker
     {
+        /// <summary>
+        /// Based on the information inside sonarlint-visualstudio\src\IssueViz\Monikers.imagemanifest
+        /// </summary>
+        public static SonarLintImageMoniker OfficialSonarLintMoniker = new SonarLintImageMoniker(new Guid("19682eb3-4922-49e2-8f9b-e3edb43f656c"), 0);
```
rita-sonarsource, line 33: I'm not sure if this is the best place for it but it was either this or a random new constants class.

src/Core/Notifications/INotificationService.cs
```diff
@@ -18,10 +18,89 @@
...
+
 namespace SonarLint.VisualStudio.Core.Notifications
 {
-    public interface INotificationService
+    public interface INotificationService : IDisposable
     {
         void ShowNotification(INotification notification);
     }
+    
+    public sealed class NotificationService : INotificationService
```
rita-sonarsource, line 32: Implementation similar to https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/IssueViz.Security/OpenInIDE/Api/OpenInIDEFailureInfoBar.cs

src/Core/SonarLintImageMoniker.cs
rita-sonarsource: Also will fix the SC comment in the next commit

src/Core/Notifications/INotificationService.cs
```diff
@@ -18,10 +18,89 @@
...
+    {
+        /// <summary>
+        /// Taken from "ToolWindowGuids80.ErrorList"
+        /// </summary>
+        internal static readonly Guid ErrorListToolWindowGuid = new Guid("D78612C7-9962-4B83-95D9-268046DAD23A");
```
duncanp-sonar, line 37: The error list makes sense for this notification, but in the future I think we'll need to let the caller specify where it should appear.

src/Core/Notifications/INotificationService.cs
```diff
@@ -18,10 +18,89 @@
...
+        {
+            if (notification == null)
+            {
+                throw new ArgumentNullException(nameof(notification));
+            }
+            // todo: check if already shown
+
+            threadHandling.RunOnUIThread(() =>
+            {
+                RemoveExistingInfoBar();
```
duncanp-sonar, line 60: This will need more work later too. There will probably be occasions when we want to display more than one (different) notification at the same time.

src/Core/Notifications/INotificationService.cs
```diff
@@ -18,10 +18,89 @@
...
+
+                activeNotification = new Tuple<IInfoBar, INotification>(infoBar, notification); 
+                activeNotification.Item1.ButtonClick += CurrentInfoBar_ButtonClick;
+                activeNotification.Item1.Closed += CurrentInfoBar_Closed;
+            });
+        }
+
+        private void CurrentInfoBar_ButtonClick(object sender, InfoBarButtonClickedEventArgs e)
+        {
+            var matchingAction = activeNotification.Item2.Actions.FirstOrDefault(x => x.CommandText == e.ClickedButtonText);
```
duncanp-sonar, line 77: This will be on the UI thread, won't it?
If so, we should guard against unhandled exceptions so we don't crash VS.

The same applies to the "closed" handler, but it's less likely to throw since it's not performing an arbitrary task (the existing infobar code doesn't have error handling for the close).

src/Core/Notifications/INotificationService.cs
rita-sonarsource: Yep, this is kind of a "temp" solution. I don't want to overcomplicate it now.

src/Core/Notifications/INotificationService.cs
```diff
@@ -18,10 +18,89 @@
...
+        /// Taken from "ToolWindowGuids80.ErrorList"
+        /// </summary>
+        internal static readonly Guid ErrorListToolWindowGuid = new Guid("D78612C7-9962-4B83-95D9-268046DAD23A");
+
+        private readonly IInfoBarManager infoBarManager;
+        private readonly IThreadHandling threadHandling;
+
+        private Tuple<IInfoBar, INotification> activeNotification;
+
+        public NotificationService(IInfoBarManager infoBarManager, IThreadHandling threadHandling)
```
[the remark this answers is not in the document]
rita-sonarsource: Good question, it means the implementation would have to move to Integration (or we will need a MEF wrapper in Integration). I've been delaying the decision.. :) It doesn't make sense to make to move all this generic code into Integration just for MEF, but we have to MEF-import it because `InfoBarManager` needs VS stuff. Perhaps introducing MEF into Core would simplify things.

rita-sonarsource:
@duncanp-sonar I added error handling, please re-review

duncanp-sonar (approved):

Issue #3185, opened 2022-08-15
Send keep-alive to the eslint-bridge process


duncanp-sonar:
### Description

Send a keep-alive every five seconds to the eslint-bridge process to prevent it from terminating.
We start separate eslint-bridge instances for JavaScript and TypeScript, so this need to be done for both.

Depends on [SonarJS-#3131](https://github.com/SonarSource/SonarJS/issues/3131). The implementation details are in the ticket.

Note: this ticket is dependent on the current JS/TS sprint finishing and releasing v9.7 i.e. [this milestone](https://github.com/SonarSource/SonarJS/milestone/59).

victor-diez-sonarsource:
Hello @duncanp-sonar,

we will reuse the existing endpoint `/status`, instead of creating a new one just for the heartbeat, given that any request to any endpoint will reset the timeout.

PR #3198, opened 2022-08-19
Add GetAsync method to EslintBridgeHttpWrapper

duncanp-sonar:
Prep for #3185

### DIFF
--- a/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeHttpWrapperTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeHttpWrapperTests.cs
@@ -136,6 +136,67 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
             await act.Should().ThrowAsync<NotImplementedException>();
         }
 
+        [TestMethod]
+        public async Task GetAsync_ExecutesRequestsOnTheUrl()
+        {
+            Uri requestUri = null;
+
+            var httpMessageHandler = SetupHttpMessageHandler("some response", message =>
+            {
+                requestUri = message.RequestUri;
+            });
+
+            var testSubject = CreateTestSubject(httpMessageHandler);
+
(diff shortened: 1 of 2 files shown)


src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeHttpWrapperTests.cs
```diff
@@ -136,6 +136,67 @@ public async Task PostAsync_Exception_ExceptionNotCaught()
             await act.Should().ThrowAsync<NotImplementedException>();
         }
 
+        [TestMethod]
+        public async Task GetAsync_ExecutesRequestsOnTheUrl()
```
duncanp-sonar, line 140: These three tests are copies of the equivalent POST ones, but with `content` argument removed.

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
```diff
@@ -55,13 +57,25 @@ public async Task<string> PostAsync(Uri serverEndpoint, object request, Cancella
...
+        public async Task<string> GetAsync(Uri serverEndpoint, CancellationToken cancellationToken)
+        {
+            logger.LogDebug("[eslint-bridge GET] Endpoint: {0}{1}", serverEndpoint, Environment.NewLine);
+
+            var response = await httpClient.GetAsync(serverEndpoint, cancellationToken);
+            var responseString = await response.Content.ReadAsStringAsync();
+
+            logger.LogDebug("[eslint-bridge GET] Endpoint: {0}, response:{1}{2}", serverEndpoint, Environment.NewLine, responseString);
 
             return responseString;
```
duncanp-sonar, line 80: I looked at whether I could have a common helper method that `PostAsync` and `GetAsync` could delegate the work too, but the methods signatures are different.

duncanp-sonar:
The eslint-bridge keep-alive message needs GET rather than POST.

rita-sonarsource (approved):

PR #3199, opened 2022-08-22
Add locking to NotificationService

rita-sonarsource:
I'm not 100% sure this is PR is even needed. Everything should run on the UI thread. Perhaps the only caveat is Dispose, but that would be called by VS during shutdown so it won't matter anyway.

Perhaps for readability, it's better not to merge it, I don't know it. I'm not sure if a problem can really happen.

### DIFF
--- a/src/Core/Notifications/INotificationService.cs
+++ b/src/Core/Notifications/INotificationService.cs
@@ -37,6 +37,8 @@ namespace SonarLint.VisualStudio.Core.Notifications
         /// </summary>
         internal static readonly Guid ErrorListToolWindowGuid = new Guid("D78612C7-9962-4B83-95D9-268046DAD23A");
 
+        private static readonly object Locker = new object();
+
         private readonly IInfoBarManager infoBarManager;
         private readonly IThreadHandling threadHandling;
         private readonly ILogger logger;
@@ -60,24 +62,10 @@ namespace SonarLint.VisualStudio.Core.Notifications
 
             threadHandling.RunOnUIThread(() =>
             {
-                try
+                lock (Locker)
                 {
                     RemoveExistingInfoBar();
-
-                    var buttonTexts = notification.Actions.Select(x => x.CommandText).ToArray();
-
-                    var infoBar = infoBarManager.AttachInfoBarWithButtons(ErrorListToolWindowGuid,
-                        notification.Message,
-                        buttonTexts,
-                        SonarLintImageMoniker.OfficialSonarLintMoniker);
-
-                    activeNotification = new Tuple<IInfoBar, INotification>(infoBar, notification);
-                    activeNotification.Item1.ButtonClick += CurrentInfoBar_ButtonClick;
-                    activeNotification.Item1.Closed += CurrentInfoBar_Closed;
-                }
-                catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
-                {
-                    logger.WriteLine(CoreStrings.Notifications_FailedToDisplay, notification.Id, ex);
+                    ShowInfoBar(notification);
                 }
             });
         }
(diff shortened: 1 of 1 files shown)


src/Core/Notifications/INotificationService.cs
```diff
@@ -60,24 +62,10 @@ public void ShowNotification(INotification notification)
 
             threadHandling.RunOnUIThread(() =>
             {
-                try
+                lock (Locker)
                 {
                     RemoveExistingInfoBar();
-
-                    var buttonTexts = notification.Actions.Select(x => x.CommandText).ToArray();
```
rita-sonarsource, line 67: Extracted as-is

src/Core/Notifications/INotificationService.cs
```diff
@@ -60,24 +62,10 @@ public void ShowNotification(INotification notification)
...
-                    var infoBar = infoBarManager.AttachInfoBarWithButtons(ErrorListToolWindowGuid,
-                        notification.Message,
-                        buttonTexts,
-                        SonarLintImageMoniker.OfficialSonarLintMoniker);
-
-                    activeNotification = new Tuple<IInfoBar, INotification>(infoBar, notification);
-                    activeNotification.Item1.ButtonClick += CurrentInfoBar_ButtonClick;
-                    activeNotification.Item1.Closed += CurrentInfoBar_Closed;
-                }
-                catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
```
rita-sonarsource, line 78: Minor change: the catch used to cover both RemoveExisting and showing the new, but there is a try-catch inside Remove. So I changed the scope to cover Show.

src/Core/Notifications/INotificationService.cs
```diff
@@ -86,7 +74,7 @@ private void CurrentInfoBar_ButtonClick(object sender, InfoBarButtonClickedEvent
         {
             try
             {
-                var matchingAction = activeNotification.Item2.Actions.FirstOrDefault(x => x.CommandText == e.ClickedButtonText);
+                var matchingAction = activeNotification?.Item2.Actions.FirstOrDefault(x => x.CommandText == e.ClickedButtonText);
```
rita-sonarsource, line 77: I was thinking of a scenario where the user would click on a button and before the callback is scheduled, we replace the infobar with a new notification.
I don't think it can happen since both will be on the UI thread, so the callback must be executed before we can access the UI thread to replace the infobar, and while we're replacing, the thread is blocked so the user cannot click. 
But I'm paranoid so I thought it might be worth adding the check here anyway...

src/Core/Notifications/INotificationService.cs
```diff
@@ -124,9 +115,33 @@ private void RemoveExistingInfoBar()
             }
         }
 
+        private void ShowInfoBar(INotification notification)
```
duncanp-sonar, line 118: I think this is neater as a separate function.

src/Core/Notifications/INotificationService.cs
duncanp-sonar:
```
I don't think it can happen since both will be on the UI thread, so the callback must be executed before we can access the UI thread to replace the infobar
```

In general, it is possible for a thread to [temporarily yield control of the UI thread](https://github.com/microsoft/vs-threading/blob/main/doc/cookbook_vs.md#how-to-switch-to-or-use-the-ui-thread-with-a-specific-priority), which would allow other code to execute on the UI thread. In this case, our code doesn't yield control, but we don't know what happens inside any VS-code we call.

However, I don't think we need to worry about it at the moment. So far, we've only thought about displaying a single notification at a time, and there is only one possible notification we could show. To make the service more general, we would need to think the UX for displaying more than one notification at a time. Locking might well be part of that.

rita-sonarsource: Even for a single notification, it can be triggered by multiple analysis failures. (which gives me an idea for another task: not to display a notification of a notification of same id is currently shown)

duncanp-sonar: Yep, true.

Not displaying the same notification if it already being shown: there is already a more restrictive requirement in the spec document: _Only one instance of the notification should be shown per VS session._ Implementing that will require thread safety/state management.

You could either commit this change as is and tweak the locking as necessary when implementing "one notification per session", or add locking as part of that feature - do whichever is easiest for you.

duncanp-sonar (approved):

duncanp-sonar:
@rita-sonarsource, ia this PR still required?

PR #3200, opened 2022-08-22
Don't re-display same notification

rita-sonarsource:
Fixes [#3205](https://github.com/SonarSource/sonarlint-visualstudio/issues/3205)

### DIFF
--- a/src/Core/Notifications/INotificationService.cs
+++ b/src/Core/Notifications/INotificationService.cs
@@ -19,9 +19,11 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using SonarLint.VisualStudio.Core.InfoBar;
 using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.Helpers;
 
 namespace SonarLint.VisualStudio.Core.Notifications
 {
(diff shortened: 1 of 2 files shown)


src/Core/Notifications/INotificationService.cs
```diff
@@ -56,29 +56,17 @@ public void ShowNotification(INotification notification)
             {
                 throw new ArgumentNullException(nameof(notification));
             }
-            // todo: check if already shown
+
+            if (activeNotification?.Item2.Id == notification.Id)
```
rita-sonarsource, line 60: Minor optimization: the "analysis failed" notification can be triggered multiple times, there is no reason to block the UI thread each time.

src/Core/Notifications/INotificationService.cs
```diff
@@ -124,6 +112,27 @@ private void RemoveExistingInfoBar()
             }
         }
 
+        private void ShowInfoBar(INotification notification)
```
rita-sonarsource, line 115: Extracted as-is

src/Core/Notifications/INotificationService.cs
ugras-ergun-sonarsource: It makes sense for this notification but can we have a scenario where we need to trigger it multiple scenario? (I can't think of any)
Also instead of using Item2 maybe we should use field names for the tuple to increase readibility.

duncanp-sonar: As mentioned [in the other PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/3199#discussion_r951201269), there is a more restrictive check that makes this one redundant ("_Only one instance of the notification should be shown per VS session_").

PR #3202, opened 2022-08-22
Add DoNotShowAgainNotificationAction

rita-sonarsource:


src/Core/Notifications/DoNotShowAgainNotificationAction.cs
```diff
@@ -0,0 +1,51 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarLint.VisualStudio.Core.Notifications
+{
+    public interface IDoNotShowAgainNotificationAction : INotificationAction
```
rita-sonarsource, line 25: As we discussed offline, it is currently unclear who would create this action and attach it to the notification. 

Perhaps we would need a decorator or an extension method, but we can temporarily have the caller use it.

src/Core/Notifications/DoNotShowAgainNotificationAction.cs
```diff
@@ -0,0 +1,51 @@
...
+    public class DoNotShowAgainNotificationAction : IDoNotShowAgainNotificationAction
+    {
+        private readonly IDisabledNotificationsStorage disabledNotificationsStorage;
+
+        public DoNotShowAgainNotificationAction(IDisabledNotificationsStorage disabledNotificationsStorage)
+        {
+            this.disabledNotificationsStorage = disabledNotificationsStorage;
+        }
+
+        public string CommandText => "Do not show again";
```
duncanp-sonar, line 38: The wording in the VS gold bars is more informal e.g. `Don't report this extension again`.
Something like `Don't show this again` would be closer to that VS style.

It's a UI string, so it should be in a resource file.

duncanp-sonar (approved):
Minor comments on the wording of string

ugras-ergun-sonarsource (approved):

PR #3204, opened 2022-08-22
Added keep-alive handler for eslint-bridge

### LINKED ISSUE #3185: Send keep-alive to the eslint-bridge process

duncanp-sonar:
Fixes #3185

### DIFF
--- a/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
@@ -25,13 +25,15 @@ using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using Newtonsoft.Json;
+using SonarLint.VisualStudio.Core.SystemAbstractions;
+using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.TypeScript.EslintBridgeClient;
 using SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract;
 
 namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
 {
     [TestClass]
-    public class EslintBridgeClientTests
+    public partial class EslintBridgeClientTests
     {
         private const int ServerPort = 1234;
         private const string AnalyzeEndpoint = "dummy-analyze";
@@ -269,14 +271,18 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
             eslintBridgeProcess.Verify(x => x.Dispose(), Times.Never);
         }
 
-        private TypeScript.EslintBridgeClient.EslintBridgeClient CreateTestSubject(IEslintBridgeHttpWrapper httpWrapper = null,
+        private static TypeScript.EslintBridgeClient.EslintBridgeClient CreateTestSubject(IEslintBridgeHttpWrapper httpWrapper = null,
             IAnalysisConfiguration analysisConfiguration = null,
-            IEslintBridgeProcess eslintBridgeProcess = null)
+            IEslintBridgeProcess eslintBridgeProcess = null,
+            ITimer timer = null)
         {
             analysisConfiguration ??= Mock.Of<IAnalysisConfiguration>();
             eslintBridgeProcess ??= SetupServerProcess().Object;
+            httpWrapper ??= Mock.Of<IEslintBridgeHttpWrapper>();
+            timer ??= new Mock<ITimer>().Object;
 
-            return new TypeScript.EslintBridgeClient.EslintBridgeClient(AnalyzeEndpoint, eslintBridgeProcess, httpWrapper, analysisConfiguration);
+            return new TypeScript.EslintBridgeClient.EslintBridgeClient(AnalyzeEndpoint, eslintBridgeProcess, httpWrapper, analysisConfiguration,
+                timer, Mock.Of<ILogger>());
         }
 
         private static Mock<IEslintBridgeHttpWrapper> SetupHttpWrapper(string endpoint, string response = "OK!", Action<object> assertReceivedRequest = null)
@@ -291,7 +297,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
             return httpWrapper;
         }
 
-        private Mock<IEslintBridgeProcess> SetupServerProcess(bool isRunning = false)
+        private static Mock<IEslintBridgeProcess> SetupServerProcess(bool isRunning = false)
         {
             var serverProcess = new Mock<IEslintBridgeProcess>();
             serverProcess.Setup(x => x.Start()).ReturnsAsync(ServerPort);
--- a/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
+++ b/src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
@@ -217,6 +217,17 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
         }
 
         [TestMethod]
+        public void Dispose_DisposesEslintBridgeKeepAlive()
+        {
+            var keepAlive = new Mock<IEslintBridgeKeepAlive>();
+
+            var testSubject = CreateTestSubject(keepAlive: keepAlive.Object);
+            testSubject.Dispose();
+
+            keepAlive.Verify(x => x.Dispose(), Times.Once);
+        }
+
+        [TestMethod]
         public async Task Close_StopsEslintBridgeProcess()
         {
             var eslintBridgeProcess = new Mock<IEslintBridgeProcess>();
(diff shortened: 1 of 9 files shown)


src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
```diff
@@ -31,7 +31,7 @@
 namespace SonarLint.VisualStudio.TypeScript.UnitTests.EslintBridgeClient
 {
     [TestClass]
-    public class EslintBridgeClientTests
+    public partial class EslintBridgeClientTests
```
duncanp-sonar, line 34: Not required. I'll remove it.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeClientTests.cs
```diff
@@ -269,14 +280,18 @@ public async Task Close_FailsToCallClose_EslintBridgeProcessStillStopped()
             eslintBridgeProcess.Verify(x => x.Dispose(), Times.Never);
         }
 
-        private TypeScript.EslintBridgeClient.EslintBridgeClient CreateTestSubject(IEslintBridgeHttpWrapper httpWrapper = null,
+        private static TypeScript.EslintBridgeClient.EslintBridgeClient CreateTestSubject(IEslintBridgeHttpWrapper httpWrapper = null,
```
duncanp-sonar, line 283: Miscellaneous: couple of helper methods could be `static`.

src/TypeScript.UnitTests/EslintBridgeClient/EslintBridgeKeepAliveTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            // the exception is thrown.
+            Func<Task> act = async () => await context.DirectlyInvokeTimerHandlerAsync();
+
+            act.Should().ThrowExactly<StackOverflowException>().And.Message.Should().Be("this is a test");
+
+            // Should restart the timer
+            context.CheckOnlyExpectedMethods("Stop", "Start");
+        }
+
+        private class KeepAliveWorkflowContext
```
duncanp-sonar, line 127: There are several tests that need to check the order of method invocation. This helper class hides the setup-ugliness so the tests themselves can be simpler.

src/TypeScript/EslintBridgeClient/EslintBridgeHttpWrapper.cs
```diff
@@ -70,12 +70,12 @@ public async Task<string> PostAsync(Uri serverEndpoint, object request, Cancella
 
         public async Task<string> GetAsync(Uri serverEndpoint, CancellationToken cancellationToken)
         {
-            logger.LogDebug("[eslint-bridge GET] Endpoint: {0}{1}", serverEndpoint, Environment.NewLine);
+            logger.LogDebug("[eslint-bridge GET] Endpoint: {0}", serverEndpoint);
```
duncanp-sonar, line 73: Miscellaneous: improving the look of the logged output.

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -59,22 +59,26 @@ internal class EslintBridgeClient : IEslintBridgeClient
         private readonly IEslintBridgeProcess eslintBridgeProcess;
         private readonly IEslintBridgeHttpWrapper httpWrapper;
         private readonly IAnalysisConfiguration analysisConfiguration;
+        private readonly IEslintBridgeKeepAlive keepAlive;
         private bool isDisposed;
 
         public EslintBridgeClient(string analyzeEndpoint, IEslintBridgeProcess eslintBridgeProcess, ILogger logger)
-            : this(analyzeEndpoint, eslintBridgeProcess, new EslintBridgeHttpWrapper(logger), new AnalysisConfiguration())
+            : this(analyzeEndpoint, eslintBridgeProcess, new EslintBridgeHttpWrapper(logger), new AnalysisConfiguration(),
+                  new EslintBridgeKeepAlive(eslintBridgeProcess, logger))
```
duncanp-sonar, line 67: The client is just responsible for creating and disposing the keep-alive handler.

In an earlier version, the keep-alive code was all in this class. This version is simpler. See #3203 for the earlier version.

I considered adding the functionality to the `EslintBridgeProcess` class, but it is already pretty complicated.

src/TypeScript/EslintBridgeClient/EslintBridgeClient.cs
```diff
@@ -96,7 +100,7 @@ public async Task InitLinter(IEnumerable<Rule> rules, CancellationToken cancella
 
         public async Task<AnalysisResponse> Analyze(string filePath, string tsConfigFilePath, CancellationToken cancellationToken)
         {
-            var tsConfigFilePaths = tsConfigFilePath == null ? Array.Empty<string>() : new[] {tsConfigFilePath};
+            var tsConfigFilePaths = tsConfigFilePath == null ? Array.Empty<string>() : new[] { tsConfigFilePath };
```
duncanp-sonar, line 103: Miscellaneous - whitespace fixup by VS.

src/TypeScript/EslintBridgeClient/EslintBridgeKeepAlive.cs
```diff
@@ -0,0 +1,110 @@
...
+            this.httpWrapper = httpWrapper;
+
+            this.timer = timer;
+            this.timer.AutoReset = true;
+            this.timer.Interval = MillisecondsToWaitBetweenKeepAlives;
+            this.timer.Elapsed += OnKeepAliveTimerElapsed;
+
+            // Note: we're starting the timer immediately, even though the process might not be running
+            // yet (i.e. the user hasn't tried to analyse a JS/TS file)
+            this.timer.Start();
```
duncanp-sonar, line 70: This means the timer will start as soon as this class is created i.e. as soon as the `EslintBridgeClient` instance is created.

This happens when the JS/TS analyzer is MEF-imported i.e. before we know whether there are any JS/TS to analyze.

This was the one advantage of the version in #3203 - the timer was not started until an analysis was triggered.

As it is, two timers (one for JS, one for TS) will start when the `AnalyzerController` is created. They will fire in the background every five seconds, and early-out because the process is not running.

I'm not sure this is worth worrying about, though. What do you think?

src/TypeScript/EslintBridgeClient/EslintBridgeKeepAlive.cs
```diff
@@ -0,0 +1,110 @@
...
+        private readonly ITimer timer;
+        private readonly ILogger logger;
+
+        public EslintBridgeKeepAlive(IEslintBridgeProcess process, ILogger logger)
+            : this(process, logger, new EslintBridgeHttpWrapper(logger), new TimerWrapper())
+        {
+        }
+
+        internal /* for testing */ EslintBridgeKeepAlive(IEslintBridgeProcess process, ILogger logger,
+            IEslintBridgeHttpWrapper httpWrapper, ITimer timer)
```
rita-sonarsource, line 57: I feel that the dependencies between this class and EslintBridgeClient show a problem.

EslintBridgeClient is supposed to be the only one that knows how to communicate with the eslint bridge process; it's responsible for ensuring the process is up and building the server URL. It's a bit out of place that EslintBridgeClient is the one in charge of the timer.

How about a different approach:
* `EslintBridgeAnalyzer` can be the one to create and dispose the `EslintBridgeKeepAlive` class.
* `EslintBridgeKeepAlive` will only have the timer, and will call `EslintBridgeClient.SendKeepAlive()`.
* `EslintBridgeClient` then becomes simple and only does the requests, like it did before.

WDYT?

src/TypeScript/EslintBridgeClient/EslintBridgeKeepAlive.cs
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        private async Task SendKeepAliveAsync()
+        {
+            var port = await process.Start();
+            var fullServerUrl = BuildServerUri(port, "status");
+            await httpWrapper.GetAsync(fullServerUrl, CancellationToken.None);
+        }
+
+        private Uri BuildServerUri(int port, string endpoint) => new Uri($"http://localhost:{port}/{endpoint}");
```
duncanp-sonar, line 106: I forgot to comment on this: the same code exists in the `EslintBridgeClient`.
My intention was to refactor the `EslintBridgeHttpWrapper` class so that it accepted the port and endpoint and built the URL.

The alternative would be make the HttpWrapper even more strongly typed and change it to have higher-level methods e.g. InitLinter, KeepAlive, Analyze etc. However, then it would be very similar to `IEsLintBridgeClient` itself.

src/TypeScript/EslintBridgeClient/EslintBridgeKeepAlive.cs
duncanp-sonar: > It's a bit out of place that EslintBridgeClient is the one in charge of creating and disposing the timer class and that this class also knows how to build the URL and get the port.

I agree with the second bit, but I don't think it is out of place that the EsLintBridgeClient is ultimately responsible for starting the process and keeping it running.

The `EsLintBridgeClient` hides the fact that there is an external process, and handles recovery when the external process fails. I think the fact that the keep-alive is required should be hidden from callers too, including the analyzer.

 Instead of duplicating the URL building, one option is to move it to the http wrapper. The other would be to add a KeepAlive to the EsLintBridgeClient, and have the keep-alive timer class call back to that.

What do you think? (a quick call might be easier, if it's not too noisy).

duncanp-sonar: As discussed offline, we don't want to change the analyzer. I'll merge this as-is and look at changing the keep-alive class separately.

PR #3206, opened 2022-08-23
 Add INotificationStorage

### LINKED ISSUE #3201: Add INotificationStorage
Persists and retrieves notification state e.g. whether the user has said "don't show again" for a specific notification.

ugras-ergun-sonarsource:
fixes #3201


ugras-ergun-sonarsource:
File name change is just removing the extra space at the end.

src/Core/Core.csproj
```diff
@@ -20,6 +20,7 @@
   <ItemGroup>
     <Reference Include="PresentationCore" />
     <Reference Include="PresentationFramework" />
+    <Reference Include="System.ComponentModel.Composition" />
```
rita-sonarsource, line 23: Have we decided to add MEF to Core? Personally I'm in favor.
@duncanp-sonar

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -18,11 +18,84 @@
...
+        }
+        
+        private string GetFilePath()
+        {
+            string appData = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+            string fullPath = Path.Combine(appData, "SonarLint for Visual Studio", vsVersionProvider.Version.MajorInstallationVersion, "disabledNotifications.txt");
+
+            return fullPath;
+        }
+
```
rita-sonarsource, line 98: @ugras-ergun-sonarsource maybe it's time that you install some kind of code styling? 😛 
The empty lines and inconsistent spacing between `if(` and `if (` bother me too much 😆

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -18,11 +18,84 @@
...
+        private IList<string> ReadDisabledNotifications()
+        {
+            return fileSystem.File.ReadAllLines(filePath).ToList();
+        }
+
+        private void SaveNotifications()
+        {
+            fileSystem.File.WriteAllLines(filePath, disabledNotifications);
```
rita-sonarsource, line 88: What about exception handling? everywhere where we access the file system.
It would be inconvenient for the caller to catch them and there is nothing we can do about them besides logging.

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -18,11 +18,84 @@
...
+            if (disabledNotifications == null)
+            {
+                disabledNotifications = ReadDisabledNotifications();
+            }
+            return disabledNotifications.Contains(id);
+        }
+
+        private IList<string> ReadDisabledNotifications()
+        {
+            return fileSystem.File.ReadAllLines(filePath).ToList();
```
rita-sonarsource, line 83: I thought that we decided to store a json object with an array, no? for future flexibility

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -18,11 +18,84 @@
...
+
+    [Export(typeof(IDisabledNotificationsStorage))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class DisabledNotificationsStorage : IDisabledNotificationsStorage
+    {
+        private readonly IVsVersionProvider vsVersionProvider;
+        private readonly IFileSystem fileSystem;
+        private readonly string filePath;
+
+        private IList<string> disabledNotifications = null;
```
rita-sonarsource, line 45: This will throw if accessed from multiple threads

rita-sonarsource (commented):
Some comments about error handling and threading.

src/Core.UnitTests/Notifications/DisabledNotificationsStorageTests.cs
```diff
@@ -0,0 +1,136 @@
...
+        [TestMethod]
+        public void DisableNotification_IsNotDisabled_Disables()
+        {
+            var file = CreateFileMock("1", "2", "3");
+            var testSubject = CreateTestSubject(file.Object);
+
+            var result = testSubject.IsNotificationDisabled("4"); 
+            result.Should().BeFalse();
+
+            testSubject.DisableNotification("4");
```
rita-sonarsource, line 60: None of the tests check what happens if you call Disable without calling IsDisabled first

src/Core.UnitTests/Notifications/DisabledNotificationsStorageTests.cs
```diff
@@ -0,0 +1,136 @@
...
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Core.Notifications;
+using SonarLint.VisualStudio.Core.VsVersion;
+
+namespace SonarLint.VisualStudio.Core.UnitTests.Notifications
+{
+    [TestClass]
+    public class DisabledNotificationsStorageTests
```
rita-sonarsource, line 35: Needs exception scenarios

src/Core.UnitTests/Notifications/DisabledNotificationsStorageTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            var version = new Mock<IVsVersion>();
+            version.SetupGet(v => v.MajorInstallationVersion).Returns("17");
+
+            var versionProvider = new Mock<IVsVersionProvider>();
+            versionProvider.SetupGet(vp => vp.Version).Returns(version.Object);
+
+            return versionProvider.Object;
+        }
+
+        private string GetFilePath()
```
rita-sonarsource, line 126: This shouldn't be copy-pasted, maybe it should be internal in the product code, or the final property should be internal.

duncanp-sonar (commented):
I agree with @rita-sonarsource - needs exception handling when accessing the file system, and we agreed the file would be json.

src/Core.UnitTests/Notifications/DisabledNotificationsStorageTests.cs
ugras-ergun-sonarsource: I did this way so if we have unintentionally changed the save path we would have a warning.

ugras-ergun-sonarsource:
Some tests are missing and the test rita added fails. I'll fix them shortly meanwhile the main code is ready

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -30,17 +38,97 @@ public interface IDisabledNotificationsStorage
 
     [Export(typeof(IDisabledNotificationsStorage))]
     [PartCreationPolicy(CreationPolicy.Shared)]
-    internal class DisabledNotificationsStorage : IDisabledNotificationsStorage
+    public class DisabledNotificationsStorage : IDisabledNotificationsStorage
```
rita-sonarsource, line 41: why is this public?

ugras-ergun-sonarsource: merge mistake

src/Core/VsVersion/IVsVersion.cs
```diff
@@ -36,5 +36,10 @@ public interface IVsVersion
         /// VS display version, as seen in the About window. Example: "16.9.0 Preview 3.0"
         /// </summary>
         string DisplayVersion { get; }
+
+        /// <summary>
+        /// Major VS build version. Example: for "16.9.30914.41" it'll be "16"
+        /// </summary>
+        string MajorInstallationVersion { get; }
```
duncanp-sonar, line 43: I would have made the changes to the this interface first and pushed them in a separate PR. It's completely self-contained, would have been quick to do and review, and would have increased the amount of "done" work.

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -32,15 +40,95 @@ public interface IDisabledNotificationsStorage
...
+                var fileContent = JsonConvert.SerializeObject(disabledNotifications);
+                fileSystem.File.WriteAllText(filePath, fileContent);
+            }
+            catch (Exception ex)
+            {
+                logger.WriteLine($"Failed to save disabled notification config to disk. Error:{ex.Message}");
+            }
+        }
+        
+        private string GetFilePath()
```
duncanp-sonar, line 126: `GetExpectedDisableNotificationsFilePath` would be a better name.

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -32,15 +40,95 @@ public interface IDisabledNotificationsStorage
...
+        }
+
+        private void SaveNotifications()
+        {
+            try
+            {
+                var fileContent = JsonConvert.SerializeObject(disabledNotifications);
+                fileSystem.File.WriteAllText(filePath, fileContent);
+            }
+            catch (Exception ex)
```
duncanp-sonar, line 120: Don't catch critical exceptions.

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -32,15 +40,95 @@ public interface IDisabledNotificationsStorage
...
+            {
+                logger.WriteLine($"Failed to save disabled notification config to disk. Error:{ex.Message}");
```
duncanp-sonar, line 122: All non-debug messages should be in a resource file.

Also, the message should be prefixed with `[Notifications] `.

Same for the other error message.

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -32,15 +40,95 @@ public interface IDisabledNotificationsStorage
...
+        private string filePath => GetFilePath();
+
+        private DisabledNotifications disabledNotifications = null;
+        private readonly object lockObject = new object();
+
+        [ImportingConstructor]
+        public DisabledNotificationsStorage(IVsVersionProvider vsVersionProvider, ILogger logger) : this(vsVersionProvider, logger, new FileSystem())
         {
-            // todo
+
```
duncanp-sonar, line 55: Redundant empty line

src/Core.UnitTests/Notifications/DisabledNotificationsStorageTests.cs
```diff
@@ -18,19 +18,237 @@
...
+            file.Setup(f => f.ReadAllText(It.IsAny<string>())).Throws(new Exception("Test"));
+            var logger = new TestLogger();
+
+            var testSubject = CreateTestSubject(file.Object, logger);
+
+            var result = testSubject.IsNotificationDisabled("1");
+
+            result.Should().BeFalse();
+            logger.AssertOutputStrings(2);
+            logger.AssertOutputStringExists("Failed to read disabled notification config from disk. Error:Test");
```
duncanp-sonar, line 79: We don't normally check for the exact text of localizable  strings.
Instead, we use a longer error message in the test exception and check that that appears in the output.

Search the existing code for `"this is a test"` - there are lots of examples.

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -32,15 +40,95 @@ public interface IDisabledNotificationsStorage
...
+        }
         public bool IsNotificationDisabled(string id)
         {
-            // todo
-            return false;
+            lock (lockObject)
+            {
+                if (disabledNotifications == null)
+                {
+                    disabledNotifications = ReadDisabledNotifications();
```
duncanp-sonar, line 88: This is potentially confusing; it will be shown even if the file exists, but couldn't be loaded because of a serialization error.

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -32,15 +40,95 @@ public interface IDisabledNotificationsStorage
...
+                }
+                return disabledNotifications.Notifications.Any(n => n.Id == id);
+            }
+        }
+
+        private DisabledNotifications ReadDisabledNotifications()
+        {
+            try
+            {
+                var fileContent = fileSystem.File.ReadAllText(filePath);
```
duncanp-sonar, line 103: This will throw if the file doesn't exist, which is normal and expected situation if the user has never disabled a notification.

It would be better to handle the file-not-found case separately e.g. something like:

```
if (!File.Exists(filePath))
{
  logger.LogDebug($"[Notifications] Disabled notifications file does not exist. File: {filePath}");
  return new DisabledNotifications();
}
```

src/Core/Notifications/DisabledNotifications.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            Notifications = new List<DisabledNotification>();
+        }
+
+        public IList<DisabledNotification> Notifications { get; }
+
+        public void AddNotification(string id)
+        {
+            var disabledNotification = new DisabledNotification(id);
+            Notifications.Add(disabledNotification);
```
duncanp-sonar, line 38: This allows duplicates. We should at least `Debug.Assert` that the notification isn't already in the list.

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -32,15 +40,95 @@ public interface IDisabledNotificationsStorage
...
+                logger.WriteLine($"Failed to read disabled notification config from disk. Error:{ex.Message}");                
+            }
+            return null;
+        }
+
+        private void SaveNotifications()
+        {
+            try
+            {
+                var fileContent = JsonConvert.SerializeObject(disabledNotifications);
```
duncanp-sonar, line 117: What does the serialized json look like?

duncanp-sonar: We normally serialize with `Formatting.Indented` for readability.

src/Core.UnitTests/Notifications/DisabledNotificationsStorageTests.cs
duncanp-sonar: A better name would be `GetExpectedDisabledNotificationsFilePath`.

duncanp-sonar (changes requested):
See inline comments.

src/Core/Notifications/IDisabledNotificationsStorage.cs
ugras-ergun-sonarsource: I have changed the wording to "[Notifications]Couldn't load disabled notification config. Notification will be showed."

src/Core/Resources/Strings.resx
```diff
@@ -0,0 +1,129 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="DisabledNotificationReadError" xml:space="preserve">
+    <value>[Notifications]Failed to read disabled notification config from disk. Error: {0}</value>
```
rita-sonarsource, line 121: Missing spaces between ]

src/Core/Resources/Strings.resx
```diff
@@ -0,0 +1,129 @@
...
+  </data>
+  <data name="DisabledNotificationSaveError" xml:space="preserve">
+    <value>[Notifications]Failed to save disabled notification config to disk. Error: {0}</value>
+  </data>
+  <data name="DisabledNotificationsFailedToLoad" xml:space="preserve">
+    <value>[Notifications]Couldn't load disabled notification config. Notification will be showed.</value>
```
rita-sonarsource, line 127: "Disabled notifications will be shown"

src/Core/Core.csproj
rita-sonarsource, line 23: Another merge conflict? this is already included below

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -32,15 +42,101 @@ public interface IDisabledNotificationsStorage
...
+            }
+        }
+
+        private DisabledNotifications ReadDisabledNotifications()
+        {
+            try
+            {
+                if (!fileSystem.File.Exists(filePath)) 
+                { 
+                    logger.LogDebug($"[Notifications]Disabled notifications file does not exist. File: {filePath}");
```
rita-sonarsource, line 107: Space after ]

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -32,15 +42,101 @@ public interface IDisabledNotificationsStorage
...
+                }
+
+                var fileContent = fileSystem.File.ReadAllText(filePath);
+                return JsonConvert.DeserializeObject<DisabledNotifications>(fileContent);
+            }
+            catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+            {
+                logger.WriteLine(string.Format(Strings.DisabledNotificationReadError, ex.Message));                
+            }
+            return null;
```
rita-sonarsource, line 118: If the json is corrupted, we will forever return null and then never update any storage. 
I think this method should never return null, and worst case it should treat it as if the file does not exist. @duncanp-sonar wdyt?

src/Core.UnitTests/Notifications/DisabledNotificationsStorageTests.cs
```diff
@@ -18,19 +18,274 @@
...
+
+            var result = testSubject.IsNotificationDisabled("1");
+
+            result.Should().BeFalse();
+            file.Verify(f => f.Exists(ExpectedDisabledNotificationsFilePath), Times.Once);
+            file.Verify(f => f.ReadAllText(ExpectedDisabledNotificationsFilePath), Times.Never);
+        }
+
+        [TestMethod]
+        public void IsNotificationDisabled_ReadError_LogsError()
```
rita-sonarsource, line 82: Missing critical exception test

src/Core.UnitTests/Notifications/DisabledNotificationsStorageTests.cs
```diff
@@ -18,19 +18,274 @@
...
+
+            var result = testSubject.IsNotificationDisabled("1");
+
+            result.Should().BeFalse();
+            logger.AssertOutputStrings(2);
+            logger.AssertPartialOutputStringExists("this is a test");
+        }
+
+        [TestMethod]
+        public void DisableNotification_IsNotDisabled_IsDisabledNotCalled_Disables()
```
rita-sonarsource, line 98: The name of this test is very confusing.

DisableNotification_NotificationIsNotPreviouslyDisabled_NotificationIsNowDisabled

src/Core.UnitTests/Notifications/DisabledNotificationsStorageTests.cs
rita-sonarsource: What I meant was that IsDisabled should be tested in isolation.

src/Core.UnitTests/Notifications/DisabledNotificationsStorageTests.cs
```diff
@@ -18,19 +18,274 @@
...
+
+            var disabledNotificationsJson = (string)file.Invocations[1].Arguments[1];
+            var disabledNotificationsResult = JsonConvert.DeserializeObject<DisabledNotifications>(disabledNotificationsJson);
+
+            disabledNotificationsResult.Notifications.Count().Should().Be(1);
+            disabledNotificationsResult.Notifications.Any(n => n.Id == "1").Should().BeTrue();
+        }
+
+        [TestMethod]
+        public void DisableNotification_AlreadyDisabled_IsDisabledNotCalled_DoesNothing()
```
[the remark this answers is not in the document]
ugras-ergun-sonarsource: A call to "IsNotificationDisabled" populates the "DisabledNotifications" object. So I put both scenarios for whether "IsNotificationDisabled" is called or not. If it's confusing I could remove the ones that call the "IsNotificationDisabled"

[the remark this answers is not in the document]
rita-sonarsource: Perhaps a better check would be to call Disable() twice in a row with same id and verify that the file is only saved once.
Same for IsDisabled, verify that it's loaded only once.

src/Core/Resources/Strings.resx
ugras-ergun-sonarsource: [Apparently, both are correct. ](https://www.merriam-webster.com/dictionary/show) But it seems shown is more widely used.

rita-sonarsource: You can't say "be showed"

src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -32,15 +40,95 @@ public interface IDisabledNotificationsStorage
...
+        private DisabledNotifications ReadDisabledNotifications()
+        {
+            try
+            {
+                var fileContent = fileSystem.File.ReadAllText(filePath);
+                return JsonConvert.DeserializeObject<DisabledNotifications>(fileContent);
+            }
+            catch (Exception ex)
+            {
+                logger.WriteLine($"Failed to read disabled notification config from disk. Error:{ex.Message}");
```
duncanp-sonar: This should be in the resource file.

duncanp-sonar (approved):
Just the spacing and spelling issues @rita-sonarsource mentioned.

src/Core/Notifications/IDisabledNotificationsStorage.cs
ugras-ergun-sonarsource: it is already. Are you looking to an older version?

duncanp-sonar:
Yes, I was. My bad.

Duncan Pocklington | Sonar

Developer

https://sonarsource.com

Are you using SonarLint <https://www.sonarlint.org> in your IDE?

rita-sonarsource (approved):

PR #3208, opened 2022-08-23
Added missing test for CompatibleNodeLocator

duncanp-sonar:


src/TypeScript.UnitTests/NodeJSLocator/CompatibleNodeLocatorTests.cs
```diff
@@ -52,6 +52,21 @@ public void Locate_NoDetectedVersions_Null()
             result.Should().BeNull();
         }
 
+        [TestMethod]
+        public void Locate_NoCompatibleVersions_ReturnsNull()
```
duncanp-sonar, line 56: There are tests for "no _detected_ versions" and "first compatible version", but not for "no compatible versions".

ugras-ergun-sonarsource (approved):

rita-sonarsource (approved):

PR #3211, opened 2022-08-23
Call Node notification service if compatible version not found

### LINKED ISSUE #3175: Display notification if compatible NodeJS not found
### Description

A notification should be displayed if JS/TS analysis cannot be performed because an appropriate version of NodeJS cannot be found.

The notification should contain a link to more information that explains the problem and how to fix it.

The user should be able to:
* dismiss the current notification
* indicate that they do not want to see the notification again.

TODO - should the notification only be displayed for every file, or only once per VS session?
UPDATE: a total of once per VS session (i.e. if shown only once even if both JS and TS are analysed).

duncanp-sonar:
Fixes #3175

### DIFF
--- a/src/TypeScript.UnitTests/NodeJSLocator/CompatibleNodeLocatorTests.cs
+++ b/src/TypeScript.UnitTests/NodeJSLocator/CompatibleNodeLocatorTests.cs
@@ -24,7 +24,6 @@ using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Core.JsTs;
-using SonarLint.VisualStudio.Core.Notifications;
 using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.Integration.UnitTests;
 using SonarLint.VisualStudio.TypeScript.NodeJSLocator;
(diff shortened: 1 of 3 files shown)


src/TypeScript.UnitTests/NodeJSLocator/CompatibleNodeLocatorTests.cs
```diff
@@ -97,15 +130,18 @@ public void IsCompatibleVersion_ReturnsTrueFalse(int majorVersion, bool expected
...
 
-        private CompatibleNodeLocator CreateTestSubject(IReadOnlyCollection<NodeVersionInfo> candidateLocations = null)
+        private CompatibleNodeLocator CreateTestSubject(IReadOnlyCollection<NodeVersionInfo> candidateLocations = null,
+            IUnsupportedNodeVersionNotificationService unsupportedNodeNotificationService = null)
         {
             candidateLocations ??= Array.Empty<NodeVersionInfo>();
             var versionInfoProvider = new Mock<INodeVersionInfoProvider>();
             versionInfoProvider.Setup(x => x.GetAllNodeVersions()).Returns(candidateLocations);
 
+            unsupportedNodeNotificationService ??= Mock.Of<IUnsupportedNodeVersionNotificationService>();
```
duncanp-sonar, line 140: Depends on `IUnsupportedNodeVersionNotificationService` being MEF-exportable - being done separately.

rita-sonarsource: Perhaps so this commit won't break, it's worth adding a dummy implementation of this service that is mef exportable.

rita-sonarsource (approved):

PR #3214, opened 2022-08-24
Implement UnsupportedNodeVersionNotificationService

rita-sonarsource:
Fixes [#3217](https://github.com/SonarSource/sonarlint-visualstudio/issues/3217)


src/TypeScript/Notifications/IUnsupportedNodeVersionNotificationService .cs
```diff
@@ -31,9 +32,36 @@ public interface IUnsupportedNodeVersionNotificationService
     [PartCreationPolicy(CreationPolicy.Shared)]
     internal class UnsupportedNodeVersionNotificationService : IUnsupportedNodeVersionNotificationService
     {
+        private readonly INotificationService notificationService;
+        private readonly INotification notification;
+
+        private const string NotificationId = "sonarlint.nodejs.min.version.not.found.10";
+
+        [ImportingConstructor]
+        public UnsupportedNodeVersionNotificationService(INotificationService notificationService,
```
rita-sonarsource, line 41: MEF-exported in https://github.com/SonarSource/sonarlint-visualstudio/pull/3215

src/TypeScript/Notifications/IUnsupportedNodeVersionNotificationService .cs
```diff
@@ -31,9 +32,36 @@ public interface IUnsupportedNodeVersionNotificationService
...
+
         public void Show()
         {
-            // TODO
+            notificationService.ShowNotification(notification);
+        }
+
+        private void ShowMoreInfo()
+        {
+            // todo: output pane? wiki page?
```
rita-sonarsource, line 64: I'm not sure what we should do here.
Preferably it should go to a documentation page, but which link? our wiki? We've avoided linking directly to specific wiki pages.

src/TypeScript/Notifications/Resources.resx
```diff
@@ -0,0 +1,126 @@
...
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="NotificationShowMoreInfoAction" xml:space="preserve">
+    <value>Show more info</value>
+  </data>
+  <data name="NotificationUnsupportedNode" xml:space="preserve">
+    <value>SonarLint: JavaScript/TypeScript analysis failed: no supported version of Node was found.</value>
```
duncanp-sonar, line 124: Do we need the "SonarLint" text as well as the SonarLint icon?

src/TypeScript/Notifications/Resources.resx
```diff
@@ -0,0 +1,126 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="NotificationShowMoreInfoAction" xml:space="preserve">
+    <value>Show more info</value>
```
duncanp-sonar, line 121: A more action-oriented message would be something like "_How do I fix this?_".
Maybe as @marco-comi-sonarsource?

src/TypeScript/Notifications/IUnsupportedNodeVersionNotificationService .cs
duncanp-sonar: The spec says a [wiki page](https://docs.google.com/document/d/1EoG_dIMb83G1P16fgabtDHSofHh6DUNWmZlrrQ_nNJU/edit), but doesn't specify whether it should be the root page or a specific page.

We should open a discussion with ITOps about redirects for documentation (or ask @kevin-hinz whether the doc team are already working on this).

src/TypeScript/Notifications/Resources.resx
marco-comi-sonarsource: +1 for "How do I fix this?" or "How to fix it"

src/TypeScript/Notifications/Resources.resx
marco-comi-sonarsource: If not, would the user have any hint that the notification is about SonarLint?

rita-sonarsource: There would be a SonarLint icon. I've removed it for now, perhaps we can see how it looks like on dogfood and add back if it's not visible enough.

src/TypeScript/Notifications/IUnsupportedNodeVersionNotificationService .cs
kevin-hinz: We are planning for redirects, in part, with the Martech Team. At this time, the final destination URLs are not defined and no clear process is established.

There are 2 methods to use now:

1. According to [this demo](https://christianheilmann.com/2021/08/03/redirecting-github-pages/), it is possible to redirect .md pages which should include GitHub wikis. 
2. Use [this page of redirects](https://redirect.sonarsource.com/) that we own. It is likely that no one team or person controls this page; we just need to find out where it lives! In this option, we can create a new folder for SonarLint/, your hyperlink points to some key on that page, the key is redirected to the wiki (now), and to the docs.sonarlint (after migration).

The easy solution is Option 1: create a wiki, link to it, and hope the .md redirect workflow works later.
Option 2: takes a search to find/edit the redirects page, and a plan to make it work well. @peeter-piegaze-sonarsource should be involved here.

In both cases, the URL should be updated in the Product when we have the final destinations available and the content will live in a wiki now. We should meet to choose which option to follow.

PR #3225, opened 2022-08-26
Disabled notifications file is not created when folder does not exist 

### LINKED ISSUE #3224: Disabled notifications file is not created when folder does not exist

ugras-ergun-sonarsource:
fixes #3224

### DIFF
--- a/src/Core/Notifications/IDisabledNotificationsStorage.cs
+++ b/src/Core/Notifications/IDisabledNotificationsStorage.cs
@@ -105,7 +105,8 @@ namespace SonarLint.VisualStudio.Core.Notifications
             {
                 if (!fileSystem.File.Exists(FilePath)) 
                 { 
-                    logger.LogDebug($"[Notifications] Disabled notifications file does not exist. File: {FilePath}"); 
+                    logger.LogDebug($"[Notifications] Disabled notifications file does not exist. File: {FilePath}");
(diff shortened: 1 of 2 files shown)


src/Core/Notifications/IDisabledNotificationsStorage.cs
```diff
@@ -105,7 +105,8 @@ private DisabledNotifications ReadDisabledNotifications()
             {
                 if (!fileSystem.File.Exists(FilePath)) 
                 { 
-                    logger.LogDebug($"[Notifications] Disabled notifications file does not exist. File: {FilePath}"); 
+                    logger.LogDebug($"[Notifications] Disabled notifications file does not exist. File: {FilePath}");
+                    fileSystem.Directory.CreateDirectory(Path.GetDirectoryName(FilePath));
```
rita-sonarsource, line 109: why not do it on save? this way if the user never dismisses anything, we won't create an empty directory.

ugras-ergun-sonarsource: So that we will create this once per user.

ugras-ergun-sonarsource: But save will be seldom anyway, so it makes sense.

rita-sonarsource (approved):

PR #3233, opened 2022-08-29
Gold bar isn't removed after clicking don't show this again

### LINKED ISSUE #3226: Gold bar isn't removed after clicking don't show this again

ugras-ergun-sonarsource:
Fixes #3226

### DIFF
--- a/src/Core.UnitTests/Notifications/NotificationServiceTests.cs
+++ b/src/Core.UnitTests/Notifications/NotificationServiceTests.cs
@@ -88,9 +88,9 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Notifications
         {
             var notification = CreateNotification(actions: new INotificationAction[]
             {
-                new NotificationAction("notification1", _ => { }),
-                new NotificationAction("notification2", _ => { }),
-                new NotificationAction("notification2", _ => { })
+                new NotificationAction("notification1", _ => { }, false),
+                new NotificationAction("notification2", _ => { }, false),
+                new NotificationAction("notification2", _ => { }, false)
             });
 
             var infoBarManager = CreateInfoBarManager(notification, Mock.Of<IInfoBar>());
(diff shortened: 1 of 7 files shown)


src/Core/Notifications/DoNotShowAgainNotificationAction.cs
```diff
@@ -41,6 +41,8 @@ public DoNotShowAgainNotificationAction(IDisabledNotificationsStorage disabledNo
 
         public string CommandText => CoreStrings.Notifications_DontShowAgainAction;
 
+        public bool ShouldDismissAfterAction => true;
```
ugras-ergun-sonarsource, line 44: Test Coverage fails because this line is not covered. I do not see any meaningful tests for this.

src/Core.UnitTests/Notifications/NotificationServiceTests.cs
```diff
@@ -220,17 +220,43 @@ public void ShowNotification_UnknownInfoBarButtonClicked_NoException()
 
             infoBar.Raise(x => x.ButtonClick += null, new InfoBarButtonClickedEventArgs("unknown notification"));
 
+            callback.Invocations.Count.Should().Be(0);            
+        }
+
+        [TestMethod]
+        public void ShowNotification_InfoBarButtonClicked_ActionInvokedWithTheNotification_NotDismissed()
```
duncanp-sonar, line 227: This test is almost identical to the existing 'ShowNotification_InfoBarButtonClicked_ActionInvokedWithTheNotification_Dismissed`, isn't it?
i.e. the only difference is the check whether `DetachInfoBar` is null or not.

If that is the case, it would be simpler to have one data-driven test that accepts true/false and does the appropriate `DetachInfoBar` check.

(as it is, it isn't clear from the test names what the difference between the two tests is i.e. why we expect one to be dismissed and the other not).

duncanp-sonar (approved):
A single data-driven test rather than two nearly identical tests would be cleaner.

PR #3234, opened 2022-08-29
Change notification storage file format to be more general

duncanp-sonar:
Related to #3201


duncanp-sonar:
Changed the notifications file format to make it clear that it is internal (i.e. we don't expect users to edit it), and to make it more general purpose, in case we want to store other notification-related information in the future.

The file is now called `internal.notifications.json`.
The content looks like this:

```
{
  "DisabledNotifications": [
    {
      "Id": "sonarlint.nodejs.min.version.not.found.10"
    }
  ]
}
```

duncanp-sonar:
The CI is failing because of an unrelated flaky test. I've disabled the test in a separate PR (#3237).

ugras-ergun-sonarsource (approved):

PR #3238, opened 2022-08-29
Changed NodeJS notification help to a specific wiki page

duncanp-sonar:
A placeholder for the new page exist, but without any content: https://github.com/SonarSource/sonarlint-visualstudio/wiki/Supported-NodeJS-versions

### DIFF
--- a/src/TypeScript.UnitTests/Notifications/UnsupportedNodeVersionNotificationServiceTests.cs
+++ b/src/TypeScript.UnitTests/Notifications/UnsupportedNodeVersionNotificationServiceTests.cs
@@ -110,7 +110,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Notifications
 
             firstAction.Action(null);
 
-            browserService.Verify(x=> x.Navigate("https://github.com/SonarSource/sonarlint-visualstudio/wiki/Supported-NodeJS-versions"), Times.Once);
+            browserService.Verify(x=> x.Navigate("https://github.com/SonarSource/sonarlint-visualstudio/wiki/NodeJS-prerequisite-for-JavaScript-and-TypeScript-analysis"), Times.Once);
             browserService.VerifyNoOtherCalls();
         }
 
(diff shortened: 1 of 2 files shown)


src/TypeScript.UnitTests/Notifications/UnsupportedNodeVersionNotificationServiceTests.cs
```diff
@@ -110,7 +110,7 @@ public void Show_ShowMoreInfoAction_OpensBrowser()
 
             firstAction.Action(null);
 
-            browserService.Verify(x=> x.Navigate("https://github.com/SonarSource/sonarlint-visualstudio/wiki"), Times.Once);
+            browserService.Verify(x=> x.Navigate("https://github.com/SonarSource/sonarlint-visualstudio/wiki/Supported-NodeJS-versions"), Times.Once);
```
ugras-ergun-sonarsource, line 113: I think it should be "required" instead of "supported".

duncanp-sonar: Or perhaps "minimum". What do you think, @rita-sonarsource?

rita-sonarsource: Maybe something else entirely, like "nodejs-prerequisite"?

duncanp-sonar: Let's go with `NodeJs-prerequisite`.

duncanp-sonar: Or `NodeJS-prerequisite-for-JavaScript-and-TypeScript-analysis`, which makes the wiki page list look like this:

image

Just `NodeJS-prerequisite` on its own looked a bit odd, as it looked like it applied to all languages.

rita-sonarsource: LGTM 👍

PR #3251, opened 2022-09-02
invalid chars

ugras-ergun-sonarsource:


src/Infrastructure.VS/LanguageDetection/SonarLanguageRecognizer.cs
```diff
@@ -154,6 +159,16 @@ private string GetNormalizedExtention(string fileName)
             return extension.ToLowerInvariant();
         }
         
-
+        private bool IsFileNameInvalid(string fileName)
```
rita-sonarsource, line 162: https://stackoverflow.com/questions/27253394/getinvalidfilenamechars-doesnt-contain-all-illegal-chars

I think it's better to do a try-catch around GetNormalizedExtention

ugras-ergun-sonarsource: I don't think it's a concern for us. Our code will always run on a windows machine. This would create problems if we had operations run on different systems.

duncanp-sonar (approved):

PR #3255, opened 2022-09-05
Fix MefTestHelper

### LINKED ISSUE #2467: Infra: improve MEF tests
### Description

The logic in the MefTestHelper is supposed to check that the composition will fail if any required export is missing, but it doesn't.
The new version in `sonarlint-omnisharp` should do a better job.

#### Notes
The SLVS-version used "Take" to select an increasing number of the required elements from the array. However, that's not a good check.

e.g. all inputs [ A, B, C ]
Take[0] -> tests with [ ] -> should fail
Take[1] -> tests with [ A ] -> should fail
Take[1] -> tests with [ A, B ] -> should fail

However, even if all of the iterations fail as expected, the only thing we can conclude is that C is required. It's possible A and B are optional.

The new `sonarlint-omnisharp` version tests the following:
[ B, C ]
[ A, B ]
[ A, C ]
... and expects failure in each case.

duncanp-sonar:
- was not correctly checking for required exports

Fixes #2467

### DIFF
--- a/src/IssueViz.Security.UnitTests/OpenInIDE/Http/OwinPipelineProcessorTests.cs
+++ b/src/IssueViz.Security.UnitTests/OpenInIDE/Http/OwinPipelineProcessorTests.cs
@@ -19,7 +19,8 @@
  */
 
 using System;
-using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.ComponentModel.Composition.Hosting;
 using System.Net;
 using System.Threading.Tasks;
 using FluentAssertions;
@@ -35,14 +36,50 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
     [TestClass]
     public class OwinPipelineProcessorTests
     {
-        const string Origin = "http://origin";
+        private const string Origin = "http://origin";
 
         [TestMethod]
         public void MefCtor_CheckIsExported()
         {
+            // IOwinRequestHandlers are ImportMany -> optional
             MefTestHelpers.CheckTypeCanBeImported<OwinPipelineProcessor, IOwinPipelineProcessor>(
-                MefTestHelpers.CreateExport<IEnumerable<IOwinPathRequestHandler>>(Array.Empty<IOwinPathRequestHandler>()),
                 MefTestHelpers.CreateExport<ILogger>());
+        }
+
+        [TestMethod]
+        public void MefCtor_CheckIsExported_RequestHandlersAreImported()
+        {
+            var requestHandler1 = CreateHandler("/path1");
+            var requestHandler2 = CreateHandler("/path2");
+
+            var batch = new CompositionBatch();
+
+            batch.AddExport(MefTestHelpers.CreateExport<IOwinPathRequestHandler>(requestHandler1));
+            batch.AddExport(MefTestHelpers.CreateExport<IOwinPathRequestHandler>(requestHandler2));
+            batch.AddExport(MefTestHelpers.CreateExport<ILogger>());
+
+            var importer = new SingleObjectImporter<IOwinPipelineProcessor>();
+            batch.AddPart(importer);
+
+            using var catalog = new TypeCatalog(typeof(OwinPipelineProcessor));
+            using var container = new CompositionContainer(catalog);
+            container.Compose(batch);
+
+            importer.Import.Should().NotBeNull();
+
+            var actual = (OwinPipelineProcessor)importer.Import;
+            actual.PathToHandlerMap.Count.Should().Be(2);
+
+            actual.PathToHandlerMap.Keys.Should().BeEquivalentTo(new string[] { "/path1", "/path2"} );
+            actual.PathToHandlerMap["/path1"].Should().BeSameAs(requestHandler1);
+            actual.PathToHandlerMap["/path2"].Should().BeSameAs(requestHandler2);
+
+            IOwinPathRequestHandler CreateHandler(string path)
+            {
+                var handler = new Mock<IOwinPathRequestHandler>();
+                handler.Setup(x => x.ApiPath).Returns(path);
+                return handler.Object;
+            }
         }
 
         [TestMethod]
--- a/src/IssueViz.UnitTests/Editor/IssueSpanCalculatorTests.cs
+++ b/src/IssueViz.UnitTests/Editor/IssueSpanCalculatorTests.cs
@@ -46,8 +46,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor
         [TestMethod]
         public void MefCtor_CheckIsExported()
         {
-            MefTestHelpers.CheckTypeCanBeImported<IssueSpanCalculator, IIssueSpanCalculator>(
-                MefTestHelpers.CreateExport<IChecksumCalculator>());
+            MefTestHelpers.CheckTypeCanBeImported<IssueSpanCalculator, IIssueSpanCalculator>();
         }
 
         [DataTestMethod]
--- a/src/IssueViz.UnitTests/Editor/LocationTagging/IssueLocationStoreAggregatorTests.cs
+++ b/src/IssueViz.UnitTests/Editor/LocationTagging/IssueLocationStoreAggregatorTests.cs
@@ -36,13 +36,24 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.LocationTag
     public class IssueLocationStoreAggregatorTests
     {
         [TestMethod]
-        public void MefCtor_CheckExports()
+        public void MefCtor_CheckIsExported()
+        {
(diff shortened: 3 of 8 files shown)


src/IssueViz.Security.UnitTests/OpenInIDE/Http/OwinPipelineProcessorTests.cs
```diff
@@ -35,16 +36,52 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
     [TestClass]
     public class OwinPipelineProcessorTests
     {
-        const string Origin = "http://origin";
+        private const string Origin = "http://origin";
```
duncanp-sonar, line 39: Miscellaneous: missing scope.

src/TestInfrastructure/Helpers/Exceptions.cs
```diff
@@ -34,67 +34,9 @@ public static class Exceptions
...
-        {
-            if (action == null)
-            {
-                throw new ArgumentNullException("action");
-            }
-
-            try
-            {
-                action();
-                FluentAssertions.Execution.Execute.Assertion.FailWith("Expected exception " + typeof(TException).FullName);
```
duncanp-sonar, line 83: This method was buggy: I don't think it's ever worked properly.

src/TestInfrastructure/Helpers/Exceptions.cs
```diff
@@ -34,67 +34,9 @@ public static class Exceptions
...
-                    FluentAssertions.Execution.Execute.Assertion.FailWith("Expected exception " + typeof(TException).FullName + " but got " + ex.GetType().FullName + "\n" + ex.ToString());
-                }
-
-                // Perform any additional checks that were supplied.
-                if (additionalChecks != null)
-                {
-                    additionalChecks(ex);
-                }
-            }
+            action.Should().Throw<TException>();
```
duncanp-sonar, line 39: We don't need any of the other methods that were in this class.

We could get rid of this method too, but we use it in 64 places, so it would be in a different PR.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -103,10 +107,13 @@ public static Export CreateExport<T>(object exportInstance, string contractName
             for (int i = 0; i < requiredExports.Count(); i++)
             {
                 // Try importing when not all of the required exports are available -> exception
-                Exceptions.Expect<Exception>(
```
duncanp-sonar, line 106: Switching to use the `FluentAssertions` `Should().Throw` fixes one of errors in `CheckTypeCanBeImported`.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -103,10 +107,13 @@ public static Export CreateExport<T>(object exportInstance, string contractName
...
-                    null,
-                    () => TryCompose(importer, catalog, requiredExports.Take(i)),
-                    checkDerived: true);
+                var subsetOfRequiredExports = new List<Export>(requiredExports);
+                var exportToRemove = subsetOfRequiredExports[i];
```
duncanp-sonar, line 111: This fixes the other error in `CheckTypeCanBeImported` i.e. the one described in #2467.

src/TypeScript.UnitTests/NodeJSLocator/NodeVersionInfoProviderTests.cs
```diff
@@ -41,8 +40,7 @@ public class NodeVersionInfoProviderTests
         public void MefCtor_CheckIsExported()
         {
             MefTestHelpers.CheckTypeCanBeImported<NodeVersionInfoProvider, INodeVersionInfoProvider>(
-                MefTestHelpers.CreateExport<INodeLocationsProvider>(),
-                MefTestHelpers.CreateExport<ILogger>());
+                MefTestHelpers.CreateExport<INodeLocationsProvider>());
```
duncanp-sonar, line 43: This test was passing incorrectly. The `logger` import is not required.

src/IssueViz.UnitTests/Selection/IssueSelectionServiceTests.cs
```diff
@@ -34,8 +34,7 @@ public class IssueSelectionServiceTests
         [TestMethod]
         public void MefCtor_CheckIsExported()
         {
-            MefTestHelpers.CheckTypeCanBeImported<IssueSelectionService, IIssueSelectionService>(
-                MefTestHelpers.CreateExport<IAnalysisIssueSelectionService>());
+            MefTestHelpers.CheckTypeCanBeImported<IssueSelectionService, IIssueSelectionService>();
```
duncanp-sonar, line 37: This test was passing incorrectly - the type has no required imports.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -23,18 +23,22 @@
...
+using System.Diagnostics;
 using System.Linq;
+using FluentAssertions;
 using Moq;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests
 {
     public static class MefTestHelpers
     {
+        [DebuggerStepThrough]
```
duncanp-sonar, line 35: Miscellaneous: simplifies debugging through the tests. MS docs are [here](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.debuggerstepthroughattribute?view=net-6.0).

src/IssueViz.UnitTests/Editor/LocationTagging/IssueLocationStoreAggregatorTests.cs
```diff
@@ -36,13 +36,24 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.LocationTag
     public class IssueLocationStoreAggregatorTests
     {
         [TestMethod]
-        public void MefCtor_CheckExports()
+        public void MefCtor_CheckIsExported()
```
duncanp-sonar, line 39: New test - this type effectively has no required types since it is `ImportMany`.

src/IssueViz.UnitTests/Editor/LocationTagging/IssueLocationStoreAggregatorTests.cs
```diff
@@ -36,13 +36,24 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.LocationTag
...
+        {
+            // IIssueLocationStores are ImportMany -> optional
+            MefTestHelpers.CheckTypeCanBeImported<IssueLocationStoreAggregator, IIssueLocationStoreAggregator>();
+        }
+
+        [TestMethod]
+        public void MefCtor_CheckIsExported_LocationStoresAreImported()
```
duncanp-sonar, line 46: I modified the existing test to check that the expected `ImportMany` types are actually imported.

src/IssueViz.Security.UnitTests/OpenInIDE/Http/OwinPipelineProcessorTests.cs
```diff
@@ -35,16 +36,52 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.OpenInIDE
...
         public void MefCtor_CheckIsExported()
         {
+            // IOwinRequestHandlers are ImportMany -> optional
             MefTestHelpers.CheckTypeCanBeImported<OwinPipelineProcessor, IOwinPipelineProcessor>(
-                MefTestHelpers.CreateExport<IEnumerable<IOwinPathRequestHandler>>(Array.Empty<IOwinPathRequestHandler>()),
                 MefTestHelpers.CreateExport<ILogger>());
         }
 
+        [TestMethod]
+        public void MefCtor_CheckIsExported_RequestHandlersAreImported()
```
duncanp-sonar, line 50: New test to check that the `ImportMany` request handlers are imported as expected.
This test is logically the same as `MefCtor_CheckIsExported_LocationStoresAreImported` (below).

src/IssueViz.UnitTests/Editor/IssueSpanCalculatorTests.cs
```diff
@@ -46,8 +46,7 @@ public void TestInitialize()
         [TestMethod]
         public void MefCtor_CheckIsExported()
         {
-            MefTestHelpers.CheckTypeCanBeImported<IssueSpanCalculator, IIssueSpanCalculator>(
-                MefTestHelpers.CreateExport<IChecksumCalculator>());
+            MefTestHelpers.CheckTypeCanBeImported<IssueSpanCalculator, IIssueSpanCalculator>();
```
duncanp-sonar, line 49: This test was passing incorrectly before - the import isn't required.

ugras-ergun-sonarsource (approved):

PR #3259, opened 2022-09-07
DO NOT MERGE - blocked process

### LINKED ISSUE #3230: Exceptions starting the eslint-bridge process cause worker threads to be blocked
### Description

If the `eslintbridge` process does not start as expected, the `EslintBridgeProcess` `startTask` never completes.

Analysis will never succeed, and each subsequent analysis request will block a worker thread.

#### Notes
When the `eslintbridge` process is started it returns the port number in the standard output stream.
The `EslintBridgeProcess` needs to wait for this output to be returned so it can parse it and return the allocated port number.
As a result, the `EslintBridgeProcess.Start` method returns a `Task<int>`.

To make this method `async`, it starts a new task that is returned to the caller. The method that handles the output stream waits for the output, parses it to get the port number, then sets the result of the task.

However, it doesn't correctly handle the case where the process fails to start.
In that situation:
* the `startTask` is never completed.
* any subsequent analysis calls block in `EslintBridgeAnalyzer:EnsureEslintBridgeClientIsInitialized` on `serverInitLocker.WaitOne()`.
This will eventually lead to thread pool starvation.

ugras-ergun-sonarsource:
Fixes #3230

### DIFF
--- a/src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
+++ b/src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
@@ -133,6 +133,10 @@ namespace SonarLint.VisualStudio.TypeScript.Analyzer
                     shouldInitLinter = false;
                 }
             }
+            catch (Exception ex)
+            {
+                logger.WriteLine(ex.Message);
+            }
             finally
             {
                 serverInitLocker.Set();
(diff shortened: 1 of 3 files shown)


src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
```diff
@@ -133,6 +133,10 @@ private async Task EnsureEslintBridgeClientIsInitialized(CancellationToken cance
                     shouldInitLinter = false;
                 }
             }
+            catch (Exception ex)
```
ugras-ergun-sonarsource, line 136: For some reason If we don't add this, we publish whole stack trace of timeout exception to the output window. But when we add this, errors are not written at all. 🤔

rita-sonarsource:
@ugras-ergun-sonarsource @duncanp-sonar I submitted a fix that works but I feel it is a bit risky to merge it. I'll close this PR without merging and we could revisit it.

Issue #3195, opened 2022-08-18
GetGuidOfProject called on the wrong thread -> handled exception


duncanp-sonar:
### Description

The `TextBufferIssueTracker` calls `GetProjectGuid` on the wrong thread, causing VS to throw.
The exception is caught and logged -> no crash.

Unclear what the user impact is - analysis is not run for the specified file?

### Repro steps
Not sure - was debugging with standalone slvs-samples solution open and happened to notice this in the OutputWindow.

There were multiple file tabs visible, but I'd only opened a couple of files. I think the file `CognitiveComplexity.ts` file was top-most.

#### Stack trace
```
DEBUG: Failed to calculate project guid of file D:\repos\sq\slvs.samples\standalone\vs2022\NormalSolution\CSharp_SDKStyleProject_ConsoleApp\ts\CognitiveComplexity.ts, error occurred: System.Runtime.InteropServices.COMException (0x8001010E): GetGuidOfProjectFile must be called on the UI thread.
   at Microsoft.VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread(String callerMemberName)
   at Microsoft.VisualStudio.CommonIDE.Solutions.Solution.Microsoft.VisualStudio.Shell.Interop.IVsSolution5.GetGuidOfProjectFile(String pszProjectFile)
   at SonarLint.VisualStudio.Integration.Vsix.TextBufferIssueTracker.GetProjectGuid()
```

duncanp-sonar:
Another repro / call stack:
* opened a solution with open file -> analysis triggered.

#### Stack trace
```
Analyzing D:\repos\sq\slvs.samples\standalone\vs2022\NormalSolution\Cpp_WindowsProject_VS2022\Rules\S3135_v6_36.cpp
Refreshing PCH file for D:\repos\sq\slvs.samples\standalone\vs2022\NormalSolution\Cpp_WindowsProject_VS2022\Rules\S3135_v6_36.cpp. PCH file location: C:\Users\duncan.pocklington\AppData\Local\Temp\SLVS\PCH\876334b6-77f9-4f1c-9ba6-1a3611d08c5f\PCH.preamble
[...]

PR #3261, opened 2022-09-09
Added EmptyIssuesSnapshot

duncanp-sonar:
Relates to #3195


duncanp-sonar:
Related to #3195

This change adds an implementation of `IIssueSnapshot` for an empty snapshot i.e. one with no issues.

When creating an empty snapshot, we don't need the project name or guid and will never use them (when issues start arriving we create another completely new snapshot - we don't update the empty snapshot).

This removes one of the two places where we fetch the project name and guid. We still need the other one, but that will take more refactoring to fix.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -86,7 +86,7 @@ internal class TextBufferIssueTracker : IIssueTracker, ITagger<IErrorTag>, IDisp
             this.FilePath = document.FilePath;
             this.charset = document.Encoding.WebName;
 
-            this.Factory = new IssuesSnapshotFactory(new IssuesSnapshot(GetProjectName(), GetProjectGuid(), FilePath, new List<IAnalysisIssueVisualization>()));
+            this.Factory = new IssuesSnapshotFactory(EmptyIssuesSnapshot.Create(FilePath));
```
duncanp-sonar, line 89: @rita-sonarsource as discussed offline, I'll remove this parameter from the factory and make the class nested internal.

src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotFactoryTests.cs
```diff
@@ -32,12 +32,20 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.ErrorList
     [TestClass]
     public class IssuesSnapshotFactoryTests
     {
+        [TestMethod]
+        public void Create_CurrentSnapshotIsEmpty()
+        {
+            var testSubject = new IssuesSnapshotFactory("my file");
+            testSubject.CurrentSnapshot.GetType().Name.Should().Be("EmptyIssuesSnapshot");
```
duncanp-sonar, line 39: I made the nested class `private` so the test code can't see it at all -> need to check the type by name.

rita-sonarsource (approved):

PR #3267, opened 2022-09-15
Execute GetGuidOfProject on UI thread

### LINKED ISSUE #3195: GetGuidOfProject called on the wrong thread -> handled exception

duncanp-sonar:
Fixes #3195

### DIFF
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
@@ -126,7 +126,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             tagger.Should().NotBeNull();
 
             VerifyCheckedAnalysisIsSupported();
-            VerifyAnalysisWasRequested();
             mockAnalyzerController.VerifyNoOtherCalls();
         }
 
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
@@ -23,6 +23,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading;
+using System.Threading.Tasks;
 using EnvDTE;
 using EnvDTE80;
 using FluentAssertions;
@@ -85,7 +86,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintTagger
 
             return new TextBufferIssueTracker(taggerProvider.dte, taggerProvider,
                 mockedJavascriptDocumentFooJs.Object, javascriptLanguage,
-                mockSonarErrorDataSource.Object, vsSolution, issueConsumerFactory, logger);
+                mockSonarErrorDataSource.Object, vsSolution, issueConsumerFactory, logger,
+                new NoOpThreadHandler());
         }
 
         private static IIssueConsumerFactory CreateValidIssueConsumerFactory()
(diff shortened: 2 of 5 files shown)


duncanp-sonar:
The previous refactorings to fetch the project guid only once before making the request actually fixed most of the occurrences of the bug as the event handlers in `TextBufferIssueTracker` are called on the UI thread. That only left analyses triggered by disabling a rule.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -129,27 +131,19 @@ public void Ctor_CreatesFactoryWithCorrectData()
         }
 
         [TestMethod]
-        public void Ctor_FailsToRetrieveProjectGuid_NonCriticalException_ExceptionCaught()
```
duncanp-sonar, line 132: Previously the analysis was requested synchronously so we needed to catch exceptions for the initial analysis requested in the constructor to prevent crashes.

Now, the constructor still makes the analysis request, but it is handle asynchronously and exceptions aren't raised. I added a new test to check this.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -110,7 +128,7 @@ private void SafeOnFileActionOccurred(object sender, TextDocumentFileActionEvent
             }
             catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
             {
-                logger.WriteLine(Strings.Daemon_Editor_ERROR, ex);
```
duncanp-sonar, line 113: The old message wording wasn't particularly helpful so I updated it.

src/Integration.Vsix/SonarLintTagger/TextBufferIssueTracker.cs
```diff
@@ -124,18 +142,38 @@ private void SnapToNewSnapshot(IIssuesSnapshot newSnapshot)
 
         public void RequestAnalysis(IAnalyzerOptions options)
         {
-            FilePath = document.FilePath; // Refresh the stored file path in case the document has been renamed
+            RequestAnalysisAsync(options).Forget();
```
duncanp-sonar, line 145: Same pattern as for other cases where we want to switch from a `void` method to an `async` one.

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
```diff
@@ -126,7 +126,6 @@ public void CreateTagger_should_create_tracker_when_analysis_is_supported()
             tagger.Should().NotBeNull();
 
             VerifyCheckedAnalysisIsSupported();
-            VerifyAnalysisWasRequested();
```
duncanp-sonar, line 129: This is testing behaviour of the `TextBufferIssueTracker`.
It's tested in the `TextBufferIssueTrackerTests`, and it can't easily be tested here now as the request is no longer processed synchronously.

src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -338,15 +333,49 @@ public void RequestAnalysis_FileHasNoAssociatedProject_NoExceptionIsThrown()
...
+            var issueConsumerFactory = CreateIssueConsumerFactory(issueConsumer.Object);
+
+            // Clear the invocations that occurred during construction
+            testSubject = CreateTextBufferIssueTracker(issueConsumerFactory: issueConsumerFactory.Object);
+
+            issueConsumer.Setup(x => x.Accept(It.IsAny<string>(), It.IsAny<IEnumerable<IAnalysisIssue>>()))
+                .Throws(new InvalidOperationException("this is a test"));
+
+            Func<Task> act = () => testSubject.RequestAnalysisAsync(Mock.Of<IAnalyzerOptions>());
+            act.Should().NotThrow<InvalidOperationException>();
```
rita-sonarsource, line 359: Better to change it to NotThrow() (without type), so it would be more strict.

rita-sonarsource (approved):
Just a minor comment

Issue #3268, opened 2022-09-16
Embed SonarC# 8.45 and SonarVB 8.45


martin-strecker-sonarsource:
We enabled rule S2259 to use the new symbolic execution engine. With this change, we improve precision, support newer language features, and activated S2259 for VB.

Release notes: https://github.com/SonarSource/sonar-dotnet/releases/tag/8.45.0.54064

PR #3276, opened 2022-09-20
Update embedded SonarC# and SonarVB to 8.45

rita-sonarsource:
Fixes #[3268](https://github.com/SonarSource/sonarlint-visualstudio/issues/3268)


rita-sonarsource:
Closed due to regression: https://community.sonarsource.com/t/8-45-0-54064-many-false-positives-with-xx-is-null-on-at-least-one-execution-path/71593

Issue #2696, opened 2021-10-12
FileNotFoundException is thrown when analysis is invoked on temporary build files


rita-sonarsource:
#### Description

When opening a project for the first time, an exception is thrown:

```
Failed to analyze Temp.txt: System.IO.FileNotFoundException: Could not find file 'C:\Projects\sonarlint-visualstudio\src\Integration.Vsix\bin\Debug\VS2019\net46\Temp.txt'.
File name: 'C:\Projects\sonarlint-visualstudio\src\Integration.Vsix\bin\Debug\VS2019\net46\Temp.txt'
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
   at Microsoft.VisualStudio.Text.Implementation.TextDocumentFactoryService.OpenFileGuts(String filePath, DateTime& lastModifiedTimeUtc, Int64& fileSize)
   at Microsoft.VisualStudio.Text.Implementation.TextDocumentFactoryService.OpenFile(String filePath, DateTime& lastModifiedTimeUtc, Int64& fileSize)
   at Microsoft.VisualStudio.Text.Implementation.TextDocumentFactoryService.CreateAndLoadTextDocument(String filePath, IContentType contentType, Boolean attemptUtf8Detection, Boolean allowCompressedStorage, Boolean throwOnInvalidCharactersIfUnknownEncoding, Boolean& characterSubstitutionsOccurred)
   at Microsoft.VisualStudio.Text.Implementation.TextDocumentFactoryService.CreateAndLoadTextDocument(String filePath, IContentType contentType)
   at SonarLint.VisualStudio.CloudSecrets.SecretsAnalyzer.ExecuteAnalysis(String filePath, String charset, IEnumerable`1 detectedLanguages, IIssueConsumer consumer, IAnalyzerOptions analyzerOptions, CancellationToken cancellationToken)
```

TODO:
- [x] Check if there is a way not to invoke analysis on temp files
- [ ] Verify that the file exists before attempting to read it

rita-sonarsource:
[...]

PR #3279, opened 2022-09-22
Don't analyze temp files

rita-gorokhod:
Fixes [#2696](https://github.com/SonarSource/sonarlint-visualstudio/issues/2696)

### DIFF
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
@@ -125,6 +125,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             tagger.Should().NotBeNull();
 
             VerifyCheckedAnalysisIsSupported();
+            VerifyAnalysisWasRequested();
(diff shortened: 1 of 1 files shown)


src/IssueViz/Editor/TaggableBufferIndicator.cs
```diff
@@ -40,14 +54,7 @@ public bool IsTaggable(ITextBuffer buffer)
                 return false;
             }
 
-            var isOutputBuffer = buffer.ContentType.IsOfType("Output");
```
rita-sonarsource, line 43: Unfortunately this check was not enough.

When you first open a tool window in VS, VS creates behind the scenes a temp file called "Temp.txt". The content type of this file is whatever window you opened. I previously thought it happens just with output windows, hence the "IsOfType(Output)", but apparently it can happen for any tool window. So for example, if you open the command window, it will create a temp.txt with content type "Command". 

I tried to see if there is anything to identify these windows but there is nothing, and their content type inherits from "text", which we should analyze because we support secrets.

So in the end I opted for checking if it really exists or not, and that seems to work.

ugras-ergun-sonarsource (approved):

PR #3282, opened 2022-09-27
 [Infra] Consider changing pipelines to yaml

### LINKED ISSUE #2234: [Infra] Consider changing pipelines to yaml
### Description

Consider whether to change the existing build and release pipeline from the "classic" designer-based definitions to yaml.

#### Pros
* the definition is versioned and stored with the code so it's easier to make changes like enabling NuGet package locks.
* a definition can always be recovered if deleted accidentally or deliberately
* more fashionable

#### Cons
* less readable - the designer gives an better overall high-level picture of the pipeline
* involves effort, and using the classic designer hasn't caused a problem so far. Is it worth it?

ugras-ergun-sonarsource:
fixes #2234


azure-pipelines.yml
```diff
@@ -0,0 +1,321 @@
...
+    - master
+    - dev/*
+    - feature/*
+    - branch-*
+  batch: True
+jobs:
+- job: Phase_1
+  strategy:
+    maxParallel: 2
+    matrix:
```
ugras-ergun-sonarsource, line 17: running multiple configuration from a variable is not supported in yaml so I had to change it to a matrix strategy.

ugras-ergun-sonarsource:
Most of the yaml file is auto-generated by exporting pipeline to yaml in azure devops. I had to make few changes to make it work. Only one of the changes were significant though I have commented on that.

azure-pipelines.yml
```diff
@@ -0,0 +1,322 @@
+variables:
+- group: sonarsource-build-variables
+- name: BuildParameters.solution
+  value: SonarLint.VisualStudio.Integration.sln
+name: $(Build.BuildId)
```
rita-sonarsource, line 5: Is this what makes the new names "wrong"?

azure-pipelines.yml
```diff
@@ -0,0 +1,322 @@
...
+trigger:
+  branches:
+    include:
+    - master
+    - dev/*
```
rita-sonarsource, line 10: I think we can remove dev and feature, we don't use them

azure-pipelines.yml
```diff
@@ -0,0 +1,322 @@
...
+jobs:
+- job: Phase_1
+  strategy:
+    maxParallel: 2
+    matrix:
+      2019:
+        vsTargetVersion: 2019
+      2022:
+        vsTargetVersion: 2022      
+  displayName: Phase 1
```
rita-sonarsource, line 23: Can we have meaningful names for the phases? would make it easier

azure-pipelines.yml
```diff
@@ -0,0 +1,322 @@
...
+  - checkout: self
+    submodules: recursive
+    fetchTags: false
+    persistCredentials: True
+  - task: PowerShell@2
+    displayName: Calculate if should sign Vsix
+    continueOnError: True
+    inputs:
+      targetType: inline
+      script: >-
```
rita-sonarsource, line 37: It pretty messy to have the scripts inline in one file. Can we load it from disk and commit to the repo these scripts?

azure-pipelines.yml
```diff
@@ -0,0 +1,322 @@
...
+    condition: and(succeeded(), eq(variables['vsTargetVersion'], '2022'))
+    inputs:
+      command: custom
+      solution: $(BuildParameters.solution)
+      feedRestore: 399fb241-ecc7-4802-8697-dcdd01fbb832/423b576f-2263-43a1-93bd-69f4def19102
+      includeNuGetOrg: false
+      nugetConfigPath: nuget.config
+      arguments: restore -LockedMode -Verbosity detailed
+  - task: NuGetCommand@2
+    displayName: NuGet restore without LockedMode check (VS2017 and VS2019)
```
rita-sonarsource, line 101: time to update the name and remove 2017

azure-pipelines.yml
```diff
@@ -0,0 +1,322 @@
...
...    Param([string] $propertyPath, [string] $propertyValue)\n\n    Write-Host \"Setting the build property $propertyPath to $propertyValue ...\"\n\n    # NB: the WebClient class defaults to TLS v1, which is no longer supported by some online providers.\n    # This setting isn't required on MS-hosted agents, but it can be for local agents.\n    [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12\n\n    $rootApiUrl = \"$($env:SYSTEM_TEAMFOUNDATIONCOLLECTIONURI)$env:SYSTEM_TEAMPROJECTID/_apis/\" \n\n    # Update build properties\n    $url =$rootApiUrl + \"build/builds/$env:BUILD_BUILDID/properties?api-version=5.0-preview.1\"\n\n    #$jsonPatch = '[{ \"op\": \"replace\", \"path\": \"/$($propertyPath)\", \"value\": \"' + $($propertyValue) + '\" }]'\n\n    $jsonPatch = \"[{ 'op': 'replace', 'path': '/$propertyPath', 'value': '$propertyValue' }]\"\n\n    $pipeline = Invoke-RestMethod -Uri $url -Headers @{\n        \"Authorization\" = \"Bearer $env:SYSTEM_ACCESSTOKEN\";\n    } -Body $jsonPatch -Method Patch -ContentType \"application/json-patch+json\"\n\n}\n\n\nSetBuildProperty -propertyPath \"VsixVersion\" -propertyValue $env:SONAR_PROJECT_VERSION"
```
rita-sonarsource, line 160: this is unmaintainable, anything we can do?

azure-pipelines.yml
ugras-ergun-sonarsource: I've kept the original naming. But it makes sense to make it something else.

rita-sonarsource (changes requested):
I think it will be difficult to maintain in this state. We need to find a way to split it to smaller files, if possible.

pipeline/azure-pipelines.yml
```diff
@@ -0,0 +1,215 @@
...
+    inputs:
+      downloadDirectory: $(System.DefaultWorkingDirectory)\Whitesource
+      feedListDownload: 399fb241-ecc7-4802-8697-dcdd01fbb832/69e11f1b-cad7-442a-b676-37722e207221
+      packageListDownload: agent
+      versionListDownload: '*'
+  - task: PowerShell@2
+    displayName: Run WhiteSource (Mend) scan
+    condition: and(succeeded(), or(eq(variables['ForceWhiteSourceScan'], 'true'), eq(variables['Build.SourceBranch'], 'refs/heads/master')))
+    inputs:
+      targetType: inline
```
ugras-ergun-sonarsource, line 129: kept this one inline because it was already a one-liner. It didn't make sense to extract the script to a file.

pipeline/scripts/should-sign-vsix.ps1
```diff
@@ -0,0 +1,48 @@
+param([string] $SourceBranch, [string] $SignVsixIfMasterBranch, [string] $ForceSignVsix)
```
ugras-ergun-sonarsource, line 1: variables are not accessible in files. So they need to be passed as parameters.

pipeline/scripts/trigger-dogfood.ps1
```diff
@@ -0,0 +1,20 @@
+param($SourceBranch, $TriggerDogfoodReleaseIfMasterBranch)
```
ugras-ergun-sonarsource, line 1: variables are not accessible in files. So they need to be passed as parameters.

pipeline/azure-pipelines.yml
```diff
@@ -0,0 +1,215 @@
...
+  - checkout: self
+    submodules: recursive
+    fetchTags: false
+    persistCredentials: True
+  - task: PowerShell@2
+    displayName: Calculate if should sign Vsix
+    inputs:
+      targetType: filePath
+      filePath: $(System.DefaultWorkingDirectory)\pipeline\scripts\should-sign-vsix.ps1
+      arguments: -SourceBranch $(Build.SourceBranch) -SignVsixIfMasterBranch $(SignVsixIfMasterBranch) -ForceSignVsix $(ForceSignVsix)
```
ugras-ergun-sonarsource, line 35: variables are not accessible in files. So they need to be passed as parameters.

pipeline/azure-pipelines.yml
```diff
@@ -0,0 +1,215 @@
...
+  - task: SonarCloudPublish@1
+    displayName: Publish Quality Gate Result (VS2022 only)
+    condition: and(succeeded(), eq(variables['vsTargetVersion'], '2022'))
+  - task: PowerShell@2
+    displayName: Set trigger dogfood VSIX build tag (VS2022 only)
+    condition: and(succeeded(), eq(variables['vsTargetVersion'], '2022'))
+    inputs:
+      targetType: filePath
+      filePath: $(System.DefaultWorkingDirectory)\pipeline\scripts\trigger-dogfood.ps1
+      arguments: -SourceBranch $(Build.SourceBranch) -TriggerDogfoodReleaseIfMasterBranch $(TriggerDogfoodReleaseIfMasterBranch)
```
ugras-ergun-sonarsource, line 214: variables are not accessible in files. So they need to be passed as parameters.

pipeline/azure-pipelines.yml
```diff
@@ -0,0 +1,215 @@
+variables:
```
duncanp-sonar, line 1: The pipeline is hard to follow in its current form - whitespace between the sections/tasks would make it easier to read e.g. the [sonar-secrets dotnet yaml file](https://github.com/SonarSource/sonar-secrets/blob/master/sonar-secrets-dotnet/azure-pipelines-dotnet.yml).

pipeline/azure-pipelines.yml
duncanp-sonar: I thought variables were made available to PowerShell scripts as environment variables (and some of the scripts refer to pre-defined variables e.g. `$env:SYSTEM_TEAMPROJECTID`).

However, there's no harm in being explicit about which variables are being passed.

ugras-ergun-sonarsource: This was the solution I was able to find. I think using environment variables is cleaner, so I'll move to that.

pipeline/azure-pipelines.yml
```diff
@@ -0,0 +1,244 @@
...
+  displayName: "Build:"
+  cancelTimeoutInMinutes: 1
+  pool:
+    vmImage: windows-2022
+    
+  steps:
+  - checkout: self
+    submodules: recursive
+    fetchTags: false
+    persistCredentials: True
```
duncanp-sonar, line 33: Why do we need to set `persistCredentials` to true? (looks like the [default ](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-checkout?view=azure-pipelines)is false).

ugras-ergun-sonarsource: This was the result of the export.

pipeline/azure-pipelines.yml
```diff
@@ -0,0 +1,244 @@
...
+  - task: SonarCloudPrepare@1
+    displayName: Prepare analysis on SonarCloud (VS2022 only)
+    condition: and(succeeded(), eq(variables['vsTargetVersion'], '2022'))
+    inputs:
+      SonarCloud: 4d6d3285-de8e-41bb-a346-987397025edb
+      organization: sonarsource
+      projectKey: sonarlint-visualstudio
+      projectName: SonarLint for Visual Studio
+      projectVersion: $(SONAR_PROJECT_VERSION)
+      extraProperties: "# Additional properties that will be passed to the scanner, \n# Put one key=value per line, example:\n# sonar.exclusions=**/*.bin\n#sonar.cs.vstest.reportsPaths=$(Agent.TempDirectory)/**/*.trx\n#sonar.cs.vscoveragexml.reportsPaths=$(Agent.TempDirectory)/**/*.coveragexml"
```
duncanp-sonar, line 121: This doesn't look right - everything is on one line.
However, I don't think it matters since all of the extra properties are commented out. If that is correct then it would be better to just remove the `extraProperities`.

pipeline/azure-pipelines.yml
```diff
@@ -0,0 +1,244 @@
...
+
+  - task: DotNetCoreCLI@2
+    displayName: Generate cycloneDX sbom file
+    condition: and(succeeded(), eq(variables['Build.SourceBranchName'], 'master'))
+    inputs:
+      command: custom
+      projects: SonarLint.VisualStudio.Integration.sln
+      custom: CycloneDX
+      arguments: -t -j -o binaries
+  - task: DownloadSecureFile@1
```
duncanp-sonar, line 166: Minor - missing blank line between tasks

pipeline/azure-pipelines.yml
ugras-ergun-sonarsource: Again this is the result of export. And Information is copied from the job. If this would not work we can remove I guess. 

image

pipeline/azure-pipelines.yml
```diff
@@ -0,0 +1,244 @@
...
+    condition: and(succeeded(), eq(variables['Build.SourceBranchName'], 'master'))
+    inputs:
+      targetType: filePath
+      filePath: $(System.DefaultWorkingDirectory)\pipeline\scripts\sbom.ps1
+
+  - task: CopyFiles@2
+    displayName: 'Copy VSIX to: $(Build.ArtifactStagingDirectory)'
+    inputs:
+      SourceFolder: $(Build.SourcesDirectory)\binaries
+      Contents: >-
```
duncanp-sonar, line 184: I'm guessing the extra line-feed between the line settings required (i.e. line 186) is required here because you are using the [block chomping syntax](https://yaml-multiline.info/) `>-`.

It might be simpler to use the literal syntax and get rid of the extra blank line i.e. 
```
    Contents: |
     **\*.vsix
     **\SonarLint.visualstudio.sbom*.*
```

pipeline/azure-pipelines.yml
```diff
@@ -0,0 +1,244 @@
...
+
+  - task: PublishBuildArtifacts@1
+    displayName: 'Publish Artifact: vsix'
+    inputs:
+      ArtifactName: vsix
+
+  - task: VSTest@2
+    displayName: VsTest - testAssemblies
+    inputs:
+      testAssemblyVer2: >-
```
duncanp-sonar, line 219: Same comment about "block chomping syntax" here.

duncanp-sonar (commented):
Nice! There are a couple more GUIDs that could be replaced with file names, but otherwise looks good.

Have you tried forcing signing the VSIX / forcing a WhiteSource scan to check they work ok?

ugras-ergun-sonarsource:
> Have you tried forcing signing the VSIX / forcing a WhiteSource scan to check they work ok?

I am sure I did for dogfood release but not sure about WhiteSource. I'll run one with the force variable set to true just in case.

duncanp-sonar (approved):

PR #3292, opened 2022-10-11
Fix SBOM generation in pipeline

duncanp-sonar:
Relates to #2234


pipeline/scripts/sbom.ps1
```diff
@@ -1,16 +1,18 @@
 # Calculate the file name
-$sbomName = "SonarLint.visualstudio.sbom-$env:SONAR_PROJECT_VERSION.$env:BUILD_BUILDID-$(vsTargetVersion).json"
```
duncanp-sonar, line 2: Not all of the variable references had been converted to the `$env:xxx` format -> errors when building in `master`.

duncanp-sonar:
I ran the pipeline with `ForceSbomGeneration=true` to check that it works as expected:

image

ugras-ergun-sonarsource (approved):

PR #3310, opened 2022-11-14
Update embedded CFamily analyzer to 6.39.0.60440

### LINKED ISSUE #3287: Update embedded CFamily analyzer to 6.39.0.60440
6.38.0.60268 → 6.39.0.60440: bug fixes and improvements

Release notes: https://sonarsource.atlassian.net/issues/?jql=project%20%3D%2010166%20AND%20fixVersion%20%3D%2013885%20ORDER%20BY%20priority%20DESC%2C%20key%20ASC

6.37.0.60132 → 6.38.0.60268: bug fixes and improvements

Release notes: https://sonarsource.atlassian.net/issues/?jql=project%20%3D%2010166%20AND%20fixVersion%20%3D%2013853%20ORDER%20BY%20priority%20DESC%2C%20key%20ASC

ugras-ergun-sonarsource:
fixes #3287


src/CFamily.UnitTests/Subprocess/ProtocolTest.cs
```diff
@@ -67,6 +67,22 @@ public void Read_Response()
             response.Messages[0].Parts.Length.Should().Be(1);
         }
 
+        [TestMethod]
+        public void Response_WithDataFlow_DataFlowIgnored()
```
rita-sonarsource, line 71: I wonder if it would actually be easier to have a DataFlows property on the message, and to populate it? This test is a bit weak as the assertions don't test anything related to data flows. Perhaps it would be easier to populate the flows and assert them.

rita-sonarsource (changes requested):
See my comment regarding ignoring the data flows

src/CFamily.UnitTests/Subprocess/ProtocolTest.cs
ugras-ergun-sonarsource: It would increase the scope of the testing and the code with no real benefit it would be discarded anyway. The purpose of this test is to just make sure that we correctly discard the data flows.

rita-sonarsource: My problem with the test is that you can change `MockResponseWithDataFlow` to `MockResponseWithQuickFixes` and nothing would fail. There is no assertion anywhere that the data flows were actually attempted to be parsed. I'm not sure how we can assert it otherwise, without actually populating the fields. Do you have any suggestions? :/

ugras-ergun-sonarsource: This would require we would change the Message class and its ctor, which is referenced in 35 places (albeit almost all of them are tests). I do not see the need to assert that data is correctly parsed, as we do not care about it. We just need to assert that the existence of this data will not break our process, and we are asserting that.

PR #3311, opened 2022-11-14
Remove TaskScheduler.Default from CLangAnalyzer and update tests

bigfluffycookie:


src/CFamily.UnitTests/Analysis/CLangAnalyzerTests.cs
```diff
@@ -285,6 +287,27 @@ public async Task TriggerAnalysisAsync_AnalysisFailsDueToInternalMessage_Notifie
             statusNotifier.VerifyNoOtherCalls();
         }
 
+        [TestMethod]
```
bigfluffycookie, line 290: This is what I ended up with, I wasn't sure if I needed to test anything ontop of the callback order so please let me know

src/CFamily.UnitTests/Analysis/CLangAnalyzerTests.cs
```diff
@@ -309,16 +332,20 @@ private static Mock<IRequestFactoryAggregate> CreateRequestFactory(string filePa
             IAnalysisStatusNotifier statusNotifier = null,
             ICFamilyIssueConverterFactory issueConverterFactory = null,
             IRequestFactoryAggregate requestFactory = null,
-            ILogger logger = null)
+            ILogger logger = null,
+            IFileSystem fileSystem = null,
```
bigfluffycookie, line 336: I added in the filesystem here aswell as it makes sense for it to be there

src/CFamily.UnitTests/Analysis/CLangAnalyzerTests.cs
duncanp-sonar: I think that's the only new thing we want to test here.

src/CFamily/Analysis/CLangAnalyzer.cs
```diff
@@ -122,7 +126,7 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
             // For notes on VS threading, see https://github.com/microsoft/vs-threading/blob/master/doc/cookbook_vs.md
```
duncanp-sonar, line 126: Minor: I would probably delete this comment: we're now using our own abstraction layer, so the cookbook isn't directly relevant.

duncanp-sonar (commented):
Nice: just a couple of minor comments.

src/CFamily.UnitTests/Analysis/CLangAnalyzerTests.cs
```diff
@@ -285,6 +287,27 @@ public async Task TriggerAnalysisAsync_AnalysisFailsDueToInternalMessage_Notifie
             statusNotifier.VerifyNoOtherCalls();
         }
 
+        [TestMethod]
+        public async Task TriggerAnalysisAsync_ThreadHandling_CorrectCallbackOrder()
```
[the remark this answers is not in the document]
bigfluffycookie: Good point, I sometimes get stuck on the naming convention of "methodName_TestScenario_ExpectedResult" so sometimes I come up with strange names trying to match it

PR #3316, opened 2022-11-16
Lb/issue 3145

bigfluffycookie:
A lot of the code in the different classes is more or less the same however the tests are all very different. Some test classes were written with a TestInitializer and I tried my best to conform to each classes standard as to not change too much.

### DIFF
--- a/src/Core/IThreadHandling.cs
+++ b/src/Core/IThreadHandling.cs
@@ -54,6 +54,14 @@ namespace SonarLint.VisualStudio.Core
         Task RunOnUIThread(Action op);
 
         /// <summary>
+        /// Executes the operation synchronously on the main thread.
+        /// If the caller is on the main thread already then the operation is executed directly.
+        /// If the caller is not on the main thread then the method will switch to the main thread,
+        /// then resume on the caller's thread when then the operation completes.
+        /// </summary>
+        void RunOnUIThreadSync(Action op);
+
+        /// <summary>
         /// Executes the operation asynchronously on the background thread.
         /// If the caller is on a background thread already then the operation is executed directly.
         /// If the caller is not on a background thread then the method will switch to one,
--- a/src/Infrastructure.VS.UnitTests/ActiveDocumentLocatorTests.cs
+++ b/src/Infrastructure.VS.UnitTests/ActiveDocumentLocatorTests.cs
@@ -25,6 +25,7 @@ using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.Text;
 using Moq;
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration.UnitTests;
 
 namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
@@ -41,8 +42,6 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
         [TestInitialize]
         public void TestInitialize()
         {
-            ThreadHelper.SetCurrentThreadAsUIThread();
-
             monitorSelectionMock = new Mock<IVsMonitorSelection>();
             textDocumentProviderMock = new Mock<ITextDocumentProvider>();
             textDocument = Mock.Of<ITextDocument>();
@@ -52,7 +51,26 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
                 .Setup(x => x.GetService(typeof(SVsShellMonitorSelection)))
                 .Returns(monitorSelectionMock.Object);
 
-            testSubject = new ActiveDocumentLocator(serviceProviderMock.Object, textDocumentProviderMock.Object);
+            testSubject = new ActiveDocumentLocator(serviceProviderMock.Object, textDocumentProviderMock.Object, new NoOpThreadHandler());
+        }
+
+        [TestMethod]
+        public void Ctor_VerifyIsRunningOnUiThread()
+        {
+            var serviceProvider = new Mock<IServiceProvider>();
+            var threadHandling = new Mock<IThreadHandling>();
+            threadHandling.Setup(x => x.RunOnUIThread(It.IsAny<Action>()))
+                .Callback<Action>(op =>
+                {
+                    // Try to check that the product code is executed inside the "RunOnUIThread" call
+                    serviceProvider.Invocations.Count.Should().Be(0);
+                    op();
+                    serviceProvider.Invocations.Count.Should().Be(1);
+                });
+
+            _ = new ActiveDocumentLocator(serviceProvider.Object, textDocumentProviderMock.Object, threadHandling.Object);
+
+            threadHandling.Verify(x => x.RunOnUIThread(It.IsAny<Action>()), Times.Once);
         }
 
         [TestMethod]
--- a/src/Infrastructure.VS.UnitTests/ActiveDocumentLocatorTests.cs
+++ b/src/Infrastructure.VS.UnitTests/ActiveDocumentLocatorTests.cs
@@ -49,7 +49,7 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
             textDocument = Mock.Of<ITextDocument>();
             
             threadHandlingMock = new Mock<IThreadHandling>();
-            threadHandlingMock.Setup(x => x.RunOnUIThread(It.IsAny<Action>())).Callback<Action>(op => op());
+            threadHandlingMock.Setup(x => x.RunOnUIThreadSync(It.IsAny<Action>())).Callback<Action>(op => op());
 
             var serviceProviderMock = new Mock<IServiceProvider>();
             serviceProviderMock
@@ -76,7 +76,7 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
 
             _ = new ActiveDocumentLocator(serviceProvider.Object, textDocumentProviderMock.Object, threadHandling.Object);
 
-            threadHandling.Verify(x => x.RunOnUIThread(It.IsAny<Action>()), Times.Once);
+            threadHandling.Verify(x => x.RunOnUIThreadSync(It.IsAny<Action>()), Times.Once);
         }
 
         [TestMethod]
--- a/src/Infrastructure.VS.UnitTests/ActiveDocumentLocatorTests.cs
+++ b/src/Infrastructure.VS.UnitTests/ActiveDocumentLocatorTests.cs
@@ -21,6 +21,7 @@
 using System;
 using FluentAssertions;
 using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.Text;
@@ -35,6 +36,7 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
     {
         private Mock<IVsMonitorSelection> monitorSelectionMock;
         private Mock<ITextDocumentProvider> textDocumentProviderMock;
+        private Mock<IThreadHandling> threadHandlingMock;
         private ITextDocument textDocument;
 
         private ActiveDocumentLocator testSubject;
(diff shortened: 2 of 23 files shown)


src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDEFailureInfoBarTests.cs
```diff
@@ -189,6 +187,33 @@ public async Task InfoBarButtonClicked_OutputWindowIsShown()
...
+            var infoBarManager = new Mock<IInfoBarManager>();
+            infoBarManager.Setup(x => x.AttachInfoBarWithButton(ValidToolWindowId, It.IsAny<string>(), It.IsAny<string>(), default));
+
+            var threadHandling = new Mock<IThreadHandling>();
+
+            var testSubject = new OpenInIDEFailureInfoBar(infoBarManager.Object, Mock.Of<IOutputWindowService>(), threadHandling.Object);
+
+            await testSubject.ShowAsync(ValidToolWindowId);
+
+            threadHandling.Verify(x => x.RunOnUIThread(It.IsAny<Action>()), Times.Once);
```
bigfluffycookie, line 202: I can't easily test if the methods that get called are running in the ui thread as they are private so I decided to just test if RunOnUIThread gets called.

src/IssueViz.UnitTests/Editor/DocumentNavigatorTests.cs
```diff
@@ -42,14 +43,13 @@ public class DocumentNavigatorTests
 
         private Mock<IOutliningManagerService> outliningManagerServiceMock;
         private Mock<IEditorOperations> editorOperationsMock;
+        private Mock<IThreadHandling> threadHandlingMock;
```
bigfluffycookie, line 46: I tried staying consitent with how the tests were already written instead of changing everything which is why its here

src/IssueViz.UnitTests/Editor/DocumentNavigatorTests.cs
```diff
@@ -64,10 +64,15 @@ public void TestInitialize()
                 .Setup(x => x.GetEditorOperations(mockTextView))
                 .Returns(editorOperationsMock.Object);
 
+            threadHandlingMock = new Mock<IThreadHandling>();
+            threadHandlingMock.Setup(x => x.RunOnUIThread(It.IsAny<Action>())).Callback<Action>(op => op());
```
bigfluffycookie, line 68: This is not great however, I think its the best I can do with the way this testclass is written.

src/IssueViz.UnitTests/Editor/DocumentNavigatorTests.cs
```diff
@@ -125,5 +131,23 @@ public void Navigate_SpanIsSelected()
...
             editorOperationsMock.VerifyNoOtherCalls();
         }
+
+        [TestMethod]
+        public void Navigate_VerifyIsRunningOnUIThread()
+        {
+            threadHandlingMock.Reset();
+            threadHandlingMock.Setup(x => x.RunOnUIThread(It.IsAny<Action>())).Callback<Action>(op =>
+            {
+                outliningManagerServiceMock.Invocations.Count.Should().Be(0);
```
bigfluffycookie, line 141: Check if methods get invoked while running on ui thread

src/Infrastructure.VS/DocumentEvents/ActiveDocumentTracker.cs
```diff
@@ -107,7 +111,7 @@ int IVsSelectionEvents.OnElementValueChanged(uint elementId, object oldValue, ob
             ActiveDocumentChanged?.Invoke(this, new ActiveDocumentChangedEventArgs(activeTextDoc));
 
         private static bool IsDocumentFrame(IVsWindowFrame frame) =>
-            ErrorHandler.Succeeded(frame.GetProperty((int) __VSFPROPID.VSFPROPID_Type, out var frameType)) &&
+            Microsoft.VisualStudio.ErrorHandler.Succeeded(frame.GetProperty((int) __VSFPROPID.VSFPROPID_Type, out var frameType)) &&
```
bigfluffycookie, line 114: Needed to add this because of ambiguous reference issue

src/Infrastructure.VS.UnitTests/ActiveDocumentTrackerTests.cs
```diff
@@ -242,19 +271,24 @@ public void OnElementValueChanged_NoEventSubscribers_DoesNotFail(VSConstants.VSS
...
             textDocumentProvider ??= Mock.Of<ITextDocumentProvider>();
+            threadHandling ??= new NoOpThreadHandler();
 
             var serviceProviderMock = new Mock<IServiceProvider>();
             serviceProviderMock
                 .Setup(x => x.GetService(typeof(SVsShellMonitorSelection)))
                 .Returns(monitorSelection);
 
-            var testSubject = new ActiveDocumentTracker(serviceProviderMock.Object, textDocumentProvider);
+            serviceProvider ??= serviceProviderMock.Object;
```
bigfluffycookie, line 289: Added this in so you can also test it

src/Integration.Vsix.UnitTests/Helpers/StatusBarNotifierTests.cs
```diff
@@ -19,32 +19,38 @@
  */
```
bigfluffycookie, line 19: This test class I rewrote mostly as the TestInitializer wasnt needed anymore and this way it was easier for me to add more tests.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -129,8 +131,22 @@ public ITaintIssueViewModel SelectedIssue
             ITelemetryManager telemetryManager,
             IIssueSelectionService selectionService,
             ICommand navigateToDocumentationCommand,
-            IMenuCommandService menuCommandService)
+            IMenuCommandService menuCommandService) :
+            this(store, locationNavigator, activeDocumentTracker,activeDocumentLocator,showInBrowserService,
+                 telemetryManager,selectionService,navigateToDocumentationCommand,menuCommandService, new ThreadHandling()){ }
+
+        internal TaintIssuesControlViewModel(ITaintStore store,
```
bigfluffycookie, line 138: Added internal Ctor for testing purposes

src/IssueViz.UnitTests/Editor/QuickActions/IssueLocationActionsSourceTests.cs
```diff
@@ -149,6 +150,33 @@ public void OnTagsChanged_HasSubscribersToSuggestedActionsChanged_RaisesSuggeste
             eventHandler.Verify(x => x(It.IsAny<object>(), It.IsAny<EventArgs>()), Times.Once);
         }
 
+        [TestMethod]
+        public void OnTagsChanged_RunsOnUIThread()
+        {
```
bigfluffycookie, line 155: Most of this code is copied from other test methods, I've only added the threadhandling bits.

src/IssueViz/Editor/QuickActions/IssueLocationActionsSource.cs
```diff
@@ -44,17 +47,27 @@ internal sealed class IssueLocationActionsSource : ISuggestedActionsSource
...
 
-        public IssueLocationActionsSource(ILightBulbBroker lightBulbBroker, 
+        public IssueLocationActionsSource(ILightBulbBroker lightBulbBroker,
+            IVsUIShell vsUiShell,
+            IBufferTagAggregatorFactoryService bufferTagAggregatorFactoryService,
+            ITextView textView,
+            IIssueSelectionService selectionService) :
+            this (lightBulbBroker, vsUiShell, bufferTagAggregatorFactoryService, textView, selectionService, new ThreadHandling()) { }
+
+        internal IssueLocationActionsSource(ILightBulbBroker lightBulbBroker,
```
bigfluffycookie, line 59: Added internal ctor for testing purposes.

src/Infrastructure.VS.UnitTests/ActiveDocumentLocatorTests.cs
```diff
@@ -34,25 +36,57 @@ public class ActiveDocumentLocatorTests
...
         public void TestInitialize()
         {
-            ThreadHelper.SetCurrentThreadAsUIThread();
-
             monitorSelectionMock = new Mock<IVsMonitorSelection>();
             textDocumentProviderMock = new Mock<ITextDocumentProvider>();
             textDocument = Mock.Of<ITextDocument>();
+            
+            threadHandlingMock = new Mock<IThreadHandling>();
+            threadHandlingMock.Setup(x => x.RunOnUIThread(It.IsAny<Action>())).Callback<Action>(op => op());
```
rita-sonarsource, line 52: FYI, we have the class NoOpThreadHandling that essentially does exactly this.
You don't need to change it now but just FYI

src/Infrastructure.VS.UnitTests/ActiveDocumentTrackerTests.cs
```diff
@@ -110,6 +126,19 @@ public void OnSelectionChanged_DoesNotRaiseEvent()
             CheckEventNotRaised(eventHandler);
         }
 
+        [TestMethod]
+        public void OnElementValueChanged_ThrowIfNotOnUIThreadCalled()
+        {
+            var threadHandlingMock = new Mock<IThreadHandling>();
+            threadHandlingMock.Setup(x => x.RunOnUIThread(It.IsAny<Action>())).Callback<Action>(op => op());
+
+            var eventHandler = new Mock<Action<ActiveDocumentChangedEventArgs>>();
```
rita-sonarsource, line 135: eventHandler is not really used, right?

src/Infrastructure.VS/ActiveDocumentLocator.cs
```diff
@@ -40,22 +42,24 @@ public interface IActiveDocumentLocator
...
         [ImportingConstructor]
-        public ActiveDocumentLocator([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider, ITextDocumentProvider textDocumentProvider)
+        public ActiveDocumentLocator([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider, ITextDocumentProvider textDocumentProvider, IThreadHandling threadHandling)
         {
             this.textDocumentProvider = textDocumentProvider;
-            
-            RunOnUIThread.Run(() =>
+            this.threadHandling = threadHandling;
+
+            threadHandling.RunOnUIThread(() =>
```
rita-sonarsource, line 54: Looking again at this code, I think running it asynchronically is not the right way to go. I think it's better to do "throwIfNotOnUiThread" and then simply call it, because we expect all MEF ctors to be called on the UI thread. WDYT? cc @duncanp-sonar

src/Integration.Vsix/Helpers/IStatusBarNotifier.cs
```diff
@@ -36,25 +38,28 @@ internal interface IStatusBarNotifier
...
+            this.threadHandling = threadHandling;
+
+            threadHandling.RunOnUIThread(() =>
             {
                 vsStatusBar = serviceProvider.GetService(typeof(IVsStatusbar)) as IVsStatusbar;
-            });
+            }).Forget();
         }
 
         public void Notify(string message, bool showSpinner)
```
rita-sonarsource, line 54: I think we should open a ticket to change this method to be async so that the caller could decide if to call Forget or await.

src/Infrastructure.VS/ActiveDocumentLocator.cs
duncanp-sonar: @rita-sonarsource I _thought_ that was the case, but I'm no longer sure. I can't remember if it was something I read or something noticed when debugging. I'll have a look.

src/IssueViz.Security.UnitTests/OpenInIDE/Api/OpenInIDEFailureInfoBarTests.cs
rita-sonarsource: How about testing ` infoBarManager.AttachInfoBarWithButton` ?

rita-sonarsource (changes requested):
LGTM but see my comment regarding calling .Forget in the ctors. I think in practice it would be ok since we _should_ be on the UI thread, but it might be better to explicitly call ThrowIfNotOnUIThread (and would make the tests easier too).

src/Infrastructure.VS/ActiveDocumentLocator.cs
duncanp-sonar: From the [VS Threading Cookbook](https://github.com/microsoft/vs-threading/blob/main/doc/cookbook_vs.md#how-do-i-effectively-verify-that-my-code-is-fully-free-threaded):

> Code is free-threaded if it... can complete on the caller's thread, no matter which thread that is.
Important places to be free-threaded include:
> 1. Code that runs in the MEF activation code paths (e.g. importing constructors, OnImportsSatisfied callbacks, and any code called therefrom).


So MEF constructors should not assume that they are called on the main thread.

src/Core/IThreadHandling.cs
```diff
@@ -53,6 +53,14 @@ public interface IThreadHandling
         /// </summary>
```
bigfluffycookie, line 53: I will create a seperate pr where I will rename RunOnUIThread to -> RunOnUIThreadAsync and RunOnUIThreadSync to RunOnUIThread. I didn't do it here as this pr already has a lot of changes and it would crowd it.

src/Core/IThreadHandling.cs
```diff
@@ -53,6 +53,14 @@ public interface IThreadHandling
...
         Task RunOnUIThread(Action op);
 
+        /// <summary>
+        /// Executes the operation synchronously on the main thread.
+        /// If the caller is on the main thread already then the operation is executed directly.
+        /// If the caller is not on the main thread then the method will switch to the main thread,
+        /// then resume on the caller's thread when then the operation completes.
+        /// </summary>
+        void RunOnUIThreadSync(Action op);
```
bigfluffycookie, line 62: I added this as I think it makes more sense to have the run sync method and call it in the places that I had previously added "Forget" as these methods depend on it not being called async.

Issue #3305, opened 2022-11-11
Error binding solution when checking if a project contains JS/TS files


duncanp-sonar:
### Description

A null-ref exception occurs during binding, preventing the binding operation from succeeding.

Source: [this](https://community.sonarsource.com/t/having-issue-while-connecting-sonarqube-using-sonarlint-for-vs/73400/2) community forum thread.

#### Error logs

Note: these error logs were produced using a special build of SLVS, with the kind assistance of the user who reported the issue.

**Log 1**

```
SonarLint.VisualStudio.Integration.ErrorListInfoBarController: Unexpected error: System.NullReferenceException: Object reference not set to an instance of an object.
   at SonarLint.VisualStudio.Integration.Binding.JsTsProjectTypeIndicator.HasJsTsFileInProject(ProjectItems projectItems)
   at SonarLint.VisualStudio.Integration.Binding.JsTsProjectTypeIndicator.IsJsTs(Project dteProject)
   at SonarLint.VisualStudio.Integration.ProjectToLanguageMapper.GetLanguagesInProject(Project dteProject)
   at SonarLint.VisualStudio.Integration.ProjectToLanguageMapper.GetAllBindingLanguagesForProject(Project dteProject)
   at SonarLint.VisualStudio.Integration.ProjectToLanguageMapper.HasSupportedLanguage(Project project)
   at SonarLint.VisualStudio.Integration.ProjectSystemHelper.<GetSolutionProjects>d__15.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at SonarLint.VisualStudio.Integration.UnboundProjectFinder.<>c__DisplayClass10_0.<GetFilteredSolutionProjectsAsync>b__0()
```

**Log 2**
This same problem also affects the telemetry manager:

```
DEBUG: [CFamilyTelemetryManager] Failed to calculate cfamily project types: System.NullReferenceException: Object reference not set to an instance of an object.
   at SonarLint.VisualStudio.Integration.Binding.JsTsProjectTypeIndicator.HasJsTsFileInProject(ProjectItems projectItems)
   at SonarLint.VisualStudio.Integration.Binding.JsTsProjectTypeIndicator.IsJsTs(Project dteProject)
[...]

PR #3318, opened 2022-11-17
[DO NOT MERGE] Add logs to JsTsProjectTypeIndicator

rita-sonarsource:
Add debug logs for #3305


rita-sonarsource:
@duncanp-sonar please review. There is a try-catch so it shouldn't fail anymore. I added logs for as many things as I could, hopefully that will point us in the right direction.

I hope the user could test this and provide us with the logs. Then I will open a new PR with only the try-catch.

duncanp-sonar (approved):
LGTM. If that logging doesn't tell us where the error is, I don't know what will...

bigfluffycookie (approved):
Looks good, That will hopefully give enough information!

PR #3340, opened 2022-11-23
Lb/singleton

### LINKED ISSUE #3308: Add Singleton to ThreadHandling.cs
Currently, we new up the threadhandling class in the constructors. As the class does not have a state we could just add a singleton instead.

An issue is being raised because the class is also exported as a MEF singleton (so it can be imported by types in `Core`).
Once we've made a non-MEF singleton for the class we should be able to suppress the issue at a single place in the code using the `pragma` keyword.

bigfluffycookie:
Adds a singleton to ThreadHandling.cs to avoid it being newed everywhere.

Fixes #3308


duncanp-sonar (approved):
Looks good. Just a minor suggestion about re-enabling the rule.

PR #3345, opened 2022-11-24
Fix flaky behavior in SingleFileMonitor.cs

### LINKED ISSUE #3236: [Infra] Test RealFile_MultipleOperations_NoDuplicates is flaky
### Description

Fails frequently on the CI machine.

bigfluffycookie:
fixes #3236

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
@@ -189,7 +189,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
                 eventCount.Should().Be(1);
             }
         }
-    
+
         [Ignore] // Note: started failing when we started using Microsoft.VisualStudio.Sdk.TestFramework
                  // to mock VS threading/JoinableTaskContext to fix the options/dialog tests
         [TestMethod]
@@ -393,7 +393,6 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
 
         #endregion Simple file operations
 
-        [Ignore] // Flaky -> disabled https://github.com/SonarSource/sonarlint-visualstudio/issues/3236
         [TestMethod]
         public void RealFile_MultipleOperations_NoDuplicates()
         {
@@ -415,7 +414,9 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
                 {
                     File.Delete(filePathToMonitor);
                 }
-                testBody();
+                var numberOfOperations = testBody();
+                numberOfOperations.Should().BeInRange(0, 5);
+                totalEventCount += numberOfOperations;
             }
 
             // We might lose some events if they happen too close together, but we should never
@@ -425,10 +426,10 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
             TestContext.WriteLine($"Number of recorded events: {totalEventCount}");
             totalEventCount.Should().BeInRange(340, 500);
 
-            void testBody()
+            int testBody()
             {
                 const int pauseBetweenOpsInMs = 125;
-
+                var numberOfOperations = 0;
                 using (var singleFileMonitor = new SingleFileMonitor(filePathToMonitor, testLogger))
                 {
                     var testWrapper = new WaitableFileMonitor(singleFileMonitor);
@@ -455,8 +456,10 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
                     File.Delete(filePathToMonitor);
                     testWrapper.PauseForEvent(pauseBetweenOpsInMs);
 
-                    totalEventCount += testWrapper.EventCount;
+                    numberOfOperations += testWrapper.EventCount;
                 }
+
+                return numberOfOperations;
             };
         }
 
--- a/src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
@@ -414,6 +414,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
                 {
                     File.Delete(filePathToMonitor);
                 }
+
                 var numberOfOperations = testBody();
                 numberOfOperations.Should().BeInRange(0, 5);
                 totalEventCount += numberOfOperations;
--- a/src/Integration.Vsix/Analysis/SingleFileMonitor.cs
+++ b/src/Integration.Vsix/Analysis/SingleFileMonitor.cs
@@ -119,23 +119,23 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis
             }
         }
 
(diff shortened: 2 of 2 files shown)


src/Integration.Vsix.UnitTests/Analysis/SingleFileMonitorTests.cs
```diff
@@ -415,7 +414,9 @@ public void RealFile_MultipleOperations_NoDuplicates()
                 {
                     File.Delete(filePathToMonitor);
                 }
-                testBody();
+                var numberOfOperations = testBody();
+                numberOfOperations.Should().BeInRange(0, 5);
```
bigfluffycookie, line 418: As we are tasting for a range on line 427 it could still be possible for some of the iterations to report more than 5 operations and we would not catch it so I added another check here.

src/Integration.Vsix/Analysis/SingleFileMonitor.cs
```diff
@@ -119,6 +119,26 @@ private static void EnsureDirectoryExists(string filePath, IFileSystem fileSyste
             }
         }
 
+        private bool IsDuplicateChangeEvent(string fullPath)
```
bigfluffycookie, line 122: From my understanding the only event that has duplicate reports is the change event so I seperated this in a different method.

src/Integration.Vsix/Analysis/SingleFileMonitor.cs
```diff
@@ -119,6 +119,26 @@ private static void EnsureDirectoryExists(string filePath, IFileSystem fileSyste
...
+        {
+            // We're trying to ignore duplicate events by checking the last-write time.
+            // However, the precision of DateTime means that it is possible for separate events
+            // that happen very close together to report the same last-write time. If that happens,
+            // we'll be ignoring a "real" notification.
+            var currentTime = File.GetLastWriteTimeUtc(fullPath);
+            var timeDiff = currentTime - lastWriteTime;
+
+            // We are only ignoring events happening within 10 milliseconds of each other.
+            if (timeDiff.TotalMilliseconds >= 0 && timeDiff.TotalMilliseconds <= 10)
```
bigfluffycookie, line 132: The 10 millisecond is a bit of a guess, I could also make it a hundred, whichever makes more sense

src/Integration.Vsix/Analysis/SingleFileMonitor.cs
```diff
@@ -119,23 +119,23 @@ private static void EnsureDirectoryExists(string filePath, IFileSystem fileSyste
...
             // However, the precision of DateTime means that it is possible for separate events
             // that happen very close together to report the same last-write time. If that happens,
             // we'll be ignoring a "real" notification.
-            var currentTime = File.GetLastWriteTimeUtc(fullPath);
+            var currentTime = File.GetLastWriteTimeUtc(path);
             var timeDiff = currentTime - lastWriteTime;
 
-            // We are only ignoring events happening within 10 milliseconds of each other.
-            if (timeDiff.TotalMilliseconds >= 0 && timeDiff.TotalMilliseconds <= 10)
+            if (timeDiff.TotalMilliseconds >= 0 && timeDiff.TotalMilliseconds < 10)
```
bigfluffycookie, line 131: The ten milliseconds is a bit of a guess.

src/Integration.Vsix/Analysis/SingleFileMonitor.cs
```diff
@@ -119,23 +119,23 @@ private static void EnsureDirectoryExists(string filePath, IFileSystem fileSyste
             }
         }
 
-        private bool IsDuplicateChangeEvent(string fullPath)
+        private bool IsDuplicateChangeEvent(string path)
```
bigfluffycookie, line 122: From my understanding only the change event fires multiple times so I separated it into a different method

src/Integration.Vsix/Analysis/SingleFileMonitor.cs
```diff
@@ -131,18 +151,11 @@ private void OnFileChanged(object sender, System.IO.FileSystemEventArgs args)
             {
                 fileWatcher.EnableRaisingEvents = false;
 
-                // We're trying to ignore duplicate events by checking the last-write time.
-                // However, the precision of DateTime means that it is possible for separate events
-                // that happen very close together to report the same last-write time. If that happens,
-                // we'll be ignoring a "real" notification.
-                var currentTime = File.GetLastWriteTimeUtc(args.FullPath);
-                if (args.ChangeType != WatcherChangeTypes.Renamed && currentTime == lastWriteTime)
```
rita-sonarsource, line 139: Looking at `ChangeType`, it seems to be a Flags enum, which seems to suggest that the file system can aggregate multiple events.
https://learn.microsoft.com/en-us/dotnet/api/system.io.watcherchangetypes?view=net-6.0

The previous check was for `ChangeType != WatcherChangeTypes.Renamed`, whereas the new is for `ChangeType == WatcherChangeTypes.Changed`. I'm not sure either of them is the correct check, we should've been using HasFlag, although I'm not sure why we are ignoring duplicates based on change type. @duncanp-sonar do you remember what was the logic behind this?

src/Integration.Vsix/Analysis/SingleFileMonitor.cs
```diff
@@ -119,6 +119,26 @@ private static void EnsureDirectoryExists(string filePath, IFileSystem fileSyste
...
+        private bool IsDuplicateChangeEvent(string path)
+        {
+            // We're trying to ignore duplicate events by checking the last-write time.
+            // However, the precision of DateTime means that it is possible for separate events
+            // that happen very close together to report the same last-write time. If that happens,
+            // we'll be ignoring a "real" notification.
+            var currentTime = File.GetLastWriteTimeUtc(path);
+            var timeDiff = currentTime - lastWriteTime;
+
+            if (timeDiff.TotalMilliseconds >= 0 && timeDiff.TotalMilliseconds < 10)
```
rita-sonarsource, line 131: This is a bit arbitrary indeed. Does it matter if it's 10 milliseconds or 20 or 200? If the file system indeed aggregates the events, then it might take longer and it would still be ok.

In fact, I'm not 100% convinced this class should be implemented at all. It should be up to the caller to handle duplicate events and throttling. In fact, the comment at the top of the class says "we're removing duplicates but at the cost of potentially losing a few real events", which IMO is way worse. 

@duncanp-sonar I vaguely remember us having a discussion about this when we introduced `analysisConfigMonitor.ConfigChanged` for ESLint bridge. I'll try to find out if it was written somewhere.

rita-sonarsource: Edit: I believe we did have this discussion when implementing `SuppressedIssuesFileWatcher` and looks like we decided not to use this class. However I couldn't find anything written, I guess we talked about it F2F...
https://github.com/SonarSource/sonarlint-visualstudio/pull/2905

src/Integration.Vsix/Analysis/SingleFileMonitor.cs
duncanp-sonar: @rita-sonarsource no, I don't remember what the logic was, unfortunately. In theory, any kind of change (creation, deletion, rename, or change) could have an impact on the set of rules to be run.

bigfluffycookie:
Does not resolve like this.

Issue #3111, opened 2022-07-19
[UX] Make it clearer for the user which analyzer(s) has been triggered


rita-sonarsource:
### Description

With the introduction of secrets analyzer, files can be potentially analyzed multiple times, i.e., secrets + js/ts analyzer.
When it happens, two identical messages are displayed to the user that the analysis was executed. 
It's not clear from the text:
* Why the analysis is triggered multiple times (it looks like a redundant duplication)
* Why the issues count is potentially different (it looks like a mistake)

![image](https://user-images.githubusercontent.com/60586879/179705444-c907a1dd-c6eb-4310-9a6d-8fc987bf364a.png)

PR #3350, opened 2022-11-25
Include analyzer name in the output window

rita-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3111

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/AnalysisStatusNotifierTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/AnalysisStatusNotifierTests.cs
@@ -31,8 +31,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Analysis
     [TestClass]
     public class AnalysisStatusNotifierTests
     {
-        private const string AnalyzerName = "some analyzer";
-
         [TestMethod]
         [DataRow("foo-started.cpp", "foo-started.cpp")]
         [DataRow("c:\\test\\foo-started.cpp", "foo-started.cpp")]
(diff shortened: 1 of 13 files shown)


src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
```diff
@@ -67,11 +69,16 @@ public void AnalysisFailed(Exception ex)
...
 
             Notify(AnalysisStrings.Notifier_AnalysisFailed, false);
         }
 
+        private void Log(string messageFormat, params object[] args)
+        {
+            logger.WriteLine($"[{analyzerName}]: " + messageFormat, args);
+        }
+
         private void Notify(string messageFormat, bool showSpinner)
```
rita-sonarsource, line 82: StatusBar remains unchanged as it needs to be a short message. Only the output window is changed.

image

src/Integration.Vsix.UnitTests/Analysis/AnalysisStatusNotifierTests.cs
```diff
@@ -56,7 +58,8 @@ public void AnalysisStarted_LogToOutputWindow()
             testSubject.AnalysisStarted();
 
             var expectedMessage = string.Format(AnalysisStrings.MSG_AnalysisStarted, filePath);
-            logger.AssertOutputStringExists(expectedMessage);
+            logger.AssertPartialOutputStringExists(AnalyzerName);
```
duncanp-sonar, line 61: The link between the individual tests and the expected `analyzerName` is a bit indirect.
It would be better if the analyzer name was passed to `CreateTestSubject`.

duncanp-sonar (approved):
Nice. Just one comment about the slightly opaque test setup.

bigfluffycookie (approved):
Looks good :)

PR #3356, opened 2022-11-29
Implement IGitWorkspaceService 

### LINKED ISSUE #3353: Implement IGitWorkspaceService
Returns git folder for a given workspace

ugras-ergun-sonarsource:
fixes #3353


src/Infrastructure.VS/GitWorkSpaceService.cs
```diff
@@ -0,0 +1,86 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+using System.IO;
+using System.IO.Abstractions;
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.Infrastructure.VS
```
duncanp-sonar, line 26: Does this need to be in the `Infrastructure.VS` project?
It doesn't use or expose any types that are VS-specific.

src/Infrastructure.VS/Resources.resx
```diff
@@ -120,6 +120,9 @@
   <data name="FailedToCalculateVsVersion" xml:space="preserve">
     <value>Failed to calculate VS version, error: {0}</value>
   </data>
+  <data name="NoGitFolder" xml:space="preserve">
+    <value>Workspace is not under git control</value>
```
duncanp-sonar, line 124: `Solution` is a more normal term in VS than `Workspace` (does VS use the word `Workspace` anywhere? I can't find it on any of the menus).

How about _Solution/folder is not in a git repository_ ?

src/Infrastructure.VS/GitWorkSpaceService.cs
ugras-ergun-sonarsource: I couldn't think of a better place to put it. I don't think it would make sense to put it in the new assembly because this can be used for anything. And it relies on "IWorkspaceService" which is VS-specific.

src/Infrastructure.VS/GitWorkSpaceService.cs
```diff
@@ -0,0 +1,86 @@
...
+                return null;
+            }
+
+            var currentDir = new DirectoryInfo(workspaceRoot);
+
+            do
+            {
+                gitRoot = Path.Combine(currentDir.FullName, gitFolder);
+
+                if (fileSystem.Directory.Exists(gitRoot))
```
duncanp-sonar, line 73: Case-sensitivity: presumably (hopefully!) `Directory.Exists` is case-insensitive.

Unfortunately, we can't really test that our implementation is case-insensitive because we're using the file system abstraction in the tests, and we'd have to set up our mock to handle casing correctly, which wouldn't prove anything...

ugras-ergun-sonarsource: If it weren't the case (pun not intended :D) I think most of the .net applications would have trouble with the file system.

src/Infrastructure.VS/GitWorkSpaceService.cs
duncanp-sonar: Hhmm. The implementations of both `IWorkspaceService` and `IFolderWorkspaceService` are dependant on VS types/services, but the interfaces don't expose any VS types so they could be in `Core`.

I think we should put all of the interfaces into `Core` since they don't depend on VS (and it increases where they can be used).

I think I would also put the implementation of this new service in `Core` too, because it's implementation doesn't have any dependency on VS.

src/Infrastructure.VS/GitWorkSpaceService.cs
duncanp-sonar: True; but we don't have any tests that prove whether our code is handling casing correctly. We do mock the `Directory.Exists` method; I guess just adding an assertion that it is called would be enough.

duncanp-sonar (commented):
Looks good, although I think the new interface and implementation should be in `Core`.

duncanp-sonar (approved):
In retrospect, it might have been easier to move the types in a separate commit ;-)

PR #3361, opened 2022-12-01
Create new ConnectedMode assembly 

### LINKED ISSUE #3357: Create new ConnectedMode assembly

ugras-ergun-sonarsource:
fixes #3357


src/ConnectedMode.UnitTests/Usings.cs
```diff
@@ -0,0 +1,3 @@
+﻿global using Microsoft.VisualStudio.TestTools.UnitTesting;
```
ugras-ergun-sonarsource, line 1: I think it's a good idea to have these on a global usings file so we do not need to add on all classes. I did not add the copyright header because we have no "real code" here.

src/ConnectedMode/ConnectedMode.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!-- NOTE: this assembly should NOT have any dependencies on Visual Studio assemblies -->
```
duncanp-sonar, line 3: This only really applies to `Core` and `Roslyn.Suppressions`. It doesn't have to be the case for this assembly.

duncanp-sonar (approved):

PR #3364, opened 2022-12-01
Added Help submenu, Show Logs button

georgii-borovinskikh-sonarsource:
image


src/Integration.Vsix/Commands/PackageCommandId.cs
```diff
@@ -28,6 +28,7 @@ internal enum PackageCommandId
         ProjectTestPropertyAuto = 0x102,
         ProjectTestPropertyTrue = 0x103,
         ProjectTestPropertyFalse = 0x104,
+        SonarLintHelpShowLogs = 0x1025,
```
rita-sonarsource, line 31: Open question regarding the names of future commands: would we want to have the SonarLint prefix? Or would just their group suffice? i.e. HelpShowLogs, HelpViewDocumentation, HelpReportAnIssue, etc

src/Integration.Vsix/SonarLint.vsct
```diff
@@ -39,8 +42,23 @@
           <ButtonText>SonarLint</ButtonText>
         </Strings>
       </Menu>
+      <Menu guid="guidCmdSet" id="SonarLintHelpMenuItem" priority="0x9999">
```
rita-sonarsource, line 45: We should probably be consistent between your PR and mine, as well as future commands.
Perhaps it's better to drop "Item" and "SonarLint" from the name, as we are referring here to the menu. So maybe just `HelpMenu`.
WDYT?

Same for `SonarLintHelpMenuGroup` = `HelpMenuGroup` ?

src/Integration.Vsix/SonarLint.vsct
```diff
@@ -39,8 +42,23 @@
...
+        <Parent guid="guidCmdSet" id="SonarLintCommandSubMenu"/>
+        <Strings>
+          <ButtonText>Help</ButtonText>
+        </Strings>
+      </Menu>
     </Menus>
 
+    <Buttons>
+      <Button guid="guidCmdSet" id="SonarLintHelpShowLogsCommand">
```
rita-sonarsource, line 54: Let's discuss the name of commands too, I'd drop the `SonarLint<Category>` prefix and keep it just `ShowLogsCommand`.
(Also open question: Command, or CommandId? Menu vs MenuId?)

rita-sonarsource (approved):
@georgii-borovinskikh-sonarsource LGTM but let's decide on variable names convention. I'm open to suggestions.

src/Integration.Vsix/SonarLint.vsct
```diff
@@ -30,6 +30,9 @@
       <Group guid="guidCmdSet" id="SonarLintCommandSubMenu" priority="0x0100">
```
bigfluffycookie, line 30: Considering we will have a number of submenus eg. help, analysis and so on shouldn't we consider having a seperate vsct file for each? I have a feeling else this file will be super full and also it will be more difficult for us to simultaneously work on it.

Each of the vsct would need a reference to the guidCmdSet and the SonarLintCommandSubMenu so maybe those could be saved somewhere accessible instead of copy pasting the guids around. What do you think?

@rita-sonarsource and @duncanp-sonar

rita-sonarsource (approved):

bigfluffycookie (approved):
LGTM

PR #3368, opened 2022-12-02
Add view documentation button to help menu.

bigfluffycookie:


src/Integration.Vsix/Commands/PackageCommandId.cs
```diff
@@ -31,6 +31,7 @@ internal enum PackageCommandId
         // Menus
         ProjectSonarLintMenu = 0x850,
         // Help menu buttons
-        SonarLintHelpShowLogs = 0x102
+        SonarLintHelpShowLogs = 0x102,
+        ViewDocumentationCommand = 0x103
```
rita-sonarsource, line 35: I just noticed that we have a problem. This enum doesn't take into account the fact that it's the same ID but under a different set. 0x102 and 0x103 are already taken by the commands above, and I think we created a conflict here. 

I suggest that we place these IDs in the relevant command classes:
```
internal class ViewDocumentationCommand : VsCommandBase
{
 internal const int ViewDocumentationCommandId = 0x103;
}
```

WDYT?

rita-sonarsource (approved):
LGTM but see my comment regarding the IDs under the enum, I missed this previously

src/Integration.Vsix/Commands/PackageCommandId.cs
bigfluffycookie: but then how would it look in the packageCommandManager.cs? Like this?

```
var viewDocumentationCommand = new ViewDocumentationCommand();
this.RegisterCommand(CommonGuids.HelpMenuCommandSet, viewDocumentationCommand.ViewDocumentationCommandId, viewDocumentationCommand);
```

src/Integration.Vsix.UnitTests/Commands/PackageCommandManagerTests.cs
```diff
@@ -86,10 +87,11 @@ public void PackageCommandManager_Initialize()
             testSubject.Initialize(serviceProvider.GetMefService<ITeamExplorerController>(),
                 serviceProvider.GetMefService<IProjectPropertyManager>(),
                 Mock.Of<IProjectToLanguageMapper>(),
-                Mock.Of<IOutputWindowService>());
+                Mock.Of<IOutputWindowService>(),
+                Mock.Of<IShowInBrowserService>());
 
             // Assert
-            menuService.Commands.Should().HaveCount(allCommands.Count, "Unexpected number of commands");
+            menuService.Commands.Should().HaveCountGreaterOrEqualTo(allCommands.Count, "Unexpected number of commands");
```
georgii-borovinskikh-sonarsource, line 94: Same as in the other PR

src/Integration.Vsix/Commands/HelpMenu/ViewDocumentationCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    internal class ViewDocumentationCommand : VsCommandBase
+    {
+        internal const int ViewDocumentationCommandId = 0x103;
+
+        private readonly IShowInBrowserService showInBrowserService;
+
+        public ViewDocumentationCommand(IShowInBrowserService showInBrowserService)
+        {
+            this.showInBrowserService = showInBrowserService;
```
georgii-borovinskikh-sonarsource, line 33: NOTE: I will be making this class more generic, since the code of calling the method can be reused. There's going to be a conflict of Ids if we define them as fields of the class.

georgii-borovinskikh-sonarsource: Not really a conflict, but it will be a bit confusing. We need to rethink the way we define Ids

bigfluffycookie: I don't think this is an issue as in the vsct file you can still see all of the IDs for the commands in the command set

georgii-borovinskikh-sonarsource: It would be an issue in the CommandManager class, here:
`this.RegisterCommand(CommonGuids.HelpMenuCommandSet, ViewDocumentationCommand.ViewDocumentationCommandId, new ViewDocumentationCommand(showInBrowserService));`
If I make this class generic, I would need to specify which link to open (rn it only opens the docs link, but you could specify which one with a `Action<IShowInBrowserService>`). Then, having the Id be defined inside the class wouldn't really make sense, because the class is going to be instantiated as different commands with different Ids

bigfluffycookie: If you want to use a different link refer to my other PR where I use "IBrowserService" which has  navigate(string url)

bigfluffycookie: I am not sure if it is good practice to reuse a command class like this anyways although maybe @rita-sonarsource or @duncanp-sonar have more insight on this :)

rita-sonarsource: @georgii-borovinskikh-sonarsource looking at this class, I think adding an abstraction layer for one method might be an overkill and would introduce further complexity. Perhaps let's start with making a command that calls `IBrowserService.Navigate`, I feel that it already does very little. WDYT?

src/Integration.Vsix.UnitTests/Commands/PackageCommandManagerTests.cs
rita-sonarsource: Leyla and I discussed this offline, the problem is with the test expects only the commands in the enum to be registered, whereas now we have commands that are not part of the enum. Hence the `HaveCountGreaterOrEqualTo`. I think we should stop putting code in the PackageCommandManager but for now it's there so this fixes the test. WDYT?

src/Integration.Vsix/Commands/HelpMenu/ViewDocumentationCommand.cs
georgii-borovinskikh-sonarsource: I'm not sure I see how using the navigate thing directly would solve the problem of having boilerplate code. How would that look like?

rita-sonarsource: It's not really boilerplate code, it's calling one method, and it's already minimal and testable.
Are you proposing a base command that other commands can inherit from, i.e.
```
class GenericCommand
{
	ctor(string url, IBrowserService)
	{
	}

	onInvoke()
	{
		IBrowserService.NavigateToUrl(url);
	}
}

class ViewDocsCommand : GenericCommand
{
	ctor(IBrowserService) : base ("http..", browserService);
}

class ReportAProblemCommand : GenericCommand
{
	ctor(IBrowserService) : base ("http..", browserService);
}
```

In this case I don't think it saves much and it introduces complexity, plus the tests would be exactly the same.
Or are you suggesting something else?

rita-sonarsource: Edit: @georgii-borovinskikh-sonarsource I saw that you wrote above "If I make this class generic, I would need to specify which link to open (rn it only opens the docs link, but you could specify which one with a Action<IShowInBrowserService>)." --> In this case I fully disagree as I don't think the command manager should know about URLs.

georgii-borovinskikh-sonarsource: Why do you think it shouldn't know about that?

georgii-borovinskikh-sonarsource: Not about the specific links, but about the configuration of commands

rita-sonarsource: You can also do `new GenericCommand(() => showInBrowserService.ShowDocumentation())` which would be the same and "more" generic, but in both cases, I think it's an overkill and doesn't really save much.

PR #3362, opened 2022-12-01
Move taint and hotspots into SonarLint -> Connected Mode menu

rita-sonarsource:
image
Fixes [#3363](https://github.com/SonarSource/sonarlint-visualstudio/issues/3363)

PR #3373, opened 2022-12-02
Add Connected Mode menu

rita-sonarsource:
Fixes [#3372](https://github.com/SonarSource/sonarlint-visualstudio/issues/3372)

### DIFF
--- a/src/Integration.Vsix/SonarLint.vsct
+++ b/src/Integration.Vsix/SonarLint.vsct
@@ -19,6 +19,7 @@
   <!-- Include all package commands -->
   <Include href="Commands\PackageCommands.vsct"/>
   <Include href="Commands\HelpMenu\HelpMenu.vsct"/>
+  <Include href="Commands\ConnectedModeMenu\ConnectedModeMenu.vsct"/>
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/SonarLint.vsct
```diff
@@ -46,17 +47,13 @@
           <ButtonText>Help</ButtonText>
         </Strings>
       </Menu>
-    </Menus>
-
-    <!-- Example of how to add a command to the sonarlint sub menu
```
rita-sonarsource, line 51: I removed this commented sample because we're already using it, so we have the real buttons as examples.

georgii-borovinskikh-sonarsource:
You only added the menu in this PR. Will other lower level buttons be in another PR?

rita-sonarsource:
> You only added the menu in this PR. Will other lower level buttons be in another PR?

Yep, https://github.com/SonarSource/sonarlint-visualstudio/pull/3362 @georgii-borovinskikh-sonarsource

bigfluffycookie (approved):
Looks good!

Issue #3398, opened 2022-12-06
Make suppressions provider branch-aware


duncanp-sonar:
### Description

Fetching suppressions should use the correct server branch

PR #3399, opened 2022-12-06
Add CreateTestSubject method to SonarQubeIssuesProviderTests

duncanp-sonar:
- no product changes

Prep for #3398

### DIFF
--- a/src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
+++ b/src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
@@ -141,8 +141,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
             SetupSolutionBinding(isConnected: true, issues: new List<SonarQubeIssue>());
 
             // Act
-            using (var issuesProvider = new SonarQubeIssuesProvider(mockSqService.Object, "keyXXX",
-                mockTimerFactory.Object, testLogger, threadHandling.Object))
+            using (var issuesProvider = CreateTestSubject("keyXXX"))
             {
                 WaitForInitialFetchTaskToStart();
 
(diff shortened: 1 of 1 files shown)


src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -141,8 +141,7 @@ public void Constructor_Connected_StartsMonitoringAndSyncs()
             SetupSolutionBinding(isConnected: true, issues: new List<SonarQubeIssue>());
 
             // Act
-            using (var issuesProvider = new SonarQubeIssuesProvider(mockSqService.Object, "keyXXX",
```
duncanp-sonar, line 144: This is a test-only change to reduce the noise in later commits.

I've modified the existing tests tests to call a `CreateTestSubject` method instead of creating the class directly.

That way, when we add a new parameter to the `SonarQubeIssuesProvider` constructor we'll only have to change the helper method.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -160,8 +159,7 @@ public void Dispose_Disconnected_TimerDisposed()
             // Arrange
             SetupSolutionBinding(isConnected: true, issues: null);
 
-            var issuesProvider = new SonarQubeIssuesProvider(mockSqService.Object, "sqkey",
-                mockTimerFactory.Object, testLogger, threadHandling.Object);
+            var issuesProvider = CreateTestSubject();
```
duncanp-sonar, line 162: This test class still uses class-level mock variables initialized in `TestInitialize`. I haven't attempted to fix that.

The only parameter that actually needs to vary in _some_ tests is the project key that is passed in.
If a test doesn't care what the project key is (like this one), then it doesn't pass one.

ugras-ergun-sonarsource (approved):

PR #3400, opened 2022-12-06
Simplify SonarQubeIssuesProvider public constructor

duncanp-sonar:


src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -83,22 +83,19 @@ public void Constructor_ThrowsOnNull()
         {
             string validProjectKey = "key1";
 
-            Action op = () => new SonarQubeIssuesProvider(null, validProjectKey, mockTimerFactory.Object, testLogger, threadHandling.Object);
+            Action op = () => new SonarQubeIssuesProvider(null, validProjectKey, testLogger);
```
duncanp-sonar, line 86: We only need to test the public constructor arguments.
The timer factory and thread handling objects are only on the internal constructor, so they don't need to be tested here.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -55,36 +55,25 @@ public sealed class SonarQubeIssuesProvider : ISonarQubeIssuesProvider
...
+        public SonarQubeIssuesProvider(ISonarQubeService sonarQubeService, string sonarQubeProjectKey,
+            ILogger logger) : this(sonarQubeService, sonarQubeProjectKey, logger, new TimerFactory(), new ThreadHandling())
         {
-
         }
 
-
-        internal SonarQubeIssuesProvider(ISonarQubeService sonarQubeService, string sonarQubeProjectKey, ITimerFactory timerFactory,
-            ILogger logger, IThreadHandling threadHandling)
+        internal /* for testing */ SonarQubeIssuesProvider(ISonarQubeService sonarQubeService, string sonarQubeProjectKey,
```
duncanp-sonar, line 63: Re-ordered the arguments to follow the convention of having the test-only parameters last (i.e. timer factory and thread handling)

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -55,36 +55,25 @@ public sealed class SonarQubeIssuesProvider : ISonarQubeIssuesProvider
...
-                throw new ArgumentNullException(nameof(timerFactory));
-            }
-            if (logger == null)
-            {
-                throw new ArgumentNullException(nameof(logger));
-            }
 
-            this.sonarQubeService = sonarQubeService;
+            Debug.Assert(timerFactory != null, "timeFactory should not be null");
+            Debug.Assert(threadHandling != null, "threadHandling should not be null");
```
duncanp-sonar, line 72: Product code should only ever be calling the `public` constructor, so it will never pass these arguments.

`timerFactory` and `threadHandling` are test-only so asserting that they are not null is enough.

ugras-ergun-sonarsource (approved):

PR #3401, opened 2022-12-07
Place IssueViz command under SonarLint menu

### LINKED ISSUE #3394: Move Show Issue Visualization command
Move from Other Windows to SonarLint

rita-sonarsource:
Fixes  #3394

image

### DIFF
--- a/src/IssueViz/Commands/IssueVizPackage.vsct
+++ b/src/IssueViz/Commands/IssueVizPackage.vsct
@@ -3,6 +3,7 @@
 
   <Extern href="stdidcmd.h"/>
   <Extern href="vsshlids.h"/>
+  <Extern href="KnownImageIds.vsct"/>
 
   <Commands package="guidIssueVizPackage">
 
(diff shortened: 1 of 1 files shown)


src/IssueViz/Commands/IssueVizPackage.vsct
```diff
@@ -15,10 +16,11 @@
     <Buttons>
 
       <Button guid="guidIssueVizPackageCmdSet" id="IssueVisualizationToolWindowCommandId" priority="0x0100" type="Button">
-        <Parent guid="guidSHLMainMenu" id="IDG_VS_WNDO_OTRWNDWS1" />
-        <Icon guid="guidImages" id="sonarlint" />
+        <Parent guid="SonarLintCmdSet" id="SonarLintGroup"/>
+        <Icon guid="ImageCatalogGuid" id="TreeView" /> <!-- known VS icon ID -->
```
rita-sonarsource, line 20: The original command used the sonarlint icon cause it was under the Other Windows menu. Now there is no need for it so I opted for a default VS icon that looks a bit like the UI in that screen.

Btw, a very useful VS extension that allows you to see all VS icons: https://www.vsixgallery.com/extension/c38f3512-4653-4d97-a4c5-860a425209f6

src/IssueViz/Commands/IssueVizPackage.vsct
```diff
@@ -117,5 +119,9 @@
       <IDSymbol name="grpDaemonErrorList" value="0x100" />
     </GuidSymbol>
 
+    <GuidSymbol name="SonarLintCmdSet" value="{DB0701CC-1E44-41F7-97D6-29B160A70BCB}">
```
rita-sonarsource, line 122: I'm not a big fan of it but I don't have another solution right now. We might want to revisit it when we have more windows/commands under the root SonarLint menu. Right now I cannot set the priority of this, it only looks ok because the ConnectedMode and Help both have a priority, so this one is in the middle by default.

georgii-borovinskikh-sonarsource (approved):

bigfluffycookie (approved):

PR #3407, opened 2022-12-07
Make the taint "opein in browser" context menu specify if it's sonar qube or cloud

georgii-borovinskikh-sonarsource:


src/IssueViz.Security/Taint/TaintList/TaintIssuesControl.xaml
```diff
@@ -19,7 +19,8 @@
             <core:DateTimeConverter x:Key="DateTimeConverter"/>
 
             <ContextMenu x:Key="RowMenu" DataContext="{Binding PlacementTarget.DataContext, RelativeSource={RelativeSource Self}}">
-                <MenuItem Header="Open in browser" 
+                <MenuItem Header="{Binding DataContext.ServerType, Mode=OneWay, RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=DataGrid}}"
+                          HeaderStringFormat="VIew in {0}"
```
rita-sonarsource, line 23: Typo, I is capitalized

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.Caption.cs
```diff
@@ -120,6 +121,46 @@ public void ActiveDocChanged_ExpectedCaptionSet()
             testSubject.WindowCaption.Should().Be(DefaultCaption);
         }
 
+        [TestMethod]
```
rita-sonarsource, line 124: These tests are separate from TaintIssuesControlViewModelTests because they test a lot more things: they test that the filtering and the caption calculation are working properly (you should see issues only for the selected document). This is not a functionality that is required from the server type, so these tests are unnecessarily complicated.  

I think it's better to put simple tests in TaintIssuesControlViewModelTests, one for ctor and one for document changed (and it doesn't even matter what the document is and if there are any issues, since the ServerType property is standalone).

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -164,9 +173,17 @@ private void ActiveDocumentTracker_OnDocumentFocused(object sender, ActiveDocume
         {
             activeDocumentFilePath = e.ActiveTextDocument?.FilePath;
             ApplyViewFilter(ActiveDocumentFilter);
+            UpdateServerType();
```
rita-sonarsource, line 176: @duncanp-sonar we're re-calculating server type in two places: ctor, and whenever an active document changes. This is under the assumption that if you change a solution, your active document will change. WDYT? 
Perhaps it's better to import `IActiveSolutionBoundTracker` and register to `SolutionBindingChanged`. WDYT?

cc @georgii-borovinskikh-sonarsource

rita-sonarsource (changes requested):
@georgii-borovinskikh-sonarsource see my comment about simplifying the tests, and I think they show that we shouldn't register to activeDocChanged but rather to `IActiveSolutionBoundTracker.SolutionBindingChanged`. Let's see what Duncan says too.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
rita-sonarsource: @georgii-borovinskikh-sonarsource so let's go with the cache 👍

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -121,6 +126,8 @@ public ITaintIssueViewModel SelectedIssue
 
         public AnalysisInformation AnalysisInformation { get; private set; }
 
+        public string ServerType => serverType.ToString();
```
georgii-borovinskikh-sonarsource, line 129: I think it's not going to be called often, so it's fine not to cache the string

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -191,6 +192,28 @@ public void Ctor_RegisterToActiveDocumentChanges()
             activeDocumentTracker.VerifyNoOtherCalls();
         }
 
+        [TestMethod]
+        public void Ctor_ExpectedServerTypeSet()
```
rita-sonarsource, line 196: It would be better to split these into 3 different tests.

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -191,6 +192,28 @@ public void Ctor_RegisterToActiveDocumentChanges()
...
+        {
+            var sonarQubeServiceMock = new Mock<ISonarQubeService>();
+
+            // 1. SC
+            sonarQubeServiceMock.Setup(x => x.GetServerInfo()).Returns(new ServerInfo(null, ServerType.SonarCloud));
```
rita-sonarsource, line 201: You can extract this to a helper method, `CreateSonarQubeService(serverType)`

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -701,6 +724,44 @@ public void AnalysisInformation_IssuesChanged_RaisesPropertyChanged()
...
+            var testSubject = CreateTestSubject(activeDocumentTracker: activeDocumentTrackerMock.Object, sonarQubeService:sonarQubeServiceMock.Object);
+            int eventCount = 0;
+            testSubject.PropertyChanged += (sender, args) => {
+                if (args is { PropertyName: nameof(testSubject.ServerType) })
+                {
+                    eventCount++;
+                }
+            };
+
+            // 1. SC
```
rita-sonarsource, line 743: 1. always use different tests
2. I think it's better here to assert (null -> non null), (null -> null), (not null -> not null, same value), (not null -> not null, different value), and (not null -> null)

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -164,9 +174,20 @@ private void ActiveDocumentTracker_OnDocumentFocused(object sender, ActiveDocume
...
             activeDocumentFilePath = e.ActiveTextDocument?.FilePath;
             ApplyViewFilter(ActiveDocumentFilter);
+            UpdateServerType();
             UpdateCaption();
         }
 
+        private void UpdateServerType()
+        {
+            var newServerType = sonarQubeService.GetServerInfo()?.ServerType;
+            if (!newServerType.Equals(serverType))
```
rita-sonarsource, line 184: cosmetics: empty row before if

rita-sonarsource (changes requested):
@georgii-borovinskikh-sonarsource please refactor the tests

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -701,6 +716,34 @@ public void AnalysisInformation_IssuesChanged_RaisesPropertyChanged()
...
+        [DataRow(ServerType.SonarCloud, null, "", true)]
+        [DataRow(ServerType.SonarCloud, ServerType.SonarCloud, nameof(ServerType.SonarCloud), false)]
+        [DataRow(ServerType.SonarCloud, ServerType.SonarQube, nameof(ServerType.SonarQube), true)]
+        public void ActiveDocChanged_ExpectedServerTypeSetOnlyWhenItIsChanged(ServerType? originalServerType,
+            ServerType? newServerType, string expectedValue, bool expectedRaiseEvent)
+        {
+            var activeDocumentTrackerMock = new Mock<IActiveDocumentTracker>();
+            var sonarQubeServiceMock = new Mock<ISonarQubeService>();
+            SetServerType(originalServerType, sonarQubeServiceMock);
+            var testSubject = CreateTestSubject(activeDocumentTracker: activeDocumentTrackerMock.Object, sonarQubeService:sonarQubeServiceMock.Object);
```
rita-sonarsource, line 731: Cosmetics: new line before testSubject, to separate between the arrange and the act

georgii-borovinskikh-sonarsource: the test subject creation here is in the arrange

rita-sonarsource (approved):
LGTM

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
rita-sonarsource: It's messy to read imo. TestSubject is always declared separately to ease on the reader. It's one line, change it :D

georgii-borovinskikh-sonarsource: Yeah but I already have RaiseActiveDocumentChangedEvent(activeDocumentTrackerMock); separated by an empty line, adding another one would be more confusing, because it would make 4 paragraphs. The creation of test subject is not really important here to warrant an extra line, IMO. Alternatively, this is what I used to do before, you could separate the AAA by 2 empty lines and then add single empty lines in the As if the visibility becomes a problem

Issue #3406, opened 2022-12-07
Add skeleton ServerBranchProvider implementation


duncanp-sonar:
### Description

Implement first version of `IServerBranchProvider`

* define interface in `Core`
* simple no-op implementation in `ConnectedMode`
* minimal tests

PR #3408, opened 2022-12-07
Add skeleton IServerBranchProvider

duncanp-sonar:
Relates to #3406


src/ConnectedMode.UnitTests/ServerBranchProviderTests.cs
```diff
@@ -0,0 +1,52 @@
...
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.Integration.UnitTests;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests
+{
+    [TestClass]
+    public class ServerBranchProviderTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
```
duncanp-sonar, line 31: Minimal set of initial tests

src/ConnectedMode/ServerBranchProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+        public ServerBranchProvider(ILogger logger)
+        {
+            this.logger = logger;
+        }
+
+        public string GetServerBranchName()
+        {
+            string branchName = null;
+
+            // TODO: calculate branch name
```
duncanp-sonar, line 42: Minimal implementation - just returns `null` for now.

PR #3419, opened 2022-12-08
Match Git branch to SQ/SC branch 

### LINKED ISSUE #3365: Match Git branch to SQ/SC branch

ugras-ergun-sonarsource:
fixes #3365


src/ConnectedMode.UnitTests/BranchMatcherTests.cs
```diff
@@ -0,0 +1,213 @@
...
+            var repo = CreateRepo(headBranch, masterBranch, devBranch);
+
+            var testSubject = new BranchMatcher(service, repo);
+
+            var result = await testSubject.GetMatchedBranch("projectKey");
+
+            result.Should().Be("dev");
+        }
+
+        /*
```
ugras-ergun-sonarsource, line 77: Added tree only for this test because other cases are very easy to understand by just looking at the code.

src/ConnectedMode.UnitTests/LibGit2SharpWrappers/CommitWrapper.cs
```diff
@@ -0,0 +1,37 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using LibGit2Sharp;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.LibGit2SharpWrappers
+{
+    internal class CommitWrapper : Commit
+    {
+        internal CommitWrapper(byte commitID)
```
ugras-ergun-sonarsource, line 27: I think 1 byte is enough to differentiate between commits. I do not think we'll have a test case with more than 256 commits

duncanp-sonar:
@ugras-ergun-sonarsource are the correct files added to the VSIXes?
Do we need different native assemblies in the VSIXes for VS2019 and VS2022?

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using LibGit2Sharp;
+using SonarQube.Client;
+
+namespace SonarLint.VisualStudio.ConnectedMode
+{
+    public interface IBranchMatcher
+    {
+        Task<string> GetMatchedBranch(string projectKey);
```
duncanp-sonar, line 32: Needs a doc comment

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    internal class BranchMatcher : IBranchMatcher
```
duncanp-sonar, line 35: Are we using the same matching algorithm as the other SonarLint flavours?

Either way, we should document it. If we are using a different algorithm we should record why.

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -0,0 +1,105 @@
...
+        public async Task<string> GetMatchedBranch(string projectKey)
+        {
+            var head = gitRepo.Head;
+
+            if (head == null)
+            {
+                return null;
+            }
+
+            var remoteBranches = await sonarQubeService.GetProjectBranchesAsync(projectKey, CancellationToken.None);
```
duncanp-sonar, line 55: What happens if the service is not connected? (it throws?)

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
```diff
@@ -0,0 +1,211 @@
...
+using SonarQube.Client;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests
+{
+    [TestClass]
+    public class BranchMatcherTests
+    {
+        [TestMethod]
+        public async Task GetMatchedBranch_ChooseBranchWithSameName()
```
duncanp-sonar, line 36: Could be data-driven to test case-insensitivity

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            var service = CreateService("master", "branch");
+
+            var masterCommit = new CommitWrapper(1);
+            var branchCommit = new CommitWrapper(2);
+
+            var masterBranch = CreateBranch("master", masterCommit);
+            var headBranch = CreateBranch("branch", branchCommit, masterCommit);
```
duncanp-sonar, line 44: The commits are completely irrelevant in this test, aren't they? What does including them add to the test?

src/ConnectedMode.UnitTests/LibGit2SharpWrappers/BranchCollectionWrapper.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using LibGit2Sharp;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.LibGit2SharpWrappers
+{
+    internal class BranchCollectionWrapper : BranchCollection
+    {
+        private readonly IEnumerable<Branch> branches;
+
+        internal BranchCollectionWrapper(IEnumerable<Branch> branches) : base()
```
duncanp-sonar, line 30: Minor: `base()` isn't required.

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
```diff
@@ -0,0 +1,211 @@
...
+         * |    |       |
+         * |    |       |->Commit5
+         * |    |       |
+         * |    |
+         * |    |---->Branch2 (closest branch)
+         * |    |       |
+         * |    |       |->Commit6
+         * |    |       |
+         * |    |
+         * |    |---->Branch3 (Head)
```
duncanp-sonar, line 97: Could you make the test slightly more robust by including a closest commit match that does not have a matching branch name?
i.e. a local branch that is closest, but that does not match a branch on the server -> should be ignored.

Does that make sense as an additional test case?

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            var result = await testSubject.GetMatchedBranch("projectKey");
+
+            result.Should().Be("branch2");
+        }
+
+        [TestMethod]
+        public async Task GetMatchedBranch_NoMatchingBranch_ChooseMain()
+        {
+            var service = CreateService("master", "branch1", "branch2");
```
duncanp-sonar, line 142: Naming: `service` is very general. `sonarService` / `CreateSonarService` would give a bit more context.

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            var result = await testSubject.GetMatchedBranch("projectKey");
+
+            result.Should().Be("branch");
+        }
+
+        [TestMethod]
+        public async Task GetMatchedBranch_NoBranchWithSameName_ChooseClosestMatch()
+        {
+            var service = CreateService("master", "dev");
```
duncanp-sonar, line 58: Do any of the test cases have data sets where there are server-side branch names that don't have corresponding local branches?

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
```diff
@@ -0,0 +1,211 @@
...
+            var allBranches = new BranchCollectionWrapper(branchList);
+            
+            var repo = new Mock<IRepository>();
+
+            repo.SetupGet(r => r.Head).Returns(headBranch);
+            repo.SetupGet(r => r.Branches).Returns(allBranches);
+            return repo.Object;
+        }
+
+        //Order of the commits matter. Commits should be in descending order.
```
duncanp-sonar, line 188: "in descending order" = "newest first"?

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            repo.SetupGet(r => r.Head).Returns(headBranch);
+            repo.SetupGet(r => r.Branches).Returns(allBranches);
+            return repo.Object;
+        }
+
+        //Order of the commits matter. Commits should be in descending order. 
+        private static BranchWrapper CreateBranch(string branchName, params CommitWrapper[] commits)
+        {
+            var masterBranch = new BranchWrapper(branchName, new CommitLogWrapper(commits));
```
duncanp-sonar, line 191: This is just `branch`, isn't it? i.e. there's nothing that makes it a `master` branch?

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
duncanp-sonar, line 142: Could you make the name of the main branch something other than `main` or `master` please?

It makes it clearer that there is nothing magic about the actual name of the branch.

src/ConnectedMode/BranchMatcher.cs
ugras-ergun-sonarsource: Yes, it does. We expect it to be in connected mode and already has a connection to the server.

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
ugras-ergun-sonarsource: GetMatchedBranch_NoMatchingBranch_ChooseMain

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
ugras-ergun-sonarsource: yes

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
ugras-ergun-sonarsource: yes, we do not care what's master/main on git, we only care about on SQ/SC

src/ConnectedMode.UnitTests/LibGit2SharpWrappers/BranchCollectionWrapper.cs
ugras-ergun-sonarsource: leftover from initial try to use collection from base.

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
ugras-ergun-sonarsource: this was a leftover from when I was trying to create a basic scenario.

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
ugras-ergun-sonarsource: Then it would match itself. But it makes sense to create such a test to be explicit.  (maybe not in this complicated format)

duncanp-sonar: Why would it match itself? It's a local branch that doesn't have a server match so it should be ignored, regardless of how close it is.

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
duncanp-sonar: So calling the variable `masterBranch` is misleading. Could you change it please?

src/ConnectedMode/BranchMatcher.cs
duncanp-sonar: The doc comment is useful. I actually meant that we should explicitly state that we are using the same algorithm as the other flavours, since it isn't clear. Also, my understanding from our offline discussion was that we know that our implementation is different in some respects. It would be useful to record that while we still remember what the difference is.

Maybe add something like the following (assuming it's accurate):

> This is the same algorithm as the other SonarLint flavours, except in how we treated branch histories.
We treat branch histories as series of linear commits ordered by time, even if there are merges.

src/ConnectedMode/BranchMatcher.cs
duncanp-sonar: An assertion to document it would be useful. It's currently not obvious that that is the intended behaviour.

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// </remarks>
+        Task<string> GetMatchedBranch(string projectKey);
+    }
+
+    internal class BranchMatcher : IBranchMatcher
+    {
+        private readonly ISonarQubeService sonarQubeService;
+        private readonly IRepository gitRepo;
+
+        public BranchMatcher(ISonarQubeService sonarQubeService, IRepository gitRepo)
```
duncanp-sonar, line 48: Passing the repo in the constructor means we will need to create a new instance of this class every time the solution changes i.e. we will need a factory to create new instances of this class a runtime.

Alternatively, we could pass the repo in to the `GetMatchedBranch` method.

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                return remoteBranches.First(rb => rb.IsMain).Name;
+            }
+            return closestBranch;
+        }
+
+        //Commits are in descending order by time
+        private int GetDistance(Branch head, Branch branch)
+        {
+            var headCommits = head.Commits.ToList();
```
duncanp-sonar, line 104: The list of head commits will be the same every time, so we could fetch it once and pass it in, rather than allocating a list for every branch.

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -0,0 +1,119 @@
...
+            for (int i = 0; i < headCommits.Count; i++)
+            {
+                var commitID = headCommits[i].Id;
+
+                var branchCommitIndex = branch.Commits.ToList().FindIndex(bc => bc.Id == commitID);
```
duncanp-sonar, line 109: `ToList` means we're going to enumerate the commit history for the branch, even if the commit we want is first in the list. We're going to repeat that for every branch.
I'll add a TODO to look at optimising this.

PR #3420, opened 2022-12-08
Add groups to help menu for dividers.

bigfluffycookie:
image


src/Integration.Vsix/Commands/HelpMenu/HelpMenu.vsct
```diff
@@ -5,34 +5,40 @@
 
   <Commands package="guidVsPackagePkgString">
     <Groups>
-      <Group guid="HelpMenuCmdSet" id="HelpGroup">
+      <Group guid="HelpMenuCmdSet" id="HelpGroupMain">
```
rita-sonarsource, line 8: I think our convention is `blahGroup`, so it would be better to rename these to HelpMainGroup, HelpLogsGroup, HelpAboutGroup

rita-sonarsource (approved):
LGTM just the comment about the variable names

PR #3427, opened 2022-12-08
Change label of Report a Problem menu item

rita-sonarsource:
image


bigfluffycookie (approved):
Its good but I feel like that button is a bit long now, might just be my opinion though

rita-sonarsource:
> Its good but I feel like that button is a bit long now, might just be my opinion though

I agree @bigfluffycookie. I don't know how to fix it though. We can do what resharper do and say "Report a Bug" which would make it slightly shorter. WDYT? Or perhaps "Send Feedback".

rita-sonarsource:
@bigfluffycookie I tried changing the label and it didn't make any difference in the button's width 🤔 I'm not sure why.
image

PR #3429, opened 2022-12-08
Make SuppressedIssuesProvider branch-aware

### LINKED ISSUE #3398: Make suppressions provider branch-aware

duncanp-sonar:
Fixes #3398

### DIFF
--- a/src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
+++ b/src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
@@ -81,21 +81,27 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
         [TestMethod]
         public void Constructor_ThrowsOnNull()
         {
-            string validProjectKey = "key1";
+            var sqService = Mock.Of<ISonarQubeService>();
+            var serverBranchProvider = Mock.Of<IServerBranchProvider>();
+            var logger = Mock.Of<ILogger>();
+            const string validProjectKey = "key1";
 
-            Action op = () => new SonarQubeIssuesProvider(null, validProjectKey, testLogger);
+            Action op = () => new SonarQubeIssuesProvider(null, serverBranchProvider, validProjectKey, logger);
             op.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("sonarQubeService");
 
-            op = () => new SonarQubeIssuesProvider(mockSqService.Object, null, testLogger);
+            op = () => new SonarQubeIssuesProvider(sqService, null, validProjectKey, logger);
+            op.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("serverBranchProvider");
+
+            op = () => new SonarQubeIssuesProvider(sqService, serverBranchProvider, null, logger);
             op.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("sonarQubeProjectKey");
 
-            op = () => new SonarQubeIssuesProvider(mockSqService.Object, "", testLogger);
+            op = () => new SonarQubeIssuesProvider(sqService, serverBranchProvider, "", logger);
             op.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("sonarQubeProjectKey");
 
-            op = () => new SonarQubeIssuesProvider(mockSqService.Object, "\r\n ", testLogger);
+            op = () => new SonarQubeIssuesProvider(sqService, serverBranchProvider, "\r\n ", logger);
             op.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("sonarQubeProjectKey");
 
-            op = () => new SonarQubeIssuesProvider(mockSqService.Object, validProjectKey, null);
+            op = () => new SonarQubeIssuesProvider(sqService, serverBranchProvider, validProjectKey, null);
             op.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("logger");
         }
 
@@ -105,7 +111,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
             // Arrange
             // The provider will attempt to fetch issues at start up and will loop if it
             // is not connected to the server, so we'll initialise it as connected.
-            SetupSolutionBinding(isConnected: true, issues: null);
+            SetupSolutionBinding(isConnected: true);
             mockTimer.SetupSet(t => t.AutoReset = true).Verifiable();
 
             // 1. Construction -> timer initialised
@@ -153,7 +159,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
         public void Dispose_Disconnected_TimerDisposed()
         {
             // Arrange
-            SetupSolutionBinding(isConnected: true, issues: null);
+            SetupSolutionBinding(isConnected: true);
 
             var issuesProvider = CreateTestSubject();
             WaitForInitialFetchTaskToStart();
@@ -171,7 +177,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
         public void Dispose_Connected_TimerDisposed()
         {
             // Arrange
-            SetupSolutionBinding(isConnected: true, issues: new List<SonarQubeIssue>());
+            SetupSolutionBinding(isConnected: true);
 
             var issuesProvider = CreateTestSubject();
 
@@ -189,7 +195,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
         [TestMethod]
         public void GetIssues_NoIssuesOnServer_ReturnsEmptyList()
         {
-            SetupSolutionBinding(isConnected: true, issues: null);
+            SetupSolutionBinding(isConnected: true);
 
             // 1. Created -> issues fetch in background
             var issuesProvider = CreateTestSubject("sqkey");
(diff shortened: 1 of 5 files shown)


src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -105,7 +111,7 @@ public void Constructor_TimerIsInitialized()
             // Arrange
             // The provider will attempt to fetch issues at start up and will loop if it
             // is not connected to the server, so we'll initialise it as connected.
-            SetupSolutionBinding(isConnected: true, issues: null);
```
duncanp-sonar, line 108: Miscellaneous test fix: passing `issues: null` causes the product code to throw an exception, which is caught and handled internally.

However, it isn't what the tests are expecting i.e. the tests are passing but perhaps not for the expected reason. I've modified the `SetupSolutionBinding` method to use an empty list of issues instead.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -198,7 +204,7 @@ public void GetIssues_NoIssuesOnServer_ReturnsEmptyList()
             VerifyServiceGetIssues(Times.Exactly(1), "sqkey");
 
             // 2. SonarQube project key doesn't match -> no issues
-            var matches = issuesProvider.GetSuppressedIssues("any project", "any file");
+            var matches = issuesProvider.GetSuppressedIssues("any project", "c:\\anyfile");
```
duncanp-sonar, line 207: Miscellaneous: changing the setup not to be `null` executed a different code path, which caused an assertion in the product code. Passing an absolute path avoids the assertion.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -207,6 +213,20 @@ public void GetIssues_NoIssuesOnServer_ReturnsEmptyList()
             VerifyServiceGetIssues(Times.Exactly(1));
         }
 
+        [TestMethod]
+        public void GetIssues_UsesExpectedBranch()
```
duncanp-sonar, line 217: Single new test: make sure the product is asking for and using the server branch.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -759,12 +790,17 @@ private void VerifyRanOnBackgroundThread(Times expected)
 
         private void VerifyServiceGetIssues(Times expected)
         {
-            mockSqService.Verify(x => x.GetSuppressedIssuesAsync(It.IsAny<string>(), null, It.IsAny<CancellationToken>()), expected);
+            mockSqService.Verify(x => x.GetSuppressedIssuesAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()), expected);
```
duncanp-sonar, line 793: Miscellaneous: this helper method is only interested in the number of times the service was called, not the arguments passed to it, so it should use `It.IsAny<string>()` rather than `null`.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -759,12 +790,17 @@ private void VerifyRanOnBackgroundThread(Times expected)
...
         }
 
         private void VerifyServiceGetIssues(Times expected, string sonarQubeProjectKey)
         {
-            mockSqService.Verify(x => x.GetSuppressedIssuesAsync(sonarQubeProjectKey, null, It.IsAny<CancellationToken>()), expected);
+            mockSqService.Verify(x => x.GetSuppressedIssuesAsync(sonarQubeProjectKey, It.IsAny<string>(), It.IsAny<CancellationToken>()), expected);
```
duncanp-sonar, line 798: Miscellaneous: same again - this method is only checking the project key, and doesn't care about the `branch` parameter.

src/Integration.UnitTests/Suppression/SonarQubeIssuesProviderTests.cs
```diff
@@ -759,12 +790,17 @@ private void VerifyRanOnBackgroundThread(Times expected)
...
+        }
+
+        private void VerifyServiceGetIssues(Times expected, string sonarQubeProjectKey, string branch)
```
duncanp-sonar, line 801: New overload for when we want to check both project key and the branch.

I thought about updating the existing tests to so they checked both the project key and branch, but it would have been quite a bit of work for no real added value.

src/Integration.UnitTests/Suppression/SuppressedIssuesProviderTests.cs
```diff
@@ -49,6 +53,45 @@ public void TestInitialize()
             testSubject = new SuppressedIssuesProvider(activeSolutionBoundTracker, createProviderFunc.Object);
         }
 
+        [TestMethod]
+        public void MefCtor_TwoExports_Singleton()
+        {
+            // SuppressIssuesProvider is unusual as it exports two types via MEF.
+            // We expect both types to be exportable, and for them to return the
+            // same instance.
```
duncanp-sonar, line 61: The comment explains why we can't use the normal MEF test pattern.

I added an overload to `MefTestHelpers` to simplify writing this test and ones like it.

src/Integration.UnitTests/Suppression/SuppressedIssuesProviderTests.cs
```diff
@@ -49,6 +53,45 @@ public void TestInitialize()
...
+
+            providerImporter.Import.Should().BeSameAs(monitorImporter.Import);
+        }
+
+        private static Export[] GetRequiredMefDependencies()
+        {
+            // The constructor executes some initialization code, so we need
+            // to provide a valid configuration to prevent it throwing.
+            var tracker = new Mock<IActiveSolutionBoundTracker>();
+            tracker.Setup(x => x.CurrentConfiguration).Returns(BindingConfiguration.Standalone);
```
duncanp-sonar, line 84: Yuck.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -55,13 +56,20 @@ public sealed class SonarQubeIssuesProvider : ISonarQubeIssuesProvider
         private CancellationTokenSource cancellationTokenSource;
         private readonly IThreadHandling threadHandling;
 
-        public SonarQubeIssuesProvider(ISonarQubeService sonarQubeService, string sonarQubeProjectKey,
-            ILogger logger) : this(sonarQubeService, sonarQubeProjectKey, logger, new TimerFactory(), new ThreadHandling())
+        public SonarQubeIssuesProvider(ISonarQubeService sonarQubeService,
+            IServerBranchProvider serverBranchProvider,
```
duncanp-sonar, line 60: One new parameter.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -109,6 +118,7 @@ public IEnumerable<SonarQubeIssue> GetSuppressedIssues(string projectGuid, strin
             this.initialFetch?.Wait(MillisecondsToWaitForInitialFetch);
 
             if (this.suppressedFileIssues == null ||
+                this.suppressedFileIssues.Count == 0 ||
```
duncanp-sonar, line 121: Miscellaneous: added a check for empty lists here as a short-circuit.
Now that the tests use an empty list rather than `null`, they hit this new check.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -120,18 +120,37 @@ public static Export CreateExport<T>(object exportInstance, string contractName
...
+
+        /// <summary>
+        /// Performs a MEF compose with the supplied objects
+        /// </summary>
+        /// <param name="importers">The instance that is importing the exported type</param>
+        /// <param name="typesToExport">The types of export being tested</param>
+        /// <param name="additionalExports">Any other exports that are required. Can be empty.</param>
+        /// <remarks>Note: this overload is useful for testing types that have multiple exports
+        /// i.e. that export multiple types.</remarks>
+        public static void Compose(object[] importers, Type[] typesToExport, params Export[] additionalExports)
```
duncanp-sonar, line 136: Added an overload of the existing method that accepts arrays for all three arguments.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
rita-sonarsource: This changes the logic a bit. 
Previously if the suppressedFileIssues was empty, we would get to the second block of modules, and try to return module issues. We won't do that anymore.
Is there a reason for this change? How long has it been since modules were deprecated? I think the code in the current state might be slightly confusing, perhaps it would be better to do a pre-commit that removes the modules.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
```diff
@@ -198,14 +208,17 @@ private async Task SynchronizeSuppressedIssuesAsync()
                 cancellationTokenSource?.Cancel();
                 cancellationTokenSource = new CancellationTokenSource();
 
+                var serverBranch = await serverBranchProvider.GetServerBranchNameAsync();
```
rita-sonarsource, line 211: We should really have that rule that async methods need to receive cancellation tokens... 
There is cancellationTokenSource.Token right below that we could use. Does it make sense to pass it to GetServerBranchNameAsync?

src/Integration.UnitTests/Suppression/SuppressedIssuesProviderTests.cs
rita-sonarsource: https://64.media.tumblr.com/tumblr_lvlpg6KxLW1qg42pfo6_r2_250.gif

src/Integration/Suppression/SonarQubeIssuesProvider.cs
duncanp-sonar: Yes, probably. I'll add it later.

src/Integration/Suppression/SonarQubeIssuesProvider.cs
duncanp-sonar: Good point. I'll revert the change.
I don't know whether the module stuff is still relevant or not, but it's unrelated to the branch awareness sprint. I'm not going to touch it now.

PR #3433, opened 2022-12-09
WIP remove LoggerExtensions

rita-sonarsource:


src/Integration/Helpers/SonarLintOutputLogger.cs
```diff
@@ -49,5 +49,29 @@ public void WriteLine(string messageFormat, params object[] args)
         {
             VsShellUtils.WriteToSonarLintOutputPane(this.serviceProvider, messageFormat, args);
         }
+
```
rita-sonarsource, line 52: Moved as-is from the LoggerExtensions

src/Roslyn.Suppressions/Roslyn.Suppressions/Logger.cs
```diff
@@ -34,6 +34,11 @@ public void WriteLine(string messageFormat, params object[] args)
         {
             Debug.WriteLine(messageFormat, args);
         }
+
+        public void LogDebug(string messageFormat, params object[] args)
+        {
+            WriteLine(messageFormat, args);
```
rita-sonarsource, line 40: This would be a functional change, but I think it's ok. We were using Debug.WriteLine in the WriteLine implementation anyway.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -98,7 +97,7 @@ private void SuppressedIssuesMonitor_SuppressionsUpdateRequested(object sender,
             catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
             {
                 // Squash non-critical exceptions
-                logger.LogDebugExtended(ex.ToString());
```
rita-sonarsource, line 101: This was the only place that used LogDebugExtended, which was tricky to implement in the Logger:
ILogger.LogDebug receives params, and you cannot put optional args [CallerFilePath] and [CallerMemberName] after them. Since this was the only usage, I left it out for now.

src/Integration/Helpers/SonarLintOutputLogger.cs
```diff
@@ -49,5 +46,14 @@ public void WriteLine(string messageFormat, params object[] args)
...
+        public void LogDebug(string messageFormat, params object[] args)
```
rita-sonarsource, line 50: Tests are added in https://github.com/SonarSource/sonarlint-visualstudio/pull/3439

src/Integration/Helpers/SonarLintOutputLogger.cs
```diff
@@ -29,15 +29,12 @@ namespace SonarLint.VisualStudio.Integration
...
 
         [ImportingConstructor]
         public SonarLintOutputLogger([Import(typeof(SVsServiceProvider))]IServiceProvider serviceProvider)
         {
-            if (serviceProvider == null)
-            {
-                throw new ArgumentNullException(nameof(serviceProvider));
-            }
-            this.serviceProvider = serviceProvider;
+            this.serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
```
duncanp-sonar, line 37: FYI MEF will never pass a null in this case (it can for `ImportMany`, but here the composition will fail). We have this pattern in quite a few places, but I don't think it helps.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
duncanp-sonar: We only really used it for internal builds where we needed extra info.

src/Roslyn.Suppressions/Roslyn.Suppressions/Logger.cs
duncanp-sonar: Maybe prefix the message with DEBUG?

src/TestInfrastructure/Framework/TestLogger.cs
```diff
@@ -115,6 +115,11 @@ public void WriteLine(string messageFormat, params object[] args)
             WriteLine(string.Format(System.Globalization.CultureInfo.CurrentCulture, messageFormat, args));
         }
 
+        public void LogDebug(string message, params object[] args)
+        {
+            // no-op
```
duncanp-sonar, line 120: I think it's worth implementing it to dump to the console - It can be useful to see the output in the test to check that the output is being written correctly e.g.

```
            if (logToConsole)
            {
                Console.WriteLine("DEBUG: "+ ...);
            }
```

duncanp-sonar (approved):

src/Integration/Helpers/SonarLintOutputLogger.cs
rita-sonarsource: Yeah I removed it in future PRs 👍

src/Roslyn.Suppressions/Roslyn.Suppressions/Logger.cs
rita-sonarsource: Yes, it got lost in the moving, I handled it in the future PR

src/TestInfrastructure/Framework/TestLogger.cs
rita-sonarsource: Good suggestion, I'll merge as-is as I'm tired of the build failures and then open a new PR

PR #3439, opened 2022-12-09
Reuse daemon log level for LogVerbose in SonarLintOutputLogger

### LINKED ISSUE #3430: Allow the user to enable/disable debug logs without setting an environment variable
### Description

There are currently 2 separate ways to enable partial debug logs:

1. Setting an environment variable `SONARLINT_INTERNAL_LOG_DEBUG=true`: this requires to restart VS when changing values, and is too complicated for users to use.
2. The options page has logging verbosity but that applies only to the CFamily analyzer.

image

TODO:
- [x] Re-use the logging verbosity defined in the options page instead of environment variable (Re-use DaemonLogLevel for ShowDebugLogs --> verbose = true, else false)
- [x] Clean up Options page and remove the text that says the logging is only for additional languages
- [x] Remove references to the environment variable

rita-sonarsource:
Fixes #3430

Tested and it works great!

### DIFF
--- a/src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
+++ b/src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
@@ -18,8 +18,6 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System;
-using FluentAssertions;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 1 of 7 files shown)


src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
```diff
@@ -30,21 +28,12 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Helpers
     [TestClass]
     public class SonarLintOutputTests
     {
-        [TestMethod]
```
rita-sonarsource, line 33: Not needed anymore, it's a MEF ctor

bigfluffycookie (approved):
Very good, glad you changed it back to Mef import the sonarLintsettings directly again instead of using the serviceProvider. Also yay that it works!

duncanp-sonar (approved):

PR #3440, opened 2022-12-09
Fix logging label in the Options page

rita-sonarsource:
Part of https://github.com/SonarSource/sonarlint-visualstudio/issues/3430
image

### DIFF
--- a/src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml
+++ b/src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml
@@ -42,7 +42,7 @@
 
         <StackPanel x:Name="VerbosityPanel" Grid.Row="2" Grid.ColumnSpan="2">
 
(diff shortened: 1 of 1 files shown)


src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml
```diff
@@ -42,7 +42,7 @@
 
         <StackPanel x:Name="VerbosityPanel" Grid.Row="2" Grid.ColumnSpan="2">
 
-            <TextBlock Text="SonarLint output verbosity for the analysis of additional languages:"
+            <TextBlock Text="Logs output verbosity:"
```
duncanp-sonar, line 45: `Logs output verbosity` sounds slightly odd.

`Log output verbosity`, `Logger output verbosity`, or `Log verbosity` perhaps. VS has:
image

Or avoid the problem with something like `Logging level:`

duncanp-sonar (approved):

PR #3441, opened 2022-12-09
Add cancellation token to IServerBranchProvider

### LINKED ISSUE #3434: Add cancellation token to IServerBranchProvider
### Description

`GetServerBranchName` is async, but doesn't accept a cancellation token.

duncanp-sonar:
Fixes #3434

### DIFF
--- a/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
+++ b/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
@@ -95,7 +95,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
                 }
 
                 var projectKey = bindingConfiguration.Project.ProjectKey;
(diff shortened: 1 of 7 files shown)


src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -95,7 +95,7 @@ public async Task SynchronizeWithServer()
                 }
 
                 var projectKey = bindingConfiguration.Project.ProjectKey;
-                var serverBranch = await serverBranchProvider.GetServerBranchNameAsync();
+                var serverBranch = await serverBranchProvider.GetServerBranchNameAsync(CancellationToken.None);
```
duncanp-sonar, line 98: The taint synchronizer doesn't currently have a `CancellationTokenSource` so we can't pass anything except `CancellationToken.None` here.

I've added #3442 to cover adding cancellability.

rita-sonarsource (approved):

PR #3443, opened 2022-12-09
Remove env var `SONARLINT_INTERNAL_LOG_DEBUG`

rita-sonarsource:
Cleanup following #3430


georgii-borovinskikh-sonarsource (commented):
Was that method unused before? I'm a bit confused

georgii-borovinskikh-sonarsource (approved):

Issue #3431, opened 2022-12-09
Fix LibGit2Sharp packaging in VSIX


duncanp-sonar:
### Description

Problems:
- [ ] the managed assembly isn't included in the VSIX
- [ ] *all* of the native binaries are included in the VSIX (including the half-dozen Linux assemblies)
- [ ] native assemblies are included in the `lib` folder, which is where we put our CFamily files. Probably not a problem.

PR #3449, opened 2022-12-13
Add LibGit binaries to VSIX

duncanp-sonar:
TODO
- the VSIX contains the native binaries for Linux and OSX, which we don't need.
- the VS2019 and VS2022 probably only need one copy of the Windows native binaries i.e. either x86 or x64

Part of #3431

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -52,6 +52,8 @@
   <ItemGroup Label="NuGet packages">
     <PackageReference Include="Newtonsoft.Json" Condition=" $(RequiresNewtonsoft)=='true'" Version="$(NewtonsoftVersion)" />
     <PackageReference Include="System.IO.Abstractions" Condition=" $(RequiresSystemIOAbstractions)=='true'" Version="9.0.4" />
+
+    <PackageReference Include="LibGit2Sharp" Version="0.26.2" Condition="$(RequiresLibGit)=='true'" GeneratePathProperty="true" PrivateAssets="all"  />  
   </ItemGroup>
 
 </Project>
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -410,3 +410,966 @@ http://www.bouncycastle.org/licence.html
 License notice for MS NodeJsTools node.exe discovery code
 ---------------------------------------------------------
 https://github.com/microsoft/nodejstools/blob/275e85d5cd95cad9122f59a76f9e49bead66101b/LICENSE
+
+
+License notice for libgit2sharp
+-----------------------------------------------------
+https://github.com/libgit2/libgit2sharp/blob/1af49546917b69499f2d41cf844509c53b37eaca/LICENSE.md
+
+The MIT License
+
+Copyright (c) LibGit2Sharp contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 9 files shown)


THIRD-PARTY-NOTICES
```diff
@@ -409,4 +409,967 @@ http://www.bouncycastle.org/licence.html
 
 License notice for MS NodeJsTools node.exe discovery code
 ---------------------------------------------------------
-https://github.com/microsoft/nodejstools/blob/275e85d5cd95cad9122f59a76f9e49bead66101b/LICENSE
\ No newline at end of file
+https://github.com/microsoft/nodejstools/blob/275e85d5cd95cad9122f59a76f9e49bead66101b/LICENSE
```
duncanp-sonar, line 412: License notices for the LibGit2Sharp and LibGit native libraries

Directory.Build.props
```diff
@@ -52,6 +52,8 @@
   <ItemGroup Label="NuGet packages">
     <PackageReference Include="Newtonsoft.Json" Condition=" $(RequiresNewtonsoft)=='true'" Version="$(NewtonsoftVersion)" />
     <PackageReference Include="System.IO.Abstractions" Condition=" $(RequiresSystemIOAbstractions)=='true'" Version="9.0.4" />
+
+    <PackageReference Include="LibGit2Sharp" Version="0.26.2" Condition="$(RequiresLibGit)=='true'" GeneratePathProperty="true" PrivateAssets="all"  />
```
duncanp-sonar, line 56: In theory, just adding the `LibGit2Sharp` NuGet package to the `ConnectedMode` project should have been enough for the VSIX build tools to recognise that the `LibGit2Sharp` package and the `LibGit2Sharp.NativeBinaries` it depends are required in the VSIX.

However, that wasn't working. The `LibGit2Sharp.NativeBinaries` package contents were added to the VSIX, but not the `LibGit2Sharp` managed package (even with the usual hacks in the `Integration.Vsix` project file about which output groups to include).

Instead, I marked the contents of the package with `PrivateAssets=all` and added the reference to the `ConnectedMode` project, its unit tests, and the `Integration .Vsix` project.

That causes the managed LibGit binaries and all of the native LibGit binaries (i.e. for all Unix platforms as well as Windows) to be added to the VSIX. This isn't exactly what we want, but it's a start and should unblock us.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -436,6 +437,7 @@
     <ItemGroup>
       <VSIXSourceItem Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' == 'System.IO.Abstractions'" />
       <VSIXSourceItem Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' == 'BouncyCastle'" />
+      <VSIXSourceItem Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' == 'LibGit2Sharp'" />
```
duncanp-sonar, line 440: The comment in line 433 says this is a workaround for a bug. It still appears to be necessary (without it, the managed LibGitSharp assembly is not included in the VSIX).

ugras-ergun-sonarsource (approved):

Issue #3454, opened 2022-12-13
Implement IServerBranchProvider


duncanp-sonar:
### Description

* if not in connected mode, return null
~~* if no open solution, return null~~
* if no git repo, return null
* open current repo, call branch mapper, return result.

PR #3455, opened 2022-12-13
Make IBranchMatcher MEF-exportable

duncanp-sonar:
Prep for #3454

### DIFF
--- a/src/ConnectedMode/BranchMatcher.cs
+++ b/src/ConnectedMode/BranchMatcher.cs
@@ -45,6 +45,8 @@ namespace SonarLint.VisualStudio.ConnectedMode
     }
 
     [Export(typeof(IBranchMatcher))]
+    // Note: this class doesn't *need* to be a singleton; it doesn't hold any unique state.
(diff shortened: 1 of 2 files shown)


src/ConnectedMode/BranchMatcher.cs
```diff
@@ -40,21 +41,21 @@ public interface IBranchMatcher
         /// This is the same algorithm as the other SonarLint flavours (?except in how we treated branch histories.
         /// We treat branch histories as series of linear commits ordered by time, even if there are merges).
         /// </remarks>
-        Task<string> GetMatchedBranch(string projectKey);
+        Task<string> GetMatchingBranch(string projectKey, IRepository gitRepo);
```
duncanp-sonar, line 44: Two changes to make the class easier to use from the `ServerBranchMapper`:
1) MEF exportable so it can be created easily
2) passing the repo as an argument so that we don't need to create a new instance every time the git branch changes.

ugras-ergun-sonarsource (approved):

PR #3456, opened 2022-12-14
Implement ServerBranchProvider

duncanp-sonar:
Part of #3454

### DIFF
--- a/src/ConnectedMode/ServerBranchProvider.cs
+++ b/src/ConnectedMode/ServerBranchProvider.cs
@@ -21,7 +21,9 @@
 using System.ComponentModel.Composition;
 using System.Threading;
 using System.Threading.Tasks;
+using LibGit2Sharp;
 using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Integration;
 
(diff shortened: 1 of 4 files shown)


src/ConnectedMode/ServerBranchProvider.cs
```diff
@@ -21,30 +21,73 @@
...
+        public async Task<string> GetServerBranchNameAsync(CancellationToken token)
         {
             string branchName = null;
 
-            // TODO: calculate branch name
-            logger.WriteLine(Resources.BranchProvider_ServerBranchName, branchName ?? Resources.NullBranchName);
+            var config = configurationProvider.GetConfiguration();
+            if (config.Mode == SonarLintMode.Standalone)
+            {
+                logger.LogVerbose(Resources.BranchProvider_NotInConnectedMode);
```
duncanp-sonar, line 72: The tests don't check that the `LogVerbose` messages are logged - the test infrastructure doesn't currently support it, and it's a low-value check.

duncanp-sonar:
I've done a manual smoke-test that the code executes in VS, although it didn't use the version of LibGit from our extension (perhaps one was already loaded).

ugras-ergun-sonarsource (approved):

Issue #3393, opened 2022-12-06
Create `About` screen


rita-sonarsource:
### Description

* Add fancy product logo?
* Add extension version? release date? Open sonarlint.org ?

TBD, ask Andrew

@leyla

PR #3458, opened 2022-12-14
Update About SonarLint window

bigfluffycookie:
Updates following [#3393](https://github.com/SonarSource/sonarlint-visualstudio/issues/3393)

NewAbout

second:
image


src/Integration.Vsix/Commands/HelpMenu/AboutDialog.xaml
```diff
@@ -7,37 +7,42 @@
...
         <Style x:Key="TextBlockStyle" TargetType="TextBlock">
-            <Setter Property="FontSize" Value="12"/>
             <Setter Property="Foreground" Value="Black"/>
+            <Setter Property="TextOptions.TextFormattingMode" Value="Ideal"/>
+            <Setter Property="TextOptions.TextRenderingMode" Value="ClearType"/>
+            <Setter Property="FontFamily" Value="..\..\Resources\MavenPro-Regular.ttf #Maven Pro"/>
         </Style>
     </Window.Resources>
-    <Grid Margin="8,8,8,8">
+    <Grid>
```
rita-sonarsource, line 30: Having a grid that is essentially a 1x1 square defeats the purpose of the grid. It's doable of course, but it's harder to read and maintain when elements have such large margins. It is also more error-prone as it's difficult to see when elements are not aligned properly, something that is better achieved via grid rows and columns. For example, the last row (copyright and close button) should be on the alignment, but because they are standalone elements, it is harder to do.
image

I think there should be a 2x2 grid:
* Row 0, Column 0: picture and version
* Row 0, Column 1: version
* Row 1, Column 0: copyright
* Row 1, Column 1: button

With appropriate margins and alignments. WDYT?

rita-sonarsource (changes requested):
See my comment regarding grid and margins. Otherwise looks awesome.

PR #3467, opened 2022-12-15
Run unit tests right after the build step

rita-sonarsource:


rita-sonarsource:
Canceling, we might want the VSIX artefact despite the test failures

PR #3482, opened 2022-12-21
[DO NOT MERGE] Outline IGitEvents implementation

duncanp-sonar:


src/Core/Suppressions/ISuppressedIssuesEvents.cs
```diff
@@ -0,0 +1,35 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarLint.VisualStudio.Core.Suppression
```
duncanp-sonar, line 23: Not sure why this is here...

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -96,11 +96,11 @@ public void ActiveSolutionBoundTracker_ArgChecks()
         {
             // Arrange
             Exceptions.Expect<ArgumentNullException>(() =>
-                new ActiveSolutionBoundTracker(null, new ConfigurableActiveSolutionTracker(), loggerMock.Object));
+                new ActiveSolutionBoundTracker(null, new ConfigurableActiveSolutionTracker(), null, null, loggerMock.Object));
```
duncanp-sonar, line 99: Hacky test fix-up

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -52,13 +53,19 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
...
+        private readonly IGitEvents gitEvents;
+
         public event EventHandler<ActiveSolutionBindingEventArgs> SolutionBindingChanged;
         public event EventHandler SolutionBindingUpdated;
 
         public BindingConfiguration CurrentConfiguration { get; private set; }
 
         [ImportingConstructor]
-        public ActiveSolutionBoundTracker(IHost host, IActiveSolutionTracker activeSolutionTracker, ILogger logger)
+        public ActiveSolutionBoundTracker(IHost host, IActiveSolutionTracker activeSolutionTracker,
```
duncanp-sonar, line 65: The ASBT imports a list of things that need to be notified about solution changes before the events are raised.

The server branch mapper and git monitor both implement this interface.

src/ConnectedMode/CachingServerBranchProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        public async Task<string> GetServerBranchNameAsync(CancellationToken token)
+        {
+            // TODO: caching
+
+            return await branchProvider.GetServerBranchNameAsync(token);
+        }
+
+        public void OnSolutionBindingChanged()
```
rita-sonarsource, line 32: This doesn't look right to me. We are introducing two ways to be notified when a solution binding has changed, either via implementing this delegate, or via importing ActiveSolutionBoundTracker and registering to its events. Working with the delegate doesn't allow the granularity of picking which solution event to register to (before or after). This is essentially means that if you want a Before then you should inherit from this, and if you want an After, you should register to the event.

I think it is cleaner to have just the events, with a Before event added to those that need it (which is this class).

ASBT already knows of IGitEvents (because it registers to HeadChanged), so it might as well call Refresh on it.

duncanp-sonar: I'll go with the event-based approach, although it doesn't really have anything in it's favour except consistency with the current (broken) approach.

When we look at fixing the eventing model, we should consider a callback-based approach:
* it's more loosely-coupled
* it's easier to test

PR #3483, opened 2022-12-22
Listen to Git Changes and raise an event 

### LINKED ISSUE #3477: Listen to Git Changes and raise an event
### Description

TODO
* how do we detect changes? libgit2sharp, FileWatcher?
* which repo changes do we need to listen to? - talk to the Java SonarLint folks.

ugras-ergun-sonarsource:
fixes #3477


src/Core/GitEventsMonitor.cs
```diff
@@ -0,0 +1,79 @@
...
+    public class GitEventsMonitor : IGitEvents
+    {
+        public event EventHandler HeadChanged;
+        private IFileSystemWatcher fileSystemWatcher;
+
+        private const string GitFolder = ".git";
+        private const string HEADFile = "HEAD";
+
+
+        public GitEventsMonitor(IGitWorkspaceService gitWorkspaceService) : this(gitWorkspaceService, new FileSystemWatcherFactory())
```
duncanp-sonar, line 47: As discussed yesterday, why does this class need `IGitWorkspaceService`, rather than just the repo root path?

src/Core/GitEventsMonitor.cs
```diff
@@ -0,0 +1,79 @@
...
+        internal GitEventsMonitor(IGitWorkspaceService gitWorkspaceService, IFileSystemWatcherFactory fileSystemFactory)
+        {
+            WatchGitEvents(gitWorkspaceService, fileSystemFactory);
+        }
+
+        private void WatchGitEvents(IGitWorkspaceService gitWorkspaceService, IFileSystemWatcherFactory fileSystemFactory)
+        {
+            var repoFolder = gitWorkspaceService.GetRepoRoot();
+            
+            if(repoFolder == null) { return; }
```
duncanp-sonar, line 59: Trivial: missing whitespace.

src/Core/GitEventsMonitor.cs
duncanp-sonar: For future reference, please don't push more commits if a review is in progress; it makes it much harder for the reviewer.
By all means make the changes locally, but don't push them.

src/Core.UnitTests/GitEventsMonitorTests.cs
```diff
@@ -0,0 +1,70 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
+namespace SonarLint.VisualStudio.Core.UnitTests
+{
+    [TestClass]
+    public class GitEventsMonitorTests
+    {
+        [TestMethod]
+        public void HeadChanged_EventInvoked()
```
duncanp-sonar, line 34: Missing test case: should not fail if there are no listeners. Currently, it will throw in line 74 of the monitor:

```
  HeadChanged.Invoke(...)
```

It should be
```
  HeadChanged?.Invoke(...)
```

src/Core/GitEventsMonitor.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    public interface IGitEvents
+    {
+        /// <summary>
+        /// Raised when the current head changes
+        /// </summary>
+        event EventHandler HeadChanged;
+    }
+
+    public class GitEventsMonitor : IGitEvents
```
duncanp-sonar, line 38: This class is expected to be short-lived i.e. a new one created each time a solution is opened.
Since it isn't expected to live for the full lifetime of VS, it should clean up after itself

i.e. the file system watcher is disposable, so this class should be disposable and clean it up correctly (might as well make this class `sealed` as part of this, since it will make the disposal code simpler).

duncanp-sonar (changes requested):
See comments.

src/Core.UnitTests/GitEventsMonitorTests.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            var fileSystemWatcherFactory = new Mock<IFileSystemWatcherFactory>();
+            fileSystemWatcherFactory.Setup(f => f.FromPath("C:\\Some Path\\.git")).Returns(fileSystemWatcher.Object);
+
+            _ = CreateTestSubject("C:\\Some Path", fileSystemWatcherFactory);
+
+            fileSystemWatcher.VerifySet(w => w.Filter = "HEAD", Times.Once);
+            fileSystemWatcher.VerifySet(w => w.EnableRaisingEvents = true, Times.Once);
+
+            fileSystemWatcher.Raise(w => w.Changed += null, null, null);
```
duncanp-sonar, line 86: There's a `Should().NotThrow(action)` FluentAssertion method that would make the check more explicit (we don't appear to have used it so far, but perhaps we should)

src/Core.UnitTests/GitEventsMonitorTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            var fileSystemWatcher = new Mock<IFileSystemWatcher>();
+
+            var fileSystemWatcherFactory = new Mock<IFileSystemWatcherFactory>();
+            fileSystemWatcherFactory.Setup(f => f.FromPath("C:\\Some Path\\.git")).Returns(fileSystemWatcher.Object);
+
+            var testSubject = CreateTestSubject("C:\\Some Path", fileSystemWatcherFactory);
+
+            fileSystemWatcher.VerifySet(w => w.Filter = "HEAD", Times.Once);
+            fileSystemWatcher.VerifySet(w => w.EnableRaisingEvents = true, Times.Once);
```
duncanp-sonar, line 66: Copy-and-paste-itis: we don't need to check the file watcher setup in every single test.
It's just clutter and detracts from the real purpose of the test.

src/Core.UnitTests/GitEventsMonitorTests.cs
```diff
@@ -0,0 +1,104 @@
...
+    public class GitEventsMonitorTests
+    {
+        [TestMethod]
+        public void HeadChanged_EventInvoked()
+        {
+            int counter = 0;
+
+            var fileSystemWatcher = new Mock<IFileSystemWatcher>();
+
+            var fileSystemWatcherFactory = new Mock<IFileSystemWatcherFactory>();
```
duncanp-sonar, line 40: Copy-and-paste-itis: would be cleaner as a `CreateFSWFactory` method.

duncanp-sonar (approved):
Looks good, apart from thettests containing too much copy-and-paste code.

PR #3485, opened 2022-12-22
Add PreSolutionBindingChanged to IActiveSolutionBoundTracker

### LINKED ISSUE #3484: Add pre-update event to IActiveSolutionBoundTracker
### Description

Currently there are multiple classes that listen to the `IActiveSolutionBoundTracker.SolutionBindingChanged` event.

The order in which listeners are notified is not deterministic, which is problem if there are dependencies between the listeners.

A minor improvement would be to add a `PreSolutionBindingChanged` event. That would at least allow some crude sequencing of operations between listeners.

duncanp-sonar:
Fixes #3484

### DIFF
--- a/src/Core/Binding/IActiveSolutionBoundTracker.cs
+++ b/src/Core/Binding/IActiveSolutionBoundTracker.cs
@@ -52,6 +52,15 @@ namespace SonarLint.VisualStudio.Core.Binding
         /// </summary>
         event EventHandler<ActiveSolutionBindingEventArgs> SolutionBindingChanged;
 
+        /// <summary>
+        /// Raised when an existing binding has been updated, and before the <see cref="SolutionBindingUpdated"/> event
+        /// </summary>
+        event EventHandler PreSolutionBindingUpdated;
+
+        /// <summary>
+        /// Raised when an existing binding has been updated i.e. the solution is still bound to the same
+        /// Sonar project (e.g. the user updated the binding via the Team Explorer)
+        /// </summary>
         event EventHandler SolutionBindingUpdated;
     }
 
--- a/src/Core/Binding/IActiveSolutionBoundTracker.cs
+++ b/src/Core/Binding/IActiveSolutionBoundTracker.cs
@@ -36,6 +36,15 @@ namespace SonarLint.VisualStudio.Core.Binding
         BindingConfiguration CurrentConfiguration { get; }
 
         /// <summary>
+        /// Raised after the binding has been updated, and before the <see cref="SolutionBindingUpdated"/> event
+        /// </summary>
+        /// <remarks>Most listeners should use the <see cref="SolutionBindingChanged"> event.
+        /// This event is intended for components that need to recalculate/refresh data that
+        /// is then used by listeners that consume the <see cref="SolutionBindingChanged"/> event
+        /// e.g. components that cache solution-level data.</remarks>
+        event EventHandler<ActiveSolutionBindingEventArgs> PreSolutionBindingChanged;
+
+        /// <summary>
         /// Event to notify subscribers when the binding status of a solution have changed.
         /// This occurs when a new solution is opened, or the SonarQube binding status of the solution changes.
         /// The event is raised after the <see cref="ISonarQubeService"/> connection has finished
--- a/src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
+++ b/src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
@@ -183,14 +183,15 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
 
             using (var testSubject = new ActiveSolutionBoundTracker(this.host, this.activeSolutionTracker, loggerMock.Object))
             {
-                var solutionBindingChangedEventCount = 0;
-                testSubject.SolutionBindingChanged += (obj, args) => { solutionBindingChangedEventCount++; };
+                var eventCounter = new EventCounter(testSubject);
 
                 // Sanity
                 testSubject.CurrentConfiguration.Mode.Should().Be(SonarLintMode.LegacyConnected, "Initially bound");
                 this.errorListController.RefreshCalledCount.Should().Be(0);
                 this.errorListController.ResetCalledCount.Should().Be(0);
-                solutionBindingChangedEventCount.Should().Be(0, "no events raised during construction");
+                eventCounter.PreSolutionBindingChangedCount.Should().Be(0, "no events raised during construction");
+                eventCounter.SolutionBindingChangedCount.Should().Be(0, "no events raised during construction");
+                eventCounter.SolutionBindingUpdatedCount.Should().Be(0, "no events raised during construction");
                 VerifyAndResetBoundSolutionUIContextMock(isActive: true);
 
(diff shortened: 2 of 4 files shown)


src/Core/Binding/IActiveSolutionBoundTracker.cs
```diff
@@ -35,6 +35,15 @@ public interface IActiveSolutionBoundTracker
...
 
+        /// <summary>
+        /// Raised after the binding has been updated, and before the <see cref="SolutionBindingUpdated"/> event
+        /// </summary>
+        /// <remarks>Most listeners should use the <see cref="SolutionBindingChanged"> event.
+        /// This event is intended for components that need to recalculate/refresh data that
+        /// is then used by listeners that consume the <see cref="SolutionBindingChanged"/> event
+        /// e.g. components that cache solution-level data.</remarks>
+        event EventHandler<ActiveSolutionBindingEventArgs> PreSolutionBindingChanged;
+
```
duncanp-sonar, line 46: Added two new 'PreXXX' events with documentation.

Calling the events e.g. `BeforeSolutionBindingChanged` doesn't make sense because the solution binding has already changed, so I went with `Pre` as the prefix rather than `Before`.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -183,14 +184,16 @@ public void ActiveSolutionBoundTracker_Changes()
 
             using (var testSubject = new ActiveSolutionBoundTracker(this.host, this.activeSolutionTracker, loggerMock.Object))
             {
-                var solutionBindingChangedEventCount = 0;
-                testSubject.SolutionBindingChanged += (obj, args) => { solutionBindingChangedEventCount++; };
+                var eventCounter = new EventCounter(testSubject);
```
duncanp-sonar, line 187: Added a new helper class to make registering for the events and checking they are raised/not raised simpler.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -183,14 +184,16 @@ public void ActiveSolutionBoundTracker_Changes()
...
 
                 // Sanity
                 testSubject.CurrentConfiguration.Mode.Should().Be(SonarLintMode.LegacyConnected, "Initially bound");
                 this.errorListController.RefreshCalledCount.Should().Be(0);
                 this.errorListController.ResetCalledCount.Should().Be(0);
-                solutionBindingChangedEventCount.Should().Be(0, "no events raised during construction");
+                eventCounter.PreSolutionBindingChangedCount.Should().Be(0, "no events raised during construction");
+                eventCounter.SolutionBindingChangedCount.Should().Be(0, "no events raised during construction");
+                eventCounter.PreSolutionBindingUpdatedCount.Should().Be(0, "no events raised during construction");
+                eventCounter.SolutionBindingUpdatedCount.Should().Be(0, "no events raised during construction");
```
duncanp-sonar, line 196: I've updated all of the existing tests to use the new helper class, and to check all of the events.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -319,33 +336,14 @@ public void OnBindingStateChanged_NewConfiguration_EventRaised()
 
                 // Assert
                 // Different config so event should be raised
-                solutionBindingChangedEventCount.Should().Be(1);
-            }
-        }
-
-        [TestMethod]
-        public void OnBindingStateChanged_SameConfiguration_EventNotRaised()
```
duncanp-sonar, line 327: Since we're now checking all of the event counts in each test, this test is now redundant (there's another test that does the same thing below).

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -319,33 +336,14 @@ public void OnBindingStateChanged_NewConfiguration_EventRaised()
...
-
-                // Assert
-                // Same config so event should not be raised
-                solutionBindingChangedEventCount.Should().Be(0);
+                eventCounter.PreSolutionBindingChangedCount.Should().Be(1);
+                eventCounter.SolutionBindingUpdatedCount.Should().Be(0);
+                eventCounter.PreSolutionBindingChangedCount.Should().Be(1);
+                eventCounter.SolutionBindingUpdatedCount.Should().Be(0);
+
+                eventCounter.RaisedEventNames.Should().HaveCount(2);
```
duncanp-sonar, line 344: Additional check the events are raised in the expected order.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -499,31 +497,39 @@ public void SolutionBindingUpdated_WhenClearBoundProject_NotRaised()
...
                 // Assert
-                callCount.Should().Be(1);
+                eventCounter.PreSolutionBindingUpdatedCount.Should().Be(1);
+                eventCounter.SolutionBindingUpdatedCount.Should().Be(1);
+                eventCounter.PreSolutionBindingChangedCount.Should().Be(0);
+                eventCounter.SolutionBindingChangedCount.Should().Be(0);
+
+                eventCounter.RaisedEventNames.Should().HaveCount(2);
+                eventCounter.RaisedEventNames[0].Should().Be("PreSolutionBindingUpdated");
+                eventCounter.RaisedEventNames[1].Should().Be("SolutionBindingUpdated");
```
duncanp-sonar, line 532: Additional check that the events are raised in the correct order.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -568,5 +574,44 @@ private void VerifyAndResetBoundSolutionUIContextMock(bool isActive)
             vsMonitorMock.Verify(x => x.SetCmdUIContext(It.IsAny<uint>(), It.IsAny<int>()), Times.Once);
             vsMonitorMock.Invocations.Clear();
         }
+
+        private class EventCounter
```
duncanp-sonar, line 578: Simple helper - counts each of the raised events, and records the order in which they are raised.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
ugras-ergun-sonarsource: I'm not sure if this is definitive proof that it'll always work this way, but I can't think of a better way to do it, and this is better than nothing.

ugras-ergun-sonarsource (approved):

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
duncanp-sonar: The events are raised and handled synchronously on a single thread so the order should be predictable, shouldn't it? Or is there something that makes it non-deterministic?

Issue #3486, opened 2022-12-22
Add bound solution git monitor


duncanp-sonar:
### Description

The `GitEventsMonitor` class is low-level and just listens to the events from a specific repo.

We need a higher-level component that reacts as solutions are opened/closed to change the repo being tracked.

PR #3488, opened 2022-12-22
Add BoundSolutionGitMonitor

duncanp-sonar:
Part of #3486

TODO - tests

### DIFF
--- a/src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
+++ b/src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
@@ -124,16 +124,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             }
         }
 
-        private static ActiveSolutionBoundTracker CreateTestSubject(IHost host,
-            IActiveSolutionTracker solutionTracker,
-            ILogger logger = null,
-            IBoundSolutionGitMonitor gitEvents = null)
-        {
-            logger ??= new TestLogger(logToConsole: true);
-            gitEvents ??= Mock.Of<IBoundSolutionGitMonitor>();
-            return new ActiveSolutionBoundTracker(host, solutionTracker, gitEvents, logger);
-        }
-
         [TestMethod]
         public void ActiveSolutionBoundTracker_When_ConnectAsync_Throws_Write_To_Output()
         {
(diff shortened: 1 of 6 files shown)


src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -92,26 +92,14 @@ public void TestInitialize()
             serviceProvider.RegisterService(typeof(SVsShellMonitorSelection), vsMonitorMock.Object, replaceExisting: true);
         }
 
-        [TestMethod]
-        public void ActiveSolutionBoundTracker_ArgChecks()
```
duncanp-sonar, line 96: Deleted - MEF won't call the constructor with nulls, so this test is pointless.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -533,6 +521,16 @@ public void SolutionBindingUpdated_WhenSetBoundProject_EventsRaisedInExpectedOrd
             }
         }
 
+        private static ActiveSolutionBoundTracker CreateTestSubject(IHost host,
```
duncanp-sonar, line 524: Helper method to avoid having to change every test to support the new parameter.

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -60,11 +61,15 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
         public BindingConfiguration CurrentConfiguration { get; private set; }
 
         [ImportingConstructor]
-        public ActiveSolutionBoundTracker(IHost host, IActiveSolutionTracker activeSolutionTracker, ILogger logger)
+        public ActiveSolutionBoundTracker(IHost host,
+            IActiveSolutionTracker activeSolutionTracker,
+            IBoundSolutionGitMonitor gitEvents,
+            ILogger logger)
         {
-            extensionHost = host ?? throw new ArgumentNullException(nameof(host));
```
duncanp-sonar, line 65: No point in throwing on nulls - MEF won't call the constructor with nulls (unless you are using the `ImportMany` attribute).

duncanp-sonar:
@ugras-ergun-sonarsource, @rita-sonarsource FYI WIP in progress with no automated tests.

I've manually tested the following:
* the imports work in VS
* the `ActiveSolutionBoundTracker` refreshes the monitor correctly
* the monitor creates a new low-level repo monitor correctly.

However: changing the branch in VS does not cause any events to be raised. More specifically, the low-level monitor class isn't raising any events, so the problem is either:
* the high-level class isn't creating the low-level class correctly, or
* the low-level class isn't detecting the changes as expected.

I had a look on the disc, and the time-stamp of the `HEAD` file is changing when you switch branch in VS.

duncanp-sonar:
Replaced by #3492

Issue #3489, opened 2022-12-22
Add Stateful IServerBranchMapper


duncanp-sonar:
### Description

Re-calculating the appropriate Sonar server branch is a potentially expensive operation. We should re-calculate the value when the bound solution changes and cache the result.

PR #3490, opened 2022-12-22
Add skeleton IStatefulServerBranchProvider

duncanp-sonar:
Part of #3489


duncanp-sonar:
@ugras-ergun-sonarsource, @rita-sonarsource this is a no-op implementation of the class. It doesn't currently do anything, but I have manually tested the following:
* it composes correctly in VS
* it refreshes when the `ActiveSolutionBoundTracker` event fires
* the suppression and taint synchronizers call it.

duncanp-sonar:
Replaced by #3493

PR #3491, opened 2022-12-27
 Track repo events in ActiveSolutionBoundTracker

### LINKED ISSUE #3487: Track repo events in ActiveSolutionBoundTracker
### Description

The `ActiveSolutionBoundTracker` should listen for git changes and notify listeners that changes have occurred.

We could either add a specific new event, or overload one of the existing events i.e. `SolutionBindingChanged` / `SolutionBindingUpdated`.

Overriding an existing event is hackier, but means we could avoid updating the taint synchronizer and suppressed issues tracker. However, if we go with that approach then we need to check whether there are any side-effects on other classes that listen to those events.

ugras-ergun-sonarsource:
fixes #3487

### DIFF
--- a/src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
+++ b/src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
@@ -546,6 +546,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
                 var eventCounter = new EventCounter(testSubject);
 
                 // Act
+                ConfigureSolutionBinding(new BoundSonarQubeProject(new Uri("http://foo"), "projectKey", "projectName"));
                 gitEventsMonitor.Raise(x => x.HeadChanged += null, null, null);
 
                 // Assert
@@ -556,6 +557,25 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
 
                 eventCounter.RaisedEventNames.Should().HaveCount(1);
                 eventCounter.RaisedEventNames[0].Should().Be("SolutionBindingUpdated");
+            }
+        }
+
+        [TestMethod]
+        public void GitRepoUpdated_UnBoundProject_SolutionBingingUpdatedNotInvoked()
+        {
+            var gitEventsMonitor = new Mock<IGitEvents>();
+
+            // Arrange
+            using (var testSubject = new ActiveSolutionBoundTracker(this.host, this.activeSolutionTracker, loggerMock.Object, gitEventsMonitor.Object))
+            {
+                var eventCounter = new EventCounter(testSubject);
+
+                // Act
+                ConfigureSolutionBinding(null);
+                gitEventsMonitor.Raise(x => x.HeadChanged += null, null, null);
+
+                // Assert
+                eventCounter.RaisedEventNames.Should().HaveCount(0);
             }
         }
 
--- a/src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
+++ b/src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
@@ -97,11 +97,13 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         {
             // Arrange
             Exceptions.Expect<ArgumentNullException>(() =>
-                new ActiveSolutionBoundTracker(null, new ConfigurableActiveSolutionTracker(), loggerMock.Object));
+                new ActiveSolutionBoundTracker(null, new ConfigurableActiveSolutionTracker(), loggerMock.Object, Mock.Of<IGitEvents>()));
             Exceptions.Expect<ArgumentNullException>(() =>
-                new ActiveSolutionBoundTracker(this.host, null, loggerMock.Object));
+                new ActiveSolutionBoundTracker(this.host, null, loggerMock.Object, Mock.Of<IGitEvents>()));
             Exceptions.Expect<ArgumentNullException>(() =>
-                new ActiveSolutionBoundTracker(this.host, new ConfigurableActiveSolutionTracker(), null));
+                new ActiveSolutionBoundTracker(this.host, new ConfigurableActiveSolutionTracker(), null, Mock.Of<IGitEvents>()));
+            Exceptions.Expect<ArgumentNullException>(() =>
+                new ActiveSolutionBoundTracker(this.host, new ConfigurableActiveSolutionTracker(), loggerMock.Object, null));
         }
 
         [TestMethod]
@@ -111,7 +113,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
             host.VisualStateManager.ClearBoundProject();
 
             // Act
-            using (var testSubject = new ActiveSolutionBoundTracker(this.host, this.activeSolutionTracker, loggerMock.Object))
+            using (var testSubject = new ActiveSolutionBoundTracker(this.host, this.activeSolutionTracker, loggerMock.Object, Mock.Of<IGitEvents>()))
             {
                 // Assert
                 testSubject.CurrentConfiguration.Mode.Should().Be(SonarLintMode.Standalone, "Unbound solution should report false activation");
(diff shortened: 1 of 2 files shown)


src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -84,6 +85,13 @@ public ActiveSolutionBoundTracker(IHost host, IActiveSolutionTracker activeSolut
             CurrentConfiguration = configurationProvider.GetConfiguration();
 
             SetBoundSolutionUIContext();
+
+            gitEventsMonitor.HeadChanged += GitEventsMonitor_HeadChanged;
+        }
+
+        private void GitEventsMonitor_HeadChanged(object sender, EventArgs e)
+        {
+            SolutionBindingUpdated?.Invoke(this, EventArgs.Empty);
```
ugras-ergun-sonarsource, line 94: This is only used in Taint and suppressions so reusing it makes sense

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -84,6 +85,13 @@ public ActiveSolutionBoundTracker(IHost host, IActiveSolutionTracker activeSolut
             CurrentConfiguration = configurationProvider.GetConfiguration();
 
             SetBoundSolutionUIContext();
+
+            gitEventsMonitor.HeadChanged += GitEventsMonitor_HeadChanged;
```
duncanp-sonar, line 89: The class has a `Dispose` method that unhooks the other extension handlers. I'm not sure it's strictly necessary, but we should be consistent and unhook this one too.

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -60,11 +60,12 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
         public BindingConfiguration CurrentConfiguration { get; private set; }
 
         [ImportingConstructor]
-        public ActiveSolutionBoundTracker(IHost host, IActiveSolutionTracker activeSolutionTracker, ILogger logger)
+        public ActiveSolutionBoundTracker(IHost host, IActiveSolutionTracker activeSolutionTracker, ILogger logger, IGitEvents gitEventsMonitor)
         {
             extensionHost = host ?? throw new ArgumentNullException(nameof(host));
             solutionTracker = activeSolutionTracker ?? throw new ArgumentNullException(nameof(activeSolutionTracker));
             this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            if (gitEventsMonitor == null) { throw new ArgumentNullException(nameof(gitEventsMonitor)); }
```
duncanp-sonar, line 68: Checking for null parameters is pointless for MEF constructors (unless the constructor uses `ImportMany`, MEF just won't call the constructor with a `null` parameter.

I think we should just remove the null checks and unit tests in classes like this when we are making other changes to the file.

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
duncanp-sonar: I don't think that necessarily follows, but I think it's ok for now.

duncanp-sonar, line 94: This means the tracker will raise the event even for standalone solutions. Is this what we expect?

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -97,11 +97,13 @@ public void ActiveSolutionBoundTracker_ArgChecks()
         {
             // Arrange
             Exceptions.Expect<ArgumentNullException>(() =>
-                new ActiveSolutionBoundTracker(null, new ConfigurableActiveSolutionTracker(), loggerMock.Object));
+                new ActiveSolutionBoundTracker(null, new ConfigurableActiveSolutionTracker(), loggerMock.Object, Mock.Of<IGitEvents>()));
```
duncanp-sonar, line 100: I would delete this test. As explained in another comment, it doesn't make sense for most MEF constructors.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -111,7 +113,7 @@ public void ActiveSolutionBoundTracker_Initialisation_Unbound()
             host.VisualStateManager.ClearBoundProject();
 
             // Act
-            using (var testSubject = new ActiveSolutionBoundTracker(this.host, this.activeSolutionTracker, loggerMock.Object))
+            using (var testSubject = new ActiveSolutionBoundTracker(this.host, this.activeSolutionTracker, loggerMock.Object, Mock.Of<IGitEvents>()))
```
duncanp-sonar, line 116: Note for the future: it would have been cleaned to refactor the existing tests to use the `CreateTestSubject`-method-with-optional-arguments pattern first.
I wouldn't do it now.

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
duncanp-sonar, line 94: Surely this should be raising the `PreSolutionBindingUpdated` event too?

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -533,6 +535,30 @@ public void SolutionBindingUpdated_WhenSetBoundProject_EventsRaisedInExpectedOrd
...
+                gitEventsMonitor.Raise(x => x.HeadChanged += null, null, null);
+
+                // Assert
+                eventCounter.PreSolutionBindingUpdatedCount.Should().Be(0);
+                eventCounter.SolutionBindingUpdatedCount.Should().Be(1);
+                eventCounter.PreSolutionBindingChangedCount.Should().Be(0);
+                eventCounter.SolutionBindingChangedCount.Should().Be(0);
+
+                eventCounter.RaisedEventNames.Should().HaveCount(1);
+                eventCounter.RaisedEventNames[0].Should().Be("SolutionBindingUpdated");
```
duncanp-sonar, line 558: The reason for checking the event names is to make sure that the events are raised in the expected order. It's not necessary if there is only one event - lines 552-555 already check that expected events have been raised.

However, why isn't the `Pre` event being raised too?

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
duncanp-sonar: Updated: it's done in the Git monitor PR.

duncanp-sonar (changes requested):
See comments/questions.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
duncanp-sonar: Update: done as part of the `bound git events` PR

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
ugras-ergun-sonarsource: I have realised that and fixed it later PR (stateful branch)

duncanp-sonar: Update: I see you've done this as part of `ui/stateful branch mapper`.

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
ugras-ergun-sonarsource: I have realised that and fixed it later PR (stateful branch)

duncanp-sonar: It handles all solution open/closed events from the lower-level `IActiveSolutionTracker`. It will raise the `SolutionBindingChanged` event for standalone solutions/when a solution is closed.

I don't think it will raise it will raise the `SolutionBindingUpdated` event except for up bound solutions, but I could be wrong.

duncanp-sonar (approved):

PR #3492, opened 2022-12-28
Add bound solution git monitor 

### LINKED ISSUE #3486: Add bound solution git monitor

ugras-ergun-sonarsource:
fixes #3486


src/Core/GitEventsMonitor.cs
```diff
@@ -65,7 +65,7 @@ private void WatchGitEvents(string repoFolder, IFileSystemWatcherFactory fileSys
 
             fileSystemWatcher.Filter = HEADFile;
             
-            fileSystemWatcher.Changed += HeadFileChanged;
+            fileSystemWatcher.Renamed += HeadFileChanged;
```
duncanp-sonar, line 68: Why are we only looking at file-rename events?

Does this have anything to do with the main purpose of this PR, or is it a completely separate bug in the `GitEventsMonitor`? It looks more like the latter. If so, it would have been better create a separate bug for it and handle it in a separated PR and not mix unrelated changes.

ugras-ergun-sonarsource: Because when we change it in Git it does not throw a changed event. It creates a new file, deletes the old one and renames the new one to "HEAD"

src/ConnectedMode.UnitTests/BoundSolutionGitMonitorTests.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            var gitWorkspaceService = CreateGitWorkSpaceService(repoPath);
+
+            var gitEventsMonitor = new Mock<IGitEvents>();
+
+            GitEventFactory gitEventFactory = (string path) =>
+            {
+                if (path != repoPath)
+                {
+                    throw new Exception();
```
duncanp-sonar, line 52: It would be better to include a message in the error explaining that this is a test failure and why i.e. unexpected repo name supplied to the factory.

Alternatively, a simple approach would be just to store the supplied path and assert later that it matches the expected path.

src/ConnectedMode.UnitTests/BoundSolutionGitMonitorTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            var gitWorkspaceService = CreateGitWorkSpaceService(originalPath);
+
+            var originalEventsMonitor = new Mock<IGitEvents>();
+            var newEventsMonitor = new Mock<IGitEvents>();
+
+            GitEventFactory gitEventFactory = (string path) =>
+            {
+                if (path != originalPath && path != newPath)
+                {
+                    throw new Exception();
```
duncanp-sonar, line 83: A descriptive error message would help debugging test failures.

duncanp-sonar (commented):
A couple of comments/questions.

src/Core/GitEventsMonitor.cs
duncanp-sonar: Do you think that is worth documenting in the code?

PR #3493, opened 2022-12-29
Add Stateful IServerBranchMapper 

### LINKED ISSUE #3489: Add Stateful IServerBranchMapper

ugras-ergun-sonarsource:
fixes #3489


src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -95,6 +95,7 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
 
         private void GitEventsMonitor_HeadChanged(object sender, EventArgs e)
         {
+            PreSolutionBindingUpdated?.Invoke(this, EventArgs.Empty);
```
duncanp-sonar, line 98: This change is unrelated to the stateful branch mapper changes. It would have been better to put it in a separate PR.

src/ConnectedMode/StatefulServerBranchProvider.cs
```diff
@@ -0,0 +1,92 @@
...
+        private void OnPreSolutionBindingChanged(object sender, ActiveSolutionBindingEventArgs e)
+        {
+            selectedBranch = null;
+        }
+
+        public async Task<string> GetServerBranchNameAsync(CancellationToken token)
+        {
+            if(selectedBranch == null)
+            {
+                selectedBranch = await branchProvider.GetServerBranchNameAsync(token);
```
duncanp-sonar, line 62: `selectedBranch == null` is being used as an indicator that a refresh is required.

However, `GetServerBranchNameAsync` will return `null` in some cases i.e. standalone solutions, non-git solutions, so in those cases the call to `GetServerBranchNameAsync` will always be made.

I guess it doesn't matter too much since the `branchProvider` should early-out in those cases.

src/ConnectedMode.UnitTests/StatefulServerBranchProviderTests.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            activeSolutionBoundTracker.Raise(asbt => asbt.PreSolutionBindingChanged += null, null, null);
+
+            //second call: Should use cache
+            serverBranch = await testSubject.GetServerBranchNameAsync(CancellationToken.None);
+
+            serverBranch.Should().Be("NewBranch");
+
+            serverBranchProvider.Verify(sbp => sbp.GetServerBranchNameAsync(It.IsAny<CancellationToken>()), Times.Exactly(2));
+        }
```
duncanp-sonar, line 92: Was it a deliberate decision not to test the `Dispose` method?

src/ConnectedMode.UnitTests/StatefulServerBranchProviderTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            //first call: Should use IServerBranchProvider
+            var serverBranch = await testSubject.GetServerBranchNameAsync(CancellationToken.None);
+
+            serverBranch.Should().Be("Branch");
+
+            serverBranchProvider.Verify(sbp => sbp.GetServerBranchNameAsync(It.IsAny<CancellationToken>()), Times.Once);
+
+            serverBranchProvider.Setup(sbp => sbp.GetServerBranchNameAsync(It.IsAny<CancellationToken>())).ReturnsAsync("NewBranch");
+
+            activeSolutionBoundTracker.Raise(asbt => asbt.PreSolutionBindingChanged += null, null, null);
```
duncanp-sonar, line 84: Missing test: the `PreSolutionBindingUpdated` event should be tested.

duncanp-sonar (commented):
A couple of comments/questions.

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
duncanp-sonar: I raised a separate ticket and PR for this issue: #3500 and #3501

src/ConnectedMode.UnitTests/StatefulServerBranchProviderTests.cs
duncanp-sonar: Done.

duncanp-sonar (approved):

Issue #3506, opened 2023-01-04
Display branch messages in the output window


duncanp-sonar:
### Description

There should be enough information in the output window for the user to see which server branch is being used (normal verbosity), and detail on how that branch was chosen (detailed verbosity).

PR #3511, opened 2023-01-04
Add logging to branch matching

duncanp-sonar:
Part of #3506

### DIFF
--- a/src/ConnectedMode/BranchMatcher.cs
+++ b/src/ConnectedMode/BranchMatcher.cs
@@ -26,6 +26,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using LibGit2Sharp;
 using SonarLint.VisualStudio.Core.ETW;
+using SonarLint.VisualStudio.Integration;
 using SonarQube.Client;
 
 namespace SonarLint.VisualStudio.ConnectedMode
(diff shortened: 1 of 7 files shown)


src/ConnectedMode/StatefulServerBranchProvider.cs
```diff
@@ -32,40 +33,51 @@ internal sealed class StatefulServerBranchProvider : IStatefulServerBranchProvid
...
                 // In these cases we expect the serverBranchProvider to return quickly so the impact of making unnecessary
                 // calls is not significant.
                 selectedBranch = await serverBranchProvider.GetServerBranchNameAsync(token);
             }
+            else
+            {
+                logger.LogVerbose(Resources.StatefulBranchProvider_CacheHit);
+            }
 
+            logger.WriteLine(Resources.StatefulBranchProvider_ReturnValue, selectedBranch ?? Resources.NullBranchName);
```
duncanp-sonar, line 80: This is the only new "normal" verbosity message.
The others are all verbose.

ugras-ergun-sonarsource (approved):

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -51,64 +52,85 @@ public interface IBranchMatcher
...
             this.sonarQubeService = sonarQubeService;
+            this.logger = logger;
         }
 
         public async Task<string> GetMatchingBranch(string projectKey, IRepository gitRepo, CancellationToken token)
         {
             Debug.Assert(sonarQubeService.IsConnected,
                 "Not expecting GetMatchedBranch to be called unless we are in Connected Mode");
+
+            logger.LogVerbose(Resources.BranchMapper_CalculatingServerBranch_Started);
```
duncanp-sonar, line 69: @ugras-ergun-sonarsource apologies, I forgot to add review comments to this PR.

I moved all of the code that was inside the try-catch to a separate method.
There was no change to existing logic except to add the logging.

The main reason was to make it easier to add a single final log message showing which branch was chosen.

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -51,64 +52,85 @@ public interface IBranchMatcher
...
                 }
-                return closestBranch;
             }
-            finally
+
+            if (closestBranch == null)
             {
-                CodeMarkers.Instance.GetMatchingBranchStop();
+                logger.LogVerbose(Resources.BranchMapper_NoMatchingBranchFound);
+                closestBranch = remoteBranches.First(rb => rb.IsMain).Name;
```
duncanp-sonar, line 130: Actually, I did change line 130 to set `closestBranch` rather than returning directly.

I did this before I created the new internal `Do` method, which made the change unnecessary. However, both versions work.

PR #3517, opened 2023-01-04
Log the thread id in verbose mode

### LINKED ISSUE #3515: Add thread id to output messages when in verbose logging
### Description

It is often useful to know which thread a call is being made on when debugging or testing.

Automatically logging the current thread ID for all messages when the logging mode is `verbose` would make this easy to check.

duncanp-sonar:
Fixes #3515

### DIFF
--- a/src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
+++ b/src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
@@ -42,9 +42,10 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Helpers
         {
             // Arrange
             var windowMock = new ConfigurableVsOutputWindow();
+            var sonarLintSettings = CreateSonarLintSettings(DaemonLogLevel.Info);
             var serviceProviderMock = CreateConfiguredServiceProvider(windowMock);
 
-            var testSubject = CreateTestSubject(serviceProviderMock);
+            var testSubject = CreateTestSubject(serviceProviderMock, sonarLintSettings);
 
             // Act
             testSubject.WriteLine("123");
@@ -81,7 +82,11 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Helpers
 
             if (shouldLogMessage)
             {
-                outputPane.AssertOutputStrings("[DEBUG] 123 param 1 2", "[DEBUG] 1 param 2 abc");
+                var currentThreadId = System.Threading.Thread.CurrentThread.ManagedThreadId;
+
+                outputPane.AssertOutputStrings(
+                    $"[ThreadId {currentThreadId}] [DEBUG] 123 param 1 2",
+                    $"[ThreadId {currentThreadId}] [DEBUG] 1 param 2 abc");
                 outputPane.AssertOutputStrings(2);
             }
             else
@@ -94,7 +99,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Helpers
         [DataRow(DaemonLogLevel.Info)]
         [DataRow(DaemonLogLevel.Minimal)]
         [DataRow(DaemonLogLevel.Verbose)]
-        public void LogVerbose_PrefixIsAddedIfLogLevelIsVerbose(DaemonLogLevel logLevel)
+        public void LogVerbose_PrefixIsAddedToAllMessagesIfLogLevelIsVerbose(DaemonLogLevel logLevel)
         {
             // Arrange
             var windowMock = new ConfigurableVsOutputWindow();
--- a/src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
+++ b/src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
@@ -18,6 +18,8 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
+using FluentAssertions;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
@@ -41,15 +43,13 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Helpers
         {
             // Arrange
             var windowMock = new ConfigurableVsOutputWindow();
+            var serviceProviderMock = CreateConfiguredServiceProvider(windowMock);
 
-            var serviceProviderMock = new ConfigurableServiceProvider(assertOnUnexpectedServiceRequest: true);
-            serviceProviderMock.RegisterService(typeof(SVsOutputWindow), windowMock);
-
-            var logger = new SonarLintOutputLogger(serviceProviderMock, Mock.Of<ISonarLintSettings>());
+            var testSubject = CreateTestSubject(serviceProviderMock);
 
             // Act
-            logger.WriteLine("123");
-            logger.WriteLine("abc");
+            testSubject.WriteLine("123");
+            testSubject.WriteLine("abc");
 
             // Assert
             var outputPane = windowMock.AssertPaneExists(VsShellUtils.SonarLintOutputPaneGuid);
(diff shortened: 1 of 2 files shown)


src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
```diff
@@ -41,15 +42,13 @@ public void Write_OutputsToWindow()
         {
             // Arrange
             var windowMock = new ConfigurableVsOutputWindow();
+            var serviceProviderMock = CreateConfiguredServiceProvider(windowMock);
```
duncanp-sonar, line 45: I did some minor cleanup on the existing tests to make them more similar to our current style i.e. helper methods and a variable called `testSubject`.

There is one new test below.

src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
```diff
@@ -64,35 +63,95 @@ public void LogVerbose_OutputsToWindowIfLogLevelIsVerbose(DaemonLogLevel logLeve
...
+            var windowMock = new ConfigurableVsOutputWindow();
+            var serviceProviderMock = CreateConfiguredServiceProvider(windowMock);
+
+            var sonarLintSettings = CreateSonarLintSettings(logLevel);
+
+            var testSubject = CreateTestSubject(serviceProviderMock, sonarLintSettings);
+
+            testSubject.WriteLine("create window pane");
+            var outputPane = windowMock.AssertPaneExists(VsShellUtils.SonarLintOutputPaneGuid);
+            outputPane.Reset();
```
duncanp-sonar, line 109: These tests use custom mock test implementations rather than Moq mocks.
Lines 107-109 are boilerplate to set up the custom classes.

I didn't try to replace the custom classes as part of this change.

src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
```diff
@@ -64,35 +63,95 @@ public void LogVerbose_OutputsToWindowIfLogLevelIsVerbose(DaemonLogLevel logLeve
...
+
+            if (logLevel == DaemonLogLevel.Verbose)
+            {
+                outputPane.AssertOutputStrings(3);
+
+                var currentThreadId = System.Threading.Thread.CurrentThread.ManagedThreadId;
+                outputPane.AssertOutputStrings(
+                    $"[ThreadId {currentThreadId}] writeline, no params",
+                    $"[ThreadId {currentThreadId}] writeline, with params: zzz",
+                    $"[ThreadId {currentThreadId}] [DEBUG] verbose");
```
duncanp-sonar, line 124: With this change the thread id is written before all message when in `Verbose` mode e.g.

image

src/Integration/Helpers/SonarLintOutputLogger.cs
```diff
@@ -41,21 +41,32 @@ public class SonarLintOutputLogger : ILogger
...
+            VsShellUtils.WriteToSonarLintOutputPane(this.serviceProvider, prefixedMessageFormat, args);
         }
 
         public void LogVerbose(string messageFormat, params object[] args)
         {
             if (sonarLintSettings.DaemonLogLevel == DaemonLogLevel.Verbose)
             {
                 var text = args.Length == 0 ? messageFormat : string.Format(messageFormat, args);
-                WriteLine("DEBUG: " + text);
+                WriteLine("[DEBUG] " + text);
```
duncanp-sonar, line 59: Miscellaneous: all of our other prefixes use square brackets, so I changed to use there here too.

src/Integration/Helpers/SonarLintOutputLogger.cs
```diff
@@ -41,21 +41,32 @@ public class SonarLintOutputLogger : ILogger
...
         public void WriteLine(string messageFormat, params object[] args)
         {
-            VsShellUtils.WriteToSonarLintOutputPane(this.serviceProvider, messageFormat, args);
+            var prefixedMessageFormat = AddPrefixIfVerboseLogging(messageFormat);
+            VsShellUtils.WriteToSonarLintOutputPane(this.serviceProvider, prefixedMessageFormat, args);
         }
 
         public void LogVerbose(string messageFormat, params object[] args)
         {
             if (sonarLintSettings.DaemonLogLevel == DaemonLogLevel.Verbose)
```
duncanp-sonar, line 56: The check against the log level is being done twice - once here, and once when the method calls through to `WriteLine`. However, refactoring it would have been uglier, and it isn't performance critical.

src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
```diff
@@ -64,35 +64,99 @@ public void LogVerbose_OutputsToWindowIfLogLevelIsVerbose(DaemonLogLevel logLeve
...
-            logger.LogVerbose("{0} {1} abc", 1, "param 2");
+            testSubject.LogVerbose("123 {0} {1}", "param 1", 2);
+            testSubject.LogVerbose("{0} {1} abc", 1, "param 2");
 
             // Assert
 
             if (shouldLogMessage)
             {
-                outputPane.AssertOutputStrings("DEBUG: 123 param 1 2", "DEBUG: 1 param 2 abc");
+                var currentThreadId = System.Threading.Thread.CurrentThread.ManagedThreadId;
```
duncanp-sonar, line 85: Change to existing test: the thread id is expected in verbose mode.

src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
rita-sonarsource: What is the difference between this and LogVerbose_OutputsToWindowIfLogLevelIsVerbose ?

src/Integration/Helpers/SonarLintOutputLogger.cs
rita-sonarsource: LogVerbose eventually calls WriteLine, which now calls AddPrefixIfVerboseLogging.

Took me some time to realize why we would still want [Debug]. I get it now. It's not immediately clear, but I don't have a better solution. 🤔

src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
duncanp-sonar: This test checks that the thread id is written out in verbose mode, whichever logging method you use (either `WriteLine` overload, or `LogVerbose`).

The existing test checks that `LogVerbose` only logs in `Verbose` mode.

src/Integration/Helpers/SonarLintOutputLogger.cs
duncanp-sonar: Yep; the existing code is a bit messy and seems unnecessarily complicated (I'm not sure why we have two overloads for `WriteLine`), but I wanted to minimise the edits since this is a last-minute change.

src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
```diff
@@ -41,58 +42,117 @@ public void Write_OutputsToWindow()
...
 
         [TestMethod]
-        [DataRow(DaemonLogLevel.Info, false)]
-        [DataRow(DaemonLogLevel.Minimal, false)]
-        [DataRow(DaemonLogLevel.Verbose, true)]
-        public void LogVerbose_OutputsToWindowIfLogLevelIsVerbose(DaemonLogLevel logLevel, bool shouldLogMessage)
+        [DataRow(DaemonLogLevel.Info)]
+        [DataRow(DaemonLogLevel.Minimal)]
+        [DataRow(DaemonLogLevel.Verbose)]
+        public void LogVerbose_OnlyOutputsToWindowIfLogLevelIsVerbose(DaemonLogLevel logLevel)
```
duncanp-sonar, line 63: Renamed and simplified the existing test.
The test name says it's testing `LogVerbose` and only expects output if the level is `Verbose`, so there's no need for the `shouldLogMessage` parameter.

src/Integration.UnitTests/Helpers/SonarLintOutputTests.cs
duncanp-sonar: @rita-sonarsource I've renamed and simplified the test so it only checks `WriteLine`.

PR #3529, opened 2023-01-10
New assembly for education

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3532


src/Education.UnitTests/Education.UnitTests.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net4.8</TargetFramework>
```
duncanp-sonar, line 4: We have a [common properties file](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/SonarLint.Test.props) that sets common values that should be used by all test projects. One of the values it sets is `TargetFramework`, so you don't need to set it here.

You should just need:
```
  <Import Project="..\SonarLint.Test.props" />
```

src/Education.UnitTests/Education.UnitTests.csproj
```diff
@@ -0,0 +1,19 @@
...
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.6.1" />
```
duncanp-sonar, line 10: All of our test projects except `SonarQube.Client.Tests` references version `16.5.0` of this package (which is unfortunate, since I think that is the test project you looked at). It's better to reference the same version.

src/Education/Education.csproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net4.7.2</TargetFramework>
```
duncanp-sonar, line 4: We also have a common product properties file you can include, so you don't need to set the target framework explicitly.

```
  <Import Project="..\..\SonarLint.Product.props" />
```

src/Integration.Vsix/Manifests/VS2019/source.extension.vsixmanifest
```diff
@@ -45,7 +45,7 @@
         <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="CFamily" Path="|CFamily|" />
         <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="CloudSecrets" Path="|CloudSecrets|" />
         <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="Core" Path="|Core|" />
-        <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="File" Path="secrets\SonarLint.Secrets.DotNet.dll" />
+		<Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="File" Path="secrets\SonarLint.Secrets.DotNet.dll" />
```
duncanp-sonar, line 48: Minor - inconsistent whitepace.

duncanp-sonar (commented):
Nice. See comments inline.

duncanp-sonar (approved):

PR #3542, opened 2023-01-13
Add SSE channel infrastructure

georgii-borovinskikh-sonarsource:
Prep for #3539


src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
```diff
@@ -0,0 +1,144 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using SonarLint.VisualStudio.ConnectedMode.ServerSentEvents.Issues;
+using SonarLint.VisualStudio.ConnectedMode.ServerSentEvents.TaintVulnerabilities; 
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests
+{
+    // since TaintServerEventChannel and IssueChangedServerEventChannel only differ in the value of the type parameter and in mef export contract, the tests results are the same for both
```
georgii-borovinskikh-sonarsource, line 32: I tried making the test class generic, so it could run the tests for both types, but vstest does not support that (unlike NUnit). I also couldn't find a way to get rid of the 2 implementation classes and merge them into one because of MEF

src/ConnectedMode/ServerSentEvents/IServerSentEventSource.cs
```diff
@@ -26,21 +26,22 @@ namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents
     /// <summary>
     /// Source for the server sent events about certain server side changes to the SQ/SC project
     /// </summary>
+    /// <remarks>This interface is not intended to be thread safe</remarks>
     /// <typeparam name="T">Server sent event type inherited from <see cref="IServerEvent"/></typeparam>
-    internal interface IServerSentEventSource<T> where T : IServerEvent
+    internal interface IServerSentEventSource<T> where T : class, IServerEvent
```
rita-sonarsource, line 31: is it important that it's class?

rita-sonarsource, line 31: This set of interfaces might need to be implemented in the Core assembly. The implementation of Taint will not be here but in the Taint assembly.

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
```diff
@@ -0,0 +1,86 @@
...
+    internal class ServerEventFilteredChannel<T> : IServerSentEventSource<T>, IServerSentEventSourcePublisher<T> where T : class, IServerEvent
+    {
+        private bool disposed;
+        /// <summary>
+        /// Channel that is used for storing and awaiting new items
+        /// </summary>
+        /// <remarks>
+        /// The behaviour for TryWrite depends, in some aspects, on the type of the channel we use. In particular, what we are interested in is what it returns before and after marking the channel as complete.
+        /// Because we specifically create an unbound channel with the SingleReader = true, SingleWriter = true settings, we actually get a SingleConsumerUnboundedChannel instance that has the following behaviour: TryWrite always returns true before the channel is marked as complete, and always returns false afterwards.
+        /// This behaviour allows us to be sure that the only way we can interpret the false result is that the channel was closed. \
```
rita-sonarsource, line 45: Cosmetics: redundant `\`

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// This is valuable for us because the only other way of knowing that would be calling WriteAsync and catching the exception, as the ChannelReader.Completed task is only resolved when there's no more items in the channel, and that may not happen at the same time as we called ChannelWriter.Complete.
+        /// So in short, the reason for this assumption was to simplify our channel wrapper code. 
+        /// </remarks>
+        private readonly Channel<T> channel = Channel.CreateUnbounded<T>(new UnboundedChannelOptions{SingleReader = true, SingleWriter = true});
+
+        protected ServerEventFilteredChannel(){}
```
rita-sonarsource, line 51: why is this protected? I understand that we would want the inheriting interfaces to be MEF exportable, but from the point of view of this class, it shouldn't care about that. A `new ServerEventFilteredChannel()` is a valid instance.

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            await channel.Reader.WaitToReadAsync().ConfigureAwait(false);
+            return channel.Reader.TryRead(out var item) 
+                ? item 
+                : null;
+        }
+
+        public void Publish(T serverEvent)
+        {
+            if (!channel.Writer.TryWrite(serverEvent ?? throw new ArgumentNullException(nameof(serverEvent))))
```
rita-sonarsource, line 63: I think the comment on the class is better suited to be here, where you can see the context.

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
```diff
@@ -0,0 +1,86 @@
...
+{
+    /// <summary>
+    /// Base class for channels of server events divided by topics
+    /// </summary>
+    /// <remarks>
+    /// Even though <see cref="IServerSentEventSource{T}"/> and <see cref="IServerSentEventSourcePublisher{T}"/> do not allow their own methods to be called concurrently, this class was designed to allow the calls to two different interfaces' methods to be made concurrent.
+    /// This means that calling <see cref="GetNextEventOrNullAsync"/> and <see cref="Publish"/>/<see cref="Dispose"/> at the same time is allowed, while calling <see cref="GetNextEventOrNullAsync"/> and <see cref="GetNextEventOrNullAsync"/>, <see cref="Publish"/> and <see cref="Dispose"/>, etc. is not.
+    /// </remarks>
+    /// <typeparam name="T"></typeparam>
+    internal class ServerEventFilteredChannel<T> : IServerSentEventSource<T>, IServerSentEventSourcePublisher<T> where T : class, IServerEvent
```
rita-sonarsource, line 36: This will likely also need to be moved to Core

src/ConnectedMode/ServerSentEvents/TaintVulnerabilities/TaintServerEventChannel.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents.TaintVulnerabilities
+{
+    /// <summary>
+    /// A concrete mef-exportable implementation of <see cref="ServerEventFilteredChannel{T}"/> for <see cref="ITaintServerEvent"/>
+    /// </summary>
+    [Export(typeof(ITaintServerEventSource))]
+    [Export(typeof(ITaintServerEventSourcePublisher))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class TaintServerEventChannel : ServerEventFilteredChannel<ITaintServerEvent>, ITaintServerEventSource, ITaintServerEventSourcePublisher
```
rita-sonarsource, line 32: Same comment about leaving it out of this PR, we will likely move it to Taint assembly

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
rita-sonarsource: 👍 
You can delete the comment though, as from the perspective of this class, it shouldn't know that taint and issue event exist.

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
```diff
@@ -0,0 +1,152 @@
...
+    [TestClass]
+    public class ServerEventFilteredChannelTests
+    {
+        [TestMethod]
+        public void TaintAndIssuesMefCtor_CheckIsExported()
```
rita-sonarsource, line 37: Shouldn't be here, will be in the respective test classes

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            var testSubject = CreateTestSubject();
+            var serverEvents = Enumerable.Range(0, 5).Select(_ => Mock.Of<ITaintServerEvent>()).ToList();
+            var receivedEvents = new List<ITaintServerEvent>();
+
+            foreach (var serverEvent in serverEvents)
+            {
+                testSubject.Publish(serverEvent);
+            }
+            for (var i = 0; i < serverEvents.Count; i++)
```
rita-sonarsource, line 81: Verify elegant, nice 👍

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
```diff
@@ -0,0 +1,152 @@
...
+            {
+                receivedEvents.Add(await testSubject.GetNextEventOrNullAsync());
+            }
+
+            receivedEvents.Should().BeEquivalentTo(serverEvents);
+        }
+
+        [TestMethod]
+        public async Task Get_AlreadyDisposedAndNoMoreItems_ReturnsNull()
```
rita-sonarsource, line 90: Missing a test of what happens when you call Get immediately after Dispose, without calling publish.
```
testSubject.Dispose();
var receivedEvent = await testSubject.GetNextEventOrNullAsync();
```

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            var testSubject = CreateTestSubject();
+            var serverEvent = Mock.Of<ITaintServerEvent>();
+
+            testSubject.Publish(serverEvent);
+            testSubject.Dispose();
+            var receivedEvent = await testSubject.GetNextEventOrNullAsync();
+            var receivedEvent2 = await testSubject.GetNextEventOrNullAsync();
+
+            receivedEvent.Should().BeSameAs(serverEvent);
```
rita-sonarsource, line 100: Ideally, that's _not_ what we would want. If the channel was disposed, any events that are in it are worthless. But I don't think we can clear the channel or anything. Maybe before returning we can check if it's disposed, and if so, not return the item. WDYT? not sure about the threading implications.

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
```diff
@@ -0,0 +1,152 @@
...
+            receivedEvent.Should().BeNull();
+        }
+
+        [TestMethod]
+        public void Publish_AlreadyDisposed_ThrowsObjectDisposedException()
+        {
+            var testSubject = CreateTestSubject();
+            var serverEvent = Mock.Of<ITaintServerEvent>();
+
+            testSubject.Publish(serverEvent);
```
rita-sonarsource, line 122: is the first publish relevant?

src/ConnectedMode/ServerSentEvents/IServerSentEventSource.cs
georgii-borovinskikh-sonarsource: Yes, otherwise the event source can't return null

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
```diff
@@ -0,0 +1,86 @@
...
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
+
+namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents
+{
+    /// <summary>
+    /// Base class for channels of server events divided by topics
+    /// </summary>
+    /// <remarks>
+    /// Even though <see cref="IServerSentEventSource{T}"/> and <see cref="IServerSentEventSourcePublisher{T}"/> do not allow their own methods to be called concurrently, this class was designed to allow the calls to two different interfaces' methods to be made concurrent.
+    /// This means that calling <see cref="GetNextEventOrNullAsync"/> and <see cref="Publish"/>/<see cref="Dispose"/> at the same time is allowed, while calling <see cref="GetNextEventOrNullAsync"/> and <see cref="GetNextEventOrNullAsync"/>, <see cref="Publish"/> and <see cref="Dispose"/>, etc. is not.
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: WDYM by problem?

rita-sonarsource (changes requested):
LGTM but see my comments, namely about moving it to Core and excluding taint/issue for now (so this PR would only be infra)

src/ConnectedMode/ServerSentEvents/Issues/IssueChangedServerEventChannel.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents.Issues
+{
+    /// <summary>
+    /// A concrete mef-exportable implementation of <see cref="ServerEventFilteredChannel{T}"/> for <see cref="IIssueChangedServerEvent"/>
+    /// </summary>
+    [Export(typeof(IIssueChangedServerEventSource))]
+    [Export(typeof(IIssueChangedServerEventSourcePublisher))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class IssueChangedServerEventChannel : ServerEventFilteredChannel<IIssueChangedServerEvent>, IIssueChangedServerEventSource, IIssueChangedServerEventSourcePublisher
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: What do you mean by leaving it out? Drop the inheritors but keep the base class?

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
georgii-borovinskikh-sonarsource: The point was to also test that they don't interfere with each other

src/ConnectedMode/ServerSentEvents/IServerSentEventSource.cs
rita-sonarsource: https://media1.giphy.com/media/3og0IPBdowkBBNtKTK/giphy.gif

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
georgii-borovinskikh-sonarsource: The bool value is atomic so shouldn't be a problem since we don't need to stop exactly after dispose, but the sooner the better. Btw, this one is a bit unclear for me, should I make it volatile? I always struggled with that keyword 🤔

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
rita-sonarsource: Yep, definitely make it public IMO. It's a standalone class, no need for abstract or protected ctor.

src/ConnectedMode/ServerSentEvents/Issues/IssueChangedServerEventChannel.cs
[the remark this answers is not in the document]
rita-sonarsource: Yep, out of this PR

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
rita-sonarsource: That will be then in their own test classes.

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Discussed in person, will make the comment more clear

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
georgii-borovinskikh-sonarsource: Yes, to simulate that it also throws even if there are some events in the channel

rita-sonarsource: Maybe use TestCases then, `bool isEmptyChannel`

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
duncanp-sonar: @georgii-borovinskikh-sonarsource it's not clear why this conversation was marked as resolved. Can you explain?

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
duncanp-sonar: @georgii-borovinskikh-sonarsource are you planning to move the comment as @rita-sonarsource requested?

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
georgii-borovinskikh-sonarsource: I added it, just forgot to reply to the comment

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
georgii-borovinskikh-sonarsource: I made the requested changes. My bad for not replying

src/Core.UnitTests/ServerSentEvents/ServerEventFilteredChannelTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            testSubject.Dispose();
+            var receivedEvent = await testSubject.GetNextEventOrNullAsync();
+            var receivedEvent2 = await testSubject.GetNextEventOrNullAsync();
+
+            receivedEvent.Should().BeSameAs(isEmptyChannel ? null : serverEvent);
+            receivedEvent2.Should().BeNull();
+        }
+
+        [TestMethod]
+        public async Task Get_AfterDispose_ReturnsNull()
```
georgii-borovinskikh-sonarsource, line 101: @duncanp-sonar JFYI it's here

src/ConnectedMode/ServerSentEvents/ServerEventFilteredChannel.cs
georgii-borovinskikh-sonarsource: Yes, but I first need to talk to her because I'm a bit confused as to where it's supposed to be moved

src/ConnectedMode.UnitTests/ServerEventFilteredChannelTests.cs
duncanp-sonar: @georgii-borovinskikh-sonarsource delete the comment?

duncanp-sonar (approved):

PR #3550, opened 2023-01-21
Remove redundant CFamily build-time code

duncanp-sonar:
- replaced 20/12/2022 in commit 30fb3b64


build/DownloadSonarPlugin/Common.cs
```diff
@@ -127,34 +127,6 @@ public static void UnzipJar(string jarFilePath, string destinationFolder, TaskLo
             LogElapsedTime("Unzipped jar file", timer, logger);
         }
 
-        public static void UncompressAndUnzipTgx(string tarFilePath, string destinationFolder, TaskLoggingHelper logger)
```
duncanp-sonar, line 130: These methods were only used when processing the CFamily plugin, the processing of which is now handled by the `CFamilyJarPreProcessor` tool.

bigfluffycookie (approved):
Looks good :)

duncanp-sonar:
FYI I manually check that the VSIX built from this PR still contains all of the expected CFamily files.

PR #3559, opened 2023-01-24
Add a method GetToolWindow in ToolWindowService.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3560

### DIFF
--- a/src/Core/IToolWindowService.cs
+++ b/src/Core/IToolWindowService.cs
@@ -43,5 +43,7 @@ namespace SonarLint.VisualStudio.Core
         /// </para>
         /// </remarks>
         void EnsureToolWindowExists(Guid toolWindowId);
+
+        T GetToolWindow<T>() where T : class;
     }
 }
--- a/src/Core/IToolWindowService.cs
+++ b/src/Core/IToolWindowService.cs
@@ -44,6 +44,10 @@ namespace SonarLint.VisualStudio.Core
         /// </remarks>
         void EnsureToolWindowExists(Guid toolWindowId);
 
+        /// <summary>
+        /// Returns a tool window of the specified type, or null if one could not be found.
+        /// </summary>
+        /// <typeparam name="T">The type of tool window. Should be a subclass of <see cref="Microsoft.VisualStudio.Shell.ToolWindowPane"/></typeparam>
         T GetToolWindow<T>() where T : class;
     }
 }
--- a/src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
+++ b/src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
@@ -19,6 +19,7 @@
  */
 
 using System;
+using System.Runtime.InteropServices;
 using FluentAssertions;
 using Microsoft.VisualStudio;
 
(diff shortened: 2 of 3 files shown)


src/Core/IToolWindowService.cs
```diff
@@ -43,5 +43,7 @@ public interface IToolWindowService
         /// </para>
         /// </remarks>
         void EnsureToolWindowExists(Guid toolWindowId);
+
```
bigfluffycookie, line 46: Should I have added a comment? (Just because the other methods in the class already have them)

src/Integration/Helpers/ToolWIndowService.cs
```diff
@@ -51,6 +51,14 @@ public void EnsureToolWindowExists(Guid toolWindowId)
             GetOrCreateWindowFrame(toolWindowId);
         }
```
bigfluffycookie, line 53: Regarding testing, I am unsure how to mock a toolwindowpane. Is that something we do?

src/Core/IToolWindowService.cs
duncanp-sonar: Yes - something like:

```
        /// <summary>
        /// Returns a tool window of the specified type, or null if one could not be found.
        /// </summary>
        /// <typeparam name="T">The type of tool window. Should be a subclass of <see cref="Microsoft.VisualStudio.Shell.ToolWindowPane"/></typeparam>
```

In general, all interfaces should have comments, on the interface and on the methods.

We don't generally add comments on the classes that implement the interface, unless there is something implementation-specific that we want to capture.

src/Core/IToolWindowService.cs
```diff
@@ -43,5 +43,7 @@ public interface IToolWindowService
...
+        T GetToolWindow<T>() where T : class;
```
duncanp-sonar, line 47: FYI it's possible to add specific type or interface to the `where` clause e.g. 
```
T GetToolWindow<T>() where T: class, IToolWindowService
```

However, we can't do that here unless we add a reference to the VS assemblies, which is what we were trying to avoid.

duncanp-sonar (commented):
Looks good, but it needs a unit test. Let me know if you need a hand.

src/Integration/Helpers/ToolWIndowService.cs
duncanp-sonar: @bigfluffycookie apologies, I missed this comment earlier.
Moq can mock classes as well as interfaces, as long as the class is not `sealed`. However, we'd generally try to mock an interface rather than a class - that way, we know we have complete control over everything the mock is returning.
If we are mocking a class, we can only override methods that the base class has marked as `virtual`. That might not give us the level of control we need for a test.

Fortunately, almost all of the public extension points in VS are interface-based, so we can almost always code against interfaces rather than classes. In this case, the `IToolWindowService` references `IVsWindowFrame` rather than `ToolWindowPane`, so we can mock that, which is what the existing `ToolWindowServiceTests` do (`ToolWindowPane` implements `IVsWindowFrame`).

In this case, it might be simpler just to create a real dummy class that implements `IVsWindowFrame` and has a `Guid` attribute on it e.g.

```
        [Guid("1D7B3C6A-BAB5-4A0D-8583-BCA4ECAEE649")]
        private class DummyToolWindowPaneWithGuid : IVsWindowFrame
        {
            public int Show() => throw new NotImplementedException();

            public int Hide() => throw new NotImplementedException();
            
            public int IsVisible() => throw new NotImplementedException();

            public int ShowNoActivate() => throw new NotImplementedException();
            "
            public int CloseFrame(uint grfSaveOptions) => throw new NotImplementedException();

            public int SetFramePos(VSSETFRAMEPOS dwSFP, ref Guid rguidRelativeTo, int x, int y, int cx, int cy) => throw new NotImplementedException();

            public int GetFramePos(VSSETFRAMEPOS[] pdwSFP, out Guid pguidRelativeTo, out int px, out int py, out int pcx, out int pcy) => throw new NotImplementedException();

            public int GetProperty(int propid, out object pvar) => throw new NotImplementedException();

            public int SetProperty(int propid, object var) => throw new NotImplementedException();

            public int GetGuidProperty(int propid, out Guid pguid) => throw new NotImplementedException();

            public int SetGuidProperty(int propid, ref Guid rguid) => throw new NotImplementedException();

            public int QueryViewInterface(ref Guid riid, out IntPtr ppv) => throw new NotImplementedException();

            public int IsOnScreen(out int pfOnScreen) => throw new NotImplementedException();
        }
```

It's might be possible to mock the `GuidAttribute` part using Moq, but sometimes it just easier to create a "real" class.

src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
```diff
@@ -128,6 +129,27 @@ public void EnsureToolWindowExists_VsServiceCalled(int vsServiceResult)
             frameMock.Invocations.Should().BeEmpty();
         }
 
+        [TestMethod]
```
bigfluffycookie, line 132: Something about this is wrong I think. it works but I am not sure its supposed to.

The windowPane object at some point turns into a "ToolWindowPaneProxy" which is why the assertion does not test against the type "ToolWindowPane".

duncanp-sonar (approved):

duncanp-sonar:
@bigfluffycookie if you prefer, merge the PR to `branch-rules` and I'll fix the conflicts (it might be quicker since I fixed the other conflicts)

Issue #3569, opened 2023-01-25
Add IRuleHelpXamlBuilder interface and dummy Implementation


bigfluffycookie:
Will be responsible for converting html to a FlowDocument

PR #3571, opened 2023-01-25
Add IRuleHelpXamlBuilder

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3569


src/Education/IRuleHelpXamlBuilder.cs
```diff
@@ -0,0 +1,58 @@
+﻿/*
```
bigfluffycookie, line 1: I am thinking writing tests for this doesn't make much sense but please correct me if I am wrong.

src/Education/IRuleHelpXamlBuilder.cs
```diff
@@ -0,0 +1,58 @@
...
+    internal class RuleHelpXamlBuilder : IRuleHelpXamlBuilder
+    {
+        public FlowDocument CreateFromHTML(IRuleHelp ruleHelp)
+        {
+            // TODO: Convert HTML to FlowChart
+
+            var placeholder = $@"
+                <FlowDocument xml:space=""preserve"" xmlns=""http://schemas.microsoft.com/winfx/2006/xaml/presentation"">
+                    <Paragraph>Rule: {ruleHelp.RuleKey}</Paragraph>
+                    <Paragraph>Language: {ruleHelp.Language}</Paragraph>
```
duncanp-sonar, line 43: Should perhaps be `ruleHelp.Language.Name` rather than just `ruleHelp.Language`.

`Language` is a class in `Core` with multiple properties, and it doesn't override `ToString` so the default text probably isn't what we want.

src/Education/IRuleHelpXamlBuilder.cs
```diff
@@ -0,0 +1,58 @@
...
+                    <Paragraph>TODO: rule description</Paragraph>
+                    <Paragraph FontSize=""20pt"">Noncompliant Code Example</Paragraph>
+                    <Paragraph FontSize=""8pt"" TextAlignment=""Left"" FontFamily=""Courier New"">non-compliant code</Paragraph>
+                    <Paragraph FontSize=""20pt"">Compliant Solution</Paragraph>
+                    <Paragraph FontSize=""8pt"" TextAlignment=""Left"" FontFamily=""Courier New"">compliant code</Paragraph>
+                </FlowDocument>";
+
+            var stringReader = new StringReader(placeholder);
+            var xamlReader = XmlReader.Create(stringReader);
```
duncanp-sonar, line 52: `StringReader` is disposable so it should be used inside a `using` statement (it's a good habit to get into, even though this is temporary boilerplate code).

For some reason `XamlReader` is not disposable.

src/Education/IRuleHelpXamlBuilder.cs
```diff
@@ -0,0 +1,58 @@
...
+            var flowDocument = (FlowDocument)XamlReader.Load(xamlReader);
```
duncanp-sonar, line 53: FYI we'll need to think about threading at some point - I'm guessing we might have to create the XAML on a UI thread.
We don't need to worry about it yet, though.

duncanp-sonar (commented):
Looks good. A couple of minor comments inline.

src/Education/IRuleHelpXamlBuilder.cs
duncanp-sonar: I agree, it doesn't make any sense yet.

PR #3575, opened 2023-01-25
Implementation for SSE session manager and event pump, interface declaration for event filter

georgii-borovinskikh-sonarsource:


georgii-borovinskikh-sonarsource:
Recreatng the PR from scratch because of an error during creation

PR #3576, opened 2023-01-25
Implementation for SSE session manager and event pump, interface declaration for event filter 

georgii-borovinskikh-sonarsource:


src/ConnectedMode.UnitTests/SeverSentEvents/ServerSentEventPumpTests.cs
```diff
@@ -0,0 +1,129 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using SonarLint.VisualStudio.ConnectedMode.ServerSentEvents;
+using SonarLint.VisualStudio.Core.ServerSentEvents.Issues;
+using SonarLint.VisualStudio.Core.ServerSentEvents.TaintVulnerabilities;
+using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarQube.Client.Models.ServerSentEvents;
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.SeverSentEvents
```
rita-sonarsource, line 32: Typo in folder and namespace name ("sever")

src/Core/ServerSentEvents/ServerEventFilteredChannel.cs
```diff
@@ -31,6 +31,7 @@ namespace SonarLint.VisualStudio.Core.ServerSentEvents
     /// <remarks>
     /// Even though <see cref="IServerSentEventSource{T}"/> and <see cref="IServerSentEventSourcePublisher{T}"/> do not allow their own methods to be called from different threads, this class was designed to allow the calls to two different interfaces' methods to be made from different threads at the same time.
     /// This means that calling from one thread <see cref="GetNextEventOrNullAsync"/> and calling <see cref="Publish"/> or <see cref="Dispose"/> from a different thread at the same time is allowed, while calling methods of the same interface from different threads is not.
+    /// NOTE: there's an exception to this rule that was made to simplify the implementation of the events pump that permits calling Dispose and Publish concurrently, but results in an exception. For more info see <see cref="IServerSentEventSourcePublisher{T}"/>
     /// </remarks>
     /// <typeparam name="T"></typeparam>
     public class ServerEventFilteredChannel<T> : IServerSentEventSource<T>, IServerSentEventSourcePublisher<T> where T : class, IServerEvent
```
rita-sonarsource, line 37: I'm not sure that the name Filtered is correct here. It doesn't do any filtering. Perhaps just `ServerEventsChannel` ?

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -18,29 +18,109 @@
...
     {
     }
 
     [Export(typeof(IServerSentEventSessionManager))]
     [PartCreationPolicy(CreationPolicy.Shared)]
-    internal class ServerSentEventSessionManager : IServerSentEventSessionManager
+    internal sealed class ServerSentEventSessionManager : IServerSentEventSessionManager
     {
+        private readonly IActiveSolutionBoundTracker activeSolutionBoundTracker;
+        private readonly IIssueChangedServerEventSourcePublisher issueChangedServerEventSourcePublisher;
```
rita-sonarsource, line 42: Cosmetics: would be more readable to place the publishers next to each other

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -18,29 +18,109 @@
...
+            }
+
+            activeSolutionBoundTracker.SolutionBindingChanged -= OnSolutionChanged;
+            sessionTokenSource?.Cancel();
+            issueChangedServerEventSourcePublisher?.Dispose();
+            taintServerEventSourcePublisher?.Dispose();
+            disposed = true;
+        }
+
+        internal /*for testing*/ void OnSolutionChanged(object sender, ActiveSolutionBindingEventArgs activeSolutionBindingEventArgs)
```
rita-sonarsource, line 82: Any reason why this is internal, instead of raising the event in the test? Moq has a Raise event function, see NotificationServiceTests for example

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -18,29 +18,109 @@
...
+        {
+            var bindingConfiguration = activeSolutionBindingEventArgs.Configuration;
+            var isOpen = !bindingConfiguration.Equals(BindingConfiguration.Standalone);
```
rita-sonarsource, line 85: isInConnectedMode is better in terms of terminology

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -18,29 +18,109 @@
...
+
+            if (isOpen)
+            {
+                var projectKey = bindingConfiguration.Project.ProjectKey;
+                Connect(projectKey);
```
rita-sonarsource, line 90: Connect is slightly misleading, since there is a Connect method on sonarQubeClient but that's not the method we're calling. Perhaps `InitializeSession` and `EndSession` instead of Disconnect?

src/ConnectedMode/ServerSentEvents/ServerSentEventsFilter.cs
```diff
@@ -0,0 +1,29 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
+
+namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents
+{
+    internal interface IServerSentEventsFilter
+    {
+        IServerEvent GetFilteredEventOrNull(IServerEvent serverEvent);
```
rita-sonarsource, line 27: What is the purpose of this interface? it receives the event, so does it do somewhere a switch-case/if-else? if so, this is a repetition, the split should be done in a single place, before where send to the corresponding publisher.

src/ConnectedMode/ServerSentEvents/ServerSentEventPump.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        private readonly IServerSentEventsFilter serverSentEventsFilter;
+
+        [ImportingConstructor]
+        public ServerSentEventPump(IServerSentEventsFilter serverSentEventsFilter)
+        {
+            this.serverSentEventsFilter = serverSentEventsFilter;
+        }
+
+        public async Task PumpAllAsync(IServerSentEventsSession session,
```
rita-sonarsource, line 50: Why doesn't it receive the publishers in the CTOR? they are always the same.

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -18,29 +18,109 @@
...
+                Disconnect();
+            }
+        }
+
+        private void Connect(string projectKey)
+        {
+            sessionTokenSource = new CancellationTokenSource();
+
+            // fire and forget
+            threadHandling.RunOnBackgroundThread(async () =>
```
rita-sonarsource, line 103: We should handle the case of Connect and Disconnect being called multiple times.

rita-sonarsource (changes requested):
See my comments. I think that the Pump interface has some problems with the way it is currently hooked into the session manager and the filterer.

src/ConnectedMode/ServerSentEvents/ServerSentEventsFilter.cs
georgii-borovinskikh-sonarsource: I mainly introduced this to get unblocked because I didn't have the models to actually implement the filtering

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
georgii-borovinskikh-sonarsource: I decided it would be simpler to avoid triggering the events and just call the method manually

src/Core/ServerSentEvents/ServerEventFilteredChannel.cs
georgii-borovinskikh-sonarsource: Well, it's filtered because it has issues that have been filtered, not the ones it needs to filter. But I think removing the word from the name is also fine

src/ConnectedMode/ServerSentEvents/ServerSentEventPump.cs
georgii-borovinskikh-sonarsource: it's a bit of a tricky situation. the channels are disposables, but the pump isn't. I decided to limit the ownership scope to reduce the confusion, but maybe I didn't achieve it

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
georgii-borovinskikh-sonarsource: What would be the preferred way? Throw an error or just discard the previous connection?

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
rita-sonarsource: I'm not a big fan of internal testing methods, it always feels hacky to me. Raising an event is cleaner IMO and can be easily done with a helper method in the test. WDYT?

src/ConnectedMode/ServerSentEvents/ServerSentEventPump.cs
rita-sonarsource: The channels are not disposable though, they're singletons. We just no longer pump data into them if the connect ended.

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
rita-sonarsource: I think discard the previous would be better.

src/ConnectedMode/ServerSentEvents/ServerSentEventPump.cs
georgii-borovinskikh-sonarsource: But they do implement IDisposable and throw after the disposal. It only happens on shutdown, but still.

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
georgii-borovinskikh-sonarsource: OK I'll change it

src/ConnectedMode/ServerSentEvents/ServerSentEventsFilter.cs
georgii-borovinskikh-sonarsource: I will have to implement this in a separate PR right after this one.

src/ConnectedMode/ServerSentEvents/ServerSentEventPump.cs
georgii-borovinskikh-sonarsource: moved to ctor

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -18,29 +18,104 @@
...
+        private CancellationTokenSource sessionTokenSource;
+
         [ImportingConstructor]
         public ServerSentEventSessionManager(
             IActiveSolutionBoundTracker activeSolutionBoundTracker,
             ISonarQubeService sonarQubeClient,
+            IServerSentEventPump eventPump,
             IIssueChangedServerEventSourcePublisher issueChangedServerEventSourcePublisher,
-            ITaintServerEventSourcePublisher taintServerEventSourcePublisher)
+            ITaintServerEventSourcePublisher taintServerEventSourcePublisher,
```
rita-sonarsource, line 57: Need to clean up dependencies, they're not used anymore.
This is also why I think using a generic test suite is less optimal, you would've spotted this if you had to explicitly declare them in the tests.

src/ConnectedMode/ServerSentEvents/ServerSentEventsFilter.cs
rita-sonarsource: Might be better to just remove it now, since this is a new interface and we know it won't be used. So no point in committing code that won't be used.

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
georgii-borovinskikh-sonarsource: They are used in Dispose method

src/ConnectedMode/ServerSentEvents/ServerSentEventsFilter.cs
georgii-borovinskikh-sonarsource: It will break the tests

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -18,29 +18,104 @@
...
+            if (serverSentEventsSession == null)
+            {
+                return;
+            }
+
+            await eventPump.PumpAllAsync(serverSentEventsSession);
+        }
+
+        private void EndSession()
         {
```
rita-sonarsource, line 116: Is this enough?
What about the task from the previous session? `eventPump.PumpAllAsync` will be a thread that never finishes due to a while(true) loop. We need a way to terminate that thread (perhaps `while(!cancellationRequested)`)

georgii-borovinskikh-sonarsource: the session object should throw on cancellation and cause the loop to exit

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
rita-sonarsource: It doesn't make sense to receive dependencies just to dispose them

src/ConnectedMode/ServerSentEvents/ServerSentEventsFilter.cs
rita-sonarsource: Then fix them :P

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
georgii-borovinskikh-sonarsource: well, they need to be Disposed somewhere and I see this as the better place to do it

src/ConnectedMode/ServerSentEvents/ServerSentEventsFilter.cs
rita-sonarsource: To elaborate: otherwise I'm reviewing code that is irrelevant.

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
rita-sonarsource: Shouldn't the eventPump be the one disposing them?

georgii-borovinskikh-sonarsource: Hmmm yes, I'll change it

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
rita-sonarsource: Hmm. We're making an assumption here but I don't have a better suggestion for now.

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -18,29 +18,104 @@
...
+            taintServerEventSourcePublisher?.Dispose();
+            disposed = true;
+        }
+
+        private void OnSolutionChanged(object sender, ActiveSolutionBindingEventArgs activeSolutionBindingEventArgs)
+        {
+            var bindingConfiguration = activeSolutionBindingEventArgs.Configuration;
+            var isInConnectedMode = !bindingConfiguration.Equals(BindingConfiguration.Standalone);
+
+            EndSession();
```
rita-sonarsource, line 88: I think there might be a race condition problem here. 
1) User opens a solution, sessionTokenSource=null, InitializeSessionAsync sets sessionTokenSource=value
2) User closes the solution, quickly opens another
==> 
sessionTokenSource=previous initialize
session is cancelled, new InitializeSessionAsync is scheduled but not executed yet
3) User closes the solution
==> sessionTokenSource= null because InitializeSessionAsync didn't run it, and no session is cancelled. A thread remains.

We need to handle this with closure, storing the session token created by InitializeSessionAsync

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
georgii-borovinskikh-sonarsource: W/o this assumption, we also couldn't assume that getNextEvent would not block forever

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
georgii-borovinskikh-sonarsource: I think moving the sessionTokenSource initialization to event handler method to run synchronously would fix this, no?

src/ConnectedMode.UnitTests/ServerSentEvents/ServerSentEventPumpTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        public void Dispose_DisposesPublishers()
+        {
+            testSubject.Dispose();
+
+            issueChangedPublisherMock.Verify(ch => ch.Dispose(), Times.Once);
+            taintPublisherMock.Verify(ch => ch.Dispose(), Times.Once);
+        }
+
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
```
rita-sonarsource, line 62: Conventions: we try to keep the MEF test first in the class, with Dispose being last.

src/ConnectedMode.UnitTests/ServerSentEvents/ServerSentEventPumpTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            issueChangedPublisherMock.Setup(publisher => publisher.Publish(It.IsAny<IIssueChangedServerEvent>()))
+                .Throws(new ObjectDisposedException(string.Empty));
+
+            await testSubject.PumpAllAsync(serverSentEventsSessionMock.Object);
+
+            issueChangedPublisherMock.Verify(publisher => publisher.Publish(It.IsAny<IIssueChangedServerEvent>()), Times.Once);
+            taintPublisherMock.Verify(publisher => publisher.Publish(It.IsAny<ITaintServerEvent>()), Times.Never);
+        }
+
+        private void SetUpEventsSequenceThatFinishes(params IServerEvent[] inputSequence)
```
rita-sonarsource, line 109: Perhaps CreateSessionWithEvents ?

src/ConnectedMode.UnitTests/ServerSentEvents/ServerSentEventPumpTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            var sequenceSetUp = serverSentEventsSessionMock.SetupSequence(x => x.ReadAsync());
+            sequenceSetUp = inputSequence.Aggregate(sequenceSetUp, (current, serverEvent) => current.ReturnsAsync(serverEvent));
```
rita-sonarsource, line 112: Elegant 👍

src/ConnectedMode.UnitTests/ServerSentEvents/ServerSentEventPumpTests.cs
```diff
@@ -0,0 +1,116 @@
...
+                Mock.Of<ITaintVulnerabilityClosedServerEvent>(),
+                Mock.Of<ITaintVulnerabilityClosedServerEvent>(),
+                Mock.Of<IIssueChangedServerEvent>(),
+                Mock.Of<ITaintVulnerabilityRaisedServerEvent>(),
+            };
+            SetUpEventsSequenceThatFinishes(inputSequence);
+
+            await testSubject.PumpAllAsync(serverSentEventsSessionMock.Object);
+            
+            issueChangedPublisherMock.Invocations
```
rita-sonarsource, line 86: I think it would be more readable to specify exactly what you're expecting.
It's better to be explicit in the tests.
```
issueChangedPublisherMock.Invocations.Count.Should().Be(3)
issueChangedPublisherMock.Invocations[0].Arguments[0].Should().Be(inputSequence[0])
```

src/ConnectedMode.UnitTests/ServerSentEvents/ServerSentEventPumpTests.cs
```diff
@@ -0,0 +1,116 @@
...
+                .Should()
+                .BeEquivalentTo(inputSequence.Where(issuesEvent => issuesEvent is IIssueChangedServerEvent));
+            taintPublisherMock.Invocations
+                .Select(call => call.Arguments.First() as ITaintServerEvent)
+                .Should()
+                .BeEquivalentTo(inputSequence.Where(taintEvent => taintEvent is ITaintServerEvent));
+        }
+
+        [TestMethod]
+        public async Task PumpAllAsync_WhenPublisherDisposed_Finishes()
```
rita-sonarsource, line 97: This test doesn't logically make sense now that the pump is the one disposing.
Instead, there should be a test for what happens if you call PumpAll after disposing the pump.

src/ConnectedMode/ConnectedMode.csproj
```diff
@@ -1,13 +1,16 @@
...
+  <Import Project="..\SonarLint.VsPkgRefs.props" />
 
   <PropertyGroup>
     <ProjectGuid>0BE551DB-3C46-42A5-BB38-DA80E83F8ABD</ProjectGuid>
     <AssemblyName>SonarLint.VisualStudio.ConnectedMode</AssemblyName>
     <RootNamespace>SonarLint.VisualStudio.ConnectedMode</RootNamespace>
 
     <RequiresLibGit>true</RequiresLibGit>
+
+    <LangVersion>8</LangVersion>
```
duncanp-sonar, line 13: Some newer language features are implemented solely by the compiler, but some also rely on a particular target framework. For product code, we don't want to be in the situation where we accidentally use a language feature that relies on a .NET runtime version that's newer than the one used by VS2019.3.

To be on the safe side, we only set the language version for test projects.

Which C#8 languages features are you using?

src/ConnectedMode/ServerSentEvents/ServerSentEventPump.cs
```diff
@@ -0,0 +1,87 @@
...
+using SonarLint.VisualStudio.Core.ServerSentEvents.TaintVulnerabilities;
+using SonarQube.Client.Models.ServerSentEvents;
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
+using System;
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+
+namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents
+{
+    internal interface IServerSentEventPump : IDisposable
```
duncanp-sonar, line 31: Missing doc comments

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -18,29 +18,95 @@
...
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarQube.Client;
 
 namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents
 {
-    internal interface IServerSentEventSessionManager
+    internal interface IServerSentEventSessionManager : IDisposable
```
duncanp-sonar, line 32: Missing doc comments

src/ConnectedMode/ServerSentEvents/SSESessionFactory.cs
```diff
@@ -0,0 +1,183 @@
...
+    /// <summary>
+    /// Factory for <see cref="ISSESession"/>. Responsible for disposing EventSourcePublishers
+    /// </summary>
+    internal interface ISSESessionFactory : IDisposable
+    {
+        ISSESession Create(string projectKey);
+    }
+
+    /// <summary>
+    /// Represents the session entity, that is responsible for dealing with SQ Client's SSE Session reader
```
rita-sonarsource, line 42: Minor: why not a cref?

src/ConnectedMode/ServerSentEvents/SSESessionFactory.cs
```diff
@@ -0,0 +1,185 @@
...
+                await threadHandling.SwitchToBackgroundThread();
+
+                var sseStream = await sonarQubeService.CreateServerSentEventsStream(projectKey, sessionTokenSource.Token);
+
+                if (sseStream == null)
+                {
+                    return;
+                }
+
+                sseStream.BeginListening().Forget();
```
rita-sonarsource, line 143: FYI I added this following the merge conflict. I _think_ we might need to run it on another thread. I'm not sure. It uses the underlying network ReadLineAsync, and I _think_ it shouldn't block (or it won't if we use ReadLineAsync().ConfigureAwait(false)).

src/ConnectedMode/ConnectedMode.csproj
```diff
@@ -1,13 +1,16 @@
...
+  <Import Project="..\SonarLint.VsPkgRefs.props" />
 
   <PropertyGroup>
     <ProjectGuid>0BE551DB-3C46-42A5-BB38-DA80E83F8ABD</ProjectGuid>
     <AssemblyName>SonarLint.VisualStudio.ConnectedMode</AssemblyName>
     <RootNamespace>SonarLint.VisualStudio.ConnectedMode</RootNamespace>
 
     <RequiresLibGit>true</RequiresLibGit>
+
+    <LangVersion>latest</LangVersion>
```
rita-sonarsource, line 13: This is a bit problematic. We've been avoiding using it since it's possible that some of the changes are not syntactic sugar, which will then break on the client's machine. Is this for the `using namespace`? If so, let's revert it as it's not worth the risk (sadly).

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionFactoryTests.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
```
rita-sonarsource, line 1: copyright

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,230 @@
...
+            MefTestHelpers.CreateExport<IActiveSolutionBoundTracker>(),
+            MefTestHelpers.CreateExport<ISSESessionFactory>());
+    }
+
+    [TestMethod]
+    public void Ctor_SubscribesToSolutionEvents()
+    {
+        var solutionTrackerMock = new Mock<IActiveSolutionBoundTracker>();
+
+        var testSubject = new SSESessionManager(solutionTrackerMock.Object, Mock.Of<ISSESessionFactory>());
```
rita-sonarsource, line 45: testSubject unused

rita-sonarsource (changes requested):
@georgii-borovinskikh-sonarsource LGTM but see my comments regarding the TestScope and the new `BeginListening().Forget()` method on SSEStream

src/ConnectedMode/ConnectedMode.csproj
```diff
@@ -1,13 +1,16 @@
 <Project Sdk="Microsoft.NET.Sdk">
   
   <Import Project="..\SonarLint.Product.props" />
+  <Import Project="..\SonarLint.VsPkgRefs.props" />
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Yes, here's what Duncan said as well:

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: ```I think it's ok to add a reference. If I remember correctly the Forget method is actually a no-op, so we could easily add our own version in Core. However, we'll probably need to reference some VS assemblies for connected mode at some point.```

src/ConnectedMode/ConnectedMode.csproj
georgii-borovinskikh-sonarsource: I will try to set it to 7 and see if it compiles. Maybe I needed it for some stuff that I ended up not using

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        solutionTrackerMock.VerifyAdd(
+            solutionTracker => solutionTracker.SolutionBindingChanged +=
+                It.IsAny<EventHandler<ActiveSolutionBindingEventArgs>>(), Times.Once);
+    }
+
+    [DataTestMethod]
+    [DataRow(true)]
+    [DataRow(false)]
+    public void OnSolutionChanged_CorrectlyIdentifiesOpenAndClose(bool isOpen)
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: But the copy paste is going to end up in almost every test 😞 Can you elaborate on why it's hard to understand? Maybe we can find a compromise

src/ConnectedMode/ServerSentEvents/SSESessionFactory.cs
georgii-borovinskikh-sonarsource: I didn't have the interface at that time. But now that you pulled it I can fix this

src/ConnectedMode/ConnectedMode.csproj
rita-sonarsource: See Duncan's comment regarding this: https://github.com/SonarSource/sonarlint-visualstudio/pull/3576#discussion_r1091735735

georgii-borovinskikh-sonarsource: Yes, but that's why I changed to latest. Did I misunderstood him then?

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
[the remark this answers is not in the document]
rita-sonarsource: Couple of things:
* OpenNewProject and CloseProject are not the right names for what's happening. The event is bindingChanged and it's either Standalone or Connected, there is no concept of project (nor a solution). The session manager only knows that the binding changed and so the tests need to speak in the same terminology.
* Names such as SetUpMocksDispose and SetUpMockConnection -- it's not clear what they're doing, especially when they work with sequences. 
* `sessionFactory.Create(It.IsAny<string>())` it is always better to set up with specific arguments. It.IsAny is used here because this is a generic function that receives the sequence, but it makes it more complicated for the reader to understand, versus is seeing a specific set up in the code and then immediately the verification. 

I think it's ok that there is some duplication in the unit tests because it increases the readability of every test. When there are so few functions that encapsulate everything, it is very difficult to understand what actually is being tested.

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: IMO points 1 and 2 are just bad naming and point 3 can be fixed as well. "When there are so few functions that encapsulate everything" - but if you get rid of the class, you either keep the same functions but move them to the test class or get rid of them and get a lot of duplication. "it is very difficult to understand what actually is being tested." - it's in the name of the test method + verifications are exactly the same except the `testScope.*` prefix

src/ConnectedMode/ConnectedMode.csproj
rita-sonarsource: Yes, he meant that we shouldn't set this at all (only in test projects).

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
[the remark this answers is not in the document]
rita-sonarsource: Ok, try to rename and let's see. IMO the code I sent you was more explicit and easier to understand than the original.

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        solutionTrackerMock.VerifyAdd(
+            solutionTracker => solutionTracker.SolutionBindingChanged +=
+                It.IsAny<EventHandler<ActiveSolutionBindingEventArgs>>(), Times.Once);
+    }
+
+    [DataTestMethod]
+    [DataRow(true)]
+    [DataRow(false)]
+    public void OnSolutionChanged_CorrectlyIdentifiesOpenAndClose(bool switchToConnectedMode)
```
rita-sonarsource, line 55: I don't understand what the test's name was supposed to be

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        var testScope = new TestScope();
+
+
+        if (switchToConnectedMode)
```
rita-sonarsource, line 60: The word "switch" implies that there was a previous state and now you're switching to the new one. This is not the case. It's simply "isInConnectedMode", with the methods being RaiseBindingChangedToConnectedMode and RaiseBindingChangedToStandaloneMode

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
georgii-borovinskikh-sonarsource: yes, missed this. should be connected vs standalone

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+            testScope.SetUpSSEFactoryToReturnNoOpSSESession(projectKey);
+            testScope.SwitchToConnectedMode(projectKey);
+        }
+        else
+        {
+            testScope.SwitchToStandaloneMode();
+        }
+
+        testScope.SSESessionFactoryMock.Verify(factory => factory.Create(It.IsAny<string>()),
+            Times.Exactly(switchToConnectedMode ? 1 : 0));
```
rita-sonarsource, line 72: I think there is enough difference in the test that the testcases is not worth it. Just split it to 2, it will be more readable.

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    [TestMethod]
+    public void OnSolutionChanged_WhenChangesFromClosedToOpen_CreatesSessionAndLaunchesIt()
```
rita-sonarsource, line 76: Rename, no close and no open. It's connected and standalone.

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        var testScope = new TestScope();
+        testScope.SwitchToStandaloneMode();
```
rita-sonarsource, line 79: Technically this test is exactly the same as the previous one, as starting from nothing is same as starting from standalone.

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+        var sessionMock = testScope.SetUpSSEFactoryToReturnNoOpSSESession(projectKey);
+
+        testScope.SwitchToConnectedMode(projectKey);
+
+        testScope.SSESessionFactoryMock.Verify(factory => factory.Create(It.IsAny<string>()), Times.Once);
+        sessionMock.Verify(session => session.PumpAllAsync(), Times.Once);
+    }
+
+    [TestMethod]
+    public void OnSolutionChanged_WhenChangesFromOpenToClosed_CancelsSession()
```
rita-sonarsource, line 90: Names + "DisposesPreviousSession" is more clear.

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        var testScope = new TestScope();
+        var projectKey = "myproj";
+        var sessionMock = testScope.SetUpSSEFactoryToReturnNoOpSSESession(projectKey);
+        sessionMock.Setup(session => session.Dispose());
+        testScope.SwitchToConnectedMode(projectKey);
+
+        testScope.SwitchToStandaloneMode();
+
+        sessionMock.Verify(session => session.Dispose(), Times.Once);
```
rita-sonarsource, line 100: where do you verify that no new session was created?

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+        var projectKey2 = "proj2";
+        var sessionMock1 = testScope.SetUpSSEFactoryToReturnNoOpSSESession(projectKey1);
+        sessionMock1.Setup(session => session.Dispose());
+        testScope.SwitchToConnectedMode(projectKey1);
+        var sessionMock2 = testScope.SetUpSSEFactoryToReturnNoOpSSESession(projectKey2);
+
+        testScope.SwitchToConnectedMode(projectKey2);
+
+        sessionMock1.Verify(session => session.Dispose(), Times.Once);
+        sessionMock2.Verify(session => session.Dispose(), Times.Never);
```
rita-sonarsource, line 117: Missing factory verifications

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    [TestMethod]
+    public void Dispose_WhenProjectOpened_CorrectAndIdempotent()
+    {
+        var testScope = new TestScope();
+        var projectKey = "myproj";
+        var sseSession = testScope.SetUpSSEFactoryToReturnNoOpSSESession(projectKey);
+        testScope.SetUpCorrectDisposeOrder(sseSession);
```
rita-sonarsource, line 126: SetUpCorrectDisposeOrder = I'm not sure what it's supposed to do and from looking at the implementation I still don't get it. Why do you need to set up anything for Dispose order?

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+        var sessionMock2 = testScope.SetUpSSEFactoryToReturnNoOpSSESession(projectKey2);
+
+        testScope.SwitchToConnectedMode(projectKey2);
+
+        sessionMock1.Verify(session => session.Dispose(), Times.Once);
+        sessionMock2.Verify(session => session.Dispose(), Times.Never);
+    }
+
+    [TestMethod]
+    public void Dispose_WhenProjectOpened_CorrectAndIdempotent()
```
rita-sonarsource, line 121: _WhenSessionExists_SessionIsDisposed

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+        testScope.SetUpCorrectDisposeOrder(sseSession);
+        testScope.SwitchToConnectedMode(projectKey);
+
+        CallDisposeMultipleTimes(testScope);
+
+        CheckEverythingIsDisposedOnce(testScope);
+    }
+
+    [TestMethod]
+    public void Dispose_WhenProjectClosed_CorrectAndIdempotent()
```
rita-sonarsource, line 135: _WhenSessionDoesNotExist_... - I'm not sure what it's supposed to dispose and `CheckEverythingIsDisposedOnce` is really not clear

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -0,0 +1,235 @@
...
+        var testScope = new TestScope();
+        testScope.SetUpCorrectDisposeOrder(null);
+        testScope.SwitchToStandaloneMode();
+
+        CallDisposeMultipleTimes(testScope);
+
+        CheckEverythingIsDisposedOnce(testScope);
+    }
+
+    private static void CheckEverythingIsDisposedOnce(TestScope testConfiguration)
```
rita-sonarsource, line 146: Split this to 2, VerifyEventUnsubscribed, VerifySessionFactoryDisposed

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionTests.cs
```diff
@@ -0,0 +1,227 @@
...
+using SonarQube.Client.Models.ServerSentEvents;
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.ServerSentEvents;
+
+[TestClass]
+public class SSESessionTests
+{
+    [TestMethod]
+    public void PumpAllAsync_WhenSonarQubeRefusesConnection_DoesNotThrow()
```
rita-sonarsource, line 42: _WhenServerReturnsNullStream_

Also, what do you expect to happen? Do you expect the loop the break? the thread to end?

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionTests.cs
```diff
@@ -0,0 +1,227 @@
...
+    {
+        var testScope = new TestScope();
+        testScope.SetUpSwitchToBackgroundThread();
+        testScope.SonarQubeServiceMock
+            .InSequence(testScope.MainCallSequence)
```
rita-sonarsource, line 47: MainCallSequence = not clear. Why is this needed?

rita-sonarsource (approved):
LGTM, see my comments regarding name changes

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
rita-sonarsource: I'd split this test to 2, there are enough differences w

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionTests.cs
georgii-borovinskikh-sonarsource: To exit without doing anything

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
georgii-borovinskikh-sonarsource: Yeah, but it also checks that it disposes the factory and unsubscribes from the event, so I would rather keep the CorrectAndIdempotent suffix

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
georgii-borovinskikh-sonarsource: InSequence only works with Strict mocks, which throw if `.Setup(x => x.MethodName(...))` is not called

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
georgii-borovinskikh-sonarsource: It is duplicated by other tests, so I removed it and added any missing verifications to those tests

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
georgii-borovinskikh-sonarsource: I changed it

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
georgii-borovinskikh-sonarsource: I changed it

Issue #3578, opened 2023-01-25
Extract rule HTML descriptions from json


duncanp-sonar:
### Description

Extract the description fields from the per-language json metadata and save each rule description in separate file in a per-language folder under the `Rules` assembly.

```
Rules.csproj
-- Help
    -- cs
        -- S100.html
        -- S101.html
...
    -- c
        -- S100.html
        -- S101.html
...
etc

```

PR #3580, opened 2023-01-26
Add rule description extraction tool

duncanp-sonar:
Part of #3578


build/ProcessJarFiles/JarProcessing.targets
```diff
@@ -58,6 +61,16 @@
     <Exec Command="java -jar $(RuleExtractorJar) --plugins $(JarDownloadDir)\%(PluginJars.JarName) --languages=%(PluginJars.LanguageKey) --output=$(_WorkingDirectory)\%(PluginJars.LanguageKey)-rules.json" />
   </Target>
 
+
+  <Target Name="ExtractRuleDescFromJson">
+    <Message Importance="high" Text="[ProcessJarFiles] Checking for existence of rule extractor exe: $(RuleDescExtractorExe) ..." />
+    <Error Condition="!Exists($(RuleDescExtractorExe))" Text="Rule description extractor exe does not exist: $(RuleDescExtractorExe)" />
+    <Message Importance="high" Text="[ProcessJarFiles]    ... exe exists." />
+
+    <!-- TODO - call exe -->
```
duncanp-sonar, line 70: So far the targets just check that the exe has been built and can be found.

Actually calling the exe will be done in a separate PR.

duncanp-sonar:
Adds a new internal, build-time tool.

The solution file has been changed so the internal tool is built early (i.e. before it is used by ProcessJarFiles).

PR #3581, opened 2023-01-26
Embed rule descriptions as resources in Rules.csproj

duncanp-sonar:
Part of #3578


.gitignore
```diff
@@ -169,3 +169,4 @@ $RECYCLE.BIN/
 /build/DownloadSonarPlugin/ManualTestOutput
 /build/ProcessJarFiles/tmp
 /**/launchSettings.json
+/src/Rules/Embedded
```
duncanp-sonar, line 172: Ignore the generated files (there are thousands of them, and they're regenerated as part of the build).

Issue #3544, opened 2023-01-15
Handle non-parse-able HTML descriptions


duncanp-sonar:
### Description

The HTML descriptions generated for rules is not XHTML i.e. it might not be parse-able as XML.

Examples:

```
Error: file: D:\wip\RuleDescriptions\jars\c-rules.json, rule: c:S1911, error: The 'col' start tag on line 5 position 6 does not match the end tag of 'colgroup'. Line 6, position 5.

Error: file: D:\wip\RuleDescriptions\jars\cpp-rules.json, rule: cpp:S1911, error: The 'col' start tag on line 5 position 6 does not match the end tag of 'colgroup'. Line 6, position 5.

Error: file: D:\wip\RuleDescriptions\jars\cpp-rules.json, rule: cpp:S3659, error: The 'col' start tag on line 7 position 6 does not match the end tag of 'colgroup'. Line 8, position 5.

Error: file: D:\wip\RuleDescriptions\jars\csharp-rules.json, rule: csharpsquid:S4462, error: The 'col' start tag on line 16 position 6 does not match the end tag of 'colgroup'. Line 17, position 5.

Error: file: D:\wip\RuleDescriptions\jars\csharp-rules.json, rule: csharpsquid:S6422, error: The 'col' start tag on line 7 position 6 does not match the end tag of 'colgroup'. Line 8, position 5.

Error: file: D:\wip\RuleDescriptions\jars\csharp-rules.json, rule: csharpsquid:S4069, error: The 'col' start tag on line 6 position 6 does not match the end tag of 'colgroup'. Line 7, position 5.

Error: file: D:\wip\RuleDescriptions\jars\csharp-rules.json, rule: csharpsquid:S3962, error: The 'col' start tag on line 8 position 6 does not match the end tag of 'colgroup'. Line 9, position 5.

Error: file: D:\wip\RuleDescriptions\jars\csharp-rules.json, rule: csharpsquid:S2148, error: The 'col' start tag on line 7 position 6 does not match the end tag of 'colgroup'. Line 8, position 5.

Error: file: D:\wip\RuleDescriptions\jars\js-rules.json, rule: javascript:S6321, error: The 'br' start tag on line 2 position 19 does not match the end tag of 'p'. Line 3, position 107.

[...]

PR #3586, opened 2023-01-29
Add test that rule desc resources are XML

duncanp-sonar:
Part of #3544

### DIFF
--- a/build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
+++ b/build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
@@ -18,8 +18,8 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System.Diagnostics;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 
 namespace ExtractRuleDescFromJson;
 
@@ -98,8 +98,8 @@ internal class RuleDescExtractor
     {
         try
         {
-            // TODO: clean up HTML?
-            SaveRuleFile(rule);
+            var xml = EnsureHtmlIsXml(rule.Description);
+            SaveRuleFile(rule, xml);
         }
         catch (Exception ex)
         {
@@ -107,11 +107,30 @@ internal class RuleDescExtractor
         }
     }
 
-    private void SaveRuleFile(Rule rule)
+    // Regular expression that find empty "col"and "br" HTML elements
+    // e.g. <br>, <br >, <col>, <col span="123">
+    // This is valid HTML, but means we can't parse it as XML. So, we find
+    // the empty elements and replace them with elements with closing tags
+    // e.g. <br>  =>  <br/>
+    // e.g. <col span="123">  =>  <col span="123"/>
+    private static Regex cleanCol = new Regex("(?<element>(<col\\s*)|(col\\s+[^/^>]*))>", RegexOptions.Compiled);
+    private static Regex cleanBr = new Regex("(?<element>(<br\\s*)|(br\\s+[^/^>]*))>", RegexOptions.Compiled);
+
+    private static string EnsureHtmlIsXml(string html)
+    {
+        var xml = html.Replace("&nbsp;", "&#160;");
+
+        xml = cleanCol.Replace(xml, "${element}/>");
+        xml = cleanBr.Replace(xml, "${element}/>");
+
+        return xml;
+    }
+
+    private void SaveRuleFile(Rule rule, string xml)
     {
         var fullPath = CalculateRuleFileName(rule);
         Logger.LogPartialMessage($" {Path.GetFileName(fullPath)}");
-        File.WriteAllText(fullPath, rule.Description);
+        File.WriteAllText(fullPath, xml);
     }
 
     private string CalculateRuleFileName(Rule rule)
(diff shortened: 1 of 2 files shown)


build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
```diff
@@ -98,20 +98,39 @@ private void ProcessRule(Rule rule)
...
+    // e.g. <col span="123">  =>  <col span="123"/>
+    private static Regex cleanCol = new Regex("(?<element>(<col\\s*)|(col\\s+[^/^>]*))>", RegexOptions.Compiled);
+    private static Regex cleanBr = new Regex("(?<element>(<br\\s*)|(br\\s+[^/^>]*))>", RegexOptions.Compiled);
+
+    private static string EnsureHtmlIsXml(string html)
+    {
+        var xml = html.Replace("&nbsp;", "&#160;");
+
+        xml = cleanCol.Replace(xml, "${element}/>");
+        xml = cleanBr.Replace(xml, "${element}/>");
```
duncanp-sonar, line 124: The HTML generated for the rule description has a couple of things that stop it from being parseable as XML:

1) it uses entities that are not recognised by the XML reader. Here, the only one we need to worry about is `&nbsp;`, which we just replace with Unicode equivalent.

2) some HTML tags are defined as being "empty" i.e. they don't need a closing tag. The ones we need to worry about are `<col ...>` and `<br>`. We handle this by finding and replacing them by `col .../>` and `<br/>` respectively.

src/Rules.UnitTests/EmbeddedResourceTests.cs
```diff
@@ -58,6 +61,44 @@ public void CheckEmbeddedDescriptionFiles_ByLanguage(string languageKey, int atL
...
+        private static bool ReadResourceAsXml(Assembly asm, string fullResourceName)
+        {
+            try
+            {
+                var readerSettings = new XmlReaderSettings
+                {
+                    ConformanceLevel = ConformanceLevel.Fragment
+                };
+                using var stream = new StreamReader(asm.GetManifestResourceStream(fullResourceName));
+                using var reader = XmlReader.Create(stream, readerSettings);
```
duncanp-sonar, line 89: The `XmlReader` is very fast, so parsing 2000 files in the test only takes a few hundred milliseconds.

src/Rules.UnitTests/EmbeddedResourceTests.cs
```diff
@@ -58,6 +61,44 @@ public void CheckEmbeddedDescriptionFiles_ByLanguage(string languageKey, int atL
             resourceNames.Should().HaveCountGreaterThan(atLeastXRules);
         }
 
+        [TestMethod]
+        public void CheckEmbeddedDescriptionFiles_AreParseableAsXml()
```
duncanp-sonar, line 65: Tests loading all of the embedded rule descriptions and making sure we can parse them as XML.

This is a safety net in case the format of the generated rule description HTML changes - this test will fail if we update to a newer version of an analyzer with a new-format rule description.

duncanp-sonar:
Modification to the rule description extraction tool to fix up the extracted HTML so that it can be parsed as XML
i.e. makes sure that what we embedded in `Rules.csproj` is XML.

That makes writing the "HTML to XAML" converter much simpler since we can use normal .NET `XmlReader` class.

src/Rules.UnitTests/EmbeddedResourceTests.cs
```diff
@@ -58,6 +61,44 @@ public void CheckEmbeddedDescriptionFiles_ByLanguage(string languageKey, int atL
...
+            try
+            {
+                var readerSettings = new XmlReaderSettings
+                {
+                    ConformanceLevel = ConformanceLevel.Fragment
+                };
+                using var stream = new StreamReader(asm.GetManifestResourceStream(fullResourceName));
+                using var reader = XmlReader.Create(stream, readerSettings);
+
+                while (reader.Read()) { }
```
duncanp-sonar, line 91: As discussed offline, I've checked that this is actually reading the whole document, and it is. 

I tweaked the test locally to dump each node that was read, and then checked the logged output:
```
                Console.WriteLine("Reading document: " + fullResourceName);
                while (reader.Read())
                {
                    Console.WriteLine($"ElementType: {reader.NodeType}, {reader.Name}");
                }
                Console.WriteLine();
```
Start of logged output:
image

End of logged output:
image

build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
```diff
@@ -18,8 +18,8 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
```
bigfluffycookie, line 18: Silly question, I have used the word "xaml" often whereas you use "xml" is that something I should pay attention to when naming variables or?

bigfluffycookie (approved):
Looks good just 1 question.

build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
duncanp-sonar: It depends on the content. `XML` is a general-purpose markup language that specifies how a document should be structured for it to be a "well-formed" XML document. It doesn't say anything about what the content of the document means.

`XAML` refers to a specific UI technology that happens to use XML as the serialization format. If we're talking about a XAML document, we expect the content to refer to XAML UI elements e.g. listboxes, labels, panels etc.

Here, we don't expect the content of the embedded files to be XAML, but we do expect they to be parseable as XML.

Issue #3560, opened 2023-01-24
Add a method to IToolWindowService "GetToolWindow"


bigfluffycookie:
Add a generic method that returns the toolwindowpane.

PR #3587, opened 2023-01-30
Change method signature of getToolWindow in ToolWindowService.

bigfluffycookie:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3560


src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
```diff
@@ -143,7 +143,7 @@ public void GetToolWindow_ReturnsCorrectType()
             SetupFindToolWindow(serviceProviderMock, uiShellMock, VSConstants.S_OK, new Guid(MyDummyToolWindow.GuidAsString), frameMock.Object);
 
             var testSubject = new ToolWindowService(serviceProviderMock.Object);
-            var result = testSubject.GetToolWindow<MyDummyToolWindow>();
+            var result = testSubject.GetToolWindow<MyDummyToolWindow, MyDummyToolWindow>();
```
duncanp-sonar, line 146: It would be better to introduce a new `IDummyToolWindow` interface and use that in the test i.e. make the test more accurately reflect how the product code is going to be used.

duncanp-sonar (commented):
Looks good, but I'd tweak to test so the `T` and `V` are different types.

duncanp-sonar (approved):

Issue #3536, opened 2023-01-12
Add rule description service


duncanp-sonar:
### Description

Add a service that callers can use to request the display of rule help for a specific rule.

Things to consider:
* which assembly should the interface definition be in?
* which assembly should the implementing class be in?
* what should the interface look like?

e.g.
```
    public interface IEducation
    {
        void ShowRuleHelp(string language, string ruleID);
    }

    internal class Education : IEducation
    {
        public void ShowRuleHelp(string language, string ruleID)
        {
            // Calculate XAML resource name

            // Deserialize XAML string to FlowDocument

            // Update tool window

            // Ensure tool window is visible
        }
    }

```

PR #3588, opened 2023-01-30
Add Education Service

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3536


src/Education.UnitTests/EducationTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            _ = CreateEducation(toolWindowService: toolWindowService.Object);
+
+            toolWindowService.Verify(x => x.GetToolWindow<RuleDescriptionToolWindow, IRuleDescriptionToolWindow>(), Times.Once);
+        }
+
+        [TestMethod]
+        public void ShowRuleDescription_EverythingGetsCalledCorrectly()
+        {
+            var ruleMetaDataProvider = new Mock<IRuleMetadataProvider>();
+            var language = Language.Unknown;
```
bigfluffycookie, line 58: I am not sure if this is okay or not please lmk

src/Education/IRuleHelpXamlBuilder.cs
```diff
@@ -26,7 +26,7 @@
 
 namespace SonarLint.VisualStudio.Education
 {
-    internal interface IRuleHelpXamlBuilder
+    public interface IRuleHelpXamlBuilder
```
bigfluffycookie, line 29: Missed this in a previous pr.

src/Education.UnitTests/EducationTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        public void Ctor_GetsToolWindowFromService()
+        {
+            var toolWindowService = new Mock<IToolWindowService>();
+
+            _ = CreateEducation(toolWindowService: toolWindowService.Object);
+
+            toolWindowService.Verify(x => x.GetToolWindow<RuleDescriptionToolWindow, IRuleDescriptionToolWindow>(), Times.Once);
+        }
+
+        [TestMethod]
```
bigfluffycookie, line 54: As discussed this tests the whole ShowRuleDescription method.

src/Education/Commands/RuleDescriptionToolWindow.cs
```diff
@@ -20,20 +20,34 @@
...
         private const string ToolWindowIdAsString = "9E74B368-9FC3-47B0-A1C7-2DBA3A2C1762";
         public static readonly Guid ToolWindowId = new Guid(ToolWindowIdAsString);
+        private readonly FlowDocumentReader flowDocumentReader;
 
         public RuleDescriptionToolWindow()
         {
             Caption = Resources.RuleDescriptionToolWindowCaption;
-            Content = new RuleDescriptionUserControl();
+            flowDocumentReader = new FlowDocumentReader();
+            Content = flowDocumentReader;
```
duncanp-sonar, line 45: I'm not sure if there are any threading issues we need to be aware of here i.e. do we need to make sure the tool window is created/updated on the UI thread?

However, it's going to be hard to check until we're actually populating the window for real in the IDE, so I think we can leave for now (although it would be worth adding a couple of `TODO: check if UI thread requried` here and in the `UpdateContent` method).

duncanp-sonar: Ah - the thread handling is in the education service, which makes sense.

src/Education/IEducation.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                try
+                {
+                    toolWindowService.Show(RuleDescriptionToolWindow.ToolWindowId);
+                }
+                catch (Exception ex) when (!Core.ErrorHandler.IsCriticalException(ex))
+                {
+                    logger.WriteLine(string.Format(Resources.ERR_RuleDescriptionToolWindow_Exception, ex));
+                }
+            }).Forget();
```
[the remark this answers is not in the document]
duncanp-sonar: `Forget` is ok here - we're not really worried about race conditions; we've asked for the UI to be updated, so we might as well return.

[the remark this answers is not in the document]
duncanp-sonar: It does beg the question whether we should have made the service method async, although that would have just moved the problem upstream i.e. any `void` callers would have had to handle calling this `asyc` method.

src/Education/IEducation.cs
```diff
@@ -0,0 +1,91 @@
...
+            ruleDescriptionToolWindow = toolWindowService.GetToolWindow<RuleDescriptionToolWindow, IRuleDescriptionToolWindow>();
+        }
+
+        public void ShowRuleDescription(Language language, string ruleKey)
+        {
+            var ruleHelp = ruleMetadataProvider.GetRuleHelp(language, ruleKey);
+
+            var flowDocument = ruleHelpXamlBuilder.Create(ruleHelp);
+
+            ruleDescriptionToolWindow.UpdateContent(flowDocument);
```
duncanp-sonar, line 70: I think the creation of the flow document and updating of the window need to done on the UI thread too (see [here](https://learn.microsoft.com/en-us/dotnet/desktop/wpf/advanced/threading-model?view=netframeworkdesktop-4.8)). Since you already have a `RunOnUIThread` block, it might be better to refactor the methods so everything except line 66 is done inside the block.

src/Education.UnitTests/EducationTests.cs
duncanp-sonar: Yep, it's ok. `Language.Unknown` is a valid instance of the `Language` class, and you're checking it's passed to the appropriate service in lines 62 and 76.

src/Education.UnitTests/EducationTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            var toolWindowService = new Mock<IToolWindowService>();
+            toolWindowService.Setup(x => x.GetToolWindow<RuleDescriptionToolWindow, IRuleDescriptionToolWindow>()).Returns(ruleDescriptionToolWindow.Object);
+
+            var testSubject = CreateEducation(toolWindowService: toolWindowService.Object, ruleMetadataProvider: ruleMetaDataProvider.Object, ruleHelpXamlBuilder: ruleHelpXamlBuilder.Object);
+            testSubject.ShowRuleDescription(language, ruleKey);
+
+            ruleMetaDataProvider.Verify(x => x.GetRuleHelp(language, ruleKey), Times.Once);
+            ruleHelpXamlBuilder.Verify(x => x.Create(ruleHelp), Times.Once);
+            ruleDescriptionToolWindow.Verify(x => x.UpdateContent(flowDocument), Times.Once);
+        }
```
duncanp-sonar, line 79: Should also check that `toolWindowService.ShowToolWindow` is called.

duncanp-sonar (commented):
Looks good. A couple of comments about moving some of the work to the UI thread, and a missing test check.

duncanp-sonar (approved):

Issue #3591, opened 2023-01-31
Add a CommandInterceptor


bigfluffycookie:
Add a class that enables us to intercept commands before they get called.

PR #3592, opened 2023-01-31
Add a command interceptor class

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3591


src/Integration.UnitTests/CommandInterceptorTests.cs
```diff
@@ -0,0 +1,88 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
```
bigfluffycookie, line 20: Here it would probably make sense to check if the functions are called, I am just not quite sure how without making it ugly.

src/Integration/CommandInterceptor.cs
```diff
@@ -0,0 +1,77 @@
...
+            this.threadHandling = threadHandling;
+        }
+
+        public int QueryStatus(ref Guid pguidCmdGroup, uint cCmds, OLECMD[] prgCmds, IntPtr pCmdText)
+        {
+            return (int)Microsoft.VisualStudio.OLE.Interop.Constants.OLECMDERR_E_NOTSUPPORTED;
+        }
+
+        public int Exec(ref Guid pguidCmdGroup, uint nCmdID, uint nCmdexecopt, IntPtr pvaIn, IntPtr pvaOut)
+        {
```
bigfluffycookie, line 51: I've noticed that sometimes when I replaced ThreadHelper.ThrowIfNotUOnUIThread i used RunOnUIThread instead, the only reason I didn't do it here is because it was simpler this way. LMK if thats a problem.

duncanp-sonar: It depends where we want to put the responsibility for managing threading.

If we throw here, we're effectively saying it is the responsibility of the caller to make sure we are on the correct thread.

If we handle the transition in this class, we're removing the responsibility from the caller.

Here, the only caller will be VS. I would expect that we are always called on the UI thread, so throwing in this case is ok. If it turns out that VS can call the method on a background thread we'll find out pretty soon ;-)

src/Integration.UnitTests/CommandInterceptorTests.cs
```diff
@@ -0,0 +1,88 @@
...
+    [TestClass]
+    public class CommandInterceptorTests
+    {
+        [TestMethod]
+        public void Exec_CommandProgressionStops_ReturnsZero()
+        {
+            var guid = new Guid();
+            var commandID = new CommandID(guid, 0);
+
+            var func = () => { return CommandProgression.Stop; };
```
duncanp-sonar, line 38: > Here it would probably make sense to check if the functions are called, I am just not quite sure how without making it ugly.

One simple way is to create a local variable and set it in the test:

```
bool funcCalled = false;
var func = () =>
  {
    funcCalled = true;
    return CommandProgression.Stop;
  }

...


funcCalled.Should().BeTrue();
```

duncanp-sonar (commented):
Looks good. Just a comment about choosing test input data values .

duncanp-sonar:
@bigfluffycookie could you change the tests to use non-zero Guids and non-zero cmdIds please?

duncanp-sonar:
@bigfluffycookie apologies, I didn't notice that the new class was in `Integration`. It should be in `Infrastructure.VS`, which is a helper assembly for general purpose classes that depend on Visual Studio.

It's likely we'll need an `ICommandInterceptor` interface so we can mock it when testing classes that use it. However, we can wait and see, and add it if we need it.

duncanp-sonar:
(...don't forget to change the namespaces when you move the files...)

duncanp-sonar (approved):

PR #3595, opened 2023-02-01
Change EventPump into a transient instance

rita-sonarsource:
@georgii-borovinskikh-sonarsource I went over the design with @duncanp-sonar and we came up with a solution that we think is simpler and addresses a couple of problems:
* It allows the whole "session" to be disposed, whenever the binding changes, without the risk of concurrency problems
* It makes explicit usage of the cancellation token which makes it simpler to understand how the session is disposed and canceled.

Please take a look and LMK what you think.

### DIFF
--- a/src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
+++ b/src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
@@ -51,21 +51,21 @@ namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents
             this.activeSolutionBoundTracker = activeSolutionBoundTracker;
             this.sseSessionFactory = sseSessionFactory;
 
-            activeSolutionBoundTracker.SolutionBindingChanged += OnSolutionChanged;
+            activeSolutionBoundTracker.SolutionBindingChanged += SolutionBindingChanged;
         }
 
         public void Dispose()
         {
-            activeSolutionBoundTracker.SolutionBindingChanged -= OnSolutionChanged;
-            currentSession?.Dispose();
+            activeSolutionBoundTracker.SolutionBindingChanged -= SolutionBindingChanged;
+            EndCurrentSession();
         }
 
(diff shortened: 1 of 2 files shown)


src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -25,7 +25,10 @@
 using Microsoft.VisualStudio.Threading;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Core.ServerSentEvents.Issues;
+using SonarLint.VisualStudio.Core.ServerSentEvents.TaintVulnerabilities;
 using SonarQube.Client;
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
 
 namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents
```
rita-sonarsource, line 33: @duncanp-sonar as we discussed offline, LMK what you think of this

src/ConnectedMode/ServerSentEvents/ServerSentEventSessionManager.cs
```diff
@@ -38,75 +41,153 @@ internal interface IServerSentEventSessionManager : IDisposable
...
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class SSESessionFactory : ISSESessionFactory
+    {
+        private readonly ISonarQubeService sonarQubeClient;
+        private readonly IIssueChangedServerEventSourcePublisher issueChangedServerEventSourcePublisher;
+        private readonly ITaintServerEventSourcePublisher taintServerEventSourcePublisher;
+        private readonly IThreadHandling threadHandling;
+
+        [ImportingConstructor]
+        public SSESessionFactory(ISonarQubeService sonarQubeClient,
```
rita-sonarsource, line 101: I'm thinking something nicer can be done here in terms of duplication but that's the gist.

georgii-borovinskikh-sonarsource (approved):
Good change. Will continue working on it in my branch

Issue #3557, opened 2023-01-23
Convert simple rule HTML help to XAML


duncanp-sonar:
### Description

Translate simple Sonar HTML to XAML that can be displayed in the IDE.

Note: the Sonar HTML is not XHTML i.e. can't all be parsed as XML (about 12 rules out of 2000 are not XHTML)

#### TODO
* needs to handle theming
* needs to use current VS fonts/sizes
* ?make XAML async parse-able (add attribute to XAML?)
* separate formatting for code samples

#### List of all HTML elements/attributes
The complete list of elements and attributes used at time of writing is as follows:
```
p
code
h2
pre
ul
li
a, attributes: href
em
strong
h3
ol
blockquote

Rule count: 1915
Parse error count: 12
```

#### Non-XHTML rules descriptions
Most of the HTML rule descriptions are XHTML. However, are a few are not which means we can't parse them as XML.
Whether this is a problem or not depends on how we want to generate the XAML.

List of non-compliant rules:
```
Error: file: D:\wip\RuleDescriptions\jars\c-rules.json, rule: c:S1911, error: The 'col' start tag on line 5 position 6 does not match the end tag of 'colgroup'. Line 6, position 5.
Error: file: D:\wip\RuleDescriptions\jars\cpp-rules.json, rule: cpp:S1911, error: The 'col' start tag on line 5 position 6 does not match the end tag of 'colgroup'. Line 6, position 5.
Error: file: D:\wip\RuleDescriptions\jars\cpp-rules.json, rule: cpp:S3659, error: The 'col' start tag on line 7 position 6 does not match the end tag of 'colgroup'. Line 8, position 5.
Error: file: D:\wip\RuleDescriptions\jars\csharp-rules.json, rule: csharpsquid:S4462, error: The 'col' start tag on line 16 position 6 does not match the end tag of 'colgroup'. Line 17, position 5.
Error: file: D:\wip\RuleDescriptions\jars\csharp-rules.json, rule: csharpsquid:S6422, error: The 'col' start tag on line 7 position 6 does not match the end tag of 'colgroup'. Line 8, position 5.
Error: file: D:\wip\RuleDescriptions\jars\csharp-rules.json, rule: csharpsquid:S4069, error: The 'col' start tag on line 6 position 6 does not match the end tag of 'colgroup'. Line 7, position 5.
[...]

PR #3598, opened 2023-02-01
Implement rule description to XAML builder

duncanp-sonar:
Relates to #3557


src/Education.UnitTests/EducationTests.cs
```diff
@@ -22,6 +22,7 @@
 using Moq;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Education.Commands;
+using SonarLint.VisualStudio.Education.XamlGenerator;
```
duncanp-sonar, line 25: I moved the generator interface into a subfolder (because I expect the implement to required multiple separate class files) -> namespace changed.

src/Education.UnitTests/RuleHelpXamlBuilderTests.cs
```diff
@@ -0,0 +1,98 @@
...
+
+namespace SonarLint.VisualStudio.Education.UnitTests
+{
+    [TestClass]
+    public class RuleHelpXamlBuilderTests
+    {
+        private static readonly Assembly ResourceAssembly = typeof(RuleMetadataProvider).Assembly;
+
+        [TestMethod]
+        public void CreateXamlString_CheckAllEmbedded()
```
duncanp-sonar, line 39: As discussed at the standup this morning, I have only written this one test.
It doesn't check the detail of the XAML that is produced.
However, it does attempt to build the XAML for all of the embedded rule help files, and checks that the builder does not fail for any of them.

src/Education/Resources.resx
```diff
@@ -123,4 +123,10 @@
   <data name="RuleDescriptionToolWindowCaption" xml:space="preserve">
     <value>Rule Description</value>
   </data>
+  <data name="String1" xml:space="preserve">
```
duncanp-sonar, line 126: Doh! will remove this.

src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
```diff
@@ -0,0 +1,472 @@
...
+    public interface IRuleHelpXamlBuilder
+    {
+        /// <summary>
+        /// Generates a XAML document containing the help information for the specified rule
+        /// </summary>
+        /// <remarks>Assumes that the <see cref="IRuleHelp.HtmlDescription"/> is parseable as XML</remarks>
+        FlowDocument Create(IRuleHelp ruleHelp);
+    }
+
+    internal partial class RuleHelpXamlBuilder : IRuleHelpXamlBuilder
```
duncanp-sonar, line 42: I'll clean up and simplify the implementation once I've made the changes to support styling and theming.

At a high-level, the idea is fairly simple:
* read through the input HTML doc (treating it as XML)
* for each HTML element, write out at the appropriate XAML element(s) and possibly attributes.

Often there is a 1-1 mapping e.g.
* `<h2>` -> `<Paragraph FontSize="20pt">`
* `<ul` -> `<List>`
* `<br>` -> `<LineBreak>`

Lists and tables are a bit more complicated, but not much.

The builder only handles a subset of HTML tags, but it will throw for any it does not recognise.

src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
```diff
@@ -0,0 +1,472 @@
...
+
+            using (var stringReader = new StringReader(xaml))
+            {
+                var xamlReader = XmlReader.Create(stringReader);
+                var flowDocument = (FlowDocument)XamlReader.Load(xamlReader);
+                return flowDocument;
+            }
+        }
+
+        internal /* for testing */ string CreateXamlString(IRuleHelp ruleHelp)
```
duncanp-sonar, line 73: Added to make testing the produced XAML easier (it's easier to check a string than the XAML object graph). However, I've stashed the few low-level tests that I wrote, so this method isn't used by the current test.

src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
```diff
@@ -0,0 +1,472 @@
...
+                        default:
+                            var message = string.Format(Resources.XamlBuilder_UnexpectedNodeError, reader.NodeType, reader.Name, reader.Value);
+                            throw new InvalidDataException(message);
+                    }
+                }
+
+                // We've processed all of the html elements.
+                // Now, the stack should only contain the root FlowDocument element, plus an extra
+                // block element if the first tag we processed was an Inline.
+                Debug.Assert(outputXamlElementStack.Count <= 2, "Expecting at most 2 unclosed elements in the stack");
```
duncanp-sonar, line 117: Sanity check when we've finished writing that we have the expected items left on the stack.

src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
```diff
@@ -0,0 +1,472 @@
...
+        }
+
+        private void PushOutputElementInfo(string htmlElementName, bool supportsInlines)
+        {
+            outputXamlElementStack.Push(new XamlOutputElementInfo(htmlElementName, supportsInlines));
+        }
+
+        private void ProcessElement()
+        {
+            switch (reader.Name)
```
duncanp-sonar, line 158: I might be able to replace most of this code with a simple table lookup that that maps HTML elements to the corresponding XAML element(s).

For now, it's easier to leave it as it is in case I need to add any special logic to handle styling/theming.

src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
```diff
@@ -0,0 +1,472 @@
...
+
+        private void WriteEmptyElement(string name, string value = null)
+        {
+            Debug.Assert(reader.IsEmptyElement);
+            writer.WriteElementString(name, value);
+        }
+
+        private void WriteInlineElementStart(string elementName)
+        {
+            // If we are writing an inline element, we need a parent element that supports inlines.
```
duncanp-sonar, line 330: Long comments in this and the following methods to explain the one complicate part of building the XAML tree.

In summary: most XAML document elements can be classified as either `Inline` or `Block` (there are a few exceptions e.g. the table and list classes).

Some elements can't have child elements e.g. `LineBreak`.
Some elements can have one or more `Inline` children e.g. `Span` and `Paragraph`
Some elements can have one or more `Block` children e.g. `Section`, `ListItem`, `TableCell`.


Before adding a child element, we need to make sure it's parent can accept a child element of that type
e.g.
* current = `Paragraph`, want to add text -> no problem. `Paragraph` has `Inline` children, text is an `Inline`

* current = `ListItem`, want to add text -> can't. `ListItem` has `Blocks` as children. So, we first add a `Paragraph` (a block) -> `ListItem` -> `Paragraph` -> text.

* current = `Paragraph`, want to add a `List` -> can't. `List` is a `Block`. So, we need to walk back up the list of added elements closing them until we find something that accepts a `Block`. Note: there are limits on how far we can walk back up - we can only walk as far as an element that corresponds directly to an `HTML` opening tag. If we walk any further, we'll have a problem when we try to process the `HTML` element closing tag. To put it another way, we can only close extra intermediate elements we have added that don't map directly to an `HTML` tag.

duncanp-sonar:
@bigfluffycookie I suggest not spending too long now looking at the detail of the XAML builder since it will change. I'm happy to discuss how it works offline, which might be easier.

bigfluffycookie (approved):
Test look good to me!

Issue #3600, opened 2023-02-02
Change Language class to have a conversion between RepoKey and Language.


bigfluffycookie:
The language class currently only allows you to get a language if you have the language key. 

From the error list however, we only get access to the repository key so we need to add a method to convert this for us.

PR #3602, opened 2023-02-02
Add GetLanguageFromRepositoryKey to Languages.cs

bigfluffycookie:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3600

### DIFF
--- a/src/Core.UnitTests/LanguageTests.cs
+++ b/src/Core.UnitTests/LanguageTests.cs
@@ -124,13 +124,13 @@ namespace SonarLint.VisualStudio.Core.UnitTests
         [TestMethod]
         public void GetLanguageFromRepositoryKey_GetsCorrectLanguage()
         {
-            Language.GetLanguageFromRepositoryKey("csharpsquid", out Language cs);
-            Language.GetLanguageFromRepositoryKey("vbnet", out Language vbnet);
-            Language.GetLanguageFromRepositoryKey("cpp", out Language cpp);
-            Language.GetLanguageFromRepositoryKey("c", out Language c);
-            Language.GetLanguageFromRepositoryKey("javascript", out Language js);
-            Language.GetLanguageFromRepositoryKey("typescript", out Language ts);
-            Language.GetLanguageFromRepositoryKey("unknown", out Language unknown);
+            var cs = Language.GetLanguageFromRepositoryKey("csharpsquid");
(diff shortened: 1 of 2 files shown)


src/Core.UnitTests/LanguageTests.cs
```diff
@@ -100,5 +100,45 @@ public void Language_Equality()
             lang1b.Should().Be(lang1a, "Languages with the same ids should be equal");
             lang2.Should().NotBe(lang1a, "Languages with different ids should NOT be equal");
         }
+
+        [TestMethod]
```
bigfluffycookie, line 104: I noticed that this wasn't tested before so I added it.

src/Core/Language.cs
```diff
@@ -96,6 +101,19 @@ public static IEnumerable<Language> KnownLanguages
             }
         }
 
+        /// <summary>
+        /// Matches the repository name to its language object counterpart.
+        /// </summary>
+        private static readonly Dictionary<string, Language> repoKeyToLanguage = new Dictionary<string, Language>()
```
bigfluffycookie, line 107: Are we using dictionaries for these kinds of things? To me it made sense but maybe I am wrong

duncanp-sonar: Yep, makes sense.

src/Core/Language.cs
```diff
@@ -56,6 +56,11 @@ public sealed class Language : IEquatable<Language>
         public static Language GetLanguageFromLanguageKey(string languageKey) =>
             KnownLanguages.FirstOrDefault(l => languageKey.Equals(l.ServerLanguage.Key, System.StringComparison.OrdinalIgnoreCase));
 
+        /// <summary>
+        /// Tries to get the language for the specified repository key.
+        /// </summary>
+        public static bool GetLanguageFromRepositoryKey(string repoKey, out Language language) => repoKeyToLanguage.TryGetValue(repoKey, out language);
```
duncanp-sonar, line 62: It seems a bit odd that the `Get` methods have similar names (`GetLanguageFromLanguageKey`, `GetLanguageFromRepositoryKey`), but have different signatures.

Your new method is really a `TryGet`. Does it need to be?
If it does, the test needs to check the return value as well as the `out` value.

bigfluffycookie: It does not need to be, not sure why I have made it that way

duncanp-sonar (approved):

Issue #3556, opened 2023-01-23
Override the default VS "show rule help command" behaviour for Sonar issues


duncanp-sonar:
### Description

Override the default VS "Show Error Help" menu command behaviour so that the Sonar Rule Help tool window is shown for Sonar issues for all languages.

TODO: might need to split this ticket to handle Roslyn/non-Roslyn issues separately

Investigate using `IVsRegisterPriorityCommandTarget`

See https://github.com/VsixCommunity/Community.VisualStudio.Toolkit/blob/4230c40c58be62c51a9c3bf444733a0fde850f8b/src/toolkit/Community.VisualStudio.Toolkit.Shared/Commands/Commands.cs

PR #3605, opened 2023-02-02
Add class ShowRuleHelpFromErrorList which initializes command interception and handles it.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3556


src/Education.UnitTests/Commands/ShowRuleHelpFromErrorListTests.cs
```diff
@@ -0,0 +1,118 @@
...
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Education.Commands;
+using SonarLint.VisualStudio.Infrastructure.VS;
+using SonarLint.VisualStudio.Integration.UnitTests;
+
+namespace SonarLint.VisualStudio.Education.UnitTests.Commands
+{
```
bigfluffycookie, line 32: I wanted to add a test ->

```  
[TestMethod]
  {
            MefTestHelpers.CheckTypeCanBeImported<ShowHelpFromErrorList, IShowHelpFromErrorList>(
                MefTestHelpers.CreateExport<SVsServiceProvider>(),
                MefTestHelpers.CreateExport<IEducation>(),
                MefTestHelpers.CreateExport<IErrorListHelper>());
}
```

However, this calls the ctor, which calls "InitializeInterception()", which uses the threadhandler. This then throws an exception. Please let me know how to deal with this :)

src/Education/Commands/ShowRuleHelpFromErrorList.cs
```diff
@@ -0,0 +1,91 @@
...
+            priority.RegisterPriorityCommandTarget(0, interceptor, out uint cookie);
+        }
+
+        /// <summary>
+        /// If the currently selected error code is a Sonar code show the rule description and
+        /// stop progression to next command else move on.
+        /// </summary>
+        /// <returns></returns>
+        internal CommandProgression HandleInterception()
+        {
```
bigfluffycookie, line 80: I was about to add an "If language is null" check but then I realized I wasn't sure about the desired behavior, can I assume that "ShowRuleDescription" will handle it? And if not do we want to throw an exception or just return?

bigfluffycookie:
> Kudos, SonarCloud Quality Gate passed!

Why does sonarcloud pass with two "Major" code smells?

src/Education.UnitTests/Commands/ShowRuleHelpFromErrorListTests.cs
duncanp-sonar: Ah. One option would be to MEF import `IThreadHandling` , and in `CreateExport` pass in the no-op thread handling implementation.

You would need to add a new overload to `CreateExport` since the current one is hard-coded to use a dummy value:

```
        [DebuggerStepThrough]
        public static Export CreateExport<T>() where T: class
        {
            return CreateExport<T>(Mock.Of<T>());
        }
```

(there did use to be an overload that accepted the object to export, but I removed it because it wasn't being used).
However, you would still need to provide a minimal implementation for the service provider and `IVsRegisterPriorityCommandTarget`, otherwise that code will fail too.

An alternative would be to leave the test as it is, but catch the exception and check it is null-ref exception. It's a bit ugly, but it would be enough to tell us what we want to know i.e. that the class was created via MEF (if the class can't be created via MEF we'd get a `CompositionException` instead).

src/Education/Commands/ShowRuleHelpFromErrorList.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.ComponentModel.Composition;
+using System.ComponentModel.Design;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Infrastructure.VS;
+
+namespace SonarLint.VisualStudio.Education.Commands
+{
+    public interface IShowHelpFromErrorList
```
duncanp-sonar, line 31: Missing doc comment

src/Education/Commands/ShowRuleHelpFromErrorList.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    [Export(typeof(IShowHelpFromErrorList))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class ShowHelpFromErrorList : IShowHelpFromErrorList
+    {
+        private readonly IThreadHandling threadHandling;
+        private readonly IEducation education;
+        private readonly IErrorListHelper errorListHelper;
+
+        private readonly Guid showHelpCmdGuid = new Guid("4A9B7E50-AA16-11D0-A8C5-00A0C921A4D2");
```
duncanp-sonar, line 42: Probably worth adding a comment to make it clear that this is the MS-defined "Show Error Help" command.

src/Education/Commands/ShowRuleHelpFromErrorList.cs
```diff
@@ -0,0 +1,91 @@
...
+            var interceptor = new CommandInterceptor(command, () => HandleInterception());
+
+            priority.RegisterPriorityCommandTarget(0, interceptor, out uint cookie);
+        }
+
+        /// <summary>
+        /// If the currently selected error code is a Sonar code show the rule description and
+        /// stop progression to next command else move on.
+        /// </summary>
+        /// <returns></returns>
```
duncanp-sonar, line 78: Empty `<returns>`

src/Education/Commands/ShowRuleHelpFromErrorList.cs
```diff
@@ -0,0 +1,91 @@
...
+        internal CommandProgression HandleInterception()
```
duncanp-sonar, line 79: I always add `internal /* for testing */` to make it absolutely clear why it's internal.

src/Education.UnitTests/Commands/ShowRuleHelpFromErrorListTests.cs
```diff
@@ -0,0 +1,118 @@
...
+            var result = testSubject.HandleInterception();
+
+            education.Verify(x => x.ShowRuleDescription(expectedLanguage, expectedRule), Times.Once);
+            result.Should().Be(CommandProgression.Stop);
+        }
+
+        [TestMethod]
+        public void HandleInterception_RuleIsNotSonar_ReturnsContinue()
+        {
+            var errorListHelper = CreateErrorListHelper("unknown:S222", ruleIsSonar: false); ;
```
duncanp-sonar, line 77: Extra ` ;` at the end of the line

duncanp-sonar (commented):
Looks good. A few answers/comments inline.

duncanp-sonar (approved):

PR #3607, opened 2023-02-02
Tweak the XAML generator

duncanp-sonar:
- alter whitespace handling
- add table cell style

### DIFF
--- a/src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
+++ b/src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
@@ -138,7 +138,6 @@ namespace SonarLint.VisualStudio.Education.XamlGenerator
         private void WriteDocumentHeader(IRuleHelp ruleHelp)
         {
             writer.WriteStartElement("FlowDocument", XamlNamespace);
-            writer.WriteAttributeString("xml", "space", null, "preserve");
             writer.WriteAttributeString("xmlns", "http://schemas.microsoft.com/winfx/2006/xaml/presentation");
             writer.WriteElementString("Paragraph", $"Rule: {ruleHelp.RuleKey}");
             writer.WriteElementString("Paragraph", $"Language: {ruleHelp.Language.Name}");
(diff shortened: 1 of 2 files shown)


src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
```diff
@@ -217,14 +216,14 @@ private void ProcessElement()
 
                 case "p":
                     writer.WriteStartElement("Paragraph");
-                    writer.WriteAttributeString("xml", "space", null, "default");
```
duncanp-sonar, line 220: I've made a few changes to how whitespace is being handled, having looked at some sample generated rule help.

Previously, whitespace was being `preserved` at the top-level, and turned off (`default`) on specific child cells i.e. most whitespace (particularly line breaks) would be rendered in the output. However, that was leading to line breaks appearing in odd places, so I've flipped it round.

Now, whitespace between tags is ignored (the default). It is only preserved for specific child elements, namely those inside the `pre` <html> tag.

src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
```diff
@@ -297,10 +296,8 @@ private void ProcessElement()
                     writer.WriteStartElement("TableCell");
                     PushOutputElementInfo("td", false);
 
-                    if (tableAlternateRow)
-                    {
-                        ApplyStyleToStartElement(StyleResourceNames.TableBodyCellAlternateRow);
-                    }
+                    var cellStyle = tableAlternateRow ? StyleResourceNames.TableBodyCellAlternateRow : StyleResourceNames.TableBodyCell;
```
duncanp-sonar, line 299: I also added a style for "normal" table cells. That means we can define one style for the odd-numbered rows, and another for the even-numbered rows.

bigfluffycookie: Ah very nice!

src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
bigfluffycookie: Makes sense

bigfluffycookie (approved):
Looks good! :)

Issue #3594, opened 2023-02-01
Add error list event processor


duncanp-sonar:
### Description

The hyperlink in the `errorcode` column of the Error List opens the issue help page in a browser.

It uses a different mechanism from the `Show Error Help` command on the Error List context menu i.e. intercepting the `Show Error Help` command has no effect on the hyperlink navigation.

This means we also need to provide a `TableControlEventProcessorBase` subclass that handles the navigate to help behaviour.

#### Prototype code

TODO - check metadata attributes
- [ ] can we limit the the`ManagerType` to `ErrorsTable`?
- [ ] can we limit the data source type to just our data sources? (probably not, since we need to handle Roslyn issues from our analyzers too)

```
namespace SonarLint.VisualStudio.Education.Commands
{
    [Export(typeof(ITableControlEventProcessorProvider))]
    //[Name("ErrorListPackage Table Control Event Processor")]
    //[Order(After = "Default Priority", Before = "Default Table Control Event Processor")]
    [Name("SonarLint ErrorListPackage Table Control Event Processor")]
    [Order(After = "Default Priority", Before = "ErrorListPackage Table Control Event Processor")]
    [ManagerType("ErrorsTable")]
    [ManagerType("TaskTable")]
    [DataSourceType("*")]
    [DataSource("*")]
    internal class EventProcessor : ITableControlEventProcessorProvider
    {
        public ITableControlEventProcessor GetAssociatedEventProcessor(IWpfTableControl tableControl)
        {
            return (ITableControlEventProcessor)new SonarLintErrorListPackageEventProcessor(tableControl);
        }
    }

    internal class SonarLintErrorListPackageEventProcessor : TableControlEventProcessorBase
    {
        private readonly IWpfTableControl wpfTableControl;

        public SonarLintErrorListPackageEventProcessor(IWpfTableControl wpfTableControl)
        {
            this.wpfTableControl = wpfTableControl;
        }

        public override void PreprocessNavigateToHelp(
           ITableEntryHandle entryHandle,
           TableEntryEventArgs e)
        {
            Requires.NotNull<ITableEntryHandle>(entryHandle, nameof(entryHandle));


            bool handled = false;
            if (entryHandle is ITableEntry tableEntry1)
            {
                tableEntry1.TryGetValue(StandardTableKeyNames.BuildTool, out string buildTool);
                tableEntry1.TryGetValue(StandardTableKeyNames.ErrorCode, out string errorCode);
[...]

PR #3612, opened 2023-02-02
Add IErrorListHelper method to get rule id from handle

duncanp-sonar:
- moved the IErrorListHelper interface to Infrastructure.VS

Prep for #3594


src/Infrastructure.VS/IErrorListHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// </remarks>
+        bool TryGetRuleIdFromSelectedRow(out SonarCompositeRuleId ruleId);
+
+        /// <summary>
+        /// Attempts to retrieve the Sonar rule key from the specific table entry in the Error List.
+        /// </summary>
+        /// The method will only return a rule key if the row represents a Sonar analysis issue for
+        /// any supported language (including Roslyn languages i.e. C# and VB.NET)
+
+        bool TryGetRuleId(ITableEntryHandle tableEntryHandle, out SonarCompositeRuleId ruleId);
```
duncanp-sonar, line 45: Users can click the "show help" hyperlink in the ErrorList "code" column without selecting the row first.

So, we need another method in this interface that doesn't check the selected rows, but instead accepts a handle indicating the row.

Unfortunately, that's a VS-specific interface, so I've moved `IErrorListHelper` to `Infrastructure.VS` :-(

src/Infrastructure.VS/IErrorListHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.VisualStudio.Shell.TableControl;
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.Infrastructure.VS
```
duncanp-sonar, line 24: Changed the namespace

src/Infrastructure.VS/IErrorListHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <summary>
+        /// Attempts to retrieve the Sonar rule key from the selected item in the Error List.
+        /// </summary>
+        /// <remarks>
+        /// The method will only return a rule key if:
+        /// * there is a single row selected in the Error List
+        /// * the row represents a Sonar analysis issue for any supported language
+        ///   (including Roslyn languages i.e. C# and VB.NET)
+        /// </remarks>
+        bool TryGetRuleIdFromSelectedRow(out SonarCompositeRuleId ruleId);
```
duncanp-sonar, line 37: This method and comment are the same as before.

src/Infrastructure.VS.UnitTests/ErrorListHelperTests.cs
```diff
@@ -74,7 +74,39 @@ public void GetErrorCode_SingleSonarIssue_ErrorCodeReturned(string fullRuleKey,
...
+        [DataRow("c:S111", "c", "S111", "SonarLint")]
+        [DataRow("cpp:S222", "cpp", "S222", "SonarLint")]
+        [DataRow("javascript:S333", "javascript", "S333", "SonarLint")]
+        [DataRow("typescript:S444", "typescript", "S444", "SonarLint")]
+        [DataRow("secrets:S555", "secrets", "S555", "SonarLint")]
+        [DataRow("foo:bar", "foo", "bar", "SonarLint")]
+        public void TryGetRuleId_FromHandle_ErrorCodeReturned(string fullRuleKey, string expectedRepo, string expectedRule, string buildTool)
+        {
+            // Note: this is a copy of TryGetRuleIdFromSelectedRow_SingleSonarIssue_ErrorCodeReturned,
+            //       but without the serviceProvider and IErrorList setup
```
duncanp-sonar, line 89: Exactly the same as the existing test, but with less setup.
I couldn't think of a good alternative to just copying the existing tests - let me know if you can.

bigfluffycookie: Hmm, I cant really think of  a better way either

bigfluffycookie (approved):

Issue #3558, opened 2023-01-23
Handle hyperlinks in rule descriptions


duncanp-sonar:
### Description

We need at add event handlers to the flow document UI if we want the embedded hyperlinks to do anything.

Note: ?the hyperlinks embedded in rule help in other SonarLint flavours don't currently do anything(?)

`NavigateUri` is a bubbling event, so we can just add a handler to the parent control/flow reader and handle all event notifications there e.g.
```
  <FlowDocumentReader x:Name="docReaderNew" Grid.Column="1" Hyperlink.RequestNavigate="HandleRequestNavigate" />
``` 

In our case, we'd add it to the top-level user control.

PR #3618, opened 2023-02-03
Add functionality for urls to be clickable in rule description window.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3558


src/Education/Commands/RuleDescriptionUserControl.xaml
```diff
@@ -118,6 +118,11 @@
             <Paragraph>By default, this tool window is docked with the
                 <Italic>Solution Explorer</Italic>. However, you can move it any where you like, just like any other tool window.
             </Paragraph>
+            <Paragraph>
```
bigfluffycookie, line 121: Please lmk if this is the correct link. Also, I have made the whole sentence clickable but I can also change it so only the "here" is.

src/Education/Commands/RuleDescriptionToolWindow.cs
```diff
@@ -37,10 +39,12 @@ internal class RuleDescriptionToolWindow : ToolWindowPane, IRuleDescriptionToolW
         public static readonly Guid ToolWindowId = new Guid(ToolWindowIdAsString);
         private readonly RuleDescriptionUserControl control;
 
-        public RuleDescriptionToolWindow()
+        public RuleDescriptionToolWindow(IServiceProvider serviceProvider)
         {
             Caption = Resources.RuleDescriptionToolWindowCaption;
-            control = new RuleDescriptionUserControl();
+            var componentModel = serviceProvider.GetService(typeof(SComponentModel)) as IComponentModel;
```
bigfluffycookie, line 45: I dont like this very much but I am thinking thats the only way I can do it

duncanp-sonar: Yep; we basically have two options:
1) this one i.e. the tool window is responsible for fetching the required service, or
2) we make the package responsible for fetching the service and passing it to the tool window when it creates it.

I would probably have gone for option (2) since the package is a "controller"/manager that is responsible for instantiating the tool window, so it doesn't feel too odd to make it responsible for supplying the tool window with the objects it needs.
It keeps the tool window simpler, and means the code for finding and creating objects is not spread out through multiple classes - it's concentrated in the packages.

However, there isn't a much in it.

src/Education/Commands/RuleDescriptionUserControl.xaml
duncanp-sonar: Yes, the link is correct.

I would have included the url in the text, to make it more obvious that clicking the link would take the user out of VS e.g. `You can browse all available rules at https://rules.sonarsource.com.`

Let's see what Marco thinks next week.

src/Education/Commands/RuleDescriptionUserControl.xaml.cs
```diff
@@ -19,14 +19,24 @@
...
+
+        internal RuleDescriptionUserControl(IBrowserService browserService)
         {
+            this.browserService = browserService;
             InitializeComponent();
         }
+
+        public void HandleRequestNavigate(object sender, RequestNavigateEventArgs e)
+        {
+            browserService.Navigate(e.Uri.AbsoluteUri);
```
duncanp-sonar, line 39: This call is happening on the UI thread, so an unhandled exception will crash VS.

However, none of the existing uses of this service have exception handling, so I guess it's ok not to have it here.

If we did add exception handling, it would make sense to put it in the browser service itself.

duncanp-sonar (commented):
Looks good. A couple of suggestions, but nothing essential.

duncanp-sonar (approved):

Issue #3619, opened 2023-02-03
Change rule metdata service to return IRuleInfo


duncanp-sonar:
### Description

The embedded rule description is not enough - more information is required.

- [x] define a new `IRuleInfo` interface with all of the required data
- [x] change the generator to write the data to disk
- [x] change the rule metadata provider to load the embedded json
- [x] change `IRuleMetadataInfo` to return `IRuleInfo`

PR #3622, opened 2023-02-03
Change embedded rule info to json

duncanp-sonar:
Part of #3619


build/tools/ExtractRuleDescFromJson/ExtractRuleDescFromJson.csproj
```diff
@@ -13,4 +13,12 @@
     <SonarQubeExclude>true</SonarQubeExclude>
   </PropertyGroup>
 
+  <!-- We're linking to the real SLVS IRuleInfo file to make sure this generator
+       code can't get out of step with the real product code
+  -->
+  <ItemGroup Label="SLVS Linked product code">
+    <Folder Include="SLVS_LinkedProductCode\" />
+    <Compile Include="..\..\..\src\Rules\IRuleInfo.cs" Link="SLVS_LinkedProductCode\IRuleHelp.cs" />
+  </ItemGroup>
```
duncanp-sonar, line 22: This change is in two parts:
1. changes to the generator tool, so it creates a json file rather than a simple text file, and
2. the corresponding changes in the product code to load the json file

This is not a breaking change; I haven't changed what is returned by the `IRuleMetadataProvider` interface yet.

build/tools/ExtractRuleDescFromJson/PluginRule.cs
```diff
@@ -18,21 +18,26 @@
...
 
-using System.Text.Json.Serialization;
-
 namespace ExtractRuleDescFromJson;
 
-public class Rule
+/// <summary>
+/// Data-class: rule properties, as defined by the plugin API
+/// </summary>
+public class PluginRule
```
duncanp-sonar, line 26: This class is only used in the generator tool, so I made the properties get-set so I didn't have to have a constructor.

I also changed the name from `Rule` to make it clear that this the version of rule data that we get from the Sonar analyzer plugin api.

build/tools/ExtractRuleDescFromJson/PluginRule.cs
```diff
@@ -18,21 +18,26 @@
...
+
+    public string? Description { get; set; }
+
+    public bool? IsActiveByDefault { get; set; }
 
-    [JsonPropertyName("key")]
-    public string Key { get; }
+    public string? Language { get; set; }
 
-    [JsonPropertyName("description")]
```
duncanp-sonar, line 36: I changed the serializer options to ignore casing of property names, so the attributes aren't required.

build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
```diff
@@ -87,26 +88,61 @@ private void PrepareDestinationDirectory()
...
     }
 
-    private void ProcessRule(Rule rule)
+    private void ProcessRule(PluginRule pluginRule)
     {
         try
         {
-            var xml = EnsureHtmlIsXml(rule.Description);
-            SaveRuleFile(rule, xml);
+            var descAsxml = EnsureHtmlIsXml(pluginRule.Description);
```
duncanp-sonar, line 104: Previously we were just saving a text file.
Now we need to create a new json file with more info in it, in a form that can be deserialized by the product code.

build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
```diff
@@ -87,26 +88,61 @@ private void PrepareDestinationDirectory()
...
+
+            var slvsRule = new RuleInfo(
+                pluginRule.Language ?? throw new ArgumentNullException("language"),
```
duncanp-sonar, line 107: We want to know at build time if there was a problem parsing the data extracted from the jars, so all of the data retrieval methods throw if data is missing.

build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
```diff
@@ -87,26 +88,61 @@ private void PrepareDestinationDirectory()
...
+            SaveRuleFile(slvsRule);
         }
         catch (Exception ex)
         {
-            Logger.LogError($"Error processing rule. Rule key: {rule.Key}, file: {context.RuleJsonFilePath}, {ex.Message}");
+            Logger.LogError($"Error processing rule. Rule key: {pluginRule.Key}, file: {context.RuleJsonFilePath}, {ex.Message}");
         }
     }
 
+    private static RuleIssueSeverity ConvertPluginSeverity(string? pluginSeverity)
```
duncanp-sonar, line 125: Rather than writing a JsonConverter, I just wrote a couple of simple methods.
I like the new pattern-matching syntax.

build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
```diff
@@ -87,26 +88,61 @@ private void PrepareDestinationDirectory()
...
+        => pluginSeverity switch
+        {
+            "MINOR" => RuleIssueSeverity.Minor,
+            "MAJOR" => RuleIssueSeverity.Major,
+            "INFO" => RuleIssueSeverity.Info,
+            "BLOCKER" => RuleIssueSeverity.Blocker,
+            "CRITICAL" => RuleIssueSeverity.Critical,
+            _ => throw new ArgumentException("Invalid enum value for pluginSeverity" + pluginSeverity, nameof(pluginSeverity)),
```
duncanp-sonar, line 133: I probably didn't need this method - the enum value names match, so I could probably have just used `Enum.Parse`.

build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
```diff
@@ -126,24 +167,32 @@ private static string EnsureHtmlIsXml(string html)
...
         // e.g. "S123.desc"
-        var colonPos = rule.Key.IndexOf(':');
+        var colonPos = slvsRule.RuleKey.IndexOf(':');
         if (colonPos == -1)
         {
-            throw new InvalidOperationException("Invalid rule key: " + rule.Key);
+            throw new InvalidOperationException("Invalid rule key: " + slvsRule.RuleKey);
         }
 
-        var ruleKeyWithoutLanguage = rule.Key.Substring(colonPos + 1);
```
duncanp-sonar, line 145: I changed the variable name `ruleKeyWithoutLanguage` since it was misleading - it's the repo key, not the language key.

src/Education.UnitTests/RuleHelpXamlBuilderTests.cs
```diff
@@ -42,7 +43,7 @@ public void Create_CheckAllEmbedded()
             // XAML document for them, but it still only takes a around 3 seconds
             // to run.
             var resourceNames = ResourceAssembly.GetManifestResourceNames()
-                .Where(x => x.EndsWith(".desc"));
+                .Where(x => x.EndsWith(".json"));
```
duncanp-sonar, line 46: Fixed up tests to look for and load the embedded `.json` files.

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -196,8 +196,8 @@ Relative path: 'SonarLint.VisualStudio.Education.dll'
 
 Referenced assemblies:
 - 'Microsoft.VisualStudio.ComponentModelHost, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
-- 'Microsoft.VisualStudio.Interop, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```
duncanp-sonar, line 199: This `Microsoft.VisualStudio.Interop` is still referenced, but it's position in the file has changed.

I'm guessing the order changed in a merge. If I remember correctly, the AsmRef tool sorts the references into alphabetical order. The new version of the file is in sorted order.

src/Rules/IRuleInfo.cs
```diff
@@ -0,0 +1,104 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+// Note: we have multiple enum definitions for IssueSeverity and IssueType in SLVS
+// They should be consolidated if possible.
+// See https://github.com/SonarSource/sonarlint-visualstudio/issues/3617
```
duncanp-sonar, line 23: I defined completely new types for severity and issue type, and a new `IRuleInfo` class that is a superset of `IRuleHelp`.

In a later PR, I'll replace `IRuleHelp` with `IRuleInfo`. I didn't want to do that in this PR as it would create too much churn.

src/Rules/IRuleMetadataProvider.cs
```diff
@@ -39,34 +42,55 @@ public interface IRuleMetadataProvider
...
                 ? Resources.Rules_DescriptionForMissingRule
                 : GetDescription(language, ruleKey);
 
             return new RuleHelp(language, ruleKey, description);
         }
 
         private string GetDescription(Language language, string ruleKey)
+        {
+            var ruleInfo = LoadRuleInfo(language, ruleKey);
+            return ruleInfo?.Description ?? Resources.Rules_DescriptionForMissingRule;
```
duncanp-sonar, line 65: I avoided changing the `IRuleMetadataProvider` interface to limit the scope of the changes in this PR.

The class now uses the new `IRuleInfo` class internally to load the embedded json, but just uses it to create and return an instance of `IRuleHelp`.

src/Rules/IRuleMetadataProvider.cs
```diff
@@ -39,34 +42,55 @@ public interface IRuleMetadataProvider
...
+                            return JsonConvert.DeserializeObject<RuleInfo>(data);
+                        }
                     }
                 }
             }
-
-            return Resources.Rules_DescriptionForMissingRule;
+            catch (Exception ex)
+            {
+                logger.LogVerbose(Resources.MetadataProvider_ErrorLoadingJson, resourcePath, ex);
```
duncanp-sonar, line 87: Difficult to test this bit...

bigfluffycookie: Hmm yea I can't think of any simple ways either :/

bigfluffycookie (approved):
Good work!

Issue #3615, opened 2023-02-03
Add title and subtitle to help doc


duncanp-sonar:
### Description

The help doc is missing a title and subtitle.

It should look like the the SonarLint help in IntelliJ e.g.

![image](https://user-images.githubusercontent.com/10757559/216789167-51f04050-e350-4f56-b17a-b6671d6f7f24.png)

PR #3628, opened 2023-02-05
Add severity and issue type images to rule help

duncanp-sonar:
Part of #3615


src/Education/Commands/RuleDescriptionUserControl.xaml
```diff
@@ -42,7 +42,14 @@
...
             <Setter Property="FontSize" Value="{DynamicResource {x:Static vsshell:VsFonts.CaptionFontSizeKey}}"/>
-            <Setter Property="FontWeight" Value="{DynamicResource {x:Static vsshell:VsFonts.CaptionFontWeightKey}}"/>
+            <Setter Property="FontWeight" Value="Bold"/>
+        </Style>
+        <Style x:Key="SubtitleElement_Image" TargetType="Image">
+            <Setter Property="Margin" Value="0,0,3,0" />
+            <Setter Property="Stretch" Value="None" />
+            <!--<Setter Property="Stretch" Value="UniformToFill" />
+            <Setter Property="Height" Value="12" />
+            <Setter Property="Width" Value="12" />-->
```
duncanp-sonar, line 52: I added another style so we can control the appearance of the images from the XAML.
I left the properties setters for `Stretch`, `Height` and `Width` for now as a reminder of alternatives for controlling the image size in case it needs any tweaking.

src/Education/Commands/RuleDescriptionUserControl.xaml
```diff
@@ -130,7 +137,8 @@
             <Paragraph>By default, this tool window is docked with the
                 <Italic>Solution Explorer</Italic>. However, you can move it any where you like, just like any other tool window.
             </Paragraph>
-            <Paragraph>You can browse all available rules at <Hyperlink NavigateUri="https://rules.sonarsource.com/">https://rules.sonarsource.com</Hyperlink>
+            <Paragraph TextAlignment="Left">You can browse all available rules at
```
duncanp-sonar, line 140: Miscellaneous: changed the text alignment to `Left`. Before, it was being justified, which looked at bit weird depending on the width of the tool window.

src/Education/Education.csproj
```diff
@@ -29,6 +29,14 @@
...
   </ItemGroup>
 
+  <ItemGroup Label="Images">
+    <None Remove="XamlGenerator\images\severity\*.png" />
+    <Resource Include="XamlGenerator\images\severity\*.png" />
+
+    <None Remove="XamlGenerator\images\type\*.png" />
+    <Resource Include="XamlGenerator\images\type\*.png" />
+  </ItemGroup>
+
```
duncanp-sonar, line 39: I copied all of the relevant images from SLI. There are multiple versions of each each in different sizes; currently we only use one size of each.

src/Education/XamlGenerator/IRuleHelpXamlBuilder.cs
```diff
@@ -170,14 +170,33 @@ private void WriteSubTitle(IRuleInfo ruleInfo)
...
+
+        private void WriteSubTitleElementWithImage(SubTitleImageInfo imageInfo)
+        {
+            writer.WriteStartElement("Span");
+            ApplyStyleToStartElement(StyleResourceNames.SubtitleElement_Span);
+
+            if (imageInfo.ImageResourcePath != null)
+            {
+                writer.WriteStartElement("InlineUIContainer");
+                writer.WriteStartElement("Image");
```
duncanp-sonar, line 191: `FlowDocument` can only contain elements deriving from `TextElement` e.g. `Paragraph`, `Section`, `Inline`.

`Image` is a "normal" XAML UI control that inherits from `UIElement`, which means it can't be hosted directly in the document. So, we have to add an `InlineUIContainer` first, which allows us to host non-`TextElement` XAML classes in documents.

There is also a `BlockUIContainer` class that we could use to host "normal" XAML classes at `Block` level if we needed to.

src/Education/XamlGenerator/SubTitleImageInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <summary>
+    /// Data class - holds the image path and display text to use for sub-element items
+    /// with icons i.e. severity and issue type
+    /// </summary>
+    public class SubTitleImageInfo
+    {
+        public static readonly IReadOnlyDictionary<RuleIssueSeverity, SubTitleImageInfo> SeverityImages = new Dictionary<RuleIssueSeverity, SubTitleImageInfo>()
+        {
+            { RuleIssueSeverity.Critical, new SubTitleImageInfo("/SonarLint.VisualStudio.Education;component/XamlGenerator/images/severity/critical.png", "Critical" )},
+            { RuleIssueSeverity.Blocker, new SubTitleImageInfo("/SonarLint.VisualStudio.Education;component/XamlGenerator/images/severity/blocker.png", "Blocker" )},
```
duncanp-sonar, line 35: The text we want to display in the UI and the name of the embedded image file don't always directly map to the corresponding enum name, so I added these dictionaries as lookups to find the values to use.

e.g. `Blocker` is the same in all cases, but for code smells the enum value is `CodeSmell`, whereas the file is called `code_smell` and we want to display `Code Smell` in the UI.

src/Education/XamlGenerator/SubTitleImageInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+            { RuleIssueSeverity.Minor, new SubTitleImageInfo("/SonarLint.VisualStudio.Education;component/XamlGenerator/images/severity/minor.png", "Minor" )},
+            { RuleIssueSeverity.Info, new SubTitleImageInfo("/SonarLint.VisualStudio.Education;component/XamlGenerator/images/severity/info.png", "Info" )}
+        };
+
+        public static readonly IReadOnlyDictionary<RuleIssueType, SubTitleImageInfo> IssueTypeImages = new Dictionary<RuleIssueType, SubTitleImageInfo>()
+        {
+            { RuleIssueType.Vulnerability, new SubTitleImageInfo("/SonarLint.VisualStudio.Education;component/XamlGenerator/images/type/vulnerability.png", "Vulnerability" )},
+            { RuleIssueType.CodeSmell, new SubTitleImageInfo("/SonarLint.VisualStudio.Education;component/XamlGenerator/images/type/code_smell.png", "Code Smell" )},
+            { RuleIssueType.Bug, new SubTitleImageInfo("/SonarLint.VisualStudio.Education;component/XamlGenerator/images/type/bug.png", "Bug" )},
+            { RuleIssueType.Hotspot, new SubTitleImageInfo(null /* TODO - image for hotspot */, "Hotspot" )},
```
duncanp-sonar, line 46: SLI doesn't have an icon for hotspot yet.
However, we're only displaying help for "normal", non-hotspot rules at the moment, so it won't matter. Even if we were displaying hotspots, only the icon would be missing. The `Hotspot` text would still appear.

src/Education/Commands/RuleDescriptionUserControl.xaml
bigfluffycookie: Ah good spot!

bigfluffycookie (approved):
Yay Icons!

PR #3635, opened 2023-02-06
Fix alignment of icons

bigfluffycookie:
part of https://github.com/SonarSource/sonarlint-visualstudio/issues/3615

Before:
before

After:
result


duncanp-sonar (approved):
Nice - looks much better!

PR #3638, opened 2023-02-06
Handle unknown rules in IEducation.ShowRuleHelp

### LINKED ISSUE #3637: Handle unknown rules in IEducation.ShowRuleHelp
### Description

Previously, if the `IRuleMetadataProvider` service was asked for an unknown rule it would return a simple "rule not found" message. Now it returns `null`, which will cause `Education.ShowRuleHelp` would cause throw.

There shouldn't be any circumstances in which the education service is asked to display help for an unknown rule, but it would be safer to handle this case explicitly.

duncanp-sonar:
Fixes #3637

### DIFF
--- a/src/Education/IEducation.cs
+++ b/src/Education/IEducation.cs
@@ -69,6 +69,12 @@ namespace SonarLint.VisualStudio.Education
         {
             var ruleInfo = ruleMetadataProvider.GetRuleInfo(ruleId);
 
+            if (ruleInfo == null)
+            {
+                logger.WriteLine(Resources.Education_NoRuleInfo, ruleId.ErrorListErrorCode);
+                return;
+            }
+
             threadHandling.RunOnUIThread(() =>
             {
                 var flowDocument = ruleHelpXamlBuilder.Create(ruleInfo);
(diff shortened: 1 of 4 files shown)


src/Education/Resources.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="Education_NoRuleInfo" xml:space="preserve">
+    <value>[Education] Unable to locate help for rule '{0}'. Help should be available at https://rules.sonarsource.com.</value>
```
duncanp-sonar, line 121: This message is pretty much as the one that used to be returned by `IRuleMetadataProvider`.

The difference is that this message is only shown in the Output Window, whereas previously the tool window document was updated.

However, updating the tool window would take more work; this case shouldn't really occur anyway, we're just being cautious; and we're trying to finish of the sprint, so we should be minimising the amount of change at this stage.

src/Education/IEducation.cs
```diff
@@ -69,6 +69,12 @@ public void ShowRuleHelp(SonarCompositeRuleId ruleId)
         {
             var ruleInfo = ruleMetadataProvider.GetRuleInfo(ruleId);
 
+            if (ruleInfo == null)
+            {
+                logger.WriteLine(Resources.Education_NoRuleInfo, ruleId.ErrorListErrorCode);
+                return;
+            }
```
duncanp-sonar, line 76: Just erring on the side of caution. This shouldn't happen in practice, but if it does we end with a null-ref exception on the UI thread.

bigfluffycookie (approved):

PR #3641, opened 2023-02-06
Lazily fetch the tool window on the UI thread

### LINKED ISSUE #3640: Ensure `Education` constructor follows threading rules
### Description

Currently we indirectly call the `IVsUIShell` service from the constructor without checking which thread we are on (this might be contributing the issue in #3620).

We have an instance of `IThreadHandling` so we could ensure we are on the UI thread before making the call.
Alternatively, we could delay fetching the tool window until we need to display it.

duncanp-sonar:
Fixes #3640

### DIFF
--- a/src/Education.UnitTests/EducationTests.cs
+++ b/src/Education.UnitTests/EducationTests.cs
@@ -44,13 +44,13 @@ namespace SonarLint.VisualStudio.Education.UnitTests
         }
 
         [TestMethod]
-        public void Ctor_GetsToolWindowFromService()
+        public void Ctor_DoesNotFetchToolWindow()
         {
             var toolWindowService = new Mock<IToolWindowService>();
 
             _ = CreateEducation(toolWindowService: toolWindowService.Object);
 
-            toolWindowService.Verify(x => x.GetToolWindow<RuleHelpToolWindow, IRuleHelpToolWindow>(), Times.Once);
+            toolWindowService.Invocations.Should().HaveCount(0);
         }
 
         [TestMethod]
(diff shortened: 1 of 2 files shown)


src/Education.UnitTests/EducationTests.cs
```diff
@@ -72,6 +72,10 @@ public void ShowRuleHelp_EverythingGetsCalledCorrectly()
             toolWindowService.Setup(x => x.GetToolWindow<RuleHelpToolWindow, IRuleHelpToolWindow>()).Returns(ruleDescriptionToolWindow.Object);
 
             var testSubject = CreateEducation(toolWindowService: toolWindowService.Object, ruleMetadataProvider: ruleMetaDataProvider.Object, ruleHelpXamlBuilder: ruleHelpXamlBuilder.Object);
+            // Sanity check - tool window not yet fetched
+            toolWindowService.Invocations.Should().HaveCount(0);
```
duncanp-sonar, line 76: This makes the old test `Ctor_GetsToolWindowFromService` redundant.

src/Education/IEducation.cs
```diff
@@ -61,8 +62,6 @@ public Education(IToolWindowService toolWindowService, IRuleMetadataProvider rul
             this.ruleMetadataProvider = ruleMetadataProvider;
             this.logger = logger;
             this.threadHandling = threadHandling;
-
-            ruleHelpToolWindow = toolWindowService.GetToolWindow<RuleHelpToolWindow, IRuleHelpToolWindow>();
```
duncanp-sonar, line 65: We should be on the UI thread when making this since it indirectly calls through to `IVsUIShell`.

We have an `IThreadHandling` instance so we could switch to the UI thread here. However, we already have a switch to the UI thread in the `ShowRuleHelp` method, so we can just refer the work until then i.e. lazily-fetch the tool window.

PR #3643, opened 2023-02-06
Drop show error help command interceptor

### LINKED ISSUE #3642: Drop `Show Error Help` command interceptor
### Description

The `Error List` routes the `Show Error Help` context menu command to the MEF-based event processor, so we don't need both.

This also means we can close #3620.

duncanp-sonar:
Fixes #3642


duncanp-sonar:
This is a straightforward deletion of one class, so I'll go ahead and commit it so I can test the final artefacts.

PR #3647, opened 2023-02-07
Make the images slightly less blurry.

bigfluffycookie:
Before 
before1

After
after1


duncanp-sonar (approved):
Nice - looks much better!

Issue #3651, opened 2023-02-07
[RuleHelp] Clicking on Sonar hyperlinks in issue tooltips should show in-IDE rule help 


duncanp-sonar:
### Description

Clicking on the error code hyperlink in an issue tooltip should show the in-IDE rule help, if possible.

![image](https://user-images.githubusercontent.com/10757559/217317971-6f36706f-2002-44b2-8b4e-0f124f2c481c.png)

Currently it still navigates to `https://rules.sonarsource.com`.

PR #3652, opened 2023-02-07
Change IEducation handling of non-local rules

duncanp-sonar:
- changed to navigate to rules.sonarsource.com

Prep for #3651


src/Education/InternalsVisibleTo.cs
```diff
@@ -22,8 +22,10 @@
 
 #if SignAssembly
 [assembly: InternalsVisibleTo("SonarLint.VisualStudio.Education.UnitTests,PublicKey=002400000480000094000000060200000024000052534131000400000100010081b4345a022cc0f4b42bdc795a5a7a1623c1e58dc2246645d751ad41ba98f2749dc5c4e0da3a9e09febcb2cd5b088a0f041f8ac24b20e736d8ae523061733782f9c4cd75b44f17a63714aced0b29a59cd1ce58d8e10ccdb6012c7098c39871043b7241ac4ab9f6b34f183db716082cd57c1ff648135bece256357ba735e67dc6")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```
duncanp-sonar, line 25: We're now mocking an internal class, so we need to give Moq access to it.

src/Education/IEducation.cs
```diff
@@ -68,34 +77,43 @@ public void ShowRuleHelp(SonarCompositeRuleId ruleId)
...
-            }
-
             threadHandling.RunOnUIThread(() =>
             {
-                // Lazily fetch the tool window from a UI thread
-                if (ruleHelpToolWindow == null)
+                if (ruleInfo == null)
                 {
-                    ruleHelpToolWindow = toolWindowService.GetToolWindow<RuleHelpToolWindow, IRuleHelpToolWindow>();                    
+                    showRuleInBrowser.ShowRuleDescription(ruleId);
```
duncanp-sonar, line 84: If we can't find the rule info locally, fallback on showing it in the IDE.

This is a better general behaviour that doing nothing. 
More importantly, it means that `IEducation` will now also be able to handle requests for taint rules. We don't have the info locally, but at least it will be shown in the browser.

bigfluffycookie (approved):

Issue #3663, opened 2023-02-08
Add Secrets Language


bigfluffycookie:
Add secrets as a language.

PR #3664, opened 2023-02-08
Add secrets to languages

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3663

### DIFF
--- a/src/Core.UnitTests/LanguageTests.cs
+++ b/src/Core.UnitTests/LanguageTests.cs
@@ -67,6 +67,9 @@ namespace SonarLint.VisualStudio.Core.UnitTests
             Language.VBNET.IsSupported.Should().BeTrue();
             Language.Cpp.IsSupported.Should().BeTrue();
             Language.C.IsSupported.Should().BeTrue();
+            Language.Js.IsSupported.Should().BeTrue();
+            Language.Ts.IsSupported.Should().BeTrue();
+            Language.Secrets.IsSupported.Should().BeTrue();
         }
 
         [TestMethod]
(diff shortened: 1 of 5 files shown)


src/Core.UnitTests/LanguageTests.cs
```diff
@@ -67,6 +67,9 @@ public void Language_IsSupported_SupportedLanguage_IsTrue()
             Language.VBNET.IsSupported.Should().BeTrue();
             Language.Cpp.IsSupported.Should().BeTrue();
             Language.C.IsSupported.Should().BeTrue();
```
bigfluffycookie, line 69: I felt weird only adding the secret tests so I added the js and ts stuff aswell.

bigfluffycookie:
Unsure if I updated the submodule correctly but I think i did not

src/Core.UnitTests/LanguageTests.cs
duncanp-sonar: Thanks! If you spot a test hole when changing a component, it's a practice to add some tests if it doesn't take too much effort.

src/Integration/Binding/BindingController.cs
```diff
@@ -178,7 +178,8 @@ void IBindingWorkflowExecutor.BindProject(BindCommandArgs bindingArgs)
                     Language.C,
                     Language.Cpp,
                     Language.Js,
-                    Language.Ts
+                    Language.Ts,
+                    Language.Secrets
```
duncanp-sonar, line 182: Did you mean to include this change in the PR?

duncanp-sonar:
> Unsure if I updated the submodule correctly but I think i did not

I think it's ok. We can discuss it tomorrow.

In hindsight, we should have created a feature branch in the `sonarqube-webclient-dotnet` repo and merged the language change to that first (in the same way Rita and Georgii have a `branch-sse` feature branch  in that repo for their changes).
However, our change is minor and non-impacting, so it won't have done any harm.

src/Integration/Binding/BindingController.cs
bigfluffycookie: Hm actually, I am not sure as I dont fully understand what the binding config provider is. Do you think it makes more sense to include it in the next pr?

duncanp-sonar (approved):

PR #3666, opened 2023-02-08
Initialize SSESessionManager

rita-sonarsource:
Need the VSIX package to mef-import ISSESessionManager, in order for the whole module to start.

Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3669


src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
```diff
@@ -28,7 +28,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents
     /// <summary>
     /// Reacts to project changes and opens/closes Server Sent Events sessions
     /// </summary>
-    internal interface ISSESessionManager : IDisposable
+    public interface ISSESessionManager : IDisposable
```
rita-sonarsource, line 31: Perhaps we should move the interface to Core. I'm not sure. IntegrationVSIX needs to call it so the listeners would be hooked up.

src/Core/Core.csproj
```diff
@@ -11,12 +11,9 @@
 
     <RequiresNewtonsoft>true</RequiresNewtonsoft>
     <RequiresSystemIOAbstractions>true</RequiresSystemIOAbstractions>
+    <RequiresChannels>true</RequiresChannels>
```
rita-sonarsource, line 14: Our convention for some packages that are references in more than one place, to make sure we're referencing the same version.

src/Core/ServerSentEvents/Issues/IssueChangedServerEventChannel.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System.ComponentModel.Composition;
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
+
+namespace SonarLint.VisualStudio.Core.ServerSentEvents.Issues
+{
+    [Export(typeof(IIssueChangedServerEventSource))]
+    [Export(typeof(IIssueChangedServerEventSourcePublisher))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class IssueChangedServerEventChannel : ServerEventChannel<IIssueChangedServerEvent>, IIssueChangedServerEventSource, IIssueChangedServerEventSourcePublisher
```
rita-sonarsource, line 29: Dummy implementation; it probably shouldn't be in Core but I left it for now so that the MEF import works.

src/Core/ServerSentEvents/TaintVulnerabilities/TaintServerEventChannel.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System.ComponentModel.Composition;
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
+
+namespace SonarLint.VisualStudio.Core.ServerSentEvents.TaintVulnerabilities
+{
+    [Export(typeof(ITaintServerEventSource))]
+    [Export(typeof(ITaintServerEventSourcePublisher))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class TaintServerEventChannel : ServerEventChannel<ITaintServerEvent>, ITaintServerEventSource, ITaintServerEventSourcePublisher
```
rita-sonarsource, line 29: Same comment as for IssueChangedServerEventChannel; doing this just for the MEF export.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -441,6 +442,8 @@
       <VSIXSourceItem Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' == 'System.IO.Abstractions'" />
       <VSIXSourceItem Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' == 'BouncyCastle'" />
       <VSIXSourceItem Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' == 'LibGit2Sharp'" />
+      <VSIXSourceItem Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' == 'System.Threading.Channels'" />
```
rita-sonarsource, line 445: Very ugly and without this, the assemblies are not included in the VSIX and fail on runtime.

src/Integration.Vsix/SonarLintNotificationsPackage.cs
```diff
@@ -75,6 +76,8 @@ private async System.Threading.Tasks.Task InitAsync()
             logger = await this.GetMefServiceAsync<ILogger>();
             logger.WriteLine(Resources.Strings.Notifications_Initializing);
 
+            sseSessionManager = await this.GetMefServiceAsync<ISSESessionManager>();
```
rita-sonarsource, line 79: Probably not the right package for this. Not sure where else to put it. Our packages are a bit of a mess.

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
duncanp-sonar: Or `ConnectedMode`?
All of these classes are specific to Connected Mode - could they all be moved there?

It would mean `Connected Mode` would have to reference VS, but I think that's ok.

duncanp-sonar (commented):
Looks ok. Let's have a discussion about whether to put more stuff into the `ConnectedMode` project.

From a design point of view, I think we have a couple of high-level goals:
1) co-locate feature code as much as possible (or to put it another way, isolate separate features as much as possible), and 

2) isolate our code from VS as much as possible.

These are not mutually exclusive but can cause conflicts.
(1) suggests putting the new code in `ConnectedMode`, but that makes it harder to achieve (2): `ConnectedMode` would then need to reference the VSSSDK so it would be harder to stop VS types leaking into classes that could be VS-independent.

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
rita-sonarsource: It's already in ConnectedMode, and VSIX indeed references it now. Is that what you meant?

duncanp-sonar: Yes - put everything in `ConnectedMode`

duncanp-sonar: As discussed offline, we could have `IssueViz.Security` depend on `ConnectedMode` and move everything else out of `Core` and all of the non-packaging stuff out of `Integration.VSIX`.

src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.VisualStudio.Shell;
+using SonarLint.VisualStudio.ConnectedMode.ServerSentEvents;
+using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.ConnectedMode
+{
+    [ExcludeFromCodeCoverage]
+    [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
+    [ProvideAutoLoad(BoundSolutionUIContext.GuidString, PackageAutoLoadFlags.BackgroundLoad)]
```
rita-sonarsource, line 36: This will start only when we're in connected mode. Same logic as in the taintSync package.

src/Core/Core.csproj
duncanp-sonar: Does `Core` still need channels?

src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -0,0 +1,65 @@
...
+            var componentModel = await GetServiceAsync(typeof(SComponentModel)) as IComponentModel;
+            var logger = componentModel.GetService<ILogger>();
+
+            logger.WriteLine(Resources.Package_Initializing);
+
+            sseSessionManager = componentModel.GetService<ISSESessionManager>();
+
+            logger.WriteLine(Resources.Package_Initialized);
+
+            await JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
```
duncanp-sonar, line 53: The thread transition should happen first.

src/Core/Core.csproj
duncanp-sonar: Ah - yes, because the events are still in `Core`. Could they be in `ConnectedMode`?

src/ConnectedMode/ConnectedMode.csproj
```diff
@@ -3,12 +3,27 @@
...
+  </PropertyGroup>
+
   <PropertyGroup>
     <ProjectGuid>0BE551DB-3C46-42A5-BB38-DA80E83F8ABD</ProjectGuid>
     <AssemblyName>SonarLint.VisualStudio.ConnectedMode</AssemblyName>
     <RootNamespace>SonarLint.VisualStudio.ConnectedMode</RootNamespace>
-
     <RequiresLibGit>true</RequiresLibGit>
+
+    <CopyBuildOutputToOutputDirectory>true</CopyBuildOutputToOutputDirectory>
```
duncanp-sonar, line 20: I thought this would be `true` by default (even if it is, being explicit doesn't hurt)

duncanp-sonar (commented):
See inline: the package initialize thread transition is in the wrong place.

src/Core/Core.csproj
rita-sonarsource: I will revisit once I implement the Taint consumer

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
rita-sonarsource: I will revisit once I implement the Taint consumer

PR #3667, opened 2023-02-08
Initialize SSESession if already in connected mode when loaded

rita-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3668

### DIFF
--- a/src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
+++ b/src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
@@ -50,6 +50,8 @@ namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents
             this.sseSessionFactory = sseSessionFactory;
 
             activeSolutionBoundTracker.SolutionBindingChanged += SolutionBindingChanged;
+
+            CreateSessionIfInConnectedMode(activeSolutionBoundTracker.CurrentConfiguration);
         }
 
         public void Dispose()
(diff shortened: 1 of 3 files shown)


src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
```diff
@@ -50,6 +50,8 @@ public SSESessionManager(IActiveSolutionBoundTracker activeSolutionBoundTracker,
             this.sseSessionFactory = sseSessionFactory;
 
             activeSolutionBoundTracker.SolutionBindingChanged += SolutionBindingChanged;
+
+            CreateSessionIfInConnectedMode(activeSolutionBoundTracker.CurrentConfiguration);
```
rita-sonarsource, line 54: @georgii-borovinskikh-sonarsource spotted this bug: due to the async nature of the packages load order, it's possible that when this module starts, we're already in connected mode and missed the event. So we need to also explicitly check in the ctor if we're in connected mode, and if so, start a new session.

I wanted to write tests for this, but due to the nature of TestScope, I couldn't -- as the TestScope's ctor already initializes the mocks and the class, not giving me a hook. I can pass parameters to TestScope's ctor, but it starts to feel hacky and makes the tests not clear. LMK what you think.

georgii-borovinskikh-sonarsource: Well, a good solution would be to move the TestSubject creation from the test scope constructor to a different method (aka CreateTestSubject) so you could setup the mocks before you create it

georgii-borovinskikh-sonarsource: Or make the TestSubject -> Lazy<TestSubject>, so it's less copypaste

Issue #3670, opened 2023-02-09
Create interface for IConnectedModeSecrets


bigfluffycookie:
Create a new class that checks if secrets are available in connected mode.

PR #3671, opened 2023-02-09
Add interface IConnecteDModeSecrets which checks if secrets are available in connected mode.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3670


src/Core/Secrets/IConnectedModeSecrets.cs
```diff
@@ -0,0 +1,27 @@
+﻿/*
```
bigfluffycookie, line 1: I made a new folder "Secrets" in core hope thats oki

duncanp-sonar: I would have done the same, even though there is currently only one file in it.

src/CloudSecrets/ConnectedModeSecrets.cs
```diff
@@ -0,0 +1,52 @@
...
+            this.sonarQubeService = sonarQubeService;
+        }
+
+        public bool AreSecretsAvailable()
+        {
+            if (sonarQubeService.IsConnected)
+            {
+                var serverInfo = sonarQubeService.GetServerInfo();
+
+                return serverInfo.ServerType == ServerType.SonarCloud || serverInfo.Version >= MinimumRequiredSonarQubeVersion;
```
bigfluffycookie, line 46: Logically, if the servertype is not sonarcloud it has to be sonarqube however, I am not sure if it might be better to write it like this so its clearer:

```
serverInfo.ServerType == ServerType.SonarCloud ||
 (serverInfo.ServerType == ServerType.SonarQube && serverInfo.Version >= MinimumRequiredSonarQubeVersion);
```

src/CloudSecrets.UnitTests/ConnectedModeSecretsTests.cs
```diff
@@ -0,0 +1,118 @@
+﻿/*
```
bigfluffycookie, line 1: Please let me know if this & the other class were supposed to go into the folder CloudSecrets or not.

src/Core/Secrets/IConnectedModeSecrets.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace SonarLint.VisualStudio.Core.Secrets
+{
+    public interface IConnectedModeSecrets
+    {
+        /// <summary>
+        /// Returns true if the project is bound to either SonarCloud, or to SonarQube version >= 9.9
+        /// Returns false if project is unbound or bound to SonarQube version <= 9.8
+        /// </summary>
+        bool AreSecretsAvailable();
```
duncanp-sonar, line 29: The summary is a bit too detailed - I would probably have made it simpler e.g.

```
<summary>Returns whether Secrets are supported in Connected Mode</summary>
```

... and moved your existing comment to `<remarks>` section.

src/CloudSecrets/ConnectedModeSecrets.cs
duncanp-sonar: Hhhmm. I don't have a strong preference, but for me the more-explicit version is easier to understand because it is more explicit and requires less thinking. Go with whichever you prefer.

src/CloudSecrets.UnitTests/ConnectedModeSecretsTests.cs
```diff
@@ -0,0 +1,118 @@
...
+            var testSubject = new ConnectedModeSecrets(sonarQubeService.Object);
+
+            var result = testSubject.AreSecretsAvailable();
+
+            sonarQubeService.Verify(x => x.IsConnected, Times.Once);
+            sonarQubeService.Verify(x => x.GetServerInfo(), Times.Once);
+            result.Should().BeTrue();
+        }
+
+        private ServerInfo CreateServerInfo(ServerType serverType, Version version = null)
```
duncanp-sonar, line 110: You could expand this method (or create another one that wraps it) that creates and returns the configure `SonarQubeService`. That would simplify the setup of each test quite a bit.

src/CloudSecrets.UnitTests/ConnectedModeSecretsTests.cs
```diff
@@ -0,0 +1,118 @@
...
+            result.Should().BeTrue();
+        }
+
+        [TestMethod]
+        public void AreSecretsAvailable_ProjectIsConnectedToSonarQube_BelowRequiredVersion_NotAvailable()
+        {
+            var sonarQubeService = new Mock<ISonarQubeService>();
+            sonarQubeService.Setup(x => x.IsConnected).Returns(true);
+
+            var serverInfo = CreateServerInfo(ServerType.SonarQube, new Version(1, 5));
```
duncanp-sonar, line 80: It would be better to test the boundary condition i.e. the highest version that should not be supported e.g. 9.8. That would check that the product code is doing the comparison correctly.

One option is to have a single data-driven test e.g.

`
void AreSecretsAvailable_ConnectedToSonarQube_CheckPluginVersion(string version, bool expectedResult)
`

src/CloudSecrets.UnitTests/ConnectedModeSecretsTests.cs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<ConnectedModeSecrets, IConnectedModeSecrets>(
+                MefTestHelpers.CreateExport<ISonarQubeService>());
+        }
+
+        [TestMethod]
+        public void AreSecretsAvailable_ProjectIsUnbound_NotAvailable()
```
duncanp-sonar, line 42: Strictly speaking the test names are a bit misleading - it's not checking whether the projects are bound or not, it's checking whether the `SonarQubeClient` is connected. Using something like `ServiceIsConnected/IsNotConnected` would be clearer.

duncanp-sonar (approved):
Looks good. I'd just rename the tests that refer to `Project` in the name.

PR #3679, opened 2023-02-10
Add methods Add/Remove to ITaintStore

rita-sonarsource:
Fixes [#3678](https://github.com/SonarSource/sonarlint-visualstudio/issues/3678)

### DIFF
--- a/src/IssueViz.Security/Taint/TaintStore.cs
+++ b/src/IssueViz.Security/Taint/TaintStore.cs
@@ -59,13 +59,23 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
     [PartCreationPolicy(CreationPolicy.Shared)]
     internal sealed class TaintStore : ITaintStore
     {
+        public event EventHandler<IssuesChangedEventArgs> IssuesChanged;
+
+        private static readonly object Locker = new object();
+
         private List<IAnalysisIssueVisualization> taintVulnerabilities = Array.Empty<IAnalysisIssueVisualization>().ToList();
         private AnalysisInformation analysisInformation;
 
-        public IReadOnlyCollection<IAnalysisIssueVisualization> GetAll() => taintVulnerabilities;
-        
+        public IReadOnlyCollection<IAnalysisIssueVisualization> GetAll()
+        {
+            lock (Locker)
+            {
+                return taintVulnerabilities;
+            }
+        }
+
         public AnalysisInformation GetAnalysisInformation() => analysisInformation;
-        
+
         public void Add(IAnalysisIssueVisualization issueVisualization)
         {
             if (issueVisualization == null)
@@ -78,9 +88,12 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
                 return;
             }
 
-            taintVulnerabilities.Add(issueVisualization);
+            lock (Locker)
+            {
+                taintVulnerabilities.Add(issueVisualization);
 
-            NotifyIssuesChanged(Array.Empty<IAnalysisIssueVisualization>(), new[] {issueVisualization});
+                NotifyIssuesChanged(Array.Empty<IAnalysisIssueVisualization>(), new[] { issueVisualization });
+            }
         }
 
         public void Remove(string issueKey)
@@ -90,21 +103,22 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
                 throw new ArgumentNullException(nameof(issueKey));
             }
 
-            var indexToRemove =
-                taintVulnerabilities.FindIndex(issueViz => ((ITaintIssue) issueViz.Issue).IssueKey.Equals(issueKey));
-
-            if (indexToRemove == -1)
+            lock (Locker)
             {
-                return;
+                var indexToRemove =
+                    taintVulnerabilities.FindIndex(issueViz => ((ITaintIssue)issueViz.Issue).IssueKey.Equals(issueKey));
+
+                if (indexToRemove == -1)
+                {
+                    return;
+                }
+
+                var valueToRemove = taintVulnerabilities[indexToRemove];
+                taintVulnerabilities.RemoveAt(indexToRemove);
+
+                NotifyIssuesChanged(new[] { valueToRemove }, Array.Empty<IAnalysisIssueVisualization>());
             }
-
-            var valueToRemove = taintVulnerabilities[indexToRemove];
-            taintVulnerabilities.RemoveAt(indexToRemove);
-
-            NotifyIssuesChanged(new[] {valueToRemove}, Array.Empty<IAnalysisIssueVisualization>());
         }
-
-        public event EventHandler<IssuesChangedEventArgs> IssuesChanged;
 
         public void Set(IEnumerable<IAnalysisIssueVisualization> issueVisualizations, AnalysisInformation analysisInformation)
         {
@@ -113,15 +127,18 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
                 throw new ArgumentNullException(nameof(issueVisualizations));
             }
 
-            this.analysisInformation = analysisInformation;
+            lock (Locker)
+            {
+                this.analysisInformation = analysisInformation;
 
-            var oldIssues = taintVulnerabilities;
-            taintVulnerabilities = issueVisualizations.ToList();
+                var oldIssues = taintVulnerabilities;
+                taintVulnerabilities = issueVisualizations.ToList();
 
-            var removedIssues = oldIssues.Except(taintVulnerabilities).ToArray();
-            var addedIssues = taintVulnerabilities.Except(oldIssues).ToArray();
+                var removedIssues = oldIssues.Except(taintVulnerabilities).ToArray();
+                var addedIssues = taintVulnerabilities.Except(oldIssues).ToArray();
 
-            NotifyIssuesChanged(removedIssues, addedIssues);
+                NotifyIssuesChanged(removedIssues, addedIssues);
+            }
         }
 
         private void NotifyIssuesChanged(
--- a/src/IssueViz.Security/Taint/TaintStore.cs
+++ b/src/IssueViz.Security/Taint/TaintStore.cs
@@ -63,14 +63,14 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
 
         private static readonly object Locker = new object();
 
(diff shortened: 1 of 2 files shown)


src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -39,21 +40,85 @@ internal interface ITaintStore : IIssuesStore
...
     [Export(typeof(IIssuesStore))]
     [PartCreationPolicy(CreationPolicy.Shared)]
     internal sealed class TaintStore : ITaintStore
     {
-        private IAnalysisIssueVisualization[] taintVulnerabilities = Array.Empty<IAnalysisIssueVisualization>();
+        public event EventHandler<IssuesChangedEventArgs> IssuesChanged;
+
+        private static readonly object Locker = new object();
+
+        private List<IAnalysisIssueVisualization> taintVulnerabilities = Array.Empty<IAnalysisIssueVisualization>().ToList();
```
duncanp-sonar, line 66: Just `new List<...>` would do, wouldn't it?

rita-sonarsource: This is why we do code reviews :D

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -39,21 +40,85 @@ internal interface ITaintStore : IIssuesStore
...
         private AnalysisInformation analysisInformation;
 
-        public IReadOnlyCollection<IAnalysisIssueVisualization> GetAll() => taintVulnerabilities;
-        
+        public IReadOnlyCollection<IAnalysisIssueVisualization> GetAll()
+        {
+            lock (Locker)
```
duncanp-sonar, line 71: I don't think the lock here provides any protection; even though the return type is `IReadOnlyCollection<>`, we are returning the real `List` object, so if we modify later in this class any consumers will see the modified list.

Somewhat surprisingly, even calling `.AsReadOnly()` returns a copy that can be modified:
```
var list = new List<string>() { "1", "2" };
IReadOnlyCollection<string> readOnly = list.AsReadOnly();
Console.WriteLine(($"readOnly count: {readOnly.Count}")); // Writes "2"
list.Add("3");
Console.WriteLine(($"readOnly count: {readOnly.Count}")); // Writes "3"
```

If we really want to return a fixed list we should explicitly copy the list and return the copy.

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -39,21 +40,85 @@ internal interface ITaintStore : IIssuesStore
...
 
-        public event EventHandler<IssuesChangedEventArgs> IssuesChanged;
+        public void Add(IAnalysisIssueVisualization issueVisualization)
+        {
+            if (issueVisualization == null)
+            {
+                throw new ArgumentNullException(nameof(issueVisualization));
+            }
+
+            if (analysisInformation == null)
```
duncanp-sonar, line 86: I think there are a couple of possible problems here.

It looks like the `TaintSynchronizer` will clear the store when a standalone solution is opened, and fetch taint issues when a connected solution is opened/updated. Until the fetch has been completed successfully, `analysisInformation` will be null.

A) possible race condition: we get a taint event notification before the `TaintSynchronizer` has fetched the initial set of taint issues -> `analysisInformation` will still be null -> event will be lost.

B) the `TaintSynchronizer` has finished its initial sync, but there are no taint issues to start with, so `analysisInformation` never gets set. A taint issue is added to the server -> taint event -> `analysisInformation` is `null` -> event is ignored.

I think (A) is a corner case that we can probably ignore. Hopefully the ignored event would be for a taint issue that would be in the batch fetch anyway.

(B) is a problem, but should be easy enough to fix.

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -39,21 +40,85 @@ internal interface ITaintStore : IIssuesStore
...
+            lock (Locker)
+            {
+                return taintVulnerabilities;
+            }
+        }
+
         public AnalysisInformation GetAnalysisInformation() => analysisInformation;
 
-        public event EventHandler<IssuesChangedEventArgs> IssuesChanged;
+        public void Add(IAnalysisIssueVisualization issueVisualization)
```
duncanp-sonar, line 79: Don't we get the taint event notifications in batch? If so, do we want to update the store in batch, or one item at a time?

Batch would be more efficient, but one at a time is probably ok for now. It might trigger multiple UI updates if the taint tool window is visible,  but the number of events should be relatively small so it should cause a problem.

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -39,21 +40,85 @@ internal interface ITaintStore : IIssuesStore
...
+            }
+
+            if (analysisInformation == null)
+            {
+                return;
+            }
+
+            lock (Locker)
+            {
+                taintVulnerabilities.Add(issueVisualization);
```
duncanp-sonar, line 93: Is it possible that we already have this issue viz in the store?
Perhaps worth asserting that we don't.

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -62,19 +127,24 @@ public void Set(IEnumerable<IAnalysisIssueVisualization> issueVisualizations, An
...
+            lock (Locker)
+            {
+                this.analysisInformation = analysisInformation;
+
+                var oldIssues = taintVulnerabilities;
+                taintVulnerabilities = issueVisualizations.ToList();
 
-            var oldIssues = taintVulnerabilities;
-            taintVulnerabilities = issueVisualizations.ToArray();
+                var removedIssues = oldIssues.Except(taintVulnerabilities).ToArray();
```
duncanp-sonar, line 137: This is comparing based on reference equality. Is this what we want?

src/IssueViz.Security/Taint/TaintStore.cs
rita-sonarsource: Unfortunately batch is only for issues. Taint is one by one. Consistency...

src/IssueViz.Security/Taint/TaintStore.cs
rita-sonarsource: Hmm. Good question. I suspect then that we always had a bug here. This code is only called by Set(), when the monitor fetches new issues from the server. I think in reality it was always putting all old as removed and all new as added.

src/IssueViz.Security/Taint/TaintStore.cs
rita-sonarsource: Hmm, B is indeed a problem. `analysisInformation` is used in the UI to display what branch we're on and when was the last analysis. Based on information we have in the event, we can know the branch name, but we don't know when the analysis happened. Perhaps we can use DateTime.Now? 

IMO the root cause here is that we want the list to be empty when there are no issues for the given file, so we want to distinguish between "there are no issues, at all" (= no analysis info), and "there are no issues in this file" (=has analysis info). I'm not sure how to tackle this.

src/IssueViz.Security/Taint/TaintStore.cs
duncanp-sonar: I think the old code just assumed we were doing a complete replacement of issues every time i.e. it was just a batch update.

src/IssueViz.Security/Taint/TaintStore.cs
duncanp-sonar: If we ignore the possible race condition in (A) (which I think is ok), we just need to change the `SynchronizeWithServer` method so that it always calls `GetAnalysisInformation(...)`, even if there are currently no issues.

There's a separate, unrelated issue: the taint synchronizer isn't branch-aware. We missed this case when doing the branch-awareness work :-(
I've created #3683.

rita-sonarsource: @duncanp-sonar it looks like taint itself _is_ branch aware, but the UI is displaying wrong information. https://github.com/SonarSource/sonarlint-visualstudio/blob/c27cd936cb2ff6ebf091d74aab884e1b68ca9bf3/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs#L101

src/IssueViz.Security/Taint/TaintStore.cs
georgii-borovinskikh-sonarsource: .ToList() also works :-)

src/IssueViz.Security/Taint/TaintStore.cs
georgii-borovinskikh-sonarsource: In both cases, I think, we can rely on the fact that we sync regularly and the missing issue is going to be picked up by the next sync

duncanp-sonar (commented):
Product code Looks good - comments inline.
I haven't looked at the tests.

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -83,13 +83,18 @@ public void Add(IAnalysisIssueVisualization issueVisualization)
...
             lock (Locker)
             {
+                if (analysisInformation == null)
+                {
+                    return;
+                }
+
+                if (taintVulnerabilities.Contains(issueVisualization, TaintAnalysisIssueVisualizationByIssueKeyEqualityComparer.Instance))
+                {
+                    return;
```
duncanp-sonar, line 95: Could the state of the issue have changed, and if so would we care?

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -62,23 +132,60 @@ public void Set(IEnumerable<IAnalysisIssueVisualization> issueVisualizations, An
...
-            var removedIssues = oldIssues.Except(taintVulnerabilities).ToArray();
-            var addedIssues = taintVulnerabilities.Except(oldIssues).ToArray();
-
             if (removedIssues.Any() || addedIssues.Any())
             {
                 IssuesChanged?.Invoke(this, new IssuesChangedEventArgs(removedIssues, addedIssues));
             }
         }
+
+        private class TaintAnalysisIssueVisualizationByIssueKeyEqualityComparer : IEqualityComparer<IAnalysisIssueVisualization>
```
rita-sonarsource, line 159: They're not going to be the same instance, so it's really only the `IssueKey` that matters. Why bother with a converter? We do the casting inside FindIndex for Remove, so why not do the same in Add?

src/IssueViz.Security/Taint/TaintStore.cs
georgii-borovinskikh-sonarsource: Clarified with Damien - the answer is no

Issue #3673, opened 2023-02-09
Update ProjectLanguageMapper to handle secrets


bigfluffycookie:
This requires changes in a few places:

- [x] Add Secrets to SonarArtefactKeys.cs -> SonarPluginKeys
- [x] Add Secrets in MinimumSupportedSonarQubePlugin
- [x] Add secrets configprovider to BindingController
- [x] Import IConnectedModeSecrets and check if they are available.

PR #3681, opened 2023-02-10
Add check for secrets in ProjectToLanguageMapper

bigfluffycookie:
fixes part 4 of https://github.com/SonarSource/sonarlint-visualstudio/issues/3673

When connecting to SonarCloud / SonarQube >= v9.9 secrets settings get downloaded.
When the slconfig secrets time stamp is out of date a gold bar asks to update it.
When updating it the slconfig gets correctly updated.

### DIFF
--- a/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
@@ -38,6 +38,7 @@ using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Analysis;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Core.Secrets;
 using SonarLint.VisualStudio.Integration.Binding;
 using SonarLint.VisualStudio.Integration.NewConnectedMode;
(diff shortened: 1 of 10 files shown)


src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -38,6 +38,7 @@
 using SonarLint.VisualStudio.Core.Analysis;
```
bigfluffycookie, line 38: As discussed offline, all files that were changed besides ProjectToLanguageMapper and ProjectToLanguageMapperTests contain the same changes which is adding a mock for IConnectedModeSecrets to the construction in testintialize.

src/Integration.UnitTests/ProjectToLanguageMapperTests.cs
```diff
@@ -144,6 +146,34 @@ public void GetAllBindingLanguagesForProject_NotCMakeProject_UnknownLanguage()
             actualLanguage.Should().BeEquivalentTo(Language.Unknown);
         }
 
+        [TestMethod]
```
bigfluffycookie, line 149: All new tests for copied from the CMake / JSTS tests and changed to check secrets.

duncanp-sonar (approved):
Nice!

PR #3684, opened 2023-02-10
Initialize SSESession if already in connected mode when loaded

### LINKED ISSUE #3668: Initialize SSESession if already in connected mode when package loads
### Description

Due to the async nature of the packages load order, it's possible that when this module starts, we're already in connected mode and missed the event. So we need to also explicitly check in the ctor if we're in connected mode, and if so, start a new session.

georgii-borovinskikh-sonarsource:
Closes #3668

### DIFF
--- a/src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
+++ b/src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
@@ -20,6 +20,7 @@
 
 using System;
 using System.Threading.Tasks;
+using Moq.Language.Flow;
 using SonarLint.VisualStudio.ConnectedMode.ServerSentEvents;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Integration.UnitTests;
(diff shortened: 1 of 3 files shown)


src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -33,9 +33,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
     public static class MefTestHelpers
     {
         [DebuggerStepThrough]
-        public static Export CreateExport<T>() where T: class
+        public static Export CreateExport<T>(T value = null) where T : class
```
rita-sonarsource, line 36: Why was this needed? there is a method below that accepts a value

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
```diff
@@ -36,10 +36,12 @@ internal interface ISSESessionManager : IDisposable
     [PartCreationPolicy(CreationPolicy.Shared)]
     internal sealed class SSESessionManager : ISSESessionManager
     {
+        private readonly object syncRoot = new object();
```
rita-sonarsource, line 39: It's a singleton but still better to be explicit and mark it static.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
georgii-borovinskikh-sonarsource: Hmm, true, I just got thrown off by the fact that that method received an object. Will revert this change

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
```diff
@@ -67,25 +71,49 @@ public void Dispose()
...
+                currentSession = sessionToLaunch = sseSessionFactory.Create(bindingConfiguration.Project.ProjectKey);
+            }
+
+            try
+            {
+                sessionToLaunch.PumpAllAsync().Forget();
+            }
+            catch (ObjectDisposedException)
+            {
+                // ignore
```
georgii-borovinskikh-sonarsource, line 106: @rita-sonarsource I've only just realized that because the manager doesn't await the Task, it doesn't need the catch

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
rita-sonarsource: The fields can't be static because they're MEF imported.
I think that actually avoiding deadlocks is a problem -- I'd rather get the deadlock and see it so I could fix it.

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
```diff
@@ -67,25 +71,42 @@ public void Dispose()
...
+                if (bindingConfiguration.Equals(currentBindingConfiguration))
+                {
+                    return;
+                }
 
-            currentSession = sseSessionFactory.Create(bindingConfiguration.Project.ProjectKey);
+                EndCurrentSession();
 
-            currentSession.PumpAllAsync().Forget();
+                currentBindingConfiguration = bindingConfiguration;
```
rita-sonarsource, line 89: I think we should not cache this.
This solves a scenario that can only happen during first time initialization, and it is unlikely to happen in practice. If we don't cache, then worst case we will throw away the session and make a new one which is identical.
I think this is preferred as it simplifies the code and the impact is not large.

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -29,48 +30,85 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.ServerSentEvents;
...
+
+    [TestMethod]
+    public void Ctor_WhenInConnectedModeOnCreation_CreatesSession()
+    {
+        var testScope = new TestScope(TestScope.CreateConnectedModeBindingConfiguration(DefaultProjectKey));
+        testScope.SetUpSSEFactoryToReturnNoOpSSESession(DefaultProjectKey);
+
+        var _ = testScope.CreateTestSubject();
+
+        testScope.SSESessionFactoryMock.Verify(factory => factory.Create(DefaultProjectKey), Times.Once);
```
rita-sonarsource, line 87: I think we should also check that the session was started

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -29,48 +30,85 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.ServerSentEvents;
...
+        testScope.SetUpSSEFactoryToReturnNoOpSSESession(DefaultProjectKey, () => isSubscribed.Should().BeTrue());
+
+        var _ = testScope.CreateTestSubject();
+
+        testScope.ActiveSolutionBoundTrackerMock.VerifyAdd(tracker => tracker.SolutionBindingChanged += It.IsAny<EventHandler<ActiveSolutionBindingEventArgs>>(), Times.Once);
+        testScope.SSESessionFactoryMock.Verify(factory => factory.Create(DefaultProjectKey), Times.Once);
+    }
+
+    [TestMethod]
+    public void Ctor_WhenInConnectedModeOnCreation_CreatesSession()
```
rita-sonarsource, line 80: What about the standalone test? with explicit verification that session factory was not called.

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
```diff
@@ -97,16 +136,31 @@ public void OnSolutionChanged_WhenChangesFromConnectedToConnected_CancelsSession
         sessionMock2.Verify(session => session.Dispose(), Times.Never);
     }
 
+    [TestMethod]
+    public void OnSolutionChanged_WhenAttemptsToConnectToAlreadyConnectedProject_KeepsPreviousSessionAlive()
```
rita-sonarsource, line 140: See my comment in the product code. I think this is dangerous.
The user can force a connection to restart by right-clicking in the team explorer and picking "Update". This will disable it.

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
rita-sonarsource: Edit: I think it's also dangerous. The user can force an update by clicking on "Update" in the team explorer. I _think_ that doing that fires this event too. This will kill the force-update functionality.

rita-sonarsource (approved):
LGTM but I'd remove the caching of the bindingConfiguration

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
georgii-borovinskikh-sonarsource: I hard disagree. What if the deadlock happens on a user's machine? How are you going to repro and fix it?

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
georgii-borovinskikh-sonarsource: Will the active solution tracker even raise an event in this case? Since it's the same binding configuration, I'm pretty sure it's not going to

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionManagerTests.cs
georgii-borovinskikh-sonarsource: see my reply in that thread

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
rita-sonarsource: You can check; put a breakpoint in the event and click on update.

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
rita-sonarsource: It's a minor issue so agree to disagree and keep it however.

PR #3685, opened 2023-02-10
[WIP] Process Taint server sent events

rita-sonarsource:
Fixes [3636](https://github.com/SonarSource/sonarlint-visualstudio/issues/3636)


src/IssueViz.Security/Taint/ServerSentEvents/ITaintServerEventSource.cs
```diff
@@ -18,9 +18,10 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using SonarLint.VisualStudio.Core.ServerSentEvents;
 using SonarQube.Client.Models.ServerSentEvents.ClientContract;
 
-namespace SonarLint.VisualStudio.Core.ServerSentEvents.TaintVulnerabilities
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint.ServerSentEvents
```
rita-sonarsource, line 24: I moved this to IssueViz.Security assembly and it caused some changes... I should've done it in a separate PR 🙈

src/IssueViz.Security/Taint/TaintStore.cs
```diff
@@ -52,6 +55,15 @@ internal sealed class TaintStore : ITaintStore
         public IReadOnlyCollection<IAnalysisIssueVisualization> GetAll() => taintVulnerabilities;
         
         public AnalysisInformation GetAnalysisInformation() => analysisInformation;
+        public void Add(IAnalysisIssueVisualization taintIssue)
```
rita-sonarsource, line 58: This will be replaced with real implementation in https://github.com/SonarSource/sonarlint-visualstudio/issues/3678

src/IssueViz.Security/Taint/ServerSentEvents/ITaintServerEventSource.cs
duncanp-sonar: It seems odd that `ConnectedMode` needs to know about `IssueViz.Security`.

Would it make sense to invert the dependency by moving the interface only to `ConnectedMode`?  The interface is specific to connected mode, so it seems more natural for it to be defined there.

The implementation is visualisation-specific, so that belongs in `IssueViz.Security`

src/IssueViz.Security/Taint/ServerSentEvents/TaintServerEventsListener.cs
```diff
@@ -0,0 +1,115 @@
...
+            ILogger logger)
+        {
+            this.serverBranchProvider = serverBranchProvider;
+            this.taintServerEventSource = taintServerEventSource;
+            this.taintStore = taintStore;
+            this.threadHandling = threadHandling;
+            this.taintToIssueVizConverter = taintToIssueVizConverter;
+            this.logger = logger;
+
+            cancellationTokenSource = new CancellationTokenSource();
```
duncanp-sonar, line 67: Is this only ever cancelled on dispose?

duncanp-sonar (commented):
Looks good - just the suggestion about changing the direction of the dependency.

src/IssueViz.Security/Taint/ServerSentEvents/TaintServerEventsListener.cs
rita-sonarsource: Yes, might not be needed, the class is a singleton so dispose = vs shutdown.

src/IssueViz.Security/Taint/ServerSentEvents/TaintServerEventsListener.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            while (!cancellationTokenSource.IsCancellationRequested)
+            {
+                var taintServerEvent = await taintServerEventSource.GetNextEventOrNullAsync();
+
+                switch (taintServerEvent)
+                {
+                    case null:
+                    {
+                        return;
```
georgii-borovinskikh-sonarsource, line 82: Can we rely on it being null when and only when we're done?

georgii-borovinskikh-sonarsource (changes requested):

src/IssueViz.Security/Taint/ServerSentEvents/TaintServerEventsListener.cs
rita-sonarsource: That should be the behavior of IServerSentEventSource - it's supposed to block indefinitely until Writer.Complete is called on the channel, which is done only on dispose.

src/IssueViz.Security.UnitTests/Taint/ServerSentEvents/TaintServerEventsListenerTests.cs
```diff
@@ -0,0 +1,240 @@
...
+        [TestMethod]
+        public async Task OnEvent_UnrecognizedServerEvent_EventIsIgnored()
+        {
+            var taintStore = new Mock<ITaintStore>();
+            var logger = new Mock<ILogger>();
+            var taintServerEventSource = SetupTaintServerEventSource(Mock.Of<ITestEvent>());
+
+            var testSubject = CreateTestSubject(
+                taintServerEventSource: taintServerEventSource.Object,
+                taintStore: taintStore.Object,
```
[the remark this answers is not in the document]
rita-sonarsource: Yes because I'm asserting it in every test

src/IssueViz.Security.UnitTests/Taint/ServerSentEvents/TaintServerEventsListenerTests.cs
```diff
@@ -0,0 +1,240 @@
...
+
+            await testSubject.ListenAsync();
+
+            taintStore.Verify(x => x.Add(convertedIssueViz), Times.Once);
+            taintStore.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public async Task Dispose_StopsListeningToServerEventsSource()
+        {
```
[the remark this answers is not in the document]
rita-sonarsource: Hmm, what do you propose? 🤔

src/IssueViz.Security.UnitTests/Taint/ServerSentEvents/TaintServerEventsListenerTests.cs
```diff
@@ -0,0 +1,240 @@
...
+            var taintServerEventSource = SetupTaintServerEventSource(serverEvent);
+            var taintStore = new Mock<ITaintStore>();
+
+            var testSubject = CreateTestSubject(
+                taintServerEventSource: taintServerEventSource.Object,
+                taintStore: taintStore.Object);
+
+            await testSubject.ListenAsync();
+
+            taintStore.Verify(x => x.Remove("some issue1"), Times.Once);
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Yes, but the person would be annoyed probably :-) I feel like this isn't hard to fix and will just look cleaner + no possibility of making someone mad in the future lol

src/IssueViz.Security.UnitTests/Taint/ServerSentEvents/TaintServerEventsListenerTests.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: TestScope 🤔

src/IssueViz.Security.UnitTests/Taint/ServerSentEvents/TaintServerEventsListenerTests.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Have no new thoughts on this except what we've already discussed some time ago. Was hoping maybe you have some fresh ideas

georgii-borovinskikh-sonarsource (approved):

src/IssueViz.Security.UnitTests/Taint/ServerSentEvents/TaintServerEventsListenerTests.cs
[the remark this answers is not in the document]
rita-sonarsource: Hmm. I'll need to look at it again with fresh eyes when we do the suppressions.

PR #3686, opened 2023-02-10
Add line hash calculation to SecretsToAnalysisIssueConverter.cs

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3682


src/CloudSecrets/SecretsToAnalysisIssueConverter.cs
```diff
@@ -55,7 +65,7 @@ public IAnalysisIssue Convert(ISecret secret, ISecretDetector secretDetector, st
                         endLine: endLine,
                         startLineOffset: startLineOffset,
                         endLineOffset: endLineOffset,
```
bigfluffycookie, line 67: Not sure about testing as its such a small change.

PR #3687, opened 2023-02-10
Add linehash calculation to SecretsToAnalysisIssueConverter

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3682


src/CFamily.UnitTests/packages.lock.json
```diff
@@ -1556,7 +1556,8 @@
         "dependencies": {
           "Microsoft.VisualStudio.Sdk": "[17.0.31902.203, )",
```
bigfluffycookie, line 1557: Most files are package updates so you can ignore those, needed to add a reference to IssueVisualization.

src/CloudSecrets.UnitTests/SecretsToAnalysisIssueConverterTests.cs
```diff
@@ -91,14 +92,19 @@ public void Convert_SingleLineSecret_PositionConverted()
```
bigfluffycookie, line 92: adding the new checks into the pre existing tests made most sense to me here, let me know if you think I should write a seperate test

src/CFamily.UnitTests/packages.lock.json
duncanp-sonar: I don't think `CloudSecrets` should depend on `IssueVisualization` - they are separate feature areas.
It would be better to move `ILineHashCalculator` to `Infrastructure.VS` since it is a fairly low-level helper class (it can't go in `Core` because is uses depends on VS types).

src/CloudSecrets.UnitTests/SecretsToAnalysisIssueConverterTests.cs
duncanp-sonar: I would have added to the existing tests too.

src/CFamily.UnitTests/packages.lock.json
duncanp-sonar: Could you move the interface and implementation to `Infrastructure.VS`, and put them in a new folder called `Editor` please? (there are a couple of editor-related helpers in `Infrastructure.VS` so it would be good to have a folder we can use to group them at some point).

PR #3689, opened 2023-02-10
Use quality profile

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3688

### DIFF
--- a/src/CloudSecrets.UnitTests/SecretsAnalyzerTests.cs
+++ b/src/CloudSecrets.UnitTests/SecretsAnalyzerTests.cs
@@ -48,7 +48,7 @@ namespace SonarLint.VisualStudio.CloudSecrets.UnitTests
                 MefTestHelpers.CreateExport<IContentTypeRegistryService>(),
                 MefTestHelpers.CreateExport<IAnalysisStatusNotifierFactory>(),
                 MefTestHelpers.CreateExport<ICloudSecretsTelemetryManager>(),
-                MefTestHelpers.CreateExport<IUserSettingsProvider>());
+                MefTestHelpers.CreateExport<IRuleSettingsProviderFactory>());
         }
 
         [TestMethod]
@@ -196,7 +196,7 @@ namespace SonarLint.VisualStudio.CloudSecrets.UnitTests
             statusNotifier.VerifyNoOtherCalls();
         }
 
-        [TestMethod]
+       [TestMethod]
         public void ExecuteAnalysis_DisabledDetectorsAreNotChecked()
         {
             var textDocumentFactoryService = SetupTextDocumentFactoryService(ValidFilePath, ValidFileContent);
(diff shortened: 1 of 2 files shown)


src/CloudSecrets/SecretsAnalyzer.cs
```diff
@@ -137,8 +137,9 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
 
         private bool IsRuleActive(string ruleKey)
         {
+            var rulesSettings = ruleSettingsProvider.Get();
```
bigfluffycookie, line 140: the ruleSettingsProvider already deals with checking if we are in connected mode and if the settings file exists so we can just replace the usersettingsProvider here.

The only issue is that the RuleSettingProvider will just say "unable to load connected mode settings" if we are in SonarQube < v9.9 instead of giving a more precise log message.

bigfluffycookie, line 140: The reason I Get() the rule settings everytime we check if a rule is active is so that if we update the Quality profile at any point during runtime it gets refreshed here aswell, else we'd be stuck with the same one.

src/CloudSecrets.UnitTests/SecretsAnalyzerTests.cs
```diff
@@ -48,7 +48,7 @@ public void MefCtor_CheckIsExported()
                 MefTestHelpers.CreateExport<IContentTypeRegistryService>(),
```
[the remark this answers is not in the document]
duncanp-sonar: You're right, it's the responsibility of the `RuleSettingsProviderTests`. Here, we just need to mock `IRuleSettingsProvider` to return whatever makes sense for these tests.

src/CloudSecrets/SecretsAnalyzer.cs
```diff
@@ -137,8 +137,9 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
             // User settings override the default, if present
-            if (userSettingsProvider.UserSettings.RulesSettings.Rules.TryGetValue(ruleKey, out var ruleConfig))
+            if (rulesSettings.Rules.TryGetValue(ruleKey, out var ruleConfig))
```
duncanp-sonar, line 142: Is the comment in line 141 still required? It's a bit misleading since we're not just looking at the user settings any more.

duncanp-sonar (commented):
Looks good - just one question about what which settings re actually used in SQ <= v9.9 in Connected Mode.

src/CloudSecrets/SecretsAnalyzer.cs
bigfluffycookie: Yes, in the ruleSettingsProvider it returns "userSettingsProvider.UserSettings.RulesSettings" if it cannot find the connected mode

src/CloudSecrets/SecretsAnalyzer.cs
bigfluffycookie: Ah true, missed that!

duncanp-sonar (approved):

Issue #2798, opened 2021-12-09
[Perf] VS becomes sluggish/locks up after opening very large C++ files


duncanp-sonar:
### Description

Users have reported that VS slows down after opening very large C++ files (10k+ lines).
See the following community thread for more information : https://community.sonarsource.com/t/sonarlint-vs-extension-performance/54423

### Repro

TODO - find a project with C++ with files of 10K+ lines
Candidate: https://github.com/llvm/llvm-project

### Notes:

* the issue does not appear if SonarLint is not installed
* _"I start scrolling then the UI thread is completely blocked"_
* _"It seems to happen when i switch tabs"_
* log extract:
> Finished analyzing C:\WIP\App\Loadboard2\LB\BusLogic\QGWrQMgt\QGMgrApplyInstrConcrete.cpp, analysis time: 47.475s
> Found 672 issue(s) for C:\WIP\App\Loadboard2\LB\BusLogic\QGWrQMgt\QGMgrApplyInstrConcrete.cpp

### Areas to investigate - internal
* tagging - what is VS asking us to recalculate when scrolling? Are we recalculating more tags than we need to?
* streaming issues back from the subprocess to the Error List. Is VS trying to update the UI too often?
* PCH - how often is PCH analysis being triggered? How often cancelled? How much does it cost?
* calculating rule configuration: how long does it take? How much disc access (note: does not appear to be an issue in the community thread cited above - the log shows it using a cached configuration)

### Areas to investigate - external
* effect of anti-virus software: how many temporary files are we writing? How large are they? (e.g. PCH output)

#### Sample open source project

https://github.com/llvm/llvm-project

Instructions for generating a solution file:

```
Notes:
- Dependencies: cmake, python3, VS2022
- Run from VS 2022 x64 naitivetools command prompt

Instructions:
git clone https://github.com/llvm/llvm-project.git
cd llvm-project
mkdir build
cd build
cmake -G "Visual Studio 17 2022" -Thost=x64 -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD=X86 ..\llvm

// the last command should generate llvm-project\build\llvm.sln
// open the solution and build the project once. The build generate some file that are used in the project. It takes a lot of time to build.
// the freeze for me happen on switching between large files: type.h decl.h type.cpp.
```

```
 Directory of c:\llvm\llvm-project\clang\lib\Sema
[...]

PR #3690, opened 2023-02-11
Add IErrorTag implementation that lazily-creates tooltips

duncanp-sonar:
Relates to #2798

### DIFF
--- a/src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs
+++ b/src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs
@@ -26,6 +26,7 @@ using Microsoft.VisualStudio.PlatformUI;
 using Microsoft.VisualStudio.Shell;
 using SonarLint.VisualStudio.Core.Analysis;
 using SonarLint.VisualStudio.Infrastructure.VS;
+using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.IssueVisualization.IssueVisualizationControl.ViewModels.Commands;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Editor.ErrorTagging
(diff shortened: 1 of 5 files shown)


src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs
```diff
@@ -74,6 +80,8 @@ public object Create(IAnalysisIssueBase analysisIssueBase)
                 Foreground = GetVsThemedColor(EnvironmentColors.SystemCaptionTextBrushKey)
             };
 
+            instanceCount++;
+            logger.LogVerbose($"[ErrorTagTooltipProvider] tooltip instance count: {instanceCount}");
```
duncanp-sonar, line 84: I left the logging in for diagnostic purposes. It's low-cost, since it will only occur when tooltips are actually created.

duncanp-sonar:
The underlying issue that we are creating a lot of tags still exists. I've created #3692 to track that.
This fix at least makes the IDE responsive on very large files.

src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs
```diff
@@ -44,14 +45,19 @@ internal class ErrorTagTooltipProvider : IErrorTagTooltipProvider
...
-        public ErrorTagTooltipProvider(IVsThemeColorProvider vsThemeColorProvider, INavigateToRuleDescriptionCommand navigateToRuleDescriptionCommand)
+
+        public ErrorTagTooltipProvider(IVsThemeColorProvider vsThemeColorProvider, INavigateToRuleDescriptionCommand navigateToRuleDescriptionCommand, ILogger logger)
         {
             this.vsThemeColorProvider = vsThemeColorProvider;
             this.navigateToRuleDescriptionCommand = navigateToRuleDescriptionCommand;
+            this.logger = logger;
         }
 
+        private static long instanceCount = 0;
```
ugras-ergun-sonarsource, line 59: Wouldn't this number will always grow and could be misleading especially if a project was open for a long time?

duncanp-sonar: It will always grow, yes. The difference between "normal" behaviour and "problem" is orders of magnitude, so it will allow us to tell the difference between them. If we are creating thousands of tooltips, we have a problem.

ugras-ergun-sonarsource (approved):

src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs
duncanp-sonar: As a bonus, the recent tweaks to the verbose logging mean this also gives an easy way to see which thread the code is running on.

Issue #3683, opened 2023-02-10
Taint issue window is showing the wrong branch


duncanp-sonar:
### Description

The `TaintSynchronizer` is fetching the taint issues for the correct branch, but it is always displaying the branch name and last analysis time for the `main` branch.

See [here](https://github.com/SonarSource/sonarlint-visualstudio/blob/c27cd936cb2ff6ebf091d74aab884e1b68ca9bf3/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs#L172). The code is always using the branch marked as `main`.

PR #3693, opened 2023-02-13
Fix the displayed taint branch name

rita-sonarsource:
Fixes [#3683](https://github.com/SonarSource/sonarlint-visualstudio/issues/3683)

### DIFF
--- a/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
+++ b/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
@@ -20,6 +20,7 @@
 
 using System;
 using System.ComponentModel.Composition;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -168,12 +169,13 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint
 
         private async Task<AnalysisInformation> GetAnalysisInformation(string projectKey, string branchName)
         {
+            Debug.Assert(branchName != null, "BranchName should not be null when in Connected Mode");
+            
             var branches = await sonarQubeService.GetProjectBranchesAsync(projectKey, CancellationToken.None);
 
-            var issuesBranch = string.IsNullOrEmpty(branchName)
-                ? branches.First(x => x.IsMain)
-                : branches.FirstOrDefault(x => x.Name.Equals(branchName))
-                  ?? branches.First(x => x.IsMain);
+            var issuesBranch = branches.FirstOrDefault(x => x.Name.Equals(branchName));
+
+            Debug.Assert(issuesBranch != null, "Should always find a matching branch");
 
             return new AnalysisInformation(issuesBranch.Name, issuesBranch.LastAnalysisTimestamp);
         }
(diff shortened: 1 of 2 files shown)


src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -111,7 +111,7 @@ public async Task SynchronizeWithServer()
                 }
                 else
                 {
-                    var analysisInformation = await GetAnalysisInformation(projectKey);
+                    var analysisInformation = await GetAnalysisInformation(projectKey, serverBranch);
```
rita-sonarsource, line 114: server branch is what we get from IStatefulServerBranchProvider.GetServerBranchNameAsync -- this will return Main if it can't find branch information, right?

duncanp-sonar: Unfortunately, no. It will return `null` in that case, which the SonarQube APIs will then interpret as meaning "fetch whatever branch is marked as the main branch".

I think we should update `IServerBranchProvider`: move the logic from the taint synchronizer that locates the main branch to `IServerBranchProvider` so it can explicitly return the name of the branch to fetch in all cases.

rita-sonarsource: Ok then I will add a fallback to main branch 👍

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -166,10 +166,14 @@ private bool IsFeatureSupported(ServerInfo serverInfo)
             return false;
         }
 
-        private async Task<AnalysisInformation> GetAnalysisInformation(string projectKey)
+        private async Task<AnalysisInformation> GetAnalysisInformation(string projectKey, string branchName)
         {
             var branches = await sonarQubeService.GetProjectBranchesAsync(projectKey, CancellationToken.None);
-            var issuesBranch = branches.First(x => x.IsMain);
+
+            var issuesBranch = string.IsNullOrEmpty(branchName)
```
rita-sonarsource, line 173: With the recent change I did, branchName should not be null anymore, but still, I figured I might as well leave this. WDYT?

duncanp-sonar: If we'd documented that our API should never return `null`, then we could `Debug.Assert` instead - a `null` would mean our code isn't working as expected.

duncanp-sonar (approved):

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
rita-sonarsource: So what is the behavior we expect if we can't find a matching branch? I guess we could throw and clear the store 🤔

duncanp-sonar: It should never happen; if it does, it's because there's a bug in another part of our code and we're returning `null` when we said we wouldn't.

If it were third-party code then I'd be defensive and handle it being `null`, since we're not in control of the behaviour.

Here, we are in control, so I wouldn't be defensive. The assertion documents what we expect from our code. If we've messed it up the user will get a runtime failure.

Issue #3691, opened 2023-02-13
Update disable rule command logic for secrets


duncanp-sonar:
### Description

Users should only be able to disable Secrets rules if
a) they are in standalone mode, or
b) they are in Connected Mode but connected to SQ <= v9.9

PR #3695, opened 2023-02-13
Update disable rule command for secrets

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3691

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
@@ -174,7 +174,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
         public void CheckStatus_SingleIssue_Secrets_NotConnected_IsEnabled()
         {
             var mockUserSettingsProvider = new Mock<IUserSettingsProvider>();
-            var solutionTracker = CreateSolutionTracker(SonarLintMode.Connected);
+            var solutionTracker = CreateSolutionTracker(SonarLintMode.Standalone);
             var errorListHelper = CreateErrorListHelper("secrets:S111", ruleExists: true);
             var connectedModeSecrets = CreateConnectedModeSecrets(areSecretsAvailableInConnectedMode:false);
 
--- a/src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
@@ -26,9 +26,10 @@ using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Core;
-using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Infrastructure.VS;
 using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarQube.Client;
+
 using ThreadHelper = SonarLint.VisualStudio.Integration.UnitTests.ThreadHelper;
 
 namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
@@ -46,23 +47,23 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
             var menuCommandService = new DummyMenuCommandService();
             var errorList = Mock.Of<IErrorList>();
             var userSettingsProvider = new Mock<IUserSettingsProvider>().Object;
-            var solutionTracker = new Mock<IActiveSolutionBoundTracker>().Object;
+            var sonarQubeService = new Mock<ISonarQubeService>().Object;
             var logger = new TestLogger();
             var errorListHelper = Mock.Of<IErrorListHelper>();
 
-            Action act = () => new DisableRuleCommand(null, userSettingsProvider, solutionTracker, logger, errorListHelper);
+            Action act = () => new DisableRuleCommand(null, userSettingsProvider, sonarQubeService, logger, errorListHelper);
             act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("menuCommandService");
 
-            act = () => new DisableRuleCommand(menuCommandService, null, solutionTracker, logger, errorListHelper);
+            act = () => new DisableRuleCommand(menuCommandService, null, sonarQubeService, logger, errorListHelper);
             act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("userSettingsProvider");
 
             act = () => new DisableRuleCommand(menuCommandService, userSettingsProvider, null, logger, errorListHelper);
-            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("activeSolutionBoundTracker");
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("sonarQubeService");
 
-            act = () => new DisableRuleCommand(menuCommandService, userSettingsProvider, solutionTracker, null, errorListHelper);
+            act = () => new DisableRuleCommand(menuCommandService, userSettingsProvider, sonarQubeService, null, errorListHelper);
             act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("logger");
 
-            act = () => new DisableRuleCommand(menuCommandService, userSettingsProvider, solutionTracker, logger, null);
+            act = () => new DisableRuleCommand(menuCommandService, userSettingsProvider, sonarQubeService, logger, null);
             act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("errorListHelper");
         }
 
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/Analysis/DisableRuleCommand.cs
```diff
@@ -156,23 +156,35 @@ private void Execute(object sender, EventArgs e)
...
             // Special case: Secrets
-            // We don't support connected mode for Secrets at the moment so we allow disabling
-            // secrets rules in connected mode.
+            // We don't support connected mode for Secrets in SonarQubeVersions < 9.9 so we allow disabling
+            // secrets rules in those cases.
             if (SonarRuleRepoKeys.AreEqual(SonarRuleRepoKeys.Secrets, rule.RepoKey))
             {
                 isVisible = true;
-                isEnabled = true;
+
```
bigfluffycookie, line 169: I don't like this very much but I tried a few different options and all of them were equally ugly.

duncanp-sonar: Can't we use the new `IConnectedModeSecrets` service you created? It should have the correct logic.

src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
```diff
@@ -44,26 +45,29 @@ public class DisableRuleCommandTests
         public void Ctor_NullArguments()
         {
             var menuCommandService = new DummyMenuCommandService();
-            var errorList = Mock.Of<IErrorList>();
```
bigfluffycookie, line 47: I should have removed this awhile ago as its unused

src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
```diff
@@ -121,8 +125,6 @@ public void CheckStatusAndExecute_SingleIssue_SupportedRepo_StandaloneMode_Visib
         [DataRow("javascript:S111", SonarLintMode.LegacyConnected, false)]
         [DataRow("typescript:S111", SonarLintMode.Connected, false)]
         [DataRow("typescript:S111", SonarLintMode.LegacyConnected, false)]
-        [DataRow("secrets:S111", SonarLintMode.Connected, true)]
```
bigfluffycookie, line 124: Made a seperate test just for secrets

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
```diff
@@ -156,15 +155,15 @@ private void Execute(object sender, EventArgs e)
...
             // Special case: Secrets
-            // We don't support connected mode for Secrets at the moment so we allow disabling
-            // secrets rules in connected mode.
+            // We don't support connected mode for Secrets in SonarQubeVersions < 9.9 so we allow disabling
+            // secrets rules in those cases.
             if (SonarRuleRepoKeys.AreEqual(SonarRuleRepoKeys.Secrets, rule.RepoKey))
             {
                 isVisible = true;
-                isEnabled = true;
+                isEnabled = !isConnectedModeSupportedForSecrets;
```
bigfluffycookie, line 166: I realize it is a bit confusing to use the negative here, hence why i made the boolean name so long . However I dont think it makes it any less confusing

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
```diff
@@ -17,15 +17,16 @@
...
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-
 using System;
 using System.ComponentModel.Design;
 using System.Diagnostics;
 using System.Linq;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
+using Org.BouncyCastle.Crypto.Parameters;
```
duncanp-sonar, line 26: It seems odd that the `BouncyCastle` using is required because of this change. Is it definitely required?

bigfluffycookie: It is not, I have absolutely no idea why that was put in

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
duncanp-sonar: Sometimes it's difficult to find a good name or avoid a negation. You've added a comment and the name is descriptive, so that's probably as much as we can do.

An alternative approach would be to make the method non-`static`, and perhaps create a separate function for the secrets check and explanatory comment. That avoids having to think of a variable name for the `isConnectedModeSupportedForSecrets` boolean e.g.

```
        private void CalculateStatuses(SonarCompositeRuleId rule, out bool isVisible, out bool isEnabled)
        {
            // Special case: Secrets
            if (SonarRuleRepoKeys.AreEqual(SonarRuleRepoKeys.Secrets, rule.RepoKey))
            {
                CalculateStatusesForSecrets(out isVisible, out isEnabled);
                return;
            }

            // We don't currently support disabling rules for all repos
            if (supportedRepos.Contains(rule.RepoKey, SonarRuleRepoKeys.RepoKeyComparer))
            {
                isVisible = true;
                isEnabled = activeSolutionBoundTracker.CurrentConfiguration.Mode == SonarLintMode.Standalone;
                return;
            }

            isVisible = false;
            isEnabled = false;
        }

        private void CalculateStatusesForSecrets(out bool isVisible, out bool isEnabled)
        {
            isVisible = true;

            // We don't support connected mode for Secrets in SonarQubeVersions < 9.9 so we
            // still allow users to disable secrets rules in those cases.
            isEnabled = !connectedModeSecrets.AreSecretsAvailable();
        }
```

duncanp-sonar: Another approach is to refactor the code to avoid out parameters:

```
        private void QueryStatus(object sender, EventArgs args)
        {
            try
            {
                var isVisible = false;
                var isEnabled = false;
                if (errorListHelper.TryGetRuleIdFromSelectedRow(out var ruleId))
                {
                    isVisible = IsSonarRule(ruleId);
                    isEnabled = isVisible && IsDisablingRuleAllowed(ruleId);
                }

                menuItem.Visible = isVisible;
                menuItem.Enabled = isEnabled;
            }
            {
                logger.WriteLine(AnalysisStrings.DisableRule_ErrorCheckingCommandStatus, ex.Message);
            }
        }

        private bool IsSonarRule(SonarCompositeRuleId rule)
            => supportedRepos.Contains(rule.RepoKey, SonarRuleRepoKeys.RepoKeyComparer);

        private bool IsDisablingRuleAllowed(SonarCompositeRuleId rule)
        {
            // Special case: secrets
            // We don't support connected mode for Secrets in SonarQubeVersions < 9.9 so we
            // still allow users to disable secrets rules in those cases.
            if (SonarRuleRepoKeys.AreEqual(SonarRuleRepoKeys.Secrets, rule.RepoKey))
            {
                return !connectedModeSecrets.AreSecretsAvailable();
            }

            // Otherwise, can only disables rules in standalone mode
            return activeSolutionBoundTracker.CurrentConfiguration.Mode == SonarLintMode.Standalone;
        }
```

That's probably simplest. We have enough tests coverage, so changing the methods completely like this is fine if we want to.

bigfluffycookie: Really like the second option, much more elegant!

src/Integration.Vsix.UnitTests/Analysis/DisableRuleCommandTests.cs
```diff
@@ -144,6 +146,52 @@ public void CheckStatusAndExecute_SingleIssue_SupportedRepo_StandaloneMode_Visib
...
+            var command = CreateDisableRuleMenuCommand(mockUserSettingsProvider.Object, solutionTracker, new TestLogger(), errorListHelper, connectedModeSecrets);
+
+            // Act. Trigger the query status check
+            ThreadHelper.SetCurrentThreadAsUIThread();
+            var result = command.OleStatus;
+
+            var expectedOleStatus =  VisibleAndEnabled;
+            result.Should().Be(expectedOleStatus);
+
+            // Should always be visible, but not necessarily enabled
```
duncanp-sonar, line 190: This copy-and-pasted comment is a bit misleading in this test.

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
```diff
@@ -153,31 +151,25 @@ private void Execute(object sender, EventArgs e)
...
-            {
-                isVisible = true;
-                isEnabled = mode == SonarLintMode.Standalone;
-                return;
+                return !connectedModeSecrets.AreSecretsAvailable();
             }
 
-            isVisible = false;
-            isEnabled = false;
+            // Otherwise, can only disables rules in standalone mode
```
duncanp-sonar, line 171: Typo - `disables`. (I realise it was my typo ;-) )

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
duncanp-sonar: Probably VS trying to be helpful when you were typing, but incorrectly guessing which types you actually needed.

duncanp-sonar (approved):

Issue #3697, opened 2023-02-13
Fix ServerBranchProvider to fallback to Main server branch


rita-sonarsource:
### Description

Currently `ServerBranchProvider` returns null if we were not able to match the local branch to the server branch. This works when we then use the result to pull from other APIs, as "null" in SQ/SC api means "main branch". However, this doesn't work for places where we actually need to know the main branch's name, i.e. in Taint Vulnerabilities:

1. When we display in Taint screen which branch we are on
2. When we receive Taint events and filter to keep only the events for the current branch

TODO:
- [x] ServerBranchProvider should always return the name of the main branch, if no other matching branch was found.
- [ ] Perhaps it should also return the branch analysis information, so that we could use it in TaintIssuesSynchronizer

PR #3702, opened 2023-02-14
Return default Main branch when there is no git repo or no git head

rita-sonarsource:
Fixes [#3697](https://github.com/SonarSource/sonarlint-visualstudio/issues/3697)

I've reverted my previous commit that handled headless branches. This new commit should handle both cases (no git repo + headless).

### DIFF
--- a/src/ConnectedMode.UnitTests/BranchMatcherTests.cs
+++ b/src/ConnectedMode.UnitTests/BranchMatcherTests.cs
@@ -236,13 +236,13 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests
         }
 
         [TestMethod]
-        public async Task GetMatchedBranch_RepoHasNoHead_ReturnsMainBranch()
+        public async Task GetMatchedBranch_RepoHasNoHead_ReturnsNull()
         {
             var testSubject = CreateTestSubject(CreateSonarQubeService("any").Object);
 
             var result = await testSubject.GetMatchingBranch("projectKey", Mock.Of<IRepository>(), CancellationToken.None);
 
-            result.Should().Be("any");
+            result.Should().BeNull();
         }
 
         [TestMethod]
--- a/src/ConnectedMode/BranchMatcher.cs
+++ b/src/ConnectedMode/BranchMatcher.cs
@@ -35,10 +35,11 @@ namespace SonarLint.VisualStudio.ConnectedMode
     {
         /// <summary>
         /// Calculates the Sonar server branch that is the closest match to the head branch.
-        /// Defaults to branch marked as "main" when no match is found or when there are no head branches.
+        /// Returns null if there are no head branches
         /// </summary>
         /// <remarks>
         /// Compares branches by name and calculates distances by the number of different commits.
+        /// Defaults to branch marked as "main" when no match found.
         /// This is the same algorithm as the other SonarLint flavours (?except in how we treated branch histories.
         /// We treat branch histories as series of linear commits ordered by time, even if there are merges).
         /// </remarks>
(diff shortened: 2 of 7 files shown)


src/ConnectedMode/ServerBranchProvider.cs
```diff
@@ -39,51 +41,73 @@ internal class ServerBranchProvider : IServerBranchProvider
...
+
+            if (matchingBranchName == null)
+            {
+                logger.LogVerbose(Resources.BranchProvider_FailedToCalculateMatchingBranch);
+
+                var remoteBranches = await sonarQubeService.GetProjectBranchesAsync(config.Project.ProjectKey, token);
+                matchingBranchName = remoteBranches.First(rb => rb.IsMain).Name;
+            }
+
+            logger.WriteLine(Resources.BranchProvider_MatchingServerBranchName, matchingBranchName ?? Resources.NullBranchName);
```
ugras-ergun-sonarsource, line 94: I think we should remove the null check here since it can not be null.

ugras-ergun-sonarsource (approved):
It looks good, just a redundant null check.

src/ConnectedMode.UnitTests/BranchMatcherTests.cs
```diff
@@ -236,13 +236,13 @@ public async Task GetMatchedBranch_NoMatchingBranch_ChooseMain()
...
         [TestMethod]
-        public async Task GetMatchedBranch_RepoHasNoHead_ReturnsMainBranch()
+        public async Task GetMatchedBranch_RepoHasNoHead_ReturnsNull()
         {
             var testSubject = CreateTestSubject(CreateSonarQubeService("any").Object);
 
             var result = await testSubject.GetMatchingBranch("projectKey", Mock.Of<IRepository>(), CancellationToken.None);
 
-            result.Should().Be("any");
+            result.Should().BeNull();
```
duncanp-sonar, line 245: The name and behaviour of this test doesn't match the description or the description in the bug, both of which seem to say we should return [main] for headless repos. Which is correct?

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -35,10 +35,11 @@ public interface IBranchMatcher
     {
         /// <summary>
         /// Calculates the Sonar server branch that is the closest match to the head branch.
-        /// Defaults to branch marked as "main" when no match is found or when there are no head branches.
+        /// Returns null if there are no head branches
```
duncanp-sonar, line 38: Does this mean that there are no branches at all, or just that the repo is currently on a headless commit?

src/ConnectedMode/Resources.resx
```diff
@@ -141,6 +141,9 @@
   <data name="BranchProvider_CouldNotDetectGitRepo" xml:space="preserve">
     <value>[ConnectedMode/BranchMapping] Could not detect a git repo for the current solution</value>
   </data>
+  <data name="BranchProvider_FailedToCalculateMatchingBranch" xml:space="preserve">
+    <value>[ConnectedMode/BranchMapping] Failed to calculate closest local git branch, defaulting to server main branch.</value>
```
duncanp-sonar, line 145: It might be worth saying `... defaulting to Sonar main branch.` (or `Sonar server main branch`), just to make it absolutely clear.

Since the message is talking about local git branches then server branches, my first thought is that "server" means git server. I have to remind myself it doesn't.

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -35,10 +35,11 @@ public interface IBranchMatcher
...
         /// </summary>
         /// <remarks>
         /// Compares branches by name and calculates distances by the number of different commits.
+        /// Defaults to branch marked as "main" when no match found.
```
duncanp-sonar, line 42: Looks like the behaviour is to return `null` for detached heads
i.e. it returns `null` for detached heads, the closest match if it could find one, or `main` if it couldn't.

If the caller now has to handle `null` (and has the logic to find `main` if the branch matcher returns `null`), this class could be simpler: return closest match, or null in all other cases.

duncanp-sonar (approved):
Just some comments about comments, and a possible simplification of the branch matcher.

src/ConnectedMode/BranchMatcher.cs
rita-sonarsource: I'll open a separate issue for it 👍

Issue #3701, opened 2023-02-14
Reanalysis is not triggered automatically for secrets when a rule is disabled


duncanp-sonar:
### Description

Note: this only applies when all of the following conditions are met:
* secrets rules
* in Connected Mode
* when connected to SQ <= v9.9  

### Repro steps
* bind a solution to SQ v9.8
* open a file with secrets issues
* disable a secrets rule (from the Error List context menu)

### Expected behaviour
Re-analysis of open files is triggered automatically when a secrets rule is disabled -> issues disappear from the Error List immediately

### Actual behaviour
Re-analysis is not triggered automatically.
The

### Workaround
Trigger re-analysis manually i.e. save or re-open the file

PR #3703, opened 2023-02-14
Bug Fix

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3701

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
@@ -19,13 +19,11 @@
  */
 
 using System;
-using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Analysis;
-using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Core.Suppression;
 using SonarLint.VisualStudio.Integration.UnitTests;
 using static SonarLint.VisualStudio.Integration.UnitTests.NoOpThreadHandler;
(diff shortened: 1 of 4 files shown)


src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -75,15 +71,8 @@ internal sealed class AnalysisConfigMonitor : IAnalysisConfigMonitor, IDisposabl
...
-            if (activeSolutionBoundTracker.CurrentConfiguration.Mode == SonarLintMode.Standalone)
-            {
-                logger.WriteLine(AnalysisStrings.ConfigMonitor_UserSettingsChanged);
-                OnSettingsChangedAsync().Forget();
-            }
-            else
-            {
-                logger.WriteLine(AnalysisStrings.ConfigMonitor_UserSettingsIgnoredForConnectedModeLanguages);
-            }
+            logger.WriteLine(AnalysisStrings.ConfigMonitor_UserSettingsChanged);
```
bigfluffycookie, line 74: I think its fine to leave this message as if user specified settings are not supported it'll print this later "[RuleSettings] Using connected mode settings. User-specified settings in settings.json will be ignored."

src/Integration.Vsix.UnitTests/Analysis/AnalysisConfigMonitorTests.cs
```diff
@@ -19,13 +19,11 @@
  */
```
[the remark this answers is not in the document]
duncanp-sonar: The best kind of change...

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -75,15 +71,8 @@ internal sealed class AnalysisConfigMonitor : IAnalysisConfigMonitor, IDisposabl
...
         private void OnUserSettingsChanged(object sender, EventArgs e)
         {
-            if (activeSolutionBoundTracker.CurrentConfiguration.Mode == SonarLintMode.Standalone)
-            {
-                logger.WriteLine(AnalysisStrings.ConfigMonitor_UserSettingsChanged);
-                OnSettingsChangedAsync().Forget();
-            }
-            else
-            {
-                logger.WriteLine(AnalysisStrings.ConfigMonitor_UserSettingsIgnoredForConnectedModeLanguages);
```
duncanp-sonar, line 85: Is this resource string still used?

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
duncanp-sonar, line 74: I think it's worth adding a comment explaining why we might trigger re-analysis in Connected Mode e.g. 

> There is a corner-case where we want to raise the the event even in Connected Mode - see #3701.

duncanp-sonar (approved):
Nice. Just comments about comments.

duncanp-sonar:
@bigfluffycookie I tested the fix and it works as expected. @

Issue #3708, opened 2023-02-15
New taint issue is displayed as non-navigable


rita-sonarsource:
### Description

The taint issue arriving from the server event is displayed as non navigable, even though the code locally is identical to the code on the server. This is due to the server sending us a different hash for the same text range: we calculate the line hash for the whole line, whereas the server is sending us the hash of the exact text range.

Correct line hash of the whole line (calculated when we pull taint and calculate it ourselves):

![image](https://user-images.githubusercontent.com/60586879/218991792-a71b9ff2-2faf-483c-a2c1-b2de307db2e6.png)

Incorrect line hash arriving from the taint event, which is calculated for the exact text range:

![image](https://user-images.githubusercontent.com/60586879/218991883-f82f0de6-7bcc-4170-9a7c-edae5e6fc9ec.png)

PR #3712, opened 2023-02-15
Calculate span hash for the given range

rita-sonarsource:
Fixes [#3708](https://github.com/SonarSource/sonarlint-visualstudio/issues/3708)


src/IssueViz/Editor/IIssueSpanCalculator.cs
```diff
@@ -97,21 +89,28 @@ internal IssueSpanCalculator(IChecksumCalculator checksumCalculator)
 
             var start = new SnapshotPoint(currentSnapshot, startPos);
             var end = new SnapshotPoint(currentSnapshot, endPos);
+            var snapshotSpan = new SnapshotSpan(start, end);
+
+            if (IsHashDifferent(range.LineHash, snapshotSpan.GetText()) &&
```
rita-sonarsource, line 94: I'm really not sure how to test it. We're mocking the entire VS line :/

src/IssueViz/Editor/IIssueSpanCalculator.cs
```diff
@@ -97,21 +89,28 @@ internal IssueSpanCalculator(IChecksumCalculator checksumCalculator)
...
 
-            return new SnapshotSpan(start, end);
+            return snapshotSpan;
         }
 
-        private bool IsLineHashDifferent(ITextRange range, ITextSnapshotLine snapshotLine)
+        private bool IsHashDifferent(string issueHash, string documentText)
         {
-            if (string.IsNullOrEmpty(range.LineHash))
+            if (string.IsNullOrEmpty(issueHash))
```
georgii-borovinskikh-sonarsource, line 106: this check gets called 2 times + looks kind of out of place => remove it out of this method and place it before the main condition `if (IsHashDifferent(....`

src/IssueViz/Editor/IIssueSpanCalculator.cs
```diff
@@ -97,21 +89,28 @@ internal IssueSpanCalculator(IChecksumCalculator checksumCalculator)
...
             {
                 return false;
             }
 
-            var textInSnapshot = snapshotLine.GetText();
-            var snapshotLineHash = checksumCalculator.Calculate(textInSnapshot);
+            var documentTextHash = checksumCalculator.Calculate(documentText);
 
-            return snapshotLineHash != range.LineHash;
+            return documentTextHash != issueHash;
```
georgii-borovinskikh-sonarsource, line 113: Just a question: why is line hash a string and not a number? Looks so weird

src/IssueViz/Editor/IIssueSpanCalculator.cs
georgii-borovinskikh-sonarsource: WDYM by mocking a vs line? What does "line" mean?

duncanp-sonar: "line" = `Microsoft.VisualStudio.Text.ITextSnapshotLine`

src/IssueViz/Editor/IIssueSpanCalculator.cs
duncanp-sonar: No particular reason. The SQ APIs return the hashes as strings and we didn't change the format when we deserialized them.

duncanp-sonar (approved):

duncanp-sonar:
Replaced by #3732

Issue #3705, opened 2023-02-15
Taint event's absolute file path is not being calculated


rita-sonarsource:
### Description

The server includes in the taint event the relative file path, which we then use to calculate the absolute file path in the solution.
However, the server sends forward-slashes, and AbsoluteFilePathLocator needs back-slashes. 

TODO:
- [ ] Add an assertion to AbsoluteFIlePathLocator that the string is not back-slashed, otherwise it fails silently.
- [ ] Fix the sub module to return correct file paths (call FilePathNormalizer)

PR #3713, opened 2023-02-15
Fix AbsoluteFilePathLocator to handle forward slashes

rita-sonarsource:
Part of [#3705](https://github.com/SonarSource/sonarlint-visualstudio/issues/3705)

### DIFF
--- a/src/Integration/Helpers/AbsoluteFilePathLocator.cs
+++ b/src/Integration/Helpers/AbsoluteFilePathLocator.cs
@@ -45,6 +45,7 @@ namespace SonarLint.VisualStudio.Integration.Helpers
                 throw new ArgumentNullException(nameof(relativeFilePath));
             }
 
+            relativeFilePath = relativeFilePath.Replace("/", "\\");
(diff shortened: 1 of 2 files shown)


src/Integration/Helpers/AbsoluteFilePathLocator.cs
```diff
@@ -45,6 +45,7 @@ public string Locate(string relativeFilePath)
                 throw new ArgumentNullException(nameof(relativeFilePath));
             }
 
+            relativeFilePath = relativeFilePath.Replace("/", "\\");
             relativeFilePath = relativeFilePath.TrimStart('\\');
 
             var absoluteFilePaths = new string[1];
```
rita-sonarsource, line 51: This class is a wrapper around VS API. It will return null if you give it something with forward slashes.
I think we should either handle it and fix to backslashes, or throw if it's not backslashed.
IMO it's better to fix. WDYT?

georgii-borovinskikh-sonarsource (approved):
lgtm

duncanp-sonar (approved):

PR #3714, opened 2023-02-15
Fetch secrets NuGet package from Repo

### LINKED ISSUE #3710: [Infra] Fetch secrets NuGet package from Repox
### Description

The Secrets NuGet package is now owned by the Language Team (code is now in sonar-text).
The NuGet package is now published to https://repox.jfrog.io/repox/api/nuget/sonarsource-nuget-releases.

TODO
- [x] Update NuGet config to fetch secrets from Repox. Might need to trust a new publisher.
- [x] Update package version
- [x] Manually test the new Secrets binary (built from different repo using a different build pipeline)

duncanp-sonar:
Fixes #3710


duncanp-sonar:
FYI the first commit has the config changes.
The second commit is just the package updates.

src/EmbeddedSonarAnalyzer.props
```diff
@@ -12,7 +12,7 @@
 	     Secrets are special case: we get the analyzer binaries from a NuGet package, and the
 		 rule help from an "normal" Java plugin. The two artefacts are versioned separately.
 	-->
-    <EmbeddedSonarSecretsNuGetVersion>1.0.3</EmbeddedSonarSecretsNuGetVersion>
+    <EmbeddedSonarSecretsNuGetVersion>1.0.4</EmbeddedSonarSecretsNuGetVersion>
     <EmbeddedSonarSecretsJarVersion>2.0.1.611</EmbeddedSonarSecretsJarVersion>
```
rita-sonarsource, line 16: The jar is the same I presume?

nuget.config
```diff
@@ -11,15 +11,18 @@
 
   <packageSources>
     <clear />
-    <add key="slvs_input" value="https://sonarsource.pkgs.visualstudio.com/399fb241-ecc7-4802-8697-dcdd01fbb832/_packaging/slvs_input/nuget/v3/index.json" />
+    <!-- SonarSecrets package published by the Language Team -->
+    <add key="repox_releases" value="https://repox.jfrog.io/repox/api/nuget/sonarsource-nuget-releases" />
```
rita-sonarsource, line 15: Are they signed with the same certificate as the dotnet analyzers? If not, the trustedSigners need to be updated.

rita-sonarsource (approved):
LGTM just the question about the trustedSigners

nuget.config
duncanp-sonar: Sorry, forgot to mention that. The signature is the same. As a test, I commented it out and as expected the NuGet fetch failed.

src/EmbeddedSonarAnalyzer.props
duncanp-sonar: Yes. We're now using the latest released versions of the jar and the NuGet package.

PR #3716, opened 2023-02-15
Smoke test that secret detectors can be imported

duncanp-sonar:
Relates to #3710


src/CloudSecrets.UnitTests/SecretsSmokeTests.cs
```diff
@@ -0,0 +1,45 @@
...
+
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.Secrets.DotNet;
+using SonarLint.VisualStudio.Integration.UnitTests;
+
+namespace SonarLint.VisualStudio.CloudSecrets.UnitTests
+{
+    [TestClass]
+    public class SecretsSmokeTests
```
duncanp-sonar, line 29: A simple automated check that at least some secret detectors can be found in the Secrets NuGet package.

rita-sonarsource (approved):

PR #3721, opened 2023-02-15
Update embedded JS/TS analyzer to 10.0.0.20728 

### LINKED ISSUE #3543: Update embedded JS/TS analyzer to 10.0.0.20728
## What's Changed in 9.13.0
https://github.com/SonarSource/SonarJS/issues?q=is%3Aclosed+milestone%3A9.13

* Fix FNs caused by `node:` imports in several rules using `getFullyQualifiedName()` by @ilia-kebets-sonarsource in https://github.com/SonarSource/SonarJS/pull/3618
* Fix FP S2424 (`no-built-in-override`): Remove System from globals by @ilia-kebets-sonarsource in https://github.com/SonarSource/SonarJS/pull/3620
* Fix FP S6442 (`no-hook-setter-in-body`): Shouldn't raise when used conditionally by @saberduck in https://github.com/SonarSource/SonarJS/pull/3622
* Deprecate S2814 (`no-redeclare`) for TypeScript by @victor-diez-sonarsource in https://github.com/SonarSource/SonarJS/pull/3635
* Upgrade ESLint, TypeScript, and TypeScript ESLint dependencies by @yassin-kammoun-sonarsource in https://github.com/SonarSource/SonarJS/pull/3631
* Fix recursion in FQN resolution by @victor-diez-sonarsource in https://github.com/SonarSource/SonarJS/pull/3642

ugras-ergun-sonarsource:
fixes #3543

### DIFF
--- a/src/EmbeddedVsix.UnitTests/TypeScriptRuleDefinitions_SanityChecks.cs
+++ b/src/EmbeddedVsix.UnitTests/TypeScriptRuleDefinitions_SanityChecks.cs
@@ -95,6 +95,6 @@ namespace SonarLint.VisualStudio.AdditionalFiles.UnitTests
             ruleDefinition.RuleKey.EndsWith(":S2260");
 
         private static bool HasDefaultParameters(RuleDefinition ruleDefinition) =>
-                ruleDefinition.DefaultParams.Length > 0;
+                ruleDefinition.DefaultParams?.Length > 0;
     }
 }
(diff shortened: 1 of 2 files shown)


src/EmbeddedVsix.UnitTests/TypeScriptRuleDefinitions_SanityChecks.cs
```diff
@@ -95,6 +95,6 @@ private static void CheckRules(string language, IEnumerable<RuleDefinition> rule
             ruleDefinition.RuleKey.EndsWith(":S2260");
 
         private static bool HasDefaultParameters(RuleDefinition ruleDefinition) =>
-                ruleDefinition.DefaultParams.Length > 0;
+                ruleDefinition.DefaultParams?.Length > 0;
```
ugras-ergun-sonarsource, line 98: Not sure if this is the best way to solve the error in the test.

duncanp-sonar: If I understand correctly, you've changed the test so that it won't fail if a rule doesn't have default parameters.

Does our product handle this case? Presumably before it was assuming that the default parameters might be empty but not null.

Is this actually a change of format by the JS/TS bubble, and if so was it intended?

Issue #3650, opened 2023-02-07
Hyperlinks in issue tooltips do not appear to be clickable


duncanp-sonar:
### Description

The appearance of the hyperlinks in non-Roslyn SonarLint issues in error tooltips do not change when the cursor is over them.
In addition, the links appear to two clicks to activate, rather than one.

cf the Microsoft issues, the appearance of which does change when the cursor is over the tooltip.

#### Screenshot - hovering over cpp:S3135
Appearance does not change i.e. no underline to show the element is clickable (although it can be clicked on)

![image](https://user-images.githubusercontent.com/10757559/217316388-fdbb66b9-493c-4fe5-9fc3-615e705841ac.png)

#### Screenshot - hovering over C6834
Underline appears.

![image](https://user-images.githubusercontent.com/10757559/217316532-e9898d0d-75db-43ad-87e7-9336aaa0e2ec.png)

duncanp-sonar:
We have a class that generates the tooltip WPF: [ErrorTagTooltipProvider](https://github.com/SonarSource/sonarlint-visualstudio/blob/e5362abad51eb7655cf8b310bcc3801217b0a054/src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs#L62)

There are two issues:
1) styling
2) requires double-click.

The second issue would be fixed if we set the URI for the link and handled `RequestNavigate` instead of executing a command, which is what we are currently doing.

The first might be fixed by finding and applying the correct style from VS.

PR #3726, opened 2023-02-16
Fix styling of tooltip hyperlinks

duncanp-sonar:
Part of #3650


duncanp-sonar:
Screenshot of the new tooltips. There are two tooltips, and the mouse is over the second one (although the mouse pointer doesn't appear in the image).

First tooltip: default appearance i.e. mouse not over the link
Second tooltip: mouse over -> link is highlighted.

image

src/IssueViz.UnitTests/Editor/ErrorTagging/ErrorTagTooltipProviderTests.cs
```diff
@@ -71,6 +71,8 @@ public void Create_CreatesTooltipWithHyperlink()
             var hyperlink = (Hyperlink)inlines[0];
             hyperlink.Command.Should().Be(navigateCommand);
             hyperlink.CommandParameter.Should().Be("some rule");
+
+            hyperlink.TextDecorations.Should().BeNull(); // no decorations by default i.e. not underlined
```
duncanp-sonar, line 75: We can't easily check the mouse-over behaviour, but we can at least check that the tooltip is not underlined by default.

PR #3728, opened 2023-02-16
Change tooltip command handling to use NavigateUri event

duncanp-sonar:
Part of #3650

### DIFF
--- a/src/IssueViz.UnitTests/Editor/ErrorTagging/ErrorTagTooltipProviderTests.cs
+++ b/src/IssueViz.UnitTests/Editor/ErrorTagging/ErrorTagTooltipProviderTests.cs
@@ -69,8 +69,12 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.ErrorTaggin
             (inlines[2] as Run).Text.Should().Be("some message");
 
             var hyperlink = (Hyperlink)inlines[0];
-            hyperlink.Command.Should().Be(navigateCommand);
-            hyperlink.CommandParameter.Should().Be("some rule");
+
+            // Click-handling
+            // Can't easily test that the NavigateUri event is wired up correctly
+            // - relying on manual test
+            hyperlink.Command.Should().BeNull();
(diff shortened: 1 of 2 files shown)


src/IssueViz.UnitTests/Editor/ErrorTagging/ErrorTagTooltipProviderTests.cs
```diff
@@ -69,8 +69,12 @@ public void Create_CreatesTooltipWithHyperlink()
             (inlines[2] as Run).Text.Should().Be("some message");
 
             var hyperlink = (Hyperlink)inlines[0];
-            hyperlink.Command.Should().Be(navigateCommand);
-            hyperlink.CommandParameter.Should().Be("some rule");
+
+            // Click-handling
+            // Can't easily test that the NavigateUri event is wired up correctly
+            // - relying on manual test
```
duncanp-sonar, line 75: I've manually tested the change and it works as expected.

src/IssueViz/Editor/ErrorTagging/ErrorTagTooltipProvider.cs
```diff
@@ -44,6 +45,7 @@ internal interface IErrorTagTooltipProvider
     [PartCreationPolicy(CreationPolicy.Shared)]
     internal class ErrorTagTooltipProvider : IErrorTagTooltipProvider
     {
+        private static readonly Uri dummyUri = new Uri("rules://sonarlint");
```
ugras-ergun-sonarsource, line 48: Should we make it a real url just in case if we fail to override it due to a bug or something. Maybe "www.sonarsource.com" ? Just in case?

duncanp-sonar: I wondered about using a real URL but thought that might suggest to a casual reader that we expect to navigate to it.

Just setting the Uri property does nothing on its own - nothing will happen unless something registers to handle the event. So not having a real URL should be ok.

ugras-ergun-sonarsource (approved):

PR #3738, opened 2023-02-17
Ue/sonarruleshelp

ugras-ergun-sonarsource:


duncanp-sonar (commented):
Looks good. Did you ask @marco-comi-sonarsource for his opinion?

ugras-ergun-sonarsource:
> Looks good. Did you ask @marco-comi-sonarsource for his opinion?

Yes he said "I think Sonar makes sense at least in places where the concept is clearly not specific to SonarLint - and the rule description is one of those 
for the second point, i don't have any strong preference TBH - maybe in doubt we can drop the title in the document"

duncanp-sonar (approved):

PR #3741, opened 2023-02-17
Expect exceptions from SSEStreamReader

rita-sonarsource:


src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionTests.cs
```diff
@@ -124,20 +125,34 @@ public async Task PumpAllAsync_WhenUnsupportedEvent_Ignores()
...
+        testScope.CapturedSessionToken.Value.IsCancellationRequested.Should().BeTrue();
+    }
+
+    [TestMethod]
+    public async Task PumpAllAsync_WhenPublisherErrors_CriticalError_SessionThrows()
+    {
+        var testScope = new TestScope();
+        testScope.SetUpSwitchToBackgroundThread();
+        var sseStreamMock = testScope.SetUpSQServiceToSuccessfullyReturnSSEStreamReader();
+        sseStreamMock.Setup(x => x.ReadAsync()).Throws(new StackOverflowException("this is a test"));
```
georgii-borovinskikh-sonarsource, line 149: That's probably one of the worst exceptions to actually use as an example:

```
Starting with 2.0 a StackOverflow Exception can only be caught in the following circumstances.

The CLR is being run in a hosted environment* where the host specifically allows for StackOverflow exceptions to be handled
The stackoverflow exception is thrown by user code and not due to an actual stack overflow situation ([Reference](https://learn.microsoft.com/en-us/archive/blogs/jaredpar/when-can-you-catch-a-stackoverflowexception))
*"hosted environment" as in "my code hosts CLR and I configure CLR's options" and not "my code runs on shared hosting"
```

Choose a different one please

src/ConnectedMode/ServerSentEvents/SSESessionFactory.cs
```diff
@@ -149,8 +157,10 @@ public async Task PumpAllAsync()
                                 break;
                         }
                     }
-                    catch (ObjectDisposedException)
+                    catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
                     {
+                        logger.LogVerbose($"[SSESession] Failed to handle events: {ex}");
```
georgii-borovinskikh-sonarsource, line 162: This logs in Debug, right?

georgii-borovinskikh-sonarsource (changes requested):

duncanp-sonar (approved):

src/ConnectedMode/ServerSentEvents/SSESessionFactory.cs
rita-sonarsource: Yep

Issue #3680, opened 2023-02-10
Fix tests to use interface of ProjectToLanguageMapper instead of using the actual class.


bigfluffycookie:
There are a few tests that have this in a TestInitialize

```
var projectToLanguageMapper = new ProjectToLanguageMapper(Mock.Of<ICMakeProjectTypeIndicator>(),  Mock.Of<IJsTsProjectTypeIndicator>(),  Mock.Of<IConnectedModeSecrets>());
var mefHost = ConfigurableComponentModel.CreateWithExports(MefTestHelpers.CreateExport<IProjectToLanguageMapper>(projectToLanguageMapper));
```

We should change this to instead set up a Mock with IProjectToLanguageMapper instead.

Classes to fix:

- [ ] BindingProcessImplTests
- [ ] CSharpVBBindingOperationTests
- [ ] NuGetBindingOperationTests
- [ ] SolutionBindingOperationTests
- [ ] ConnectionWorkflowTests
- [ ] ErrorListInfoBarController.QualityProfileBackgroundProcessorTests
- [ ] ErrorListInforBarControllerTests
- [ ] ConflictsManagerTests

PR #3742, opened 2023-02-17
fix BindingProcessingImplTests a little bit

bigfluffycookie:
Part of https://github.com/SonarSource/sonarlint-visualstudio/issues/3680

### DIFF
--- a/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
@@ -37,8 +37,6 @@ using NuGet;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Analysis;
 using SonarLint.VisualStudio.Core.Binding;
-using SonarLint.VisualStudio.Core.CFamily;
-using SonarLint.VisualStudio.Core.Secrets;
 using SonarLint.VisualStudio.Integration.Binding;
 using SonarLint.VisualStudio.Integration.NewConnectedMode;
 using SonarLint.VisualStudio.Integration.Resources;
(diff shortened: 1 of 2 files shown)


src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -822,15 +836,15 @@ public void InitializeSolutionBinding_Update_NotAllUpToDate_SomeProjectsUpdated(
             // Arrange
```
bigfluffycookie, line 836: I dont know if I should have left it with names, but as we test the output strings i figured it makes sense like this

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -58,7 +56,7 @@ public class BindingProcessImplTests
         private ConfigurableSolutionRuleSetsInformationProvider ruleSetsInformationProvider;
         private ConfigurableHost host;
         private Mock<IFolderWorkspaceService> folderWorkspaceService;
-        private Mock<IJsTsProjectTypeIndicator> jstsIndicator;
```
duncanp-sonar, line 61: Getting rid of `IJsTsProjectTypeIndicator` - nice.

That should have been an indicator to whoever originally wrote the tests that something was wrong, since `IJsTsProjectTypeIndicator` is not a dependency of the type being tested here, `BindingProcessImpl`.

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -772,9 +784,9 @@ public void GetProjectsForRulesetBinding_FirstBinding_AllProjectsBound()
             // Arrange
             var allProjects = new Project[]
             {
-                new ProjectMock("cs1.csproj"),
```
duncanp-sonar, line 775: You've gotten rid of all of the uses of `ProjectMock` - nice.

duncanp-sonar (approved):
Looks good!

Issue #3767, opened 2023-02-22
Merge webclient submodule into the main repo


rita-sonarsource:
### WHY

The [webclient repo ](https://github.com/SonarSource/sonarqube-webclient-dotnet) was created originally as a separate repository in order to re-use code between .net components in the ecosystem. In practice, it's been years and it is used only by SLVS. Having it a separate submodule adds unnecessary complexity and makes development harder when adding and modifying server API endpoints.

### WHAT

Since the module is used only by SLVS, it would make sense for it to be along side SLVS code.

### HOW

Merge the git submodule into the main repository.

PR #3777, opened 2023-02-22
Cleanup webclient submodule removal

rita-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3767


duncanp-sonar:
A couple of things to check:
* the submodule repo had a [Directory.Build.props](https://github.com/SonarSource/sonarqube-webclient-dotnet/blob/master/Directory.Build.props). I don't think we need it - the settings should come from the SLVS versions.
* the submodule repo had a [hardcoded product framework version](https://github.com/SonarSource/sonarqube-webclient-dotnet/blob/bd8acc28b28936d62a155c4b050d42779058b238/SonarQube.Client/SonarQube.Client.csproj#L4). This shouldn't be required either.

duncanp-sonar (commented):
Looks good. A couple of suggestions for things to check.

rita-sonarsource:
> * the submodule repo had a [Directory.Build.props](https://github.com/SonarSource/sonarqube-webclient-dotnet/blob/master/Directory.Build.props). I don't think we need it - the settings should come from the SLVS versions.

I removed the Directory.Build.props, there are now 2 instead of 3: one under src which will apply to SLVS and client, and outer one under main root, which applies to all + build projects.

> * the submodule repo had a [hardcoded product framework version](https://github.com/SonarSource/sonarqube-webclient-dotnet/blob/bd8acc28b28936d62a155c4b050d42779058b238/SonarQube.Client/SonarQube.Client.csproj#L4). This shouldn't be required either.

Done

Issue #3778, opened 2023-02-22
Do not throw away suppressed issues


bigfluffycookie:
- Add suppression state to IAnalysisIssueVisualization
- Change IssueFilter to return all suppressed issues
- Change IssueConsumer to mark issues as suppressed

PR #3779, opened 2023-02-22
Change code so that suppressed issues will no longer be thrown away-

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3778

### DIFF
--- a/src/Core/Suppressions/IIssuesFilter.cs
+++ b/src/Core/Suppressions/IIssuesFilter.cs
@@ -24,6 +24,6 @@ namespace SonarLint.VisualStudio.Core.Suppression
 {
     public interface IIssuesFilter
     {
-        IEnumerable<IFilterableIssue> Filter(IEnumerable<IFilterableIssue> issues);
+        IEnumerable<IFilterableIssue> GetMatches(IEnumerable<IFilterableIssue> issues);
     }
 }
--- a/src/Integration.UnitTests/Suppression/IssuesFilterTests.cs
+++ b/src/Integration.UnitTests/Suppression/IssuesFilterTests.cs
@@ -64,50 +64,50 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Suppression
         [TestMethod]
         public void Filter_NullIssues_ThrowsArgumentNullException()
         {
-            Action act = () => testSubject.Filter(null);
+            Action act = () => testSubject.GetMatches(null);
             act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("issues");
         }
 
         [TestMethod]
-        public void Filter_NoMatchingIssues_ReturnsOriginalIssues()
+        public void Filter_NoMatchingIssues_ReturnsNoIssues()
         {
             // Arrange
             var inputIssues = new[] { Issue1, Issue2, Issue3 };
             SetMatchableIssues(Issue4, Issue5);
 
             // Act
-            var result = testSubject.Filter(inputIssues);
+            var result = testSubject.GetMatches(inputIssues);
 
             // Assert
-            result.Should().BeEquivalentTo(inputIssues);
+            result.Should().BeEmpty();
         }
 
         [TestMethod]
-        public void Filter_SomeIssuesMatch_ReturnsUnmatchedIssues()
+        public void Filter_SomeIssuesMatch_ReturnsMatchedIssues()
         {
             // Arrange
             var inputIssues = new[] { Issue1, Issue2, Issue3, Issue4 };
             SetMatchableIssues(Issue1, Issue2, Issue5);
 
             // Act
-            var result = testSubject.Filter(inputIssues);
+            var result = testSubject.GetMatches(inputIssues);
 
             // Assert
-            result.Should().BeEquivalentTo(Issue3, Issue4);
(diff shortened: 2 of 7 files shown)


src/Core/Suppressions/IIssuesFilter.cs
```diff
@@ -24,6 +24,6 @@ namespace SonarLint.VisualStudio.Core.Suppression
 {
     public interface IIssuesFilter
     {
-        IEnumerable<IFilterableIssue> Filter(IEnumerable<IFilterableIssue> issues);
```
bigfluffycookie, line 27: As discussed offline, renamed it so its a bit clearer what this method does.

src/Integration.UnitTests/Suppression/IssuesFilterTests.cs
```diff
@@ -64,50 +64,50 @@ public void Ctor_NullSonarQubeIssuesProvider_ThrowsArgumentNullException()
         [TestMethod]
```
bigfluffycookie, line 64: In this class I basically just reverted all tests to match the new code.

src/Integration.Vsix.UnitTests/Analysis/IssueHandlerTests.cs
```diff
@@ -177,37 +168,38 @@ public void HandleNewIssues_HasFileLevelIssue_IssueIsReturned()
...
             notificationHandler.InvocationCount.Should().Be(1);
-            notificationHandler.UpdatedSnapshot.Issues.Count().Should().Be(0);
-        }
+            notificationHandler.UpdatedSnapshot.Issues.Count().Should().Be(4);
 
-        private static IIssuesFilter CreatePassthroughIssuesFilter()
-            => CreateIssuesFilter(null).Object;
+            issue1.IsSuppressed = true;
 
-        private static Mock<IIssuesFilter> CreateIssuesFilter(IEnumerable<IFilterableIssue> optionalDataToReturn = null)
```
bigfluffycookie, line 203: This is not really needed anymore as calling the issue filter does not alter the list of issues anymore.

src/Integration.Vsix.UnitTests/Analysis/IssueHandlerTests.cs
```diff
@@ -79,34 +79,30 @@ public void HandleNewIssues_UpdatedSnapshotHasExpectedValues()
         }
```
bigfluffycookie, line 80: Here I also mostly inverted tests or in some cases even took out issuefilter tests as they were testing if the list has changed which, now it no longer does. I added 1 test to check if the issues get marked correctly as suppressed correctly

src/Integration.Vsix.UnitTests/Analysis/IssueHandlerTests.cs
```diff
@@ -177,37 +168,38 @@ public void HandleNewIssues_HasFileLevelIssue_IssueIsReturned()
...
             notificationHandler.InvocationCount.Should().Be(1);
-            notificationHandler.UpdatedSnapshot.Issues.Count().Should().Be(1);
+            notificationHandler.UpdatedSnapshot.Issues.Count().Should().Be(0);
         }
 
         [TestMethod]
-        public void HandleNewIssues_NoIssues_NoIssuesReturned()
+        public void HandleNewIssues_SomeSuppressedIssues_IssuesGetMarkedCorrectly()
         {
             // Arrange
```
bigfluffycookie, line 177: There isnt really a good reason why I chose 4 issues instead of 2 but it felt better to me

duncanp-sonar (approved):
Looks good. Can you easily fix the low code coverage? (if so, it would be worth doing)

src/IssueViz.UnitTests/Models/AnalysisIssueVisualizationTests.cs
```diff
@@ -190,6 +190,22 @@ public void SetSpan_HasSubscribers_NotifiesSubscribers()
             testSubject.Span.Should().Be(newSpan);
         }
 
+        [TestMethod]
+        public void SetIsSuppressed_HasSubscribers_VerifyRaised()
+        {
+            var testSubject = CreateTestSubject();
+
```
duncanp-sonar, line 197: I would add an assertion that `IsSuppressed` is `false` to start with, otherwise the test doesn't guarantee that the value has changed.

duncanp-sonar (approved):
Looks good, and apologies for missing the "notify property changed" part earlier.

One minor suggestion about adding another assertion.

src/IssueViz.UnitTests/Models/AnalysisIssueVisualizationTests.cs
```diff
@@ -194,7 +194,9 @@ public void SetSpan_HasSubscribers_NotifiesSubscribers()
         public void SetIsSuppressed_HasSubscribers_VerifyRaised()
         {
             var testSubject = CreateTestSubject();
-            
+
+            Assert.IsFalse(testSubject.IsSuppressed);
```
duncanp-sonar, line 198: Minor: `testSubject.IsSuppressed.Should().BeFalse();` would be better (it's cleaner not to mix old-style `Assert.IsTrue/False` with `FluentAssertions`).

PR #3780, opened 2023-02-22
 Add new fields to IRuleInfo

### LINKED ISSUE #3758: add new fields to IRuleInfo
- descriptionSections
- educationPrinciples

ugras-ergun-sonarsource:
fixes #3758

### DIFF
--- a/build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
+++ b/build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
@@ -111,7 +111,9 @@ internal class RuleDescExtractor
                 ConvertPluginSeverity(pluginRule.DefaultSeverity),
                 ConvertPluginIssueType(pluginRule.Type),
                 Convert.ToBoolean(pluginRule.IsActiveByDefault),
-                pluginRule.Tags ?? Array.Empty<string>()
+                pluginRule.Tags ?? Array.Empty<string>(),
+                null,
+                null
                 );
 
             SaveRuleFile(slvsRule);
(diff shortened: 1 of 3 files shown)


build/tools/ExtractRuleDescFromJson/RuleDescExtractor.cs
```diff
@@ -111,7 +111,9 @@ private void ProcessRule(PluginRule pluginRule)
                 ConvertPluginSeverity(pluginRule.DefaultSeverity),
                 ConvertPluginIssueType(pluginRule.Type),
                 Convert.ToBoolean(pluginRule.IsActiveByDefault),
-                pluginRule.Tags ?? Array.Empty<string>()
+                pluginRule.Tags ?? Array.Empty<string>(),
+                null,
```
ugras-ergun-sonarsource, line 115: for now it's null to avoid compile errors. It should be implemented with extraction task.

src/Rules/IRuleInfo.cs
```diff
@@ -44,6 +44,45 @@ public enum RuleIssueType
...
+        public DescriptionSection(string key, string content, IReadOnlyList<IContext> context = null)
+        {
+            Key = key;
+            Content = content;
+            Context = context;
+        }
+
+        public string Key { get; }
+        public string Content { get; }
+        public IReadOnlyList<IContext> Context { get; }
```
georgii-borovinskikh-sonarsource, line 83: Have we considered using JetBrains Annotations? [CanBeNull] here would be helpful for understanding

georgii-borovinskikh-sonarsource (approved):

src/Rules/IRuleInfo.cs
ugras-ergun-sonarsource: I don't think adding a new dependency for this purpose makes sense but I have added some remarks to make it clearer.

ugras-ergun-sonarsource:
Changed the field names according to the plugin-api

georgii-borovinskikh-sonarsource (approved):

Issue #3782, opened 2023-02-22
Change ErrorTagger to not show issues that are suppressed.


bigfluffycookie:
- Add it to Filter()

PR #3785, opened 2023-02-23
Change errorTagger to filter suppressed issues-

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3782

### DIFF
--- a/src/IssueViz.UnitTests/Editor/ErrorTagging/ErrorTaggerTests.cs
+++ b/src/IssueViz.UnitTests/Editor/ErrorTagging/ErrorTaggerTests.cs
@@ -42,16 +42,24 @@ namespace SonarLint.VisualStudio.IssueVisualization.UnitTests.Editor.ErrorTaggin
             var inputSpans = CreateSpanCollectionSpanningWholeSnapshot(snapshot);
 
             var primary1 = CreateTagSpanWithPrimaryLocation(snapshot, new Span(1, 5), message: "error message 1", ruleKey: "cpp:S5350");
+            var primary1_Suppressed = CreateTagSpanWithPrimaryLocation(snapshot, new Span(2, 6));
             var primary2 = CreateTagSpanWithPrimaryLocation(snapshot, new Span(10, 5), message: "error message 2", ruleKey: "cpp:emptyCompoundStatement");
+            var primary2_Suppressed = CreateTagSpanWithPrimaryLocation(snapshot, new Span(8, 15));
             var secondary1 = CreateTagSpanWithSecondaryLocation(snapshot, new Span(20, 5));
             var secondary2 = CreateTagSpanWithSecondaryLocation(snapshot, new Span(30, 5));
-            var aggregator = CreateAggregator(primary1, secondary1, primary2, secondary2);
+            var aggregator = CreateAggregator(primary1, secondary1, primary1_Suppressed, primary2, secondary2, primary2_Suppressed);
 
             var tooltipProvider = new Mock<IErrorTagTooltipProvider>();
             var issue1 = (primary1.Tag.Location as IAnalysisIssueVisualization).Issue;
             var issue2 = (primary2.Tag.Location as IAnalysisIssueVisualization).Issue;
+
             tooltipProvider.Setup(x => x.Create(issue1)).Returns("some tooltip1");
             tooltipProvider.Setup(x => x.Create(issue2)).Returns("some tooltip2");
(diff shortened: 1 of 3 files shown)


src/IssueViz.UnitTests/Editor/ErrorTagging/ErrorTaggerTests.cs
```diff
@@ -42,17 +42,25 @@ public void GetTags_FilterIsApplied_ExpectedTagsCreated()
             var inputSpans = CreateSpanCollectionSpanningWholeSnapshot(snapshot);
```
bigfluffycookie, line 42: I added to more tag spans that are both suppressed however now that I am looking at it it feels a bit ugly that I do not set the others to NotSuppressed. What are your thoughts? Also is this adequate test coverage in your opinion?

duncanp-sonar: You could add an optional property `bool isSuppressed = false` to the helper methods `CreateTagSpanWithPrimaryLocation` and `CreateIssueViz`.

That would also avoid the ugly casting in lines 59 - 62 to set the `IsSuppressedProperty`.

src/IssueViz/Editor/ErrorTagging/ErrorTagger.cs
```diff
@@ -45,9 +45,15 @@ protected override TagSpan<IErrorTag> CreateTagSpan(IIssueLocationTag trackedTag
         }
 
         protected override IEnumerable<IMappingTagSpan<IIssueLocationTag>> Filter(IEnumerable<IMappingTagSpan<IIssueLocationTag>> trackedTagSpans) =>
-            trackedTagSpans.Where(x => IsValidPrimaryLocation(x.Tag.Location));
+            trackedTagSpans.Where(x => IsValidPrimaryLocation(x.Tag.Location) && IsNotSuppressed(x.Tag.Location));
```
duncanp-sonar, line 48: You could write the checks as follows (which I think should give the same result):

```
        protected override IEnumerable<IMappingTagSpan<IIssueLocationTag>> Filter(IEnumerable<IMappingTagSpan<IIssueLocationTag>> trackedTagSpans) =>
            trackedTagSpans.Where(x => x.Tag.Location is IAnalysisIssueVisualization issueViz
                && !issueViz.IsSuppressed
                && IsValidPrimaryLocation(issueViz));

        private static bool IsValidPrimaryLocation(IAnalysisIssueVisualization viz) => viz.Span.IsNavigable();
```

There are a couple of advantages:
* the cast to `IAnalysisIssueVisualization` is only done once
* there is a minor performance gain in checking `IsSuppressed` first, since it is a simple property. It's only a small difference, but this method is called thousands to times so it will add up. 

I didn't add a helper method for `!issueViz.IsSuppressed` as I though it was simple enough not to need one.

src/IssueViz/Editor/ErrorTagging/ErrorTagger.cs
```diff
@@ -45,9 +45,15 @@ protected override TagSpan<IErrorTag> CreateTagSpan(IIssueLocationTag trackedTag
...
 
         private static bool IsValidPrimaryLocation(IAnalysisIssueLocationVisualization locViz) =>
             locViz is IAnalysisIssueVisualization && locViz.Span.IsNavigable();
+
+        private static bool IsNotSuppressed(IAnalysisIssueLocationVisualization locViz)
+        {
+            var issueViz = (IAnalysisIssueVisualization)locViz;
```
duncanp-sonar, line 55: `(IAnalysisIssueVisualization)locViz` is a hard cast, which will fail if the types don't match.
The old `IsValidPrimaryLocation` used `is IAnalysisIssueVisualization` which is a soft cast, and will return `false` if the types don't match.

The suggested new version of the code just uses `is`.

duncanp-sonar (commented):
See inline for a suggested alternative implementation.

src/IssueViz/Editor/ErrorTagging/ErrorTagger.cs
bigfluffycookie: Ahh okay, yea the reason I wrote a helper method is because I was unaware of this syntax -> x.Tag.Location **is** 
so it was the only way for me to cast things but this is much better!

duncanp-sonar (approved):

PR #3787, opened 2023-02-23
Move SonarQube.Client to src folder

### LINKED ISSUE #3767: Merge webclient submodule into the main repo

rita-sonarsource:
Fixes #3767


src/SonarQube.Client.Tests/SonarQube.Client.Tests.csproj
```diff
@@ -1,5 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
 
+  <Import Project="..\SonarLint.Test.props" />
+
   <PropertyGroup>
     <TargetFramework>net472</TargetFramework>
```
duncanp-sonar, line 6: This is set in `SonarLint.Test.props`. Deleting it _shouldn't_ cause any problems, but given the issues you've had so far with this PR I'm wouldn't take anything for granted.

I've create a [separate branch and PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/3799) with just that change to see happens on the CI machine. If it works then we can commit that PR, otherwise this one.

Issue #3795, opened 2023-02-27
Add IssueServerEventsListener interface


bigfluffycookie:
- add a skeleton implementation

PR #3796, opened 2023-02-27
Add IssueServerEventsListener

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3795

### DIFF
--- a/src/ConnectedMode.UnitTests/Suppressions/IssueServerEventsListenerTests.cs
+++ b/src/ConnectedMode.UnitTests/Suppressions/IssueServerEventsListenerTests.cs
@@ -1,0 +1,103 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Threading.Tasks;
+using SonarLint.VisualStudio.ConnectedMode.Suppressions;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.TestInfrastructure;
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
+{
+    [TestClass]
+    public class IssueServerEventsListenerTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<IssueServerEventsListener, IIssueServerEventsListener>(
+                MefTestHelpers.CreateExport<IIssueServerEventSource>(),
+                MefTestHelpers.CreateExport<IThreadHandling>(),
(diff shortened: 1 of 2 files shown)


src/ConnectedMode.UnitTests/Suppressions/IssueServerEventsListenerTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<IssueServerEventsListener, IIssueServerEventsListener>(
+                MefTestHelpers.CreateExport<IIssueServerEventSource>(),
+                MefTestHelpers.CreateExport<IThreadHandling>(),
+                MefTestHelpers.CreateExport<ILogger>());
+        }
+
+        [TestMethod]
```
bigfluffycookie, line 43: I am missing a test for non critical exceptions however because the code is so bare I couldn't find a reasonable way to add it (The only method that could throw is issueServerEventSource.GetNextEventOrNullAsync and if that one throws the code will be in an infinite loop unless I dispose it)

src/ConnectedMode.UnitTests/Suppressions/IssueServerEventsListenerTests.cs
```diff
@@ -0,0 +1,103 @@
+﻿/*
```
bigfluffycookie, line 1: These test were copied from TainServerEventsListenerTests and adjusted to match the new skeleton.

src/ConnectedMode.UnitTests/Suppressions/IssueServerEventsListenerTests.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            var listenTask = testSubject.ListenAsync();
+            testSubject.Dispose();
+
+            await listenTask;
+
+            issueServerEventSource.Verify(x => x.GetNextEventOrNullAsync(), Times.Once);
+        }
+
+        private static Mock<IIssueServerEventSource> SetupIssueServerEventSource(params IIssueChangedServerEvent[] serverEvents)
```
bigfluffycookie, line 74: I realize now that we dont actually need this method yet but I felt like it made sense to already add it. Lmk if you disagree.

src/ConnectedMode/Suppressions/IssueServerEventsListener.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    /// <summary>
+    /// Handles <see cref="IIssueServerEvent"/> coming from the server.
+    /// </summary>
+    internal interface IIssueServerEventsListener : IDisposable
+    {
+        Task ListenAsync();
+    }
+
+    [Export(typeof(IIssueServerEventsListener))]
```
bigfluffycookie, line 39: I tried making it as bare as possible but I am not sure if I added too much already.

src/ConnectedMode/Suppressions/IssueServerEventsListener.cs
```diff
@@ -0,0 +1,91 @@
...
+            cancellationTokenSource = new CancellationTokenSource();
+        }
+
+        public async Task ListenAsync()
+        {
+            await threadHandling.SwitchToBackgroundThread();
+
+            while (!cancellationTokenSource.IsCancellationRequested)
+            {
+                try
```
bigfluffycookie, line 66: In the Taint version of this class there is a switch statement however, I don't think in our case it makes much sense. (Maybe we need to change it later)

src/ConnectedMode/Suppressions/IssueServerEventsListener.cs
duncanp-sonar: I think it's ok. If we find something we don't need we can take it out later, but I think we'll need everything that is there.

src/ConnectedMode.UnitTests/Suppressions/IssueServerEventsListenerTests.cs
duncanp-sonar: We will need it soon so may as well keep it.

src/ConnectedMode/Suppressions/IssueServerEventsListener.cs
```diff
@@ -0,0 +1,91 @@
...
+                }
+            }
+        }
+
+        public void Dispose()
+        {
+            cancellationTokenSource.Cancel();
+            cancellationTokenSource?.Dispose();
+        }
+    }
```
duncanp-sonar, line 90: There's a lot of overlap between the taint and issue tests and implementation.
We should consider refactoring later to create a base class if it turns out that the only differences in behaviour are effectively in line 76 when recognised events are processed.

PR #3798, opened 2023-02-27
Support suppressions in the Error List

### LINKED ISSUE #3781: Support suppressions in ErrorList:
- add TableColumnDefinition in SonarErrorListDataSource 
- adjust issuesnsapshot to return stuff

duncanp-sonar:
Fixes #3781


duncanp-sonar:
### VS2022 screenshots
Works as expected:
image

image

### VS2019.11 screenshots
image

image

### VS2019.3 screenshots
There is one oddity in VS2019.3 - the "Suppressions State" column appears twice in the context menu, and the user can show both. However, the second column is always empty.

image

image

There is only one entry in the context menu in VS2019.11, so it looks like a VS bug that has been fixed. In any event, it's cosmetic, and both columns will only appear if the user chooses to show both.

image

src/Integration.Vsix.UnitTests/ErrorList/IssuesSnapshotTests _GetValue.cs
```diff
@@ -251,6 +252,15 @@ public void GetValue_Issue()
             GetValue(SonarLintTableControlConstants.IssueVizColumnName).Should().BeSameAs(issueViz);
         }
 
+        [TestMethod]
+        public void GetValue_SuppressionState_Is_SuppressionState()
+        {
+            issueViz.IsSuppressed = true;
+            GetValue(SuppressionsColumnHelper.SuppressionStateColumnName).Should().BeSameAs(SuppressionsColumnHelper.SuppressionState_Suppressed);
```
duncanp-sonar, line 259: @bigfluffycookie the first commit is your original one. The only changes I have made are to add the `SuppressionColumnHelper` and change your version to use those constants/properties instead of the VS version

e.g. here, using `SuppressionsColumnHelper.SuppresionStateColumnName` instead of `StandardTableKeyNames.SuppressionState`.

src/Integration.Vsix/ErrorList/SuppressionsColumnHelper.cs
```diff
@@ -0,0 +1,57 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.VisualStudio.Shell.TableManager;
```
duncanp-sonar, line 21: In the end I decided to use conditional compilation. That means in VS2022 we are directly using the correct VS types/constants, and the hacky solution is only necessary for VS2019 (which we expect fewer people to be using, and for which we will drop support at some point).

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -59,6 +59,13 @@
          used to build the project/solution.
   -->
 
+  <PropertyGroup>
+    <!-- Conditional compilation constant for VS2019-specific code.
+         We'd prefer to set this the shared SonarLint.VSSpecificAssemblies.props file, but
+         the IDE isn't picking up the setting if we do that. -->
+    <DefineConstants Condition="($(VisualStudioVersion)=='17.0' AND $(BuildingInsideVisualStudio)=='true') OR $(VSTargetVersion)=='2022'">$(DefineConstants);VS2022</DefineConstants>
+  </PropertyGroup>
```
duncanp-sonar, line 67: This is copied from one of the other projects that uses conditional compilation e.g. [Infrastructure.VS](https://github.com/SonarSource/sonarlint-visualstudio/blob/ae2101c36530f705e198e837b6c6302d8b61d4ae/src/Infrastructure.VS/Infrastructure.VS.csproj#L10).

src/Integration.Vsix/ErrorList/SuppressionsColumnHelper.cs
bigfluffycookie: Yea I think that makes most sense!

bigfluffycookie (approved):
Looks good!

PR #3810, opened 2023-03-01
Add IClientIssueStore Interface

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3809


src/ConnectedMode/IClientIssueStore.cs
```diff
@@ -0,0 +1,33 @@
+﻿/*
```
bigfluffycookie, line 1: The actual implementation will be in seperate pr.

src/ConnectedMode/IClientIssueStore.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System.Collections.Generic;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.ConnectedMode
+{
+    public interface IClientIssueStore
+    {
+        /// <summary>
+        /// Returns all IAnalysisIssueVisualizations.
```
duncanp-sonar, line 29: Or "/// Returns all issues that have been detected locally".

Historically we haven't made a very clear separation between server-side and "local" issue data, so being more explicit wouldn't hurt.

src/ConnectedMode/IClientIssueStore.cs
```diff
@@ -0,0 +1,33 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.ConnectedMode
+{
+    public interface IClientIssueStore
```
duncanp-sonar, line 26: This interface isn't specific to Connected Mode, so perhaps it should be somewhere else.

Ideally it would be in Core, but as discussed we can't do that yet. For now, perhaps putting it in IssueVisualization would be better. What do you think?

duncanp-sonar (commented):
Just a question about where the new interface should go

duncanp-sonar (approved):

PR #3812, opened 2023-03-01
Add implementation of IClientIssueStore

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3811


src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_ClientIssueStoreTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            testSubject.AddFactory(factory);
+            Get_ShouldContainIssues(testSubject, issue1, issue2);
+
+            testSubject.RemoveFactory(factory);
+            Get_ShouldContainIssues(testSubject);
+        }
+
+        [TestMethod]
+        public void Get_ReturnsCorrectIssuesTwoFactorîes()
```
bigfluffycookie, line 56: I'll fix the "î" in factories with other change request.

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_ClientIssueStoreTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            Get_ShouldContainIssues(testSubject, issue1, issue2);
+
+            testSubject.AddFactory(factory2);
+            Get_ShouldContainIssues(testSubject, issue1, issue2, issue3, issue4);
+
+            testSubject.RemoveFactory(factory);
+            Get_ShouldContainIssues(testSubject, issue3, issue4);
+        }
+
+        private void Get_ShouldContainIssues(SonarErrorListDataSource testSubject, params IAnalysisIssueVisualization[] issues)
```
bigfluffycookie, line 80: I know we have some test classes with a syntax similar to this however, I couldn't remember any examples so I did it like this, please lmk if its oki

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
```diff
@@ -73,8 +74,10 @@ public void MefCtor_CheckExports()
             // Set up importers for each of the interfaces exported by the test subject
```
bigfluffycookie, line 74: Usually we do not check mef exports like this do we?

src/Integration.Vsix/ErrorList/SonarErrorListDataSource.cs
```diff
@@ -249,6 +251,21 @@ public bool Contains(IAnalysisIssueVisualization issueVisualization)
 
         #endregion IIssueLocationStore implementation
 
+        #region IClientIssueStore implementation
+
+        public IEnumerable<IAnalysisIssueVisualization> Get()
+        {
```
bigfluffycookie, line 257: I forgot if I was supposed to add a "var currentFactories = this.factories" before the loop or not

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
duncanp-sonar: This one is a special case because it exports more than one interface and is a singleton, so the usual test helper doesn't do what we need.
We want to check two things:
1) it exports all of the expected interfaces, and 
2) all of those exports return the same instance of the class.

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSourceTests.cs
```diff
@@ -86,7 +89,7 @@ public void MefCtor_CheckExports()
                 // Both imports should be satisfied...
                 errorDataSourceImporter.Import.Should().NotBeNull();
                 issueLocationStoreImporter.Import.Should().NotBeNull();
-
+                clientIssueStore.Import.Should().NotBeNull();
```
duncanp-sonar, line 92: To check point (2) mentioned in the comment above, you could also add the following:

`
clientIssueStore.Import.Should().BeSameAs(errorDataSourceImporter.Import);
`

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_ClientIssueStoreTests.cs
```diff
@@ -0,0 +1,112 @@
+﻿/*
```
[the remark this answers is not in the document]
duncanp-sonar: Yep; keeps things simpler.

If a class implements multiple interfaces, then sometimes it's worth making the product code class `partial` and splitting it across multiple files (one file per interface). This new interface is simple enough that I don't think it's necessary to split the product code class in this case.

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_ClientIssueStoreTests.cs
duncanp-sonar: Or maybe we should use `î` everywhere...

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_ClientIssueStoreTests.cs
duncanp-sonar: Have a do-and-check method is ok.

Here, there are FluentAssertions [that compare collections](https://fluentassertions.com/collections/) are equivalent, so you could perhaps do the whole check in single line
e.g. `testSubject.Get().Should().Equal(issues)`

duncanp-sonar (approved):

duncanp-sonar (approved):

PR #3815, opened 2023-03-01
Rich rule description logical representation and visual representation generation

georgii-borovinskikh-sonarsource:


src/Education.UnitTests/Layout/TabGroupTests.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        private static void VerifyButtonsGeneratedCorrectly(Section visualizationRoot, string tabGroupName, string[] tabNames, string[] tabDisplayNames)
+        {
+            var buttonsBlock = (BlockUIContainer)visualizationRoot.Blocks.FirstBlock;
+            buttonsBlock.Name.Should().Be(tabGroupName);
+            buttonsBlock.Child.Should().BeOfType<StackPanel>();
+            var buttons = ((StackPanel)buttonsBlock.Child).Children.Cast<ToggleButton>().ToList();
+            buttons.Count.Should().Be(tabNames.Length);
+
+            for (var i = 0; i < buttons.Count; i++)
```
ugras-ergun-sonarsource, line 68: instead maybe use Should().BeEquivalentTo()

src/Education.UnitTests/Layout/MultiBlockSectionTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            var testSubject =
+                new MultiBlockSection(headerMock.Object, child1Node.Object, child2Node.Object);
+
+            var visualization = testSubject.CreateVisualization();
+
+            visualization.Should().BeOfType<Section>();
+            var section = visualization as Section;
+            section.Blocks.Count.Should().Be(3);
+            section.Blocks.FirstBlock.Should().BeSameAs(header);
```
ugras-ergun-sonarsource, line 56: instead of checking all one by one maybe use Should().BeEquivalentTo()

src/Education/XamlParser/IXamlBlockContent.cs
```diff
@@ -0,0 +1,51 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Windows.Documents;
+
+namespace SonarLint.VisualStudio.Education.XamlParser
```
ugras-ergun-sonarsource, line 23: I am ok with either way but maybe it's better to move this to XamlGenerator

ugras-ergun-sonarsource (approved):
looks good I have a few suggestions on tests and one about namespace changes. I assume some parts will change based on the builder implementation.

src/Education/XamlParser/IXamlBlockContent.cs
georgii-borovinskikh-sonarsource: I think it's better to split the html to xaml and xaml to wpf into 2 folders

src/Education.UnitTests/Layout/TabGroupTests.cs
georgii-borovinskikh-sonarsource: I think it's too complicated to do it like that. I would still need the for loop, but instead of checking the relevant properties inside the loop, I would need to put the elements in a list and then also set up BeEquivalentTo() to only check those 2 fields

src/Education.UnitTests/Layout/MultiBlockSectionTests.cs
georgii-borovinskikh-sonarsource: Hmm, I tried changing it to that, but it just hangs the test

Issue #3813, opened 2023-03-01
Add new Interface IClientSuppressionSynchronizer


bigfluffycookie:
which Imports an IClientIssueStore and an ISuppressedIssueMatcher

PR #3816, opened 2023-03-01
Add and implement IClientSuppressionSynchronizer.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3813


src/ConnectedMode/Suppressions/ClientSuppressionSynchronizer.cs
```diff
@@ -0,0 +1,59 @@
...
+using SonarLint.VisualStudio.IssueVisualization.Models;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Suppressions
+{
+    [Export(typeof(IClientSuppressionSynchronizer))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class ClientSuppressionSynchronizer : IClientSuppressionSynchronizer
+    {
+        private readonly IClientIssueStore clientIssueStore;
+        private readonly IIssuesFilter issueFilter;
```
bigfluffycookie, line 35: Earlier I was mistaken in the assumption that I could use the ISuppressedIssueMatcher, we need to use the issueFilter instead.

src/ConnectedMode/Suppressions/ClientSuppressionSynchronizer.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        [ImportingConstructor]
+        public ClientSuppressionSynchronizer(IClientIssueStore clientSideIssueStore, IIssuesFilter issueFilter)
+        {
+            this.clientIssueStore = clientSideIssueStore;
+            this.issueFilter = issueFilter;
+        }
+
+        public void SynchronizeSuppressedIssues()
```
bigfluffycookie, line 44: This method is extremely similar to the method "MarkSuppressedIssues" in "IssueConsumerFactory_IssueHandler.cs" I've added a note on our board incase we should change the other class to also use this one.

src/ConnectedMode.UnitTests/Suppressions/ClientSuppressionSynchronizerTests.cs
```diff
@@ -0,0 +1,132 @@
...
+    public class ClientSuppressionSynchronizerTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<ClientSuppressionSynchronizer, IClientSuppressionSynchronizer>(
+                MefTestHelpers.CreateExport<IClientIssueStore>(),
+                MefTestHelpers.CreateExport<IIssuesFilter>());
+        }
+
```
bigfluffycookie, line 41: The three tests are quite similar to be honest, the reason why i made three is to first check only if an unsuppressed issue gets suppressed, the second test is for the opposite and then the third is just for checking multiple issues (However I could just remove the first 2 tests)

src/ConnectedMode/Suppressions/ClientSuppressionSynchronizer.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            var clientIssues = clientIssueStore.Get().ToArray();
+            var filterableIssues = clientIssues.OfType<IFilterableIssue>().ToArray();
```
bigfluffycookie, line 47: I just saw this now, I am using the client issues instead of the filterable issues in .GetMatches, I'll fix it along with other issues.

duncanp-sonar, line 47: You could combine line 46 and 47 to avoid using "ToArray" twice i.e. so we only build one fixed list  e.g.

```
    var filterableIssues = clientIssueStore.Get()
               .OfType<IFilterableIssue>()
               .ToArray();
  
```

src/ConnectedMode.UnitTests/Suppressions/ClientSuppressionSynchronizerTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        [TestMethod]
+        public void SynchronizeIssues_IssueGetsSuppressedOnServer_LocalIssueGetUpdated()
+        {
+            var issue = CreateIssue(isSuppressedLocally: false);
+
+            var localIssues = new[] { issue };
```
duncanp-sonar, line 47: Minor: you've used `params` in the `CreateClientIssueStore` method so you don't need to explicitly create the `matches` array - you could just say:

```
  var issueFilter = CreateIssuesFilter(localIssues, issue);
```

src/ConnectedMode.UnitTests/Suppressions/ClientSuppressionSynchronizerTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        private static IClientIssueStore CreateClientIssueStore(params IAnalysisIssueVisualization[] localIssues)
+        {
+            var clientIssueStore = new Mock<IClientIssueStore>();
+
+            clientIssueStore.Setup(x => x.Get()).Returns(localIssues);
+
+            return clientIssueStore.Object;
+        }
+
+        private static IIssuesFilter CreateIssuesFilter(IEnumerable<IAnalysisIssueVisualization> localIssues, IEnumerable<IFilterableIssue> matches = null)
```
duncanp-sonar, line 114: Did you decide not to use `params` here too?

(there is as slight difference: if you don't pass anything to a params variable it will be set to an empty array rather than null, but that may not matter here).

src/ConnectedMode.UnitTests/Suppressions/ClientSuppressionSynchronizerTests.cs
duncanp-sonar: Difficult to know, really - there are pros and cons for both. You definitely need the "multiple" test, and as you say it does the work for the other two.

I'd be inclined to keep them all, although you could combine the first two into a data-driven test:

```
[TestMethod]
[DataRow(true)]
[DataRow(false)]
public void SychronizeIssues_SingleIssue_IsSuppressedOnServer(bool suppressedOnServer)
{
...
}
```

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithoutStrongNames.txt
```diff
@@ -171,11 +171,12 @@ Referenced assemblies:
 - 'Microsoft.VisualStudio.Threading, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
 - 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
 - 'SonarLint.VisualStudio.Core, Version=6.14.0.0, Culture=neutral, PublicKeyToken=null'
+- 'SonarLint.VisualStudio.IssueVisualization, Version=6.14.0.0, Culture=neutral, PublicKeyToken=null'
```
duncanp-sonar, line 174: The project file hasn't changed in this PR, but the list of referenced assemblies *has* changed.

I'm guessing this is because the previously we we're actually using anything from IssueViz, so the compiled assembly didn't actual need to reference it before.

src/Core/Suppressions/IClientSuppressionSynchronizer.cs
```diff
@@ -0,0 +1,30 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Core.Suppressions
+{
+    public interface IClientSuppressionSynchronizer
```
duncanp-sonar, line 23: I'm hoping that we'll be able to move the suppressions-related stuff out of `Core` later and into `ConnectedMode` (I'll add a note to the board).
However, it makes sense to put it with the other suppression-related interfaces now.

src/ConnectedMode.UnitTests/Suppressions/ClientSuppressionSynchronizerTests.cs
bigfluffycookie: But I can only use params once no? I have 2 lists, that's also why I dont have params for the other method as I have to make the list regardless or maybe I am misunderstanding this

src/Core/Suppressions/IClientSuppressionSynchronizer.cs
bigfluffycookie: Confused about this statement, Didnt we want this in core?

src/ConnectedMode.UnitTests/Suppressions/ClientSuppressionSynchronizerTests.cs
duncanp-sonar: Ah, yes; you could make one the arguments `params`, which would make passing one of the list simpler. But not both.

src/Core/Suppressions/IClientSuppressionSynchronizer.cs
duncanp-sonar: Ideally, all feature-specific stuff would be in feature-specific assemblies, unless that gives us problems with having circular dependencies between them, in which case we'd need to put the interfaces in a "neutral" assembly like Core.
However, it's not a problem for now.

duncanp-sonar (approved):

Issue #3819, opened 2023-03-02
Create SuppressionChangedHandler


duncanp-sonar:
### Description

Responsibility: listen to suppression changed events and trigger issue synchronization

TODO: something needs to create it. The Connected Mode package?

PR #3824, opened 2023-03-03
Add ServerSuppressionsChangedHandler

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3819


src/ConnectedMode/Suppressions/ServerSuppressionsChangedHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        private void ServerSuppressionsChanged(object sender, EventArgs e)
+        {
+            clientSuppressionSynchronizer.SynchronizeSuppressedIssues();
+        }
+
+        public void Dispose()
+        {
+            suppressedIssuesMonitor.SuppressionsUpdateRequested -= ServerSuppressionsChanged;
+        }
+
```
bigfluffycookie, line 52: Missed this, will fix it with other pr errors

src/ConnectedMode/Suppressions/ServerSuppressionsChangedHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            this.clientSuppressionSynchronizer = clientSuppressionSynchronizer;
+            this.suppressedIssuesMonitor = suppressedIssuesMonitor;
+
+            suppressedIssuesMonitor.SuppressionsUpdateRequested += ServerSuppressionsChanged;
+        }
+
+        private void ServerSuppressionsChanged(object sender, EventArgs e)
+        {
+            clientSuppressionSynchronizer.SynchronizeSuppressedIssues();
```
bigfluffycookie, line 45: Note here to maybe add logging

duncanp-sonar:
@bigfluffycookie, just to check - this PR is just about the last commit?

bigfluffycookie:
@duncanp-sonar 
Yes, the other two commits are from the automatic branch change.

src/ConnectedMode/Suppressions/ServerSuppressionsChangedHandler.cs
duncanp-sonar: Either here, or in the synchronizer (or perhaps both).
We need to decide what information would be useful in normal operation (e.g. "_Re-apply server-side suppressions_") and what would help in troubleshooting/support cases (e.g. "x issues suppressed/updated").
I'll add a task to the board.

PR #3828, opened 2023-03-06
Move server sent event interfaces to subfolders

duncanp-sonar:


duncanp-sonar:
I moved the code and related tests. The only changes are due to changes in namespace.

Issue #3833, opened 2023-03-06
Implement suppression store updater "fetch all"


duncanp-sonar:
### Description

Implement `ISuppressionIssueStoreUpdater`.

TODO
-[ ] implement simple "fetch all method"

Note: the "fetch from timestamp" method depends on #3830

PR #3841, opened 2023-03-07
Add helper class to simplify making branch-aware web calls

duncanp-sonar:
Prep for #3833

### DIFF
--- a/src/ConnectedMode.UnitTests/ServerQueryInfoProviderTests.cs
+++ b/src/ConnectedMode.UnitTests/ServerQueryInfoProviderTests.cs
@@ -40,7 +40,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests
 
         [TestMethod]
         [DataRow(SonarLintMode.Connected)]
-        [DataRow(SonarLintMode.Connected)]
+        [DataRow(SonarLintMode.LegacyConnected)]
         public async Task Get_IsConnectedMode_ReturnsExpectedValues(SonarLintMode mode)
         {
             // Happy path - in connected mode, has branch
--- a/src/ConnectedMode.UnitTests/ServerQueryInfoProviderTests.cs
+++ b/src/ConnectedMode.UnitTests/ServerQueryInfoProviderTests.cs
@@ -1,0 +1,135 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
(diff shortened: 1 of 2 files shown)


src/ConnectedMode/ServerQueryInfoProvider.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// <remarks>Facade to simplify classes that need to make branch-aware calls.
+    /// The projectKey and server branch are provided by different interfaces. This interface
+    /// simplifies getting the information.</remarks>
+    internal interface IServerQueryInfoProvider
+    {
+        /// <summary>
+        /// Returns the projectKey and branch name for the current bound solution, or
+        /// (null, null) if the solution is not bound
+        /// </summary>
+        Task<(string projectKey, string branchName)> GetProjectKeyAndBranchAsync(CancellationToken token);
```
duncanp-sonar, line 41: The web calls to fetch the suppressions need to know the current project key and branch name, and need to handle the details not being available.

Originally, I'd put this functionality in the `SuppressionIssueStoreUpdater` class but it made it more complicated than necessary, so I broke it out into a separate class. It's likely that we could re-use it to simpligy other classes that call the server.

src/ConnectedMode.UnitTests/ServerQueryInfoProviderTests.cs
```diff
@@ -0,0 +1,135 @@
...
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<ServerQueryInfoProvider, IServerQueryInfoProvider>(
+                MefTestHelpers.CreateExport<IConfigurationProvider>(),
+                MefTestHelpers.CreateExport<IStatefulServerBranchProvider>());
+        }
+
+        [TestMethod]
+        [DataRow(SonarLintMode.Connected)]
+        [DataRow(SonarLintMode.Connected)]
```
bigfluffycookie, line 43: Is this a mistake or is there some reason to run the same thing twice

bigfluffycookie (approved):
Looks good, just one question regarding a test

src/ConnectedMode.UnitTests/ServerQueryInfoProviderTests.cs
duncanp-sonar: Oops - mistake. Thanks!

Issue #3832, opened 2023-03-06
Implement IServerIssueStoreWriter


duncanp-sonar:
### Description

Implement the write side of the store interfaces.

TODO
-[ ] implement methods
-[ ] tests]

PR #3842, opened 2023-03-07
Implement ServerIssueStoreWriter

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3832

### DIFF
--- a/src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
+++ b/src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
@@ -18,8 +18,12 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using SonarLint.VisualStudio.ConnectedMode.Suppressions;
 using SonarLint.VisualStudio.TestInfrastructure;
+using SonarQube.Client.Models;
 
 namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
 {
@@ -30,6 +34,84 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
         public void MefCtor_CheckIsExported()
         {
             MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStore>();
+            MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStoreWriter>();
+        }
+
+        [TestMethod]
+        public void AddIssues_ListOfIssuesAsExpected()
+        {
+            var issue1 = CreateIssue("1", false);
+            var issue2 = CreateIssue("2", false);
+            var issue3 = CreateIssue("3", false);
+
+            var testSubject = new ServerIssuesStore();
+
+            var result = testSubject.Get();
+            result.Count().Should().Be(0);
+
+            testSubject.AddIssues(new List<SonarQubeIssue>() { issue1, issue2 }, clearAllExistingIssues: false);
+
+            result = testSubject.Get();
+            result.Count().Should().Be(2);
+            result.Should().Contain(issue1);
+            result.Should().Contain(issue2);
+
+            testSubject.AddIssues(new List<SonarQubeIssue>() { issue3 }, clearAllExistingIssues: true);
+
+            result = testSubject.Get();
+            result.Count().Should().Be(1);
+            result.Should().Contain(issue3);
+        }
+
+        [TestMethod]
+        public void UpdateIssues_ListOfIssuesAsExpected()
+        {
+            var issue1 = CreateIssue("issue1", true);
+            var issue2 = CreateIssue("issue2", false);
+
+            var testSubject = new ServerIssuesStore();
+
+            testSubject.AddIssues(new List<SonarQubeIssue>() { issue1, issue2 }, clearAllExistingIssues: false);
+
+            var result = testSubject.Get();
+            result.Count().Should().Be(2);
+            result.Should().Contain(issue1);
+            result.Should().Contain(issue2);
+            issue1.IsResolved.Should().BeTrue();
+            issue2.IsResolved.Should().BeFalse();
+
+            testSubject.UpdateIssue("issue1", false);
+            issue1.IsResolved.Should().BeFalse();
+
+            testSubject.UpdateIssue("issue2", true);
+            issue2.IsResolved.Should().BeTrue();
+
+            // This test is to insure it doesn't just flip the state.
+            testSubject.UpdateIssue("issue2", true);
+            issue2.IsResolved.Should().BeTrue();
+        }
+
+        [TestMethod]
+        public void AddOrUpdateIssues_EventIsInvoked()
+        {
+            var issue1 = CreateIssue("issue1", true);
+
+            var testSubject = new ServerIssuesStore();
+            var eventMock = new Mock<EventHandler>();
+            testSubject.ServerIssuesChanged += eventMock.Object;
+
+            testSubject.AddIssues(new List<SonarQubeIssue>() { issue1 }, clearAllExistingIssues: false);
+            eventMock.Verify(x => x(testSubject, EventArgs.Empty), Times.Once);
+
+            testSubject.UpdateIssue("issue1", false);
+            eventMock.Verify(x => x(testSubject, EventArgs.Empty), Times.Exactly(2));
+        }
+
+        private static SonarQubeIssue CreateIssue(string key, bool isResolved)
+        {
+            var issue = new SonarQubeIssue(key, "", "", "", "", "", isResolved, SonarQubeIssueSeverity.Info, DateTimeOffset.MinValue, DateTimeOffset.MinValue, null, null);
+
+            return issue;
         }
     }
 }
--- a/src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
+++ b/src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
@@ -121,7 +121,9 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
         }
 
         [TestMethod]
-        public void AddIssues_EventIsInvoked()
+        [DataRow(true)]
+        [DataRow(false)]
+        public void AddIssues_EventIsInvoked(bool clearAllExistingIssues)
         {
             var issue1 = CreateIssue("issue1", true);
 
@@ -129,7 +131,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
             var eventMock = new Mock<EventHandler>();
             testSubject.ServerIssuesChanged += eventMock.Object;
 
-            testSubject.AddIssues(new List<SonarQubeIssue>() { issue1 }, clearAllExistingIssues: false);
+            testSubject.AddIssues(new List<SonarQubeIssue>() { issue1 }, clearAllExistingIssues);
             eventMock.Verify(x => x(testSubject, EventArgs.Empty), Times.Once);
         }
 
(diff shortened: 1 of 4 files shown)


src/ConnectedMode/Suppressions/IServerIssuesStore.cs
```diff
@@ -45,7 +45,7 @@ internal interface IServerIssuesStore
     /// <summary>
     /// Write-interface for <see cref="IServerIssuesStore"/> i.e. contains methods to update the store
     /// </summary>
-    internal interface IServerIssueStoreWriter
+    internal interface IServerIssuesStoreWriter
```
bigfluffycookie, line 48: Missed earlier that is was "Issue" instead of "Issue**s**"

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
```diff
@@ -19,24 +19,59 @@
...
             return serverIssues ?? Enumerable.Empty<SonarQubeIssue>();
         }
+
+        #endregion IServerIssuesStore implementation
+
+        #region IServerIssueStoreWriter implementation
+
+        public void AddIssues(IEnumerable<SonarQubeIssue> issues, bool clearAllExistingIssues)
+        {
+            if (clearAllExistingIssues || serverIssues == null)
```
bigfluffycookie, line 51: Not a massive fan of the serverIssues == null here however I wasn't sure where else to initialize it.

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
```diff
@@ -19,24 +19,59 @@
...
     [PartCreationPolicy(CreationPolicy.Shared)]
-    internal class ServerIssuesStore : IServerIssuesStore
+    internal class ServerIssuesStore : IServerIssuesStore, IServerIssuesStoreWriter
     {
-        private IEnumerable<SonarQubeIssue> serverIssues;
+        private List<SonarQubeIssue> serverIssues;
 
         public event EventHandler ServerIssuesChanged;
 
+        #region IServerIssuesStore implementation
```
bigfluffycookie, line 38: I added the #region thing because it made sense to me this way its easier to know which part is which but I can also remove it if you think thats better

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
```diff
@@ -30,6 +34,84 @@ public class ServerIssuesStoreTests
...
+            testSubject.ServerIssuesChanged += eventMock.Object;
+
+            testSubject.AddIssues(new List<SonarQubeIssue>() { issue1 }, clearAllExistingIssues: false);
+            eventMock.Verify(x => x(testSubject, EventArgs.Empty), Times.Once);
+
+            testSubject.UpdateIssue("issue1", false);
+            eventMock.Verify(x => x(testSubject, EventArgs.Empty), Times.Exactly(2));
+        }
+
+        private static SonarQubeIssue CreateIssue(string key, bool isResolved)
```
bigfluffycookie, line 110: Is there a reason why we dont have an interface to mock this?

src/SonarQube.Client/Models/SonarQubeIssue.cs
```diff
@@ -59,7 +59,7 @@ public class SonarQubeIssue
         public string Message { get; }
         public string ModuleKey { get; }
         public string RuleId { get; }
-        public bool IsResolved { get; }
+        public bool IsResolved { get; set; }
```
duncanp-sonar, line 62: Probably worth commenting that this property needs to be mutable since it can be changed later by  SLVS.

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
duncanp-sonar: I use `#region` in the same way, to separate interfaces implementations. An alternative is to make it a `partial` class and put the implementations in different files.

Rita wasn't a big fan of regions because she had seen them being used badly previously.

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
```diff
@@ -19,24 +19,59 @@
...
         public IEnumerable<SonarQubeIssue> Get()
         {
             return serverIssues ?? Enumerable.Empty<SonarQubeIssue>();
         }
+
+        #endregion IServerIssuesStore implementation
+
+        #region IServerIssueStoreWriter implementation
+
+        public void AddIssues(IEnumerable<SonarQubeIssue> issues, bool clearAllExistingIssues)
```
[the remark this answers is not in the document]
duncanp-sonar: Yes, I think we do need to handle duplicates, unfortunately.
If two issues have the same `issueKey` we should consider them to be them same. We might need to implement a custom `IEqualityComparer<SonarQubeIssue>`, which we can then use to handle the differences.

I suggest handling duplicates in a separate PR. Let me know if you'd like to discuss implementation options.

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
duncanp-sonar: You could have initialised as part of the definition(
```
private List<SonarQubeIssue> serverIssues = new List<SonarQubeIssue>();
```

That would also have allowed you to skip the `null` handling in the `Get` method.
However, it doesn't make too much difference.

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
```diff
@@ -19,24 +19,59 @@
...
+        {
+            if (clearAllExistingIssues || serverIssues == null)
+            {
+                serverIssues = new List<SonarQubeIssue>();
+            }
+
+            serverIssues.AddRange(issues);
```
duncanp-sonar, line 56: Does this work if `issues` is `null`?

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
duncanp-sonar: Not really. It's a data-only class so there was less need to i.e. no functionality to mock, so I guess we never got round to it.

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
```diff
@@ -30,6 +34,88 @@ public class ServerIssuesStoreTests
...
+            testSubject.UpdateIssue("issue2", true);
+            issue2.IsResolved.Should().BeTrue();
+
+            // This test is to insure it doesn't just flip the state.
+            testSubject.UpdateIssue("issue2", true);
+            issue2.IsResolved.Should().BeTrue();
+        }
+
+        [TestMethod]
+        public void AddOrUpdateIssues_EventIsInvoked()
```
duncanp-sonar, line 99: Missing test: don't invoke the event if the issue key didn't match.

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
```diff
@@ -30,6 +34,88 @@ public class ServerIssuesStoreTests
         public void MefCtor_CheckIsExported()
         {
             MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStore>();
+            MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStoreWriter>();
+        }
+
+        [TestMethod]
+        public void AddIssues_ListOfIssuesAsExpected()
```
duncanp-sonar, line 41: I would probably split these into separate tests:
- dat-driven test: add to empty list, replace=true/false
- add to non-empty list, replace=true -> old issues replaced etc.
- add to non-empty list, replace=false -> old issues retained

I think you've covered all these cases, but it's a bit hard to tell with them all in the same test.

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
```diff
@@ -30,6 +34,88 @@ public class ServerIssuesStoreTests
...
+
+            testSubject.AddIssues(new List<SonarQubeIssue>() { issue3 }, clearAllExistingIssues: true);
+
+            result = testSubject.Get();
+            result.Count().Should().Be(1);
+            result.Should().Contain(issue3);
+        }
+
+        [TestMethod]
+        public void UpdateIssues_ListOfIssuesAsExpected()
```
duncanp-sonar, line 67: I'd split this too:
- empty list -> no-op, no event
- non-empty list, no match -> no event
- non-empty list, match -> event, and property changed (if you were being really strict you'd also check that the non-matching items hadn't been changed, only the matching item)

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
duncanp-sonar: You could add an "event raised" check to all of these tests too, then you wouldn't need the separate `AddOrUpdateIssues_EventIsInvoked` test below.

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
duncanp-sonar: (...which could be added in one of the new separation of tests I suggested above).

duncanp-sonar (commented):
Looks good. I'd split the tests to make it easier to see what is covered.

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
bigfluffycookie: no

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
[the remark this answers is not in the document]
bigfluffycookie: Does this mean I can assume in the UpdateIssue method that there will never be any duplicates?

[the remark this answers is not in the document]
duncanp-sonar: Yes, I think so. As it is written at the moment, it will throw if there are duplicates since you are using `SingleOrDefault` (`FirstOrDefault` doesn't throw if there are multiples).

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
```diff
@@ -30,6 +34,165 @@ public class ServerIssuesStoreTests
...
+            var testSubject = new ServerIssuesStore();
+            var eventMock = new Mock<EventHandler>();
+            testSubject.ServerIssuesChanged += eventMock.Object;
+
+            testSubject.AddIssues(new List<SonarQubeIssue>() { issue1 }, clearAllExistingIssues: false);
+            eventMock.Verify(x => x(testSubject, EventArgs.Empty), Times.Once);
+        }
+
+        [TestMethod]
+        public void UpdateIssues_EmptyList_DoesNotInvokeEvent()
```
duncanp-sonar, line 137: Nice.

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
```diff
@@ -30,6 +34,165 @@ public class ServerIssuesStoreTests
...
+            var eventMock = new Mock<EventHandler>();
+            testSubject.ServerIssuesChanged += eventMock.Object;
+
+            testSubject.AddIssues(null, false);
+
+            eventMock.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public void AddIssues_EventIsInvoked()
```
duncanp-sonar, line 124: Being really picky here - logically this test only checks that no event is raised if `clearAllExistingIssues` is `false`. It would be easy to make this a data-driven test that covered both cases.

duncanp-sonar (approved):
Nice. If you are feeling picky, you could make `AddIssues_EventIsInvoked` data-driven to test both `true` and `false`, but it's not essential.

PR #3843, opened 2023-03-07
Add conversion from RuleInfo to logical pages models

georgii-borovinskikh-sonarsource:


src/Education/Layout/Logical/IRuleInfoTranslator.cs
```diff
@@ -0,0 +1,86 @@
...
+        public IEnumerable<IRichRuleDescriptionSection> GetRuleDescriptionSections(IRuleInfo ruleInfo)
+        {
+
+            var sectionsByKey = ruleInfo.DescriptionSections
+                .GroupBy(x => x.Key)
+                .ToDictionary(x => x.Key, x => x.ToList());
+
+            if (sectionsByKey.TryGetValue(RootCauseSection.RuleInfoKey, out var rootCauseSectionContent))
+            {
+                yield return new RootCauseSection(GeneratePartialXaml(rootCauseSectionContent[0].HtmlContent), /*todo in future PR*/false);
```
ugras-ergun-sonarsource, line 51: We're expecting to get only 1 result here, maybe add some asserts for that.

ugras-ergun-sonarsource: also for the rest of the file where applicable

src/Education/Layout/Logical/HowToFixItSection.cs
```diff
@@ -0,0 +1,69 @@
...
+        public string Title => "How can I fix it?";
+
+        public IAbstractVisualizationTreeNode GetVisualizationTreeNode(ITabsRepository tabsRepository)
+        {
+            throw new NotImplementedException();
+            // todo in the next PR
+
+        }
+
+        public class Context
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Can't make it private. Why do you suggest moving it?

src/Education.UnitTests/Layout/RichRuleDescriptionSectionsTests.cs
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.Education.Layout.Logical;
+
+namespace SonarLint.VisualStudio.Education.UnitTests.Layout
+{
+    [TestClass]
+    public class RichRuleDescriptionSectionsTests
```
ugras-ergun-sonarsource, line 28: Maybe we should separate these tests. They're testing different classes. And when we implement `GetVisualizationTreeNode` we will add more tests and it will be doing too many things.

georgii-borovinskikh-sonarsource: Agree. Will fix.

src/Education/Layout/Logical/HowToFixItSection.cs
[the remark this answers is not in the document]
ugras-ergun-sonarsource: if you need to access it outside of the class it shouldn't be inside the class. If they're tightly coupled you can put it in the same file but as a separate class.

src/Education/Layout/Logical/IRuleInfoTranslator.cs
```diff
@@ -0,0 +1,86 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using SonarLint.VisualStudio.Education.XamlGenerator;
+using System.Linq;
+using SonarLint.VisualStudio.Rules;
+
+namespace SonarLint.VisualStudio.Education.Layout.Logical
```
ugras-ergun-sonarsource, line 26: As our naming convention, filename should be RuleInfoTranslator.cs

georgii-borovinskikh-sonarsource: Will fix

src/Education/Layout/Logical/HowToFixItSection.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: >  it shouldn't be inside the class
Why not? I can move it no problem, but it's interesting for me to figure out why exactly.

src/Education/Layout/Logical/IRuleInfoTranslator.cs
georgii-borovinskikh-sonarsource: The idea was to not fail when we encounter that and display something. Should have added a test for that(+ logging maybe). Will fix.

ugras-ergun-sonarsource (approved):

PR #3844, opened 2023-03-07
Implement fetch all suppressed issues

### LINKED ISSUE #3833: Implement suppression store updater "fetch all"

duncanp-sonar:
Fixes #3833


src/ConnectedMode/Suppressions/SuppressionIssueStoreUpdater.cs
```diff
@@ -30,16 +37,121 @@ internal interface ISuppressionIssueStoreUpdater
         /// <summary>
         /// Fetches all available suppressions from the server and updates the server issues store
         /// </summary>
-        void FetchAllServerSuppressions();
+        Task UpdateAllServerSuppressionsAsync();
```
duncanp-sonar, line 40: I changed the methods to be `async` - we're making network calls, so they should be `async`.

src/ConnectedMode/Suppressions/SuppressionIssueStoreUpdater.cs
```diff
@@ -30,16 +37,121 @@ internal interface ISuppressionIssueStoreUpdater
...
         /// </summary>
         void Clear();
     }
+
+    [Export(typeof(ISuppressionIssueStoreUpdater))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class SuppressionIssueStoreUpdater : ISuppressionIssueStoreUpdater, IDisposable
+    {
+        private readonly ISonarQubeService server;
+        private readonly IServerQueryInfoProvider serverQueryInfoProvider;
```
duncanp-sonar, line 58: Using the new helper class I added in the previous PR to make life easier.

src/ConnectedMode/Suppressions/SuppressionIssueStoreUpdater.cs
```diff
@@ -30,16 +37,121 @@ internal interface ISuppressionIssueStoreUpdater
...
 
         /// <summary>
         /// Clears all issues from the store
         /// </summary>
         void Clear();
     }
+
+    [Export(typeof(ISuppressionIssueStoreUpdater))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class SuppressionIssueStoreUpdater : ISuppressionIssueStoreUpdater, IDisposable
```
duncanp-sonar, line 55: This class contains a subset of the functionality that was/is in `SonarQubeIssueProvider` i.e. fetching all suppressions from the server.

I've dropped a couple of things:
* all of the module-related stuff, since I don't think we need it any more
* all of the timer-related code, since that will be handled by a different class

src/ConnectedMode/Suppressions/SuppressionIssueStoreUpdater.cs
```diff
@@ -30,16 +37,121 @@ internal interface ISuppressionIssueStoreUpdater
...
+
+        public async Task UpdateAllServerSuppressionsAsync()
+        {
+            await threadHandling.SwitchToBackgroundThread();
+
+            try
+            {
+                logger.WriteLine(Resources.Suppressions_Fetch_AllIssues);
+
+                CancelCurrentOperation();
```
duncanp-sonar, line 98: The old class had this logic - we only want to have one "fetch all" operation running, so any time we start one, cancel the previous one if there is one.

However, I don't think the logic was implemented correctly before, and I couldn't find any tests for it. This version is at least tested.

src/ConnectedMode/Suppressions/SuppressionIssueStoreUpdater.cs
```diff
@@ -30,16 +37,121 @@ internal interface ISuppressionIssueStoreUpdater
...
+                localCopyOfCts.Token.ThrowIfCancellationRequested();
+
+                var allSuppressedIssues = await server.GetSuppressedIssuesAsync(queryInfo.projectKey, queryInfo.serverBranch, localCopyOfCts.Token);
+                storeWriter.AddIssues(allSuppressedIssues, clearAllExistingIssues: true);
+
+                logger.WriteLine(Resources.Suppression_Fetch_AllIssues_Finished);
+            }
+            catch (OperationCanceledException)
+            {
+                logger.WriteLine(Resources.Suppressions_FetchOperationCancelled);
```
duncanp-sonar, line 119: Logging a specific message in case of cancellation since it's expected (or at least not completely unexpected).

src/ConnectedMode/Suppressions/SuppressionIssueStoreUpdater.cs
```diff
@@ -30,16 +37,121 @@ internal interface ISuppressionIssueStoreUpdater
...
+            }
+            catch(Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+            {
+                logger.LogVerbose(Resources.Suppression_FetchError_Verbose, ex);
+                logger.WriteLine(Resources.Suppressions_FetchError_Short, ex.Message);
```
duncanp-sonar, line 124: I think I've done this in a few places more recently - log a short error message that will appear at normal verbosity, and a second verbose message with all of the detail to help with support/debugging.

src/ConnectedMode.UnitTests/Suppressions/SuppressionIssueStoreUpdaterTests.cs
```diff
@@ -0,0 +1,295 @@
...
+            writer.Invocations.Should().HaveCount(1);
+        }
+
+        [TestMethod]
+        public async Task UpdateAll_RunOnBackgroundThread()
+        {
+            var callSequence = new List<string>();
+
+            var queryInfo = new Mock<IServerQueryInfoProvider>();
+            var threadHandling = new Mock<IThreadHandling>();
```
duncanp-sonar, line 99: A more robust check would check that all of the services are called in the expected order, but this is enough for now.

src/ConnectedMode.UnitTests/Suppressions/SuppressionIssueStoreUpdaterTests.cs
```diff
@@ -0,0 +1,295 @@
...
+            var testSubject = CreateTestSubject(queryInfo.Object, logger: logger);
+
+            Func<Task> operation = testSubject.UpdateAllServerSuppressionsAsync;
+            operation.Should().NotThrow<InvalidOperationException>();
+
+            logger.AssertPartialOutputStringExists("thrown in a test");
+        }
+
+        [TestMethod]
+        public void UpdateAll_OperationCancelledException_CancellationMessageLogged()
```
duncanp-sonar, line 154: Tests that we get the simpler "operation cancelled" message if a call is cancelled by a second call to "fetch all".

src/ConnectedMode.UnitTests/Suppressions/SuppressionIssueStoreUpdaterTests.cs
```diff
@@ -0,0 +1,295 @@
...
+
+            Func<Task> operation = testSubject.UpdateAllServerSuppressionsAsync;
+            operation.Should().NotThrow<OperationCanceledException>();
+
+            logger.AssertPartialOutputStringDoesNotExist("thrown in a test");
+            logger.AssertOutputStringExists(Resources.Suppressions_FetchOperationCancelled);
+        }
+
+        [TestMethod]
+        public async Task UpdateAll_CallInProgress_CallIsCancelled()
```
duncanp-sonar, line 172: This test is complicated, but it's the only way to force the operations to occur in the expected order i.e. 

1. first call starts
2. second call starts and cancels first call
3. first call notices it has been cancelled and aborts
4. second call finishes.

To make sure the operations happen in a specific order we need to use `ManualResetEvent`s to block the execution at various points.

src/ConnectedMode/Suppressions/SuppressionIssueStoreUpdater.cs
```diff
@@ -30,16 +37,121 @@ internal interface ISuppressionIssueStoreUpdater
...
+        public async Task UpdateAllServerSuppressionsAsync()
+        {
+            await threadHandling.SwitchToBackgroundThread();
+
+            try
+            {
+                logger.WriteLine(Resources.Suppressions_Fetch_AllIssues);
+
+                CancelCurrentOperation();
+                var localCopyOfCts = updateAllCancellationTokenSource = new CancellationTokenSource();
```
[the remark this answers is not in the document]
duncanp-sonar: Actually, I don't this version is completely correct either. However, it's better than before.
If I can come up with a better solution I'll do it in a separate PR.

src/ConnectedMode/Suppressions/SuppressionIssueStoreUpdater.cs
bigfluffycookie: Yea I think that makes sense

src/ConnectedMode/Suppressions/SuppressionIssueStoreUpdater.cs
```diff
@@ -30,16 +37,121 @@ internal interface ISuppressionIssueStoreUpdater
...
+        }
+
+        public void Clear()
+        {
+            throw new NotImplementedException();
+        }
+
+        private void CancelCurrentOperation()
+        {
+            // We don't want multiple "fetch all" operations running at once (e.g. if the
```
bigfluffycookie, line 140: nit pick but I think you missed the word "user" in the sentence -> (e.g. if the opens a solution then clicks "update").

src/ConnectedMode.UnitTests/Suppressions/SuppressionIssueStoreUpdaterTests.cs
bigfluffycookie: Ok yes that is a long test but the way its written and commented on its understandable so Its okay

src/ConnectedMode.UnitTests/Suppressions/SuppressionIssueStoreUpdaterTests.cs
```diff
@@ -0,0 +1,295 @@
...
+            // If cancellation worked then we should only have made one call to each
+            // of the SonarQubeService and store writer
+            server.Invocations.Should().HaveCount(1);
+            writer.Invocations.Should().HaveCount(1);
+
+            void Log(string message)
+            {
+                Console.WriteLine($"[Thread {System.Threading.Thread.CurrentThread.ManagedThreadId}] {message}");
+            }
+
```
bigfluffycookie, line 250: nitpick, empty line

bigfluffycookie (approved):
Good except for tiny tiny comments

duncanp-sonar:
Replaced by #3848 - same commits, with two more to fix merge errors.

Issue #3846, opened 2023-03-07
Implement bound solution changed update handler


duncanp-sonar:
### Description

Implement a subset of the behaviour of the old `SuppressedIssuesProvider` class

i.e.  when the bound solution is changed or updated, trigger an update the list of suppressed issues.

PR #3854, opened 2023-03-10
Add BoundSolutionUpdateHandler

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3846

### DIFF
--- a/src/ConnectedMode.UnitTests/BoundSolutionUpdateHandlerTests.cs
+++ b/src/ConnectedMode.UnitTests/BoundSolutionUpdateHandlerTests.cs
@@ -36,7 +36,6 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests
                 MefTestHelpers.CreateExport<ISuppressionIssueStoreUpdater>());
         }
 
-
         [TestMethod]
         public void Ctor_SubscribesToEvents()
         {
@@ -57,10 +56,10 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests
             _ = new BoundSolutionUpdateHandler(activeSolutionTracker.Object, suppressionIssueStoreUpdater.Object);
 
             activeSolutionTracker.Raise(x => x.SolutionBindingChanged += null, new ActiveSolutionBindingEventArgs(BindingConfiguration.Standalone));
-            suppressionIssueStoreUpdater.Verify(x => x.FetchAllServerSuppressions(), Times.Once);
+            suppressionIssueStoreUpdater.Verify(x => x.UpdateAllServerSuppressionsAsync(), Times.Once);
 
             activeSolutionTracker.Raise(x => x.SolutionBindingUpdated += null, EventArgs.Empty);
-            suppressionIssueStoreUpdater.Verify(x => x.FetchAllServerSuppressions(), Times.Exactly(2));
+            suppressionIssueStoreUpdater.Verify(x => x.UpdateAllServerSuppressionsAsync(), Times.Exactly(2));
         }
 
         [TestMethod]
--- a/src/ConnectedMode.UnitTests/BoundSolutionUpdateHandlerTests.cs
+++ b/src/ConnectedMode.UnitTests/BoundSolutionUpdateHandlerTests.cs
@@ -1,0 +1,79 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using SonarLint.VisualStudio.ConnectedMode.Suppressions;
+using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.TestInfrastructure;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests
+{
+    [TestClass]
+    public class BoundSolutionUpdateHandlerTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<BoundSolutionUpdateHandler, BoundSolutionUpdateHandler>(
+                MefTestHelpers.CreateExport<IActiveSolutionBoundTracker>(),
+                MefTestHelpers.CreateExport<ISuppressionIssueStoreUpdater>());
+        }
+
+
+        [TestMethod]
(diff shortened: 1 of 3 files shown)


src/ConnectedMode.UnitTests/BoundSolutionUpdateHandlerTests.cs
```diff
@@ -0,0 +1,79 @@
...
+    public class BoundSolutionUpdateHandlerTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<BoundSolutionUpdateHandler, BoundSolutionUpdateHandler>(
+                MefTestHelpers.CreateExport<IActiveSolutionBoundTracker>(),
+                MefTestHelpers.CreateExport<ISuppressionIssueStoreUpdater>());
+        }
+
```
bigfluffycookie, line 38: Will remove empty line with other requested changes.

src/ConnectedMode.UnitTests/BoundSolutionUpdateHandlerTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            activeSolutionTracker.Raise(x => x.SolutionBindingChanged += null, new ActiveSolutionBindingEventArgs(BindingConfiguration.Standalone));
+            suppressionIssueStoreUpdater.Verify(x => x.UpdateAllServerSuppressionsAsync(), Times.Once);
+
+            activeSolutionTracker.Raise(x => x.SolutionBindingUpdated += null, EventArgs.Empty);
+            suppressionIssueStoreUpdater.Verify(x => x.UpdateAllServerSuppressionsAsync(), Times.Exactly(2));
+        }
+
+        [TestMethod]
+        public void Dispose_UnsubscribesToEvent()
```
bigfluffycookie, line 66: Wouldit be a thing to test that if I raise an event after unsubscribing it doesnt call my method anymore? 

ie. 

Dispose()
activeSolutionTracker.BindingChanged.Raise()
suppressionIssueStoreUpdater.Verify((x => x.UpdateAllServerSuppressionsAsync(), Times.Never)

src/ConnectedMode.UnitTests/BoundSolutionUpdateHandlerTests.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        [TestMethod]
+        public void Ctor_SubscribesToEvents()
+        {
+            var activeSolutionTracker = new Mock<IActiveSolutionBoundTracker>();
+
+            _ = new BoundSolutionUpdateHandler(activeSolutionTracker.Object, Mock.Of<ISuppressionIssueStoreUpdater>());
+
+            activeSolutionTracker.VerifyAdd(x => x.SolutionBindingChanged += It.IsAny<EventHandler<ActiveSolutionBindingEventArgs>>(), Times.Once());
+            activeSolutionTracker.VerifyAdd(x => x.SolutionBindingUpdated += It.IsAny<EventHandler>(), Times.Once());
```
bigfluffycookie, line 47: I'll remove the parentheses from Times.Once() with other changes.

src/ConnectedMode.UnitTests/BoundSolutionUpdateHandlerTests.cs
```diff
@@ -0,0 +1,78 @@
...
+        public void Dispose_UnsubscribesToEvent()
+        {
+            var activeSolutionTracker = new Mock<IActiveSolutionBoundTracker>();
+
+            var testSubject = new BoundSolutionUpdateHandler(activeSolutionTracker.Object, Mock.Of<ISuppressionIssueStoreUpdater>());
+
+            testSubject.Dispose();
+
+            activeSolutionTracker.VerifyRemove(x => x.SolutionBindingChanged -= It.IsAny<EventHandler<ActiveSolutionBindingEventArgs>>(), Times.Once());
+            activeSolutionTracker.VerifyRemove(x => x.SolutionBindingUpdated -= It.IsAny<EventHandler>(), Times.Once());
```
bigfluffycookie, line 75: also remove these parentheses

src/ConnectedMode/BoundSolutionUpdateHandler.cs
```diff
@@ -0,0 +1,81 @@
...
+        public BoundSolutionUpdateHandler(IActiveSolutionBoundTracker activeSolutionBoundTracker, ISuppressionIssueStoreUpdater suppressionIssueStoreUpdater)
+        {
+            this.activeSolutionBoundTracker = activeSolutionBoundTracker;
+            this.suppressionIssueStoreUpdater = suppressionIssueStoreUpdater;
+
+            this.activeSolutionBoundTracker.SolutionBindingChanged += OnSolutionBindingChanged;
+            this.activeSolutionBoundTracker.SolutionBindingUpdated += OnSolutionBindingUpdated;
+        }
+
+        private void OnSolutionBindingUpdated(object sender, EventArgs e)
```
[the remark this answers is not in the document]
duncanp-sonar: The alternative is to move the call to "UpdateXXX" to a new method called e.g. "Refresh"/"TriggerUpdate".

The new method would only contain a single line, so the benefit of creating a new method isn't huge (there would be more reason to do it if e.g. we also had error handling / logging)

However, I think it's still worth creating the method. It provides slightly cleaner separation between handling the event and acting on it. It also makes it less likely that the implementation of the two event-handling methods could get out of step in the future if/when we need to make changes.

src/ConnectedMode/Suppressions/ServerSuppressionsChangedHandler.cs
```diff
@@ -26,11 +26,13 @@ namespace SonarLint.VisualStudio.ConnectedMode.Suppressions
 {
```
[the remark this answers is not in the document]
duncanp-sonar: Fixing it here is fine.

Your implementation of the the Dispose pattern is text-book correct i.e. it is correct and follows the recommended full pattern.

However, at the risk of confusing you about the implementation of the dispose pattern, the simple implementation that existed before was very nearly ok too - the only thing that needed to change was to make the class `sealed`.
If the class is sealed, then the previous "short-form" version would be ok too.

In this case, the short version is preferable in some ways - the "full" Dispose pattern implementation is longer than the rest of the code in the class, so it makes it harder to see what the class is actually doing.

So, here I would revert to the old, short implementation, but make the class `sealed`.
I would do the same for the implementation in `BoundSolutionUpdateHandler` too.


There are some [useful posts](https://blog.stephencleary.com/2009/08/how-to-implement-idisposable-and.html) on the Dispose pattern by Stephen Cleary which explain when it is needed/not needed more clearly than the MS docs (although his posts don't mention event handlers).

There is an [answer by Jon Skeet](https://stackoverflow.com/questions/17399991/should-i-always-disconnect-event-handlers-in-the-dispose-method) on StackOverflow that relates to unregistering event handlers in `Dispose`, which in essence says if you are sure the publisher object doesn't live longer than the subscriber then you don't need to. 

Let me know if you'd like to discuss the pattern further offline.

src/ConnectedMode.UnitTests/BoundSolutionUpdateHandlerTests.cs
duncanp-sonar: I think we have a couple of tests that do that, although I think it's unnecessary here since you are calling `VerifyRemove` to check that the event handler is unregistered.

Essentially, one is white-box test (i.e. the test has some knowledge of how the method is implemented), and the other is a black-box test (i.e. the test knows nothing about how the test is implemented).

Yours is a white-box version - it checks that the product code unregisters a specific event handler.

The alternative code you suggested above is a black-box version: it doesn't know what the product code does internally, but it observes the externally visible effects (i.e. whether `UpdateAllServerSuppressionsAsync` is called or not).

duncanp-sonar (approved):
Looks good.

I've added some remarks about the implementation of Dispose, but nothing major.

PR #3858, opened 2023-03-10
Add rule description page to visualization node convertion, add missing static resources

georgii-borovinskikh-sonarsource:


src/Education.UnitTests/Layout/ContentSectionTests.cs
```diff
@@ -18,29 +18,29 @@
...
-            xamlContentMock.Setup(x => x.GetObjectRepresentation()).Returns(content);
-            var testSubject = new ContentSection(xamlContentMock.Object);
-
-            var visualization = testSubject.CreateVisualization();
-
-            visualization.Should().BeSameAs(content);
-        }
-    }
-}
+// using System.Windows.Documents;
```
ugras-ergun-sonarsource, line 21: why not delete this?

src/Education/XamlParser/IXamlBlockContent.cs
```diff
@@ -17,35 +17,31 @@
...
-    internal class StaticXamlBlockContent : IXamlBlockContent
-    {
-        public Block GetObjectRepresentation()
-        {
-            // todo something like XamlReader.Load(someXamlResource)
-            throw new System.NotImplementedException();
-        }
-    }
-}
+//
```
ugras-ergun-sonarsource, line 20: Why not delete this?

src/Education/Layout/Logical/HowToFixItSection.cs
```diff
@@ -45,10 +46,21 @@ public HowToFixItSection(List<HowToFixItSectionContext> contexts)
 
         public string Title => "How can I fix it?";
 
-        public IAbstractVisualizationTreeNode GetVisualizationTreeNode(ITabsRepository tabsRepository)
+        public IAbstractVisualizationTreeNode GetVisualizationTreeNode(IStaticXamlStorage staticXamlStorage)
         {
-            throw new NotImplementedException();
-            // todo in the next PR
+            if (partialXamlContent != null)
```
ugras-ergun-sonarsource, line 51: This feels wrong. The class does two different things. But it's very simple so splitting the class could be overoptimization. Have you considered a quick way to do it?

ugras-ergun-sonarsource (commented):
I have a comment about a class and there are many commented out code instead of deletion. What is the reasoning behind that?

src/Education/Layout/Logical/HowToFixItSection.cs
georgii-borovinskikh-sonarsource: I chose this way because it's less boilerplate. I'm not sure if splitting this into two classes would add any value

ugras-ergun-sonarsource (approved):

Issue #3865, opened 2023-03-13
Change existing code to use IServerIssuesStore event instead of ISuppressedIssuesMonitor


duncanp-sonar:
### Description

- Roslyn suppressions handler
- ClientSuppressionsSync?

PR #3867, opened 2023-03-13
Change existing code to use IServerIssuesStore event instead of ISuppressedIssuesMonitor event

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3865


src/ConnectedMode/Suppressions/IServerIssuesStore.cs
```diff
@@ -28,7 +28,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Suppressions
     /// Stores all server issues known to the client for the current branch i.e. that have been fetched from the server.
     /// </summary>
     /// <returns>There may be other server-side issues that have not been fetched</returns>
-    internal interface IServerIssuesStore
```
bigfluffycookie, line 31: Is it okay to make this public?

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/InProcess/RoslynSettingsFileSynchronizerTests.cs
```diff
@@ -24,6 +24,7 @@
 using FluentAssertions;
```
bigfluffycookie, line 24: Just replaced IssueMonitor with the serverissuesStore

src/ConnectedMode/Suppressions/IServerIssuesStore.cs
duncanp-sonar: It probably needs to be now.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -68,24 +69,24 @@ internal sealed class RoslynSettingsFileSynchronizer : IRoslynSettingsFileSynchr
...
             this.activeSolutionBoundTracker = activeSolutionBoundTracker;
             this.logger = logger;
             this.threadHandling = threadHandling;
 
-            suppressedIssuesMonitor.ServerSuppressionsChanged += SuppressedIssuesMonitor_SuppressionsUpdateRequested;
+            serverIssuesStore.ServerIssuesChanged += ServerIssuesChanged;
         }
 
-        private void SuppressedIssuesMonitor_SuppressionsUpdateRequested(object sender, EventArgs e)
+        private void ServerIssuesChanged(object sender, EventArgs e)
```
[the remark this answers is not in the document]
duncanp-sonar: We don't have a consistent naming pattern for event handlers, but two common ones are "OnXXX" and "HandleXXX" e.g. `OnServerIssuesChanged` and `HandleServerIssuesChanged`.

Using exactly the same name as the event can be a bit confusing e.g. if you do a text search, it might not be obvious whether you are looking at the event or the handler.

duncanp-sonar (approved):

PR #3874, opened 2023-03-14
Replace SonarQubeIssueProvider with IServeIssuesStore

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3873


src/Integration/Suppression/SuppressedIssueMatcher.cs
```diff
@@ -51,13 +51,12 @@ public bool SuppressionExists(IFilterableIssue issue)
 
             // File-level issues never match non-file-level issues.
 
-            // Retrieve all issues relating to this file (file level or precise location) or project (for module level issues)
-            var serverIssues = issuesProvider.GetSuppressedIssues(issue.ProjectGuid, issue.FilePath);
+            var serverIssues = serverIssuesStore.Get();
 
             // Try to find an issue with the same ID and either the same line number or some line hash
-            bool matchFound = serverIssues.Any(s => IsMatch(issue, s));
+            bool foundIssueIsSuppressed = serverIssues.Any(s => IsMatch(issue, s) && s.IsResolved);
```
bigfluffycookie, line 57: Not a huge fan of the name but I felt like it needed to be changed

src/Integration.UnitTests/Suppression/SuppressedIssueMatcherTests.cs
```diff
@@ -130,20 +145,25 @@ private IFilterableIssue CreateIssueToMatch(string ruleId, int? startLine, strin
...
+                     ? new IssueTextRange(startLine.Value, 1, 1, 1)
+                     : null,
+                 flows: null);
+
+            sonarQubeIssue.IsResolved = isSuppressed;
+
+            return sonarQubeIssue;
+        }
 
         private void ConfigureServerIssues(
```
bigfluffycookie, line 162: Removed issueToMatch as we no longer needed it (The new Get() does not have any parameters)

src/Integration.UnitTests/Suppression/SuppressedIssueMatcherTests.cs
```diff
@@ -22,6 +22,7 @@
 using FluentAssertions;
```
bigfluffycookie, line 22: For all the existing tests I made the server side issue be suppressed=true simply so it doesnt affect what the tests were doing. Then I added a new test that only checks the isSuppressed state.

src/Integration/Suppression/SuppressedIssueMatcher.cs
duncanp-sonar: Alternatives: `isSuppressed`, or `matchesSuppressedIssue`.

Perf: checking `s.IsResolved` will be fast because it is just checking a property rather than calling a method, so it would be better to put that check first.

src/Integration.UnitTests/Suppression/SuppressedIssueMatcherTests.cs
```diff
@@ -111,10 +111,23 @@ public void MatchExists_MultipleServerIssues(string localRuleId, int localIssueL
...
+                CreateServerIssue("bbb", 333, "bbb hash", isSuppressed: true),
+                CreateServerIssue("ccc", 444, "ccc hash", isSuppressed: true));
+
+            // Act and assert
+            testSubject.SuppressionExists(issueToMatch).Should().Be(expectedResult);
+        }
+
+        [DataTestMethod]
+        [DataRow("RightRuleId", null, null, true, true)]
+        [DataRow("RightRuleId", null, null, false, false)]
```
duncanp-sonar, line 125: Minor: "Right" is potentially ambiguous - "Correct" or "Matching" are less ambiguous.

duncanp-sonar (approved):
Looks good. Just a couple of minor suggestions.

PR #3876, opened 2023-03-14
Remove ISonarQubeIssueProvider from RoslynSettingsSynchronizer

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3875


src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/InProcess/RoslynSettingsFileSynchronizerTests.cs
```diff
@@ -47,7 +47,6 @@ public void MefCtor_CheckIsExported()
         {
```
bigfluffycookie, line 47: Simply remove Isonarqubeissueprovider and used the issue store instead.

duncanp-sonar (approved):

Issue #3868, opened 2023-03-13
Check RoslynSettingsFileSynchronizer only stores suppressed issues


duncanp-sonar:
### Description

So far the `RoslynSettingsFileSynchronizer` has used the `ISuppressedIssuesProvider`, which only return suppressed issues.

Once we've switched it to use `IServerIssuesStore` we'll need to make sure it is only saving suppressed issues to the file, not all issues.

PR #3878, opened 2023-03-14
Change RoslynSettingsFileSynchronizer to only save suppressed issues

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3868

### DIFF
--- a/src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/InProcess/RoslynSettingsFileSynchronizerTests.cs
+++ b/src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/InProcess/RoslynSettingsFileSynchronizerTests.cs
@@ -194,6 +194,46 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions.UnitTests.InProcess
             actualSuppressions[1].RoslynRuleId.Should().Be("S222");
         }
 
+        [TestMethod]
+        public async Task UpdateFileStorage_OnySuppressedIssuesAreInSettings()
+        {
+            var configuration = CreateConnectedConfiguration("some project key");
+            var activeSolutionBoundTracker = CreateActiveSolutionBoundTracker(configuration);
+
+            var sonarIssues = new[]
+            {
+                CreateSonarQubeIssue(ruleId: "csharpsquid:S111", isSuppressed:false),
+                CreateSonarQubeIssue(ruleId: "vbnet:S222", isSuppressed:true),
+                CreateSonarQubeIssue(ruleId: "csharpsquid:S333", isSuppressed:true),
+                CreateSonarQubeIssue(ruleId: "vbnet:S444", isSuppressed:false),
+            };
+
+            var serverIssuesStore = CreateServerIssuesStore(sonarIssues);
+
+            RoslynSettings actualSettings = null;
+            var roslynSettingsFileStorage = new Mock<IRoslynSettingsFileStorage>();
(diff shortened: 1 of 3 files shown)


src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/InProcess/RoslynSettingsFileSynchronizerTests.cs
```diff
@@ -194,6 +194,46 @@ public async Task UpdateFileStorage_IssuesAreConvertedAndFiltered()
             actualSuppressions[1].RoslynRuleId.Should().Be("S222");
         }
 
+        [TestMethod]
```
bigfluffycookie, line 197: This test is almost a complete copy from the test above. The only change is that it tests if suppressed issues get filtered

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/TestHelper.cs
```diff
@@ -37,11 +37,13 @@ internal static class TestHelper
...
-    public static SonarQubeIssue CreateSonarQubeIssue(string ruleId = "any",
-            int? line = null,
-            string filePath = "filePath",
-            string hash = "hash") =>
-            new SonarQubeIssue(
+        public static SonarQubeIssue CreateSonarQubeIssue(string ruleId = "any",
+                int? line = null,
+                string filePath = "filePath",
+                string hash = "hash",
+                bool isSuppressed = true)
```
bigfluffycookie, line 44: Made the default for isSuppressed true so it does not interfere with existing tests.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -112,7 +112,7 @@ public async Task UpdateFileStorageAsync()
 
                 if (!string.IsNullOrEmpty(sonarProjectKey))
                 {
-                    var allSuppressedIssues = serverIssuesStore.Get();
+                    var allSuppressedIssues = serverIssuesStore.Get().Where(x => x.IsResolved).ToArray();
```
bigfluffycookie, line 115: Still a bit unsure about linq syntax so if this is wrong lmk

duncanp-sonar: The syntax is fine.

You could skip the `ToArray` here, though; `allSuppressedIssues` is used in another LINQ expression immediately below which also has a `ToArray`, so we probably don't need it here as well.

An alternative approach would be to add the check into the LINQ expression below i.e.

```
       var allIssues = serverIssuesStore.Get();

       var settings = new RoslynSettings
                    {
                        SonarProjectKey = sonarProjectKey,
                        Suppressions = allIssues
                                            .Where(x => x.IsSuppressed)        <----- CHECK ADDED HERE
                                            .Select(x => IssueConverter.Convert(x))
                                            .Where(x => x.RoslynLanguage != RoslynLanguage.Unknown && !string.IsNullOrEmpty(x.RoslynRuleId))
                                            .ToArray(),
                    };

```

(I would probably go with the second approach and put all of the LINQ conditions in one place).

duncanp-sonar (approved):
Looks good. One comment/suggestion about the LINQ expression.

PR #3882, opened 2023-03-14
Extract xaml header and closer generation to a separate class

georgii-borovinskikh-sonarsource:
It is going to be reused by Simple and Rich layout generators.

NOTE: no new code, just moved the bits to a separate class

### DIFF
--- a/src/Education/XamlGenerator/XamlGeneratorHelper.cs
+++ b/src/Education/XamlGenerator/XamlGeneratorHelper.cs
@@ -1,0 +1,125 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


src/Education/XamlGenerator/XamlGeneratorHelper.cs
```diff
@@ -0,0 +1,125 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Xml;
+using SonarLint.VisualStudio.Rules;
+
+namespace SonarLint.VisualStudio.Education.XamlGenerator
+{
+    internal class XamlGeneratorHelper
```
ugras-ergun-sonarsource, line 26: Since this is now separated it needs it's own tests

ugras-ergun-sonarsource: Also it needs an interface

ugras-ergun-sonarsource (commented):
I think new class should have tests and an interface.

src/Education/XamlGenerator/XamlGeneratorHelper.cs
georgii-borovinskikh-sonarsource: I agree with adding test (although I didn't find any for the Simple layout generator, which is why I didn't add any), but I don't think adding an interface here would add any value. It's not going to be used in any way, not even in tests

ugras-ergun-sonarsource (approved):

PR #3887, opened 2023-03-14
Trigger a suppressions update when the package loads

### LINKED ISSUE #3886: Issues are not suppressed immediately when a bound solution is opened
### Description

We expect server-side issues suppressions to be applied when a bound solution is opened (or at least within a few seconds, since the issues probably need to be fetched from the server). However, there can be a considerable delay (multiple minutes) after the first bound solution is opened before issues are suppressed.

### Repro
1. Open VS
2. Open a bound solution with suppressed issues
3. Open a file with suppressed issues

### Expected behaviour
The suppressed issues are actually suppressed (i.e. no squigglies in the editor, and the issues appear as suppressed in the Error List).

### Actual behaviour
The issues are not suppressed immediately, and won't be suppressed until the suppression update timer elapses.

### Notes
The Connected Mode package creates the components that fetch suppressions and listen for solution change notifications.

However, the package is delay-loaded i.e. it isn't loaded until Connected Mode has been established for the first time.

This means the `BoundSolutionUpdateHandler` isn't created until after the "solution changed" event from the `ActiveSolutionBoundTracker` has fired -> it doesn't trigger a fetch.

Suppressions won't be fetched until either the binding is updated, the update timer elapses, or another bound solution is opened.

### Solution
Trigger an suppression update when the package loads.
This will be a no-op if the solution is not actually in Connected Mode.

duncanp-sonar:
Fixes #3886

### DIFF
--- a/src/ConnectedMode/ConnectedModePackage.cs
+++ b/src/ConnectedMode/ConnectedModePackage.cs
@@ -64,6 +64,12 @@ namespace SonarLint.VisualStudio.ConnectedMode
             boundSolutionUpdateHandler = componentModel.GetService<BoundSolutionUpdateHandler>();
             timedUpdateHandler = componentModel.GetService<TimedUpdateHandler>();
 
+            // Trigger an initial update of suppressions (we might have missed the solution binding
+            // event from the ActiveSolutionBoundTracker)
+            // See https://github.com/SonarSource/sonarlint-visualstudio/issues/3886
+            var updater = componentModel.GetService<ISuppressionIssueStoreUpdater>();
+            updater.UpdateAllServerSuppressionsAsync().Forget();
+
             logger.WriteLine(Resources.Package_Initialized);
         }
 


src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -64,6 +64,12 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
             boundSolutionUpdateHandler = componentModel.GetService<BoundSolutionUpdateHandler>();
             timedUpdateHandler = componentModel.GetService<TimedUpdateHandler>();
 
+            // Trigger an initial update of suppressions (we might have missed the solution binding
+            // event from the ActiveSolutionBoundTracker)
+            // See https://github.com/SonarSource/sonarlint-visualstudio/issues/3886
+            var updater = componentModel.GetService<ISuppressionIssueStoreUpdater>();
+            updater.UpdateAllServerSuppressionsAsync().Forget();
```
duncanp-sonar, line 71: I've manually tested that this causes suppressions to be applied the first time a bound solution is open.
There is a slight delay as the issues are fetched from the server (i.e. the issue can appear in the Error List briefly).
However, this will always be the case whenever a bound solution is opened; it's not specific to startup.

bigfluffycookie (approved):
nice

Issue #3890, opened 2023-03-15
Add suppression logging


duncanp-sonar:
### Description

Decide what log messages would be useful in 
1) normal operation
2) support/troubleshooting/debugging

PR #3891, opened 2023-03-15
Add suppression logging to event listener and store UpdateIssue

duncanp-sonar:
Part of #3890

### DIFF
--- a/src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
+++ b/src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
@@ -20,8 +20,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.Composition.Primitives;
 using System.Linq;
 using SonarLint.VisualStudio.ConnectedMode.Suppressions;
+using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.TestInfrastructure;
 using SonarQube.Client.Models;
 
@@ -33,8 +35,32 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
         [TestMethod]
         public void MefCtor_CheckIsExported()
         {
-            MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStore>();
-            MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStoreWriter>();
+            MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStore>(
+                MefTestHelpers.CreateExport<ILogger>());
+
+            MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStoreWriter>(
+                MefTestHelpers.CreateExport<ILogger>());
+        }
+
+        [TestMethod]
+        public void MefCtor_Check_SameInstanceExported()
+        {
+            // Importing via either exported interface should return the same store instance
+
+            var storeImporter = new SingleObjectImporter<IServerIssuesStore>();
+            var storeWriterImporter = new SingleObjectImporter<IServerIssuesStoreWriter>();
+            var importers = new object[] { storeImporter, storeWriterImporter };
+
+            var loggerExport = MefTestHelpers.CreateExport<ILogger>();
+
+            var typesToExport = new Type[] { typeof(ServerIssuesStore) };
+
+            MefTestHelpers.Compose(importers, typesToExport, loggerExport);
+
+            storeImporter.AssertImportIsNotNull();
+            storeWriterImporter.AssertImportIsNotNull();
+
+            storeImporter.Import.Should().BeSameAs(storeWriterImporter.Import);
         }
 
         [TestMethod]
@@ -45,7 +71,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
             var issue1 = CreateIssue("1", false);
             var issue2 = CreateIssue("2", false);
 
-            var testSubject = new ServerIssuesStore();
+            var testSubject = CreateTestSubject();
 
             var result = testSubject.Get();
             result.Count().Should().Be(0);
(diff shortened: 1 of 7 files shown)


src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
```diff
@@ -33,8 +35,32 @@ public class ServerIssuesStoreTests
...
-            MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStoreWriter>();
+            MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStore>(
+                MefTestHelpers.CreateExport<ILogger>());
+
+            MefTestHelpers.CheckTypeCanBeImported<ServerIssuesStore, IServerIssuesStoreWriter>(
+                MefTestHelpers.CreateExport<ILogger>());
+        }
+
+        [TestMethod]
+        public void MefCtor_Check_SameInstanceExported()
```
duncanp-sonar, line 46: Miscellaneous: new test to check that the exports return the same instance rather than different instances.

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
```diff
@@ -45,7 +71,7 @@ public void AddIssues_EmptyList_ResultContainsNewIssues(bool clearAllExistingIss
             var issue1 = CreateIssue("1", false);
             var issue2 = CreateIssue("2", false);
 
-            var testSubject = new ServerIssuesStore();
+            var testSubject = CreateTestSubject();
```
duncanp-sonar, line 74: New helper method since we now need a logger.

src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs
```diff
@@ -183,11 +209,36 @@ public void UpdateIssues_NonEmptyList_Match_InvokesEventAndPropertyIsChanged()
             issue2.IsResolved.Should().BeFalse();
             issue3.IsResolved.Should().BeTrue();
 
-            // This test is to insure it doesn't just flip the state.
```
duncanp-sonar, line 186: This check is now done by the new test, so it's no longer required here

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
```diff
@@ -57,7 +65,7 @@ public void AddIssues(IEnumerable<SonarQubeIssue> issues, bool clearAllExistingI
 
             serverIssues.AddRange(issues);
 
-            ServerIssuesChanged?.Invoke(this, EventArgs.Empty);
+            RaiseEventIfRequired();
```
duncanp-sonar, line 68: I moved this to a separate method since it is now logging a message too

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
```diff
@@ -66,12 +74,30 @@ public void UpdateIssue(string issueKey, bool isResolved)
 
             if (issue == null)
             {
+                logger.LogVerbose(Resources.Store_UpdateIssue_NoMatch, issueKey);
                 return;
             }
 
-            issue.IsResolved = isResolved;
+            if (issue.IsResolved == isResolved)
```
duncanp-sonar, line 81: This is a change of behaviour - the event is only raised if the property has actually changed.

I guess I could have made this change first in a separate PR but I'd already made most of the logging changes by then, and I decided it was a small enough change that it wasn't worth the extra work.

src/SonarQube.Client/Models/ServerSentEvents/ClientContract/IIssueChangedServerEvent.cs
```diff
@@ -53,6 +54,19 @@ public IssueChangedServerEvent(string projectKey, bool isResolved, BranchAndIssu
...
+        public override string ToString()
+        {
+            var sb = new StringBuilder();
+            sb.AppendLine($"ProjectKey: {ProjectKey}, IsResolved: {IsResolved}, Issue count: " + BranchAndIssueKeys.Length);
+            foreach (var item in BranchAndIssueKeys)
+            {
+                sb.AppendLine($"\tBranch: {item.BranchName}, IssueKey: {item.IssueKey}");
+            }
+
+            return sb.ToString();
```
duncanp-sonar, line 68: Overriding `ToString` so that
```
logger.LogVerbose(Resources.Suppression_IssueChangedEventReceived, issueServerEvent);
```
... prints something useful.

This `ToString` method uses a StringBuilder i.e. creates objects and calls other functions to generate the text.

However,  the way the `LogVerbose` method is written means it will only be called when we are actually in verbose logging mode - it won't be called in "normal" logging mode.

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
bigfluffycookie: Yea makes sense!

bigfluffycookie (approved):
Looks good just a minor comment :)

src/ConnectedMode/Suppressions/ServerIssuesStore.cs
```diff
@@ -66,12 +74,30 @@ public void UpdateIssue(string issueKey, bool isResolved)
...
+            }
+        }
 
-            ServerIssuesChanged?.Invoke(this, EventArgs.Empty);
+        private void RaiseEventIfRequired()
+        {
+            if (ServerIssuesChanged != null)
+            {
+                logger.WriteLine(Resources.Store_UpdateIssue_RaisingEvent);
+                ServerIssuesChanged?.Invoke(this, EventArgs.Empty);
```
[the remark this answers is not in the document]
duncanp-sonar: I wondered about that but decided it was safer to leave it. In theory, in a multi-threaded environment the event handler could be unregistered on another thread after we have done the check on line 96, but before we get to line 99.

It's unlikely, and in our case it shouldn't happen the way the code is currently written, but it doesn't do any harm to leave it like this.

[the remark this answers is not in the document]
bigfluffycookie: Fair enough

duncanp-sonar:
@bigfluffycookie I'll add a couple of simple tests for the "ToString" method to stop SonarCloud complaining about the coverage

Issue #3892, opened 2023-03-15
Update issue store when SSE issue event is raised (simple version)


duncanp-sonar:
### Description

Handle issue changed event i.e. update the store

Note: the StoreWriter API currently expects one issue to be changed at a time, but the issue events are supposed to be raised in batch.

Need to check whether the events are actually batched, and if so adjust the StoreWriter API.

PR #3893, opened 2023-03-15
Add simple implementation of issue event handling

duncanp-sonar:
Simplistic implementation to allow further testing and investigation. More work required.

Part of #3892

### DIFF
--- a/src/ConnectedMode.UnitTests/ServerSentEvents/Issue/IssueServerEventsListenerTests.cs
+++ b/src/ConnectedMode.UnitTests/ServerSentEvents/Issue/IssueServerEventsListenerTests.cs
@@ -21,6 +21,7 @@
 using System;
 using System.Threading.Tasks;
 using SonarLint.VisualStudio.ConnectedMode.ServerSentEvents.Issue;
+using SonarLint.VisualStudio.ConnectedMode.Suppressions;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.TestInfrastructure;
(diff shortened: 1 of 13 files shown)


src/ConnectedMode/ServerSentEvents/Issue/IssueServerEventsListener.cs
```diff
@@ -72,8 +76,12 @@ public async Task ListenAsync()
                         return;
                     }
 
-                    // TODO: Implement handling of event.
                     logger.LogVerbose(Resources.Suppression_IssueChangedEventReceived, issueServerEvent);
+                    foreach(var item in issueServerEvent.BranchAndIssueKeys)
+                    {
+                        // TODO - check branch handling. Are issueKeys unique for each branch?
+                        serverIssuesStoreWriter.UpdateIssue(item.IssueKey, issueServerEvent.IsResolved);
```
duncanp-sonar, line 83: This is a simple implementation so we can do some testing with SonarQube.

We might need to change the writer API to handle bulk changes.

src/ConnectedMode.UnitTests/ServerSentEvents/Issue/IssueServerEventsListenerTests.cs
```diff
@@ -36,10 +37,56 @@ public void MefCtor_CheckIsExported()
         {
             MefTestHelpers.CheckTypeCanBeImported<IssueServerEventsListener, IIssueServerEventsListener>(
                 MefTestHelpers.CreateExport<IIssueServerEventSource>(),
+                MefTestHelpers.CreateExport<IServerIssuesStoreWriter>(),
                 MefTestHelpers.CreateExport<IThreadHandling>(),
                 MefTestHelpers.CreateExport<ILogger>());
         }
 
+        [TestMethod]
+        public async Task OnEvent_SingleEvent_MultipleTuples_StoreIsUpdated()
```
duncanp-sonar, line 46: Two new simple tests that check the writer is called when events are raised

bigfluffycookie (approved):
Nice!

PR #3906, opened 2023-03-16
Pass issue keys to api/issues/search

### LINKED ISSUE #3902: Add webclient method to fetch specific issues
### Description

Add web API call for `api/issues/search?issues=[list of issues keys]`

The API exists in both [SonarCloud ](https://sonarcloud.io/web_api/api/issues/search?deprecated=false) and [SonarQube](https://next.sonarqube.com/sonarqube/web_api/api/issues/search?deprecated=false).

rita-sonarsource:
Fixes  #3902

### DIFF
--- a/src/SonarQube.Client.Tests/SonarQubeService_GetSuppressedIssuesAsync.cs
+++ b/src/SonarQube.Client.Tests/SonarQubeService_GetSuppressedIssuesAsync.cs
@@ -52,7 +52,7 @@ namespace SonarQube.Client.Tests
                 SetupRequest("batch/issues?key=project1", reader.ReadToEnd());
             }
 
-            var result = await service.GetSuppressedIssuesAsync("project1", null, CancellationToken.None);
+            var result = await service.GetSuppressedIssuesAsync("project1", null, null, CancellationToken.None);
 
             // TODO: create a protobuf file with more than one issue with different states
             // the one above does not have suppressed issues, hence the Count==0
@@ -69,7 +69,7 @@ namespace SonarQube.Client.Tests
             SetupRequest("batch/issues?key=project1", "", HttpStatusCode.NotFound);
 
             Func<Task<IList<SonarQubeIssue>>> func = async () =>
-                await service.GetSuppressedIssuesAsync("project1", null, CancellationToken.None);
+                await service.GetSuppressedIssuesAsync("project1", null, null, CancellationToken.None);
 
             func.Should().ThrowExactly<HttpRequestException>().And
                 .Message.Should().Be("Response status code does not indicate success: 404 (Not Found).");
@@ -251,7 +251,7 @@ namespace SonarQube.Client.Tests
 }
 ");
 
-            var result = await service.GetSuppressedIssuesAsync("shared", null,CancellationToken.None);
+            var result = await service.GetSuppressedIssuesAsync("shared", null, null,CancellationToken.None);
 
             result.Should().HaveCount(4);
 
@@ -304,7 +304,7 @@ namespace SonarQube.Client.Tests
             SetupRequest("api/issues/search?projects=project1&statuses=RESOLVED&types=CODE_SMELL&p=1&ps=500", "", HttpStatusCode.NotFound);
 
             Func<Task<IList<SonarQubeIssue>>> func = async () =>
-                await service.GetSuppressedIssuesAsync("project1", null, CancellationToken.None);
+                await service.GetSuppressedIssuesAsync("project1", null, null, CancellationToken.None);
 
             func.Should().ThrowExactly<HttpRequestException>().And
                 .Message.Should().Be("Response status code does not indicate success: 404 (Not Found).");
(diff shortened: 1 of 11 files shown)


src/SonarQube.Client.Tests/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -376,12 +376,12 @@ public async Task GetSuppressedIssuesAsync_From_7_20_BranchIsNotSpecified_Branch
...
 
             SetupHttpRequest(messageHandler, EmptyGetIssuesResponse);
-            _ = await service.GetSuppressedIssuesAsync("any", emptyBranch, CancellationToken.None);
+            _ = await service.GetSuppressedIssuesAsync("any", emptyBranch, null, CancellationToken.None);
 
             // Branch is null/empty => should not be passed
             var actualRequests = messageHandler.GetSendAsyncRequests();
             actualRequests.Should().HaveCount(3);
-            actualRequests.All(x => x.RequestUri.Query.Contains("branch")).Should().BeFalse();
+            actualRequests.Any(x => x.RequestUri.Query.Contains("branch")).Should().BeFalse();
```
rita-sonarsource, line 384: I think it was a bug here, this would've been evaluated to false if it was not in one of the requests (but was in others). Realistically it wouldn't have happened, but I think this assertion is better.

src/SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
```diff
@@ -51,6 +51,9 @@ public string Statuses
         [JsonIgnore] // We don't support the branch parameter in v5.10
         public string Branch { get; set; }
 
+        [JsonIgnore] // Not needed in old API
```
duncanp-sonar, line 54: It looks like `issues` was supported in v5.5 so the comment that it is not needed is a bit confusing. Do you mean we're choosing not to support it on versions before 7.2? (which I think is ok, I'm just unclear what the comment means).

src/SonarQube.Client.Tests/SonarQubeService_GetSuppressedIssuesAsync.cs
```diff
@@ -391,23 +391,52 @@ public async Task GetSuppressedIssuesAsync_From_7_20_BranchIsSpecified_BranchInc
...
+            await ConnectToSonarQube("7.2.0.0");
+            messageHandler.Reset();
+
+            SetupHttpRequest(messageHandler, EmptyGetIssuesResponse);
+            _ = await service.GetSuppressedIssuesAsync("any", null, new[]{"issue1", "issue2"}, CancellationToken.None);
+
+            // The wrapper is expected to make three calls, for code smells, bugs, then vulnerabilities
+            var actualRequests = messageHandler.GetSendAsyncRequests();
+            actualRequests.Should().HaveCount(3);
+            actualRequests.All(x => x.RequestUri.Query.Contains("issues=issue1%2Cissue2")).Should().BeTrue();
```
duncanp-sonar, line 429: We shouldn't need to make three calls if we're fetching a known list of issues. Does it make sense from a SQ point of view to combine `issues` with other filters? If it works then I guess it's ok, but it seems a bit weird.

duncanp-sonar (commented):
Just a couple of questions about combining `issues` with other filter parameters.

src/SonarQube.Client/ISonarQubeService.cs
```diff
@@ -80,8 +80,10 @@ public interface ISonarQubeService
         /// </summary>
         /// <param name="projectKey">The project identifier</param>
         /// <param name="branch">(optional) The Sonar branch for which issues should be returned. If null/empty,
-        /// the issues for the "main" branch will be returned.</param>
-        Task<IList<SonarQubeIssue>> GetSuppressedIssuesAsync(string projectKey, string branch, CancellationToken token);
+        ///     the issues for the "main" branch will be returned.</param>
+        /// <param name="token"></param>
+        /// <param name="issueKeys">(optional) The ids of the issues to return. If empty, all issues will be returned.</param>
```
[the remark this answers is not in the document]
rita-sonarsource: I wondered about that, and I'm not sure. For taint I believe the key is unique. For issues, I'm not sure. Either way, we fetch the branch as part of our requests anyway, so I don't think it matters if we pass both (or none)

src/SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
rita-sonarsource: Ah, yes, bad wording.

src/SonarQube.Client.Tests/SonarQubeService_GetSuppressedIssuesAsync.cs
rita-sonarsource: Hmm, so what you're saying is, we can do something like
```
if issueKeys
    make 1 call with branch + issue keys
else
    make 3 calls with branch + type
```

duncanp-sonar: I would have thought so. I'm not sure how SQ handles combining the parameters.
The simplistic approach at the ISonarQubeService level would be to have a separate `GetSpecificIssues` that just branch/branch+issuesKeys (whatever is required)

duncanp-sonar (approved):

PR #3910, opened 2023-03-16
Raise single file changed event

### LINKED ISSUE #3909: Remove duplicate file changed notifications from SonarErrorListDataSource
### Description

- see new TODO in the file

duncanp-sonar:
Fixes #3909

### DIFF
--- a/src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_IssueLocationStoreTests.cs
+++ b/src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_IssueLocationStoreTests.cs
@@ -216,14 +216,13 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.ErrorList
         }
         [TestMethod]
         public void Refresh_FactoriesWithMatchingFilesUpdatedAndErrorListRefreshed()
-            => RefreshImpl_FactoriesWithMatchingFilesUpdatedAndErrorListRefreshed(CallRefresh, shouldRaiseIssuesChanged: true);
+            => RefreshImpl_FactoriesWithMatchingFilesUpdatedAndErrorListRefreshed(CallRefresh);
 
         [TestMethod]
         public void RefreshOnBufferChanged_FactoriesWithMatchingFilesUpdatedAndErrorListRefreshed()
-            => RefreshImpl_FactoriesWithMatchingFilesUpdatedAndErrorListRefreshed(CallRefreshOnBufferChanged, shouldRaiseIssuesChanged: false);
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_IssueLocationStoreTests.cs
```diff
@@ -264,6 +250,38 @@ public void RefreshOnBufferChanged_FactoriesWithMatchingFilesUpdatedAndErrorList
...
+
+            var issuesChangedEvent = CreateAndRegisterIssuesEventHandler(testSubject);
+
+            // XXX.txt and YYY.ccc won't match any of the snapshot files
+            testSubject.Refresh(new[] { "XXX.txt", "match1.aaa", "match2.bbb", "match3.ccc", "YYY.ccc", "match4.ddd", "match5.eee" });
+
+            // Should be one event with all directyl and indirectly affected files
+            CheckSingleIssuesChangedEvent(issuesChangedEvent, testSubject,
+                    "match1.aaa", "match2.bbb", "match3.ccc", "match4.ddd", "match5.eee",
+                    // this file is included because it is indirectly included i.e. is in a snapshot a file that does match
```
duncanp-sonar, line 280: I hadn't thought about indirectly included files and it wasn't tested before.
However, it's consistent with what we are doing when raising the event in other places.

src/Integration.Vsix.UnitTests/ErrorList/SonarErrorListDataSource_IssueLocationStoreTests.cs
```diff
@@ -264,6 +250,38 @@ public void RefreshOnBufferChanged_FactoriesWithMatchingFilesUpdatedAndErrorList
             CheckSinkNotifiedOfChangeToFactories(sinkMock2, factoryWithMatch1, factoryWithMatch2);
         }
 
+        [TestMethod]
+        public void RefreshImpl_SomeFactoriesWithMatchingFiles_RaisesSingleIssueChangeEventContainingExpectedFiles()
```
rita-sonarsource, line 254: I see this is only for Refresh, not RefreshOnBufferChanged. Why not use your previous way of testing both with the callback? It should just "work" since there should be only one factory, but it's worth asserting.

duncanp-sonar: The existing tests check that the event is raised/not raised in various circumstances (and `RefreshOnBufferChanged` should never raise the event).

This test is to specifically check the contents of the event args. Since `RefreshOnBufferChanged` doesn't raise the event, I thought it would be confusing to include it.

PR #3913, opened 2023-03-16
Trigger UI update when locally-suppressed issues change

### LINKED ISSUE #3912: Trigger UI update when suppressions changed
### Description

Trigger an update the whole UI (i.e. the Editor and Error List) when the set of local issues that are suppressed changes

duncanp-sonar:
Fixes #3912

### DIFF
--- a/src/ConnectedMode.UnitTests/Suppressions/ServerSuppressionsChangedHandlerTests.cs
+++ b/src/ConnectedMode.UnitTests/Suppressions/ServerSuppressionsChangedHandlerTests.cs
@@ -20,7 +20,6 @@
 
 using System;
 using SonarLint.VisualStudio.ConnectedMode.Suppressions;
-using SonarLint.VisualStudio.Core.Suppression;
 using SonarLint.VisualStudio.TestInfrastructure;
 
 namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
--- a/src/ConnectedMode/ConnectedModePackage.cs
+++ b/src/ConnectedMode/ConnectedModePackage.cs
@@ -45,6 +45,7 @@ namespace SonarLint.VisualStudio.ConnectedMode
         private ServerSuppressionsChangedHandler serverSuppressionsChangedHandler;
         private BoundSolutionUpdateHandler boundSolutionUpdateHandler;
         private TimedUpdateHandler timedUpdateHandler;
+        private LocalSuppressionsChangedHandler localSuppressionsChangedHandler;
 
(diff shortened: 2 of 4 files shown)


src/ConnectedMode.UnitTests/Suppressions/ServerSuppressionsChangedHandlerTests.cs
```diff
@@ -20,7 +20,6 @@
 
 using System;
 using SonarLint.VisualStudio.ConnectedMode.Suppressions;
-using SonarLint.VisualStudio.Core.Suppression;
```
duncanp-sonar, line 23: Miscellaneous: redundant using.

FYI the class and tests are essentially a copy of this test and associated class but with the different event sources and targets.

The `ServerSuppressionsChangedHandler` listens for updates to the list of server suppressions the client knows about and tells the `ClientSuppressionSynchronizer` to synch the local issues and server issue suppression states.

The new `LocalSuppressionsChangedHandler` listens for notifications from the  `ClientSuppressionSynchronizer` that something changed as a result of the synch, and triggers and update of the UI by calling `IIssueLocationStore.Refresh(files)`.

src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -63,6 +64,7 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
             serverSuppressionsChangedHandler = componentModel.GetService<ServerSuppressionsChangedHandler>();
             boundSolutionUpdateHandler = componentModel.GetService<BoundSolutionUpdateHandler>();
             timedUpdateHandler = componentModel.GetService<TimedUpdateHandler>();
+            localSuppressionsChangedHandler = componentModel.GetService<LocalSuppressionsChangedHandler>();
```
duncanp-sonar, line 67: The new class is MEF-created by the `ConnectedMode` package. It isn't referenced anywhere else, and we don't test the `ConnectedMode` package initialization, so I didn't create an interface for the type (same as for the other three new types in the lines immediately above).

src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -82,6 +84,7 @@ protected override void Dispose(bool disposing)
                 serverSuppressionsChangedHandler?.Dispose();
                 boundSolutionUpdateHandler?.Dispose();
                 timedUpdateHandler?.Dispose();
+                localSuppressionsChangedHandler.Dispose();
```
rita-sonarsource, line 87: missing nullable in case we didn't get to initialize it

rita-sonarsource (approved):

PR #3916, opened 2023-03-17
Handle adding existing issues to the store

### LINKED ISSUE #3915: Handle duplicates in ServerIssueStore
### Description

The issue store allows duplicate issues to be added via the `AddIssues` method. It should only be possible to add an item once. Adding the item again should replace the existing item.

duncanp-sonar:
Fixes #3915


duncanp-sonar:
@rita-sonarsource I'll sync with you before committing this since I think we're both changing this file.

rita-sonarsource (approved):

PR #3926, opened 2023-03-17
Update issues in batch

### LINKED ISSUE #3894: Handle batch issue notifications
### Description

Currently issue events are processed on the client one at a time.

However, a single event could affect multiple issues if the user has selected "Bulk Update" in the UI.

TODO
- decide whether to change the store writer API to handle bulk updates, or whether to leave them to be processed one at a time.

Handling bulk events one at a time will cause the Error List and taggers to be updated multiple times.

duncanp-sonar:
Fixes #3894

### DIFF
--- a/src/ConnectedMode/Suppressions/IServerIssuesStore.cs
+++ b/src/ConnectedMode/Suppressions/IServerIssuesStore.cs
@@ -51,9 +51,6 @@ namespace SonarLint.VisualStudio.ConnectedMode.Suppressions
         /// Updates the IsResolved status of the issues with the specified keys
         /// </summary>
         /// <remarks>If an issue key cannot be matched to an issue in the store it will be ignored.
-        /// Design decision: if we can't find an issue we could have decided to go to the server and
-        /// fetch it, but there could be dozens or even hundreds of IDEs listening to the server
-        /// events, so we don't want to trigger hundreds of requests to the server.
         /// </remarks>
         void UpdateIssues(bool isResolved, IEnumerable<string> issueKeys);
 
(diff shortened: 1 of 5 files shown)


src/ConnectedMode/Suppressions/ServerIssuesStore.cs
```diff
@@ -78,30 +78,36 @@ public void AddIssues(IEnumerable<SonarQubeIssue> issues, bool clearAllExistingI
...
         }
 
-        public void UpdateIssue(string issueKey, bool isResolved)
+        public void UpdateIssues(bool isResolved, IEnumerable<string> issueKeys)
         {
-            SonarQubeIssue issue = null;
+            bool issuesChanged = false;
             lock (serverIssuesLock)
             {
-                // Only need to lock when reading the dictionary
```
duncanp-sonar, line 86: Everything except raising the event is now inside the lock.

src/ConnectedMode/Suppressions/IServerIssuesStore.cs
```diff
@@ -48,14 +48,14 @@ public interface IServerIssuesStore
     internal interface IServerIssuesStoreWriter : IServerIssuesStore
     {
         /// <summary>
-        /// Updates the IsResolved status of the issue with the specified key 
+        /// Updates the IsResolved status of the issues with the specified keys
         /// </summary>
-        /// <remarks>If the issue key cannot be matched to an issue in the store it will be ignored.
+        /// <remarks>If an issue key cannot be matched to an issue in the store it will be ignored.
         /// Design decision: if we can't find an issue we could have decided to go to the server and
```
rita-sonarsource, line 54: Minor: I think the design decision comment is irrelevant now that we do do it in the updater

rita-sonarsource (approved):

PR #3936, opened 2023-03-21
Add SSE logs + start session inside the lock

rita-sonarsource:


src/ConnectedMode/ServerSentEvents/SSESessionFactory.cs
```diff
@@ -160,14 +162,22 @@ public async Task PumpAllAsync()
                             continue;
```
duncanp-sonar, line 162: Is worth logging the "null" event, if that represents a heartbeat?

duncanp-sonar (approved):

bigfluffycookie (approved):
Nice

PR #3938, opened 2023-03-21
fix xaml generation

georgii-borovinskikh-sonarsource:


duncanp-sonar:
The product changes look good; just need to fix up the tests.

ugras-ergun-sonarsource (approved):

PR #3940, opened 2023-03-21
Create wrapper for server and local meta data provider 

### LINKED ISSUE #3763: Create wrapper for server and local meta data provider

ugras-ergun-sonarsource:
fixes #3763


src/Rules/IRuleInfo.cs
georgii-borovinskikh-sonarsource, line 1: why add setters to the interface? I would keep it immutable, but add a method like .With(htmlNote, severity) that returns a new instance

georgii-borovinskikh-sonarsource (approved):

src/Rules/IRuleInfo.cs
georgii-borovinskikh-sonarsource: Also the name default severity now makes less sense because it's overriden by QP. In SQ default means Sonar way, I think

ugras-ergun-sonarsource: Yeah, that was a temp thing I did on Friday and forgot. Will fix it. 
For the default severity, name comes from the SonarAPI. But we can have a jsonproperty with default severity

src/Rules/IRuleInfo.cs
```diff
@@ -170,5 +176,15 @@ public class RuleInfo : IRuleInfo
         public IReadOnlyList<string> EducationPrinciples { get; }
 
         public string HtmlNote { get; set; }
+
+        public IRuleInfo WithOverridenSeverity(RuleIssueSeverity newSeverity)
```
georgii-borovinskikh-sonarsource, line 180: Maybe combine the 2 override methods? They're not super useful on their own and create an extra unnecessary object that's easy to avoid

georgii-borovinskikh-sonarsource (approved):

src/Rules/IRuleInfo.cs
georgii-borovinskikh-sonarsource: they're not used separately is what I rather meant

Issue #3944, opened 2023-03-22
Consider dropping issue filtering from the issue consumer


rita-sonarsource:
- could use the client issue synchronizer later to mark suppressed issues, rather than having the issue consumer do it
- think about impact on UX (ideally don't want issues appearing the Error List then disappearing just after when they are marked as suppressed)

PR #3959, opened 2023-03-23
Replace issue filter in IssueConsumer_IssueHandler with IClientSuppressionSynchronizer.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3944

### DIFF
--- a/src/ConnectedMode/Suppressions/IssuesFilter.cs
+++ b/src/ConnectedMode/Suppressions/IssuesFilter.cs
@@ -26,7 +26,7 @@ using SonarLint.VisualStudio.Core.Suppression;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Suppressions
 {
-    public interface IIssuesFilter
+    internal interface IIssuesFilter
     {
         /// <summary>
         /// Returns the list of supplied objects that match the filter
@@ -37,7 +37,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Suppressions
 
     [Export(typeof(IIssuesFilter))]
     [PartCreationPolicy(CreationPolicy.Shared)]
-    public class IssuesFilter : IIssuesFilter
+    internal class IssuesFilter : IIssuesFilter
     {
         private readonly ISuppressedIssueMatcher issueMatcher;
 
--- a/src/Integration.Vsix.UnitTests/Analysis/IssueHandlerTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/IssueHandlerTests.cs
@@ -79,30 +79,19 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Analysis
         }
 
         [TestMethod]
-        public void HandleNewIssues_IssuesGetMatchesIsCalled()
+        public void HandleNewIssues_ClientSuppressionSynchronizerIsCalled()
         {
-            var inputIssues = new[]
-            {
-                CreateIssue("S111", startLine: 1, endLine: 1),
-                CreateIssue("S222", startLine: 2, endLine: 2)
-            };
-
             var notificationHandler = new SnapshotChangeHandler();
-            var issuesFilter = new Mock<IIssuesFilter>();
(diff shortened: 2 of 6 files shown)


src/ConnectedMode/Suppressions/IssuesFilter.cs
```diff
@@ -26,7 +26,7 @@
 
 namespace SonarLint.VisualStudio.ConnectedMode.Suppressions
 {
```
bigfluffycookie, line 28: Does not need to be public anymore

src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs
```diff
@@ -42,39 +40,39 @@ internal class IssueHandler
...
+                this.clientSuppressionSynchronizer = clientSuppressionSynchronizer;
                 this.onSnapshotChanged = onSnapshotChanged;
 
                 this.translateSpans = translateSpans;
             }
 
             internal /* for testing */ void HandleNewIssues(IEnumerable<IAnalysisIssueVisualization> issues)
             {
-                MarkSuppressedIssues(issues);
+                clientSuppressionSynchronizer.SynchronizeSuppressedIssues();
```
bigfluffycookie, line 75: The one thing I don't like about this is that now it will synchronize all issues, instead of just the ones that we receive here. Should we consider adding an extra method for SynchronizeSuppressedIssues that has a parameter?

src/Integration.Vsix.UnitTests/Analysis/IssueHandlerTests.cs
```diff
@@ -171,37 +160,6 @@ public void HandleNewIssues_NoIssues_NoIssuesReturned()
             notificationHandler.UpdatedSnapshot.Issues.Count().Should().Be(0);
         }
 
-        [TestMethod]
```
bigfluffycookie, line 174: We dont need to test this anymore, it doesn't really concern this class.

src/Integration.Vsix.UnitTests/Analysis/IssueHandlerTests.cs
```diff
@@ -79,30 +79,19 @@ public void HandleNewIssues_UpdatedSnapshotHasExpectedValues()
...
+            testSubject.HandleNewIssues(new List<IAnalysisIssueVisualization>());
 
             // Assert
-            issuesFilter.Verify(x => x.GetMatches(inputIssues), Times.Once);
+            clientSuppressionSynchronizer.Verify(x => x.SynchronizeSuppressedIssues(), Times.Once);
             notificationHandler.InvocationCount.Should().Be(1);
-
-            // Check the updated issues
-            notificationHandler.UpdatedSnapshot.Issues.Count().Should().Be(2);
-            notificationHandler.UpdatedSnapshot.Issues.Should().BeEquivalentTo(inputIssues);
```
bigfluffycookie, line 105: We dont need to check the issue count as we aren't changing the incoming list of issues anywhere

src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs
duncanp-sonar: > The one thing I don't like about this is that now it will synchronize all issues, instead of just the ones that we receive here

That's definitely a problem for two reasons:
1) the client synchronizer raises an event which will trigger refreshes of bits of the UI, and
2) some analyzers stream issues i.e. report one issue at a time to the `IssueConsumer`. The combination of (1) and (2) would mean for a file with 100 issue the UI refresh would be triggered and additional 100 times.


Adding an overload doesn't feel quite right either. It wouldn't need to raise the `LocalSuppressionChanged` event, so the client synchronizer would then have two methods: one that synch-ed the issues from the store and raised and event, and one that sync-ed a supplied list of issues and didn't raise an event.

There is something wrong with the existing split of responsibilities, but changing the issue handler to use the synchronizer doesn't feel quite right either. Perhaps we should discuss this on Monday.

PR #3975, opened 2023-03-24
added taint repo keys

ugras-ergun-sonarsource:


src/Core/Language.cs
```diff
@@ -126,11 +126,14 @@ public static IEnumerable<Language> KnownLanguages
         private static readonly Dictionary<string, Language> repoKeyToLanguage = new Dictionary<string, Language>()
         {
             { "csharpsquid", CSharp },
+            { "roslyn.sonaranalyzer.security.cs", CSharp },
             { "vbnet", VBNET },
```
ugras-ergun-sonarsource, line 130: There are no security rules for vb.net

georgii-borovinskikh-sonarsource (approved):

PR #3979, opened 2023-03-28
Suppress taint issues

rita-sonarsource:
Part of https://github.com/SonarSource/sonarlint-visualstudio/issues/3942


src/IssueViz/Editor/LocationTagging/IIssueLocationStore.cs
```diff
@@ -36,10 +36,16 @@ public interface IIssueLocationStore
         /// </summary>
         IEnumerable<IAnalysisIssueLocationVisualization> GetLocations(string filePath);
 
+        /// <summary>
+        /// Returns all the primary issues
+        /// </summary>
+        IEnumerable<IAnalysisIssueVisualization> Get();
```
duncanp-sonar, line 42: `GetIssues` might be a better name - more consistent with `GetLocations`, and it would be reasonable to expect a `IIssueLocationStore.Get` to return locations rather issues.

duncanp-sonar (approved):
Looks good. Just a comment about the name of the `Get` method.

PR #3981, opened 2023-03-28
Regenerate T4 template on build

### LINKED ISSUE #3335: [Infra] Auto-regenerate the list of Sonar Roslyn rules that can be suppressed on build
### Description

The list of issues C#/VB rules that can be suppress in Connected Mode is generated using a [T4 template](https://github.com/SonarSource/sonarlint-visualstudio/blob/69de0102f4155f7f45aafebbc41b0b79f2958af2/src/Roslyn.Suppressions/Roslyn.Suppressions/SupportedSuppressionBuilder.tt#L1).

At present the code is only regenerated manually i.e. we have to remember to do it each time we update the Roslyn analyzers.
We should use the T4 build integration to cause the file to be updated automatically on build.

duncanp-sonar:
Fixes #3335

### DIFF
--- a/src/Roslyn.Suppressions/Roslyn.Suppressions/SupportedSuppressionBuilder.tt
+++ b/src/Roslyn.Suppressions/Roslyn.Suppressions/SupportedSuppressionBuilder.tt
@@ -4,6 +4,8 @@
 <#@ assembly name="$(SonarCSharpDllPath)" #>
 <#@ assembly name="$(SonarVBDllPath)" #>
 <#@ assembly name="$(MSCodeAnalysisCommonDllPath)" #>
+<#@ assembly name="$(MSCodeAnalysisWorkspacesDllPath)" #>
+<#@ assembly name="$(SystemCompositionAttributedModelDllPath)" #>
 <#@ assembly name="$(NuGetPackageRoot)system.collections.immutable\1.5.0\lib\netstandard2.0\System.Collections.Immutable.dll" #>
 <#@ assembly name="netstandard" #>
 <#@ assembly name="System.Runtime" #>
@@ -64,6 +66,8 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions
                 // investigate! There may be a problem with code generator.
                 // *************************************************************************************************
 <#
+                        System.AppDomain.CurrentDomain.AssemblyResolve += CurrentDomain_AssemblyResolve;
+
                         var idList = GetFilteredDiagnosticIds();
                         WriteLine("                // Number of unique diagnostic ids (C# and VB.NET): " + idList.Count());
                         foreach(var id in idList)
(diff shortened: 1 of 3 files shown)


src/Roslyn.Suppressions/Roslyn.Suppressions/SupportedSuppressionBuilder.tt
```diff
@@ -4,7 +4,8 @@
 <#@ assembly name="$(SonarCSharpDllPath)" #>
 <#@ assembly name="$(SonarVBDllPath)" #>
 <#@ assembly name="$(MSCodeAnalysisCommonDllPath)" #>
-<#@ assembly name="$(NuGetPackageRoot)system.collections.immutable\1.5.0\lib\netstandard2.0\System.Collections.Immutable.dll" #>
+<#@ assembly name="$(MSCodeAnalysisWorkspacesDllPath)" #>
+<#@ assembly name="$(SystemCompositionAttributedModelDllPath)" #>
```
duncanp-sonar, line 8: See the note on line 100 below about why these are required when running from the command line

src/Roslyn.Suppressions/Roslyn.Suppressions/SupportedSuppressionBuilder.tt
```diff
@@ -115,24 +137,43 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions
                 .OrderBy(x => x);
         }
 
-        private IEnumerable<DiagnosticDescriptor> GetDiagnosticDescriptors(Type t)
+        private IEnumerable<string> GetIdsUsingReflection(Type t)
```
duncanp-sonar, line 140: Logically this code is the same as before; it just uses reflection instead of strong types, since at transformation the set of Roslyn assemblies being loaded is different and we get type mismatches (i.e. we have two different versions of `DiagnosticAnalyzer`, and we can't cast between them).

Risk-wise: there shouldn't really be any risk. We have unit tests that checks the generated code returns 400+ analyzer ids. If there is transformation error, the build will fail. If the transformation doesn't fail but we can't find any analyzers, the test will fail.

src/Roslyn.Suppressions/Roslyn.Suppressions/SupportedSuppressionBuilder.tt
```diff
@@ -115,24 +137,43 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions
...
-        private static readonly Type DiagnosticAnalyzerAttribute = typeof(DiagnosticAnalyzerAttribute);
+        private bool IsDiagnosticAnalyzer(Type t)
+        {
+            if (t.IsAbstract) { return false; }
+            return t.GetCustomAttributes(true).Any(x => x.GetType().Name == "DiagnosticAnalyzerAttribute");
+        }
+
+        private IEnumerable<string> GetDiagnosticIds(object analyzerInstance)
+        {
+            var propertyInfo = analyzerInstance.GetType().GetProperty("SupportedDiagnostics");
```
duncanp-sonar, line 168: Fetching the property info every time isn't efficient, but the transformation is so fast it's not worth bothering about.

Also, the transformation is only run when the input files are updated (the input files being any files that are used during the transform i.e. this .tt, and the input binaries. So the transformation will be run once when we update the analyzer version).

src/Roslyn.Suppressions/Roslyn.Suppressions/SuppressorCodeGen.targets
```diff
@@ -17,6 +18,10 @@
     <!-- Reference the Sonar analyzers we want to import, but don't include them as analyzers for this project -->
     <PackageReference Include="SonarAnalyzer.CSharp" Version="$(EmbeddedSonarAnalyzerVersion)" ExcludeAssets="all" PrivateAssets="all" GeneratePathProperty="true" />
     <PackageReference Include="SonarAnalyzer.VisualBasic" Version="$(EmbeddedSonarAnalyzerVersion)" ExcludeAssets="all" PrivateAssets="all" GeneratePathProperty="true" />
+
+    <!-- NuGet packages required during the T4 transformation -->
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="3.3.1" PrivateAssets="all" IncludeAssets="none" GeneratePathProperty="true" />
+    <PackageReference Include="System.Composition.AttributedModel" Version="1.0.31" PrivateAssets="all" IncludeAssets="none" GeneratePathProperty="true" />
```
duncanp-sonar, line 24: The packages are marked with `IncludeAssets="none"` so they won't be referenced transitively by any dependant projects.
I've checked the set of assemblies included in the VSIX has not changed.

rita-sonarsource (approved):

PR #3984, opened 2023-03-28
add rich vs simple layout picker

georgii-borovinskikh-sonarsource:


src/Rules/RuleInfoExtensions.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace SonarLint.VisualStudio.Rules
+{
+    public static class RuleInfoExtensions
```
ugras-ergun-sonarsource, line 3: Why create a new class instead of using RuleExtensions class?

georgii-borovinskikh-sonarsource: Should have been named RuleInfoExtensions and not RuleExtensions. That's why I didn't see there was a class for that already :-P

ugras-ergun-sonarsource: it's not just about RuleInfo that's why it was not RuleInfoExtensions. I wrote extensions for 3 different classes and I didn't want to create a class for each extension so I have aggregated them in RuleExtensions.

ugras-ergun-sonarsource (approved):

PR #3985, opened 2023-03-28
Fetch suppressed taint issues

rita-sonarsource:


src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -189,14 +189,21 @@ private void UpdateServerType()
...
                 return false;
             }
 
-            var allFilePaths = ((ITaintIssueViewModel) viewModel).TaintIssueViz.GetAllLocations()
+            var issueViz = ((ITaintIssueViewModel) viewModel).TaintIssueViz;
+
+            if (issueViz.IsSuppressed)
+            {
+                return false;
+            }
```
rita-sonarsource, line 204: This current change is not reactive; if SSE changes it, the list won't auto-refresh until document changed event is raised. I will fix it in the next PR.

src/SonarQube.Client.Tests/SonarQubeService_GetTaintVulnerabilitiesRequest.cs
```diff
@@ -54,7 +54,7 @@ public async Task GetTaintVulnerabilitiesAsync_Response_From_SonarQube()
         {
             await ConnectToSonarQube("8.6.0.0");
 
-            SetupRequest("api/issues/search?projects=shared&statuses=OPEN%2CCONFIRMED%2CREOPENED&types=VULNERABILITY&p=1&ps=500", @"
+            SetupRequest("api/issues/search?projects=shared&statuses=OPEN%2CCONFIRMED%2CREOPENED%2CRESOLVED&types=VULNERABILITY&p=1&ps=500", @"
```
rita-sonarsource, line 57: I tried passing no statuses at all, and bizarrely I then get back more issues with corrupted info that I cannot fetch the source code for them. This is what we use to fetch suppressions normally so I added it, and that works.

src/IssueViz.Security.UnitTests/Taint/TaintIssueToIssueVisualizationConverterTests.cs
```diff
@@ -288,6 +288,48 @@ public void Convert_FromTaintIssue_IssueVizConverterCalledWithCorrectParameters_
...
+
+            var testSubject = CreateTestSubject(issueVizConverter.Object);
+            var result = testSubject.Convert(issue);
+
+            result.Should().Be(expectedConvertedIssueViz);
+            result.IsSuppressed.Should().Be(isIssueSuppressed);
+        }
+
+        [TestMethod]
+        public void Convert_FromTaintIssue_IssueVizIsNotSuppressed()
```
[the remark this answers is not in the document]
duncanp-sonar: If I remember what Julien said correctly, the taint server events are triggered as a result of new analysis results being uploaded to the server. So we get taint issues added or removed - the user won't have had the opportunity to mark an issue as suppressed (although I wonder what would happen if the analysis was the result of a PR being merged, if the user had marked an issue as "won't fix" in the PR...).

The issue status change events are a result of the user making status changes in the UI.

duncanp-sonar (approved):

Issue #3718, opened 2023-02-15
SSESession is not restarted when stream request fails


rita-sonarsource:
### Description

If the request stream suddenly dies, SSESession is terminated but a new one is not created until the user closes and reopens the solution.

Prototype [PR here](https://github.com/SonarSource/sonarqube-webclient-dotnet/pull/161)

#### Sample call stack

```
[ThreadId 60] [DEBUG] [SSESession] Failed to handle events: System.IO.IOException: The read operation failed, see inner exception. ---> System.Net.WebException: The request was aborted: The request was canceled.
   at System.Net.ConnectStream.Read(Byte[] buffer, Int32 offset, Int32 size)
   at System.Net.Http.HttpClientHandler.WebExceptionWrapperStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   --- End of inner exception stack trace ---
   at System.Net.Http.HttpClientHandler.WebExceptionWrapperStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.StreamReader.ReadBuffer()
   at System.IO.StreamReader.get_EndOfStream()
   at SonarQube.Client.Models.ServerSentEvents.ServerContract.SqSSEStreamReader.<ReadAsync>d__5.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at SonarQube.Client.Models.ServerSentEvents.SSEStreamReader.<ReadNextEventAsync>d__5.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at SonarQube.Client.Models.ServerSentEvents.SSEStreamReader.<ReadAsync>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at SonarLint.VisualStudio.ConnectedMode.ServerSentEvents.SSESessionFactory.SSESession.<PumpAllAsync>d__9.MoveNext()
```

PR #3986, opened 2023-03-29
Deal with sse session restart

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3718


src/SonarQube.Client/ISSEConnectionFactory.cs
```diff
@@ -0,0 +1,31 @@
+﻿/*
```
bigfluffycookie, line 1: Copied directly from the prototype PR -> https://github.com/SonarSource/sonarqube-webclient-dotnet/pull/161

PR #3998, opened 2023-03-29
Format json when serializing

### LINKED ISSUE #3990: [Infra] Serialize json with formatting
### Description

In most places we write formatted json to files, which makes it easier to compare when debugging or handling support calls.

There are a couple of places were we don't format the json, which makes it harder to work with e.g. `ExclusionSettingsStorage.cs` and `RoslynSettingsFileStorage.cs`. See the screenshots below for a comparison.

The fix is trivial - add `Formatting.Indented` e.g. 

```
var fileContent = JsonConvert.SerializeObject(settings, Formatting.Indented);
```

#### Without formatting
image

#### With formatting
image

duncanp-sonar:
Fixes #3990

### DIFF
--- a/src/Integration/Exclusions/ExclusionSettingsStorage.cs
+++ b/src/Integration/Exclusions/ExclusionSettingsStorage.cs
@@ -66,7 +66,7 @@ namespace SonarLint.VisualStudio.Integration.Exclusions
                 throw new InvalidOperationException("Cannot save exclusions in Standalone mode.");
             }
 
-            var fileContent = JsonConvert.SerializeObject(settings);
+            var fileContent = JsonConvert.SerializeObject(settings, Formatting.Indented);
             var exclusionsFilePath = GetFilePath(bindingConfiguration);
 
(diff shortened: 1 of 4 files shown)


src/Integration/Exclusions/ExclusionSettingsStorage.cs
```diff
@@ -66,7 +66,7 @@ public void SaveSettings(ServerExclusions settings)
                 throw new InvalidOperationException("Cannot save exclusions in Standalone mode.");
             }
 
-            var fileContent = JsonConvert.SerializeObject(settings);
+            var fileContent = JsonConvert.SerializeObject(settings, Formatting.Indented);
```
duncanp-sonar, line 69: Minor tweak - I was looking at how Roslyn suppressions are handled, and found the un-indented json a pain to work with. Most of the time we indent it. This PR changes the two places we write un-indented code.

rita-sonarsource (approved):

PR #3999, opened 2023-03-30
Restart SSESession when failure occurs

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3718

### DIFF
--- a/src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionTests.cs
+++ b/src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionTests.cs
@@ -194,11 +194,15 @@ public class SSESessionTests
     }
 
     [TestMethod]
-    public void Dispose_InvokesOnSessionFailedAsync()
+    public async Task NonCriticalException_InvokesOnSessionFailedAsync()
     {
         var testScope = new TestScope();
         testScope.SetUpSwitchToBackgroundThread();
-        testScope.TestSubject.Dispose();
+
+        var sseStreamMock = testScope.SetUpSQServiceToSuccessfullyReturnSSEStreamReader();
+        sseStreamMock.Setup(x => x.ReadAsync()).Throws(new NotImplementedException("this is a test"));
+
+        await testScope.TestSubject.PumpAllAsync();
 
         testScope.OnSessionFailedAsyncMock.Invocations.Should().HaveCount(1);
     }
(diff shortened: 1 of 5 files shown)


src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionTests.cs
```diff
@@ -193,8 +193,25 @@ public async Task Dispose_FinishesSession()
...
+        var testScope = new TestScope();
+        testScope.SetUpSwitchToBackgroundThread();
+        testScope.TestSubject.Dispose();
+
+        testScope.OnSessionFailedAsyncMock.Invocations.Should().HaveCount(1);
+    }
+
     public interface IDummyServerEvent : IServerEvent { }
 
+    private static Mock<OnSessionFailedAsync> GetOnSessionFailedAsyncMock()
```
bigfluffycookie, line 208: Not sure if this is the right way to do it but its the only way I could figure out :)

src/ConnectedMode/ServerSentEvents/SSESessionFactory.cs
```diff
@@ -200,6 +207,8 @@ public void Dispose()
                     return;
                 }
 
+                onSessionFailedCallback(this).Forget();
```
duncanp-sonar, line 210: It seems a bit odd to be calling "on failed" from within `Dispose`.
Presumably there are non-failure cases in which the session will be disposed/

It might just be a naming issue - it might be less confusing if the delegate were called `OnSessionDisposedCallback`.

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionTests.cs
duncanp-sonar: It looks like the setup is a no-op; it would be needed if the mock is being created using `strict`.

src/ConnectedMode/ServerSentEvents/SSESessionFactory.cs
bigfluffycookie: Hmm I see what you mean, now that I think about it wouldn't it instead make more sense to move the callback handling into the catch block? Do we actually want it to happen in the dispose?

src/ConnectedMode.UnitTests/ServerSentEvents/SSESessionTests.cs
```diff
@@ -193,8 +193,25 @@ public async Task Dispose_FinishesSession()
         sessionToken.Value.IsCancellationRequested.Should().BeTrue();
     }
 
+    [TestMethod]
+    public void Dispose_InvokesOnSessionFailedAsync()
```
duncanp-sonar, line 197: This seems weird - the session should be disposed e.g. when a bound solution is closed. Do we want to log a failure message and try to create another session in that case?

bigfluffycookie: I asked that in a comment above :)

src/ConnectedMode/ServerSentEvents/SSESessionFactory.cs
rita-sonarsource: @bigfluffycookie I think originally it was in the try-catch, and for some reason I moved it to the dispose. Was a mistake.

PR #4000, opened 2023-03-30
Remove IssuesFilter

### LINKED ISSUE #3944: Consider dropping issue filtering from the issue consumer

rita-gorokhod:
Fixes #3944

### DIFF
--- a/src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs
+++ b/src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs
@@ -20,12 +20,9 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using Microsoft.VisualStudio.Text;
 using SonarLint.VisualStudio.ConnectedMode.Suppressions;
-using SonarLint.VisualStudio.Core.Analysis;
-using SonarLint.VisualStudio.Core.Suppressions;
 using SonarLint.VisualStudio.IssueVisualization.Models;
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs
```diff
@@ -89,7 +89,10 @@ private void MarkSuppressedIssues(IEnumerable<IAnalysisIssueVisualization> issue
             {
                 var filterableIssues = issues.OfType<IFilterableIssue>().ToArray();
 
-                var matches = issuesFilter.GetMatches(filterableIssues);
+                var matches = filterableIssues
```
duncanp-sonar, line 92: Can't we just do the check inside the `foreach` in line 98+ and get rid of the `matches` list?

i.e. `issue.IsSuppressed = suppressedIssueMatcher.SuppressionExists(issue)`

duncanp-sonar (commented):
Looks good. I'm not sure we still need the intermediate `matches` lists, though.

src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs
rita-sonarsource: Oops, yeah that is simpler :D

PR #4002, opened 2023-03-30
Handle dispose exception when closing VS

rita-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3946


src/ConnectedMode/ServerSentEvents/Issue/IssueServerEventsListener.cs
```diff
@@ -103,10 +103,18 @@ public async Task ListenAsync()
             }
         }
 
+        private bool disposed;
+
         public void Dispose()
         {
+            if (disposed)
```
rita-sonarsource, line 110: It's called once by us, when disposing the ConnectedMode package, and a second time when VS decides to dispose MEF components.
I think we don't actually need to call dispose on stuff that are MEF components, since VS seems to call it for us, but I left it for now.

duncanp-sonar (approved):

Issue #3093, opened 2022-06-28
Apply server settings in connected mode for roslyn languages


rita-sonarsource:
### Description

#1335 added support inclusions/exclusions for non-roslyn languages (JS/TS/C/C++). 
This ticket is for adding the same functionality for roslyn languages.

Related: [MMF-1230](https://jira.sonarsource.com/browse/MMF-1230)

PR #4014, opened 2023-03-31
Write server inclusions/exclusions to SonarLint.xml

rita-sonarsource:
Fixes [#3093](https://github.com/SonarSource/sonarlint-visualstudio/issues/3093)

### DIFF
--- a/src/Integration.UnitTests/Exclusions/AnalyzableFileIndicatorTests.cs
+++ b/src/Integration.UnitTests/Exclusions/AnalyzableFileIndicatorTests.cs
@@ -65,8 +65,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Exclusions
             bool globalExclusionsApply,
             bool expectedResult)
         {
-            var projectExclusions = new[] { "exclusion1", "exclusion2" };
-            var globalExclusions = new[] { "exclusion3", "exclusion4" };
+            var projectExclusions = new[] { "**/exclusion1", "**/exclusion2" };
+            var globalExclusions = new[] { "**/exclusion3", "**/exclusion4" };
             var exclusionConfig = CreateServerExclusions(
                 inclusions: null,
                 exclusions: projectExclusions,
@@ -90,7 +90,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Exclusions
         [DataRow(false, false)]
         public void ShouldAnalyze_HasInclusions_NoExclusions_ReturnsIfIncluded(bool inclusionsApply, bool expectedResult)
         {
-            var inclusions = new[] { "inclusion1", "inclusion2" };
+            var inclusions = new[] { "**/inclusion1", "**/inclusion2" };
             var exclusionConfig = CreateServerExclusions(
                 inclusions: inclusions, 
                 exclusions: null,
@@ -119,9 +119,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Exclusions
             bool globalExclusionsApply,
             bool expectedResult)
         {
-            var inclusions = new[] { "inclusion1", "inclusion2" };
-            var projectExclusions = new[] { "exclusion1", "exclusion2" };
-            var globalExclusions = new[] { "exclusion3", "exclusion4" };
+            var inclusions = new[] { "**/inclusion1", "**/inclusion2" };
+            var projectExclusions = new[] { "**/exclusion1", "**/exclusion2" };
+            var globalExclusions = new[] { "**/exclusion3", "**/exclusion4" };
             var exclusionConfig = CreateServerExclusions(
                 inclusions: inclusions,
                 exclusions: projectExclusions,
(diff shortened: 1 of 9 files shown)


src/Integration/Binding/CSharpVBBindingConfigProvider.cs
```diff
@@ -109,13 +109,23 @@ private async Task<IBindingConfig> DoGetConfigurationAsync(SonarQubeQualityProfi
...
+            var exclusions = await FetchInclusionsExclusionsAsync(bindingConfiguration.Project.ProjectKey, cancellationToken);
 
             var ruleset = GetRulesetFile(qualityProfile, language, bindingConfiguration, activeRules, inactiveRules, sonarProperties);
-            var additionalFile = GetAdditionalFile(language, bindingConfiguration, activeRules, sonarProperties);
+            var additionalFile = GetAdditionalFile(language, bindingConfiguration, activeRules, sonarProperties, exclusions);
 
             return new CSharpVBBindingConfig(ruleset, additionalFile);
         }
 
+        private async Task<ServerExclusions> FetchInclusionsExclusionsAsync(string projectKey,
```
rita-sonarsource, line 120: We do this exact call in BindingProcessImpl, to generate sonar.settings.json.
I think since they're different format and files, it's simpler to just re-fetch it.

src/SonarQube.Client.Tests/Models/ServerExclusionsTests.cs
```diff
@@ -112,5 +114,120 @@ public void Equals_SameReference_True()
...
+                globalExclusions: new[] { "path1", "**\\path2", "**/path3", "**path4", "*/*" },
+                inclusions: Array.Empty<string>());
+
+            testSubject.Exclusions.Should().BeEmpty();
+            testSubject.GlobalExclusions.Should().BeEquivalentTo(
+                "**\\path1",
+                "**\\path2",
+                "**/path3",
+                "**\\**path4",
+                "**\\*/*");
```
rita-sonarsource, line 150: I asked Caba about this, and they do it per char at a time, so it doesn't matter if some are forward and some are back.
https://github.com/SonarSource/sonar-dotnet/blob/a994db89dfb6136f75865ebc66306a19e5f23e76/analyzers/src/SonarAnalyzer.Common/Helpers/WildcardPatternMatcher.cs#L87

src/Integration.UnitTests/Exclusions/AnalyzableFileIndicatorTests.cs
```diff
@@ -65,8 +65,8 @@ public void ShouldAnalyze_NoExclusionSettings_True()
             bool globalExclusionsApply,
             bool expectedResult)
         {
-            var projectExclusions = new[] { "exclusion1", "exclusion2" };
-            var globalExclusions = new[] { "exclusion3", "exclusion4" };
+            var projectExclusions = new[] { "**/exclusion1", "**/exclusion2" };
```
rita-sonarsource, line 68: I really should've created IServerExclusions and mocked it, but I didn't think about it until later...

src/SonarQube.Client/Models/ServerExclusions.cs
```diff
@@ -32,15 +35,21 @@ public ServerExclusions()
...
+            Exclusions = AddPathPrefixIfNeeded(exclusions) ?? EmptyValues;
+            GlobalExclusions = AddPathPrefixIfNeeded(globalExclusions) ?? EmptyValues;
+            Inclusions = AddPathPrefixIfNeeded(inclusions) ?? EmptyValues;
         }
 
+        /// <summary>
+        /// Similarly to the other SL flavors, we will prefix everything with "**/" it's not already prefixed.
+        /// </summary>
+        private static string[] AddPathPrefixIfNeeded(IEnumerable<string> paths) =>
+            paths?.Select(path => path.StartsWith("**/") || path.StartsWith("**\\") ? path : Path.Combine("**", path)).ToArray();
```
duncanp-sonar, line 51: I think we should prefix with `**/` rather than `**\` - it will create less confusion for users if we stick to the syntax specified in the documentation.

src/SonarQube.Client.Tests/Models/ServerExclusionsTests.cs
duncanp-sonar: Thinking about it again, it seems a bit weird that they support syntax that isn't supported by SonarQube/SonarCloud. I think we should stick to adding only documented syntax.

src/Integration/Binding/CSharpVBBindingConfigProvider.cs
duncanp-sonar: We've actually already fetch all of the properties for the project in line 101:
```
var sonarProperties = await FetchPropertiesAsync(bindingConfiguration.Project.ProjectKey, cancellationToken);
```

... which include the `sonar.exclusions` etc. It's something we could perhaps simplify in the future.

duncanp-sonar (approved):

PR #4017, opened 2023-04-03
Fix Roslyn suppression path compression bug

### LINKED ISSUE #3997: Diagnostic suppressor incorrectly matches relative server paths to local absolute paths
### Description

The file-matching logic isn't completely correctly. It's supposed to tail-match the relative server path to the local absolute path

e.g. given a server path of  `aaa\foo.txt`, it should match all of the following:
* `c:\aaa\foo.txt`
* `c:\bbb\aaa\foo.txt`
* `c:\ccc\bbb\aaa\foo.txt`
* `d:\aaa\foo.txt`

However, it should not match e.g. `c:\XXXaaa\foo.txt`, but it does.

Related: the equivalent code for matching non-Roslyn issues should behave the same. See #3991.

duncanp-sonar:
Fixes #3997

### DIFF
--- a/src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressionCheckerTests.cs
+++ b/src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressionCheckerTests.cs
@@ -20,10 +20,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using FluentAssertions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Text;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.Text.Editor.Commanding.Commands;
 using Moq;
 using SonarLint.VisualStudio.Roslyn.Suppressions.Settings.Cache;
 using SonarQube.Client;
@@ -104,7 +106,7 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions.UnitTests
 
             var suppressedIssues = new SuppressedIssue[]
             {
-                CreateIssue("S111", "c:\\wrongFile1.txt")
+                CreateIssue("S111", "wrongFile1.txt")
             };
 
             var cache = CreateSettingsCache("settingsKey", suppressedIssues);
@@ -150,7 +152,7 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions.UnitTests
             // Sanity check the diagnostic location is on the expected line
             diag.Location.GetLineSpan().StartLinePosition.Line.Should().Be(0, "Test setup error");
 
-            var issue = CreateIssue("S999", "c:\\issueFile.cs", 1, "hash");
+            var issue = CreateIssue("S999", "issueFile.cs", 1, "hash");
 
             SuppressionChecker.IsMatch(diag, issue, checksumCalculator.Object)
                 .Should().BeFalse();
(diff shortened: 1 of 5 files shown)


src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressionCheckerTests.cs
```diff
@@ -104,7 +106,7 @@ public void IsSuppressed_HasIssues_NoMatches_ReturnsFalse()
 
             var suppressedIssues = new SuppressedIssue[]
             {
-                CreateIssue("S111", "c:\\wrongFile1.txt")
+                CreateIssue("S111", "wrongFile1.txt")
```
duncanp-sonar, line 109: The product code change is simple. However, the existing tests needed a bit more fixing up because the product code is now stricter and contains multiple assertions about the input values i.e. the local path should be absolute and the server issue path should be relative.

The main required change was to update the server paths set in the tests to be relative.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressionCheckerTests.cs
```diff
@@ -181,6 +183,7 @@ public void IsMatch_RuleIdsAreDifferent_ReturnsFalse()
         // Constants used in for the diagnostic to match against in IsMatch_ReturnsExpectedValue
         // They are defined as constants here so we can use them in the [DataRow]s for the test.
         private const string DiagFileName = "c:\\diag.txt";
+        private const string MatchingServerFileName = "diag.txt";
```
duncanp-sonar, line 186: Relative path that should match `DiagFileName`.

src/Roslyn.Suppressions/Roslyn.Suppressions.UnitTests/SuppressionCheckerTests.cs
```diff
@@ -314,12 +320,25 @@ private static Mock<IChecksumCalculator> CreateChecksumCalculator(string hashToR
...
+            // The matching server should not be rooted, so we'll strip off the
+            // root and return the rest of the path. The product code should be
+            // considered that to be a valid match for the original full path.
+            if (!Path.IsPathRooted(absolutePath))
+            {
+                throw new ArgumentException(nameof(absolutePath), $"Test setup error: path should be rooted. Actual: {absolutePath}");
+            }
+            var root = Path.GetPathRoot(absolutePath);
+            return absolutePath.Substring(root.Length);
+        }
```
duncanp-sonar, line 341: The `CreateIssueFromDiagnostic` method is a test helper method that creates a matching server issue from a test diagnostic. Previously it was ok if the paths were the same; now the server path has to be relative.

src/Roslyn.Suppressions/Roslyn.Suppressions/SuppressionChecker.cs
```diff
@@ -117,8 +118,7 @@ public bool IsSuppressed(Diagnostic reportedDiagnostic, string settingsKey)
         }
 
         private static bool IsSameFile(Diagnostic diagnostic, SuppressedIssue suppressedIssue) =>
-            diagnostic.Location.SourceTree?.FilePath.EndsWith(suppressedIssue.FilePath, StringComparison.OrdinalIgnoreCase)
```
duncanp-sonar, line 120: The old version of the code would incorrectly match `aaa\file.txt` with `c:\XXXaaa\file.txt`.
The new `IsServerFileMatch` handles this case correctly.

bigfluffycookie (approved):
Very good!

Issue #4011, opened 2023-03-31
Create rule cross-reference parser


duncanp-sonar:
### Description

Text descriptions in rule help can include Sonar-specific markup to indicate a cross-reference to another Sonar rule.

The format is `{rule:[repository id]:[rule id]}`.
Or to put it more simply `{rule:[composite rule key]}`

e.g. `{rule:c:S131}`

The parser should accept a text string and return a list of chunks, where each chunk is either a block of text or a rule cross-reference.

PR #4018, opened 2023-04-03
Add a rule cross-reference parser

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4011


src/Education.UnitTests/XamlGenerator/RawTextProcessorTests.cs
```diff
@@ -0,0 +1,96 @@
+﻿using FluentAssertions;
```
bigfluffycookie, line 1: Forgot copyright, will add it with any other necessary changes (or standalone if there are none.)

src/Education/XamlGenerator/RawTextProcessor.cs
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text.RegularExpressions;
+
```
bigfluffycookie, line 24: Lots of this class was stolen from this pr -> https://github.com/SonarSource/sonarlint-visualstudio/commit/e6a0cd5089019cf2bc2e2fc0eca8ebc2deb78c42

I changed it up to instead return a list

src/Education/XamlGenerator/RawTextProcessor.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace SonarLint.VisualStudio.Education.XamlGenerator
+{
+    /// <summary>
+    /// Processes special internal markup embedded in normal text
+    /// </summary>
+    internal class RawTextProcessor
```
bigfluffycookie, line 30: I don't think I like the name very much

src/Education/XamlGenerator/RawTextProcessor.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    public interface IRawTextItem
+    {
+        string Text { get; }
+        bool IsCrossReference { get; }
+    }
+
+    public class RawTextItem : IRawTextItem
```
bigfluffycookie, line 81: Same with this, does RawTextItem even mean anything? Maybe just TextItem? I dont know

src/Education/XamlGenerator/RawTextProcessor.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            foreach (Match match in matches)
+            {
+                Debug.Assert(match.Success, "Expecting match to have succeeded");
+                var matchGroup = match.Groups[0];
+
+                var simpleText = rawText.Substring(endOfLastMatch, matchGroup.Index - endOfLastMatch);
+                AddSimpleText(rawTextItems, simpleText);
+
+                rawTextItems.Add(new RawTextItem(matchGroup.Value, true));
```
bigfluffycookie, line 53: The matchGroup.Value will simply be -> "{rule:cpp:S123}"

src/Education/XamlGenerator/RawTextProcessor.cs
duncanp-sonar: Don't return concrete types directly if there is an available interface i.e. return `IList` rather than `List`.

Also, use the simplest data structure possible that has the functionality required by the caller.
For example, an `IList` is writable, so returning `IList` implies it's ok for the caller to change the contents. If the result should be readonly, we can use `IReadOnlyList<>`.
Also, lists are indexable i.e. the caller can access `result[1]`. 

Here, we want the returned data to be read-only, so `IReadOnlyList<>` is an option.
The caller only really needs to iterate through the list once from the start, so `IEnumerable<>` would also do since indexing is not required.

src/Education/XamlGenerator/RawTextProcessor.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    /// <summary>
+    /// Processes special internal markup embedded in normal text
+    /// </summary>
+    internal class RawTextProcessor
+    {
+        /// <summary>
+        /// Processes a text string that might/might not contain rule cross-references
+        /// </summary>
+        public static List<IRawTextItem> ProcessRawText(string rawText)
```
duncanp-sonar, line 35: `Parse` might be a better name now, given the changed responsibilities.

src/Education/XamlGenerator/RawTextProcessor.cs
duncanp-sonar: Is the caller going to be able to use this value directly, or are they going to have to do some more parsing to turn it into a usable form? e.g. stripping off the curly brackets, and perhaps extracting the composite rule id?

Ideally, the caller just wants to know that this is a cross reference and what the referenced composite rule id is without having to do any extra work.

Looking at it another way, the parser is the only thing that should need to know exactly how the cross-references are encoded in the raw text. If the format of that encoding changed (to e.g. `[CROSSREF repo=cs;rule=S123]`), only the parser should need to change.

src/Education.UnitTests/XamlGenerator/RawTextProcessorTests.cs
```diff
@@ -0,0 +1,96 @@
...
+    public class RawTextProcessorTests
+    {
+        [TestMethod]
+        public void ProcessRawText_OneCrossReference_WithSurroundingText_ReturnsListWithCorrectItems()
+        {
+            string text = " {rule:javascript:S1656} - Implements a check on";
+
+            var list = RawTextProcessor.ProcessRawText(text);
+
+            list.Count.Should().Be(3);
```
duncanp-sonar, line 17: Minor: there's a FluentAssertion for collections `Should().HaveCount(x)`.

I think there is a diagnostic analyzer that complains if we don't use it.

src/Education/XamlGenerator/RawTextProcessor.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace SonarLint.VisualStudio.Education.XamlGenerator
+{
+    /// <summary>
+    /// Processes special internal markup embedded in normal text
+    /// </summary>
+    internal class RawTextProcessor
+    {
+        /// <summary>
+        /// Processes a text string that might/might not contain rule cross-references
+        /// </summary>
```
duncanp-sonar, line 34: Perhaps "Parses a text string..." rather than "Processes a text string..."

src/Education/XamlGenerator/RawTextProcessor.cs
bigfluffycookie: Hm those are good points, I was thinking the parse doesn't know what to actually do with the information so it should leave it whole but the way you explain it makes more sense

src/Education/XamlGenerator/CustomTextMarkupParser.cs
```diff
@@ -0,0 +1,114 @@
...
+ *   The SonarQube and SonarCloud suppport a special embedded syntax to indicate
+ *   cross-references between links
+ *   
+ *   e.g. https://sonarcloud.io/organizations/sonarsource/rules?languages=cs&open=csharpsquid%3AS4002&q=S4002
+ * 
+ *   The cross-references are embedded in the rule description text as follows:
+ *      {rule:[repoKey]:[ruleKey]}
+ * 
+ *   This class parses simple text strings looking for the embedded syntax and
+ *   calls the appropriate callback so the caller can render the appropriate output.
```
duncanp-sonar, line 37: The comment about the callback applied to the prototype but not this class (I thought I'd cut that bit out).

src/Education/XamlGenerator/CustomTextMarkupParser.cs
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+namespace SonarLint.VisualStudio.Education.XamlGenerator
+{
+    /// <summary>
+    /// Processes special internal markup embedded in normal text
+    /// </summary>
+    internal static class CustomTextMarkupParser
```
duncanp-sonar, line 45: We'll probably need to add an interface for this to make it easier to test the classes that use it.
However, that can be done in a later PR.

src/Education/XamlGenerator/CustomTextMarkupParser.cs
```diff
@@ -0,0 +1,114 @@
...
+    internal interface ISimpleText : ITextToken { string Text { get; } }
+    internal interface IRuleCrossRef : ITextToken { SonarCompositeRuleId CompositeRuleId { get; } }
+
+    internal class SimpleText : ISimpleText
+    {
+        public string Text { get; }
+
+        public SimpleText(string text)
+        {
+            Text = text;
```
duncanp-sonar, line 101: We could throw if the `text` is null/empty. Same for the `repoKey` and `ruleKey` in the `RuleCrossRef` class. That way we can be sure that the created objects will always be valid when we consume them later.

src/Education/XamlGenerator/CustomTextMarkupParser.cs
```diff
@@ -0,0 +1,114 @@
...
+            var matches = Regex.Matches(text, "{rule:(?<repoKey>[A-Za-z]+):(?<ruleKey>[\\w]+)}");
+
+            int endOfLastMatch = 0;
+
+            foreach (Match match in matches)
+            {
+                Debug.Assert(match.Success, "Expecting match to have succeeded");
+                var matchGroup = match.Groups[0];
+
+                var simpleText = text.Substring(endOfLastMatch, matchGroup.Index - endOfLastMatch);
```
duncanp-sonar, line 64: It took me a while to work out what this code is doing and why it works (even though I wrote the prototype). A couple of comments might help anyone looking at it in the future.

e.g. 
```
// Each regex match is a an embedded rule cross reference. There might be simple text before the current group and the previous group, so we need to extract that too.
```

src/Education/XamlGenerator/CustomTextMarkupParser.cs
```diff
@@ -0,0 +1,114 @@
...
+                AddSimpleText(textTokens, simpleText);
+
+                var repoKey = match.Groups["repoKey"].Value;
+                var ruleKey = match.Groups["ruleKey"].Value;
+                textTokens.Add(new RuleCrossRef(repoKey, ruleKey));
+
+                endOfLastMatch = matchGroup.Index + matchGroup.Length;
+            }
+
+            var endText = text.Substring(endOfLastMatch);
```
duncanp-sonar, line 74:
```
// Capture any simple text after the last rule cross reference.
```

src/Education.UnitTests/XamlGenerator/CustomTextMarkupParserTests.cs
```diff
@@ -0,0 +1,164 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarLint.VisualStudio.Education.XamlGenerator;
+
+namespace SonarLint.VisualStudio.Education.UnitTests
+{
+    [TestClass]
+    public class CustomTextMarkupParserTests
+    {
+        [TestMethod]
+        public void ProcessRawText_Mulitline_OneCrossReference_WithSurroundingText_ReturnsResultWithCorrectItems()
```
duncanp-sonar, line 32: Typo - `Mulit`

duncanp-sonar (approved):
Looks good. Just a suggestion to add a couple of comments.

src/Education/XamlGenerator/CustomTextMarkupParser.cs
bigfluffycookie: You did cut it out but for some reason I copied it directly from the prototype pr instead of your comment

src/Education/XamlGenerator/CustomTextMarkupParser.cs
```diff
@@ -98,6 +98,11 @@ internal class SimpleText : ISimpleText
 
         public SimpleText(string text)
         {
+            if (string.IsNullOrEmpty(text))
+            {
+                throw new ArgumentNullException("text");
```
duncanp-sonar, line 103: Use `nameof` where possible, rather the name of the argument as a string.


Also, there is a short-form for this using `??`:

```
Text = text ?? throw new ArgumentNullException(nameof(text))
```

PR #4019, opened 2023-04-03
Support DescriptionSections and EducationPrinciples in SQ Client with versioning

### LINKED ISSUE #4009: Support DescriptionSections and EducationPrinciples in SQ Client with versioning.

ugras-ergun-sonarsource:
fixes #4009

### DIFF
--- a/src/SonarQube.Client.Tests/Requests/DefaultConfiguration_Configure.cs
+++ b/src/SonarQube.Client.Tests/Requests/DefaultConfiguration_Configure.cs
@@ -25,7 +25,6 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarQube.Client.Api;
 using SonarQube.Client.Requests;
 using SonarQube.Client.Tests.Infra;
-using static Google.Protobuf.WellKnownTypes.Field.Types;
 
 namespace SonarQube.Client.Tests.Requests
 {
@@ -68,6 +67,8 @@ namespace SonarQube.Client.Tests.Requests
                     "Registered SonarQube.Client.Api.V8_6.GetTaintVulnerabilitiesRequest for 8.6",
                     "Registered SonarQube.Client.Api.V7_20.GetExclusionsRequest for 7.2",
                     "Registered SonarQube.Client.Api.V9_4.GetSonarLintEventStream for 9.4",
+                    "Registered SonarQube.Client.Api.V9_5.GetRulesWithDescriptionSectionsRequest for 9.5",
+                    "Registered SonarQube.Client.Api.V9_6.GetRulesWithEducationPrinciplesRequest for 9.6"
                 };
 
             DefaultConfiguration.ConfigureSonarQube(new RequestFactory(logger));
@@ -90,7 +91,7 @@ namespace SonarQube.Client.Tests.Requests
                     "Registered SonarQube.Client.Api.V5_00.GetSourceCodeRequest",
                     "Registered SonarQube.Client.Api.V5_10.GetLanguagesRequest",
                     "Registered SonarQube.Client.Api.V5_40.GetModulesRequest",
-                    "Registered SonarQube.Client.Api.V5_50.GetRulesRequest",
+                    "Registered SonarQube.Client.Api.V9_6.GetRulesWithEducationPrinciplesRequest",
                     "Registered SonarQube.Client.Api.V5_50.DownloadStaticFile",
                     "Registered SonarQube.Client.Api.V6_20.GetProjectsRequest",
                     "Registered SonarQube.Client.Api.V6_30.GetPluginsRequest",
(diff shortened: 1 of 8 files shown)


ugras-ergun-sonarsource:
Analysis failed due to code duplication but for versioning, we need to do this.

src/SonarQube.Client/Api/V5_50/GetRulesRequest.cs
```diff
@@ -44,8 +44,13 @@ public class GetRulesRequest : PagedRequestBase<SonarQubeRule>, IGetRulesRequest
         // that the field names here do not always correspond to the actual field names
         // in the response! For example 'internalKey' in the request corresponds to 'key'
         // in the response. The server error message (400) returns all supported fields.
+        // Also Make sure the field is supported in this version of the API.
+        // If not add a new request for API version that supports. e.g. "GetRulesWithDescriptionSectionsRequest"
+        [JsonIgnore]
+        public IList<string> ResponseFieldsList { get; set; } = new List<string> { "repo", "internalKey", "params", "actives", "htmlDesc", "tags", "name", "htmlNote" };
```
georgii-borovinskikh-sonarsource, line 50: the list can be made static since it's the same for every instance of the request

src/SonarQube.Client/Api/V9_5/GetRulesWithDescriptionSectionsRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+        public IList<string> ResponseFieldsList { get; set; }
+
+        public async Task<SonarQubeRule[]> InvokeAsync(HttpClient httpClient, CancellationToken token)
+        {
+            innerRequest.IsActive = this.IsActive;
+            innerRequest.QualityProfileKey = this.QualityProfileKey;
+            innerRequest.RuleKey = this.RuleKey;
+            innerRequest.Logger = this.Logger;
+
+            innerRequest.ResponseFieldsList.Add("descriptionSections");
```
georgii-borovinskikh-sonarsource, line 51: reuse the same list as in the v5, but modify it in the constructor/initializer. `GetRulesRequest.ResponseFields.Concat(new []{"descSec"}).ToList()`

src/SonarQube.Client/Api/V9_5/GetRulesWithDescriptionSectionsRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+        public string RuleKey { get; set; }
+        public int Page { get; set; }
+        public int PageSize { get; set; }
+        public int ItemsLimit { get; set; }
+        public ILogger Logger { get; set; }
+        public IList<string> ResponseFieldsList { get; set; }
+
+        public async Task<SonarQubeRule[]> InvokeAsync(HttpClient httpClient, CancellationToken token)
+        {
+            innerRequest.IsActive = this.IsActive;
```
georgii-borovinskikh-sonarsource, line 46: Can do this directly in the setters

src/SonarQube.Client/Api/V9_6/GetRulesWithEducationPrinciplesRequest.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Threading.Tasks;
+using SonarQube.Client.Api.V5_50;
+using SonarQube.Client.Logging;
+using SonarQube.Client.Models;
+
+namespace SonarQube.Client.Api.V9_6
+{
+    public class GetRulesWithEducationPrinciplesRequest : IGetRulesRequest
+    {
+        private readonly GetRulesRequest innerRequest = new GetRulesRequest();
```
georgii-borovinskikh-sonarsource, line 33: Could use v9.5 here instead of v5, less stuff to modify. Inheritance could work here. It has it's drawbacks (for example, if we in the future would need to remove a field instead of adding), but it's unlikely it'll cause problems or we can deal with them later IMO

src/SonarQube.Client.Tests/Requests/DefaultConfiguration_Configure.cs
```diff
@@ -90,7 +91,7 @@ public void ConfigureSonarCloud_Writes_Debug_Messages()
                     "Registered SonarQube.Client.Api.V5_00.GetSourceCodeRequest",
                     "Registered SonarQube.Client.Api.V5_10.GetLanguagesRequest",
                     "Registered SonarQube.Client.Api.V5_40.GetModulesRequest",
-                    "Registered SonarQube.Client.Api.V5_50.GetRulesRequest",
+                    "Registered SonarQube.Client.Api.V9_6.GetRulesWithEducationPrinciplesRequest",
```
georgii-borovinskikh-sonarsource, line 94: Only 9.6 here? what about 9.5?

ugras-ergun-sonarsource: Because SC always uses the latest version.

src/SonarQube.Client/Api/V9_5/GetRulesWithDescriptionSectionsRequest.cs
ugras-ergun-sonarsource: As discussed offline, this would need to make a change to the design and it's not worth it since it'll create other set of problems.

georgii-borovinskikh-sonarsource (approved):

Issue #3973, opened 2023-03-24
Rule help is not shown for Secrets in Connected Mode


duncanp-sonar:
### Description

Rule help is shown for Secrets in standalone mode, but not in Connected Mode.

### Repro
1. Open a bound solution with secrets e.g. `SLVS_Bound_VS2019`
2. Open a file containing secrets issues e.g. `SecretsSuppressedAndDisabled.txt`
3. Show error help for a secret issue (either from the Error List context menu, or by clicking on the rule key in "Code" column)

The rule help for other languages is show correctly in Connected Mode i.e. C#, VB.NET, C++, JavaScript and TypeScript.

### Expected
The `Sonar Rule Help` tool window opens and displays the help for the rule.

### Actual
Nothing happens.

#### Example log output
The log shows two actions:
[1] show help for C# issue - works correctly i.e. the rule key is recognised, and the rule info is fetched from the server
[2] show help for secret issue - the rule key is recognised, but that is all.

```
*** [1] Clicking to show C# rule help - works correctly

[ThreadId 1] [DEBUG] [ErrorList] Event processor: Sonar rule detected: csharpsquid:S3776
[ThreadId 39] [DEBUG] Looking up implementation of 'IGetRulesRequest' on thread '39'
[ThreadId 39] [DEBUG] Created request of type 'SonarQube.Client.Api.V5_50.GetRulesRequest'.
[ThreadId 39] [DEBUG] Sending Http request:
[ThreadId 39] [DEBUG] Method: GET, RequestUri: 'api/rules/search?qprofile=AX-IPZuknTk2GEVJ9VXQ&rule_key=csharpsquid%3AS3776&f=repo%2CinternalKey%2Cparams%2Cactives%2ChtmlDesc%2CdescriptionSections%2CeducationPrinciples%2Ctags%2Cname%2ChtmlNote&p=1&ps=500', Version: 1.1, Content: <null>, Headers:
{
}
[ThreadId 57] [DEBUG] Response with HTTP status code 'OK' received.
[...]

PR #4028, opened 2023-04-03
Add secrets to repo key mapping

duncanp-sonar:
Part of #3973

### DIFF
--- a/src/Rules/RuleMetaDataProvider.cs
+++ b/src/Rules/RuleMetaDataProvider.cs
@@ -19,6 +19,7 @@
  */
 
 using System.ComponentModel.Composition;
+using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using SonarLint.VisualStudio.Core;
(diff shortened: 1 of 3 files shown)


src/Rules/RuleMetaDataProvider.cs
```diff
@@ -55,6 +56,8 @@ public async Task<IRuleInfo> GetRuleInfoAsync(SonarCompositeRuleId ruleId, Cance
 
             //TODO: this does not seem to support taint.
             var language = Language.GetLanguageFromRepositoryKey(ruleId.RepoKey);
+            Debug.Assert(language != null, $"Unable to determine the language from the repo key. RepoKey: {ruleId.RepoKey}");
```
duncanp-sonar, line 59: Added an assertion to make this kind of failure more obvious during debugging/local testing.

duncanp-sonar: FYI previously, `language` would be null which would cause an unhandled exception in line 63 below.

This isn't the only change required to fix the bug - secrets rule info returns null for one of the fields, which causes an different error later. However, I'll create a separate PR for that.

ugras-ergun-sonarsource (approved):

PR #4031, opened 2023-04-03
Handle null collections when parsing server rule info

duncanp-sonar:
Handle null collections when processing server rule info

Last part of the fix for #3973

### DIFF
--- a/src/Rules/IRuleInfo.cs
+++ b/src/Rules/IRuleInfo.cs
@@ -120,13 +120,19 @@ namespace SonarLint.VisualStudio.Rules
         /// </summary>
         string Description { get; }
 
+        /// <summary>
+        /// List of tags. Can be empty. Will not be null.
+        /// </summary>
         IReadOnlyList<string> Tags { get; }
 
         /// <summary>
-        /// Tabs for new educational format
+        /// Tabs for new educational format. Can be empty. Will not be null.
         /// </summary>
         IReadOnlyList<IDescriptionSection> DescriptionSections { get; }
 
+        /// <summary>
+        /// Education principles for the new educational format. Can be empty. Will not be null.
+        /// </summary>
         IReadOnlyList<string> EducationPrinciples { get; }
 
         string HtmlNote { get; }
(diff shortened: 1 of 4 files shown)


src/Rules/IRuleInfo.cs
```diff
@@ -120,13 +120,19 @@ public interface IRuleInfo
         /// </summary>
         string Description { get; }
 
+        /// <summary>
+        /// List of tags. Can be empty. Will not be null.
+        /// </summary>
```
duncanp-sonar, line 125: I've changed `RuleInfo` so `Tags`, `DescriptionSections` and `EducationPrinciples` will return empty collections rather than nulls. That will make them easier to work with in the rest of the client code.

src/Rules/ServerRuleMetadataProvider.cs
```diff
@@ -61,30 +61,32 @@ public ServerRuleMetadataProvider(ISonarQubeService service, ILogger logger)
...
-                sqRule.GetCompositeKey(),
-                HtmlXmlCompatibilityHelper.EnsureHtmlIsXml(sqRule.Description),
-                sqRule.Name,
-                sqRule.Severity.ToRuleIssueSeverity(),
-                sqRule.IssueType.ToRuleIssueType(),
-                sqRule.IsActive,
-                sqRule.Tags,
-                descriptionSections,
-                sqRule.EducationPrinciples,
-                HtmlXmlCompatibilityHelper.EnsureHtmlIsXml(sqRule.HtmlNote));
```
duncanp-sonar, line 87: I moved the rest of this conversion code inside the exception handler. The new behaviour is that null will be returned in the event of any exceptions when trying to fetch the rule info from the server.

Previously, an exception was being thrown in line 75 of this class. It was outside the exception handler, so wasn't caught in this method. In fact, it wasn't caught anywhere. Fortunately, it was happening on a background thread, so VS was silently swallowing it.

The impact was that if the user clicked to see error help for a secret in Connected Mode, nothing would happen from the user's point of view (the exception would be thrown and silently squashed).

The new behaviour is that the exception will be caught and logged. This service will return `null`, and the education service that calls it will fall back on displaying the local help, if it exists.

Testing: there is already a unit test that checks that exceptions from the SonarQube service are caught and logged. Unfortunately, there aren't any injection points in the product code after the service call, so we can't unit test the new behaviour. However, I have manually stepped through it when debugging.

src/Rules/ServerRuleMetadataProvider.cs
```diff
@@ -61,30 +61,32 @@ public ServerRuleMetadataProvider(ISonarQubeService service, ILogger logger)
...
+                    sqRule.Tags,
+                    descriptionSections,
+                    sqRule.EducationPrinciples,
+                    HtmlXmlCompatibilityHelper.EnsureHtmlIsXml(sqRule.HtmlNote));
             }
             catch (Exception ex)
             {
                 logger.WriteLine(Resources.ServerMetadataProvider_GetRulesError, ruleId.ToString(), ex.Message);
             }
-            if (sqRule == null) return null;
```
duncanp-sonar, line 73: We always put curly braces round statements, even if there is only one.

ugras-ergun-sonarsource (approved):

Issue #4012, opened 2023-03-31
Render rule cross-references as hyperlinks in XAML flow document


duncanp-sonar:
### Description

The `XamlGeneratorHelper` turns rules descriptions into XAML FlowDocument elements. For blocks of text, the helper should use the rule cross-reference parser to extract rules cross-references.

The cross-references should be rendered as hyperlinks.

TODO:
* decide the format of the URI to embed in the hyperlink. The URI needs to encode the composite rule key.

PR #4039, opened 2023-04-04
Display rule cross reference as Hyperlink in rule description

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4012

### DIFF
--- a/src/Education/CrossReferenceRuleHelper.cs
+++ b/src/Education/CrossReferenceRuleHelper.cs
@@ -1,0 +1,40 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.Education
+{
+    /// <summary>
+    /// This is a helper class to encode a SonarCompositeRuleId into a valid URI and decode it vice versa.
+    /// </summary>
+    internal static class CrossReferenceRuleHelper
+    {
+        private const string prefix = "sonarlintrulecrossref";
+
+        public static Uri EncodeToUri(SonarCompositeRuleId compositeRuleId)
+        {
+            var link = $"{prefix}://{compositeRuleId.RepoKey}/{compositeRuleId.RuleKey}";
+
+            return new Uri(link);
+        }
+    }
+}
--- a/src/Education/XamlGenerator/RuleHelpXamlTranslator.cs
+++ b/src/Education/XamlGenerator/RuleHelpXamlTranslator.cs
@@ -24,6 +24,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Text;
 using System.Xml;
+using SonarLint.VisualStudio.Core;
 
 namespace SonarLint.VisualStudio.Education.XamlGenerator
 {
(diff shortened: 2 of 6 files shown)


src/Education/CrossReferenceRuleHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.Education
+{
+    /// <summary>
+    /// This is a helper class to encode a SonarCompositeRuleId into a valid URI and decode it vice versa.
+    /// </summary>
+    internal static class CrossReferenceRuleHelper
+    {
+        private const string prefix = "sonarlintrulecrossref";
```
bigfluffycookie, line 31: When creating a URI all capitalization gets removed, to make it less confusing I removed it here too simply so that when we test and when we decode we won't have to think about capitalization.

src/Education/CrossReferenceRuleHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.Education
+{
+    /// <summary>
+    /// This is a helper class to encode a SonarCompositeRuleId into a valid URI and decode it vice versa.
+    /// </summary>
+    internal static class CrossReferenceRuleHelper
```
bigfluffycookie, line 29: Hope this name is oki, couldn't think of anything better

src/Education/XamlGenerator/RuleHelpXamlTranslator.cs
```diff
@@ -356,9 +357,34 @@ private void WriteText(string text)
...
+                    // Note: we could explicitly wrap the text in a <Run>. However, that will happen implicitly
+                    // when the XAML is parsed, and it won't make any difference to the rendered output.
+                    writer.WriteString(simpleText.Text);
+                }
+                else if (textToken is IRuleCrossRef ruleCrossRef)
+                {
+                    HandleCrossReference(ruleCrossRef.CompositeRuleId);
+                }
+
+                Debug.Assert(textToken is IRuleCrossRef || textToken is ISimpleText, "Unknown text token type.");
```
bigfluffycookie, line 376: I could have added and else{} and added the debug assert in there but that made it look super crowded so I did it this way

src/Education/CrossReferenceRuleHelper.cs
duncanp-sonar: We could just call it something like `SonarRuleIdUriHelper` / `SonarRuleIdUriEncoderDecoder`.

It just so happens that we are using this class when handling rule cross-references embedded in help descriptions, but it's use isn't specific to cross-references.

src/Education/CrossReferenceRuleHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        private const string prefix = "sonarlintrulecrossref";
+
+        public static Uri EncodeToUri(SonarCompositeRuleId compositeRuleId)
```
duncanp-sonar, line 33: We could make it an extension method, which would allow us to write code like
```
var uri = ruleId.EncodeToUri();

...

if (uri.TryGetRuleId(out ruleId)) ...
```

However, it's just syntactic sugar.
`

It would look a

src/Education/XamlGenerator/RuleHelpXamlTranslator.cs
duncanp-sonar, line 376: It's always worth putting as much info into the assert message as possible e.g.

`$"Unknown text token type. Type: {textToken.GetType().FullName}"`

src/Education/CrossReferenceRuleHelper.cs
duncanp-sonar: What happened to `ahhhhh` ?! ;-)

duncanp-sonar (approved):
Nice! A couple of minor suggestions about naming/assert messages.

src/Education/CrossReferenceRuleHelper.cs
bigfluffycookie: I think you forgot to finish the sentence - but regardless of that as I do not like extension methods because I think they are confusing so I will leave it as is

src/Education/XamlGenerator/RuleHelpXamlTranslator.cs
bigfluffycookie: About this, I just realized should I have a test that covers this happening? So return a textToken thats neither of these types and see if it debug asserts?

duncanp-sonar: Yep, I'd have put the assertion outside the `if/else` too.

Assertions are not included in release builds, so the released code would look like this if the assertion was in an `else`:

```
                if (textToken is ISimpleText simpleText)
                {
                    ...
                }
                else if (textToken is IRuleCrossRef ruleCrossRef)
                {
                    ...
                }
                else
                {
                }
```

... which would probably generate a static analysis issue on the CI machine because we have an empty statement.

More generally, we don't want to change the normal flow of control for debug cases if we can possibly help it.

duncanp-sonar: We don't need to check that the assertion happens since the assert is just a debug-time helper for us, and we don't want to have to write tests for debug-specific behaviour.

However, you could write a test to check the expected product code behaviour in this case i.e. that unknown types are ignored (you'd need to use our `AssertIgnoreScope` class in the test to stop the assertion from popping up a dialogue and interrupting the test).

PR #4041, opened 2023-04-04
Add timeouts to regular expressions

### LINKED ISSUE #4035: Fix RegEx timeout warnings (S6444)
### Description

There are a number of Security Hotspot [warnings](https://sonarcloud.io/project/security_hotspots?id=sonarlint-visualstudio) for S6444 (using RegEx without a timeout).

The risk is low because we are generally using the RegEx against input we control. The possible impact is low (Denial of Service attack against an instance of Visual Studio).

However, it would be easy to fix to remove the hotspot warning.

duncanp-sonar:
Fixes #4035

### DIFF
--- a/build/tools/ExtractRuleDescFromJson/ExtractRuleDescFromJson.csproj
+++ b/build/tools/ExtractRuleDescFromJson/ExtractRuleDescFromJson.csproj
@@ -18,6 +18,7 @@
   -->
   <ItemGroup Label="SLVS Linked product code">
     <Folder Include="SLVS_LinkedProductCode\" />
+    <Compile Include="..\..\..\src\Core\RegexConstants.cs" Link="SLVS_LinkedProductCode\RegexConstants.cs" />
     <Compile Include="..\..\..\src\Rules\IRuleInfo.cs" Link="SLVS_LinkedProductCode\IRuleHelp.cs" />
     <Compile Include="..\..\..\src\Rules\HtmlXmlCompatibilityHelper.cs" Link="SLVS_LinkedProductCode\HtmlXmlCompatibilityHelper.cs" />
   </ItemGroup>
--- a/src/CFamily/CMake/MacroEvaluationService.cs
+++ b/src/CFamily/CMake/MacroEvaluationService.cs
@@ -50,7 +50,9 @@ namespace SonarLint.VisualStudio.CFamily.CMake
                                      // The word chars are captured in a group called "prefix".
             "?" +                    // Match the previous group 0 or 1 times i.e. the prefix is optional.
             "(?<name>\\w+)" +        // Capture one or more word consecutive word chars in a group called "name".
-            "}");                    // Match the closing literal.
+            "}",                     // Match the closing literal.
+            RegexOptions.Compiled,
+            Core.RegexConstants.DefaultTimeout);
 
         public MacroEvaluationService(ILogger logger)
             :this(logger, new MacroEvaluator()) {}
--- a/src/Core/GlobPatternMatcher.cs
+++ b/src/Core/GlobPatternMatcher.cs
@@ -64,7 +64,9 @@ namespace SonarLint.VisualStudio.Core
 
             private WildcardPattern(string pattern, string directorySeparator)
             {
-                this.pattern = new Regex(toRegexp(pattern, directorySeparator));
+                this.pattern = new Regex(toRegexp(pattern, directorySeparator),
+                                RegexOptions.Compiled,
+                                RegexConstants.DefaultTimeout);
             }
 
             private static string toRegexp(string antPattern, string directorySeparator)
--- a/src/Integration.Vsix/CFamily/VcxProject/CmdBuilder.cs
+++ b/src/Integration.Vsix/CFamily/VcxProject/CmdBuilder.cs
@@ -28,6 +28,15 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject
 {
     internal class CmdBuilder
     {
+        private static readonly Regex DoubleSeparatorRegEx = new Regex(@"\\+",
+            RegexOptions.Compiled,
+            Core.RegexConstants.DefaultTimeout);
+
+        // To avoid paths with spaces
+        private static readonly Regex DoubleQuotedRegEx = new Regex("\"[^\\\"]*(\\.[^\\\"]*)*\"",
+            RegexOptions.Compiled,
+            Core.RegexConstants.DefaultTimeout);
+
         StringBuilder Cmd { get; set; } = new StringBuilder();
         bool IsHeader { get; set; }
         public string HeaderFileLang { get; set; } = "";
@@ -63,9 +72,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject
 
         internal /* for testing */ static string AddQuote(string s)
         {
-            // To avoid paths with spaces
-            Regex doubleQuoted = new Regex("\"[^\\\"]*(\\.[^\\\"]*)*\"");
-            if (!doubleQuoted.IsMatch(s))
+            if (!DoubleQuotedRegEx.IsMatch(s))
             {
                 string quote = "\"";
                 return quote + s + quote;
@@ -78,8 +85,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject
             // path cannot be empty
             path = path.Replace("/", @"\");
 
-            Regex doubleSeparator = new Regex(@"\\+");
-            path = doubleSeparator.Replace(path, @"\");
+            path = DoubleSeparatorRegEx.Replace(path, @"\");
 
             if (path[path.Length - 1] == '\\')
             {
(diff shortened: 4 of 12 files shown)


duncanp-sonar:
We're getting security hotspot warnings because we don't generally set a timeout when using Regular Expressions, so I've been through all of the places we use regular expressions and added a timeout.

The way the `RegEx` class is written means that if we want to pass a timeout to the constructor, we also have to pass an instance of [RegExOptions](https://learn.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regexoptions?view=net-7.0). Most of the time I used for `RegExOptions.None` i.e. a no-op.

There are a couple of place where I used `RegExOptions.Compiled` instead. `Compiled` regular expressions take longer to initialize the first time, but then execute faster. I used `Compiled` in a few cases where there is an expression we use a lot e.g. when checking suppressions.

Finally, there were a couple of places were a regular expression was being created inside a method each time the method was called, even if the pattern was fixed. I moved those out of the method to be class-level variables.

I don't really like making this many changes scattered throughout the code base in a single PR. However, our unit test coverage of the regular expressions is good, and I've manually tested all of the changes (set breakpoints in all of the changed code and made sure I'd been through all of the affected code paths).

build/tools/ExtractRuleDescFromJson/ExtractRuleDescFromJson.csproj
```diff
@@ -18,6 +18,7 @@
   -->
   <ItemGroup Label="SLVS Linked product code">
     <Folder Include="SLVS_LinkedProductCode\" />
+    <Compile Include="..\..\..\src\Core\RegexConstants.cs" Link="SLVS_LinkedProductCode\RegexConstants.cs" />
```
duncanp-sonar, line 21: The `ExtractRuleDescFromJson.csproj` is one of our build-time tools that extracts data from the jars. As such, it can't have project references to any of our product assemblies, as they haven't been built yet e.g. it can't reference `Core.csproj`.

However, there are a couple of produce source code files we need to reference (e.g. the class used to serialize the rule help json), so the `ExtractRuleDescFromJson` project links directly to the source code files like this.

We now need the regular expression timeout constant, so the project now needs to reference the new constants class too.

src/CFamily/CMake/MacroEvaluationService.cs
```diff
@@ -50,7 +50,9 @@ internal class MacroEvaluationService : IMacroEvaluationService
                                      // The word chars are captured in a group called "prefix".
             "?" +                    // Match the previous group 0 or 1 times i.e. the prefix is optional.
             "(?<name>\\w+)" +        // Capture one or more word consecutive word chars in a group called "name".
-            "}");                    // Match the closing literal.
+            "}",                     // Match the closing literal.
+            RegexOptions.Compiled,
+            Core.RegexConstants.DefaultTimeout);
```
duncanp-sonar, line 55: Macros are evaluated every time we analyze a C++ file or pre-process a C++ header file, which happens every time the user switches between document windows.

i.e. it can happen a lot, and the regular expression pattern is fixed, so I decided to make the expression `Compiled`.

src/Core/GlobPatternMatcher.cs
```diff
@@ -64,7 +64,9 @@ private class WildcardPattern
 
             private WildcardPattern(string pattern, string directorySeparator)
             {
-                this.pattern = new Regex(toRegexp(pattern, directorySeparator));
+                this.pattern = new Regex(toRegexp(pattern, directorySeparator),
+                                RegexOptions.Compiled,
+                                RegexConstants.DefaultTimeout);
```
duncanp-sonar, line 69: The `WildcardPattern` is used every time we analyse a file in Connected Mode to check whether the file is include/excluded by the server settings like `sonar.exclusions`

The patterns don't change very often (only when the server settings are updated) and there are used frequently so I made them `Compiled`.

src/Integration.Vsix/CFamily/VcxProject/CmdBuilder.cs
```diff
@@ -63,9 +72,7 @@ public void AddCompiler(string path)
 
         internal /* for testing */ static string AddQuote(string s)
         {
-            // To avoid paths with spaces
-            Regex doubleQuoted = new Regex("\"[^\\\"]*(\\.[^\\\"]*)*\"");
```
duncanp-sonar, line 67: This method and the next one are called multiple times for every C++ analysis. There's no reason to create a new `Regex` each time, so I moved the definitions to `static readonly` constants.

The unit test coverage of these two methods is very good.

src/Integration/LocalServices/TestProjectIndicators/ProjectNameTestProjectIndicator.cs
```diff
@@ -54,11 +55,7 @@ public void SetTestRegex(string pattern)
 
             try
             {
-                Regex.IsMatch("", pattern);
```
duncanp-sonar, line 57: This call wasn't doing anything useful so I removed it.

src/Rules/HtmlXmlCompatibilityHelper.cs
```diff
@@ -18,23 +18,25 @@
...
         // e.g. <col span="123">  =>  <col span="123"/>
-        private static readonly Regex CleanCol = new Regex("(?<element>(<col\\s*)|(col\\s+[^/^>]*))>", RegexOptions.Compiled);
+        private static readonly Regex CleanCol = new Regex("(?<element>(<col\\s*)|(col\\s+[^/^>]*))>",
+            RegexOptions.Compiled,
+            Core.RegexConstants.DefaultTimeout);
 
-        private static readonly Regex CleanBr = new Regex("(?<element>(<br\\s*)|(br\\s+[^/^>]*))>", RegexOptions.Compiled);
+        private static readonly Regex CleanBr = new Regex("(?<element>(<br\\s*)|(br\\s+[^/^>]*))>",
+            RegexOptions.Compiled,
+            Core.RegexConstants.DefaultTimeout);
```
duncanp-sonar, line 39: These two are called multiple times each time the user asks to see rule help -> `Compiled`

src/SonarQube.Client/Api/V6_30/GetPluginsRequest.cs
```diff
@@ -50,10 +54,7 @@ public class GetPluginsRequest : RequestBase<SonarQubePlugin[]>, IGetPluginsRequ
         /// <returns>Version field formatted for SonarLint</returns>
         public static string FormatForSonarLint(string version)
         {
-            const string sonarQubeVersionPattern =
-                @"(?<major>\d+)[\.]?(?<minor>\d+)?[\.]?(?<patch>\d+)?.*";
-
-            var match = Regex.Match(version, sonarQubeVersionPattern);
```
duncanp-sonar, line 56: Create a `static readonly` `RegEx` to avoid creating an new object each time.

src/SonarQube.Client/Api/V6_30/GetPluginsRequest.cs
```diff
@@ -29,6 +30,9 @@ namespace SonarQube.Client.Api.V6_30
 {
     public class GetPluginsRequest : RequestBase<SonarQubePlugin[]>, IGetPluginsRequest
     {
+        private const string sonarQubeVersionPattern = @"(?<major>\d+)[\.]?(?<minor>\d+)?[\.]?(?<patch>\d+)?.*";
+        private static readonly Regex SonarVersionRegEx = new Regex(sonarQubeVersionPattern, RegexOptions.None, TimeSpan.FromMilliseconds(300));
```
duncanp-sonar, line 34: The `SonarQube.Client` class doesn't reference `Core`, and it can't at the moment because it is reference by `Core` -> I couldn't use the new timeout constant here.

src/SonarQube.Client/ChecksumCalculator.cs
```diff
@@ -36,14 +36,18 @@ public interface IChecksumCalculator
     /// </summary>
     public class ChecksumCalculator : IChecksumCalculator
     {
+        private static readonly Regex StripWhitespaceRegEx = new Regex("\\s",
+            RegexOptions.Compiled, 
+            TimeSpan.FromMilliseconds(300));
```
duncanp-sonar, line 41: Another class in `SonarQube.Client`, so again I used a hard-coded timeout.
I thought about adding a separate timeout constant for `SonarQube.Client` but decided it wasn't worth it with only two instances.

The correct fix would be to refactor the code to reverse the dependency between `SonarQube.Client` and `Core`, but that's a much bigger job.

src/SonarQube.Client/ChecksumCalculator.cs
```diff
@@ -36,14 +36,18 @@ public interface IChecksumCalculator
...
+
         public static string Calculate(string text)
         {
             if (text == null)
             {
                 throw new ArgumentNullException(nameof(text));
             }
 
-            var content = Regex.Replace(text, "\\s", ""); // strip whitespace
```
duncanp-sonar, line 46: Created a `static readonly RegEx`.

bigfluffycookie (approved):
Very nice! I saw the security warning and as there were some old ones I just assumed that we are ignoring them - my bad for not double checking but glad you fixed it!

Issue #4013, opened 2023-03-31
Handle clicks on rule cross-reference links


duncanp-sonar:
### Description

The Rule Help control handles user-clicks on hyperlinks in the help FlowDocument.

When the clicked hyperlink is for rule cross reference it should extract the composite rule id from the URI and call the `IEducation` service.

PR #4042, opened 2023-04-05
Handle clicks on rule cross-reference links in rule description.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4013


src/Education.UnitTests/SonarRuleIdUriEncoderDecoderTests.cs
```diff
@@ -36,5 +37,30 @@ public void EncodeToUri_EncodesCompositeKeyCorrectly()
...
+            var decodableUri = new Uri($"sonarlintrulecrossref://cpp/S1234");
+
+            var result = SonarRuleIdUriEncoderDecoder.TryDecodeToCompositeRuleId(decodableUri, out SonarCompositeRuleId compositeRuleId);
+
+            result.Should().BeTrue();
+            compositeRuleId.RepoKey.Should().Be("cpp");
+            compositeRuleId.RuleKey.Should().Be("S1234");
+        }
+
+        [TestMethod]
```
bigfluffycookie, line 53: Two test cases so its clear that it doesn't need to be in https format to fail, it just needs to be anything BUT the correct prefix

src/Education/SonarRuleIdUriEncoderDecoder.cs
```diff
@@ -36,5 +38,28 @@ public static Uri EncodeToUri(SonarCompositeRuleId compositeRuleId)
...
+            if (uri.Scheme != prefix)
+            {
+                return false;
+            }
+
+            // It is not required to use uri.Host here. It would have been fine to use any
+            // other property that had the information on the repository. 'host' is just the most simple.
+            var ruleRepo = uri.Host;
+
+            Debug.Assert(uri.AbsolutePath[0] == '/',
```
bigfluffycookie, line 56: We really aren't ever expecting this to happen however, I am not sure what to do if it does? As of right now it would just crash but I dont know until what point I can assume to get the correct input.

src/Education/SonarRuleIdUriEncoderDecoder.cs
```diff
@@ -36,5 +38,28 @@ public static Uri EncodeToUri(SonarCompositeRuleId compositeRuleId)
 
             return new Uri(link);
         }
+
+        public static bool TryDecodeToCompositeRuleId(Uri uri, out SonarCompositeRuleId compositeRuleId)
```
bigfluffycookie, line 42: I thought about instead of returning the boolean to just check if the compositeRuleID is null, is there any reason to do it this way? (Just because you suggested it with a return parameter)

duncanp-sonar: Personally, I quite like `Try` methods that return `null` if the action can't be performed - it shorter and seems quite logical to me.

However, `bool TryX(input, out output)` is a common pattern in the framework and in C# in general, so some people find it confusing if a method called `TryX` doesn't follow the pattern.

src/Education/SonarRuleIdUriEncoderDecoder.cs
duncanp-sonar: It's fine if we crash on bad input here.

The assertion has two purposes: it documents an assumption we are making about how this method is expected to be called, and acts as a notification in debug builds if that assumption is not met (i.e. highlights there is a problem and makes it easy to find). It's not intended to prevent a crash.

duncanp-sonar (approved):

Issue #4043, opened 2023-04-05
Extended rule descriptions link can cause VS to crash


bigfluffycookie:
### Description

If a user adds an extended rule description with a relative link it will cause VS to crash.

### Repo:

- Add an extended rule description in the format `[Test]("/relativeLink")`
- Click on the link in the rule description

### Workaround
None

### Affected versions
SonarLint: v6.14+

PR #4045, opened 2023-04-05
Handle relative URIs in Rule descriptions.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4043

### DIFF
--- a/src/Education/Commands/RuleHelpUserControl.xaml.cs
+++ b/src/Education/Commands/RuleHelpUserControl.xaml.cs
@@ -21,6 +21,7 @@
 using System.Windows.Controls;
 using System.Windows.Navigation;
 using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Integration;
 
 namespace SonarLint.VisualStudio.Education.Commands
 {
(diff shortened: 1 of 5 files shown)


src/Education/Commands/RuleHelpUserControl.xaml.cs
```diff
@@ -21,24 +21,35 @@
...
+            this.logger = logger;
 
             InitializeComponent();
         }
 
         public void HandleRequestNavigate(object sender, RequestNavigateEventArgs e)
         {
+            if (!e.Uri.IsAbsoluteUri)
+            {
+                logger.LogVerbose($"[RuleHelpUserControl] Failed to navigate to Uri as it is a relative path: {e.Uri}");
```
bigfluffycookie, line 47: I wanted to use the resource file but no matter what I did this class would not recognize Resources so I gave up, Do I need to do something special to be able to reference it?

duncanp-sonar: Interesting - I hadn't come across this problem before.

This is a `UserControl` so it already has a `Resources` property so just referring to `Resources` won't work. I expected using the full-qualified name `SonarLint.VisualStudio.Education.Resources` in the code to work, but it doesn't (I'm not entirely sure why - there's no reason for it not to work at the compiler level, so it might be a bug in VS). I'm guessing you tried both of those.

I found a couple of workarounds:
* change the namespace of the control so it isn't under `SonarLint.VisualStudio.Education`.
* change the name of the resource file to something other than `Resources`
* change the namespace of the `Resources` file e.g. by setting the "Custom Tool Namespace" property in the `Properties` window

* last, and simplest: add a `using` alias in the `RuleHelpUserControl` e.g. 
`using EducationResx = SonarLint.VisualStudio.Education.Resources;`

You can then reference the resource file as `EducationResx.Resources`. `Using` alias are a way of handling namespace conflicts.

src/Education/Commands/RuleHelpUserControl.xaml.cs
```diff
@@ -21,24 +21,35 @@
...
+                logger.WriteLine($"[RuleHelpUserControl] Failed to open URI: {e.Uri}");
```
duncanp-sonar, line 48: `RuleHelpUserControl` is a bit low-level for "normal" verbosity messages. We tend to use a higher-level description of the feature e.g. `[Education]`

duncanp-sonar (commented):
Looks good. See the comments; you should be able to move the string to the resource file if you add a `using` alias.

PR #4081, opened 2023-04-12
Fix embedded IRuleInfo deserialization

### LINKED ISSUE #4080: ILocalRuleMetadaProvider fails to load rules with DescriptionSections
### Description

Upgrading to v10.1.0.21143 of the JS/TS analyzer causes tests failures: the rule metadata in the new version now contains `DescriptionSections`, which the deserializer fails to handle.

### Notes
The `RuleInfo` has the following property:

```
        /// <summary>
        /// Tabs for new educational format. Can be empty. Will not be null.
        /// </summary>
        IReadOnlyList<IDescriptionSection> DescriptionSections { get; }
```

So far we have just be using `JsonConvert.DeserializeObject<RuleInfo>(data);` directly to deserialize the embedded rule help json.
That worked when there was no `DescriptionSection` information. It fails if there is data because the deserializer can't work out which concrete type to deserialize the data into given the way our classes are set up.

There are a number of ways round this. See e.g. [this StackOverflow post](https://stackoverflow.com/questions/5780888/casting-interfaces-for-deserialization-in-json-net).

duncanp-sonar:
Fixes #4080


src/Core/Helpers/InterfaceToConcreteTypeConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+using Newtonsoft.Json;
+
+namespace SonarLint.VisualStudio.Core.Helpers
+{
+    /// <summary>
+    /// Custom deserializer to help serializing/deserializing to types that have interface property types
+    /// </summary>
+    /// <typeparam name="TInterface">The interface type being serialized/deserialized</typeparam>
+    /// <typeparam name="TConcrete">The concrete type that NewtonSoft.Json should use</typeparam>
+    public class InterfaceToConcreteTypeConverter<TInterface, TConcrete> : JsonConverter where TConcrete : TInterface
```
duncanp-sonar, line 31: This will make more sense when looking at the code that uses it

src/Rules.UnitTests/LocalRuleMetadataProvider.RuleInfoDeserializerTests.cs
```diff
@@ -0,0 +1,80 @@
...
+  'Tags': [],
+  'DescriptionSections': [
```
duncanp-sonar, line 41: Previously none of the analyzers had any data in the `DescriptionSections`, so deserializing it wasn't a problem.
The newest JS/TS analyzer has data, and the deserialiation failed.

src/Rules.UnitTests/LocalRuleMetadataProvider.RuleInfoDeserializerTests.cs
```diff
@@ -0,0 +1,80 @@
...
+      'Key': 'root_cause',
+      'HtmlContent': 'content2',
+      'Context': 
+        {
+            'Key': 'context key',
+            'DisplayName': 'context display name'
+        }
+    }
+  ],
+  'EducationPrinciples': [ 'aaa', 'bbb' ],
```
duncanp-sonar, line 57: The JS/TS analyzer doesn't seem to populate the `Context` or `EducationPrinciples` fields yet, but I added some test data to make the deserialization works.

src/Education.UnitTests/SimpleRuleHelpXamlBuilderTests.cs
```diff
@@ -131,7 +131,7 @@ private static bool ProcessResource(string fullResourceName)
             try
             {
                 var data = ReadResource(fullResourceName);
-                var jsonRuleInfo = JsonConvert.DeserializeObject<RuleInfo>(data);
+                var jsonRuleInfo = LocalRuleMetadataProvider.RuleInfoJsonDeserializer.Deserialize(data);
```
duncanp-sonar, line 134: Some of the tests were deserializing the embedded json as part of the test, so they need to use the new `RuleInfo` deserializer.

src/Core/Helpers/InterfaceToConcreteTypeConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        public override bool CanConvert(Type objectType) => objectType == typeof(TInterface);
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+            => serializer.Deserialize<TConcrete>(reader);
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+            => serializer.Serialize(writer, value);
```
duncanp-sonar, line 39: I'll add a few simple tests for this; currently it's partially tested indirectly by the new `RuleInfoJsonDeserializer`.

bigfluffycookie (approved):
Nice job!

ugras-ergun-sonarsource (approved):

PR #4098, opened 2023-04-14
Create CSS analysis Response 

### LINKED ISSUE #4086: Create CSS analysis Response

ugras-ergun-sonarsource:
fixes #4086

### DIFF
--- a/src/TypeScript/EslintBridgeClient/Contract/Issue.cs
+++ b/src/TypeScript/EslintBridgeClient/Contract/Issue.cs
@@ -18,46 +18,10 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System.Collections.Generic;
 using Newtonsoft.Json;
-using SonarLint.VisualStudio.Core.Analysis;
 
 namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract
 {
-    internal class AnalysisResponse
-    {
-        // Other fields omitted
-        // i.e. Highlight[], HighlightedSymbol[], Metrics, CpdToken[]
-
-        [JsonProperty("issues")]
-        public IEnumerable<Issue> Issues { get; set; }
-
-        [JsonProperty("parsingError")]
-        public ParsingError ParsingError { get; set; }
-    }
-
-    internal class ParsingError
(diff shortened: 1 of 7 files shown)


src/TypeScript/EslintBridgeClient/Contract/Issue.cs
ugras-ergun-sonarsource, line 1: Git mixed up the change, I have extracted Issue from AnalysisResponse and renamed it to JsTsAnalysisResponse

src/TypeScript/EslintBridgeClient/Contract/JsTsAnalysisResponse.cs
ugras-ergun-sonarsource, line 1: I haven't created a base class like the request because in SonarJS `AnalysisOutput ` is empty so I would like to be consistent with SonarJS in our structure but it didn't make sense to create an empty base class. 

https://github.com/SonarSource/SonarJS/blob/a1f3c65ff6d9f31e210a2001390a2d460233b587/src/services/analysis/analysis.ts#L51

src/TypeScript/EslintBridgeClient/Contract/JsTsAnalysisResponse.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace SonarLint.VisualStudio.TypeScript.EslintBridgeClient.Contract
+{
+    internal class JsTsAnalysisResponse
+    {
+        // Other fields omitted
+        // i.e. Highlight[], HighlightedSymbol[], Metrics, CpdToken[]
+
+        [JsonProperty("issues")]
+        public IEnumerable<Issue> Issues { get; set; }
```
georgii-borovinskikh-sonarsource, line 32: This could be in the common base class with Css response tbh

georgii-borovinskikh-sonarsource (approved):

src/TypeScript/EslintBridgeClient/Contract/JsTsAnalysisResponse.cs
ugras-ergun-sonarsource: I have thought of it but I have decided to follow the structure of SonarJS. I think it's better to try to be consistent with it.

PR #4102, opened 2023-04-14
Refname EslntBridgeClient to JsTsEslintBridgeClientBase, extract common fields and disposal logic to EslintBridgeClientBase to be reused by future css clent

georgii-borovinskikh-sonarsource:


georgii-borovinskikh-sonarsource:
Code smells are from the old code I moved to a new place

ugras-ergun-sonarsource (approved):

PR #4113, opened 2023-04-18
Refactor js and ts EslintBridgeAnalyzers, extract common logic 

### LINKED ISSUE #4088: Refactor js and ts EslintBridgeAnalyzers, extract common logic

ugras-ergun-sonarsource:
fixes #4088

### DIFF
--- a/src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
+++ b/src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
@@ -95,7 +95,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
             var client = SetupEslintBridgeAnalyzer(null);
 
             var testSubject = CreateTestSubject(client.Object, telemetryManager: telemetryManager.Object);
-            await testSubject.ExecuteAnalysis(ValidFilePath, Mock.Of<IIssueConsumer>(), CancellationToken.None);
+            await testSubject.ExecuteAnalysisAsync(ValidFilePath, Mock.Of<IIssueConsumer>(), CancellationToken.None);
 
             telemetryManager.Verify(x => x.LanguageAnalyzed("js"), Times.Once);
             telemetryManager.VerifyNoOtherCalls();
@@ -109,7 +109,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
             var consumer = new Mock<IIssueConsumer>();
 
             var testSubject = CreateTestSubject(eslintBridgeAnalyzer.Object);
-            await testSubject.ExecuteAnalysis(ValidFilePath, consumer.Object, CancellationToken.None);
+            await testSubject.ExecuteAnalysisAsync(ValidFilePath, consumer.Object, CancellationToken.None);
 
             consumer.VerifyNoOtherCalls();
         }
@@ -122,7 +122,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
             var consumer = new Mock<IIssueConsumer>();
 
             var testSubject = CreateTestSubject(eslintBridgeAnalyzer.Object);
-            await testSubject.ExecuteAnalysis(ValidFilePath, consumer.Object, CancellationToken.None);
+            await testSubject.ExecuteAnalysisAsync(ValidFilePath, consumer.Object, CancellationToken.None);
 
             consumer.Verify(x => x.Accept(ValidFilePath, issues));
         }
@@ -134,7 +134,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
 
             var testSubject = CreateTestSubject(eslintBridgeAnalyzer.Object);
 
-            Func<Task> act = async () => await testSubject.ExecuteAnalysis(ValidFilePath, Mock.Of<IIssueConsumer>(), CancellationToken.None);
+            Func<Task> act = async () => await testSubject.ExecuteAnalysisAsync(ValidFilePath, Mock.Of<IIssueConsumer>(), CancellationToken.None);
             act.Should().ThrowExactly<StackOverflowException>();
         }
 
@@ -159,7 +159,7 @@ namespace SonarLint.VisualStudio.TypeScript.UnitTests.Analyzer
             var eslintBridgeAnalyzer = SetupEslintBridgeAnalyzer(exceptionToThrow: exception);
 
             var testSubject = CreateTestSubject(eslintBridgeAnalyzer.Object, statusNotifier: statusNotifier.Object);
-            await testSubject.ExecuteAnalysis(ValidFilePath, Mock.Of<IIssueConsumer>(), CancellationToken.None);
+            await testSubject.ExecuteAnalysisAsync(ValidFilePath, Mock.Of<IIssueConsumer>(), CancellationToken.None);
 
             statusNotifier.Verify(x => x.AnalysisStarted(), Times.Once);
             statusNotifier.Verify(x => x.AnalysisFailed("this is a test"), Times.Once);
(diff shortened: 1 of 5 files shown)


src/TypeScript/Analyzer/TypeScriptAnalyzer.cs
```diff
@@ -80,10 +72,10 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
         {
             Debug.Assert(IsAnalysisSupported(detectedLanguages));
 
-            ExecuteAnalysis(path, consumer, cancellationToken).Forget(); // fire and forget
+            ExecuteAnalysisAsync(path, consumer, cancellationToken).Forget(); // fire and forget
         }
 
-        internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, CancellationToken cancellationToken)
+        internal async Task ExecuteAnalysisAsync(string filePath, IIssueConsumer consumer, CancellationToken cancellationToken)
```
ugras-ergun-sonarsource, line 78: changed the name of the method to make it obvious about it's pupose

ugras-ergun-sonarsource:
Quality gate failed on coverage but it's because there's no real code change.

src/TypeScript/Analyzer/TypeScriptAnalyzer.cs
georgii-borovinskikh-sonarsource, line 78: ts config here is only used in the NullOrEmpty check and is not passed to the ExecuteAnalysis method (see (1))

src/TypeScript/Analyzer/TypeScriptAnalyzer.cs
```diff
@@ -105,15 +97,7 @@ internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, Ca
...
-                var issues = await eslintBridgeAnalyzer.Analyze(filePath, tsConfig, cancellationToken);
-                analysisStatusNotifier.AnalysisFinished(issues.Count, stopwatch.Elapsed);
-
-                if (issues.Any())
-                {
-                    consumer.Accept(filePath, issues);
-                }
+                stopwatch.Stop();
             }
             catch (TaskCanceledException)
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I think you could make a virtual method GetTsConfig which would be empty (return null) in the base and overridden in the TS analyzer. This would allow to move the invocation inside the try catch in the AnalysisBase and get rid of this try catch

georgii-borovinskikh-sonarsource (changes requested):
I think there's a bug + my proposed fix to it could solve the duplication issue as well

src/TypeScript/Analyzer/TypeScriptAnalyzer.cs
```diff
@@ -80,10 +72,10 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
             Debug.Assert(IsAnalysisSupported(detectedLanguages));
 
-            ExecuteAnalysis(path, consumer, cancellationToken).Forget(); // fire and forget
+            ExecuteAnalysisAsync(path, consumer, cancellationToken).Forget(); // fire and forget
         }
 
-        internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, CancellationToken cancellationToken)
+        internal async Task ExecuteAnalysisAsync(string filePath, IIssueConsumer consumer, CancellationToken cancellationToken)
         {
             telemetryManager.LanguageAnalyzed("ts");
```
georgii-borovinskikh-sonarsource, line 80: you could move this and the following 3 lines to the base class as well. Just instead of passing analysisStatusNotifier you would pass the 2 strings ("ts", could be called `analysedLanguageTelemetryName` and nameof(TypeScriptAnalyzer), could be called `analyzerType` or something like that

src/TypeScript/Analyzer/JavaScriptAnalyzer.cs
```diff
@@ -72,47 +60,7 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
         {
             Debug.Assert(IsAnalysisSupported(detectedLanguages));
 
-            ExecuteAnalysis(path, consumer, cancellationToken).Forget(); // fire and forget
-        }
-
-        internal async Task ExecuteAnalysis(string filePath, IIssueConsumer consumer, CancellationToken cancellationToken)
```
ugras-ergun-sonarsource, line 78: After creating the virtual GetTsConfig it was possible to move everything to the base class and there wasn't any need for an async method in the derived class.

src/TypeScript/Analyzer/AnalyzerBase.cs
```diff
@@ -0,0 +1,109 @@
...
+            analysisStatusNotifier = analysisStatusNotifierFactory.Create(analyzerName, filePath);
+
+            await threadHandling.SwitchToBackgroundThread();
+            analysisStatusNotifier.AnalysisStarted();
+
+            try
+            {
+                var tsConfig = await GetTsConfig(filePath, cancellationToken);
+
+                if (tsConfig == null && languageKey == "ts") { return; }
```
georgii-borovinskikh-sonarsource, line 78: Task<(bool isError, tsconfig)> TryGetTsConfig
return when isError == true
in other cases, continue

I don't like that we use telemetry language key for logic

src/TypeScript/Analyzer/AnalyzerBase.cs
```diff
@@ -0,0 +1,109 @@
...
+            var rulesProvider = rulesProviderFactory.Create(repoKey, language);
+            eslintBridgeAnalyzer = eslintBridgeAnalyzerFactory.Create(rulesProvider, eslintBridgeClient);
+        }
+
+        protected async virtual Task<string> GetTsConfig(string sourceFilePath, CancellationToken cancellationToken)
+        {
+            return null;
+        }
+
+        internal async Task ExecuteAsync(string languageKey, string analyzerName, string filePath, IIssueConsumer consumer, CancellationToken cancellationToken)
```
georgii-borovinskikh-sonarsource, line 66: I think it's important to specify that the language key here is for telemetry

src/TypeScript/Analyzer/TypeScriptAnalyzer.cs
```diff
@@ -80,58 +71,24 @@ public bool IsAnalysisSupported(IEnumerable<AnalysisLanguage> languages)
...
-                analysisStatusNotifier.AnalysisFailed(ex);
+                analysisStatusNotifier.AnalysisFailed(Resources.ERR_NoTsConfig);
+                return null;
             }
-        }
 
-        public void Dispose()
-        {
-            eslintBridgeAnalyzer.Dispose();
+            logger.WriteLine("[TypescriptAnalyzer] time to find ts config: " + stopwatch.ElapsedMilliseconds);
```
georgii-borovinskikh-sonarsource, line 88: nameof

src/TypeScript/Analyzer/AnalyzerBase.cs
georgii-borovinskikh-sonarsource: The default implementation would always return (false, null)

src/TypeScript.UnitTests/Analyzer/TypeScriptAnalyzerTests.cs
```diff
@@ -94,69 +94,69 @@ public void IsAnalysisSupported_HasTypeScript_True()
...
-        public async Task ExecuteAnalysis_NoTsConfig_ConsumerNotCalled()
+        public async Task ExecuteAnalysisAsync_NoTsConfig_ConsumerNotCalled()
         {
             var tsConfigProvider = SetupTsConfigProvider(result: null);
             var eslintBridgeAnalyzer = SetupEslintBridgeAnalyzer();
             var consumer = new Mock<IIssueConsumer>();
 
             var testSubject = CreateTestSubject(eslintBridgeAnalyzer.Object, tsConfigProvider);
-            await testSubject.ExecuteAnalysis(ValidFilePath, consumer.Object, CancellationToken.None);
+            await testSubject.ExecuteAsync("ts", nameof(TypeScriptAnalyzer), ValidFilePath, consumer.Object, CancellationToken.None);
```
georgii-borovinskikh-sonarsource, line 117: IMO should replace with constant fields instead

src/TypeScript.UnitTests/Analyzer/JavaScriptAnalyzerTests.cs
```diff
@@ -95,7 +95,7 @@ public async Task ExecuteAnalysis_AlwaysCollectsTelemetry()
             var client = SetupEslintBridgeAnalyzer(null);
 
             var testSubject = CreateTestSubject(client.Object, telemetryManager: telemetryManager.Object);
-            await testSubject.ExecuteAnalysis(ValidFilePath, Mock.Of<IIssueConsumer>(), CancellationToken.None);
+            await testSubject.ExecuteAsync("js", nameof(JavaScriptAnalyzer), ValidFilePath, Mock.Of<IIssueConsumer>(), CancellationToken.None);
```
georgii-borovinskikh-sonarsource, line 98: IMO should replace with constant fields instead

src/TypeScript/Analyzer/AnalyzerBase.cs
georgii-borovinskikh-sonarsource: Or some other way of indicating that's it in one of the 3 states ("not required", "all good" and "required, but missing")

georgii-borovinskikh-sonarsource (approved):

Issue #4120, opened 2023-04-24
[Binding] Drop old Connected Mode code


duncanp-sonar:
### Description

Work out which bits of the existing binding code can be dropped completely.

Some initial candidates:
* check for bindable project types
* check that the binding code is up to date (may still want the initial trigger,  but the code that walks the solution and checks each project should not be required)
* check that the C#/VB QPs have not been weakened by a custom ruleset
* "two-phase" commit code that writes changes to project files

After all of the "top-level" code has been dropped we should check the helper code e.g. do we still need everything in `ProjectSystemHelper`? (might need some of it for the migration experience code)

PR #4125, opened 2023-04-24
Drop IConflictsManager

duncanp-sonar:
- checks for conflicting Ruleset settings -> no longer required

Part of #4120


src/Integration.UnitTests/Binding/BindingControllerTests.cs
```diff
@@ -325,26 +321,14 @@ public void BindingController_BindingFinished_Navigation()
                 dteMock.ToolWindows.SolutionExplorer.Window.Active.Should().BeFalse();
             }
 
-            // Case 2: Has conflicts (should navigate to team explorer page)
```
duncanp-sonar, line 328: Fix up test since we will no longer get conflicts

bigfluffycookie (approved):
Its nice to see so many files deleted

duncanp-sonar:
@bigfluffycookie 
> Its nice to see so many files deleted

Definitely - and there should be a whole lot more that we can get rid of, along with some complicated test code.

PR #4127, opened 2023-04-24
Drop NuGet package installation code

duncanp-sonar:
- part of the original Connected Mode experience -> no longer required

Part of #4120


src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -510,39 +496,6 @@ public void InstallPackages_Succeeds_SuccessPropertyIsTrue()
             testSubject.InternalState.BindingOperationSucceeded.Should().BeTrue();
         }
 
-        [TestMethod]
-        public void InstallPackages_Succeeds_SuccessPropertyIsFalse()
```
duncanp-sonar, line 514: The implementation method is now hard-coded to return `true`, so this test is no longer required (and it was confusingly named in any event; it should have been called `InstallPackages_Fails_SuccessPropertyIsFalse`).

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -203,7 +199,8 @@ private BindingConfiguration QueueWriteBindingInformation()
 
         public void PrepareToInstallPackages()
         {
-            this.NuGetBindingOperation.PrepareOnUIThread();
+            // TODO - CM cleanup
+            // no-op
```
duncanp-sonar, line 203: A long time ago, the original binding process used to install the C#/VB.NET analyzers as NuGet packages. This code was kept for backwards compatibility, and we can drop it now.

The NuGet installation was a two-step process: "prepare" then "install".
I've made the minimal changes to `BindingProcessImpl` again - just hard-coded the methods do nothing/return success.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -144,9 +144,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="NuGet.Core" Version="2.12.0" />
-    <PackageReference Include="NuGet.VisualStudio" Version="3.3.0" />
```
duncanp-sonar, line 148: We no longer use any of the types in these packages so we can drop the references to them.

bigfluffycookie (approved):

Issue #4122, opened 2023-04-24
Create targets file in ImportsBefore


bigfluffycookie:
Create a .targets file which will be used to reference the globalconfigs.

PR #4128, opened 2023-04-25
Create .targets file in ImportBefore

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4122


src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        private void CreateTargetsFileIfNotExists()
+        {
+            var sb = new StringBuilder();
+            var xmlWriter = Create(sb);
+
+            xmlWriter.WriteStartElement("Project", "http://schemas.microsoft.com/developer/msbuild/2003");
+            xmlWriter.WriteAttributeString("xmlns", "http://schemas.microsoft.com/developer/msbuild/2003");
+
```
bigfluffycookie, line 51: I already have the code to ad din the globalAnalyzers however, I dont actually know the paths of said files here so I wasn't sure it made sense to include it. 

```
private void WriteGlobalAnalyzerConfigFile(XmlWriter xmlWriter, string condition, string pathToConfig)
        {
            xmlWriter.WriteStartElement("GlobalAnalyzerConfigFiles");
            xmlWriter.WriteAttributeString("Condition", $"'$(Language)' == '{condition}'");
            xmlWriter.WriteAttributeString("Include", pathToConfig);
            xmlWriter.WriteEndElement();
        }
```

src/ConnectedMode.UnitTests/Binding/ImportsBeforeFileGeneratorTests.cs
```diff
@@ -0,0 +1,103 @@
...
+            fileSystem.Setup(x => x.File.Exists(pathToFile)).Returns(false);
+
+            _ = new ImportBeforeFileGenerator(fileSystem.Object);
+
+            fileSystem.Verify(x => x.Directory.CreateDirectory(pathToDirectory), Times.Once);
+        }
+
+        private string GetPathToImportBefore()
+        {
+            var localAppData = Environment.GetEnvironmentVariable("LOCALAPPDATA");
```
duncanp-sonar, line 97: FYI we have an abstraction `IEnvironmentVariableProvider` that allows us to mock environment variables.

I think it's ok not to use it here since we're only using one environment variable, and it's guaranteed to exist on Windows machines. The abstraction is more useful if we are using "custom" environment variables,  or if the expected behaviour of the code changes depending on the value of the variable.

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
duncanp-sonar: Building the file content using an XmlWriter is going to be hard to read and maintain. Because the content of the file is almost completely static, we could take a simpler approach and just use a hard-coded string.
If we do need to substitute data into the file can just put a known placeholder in the text and using `String.Replace` e.g.

e.g. 

```

string content = @"<Project...>

  <ItemGroup Condition="" $(RunningInsideVisualStudio) == 'true' "">
       <SLVSConfig>[[PATH_PLACEHOLDER]]</SLVSConfig>
  </ItemGroup>

</Project>"

content = content.Replace("[[PATH_PLACEHOLDER]]", filePath);
```

We can take this a step further; instead of defining the content as a string in code, we can embed a text file in the assembly, and then just read the contents of the file at runtime using `Assembly.GetManifestResourceStream`.

That way, when we want to change the content, we just need to edit the text file.
To embed a text file, we just need to add a text file to the project and set the `Build Action` to `Embedded Resource`.

We do this in a couple of places in the product already; for example, the rule help files are all embedded as text files.
Let's have a chat about this offline after the sync.

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
```diff
@@ -0,0 +1,105 @@
...
+            var pathToImportBefore = GetPathToImportBefore();
+
+            if (!fileSystem.Directory.Exists(pathToImportBefore))
+            {
+                fileSystem.Directory.CreateDirectory(pathToImportBefore);
+            }
+
+            var fullPath = Path.Combine(pathToImportBefore, targetsFileName);
+
+            if (fileSystem.File.Exists(fullPath) && fileSystem.File.ReadAllText(fullPath) == xml)
```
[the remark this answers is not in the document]
duncanp-sonar: I think the current approach is fine. Only writing the file when we need to (i.e. missing or different) reduces the chances of us trying to write the file at the same time that is being read on another thread.

Having said that, pretty much any file operation could fail, so it would be worth wrapping all of the IO calls in an exception handler to catch and log any errors. And if we are having to add logging, we might as well add some verbose logging to help diagnose the other code paths i.e. directory being created, file already exists etc. (I wouldn't add tests for the verbose logging though)

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
```diff
@@ -0,0 +1,105 @@
...
+        public ImportBeforeFileGenerator() : this(new FileSystem()) { }
+
+        public ImportBeforeFileGenerator(IFileSystem fileSystem)
+        {
+            this.fileSystem = fileSystem;
+
+            CreateTargetsFileIfNotExists();
+        }
+
+        private void CreateTargetsFileIfNotExists()
```
duncanp-sonar, line 44: The logic in the two methods `CreateTargetsFileIfNotExists` and `WriteFile` is fine, but the way it is split between the two methods seems a bit odd.

In spite of it's name, the `CreateTargetsFileIfNotExists` doesn't actually do any of the checking for directory/file existence - it's all done in `WriteFile`.

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
```diff
@@ -0,0 +1,105 @@
...
+            fileSystem.File.WriteAllText(fullPath, xml);
+        }
+
+        private XmlWriter Create(StringBuilder sb)
+        {
+            var stringWriter = new StringWriter(sb);
+
+            var settings = new XmlWriterSettings
+            {
+                ConformanceLevel = ConformanceLevel.Fragment,
```
duncanp-sonar, line 86: We're writing a complete file so I think the `ConformanceLevel` should be `Document`. However, I wouldn't use a an XmlWriter here for the reasons explained above.

src/ConnectedMode.UnitTests/Binding/ImportsBeforeFileGeneratorTests.cs
```diff
@@ -0,0 +1,103 @@
...
+{
+    [TestClass]
+    public class ImportsBeforeFileGeneratorTests
+    {
+        [TestMethod]
+        public void FileDoesNotExist_CreatesFile()
+        {
+            string pathToDirectory = GetPathToImportBefore();
+            string pathToFile = Path.Combine(pathToDirectory, "SonarLint.targets");
+            string expectedResult = "<Project xmlns=\"http://schemas.microsoft.com/developer/msbuild/2003\">\r\n  <ItemGroup />\r\n</Project>";
```
duncanp-sonar, line 36: Checking the exact file content each time makes the tests quite fragile, as we'll have to keep updating them as we change the content.

It might be simpler to have one test that checks that the content is a valid MSBuild project file i.e. it can be parsed as XML and starts with a `Project` element with the expected attribute.

The other tests could then just check that the file system method are called as expected, without explicitly checking the content.

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
duncanp-sonar: e.g. from SimpleRuleHelpBuilderTests.cs

```
        private static string ReadResource(string fullResourceName)
        {
            using var stream = new StreamReader(ResourceAssembly.GetManifestResourceStream(fullResourceName));
            return stream.ReadToEnd();
        }

```

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
```diff
@@ -21,77 +21,79 @@
...
-        private XmlWriter Create(StringBuilder sb)
+        private string ReadResourceFile()
         {
-            var stringWriter = new StringWriter(sb);
+            logger.LogVerbose(Resources.ImportBeforeFileGenerator_LoadingResourceFile);
 
-            var settings = new XmlWriterSettings
+            var resourcePath = "SonarLint.VisualStudio.ConnectedMode.Embedded.SonarLintTargets.txt";
+
+            using (var stream = GetType().Assembly.GetManifestResourceStream(resourcePath))
```
bigfluffycookie, line 84: This could just be 

``
var resourcePath = "SonarLint.VisualStudio.ConnectedMode.Embedded.SonarLintTargets.txt";
using var stream = new 
                       StreamReader(typeof(ImportBeforeFileGenerator).Assembly.GetManifestResourceStream(resourcePath));

return stream.ReadToEnd();
```

instead of all this but I saw it done like this in a different file and as it is safe I suppose this makes more sense

File: src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs [hunk 8]
```diff
@@ -21,77 +21,79 @@
...
+                {
+                    using (var reader = new StreamReader(stream))
+                    {
+                        var data = reader.ReadToEnd();
+                        return data;
+                    }
+                }
+            }
+
+            return "";
```
bigfluffycookie, line 96: I am not actually handling this case, and I am unsure if I should. Can I assume this will not fail considering it is our own text file? Or should I check IfNullOrEmpty?

src/ConnectedMode/ConnectedMode.csproj
```diff
@@ -26,6 +26,14 @@
...
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Remove="Embedded\SonarLintTargets.txt" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Include="Embedded\SonarLintTargets.txt" />
+  </ItemGroup>
+
```
duncanp-sonar, line 36: Minor: I would merge these into the same `ItemGroup` to make it more obvious they are related.

Also, I think that adding a `Label` to the `ItemGroup` will stop Visual Studio from adding/removing unrelated entries from it when it is making changes to the project file behind the scenes e.g.

```
  <ItemGroup Label="Global targets file">
    <None Remove="Embedded\SonarLintTargets.txt" />
    <EmbeddedResource Include="Embedded\SonarLintTargets.txt" />
  </ItemGroup>
```

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
```diff
@@ -0,0 +1,107 @@
...
+                logger.WriteLine(Resources.ImportBeforeFileGenerator_FailedToWriteFile);
+                logger.LogVerbose(Resources.ImportBeforeFileGenerator_FailedToWriteFile_Verbose, ex.Message);
+            }
+        }
+
+        private string ReadResourceFile()
+        {
+            logger.LogVerbose(Resources.ImportBeforeFileGenerator_LoadingResourceFile);
+
+            var resourcePath = "SonarLint.VisualStudio.ConnectedMode.Embedded.SonarLintTargets.txt";
```
duncanp-sonar, line 82: There's probably a code smell saying this should be a constant.

By convention, we normal put constants like these at the top of the class to make them easy to find.

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
duncanp-sonar: I thought `GetManifestResourceStream` throws if it can't find the specified resource.
If that is the case, your simpler version of the code in the comment above would be better.

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
```diff
@@ -0,0 +1,107 @@
...
+            this.fileSystem = fileSystem;
+
+            WriteTargetsFileToDiskIfNotExists();
+        }
+
+        private void WriteTargetsFileToDiskIfNotExists()
+        {
+            logger.LogVerbose(Resources.ImportBeforeFileGenerator_CheckingIfFileExists);
+
+            var resource = ReadResourceFile();
```
duncanp-sonar, line 50: Minor naming suggestions: I'd probably name the variable and the method differently e.g.

```
  var fileContent = GetTargetFileContent();
```

Calling the variable `fileContent` rather than `resource` makes it clearer later in the method what the variable contains
e.g.
```
  fileSystem.File.WriteAllText(fullPath, fileContent);

  // vs

  fileSystem.File.WriteAllText(fullPath, resource);
```


Calling the method `ReadResourceFile` tells the reader where the data is coming from, but not what it represents.

src/ConnectedMode/Resources.resx
```diff
@@ -242,4 +242,25 @@
...
+    <value>[ConnectedMode/ImportsBeforeFileGenerator] Creating directory {0}</value>
+  </data>
+  <data name="ImportBeforeFileGenerator_FailedToWriteFile" xml:space="preserve">
+    <value>[ConnectedMode] Failed to write file to disk.</value>
+  </data>
+  <data name="ImportBeforeFileGenerator_FailedToWriteFile_Verbose" xml:space="preserve">
+    <value>[ConnectedMode/ImportsBeforeFileGenerator] Failed to write file to disk with exception:</value>
+  </data>
+  <data name="ImportBeforeFileGenerator_FileAlreadyExists" xml:space="preserve">
+    <value>[ConnectedMode/ImportsBeforeFileGenerator] Cancelled file writing as identical file was found.</value>
```
duncanp-sonar, line 258: Minor: I'd probably say "Skipped" rather than "Cancelled" - "Cancelled" sounds more serious.

src/ConnectedMode/Resources.resx
```diff
@@ -242,4 +242,25 @@
   <data name="Suppression_IssueChangedEvent" xml:space="preserve">
     <value>[Suppressions] Handling IssueChangedEvent, isResolved: {0}, issue keys for current branch: {1}</value>
   </data>
+  <data name="ImportBeforeFileGenerator_CheckingIfFileExists" xml:space="preserve">
+    <value>[ConnectedMode/ImportsBeforeFileGenerator] Checking if file SonarLint before imports targets file exists</value>
```
duncanp-sonar, line 246: I'd probably include the full path to the file being checked in this message to help us with debugging/support e.g.

`[ConnectedMode/ImportsBeforeFileGenerator] Checking if file SonarLint before imports targets file exists. File: {foo}`

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
```diff
@@ -0,0 +1,107 @@
...
+            catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+            {
+                logger.WriteLine(Resources.ImportBeforeFileGenerator_FailedToWriteFile);
+                logger.LogVerbose(Resources.ImportBeforeFileGenerator_FailedToWriteFile_Verbose, ex.Message);
+            }
+        }
+
+        private string ReadResourceFile()
+        {
+            logger.LogVerbose(Resources.ImportBeforeFileGenerator_LoadingResourceFile);
```
duncanp-sonar, line 80: Minor: this is a purely internal implementation detail that should never fail (unlike e.g. reading/writing from the file system).  Since it won't fail, logging it won't help us with debug/support, so it's probably not worth logging.

src/ConnectedMode.UnitTests/Binding/ImportsBeforeFileGeneratorTests.cs
```diff
@@ -0,0 +1,158 @@
...
+        public void ThrowsNonCriticalException_Catches()
+        {
+            var fileSystem = new Mock<IFileSystem>();
+            fileSystem.Setup(x => x.Directory.Exists(It.IsAny<String>())).Throws(new NotImplementedException("this is a test"));
+
+            var logger = new TestLogger();
+
+            CreateTestSubject(logger: logger, fileSystem: fileSystem.Object);
+
+            logger.AssertPartialOutputStringExists(Resources.ImportBeforeFileGenerator_FailedToWriteFile);
```
duncanp-sonar, line 109: Minor: should also check we log the error message "this is a test"

duncanp-sonar (approved):
Looks good. I've made a number of comments, but they are all about fairly nit-picky things.

PR #4131, opened 2023-04-25
Remove NuGet steps from binding process

duncanp-sonar:
- most of the implement was removed in a previous commit. This removes the entry points that called the now-removed code.

Part of #4120

### DIFF
--- a/src/Integration/Binding/BindingProcessImpl.cs
+++ b/src/Integration/Binding/BindingProcessImpl.cs
@@ -197,12 +197,6 @@ namespace SonarLint.VisualStudio.Integration.Binding
             return configurationPersister.Persist(bound, bindingMode);
         }
 
-        public void PrepareToInstallPackages()
-        {
-            // TODO - CM cleanup
-            // no-op
-        }
-
         public async Task<bool> SaveServerExclusionsAsync(CancellationToken cancellationToken)
         {
             try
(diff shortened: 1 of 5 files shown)


src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -218,12 +212,6 @@ public async Task<bool> SaveServerExclusionsAsync(CancellationToken cancellation
             return true;
         }
 
-        public void InstallPackages(IProgress<FixedStepsProgress> progress, CancellationToken cancellationToken)
-        {
-            // TODO - CM cleanup
-            this.InternalState.BindingOperationSucceeded = true;
```
duncanp-sonar, line 224: The initials state of `BindingOperationSucceeded` is `true`, so removing this didn't have any effect.

All of the remaining tests pass, and I also manually bound and updated a project to check it suceeded.

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -197,12 +197,6 @@ private BindingConfiguration QueueWriteBindingInformation()
             return configurationPersister.Persist(bound, bindingMode);
         }
 
-        public void PrepareToInstallPackages()
-        {
-            // TODO - CM cleanup
-            // no-op
-        }
```
duncanp-sonar, line 204: This step and `InstallPackages` are now effectively no-ops, as I'd removed the actual implementations in a previous commit. 

This commit just cleans up `IBindingProcess` interface and top-level implementation by removing the now-redundant steps.

bigfluffycookie (approved):

PR #4132, opened 2023-04-26
Remove project binder implementations

duncanp-sonar:
- new binding method won't make project-level changes

parrt of #4120


src/Integration/Binding/ProjectBinderFactory.cs
```diff
@@ -47,21 +47,7 @@ internal ProjectBinderFactory(IServiceProvider serviceProvider, ILogger logger,
...
-            }
-
-            var isCFamilyLanguage = languages.Contains(Core.Language.C) || languages.Contains(Core.Language.Cpp);
-
-            if (isCFamilyLanguage)
-            {
-                return new CFamilyProjectBinder(projectToLanguageMapper, logger, fileSystem);
-            }
-
+            // TODO - CM cleanup - remove IProjectBinderFactory
```
duncanp-sonar, line 50: We will no longer make project-level changes, so we don't need any of the project-level binding code.

There is a lot of it, so I'll remove it in stages to keep the PRs manageable.
This first PR just removes the directly implementations of `IProjectBinder` and their tests.

Next, I'll clean up the calling code i.e. get rid of `IProjectBinderFactory` and fix up the code that calls it.

Then, I'll remove the lower-level classes that are no longer used i.e. the classes that were only used by the `CSharpVBProjectBinder`.

src/Integration/Binding/ProjectBinderFactory.cs
```diff
@@ -47,21 +47,7 @@ internal ProjectBinderFactory(IServiceProvider serviceProvider, ILogger logger,
...
-                return new CSharpVBProjectBinder(serviceProvider, projectToLanguageMapper, fileSystem, logger);
-            }
-
-            var isCFamilyLanguage = languages.Contains(Core.Language.C) || languages.Contains(Core.Language.Cpp);
-
-            if (isCFamilyLanguage)
-            {
-                return new CFamilyProjectBinder(projectToLanguageMapper, logger, fileSystem);
-            }
-
```
duncanp-sonar, line 64: Removing this code caused a number of new code smells relating to unused variables.
I'm going to delete the entire class, so the code smells will go at that point.

bigfluffycookie (approved):

PR #4133, opened 2023-04-26
Drop IProjectBinderFactory

duncanp-sonar:
- new connected mode does not have project-level bindings

Part of #4120


duncanp-sonar:
The previous commit removed the implementations of `IProjectBinder` and changed `IProjectBinderFactory` to return an empty list.

This commit removes `IProjectBinderFactory` and fixes up the code in `BindingProcessImpl` and `SolutionBindingOperation` that used it.

The next commit will delete any lower-level classes that were previously only used by the delete `IProjectBinder` implementations.

#### Current state of the clean up
We no longer write any project-level files. The rest of the binding process still works i.e. all of the existing solution level files are still created:

image

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -216,10 +216,7 @@ public void InitializeSolutionBindingOnBackgroundThread()
         {
             this.solutionBindingOperation.RegisterKnownConfigFiles(this.InternalState.BindingConfigs);
 
-            var projectsToUpdate = GetProjectsForRulesetBinding(this.InternalState.IsFirstBinding, this.InternalState.BindingProjects.ToArray(),
-                this.host.Logger, this.threadHandling);
-
-            this.solutionBindingOperation.Initialize(projectsToUpdate);
+            this.solutionBindingOperation.Initialize();
```
duncanp-sonar, line 219: Previously `BindingProcessImpl` passed a list of projects to `SolutionBindingOperation`.

However, `SolutionBindingOperation` not longer uses the project list, so I removed the code from this class that calculated the set of projects.

src/Integration/Binding/ProjectBinderFactory.cs
```diff
@@ -1,54 +0,0 @@
...
-            this.serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
-            this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
-            this.fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
-            projectToLanguageMapper = serviceProvider.GetMefService<IProjectToLanguageMapper>();
-        }
-
-        public IProjectBinder Get(Project project)
-        {
-            // TODO - CM cleanup - remove IProjectBinderFactory
-            return null;
```
duncanp-sonar, line 51: This class was effectively a no-op after the last commit.

This commit removes it and fixes up the affected callers.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -163,34 +151,13 @@ public void Prepare(CancellationToken token)
...
             }
-
-            foreach (var project in projects)
-            {
-                if (token.IsCancellationRequested)
-                {
-                    return;
-                }
-
-                var projectBinder = projectBinderFactory.Get(project);
```
duncanp-sonar, line 174: After the last commit, `IProjectBinderFactory.Get(...)` always returned `null`.

This class didn't do anything else with the supplied list of `projects` except pass them to the factory to get binders, so I removed all of the code that referenced `projects` from this class too.

duncanp-sonar:
The Quality Gate is failing because there is one uncovered line of code (out of two). I'm not going to add more test until I've finished deleting code.

bigfluffycookie (approved):

PR #4139, opened 2023-04-26
Drop C#/VB.NET project binder helper types

duncanp-sonar:
- delete low-level types no longer used by the product code

Part o #4120


src/Integration.UnitTests/LocalServices/SolutionRuleSetsInformationProviderTests.cs
```diff
@@ -84,7 +84,7 @@ public void SolutionRuleSetsInformationProvider_GetProjectRuleSetsDeclarations_A
         public void SolutionRuleSetsInformationProvider_GetProjectRuleSetsDeclarations_ConfigurationPropertyWithDefaultValue()
         {
             // Arrange
-            PropertyMock prop1 = CreateProperty(this.projectMock, "config1", CSharpVBBindingOperation.DefaultProjectRuleSet);
```
duncanp-sonar, line 87: None of the classes that were dropped were required any longer (the only product code that used them had already been dropped). So this commit is mostly to drop the unused code.

This test method is the only place that needed to be changed as it referenced a constant from one of the classes. However, it's just a test and could have used any value - it didn't need to reference the constant.
So I changed it to another value.

bigfluffycookie (approved):

PR #4124, opened 2023-04-24
Drop UnboundProjectFinder

duncanp-sonar:
- new CM won't require project-level changes

Part of #4120

PR #4143, opened 2023-04-26
(2) Change binding to fetch QPs for all supported languages

duncanp-sonar:
Part of #4124

### DIFF
--- a/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
@@ -53,7 +53,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         private ConfigurableServiceProvider serviceProvider;
         private Mock<ISonarQubeService> sonarQubeServiceMock;
         private ConfigurableVsProjectSystemHelper projectSystemHelper;
-        private ConfigurableSolutionRuleSetsInformationProvider ruleSetsInformationProvider;
         private ConfigurableHost host;
         private Mock<IFolderWorkspaceService> folderWorkspaceService;
         private Mock<IProjectToLanguageMapper> projectToLanguageMapper;
@@ -72,13 +71,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
 
             var mockFileSystem = new MockFileSystem();
             var sccFileSystem = new ConfigurableSourceControlledFileSystem(mockFileSystem);
-            var ruleSerializer = new ConfigurableRuleSetSerializer(mockFileSystem);
-            this.ruleSetsInformationProvider = new ConfigurableSolutionRuleSetsInformationProvider();
 
             this.serviceProvider.RegisterService(typeof(ISourceControlledFileSystem), sccFileSystem);
-            this.serviceProvider.RegisterService(typeof(IRuleSetSerializer), ruleSerializer);
             this.serviceProvider.RegisterService(typeof(IProjectSystemHelper), this.projectSystemHelper);
-            this.serviceProvider.RegisterService(typeof(ISolutionRuleSetsInformationProvider), this.ruleSetsInformationProvider);
 
             projectToLanguageMapper = new Mock<IProjectToLanguageMapper>();
             folderWorkspaceService = new Mock<IFolderWorkspaceService>();
--- a/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
@@ -22,18 +22,15 @@ using System;
 using System.Collections.Generic;
 using System.IO.Abstractions.TestingHelpers;
 using System.Linq;
-using System.Text;
 using System.Threading;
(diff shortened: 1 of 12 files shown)


src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -46,31 +43,22 @@ internal class BindingProcessImpl : IBindingProcess
         private readonly ISolutionBindingOperation solutionBindingOperation;
         private readonly IBindingConfigProvider bindingConfigProvider;
         private readonly SonarLintMode bindingMode;
-        private readonly IProjectToLanguageMapper projectToLanguageMapper;
-        private readonly IThreadHandling threadHandling;
```
duncanp-sonar, line 50: Miscellaneous: `IThreadHandling` wasn't used so I removed it

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -241,64 +219,8 @@ public void SilentSaveSolutionIfDirty()
 
         #region Private methods
 
-        private void InformAboutFilteredOutProjects()
```
duncanp-sonar, line 244: We used to log a list of the "unsupported" projects i.e. those that use languages we don't support.
There's no point now we fetch settings for all languages.

Also, `Secrets` analysis is run against every text file, so there probably isn't such a thing as a project with an unsupported language type any more.

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -75,16 +70,6 @@ public void TestInitialize()
...
 
-            projectToLanguageMapper = new Mock<IProjectToLanguageMapper>();
-            folderWorkspaceService = new Mock<IFolderWorkspaceService>();
-
-            var mefProjectToLanguageMapper =
-                ConfigurableComponentModel.CreateWithExports(
-                    MefTestHelpers.CreateExport<IProjectToLanguageMapper>(projectToLanguageMapper.Object),
-                    MefTestHelpers.CreateExport<IFolderWorkspaceService>(folderWorkspaceService.Object));
-
-            serviceProvider.RegisterService(typeof(SComponentModel), mefProjectToLanguageMapper);
```
duncanp-sonar, line 86: The class no longer MEF-imports either `IProjectToLanguageMapper` or `IFolderWorkspaceService`

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -225,11 +210,8 @@ public async Task DownloadQualityProfile_Success()
             var testSubject = this.CreateTestSubject(bindingArgs, configProviderMock.Object);
 
             host.SupportedPluginLanguages.Add(language);
-            var project = Mock.Of<Project>();
-            testSubject.InternalState.BindingProjects.Add(project);
```
duncanp-sonar, line 229: For the tests to work with the old code, we had to set both `host.SupportedPluginLanguages` and also configure the `projectToLanguageMapper` to return the languages.

To fix the existing tests I just had to deleted the setup for `IProjectToLanguageMapper` since it is no longer called.

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -305,16 +282,12 @@ public async Task DownloadQualityProfile_WhenQualityProfileIsNotAvailable_Fails(
             var notifications = new ConfigurableProgressStepExecutionEvents();
             var progressAdapter = new FixedStepsProgressAdapter(notifications);
 
+            // CSharp is a supported language, but only the VB.NET QP is available
```
duncanp-sonar, line 285: Misc: new comment explaining the setup of the test. Not related to the current changes.

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -376,85 +344,14 @@ public async Task DownloadQualityProfile_WhenBindingConfigIsNull_Fails()
         }
 
         [TestMethod]
-        public void GetBindingLanguages_ReturnsDistinctLanguagesForProjects()
+        public void GetBindingLanguages_ReturnsExpectedLanguagese()
```
duncanp-sonar, line 347: I simplified the existing test to check that the supported languages match the ones returned by the host.

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -513,183 +410,6 @@ public void SilentSaveSolutionIfDirty()
             logger.AssertOutputStrings(0);
         }
 
-        [TestMethod]
-        public void DiscoverBindableProjects_AddsMatchingProjectsToBinding()
```
duncanp-sonar, line 517: I just deleted all of the `DiscoverBindableProject_XXX` tests since the method is now hard-coded to return `true`.
I'll remove it completely in the next PR.

bigfluffycookie (approved):

PR #4148, opened 2023-04-27
Add projection buffer filtering (fixes analysis bug)

georgii-borovinskikh-sonarsource:


src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -152,6 +153,11 @@ private void OnAnalysisRequested(object sender, AnalysisRequestEventArgs args)
         /// </summary>
         public ITagger<T> CreateTagger<T>(ITextBuffer buffer) where T : ITag
         {
+            if (buffer is IProjectionBuffer)
+            {
+                return null;
+            }
```
duncanp-sonar, line 159: Could you have made the fix in `ITaggableBufferIndicator` instead rather than repeating the check?
Or would that affect too many taggers?

duncanp-sonar (commented):
Question about whether `ITaggableBufferIndicator` would be a better place to make the fix.

ugras-ergun-sonarsource (approved):

ugras-ergun-sonarsource:
Maybe just add an comment as a reminder why do we ignore it.

PR #4150, opened 2023-04-27
Add error handling for css analysis

georgii-borovinskikh-sonarsource:


src/TypeScript/Analyzer/EslintBridgeAnalyzer.cs
```diff
@@ -117,6 +122,18 @@ internal sealed class EslintBridgeAnalyzer : IEslintBridgeAnalyzer
             return issues;
         }
 
+        private bool IsCssError(string filePath, AnalysisResponse analysisResponse)
+        {
+            if (analysisResponse is CssAnalysisResponse cssAnalysisResponse && cssAnalysisResponse.Error != null)
+            {
+                logger.WriteLine("Failed to analyze CSS in {0}", filePath);
```
duncanp-sonar, line 129: The non-verbose messages should be resource strings.

duncanp-sonar (approved):
Looks good, apart from the non-verbose message needing to be in a resource file (same as the other messages used by the class)

PR #4153, opened 2023-04-27
Drop ILegacyConfigFolderItemAdder

duncanp-sonar:
- auto-migration to new-new Connected Mode means saving in legacy connected mode is no longer supported

Part of #4120


src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -153,38 +150,8 @@ public void Prepare(CancellationToken token)
         }
 
         public bool CommitSolutionBinding()
-        {
```
duncanp-sonar, line 156: We used to support running an action after the main part of the "commit binding" had been done, which was only every used in "legacy connected mode".

Since we're auto-migrating settings to the new-new Connected Mode format, we don't need to support saving in this legacy format.

bigfluffycookie (approved):
Looks good except for the code smells :)

duncanp-sonar:
@bigfluffycookie could you have a look at the code smells cleanup in the second commit please?

After the first commit, the struct `WriteSettings` only had one field left in it (holding the directory path), so I've got rid of the struct completely.

src/Integration/NewConnectedMode/IConfigurationPersister.cs
```diff
@@ -95,14 +92,6 @@ private BindingConfiguration CreateBindingConfiguration(string bindingPath, Boun
                     return null;
                 }
             }
-
-            return writeSettings;
-        }
-
-        private struct WriteSettings
-        {
-            public Action<string> OnSuccessfulFileWrite { get; set; }
```
duncanp-sonar, line 115: The `OnSuccessfulFileWrite` field was not being set, so I removed the `WriteSettings` struct and changed the `GetWriteSettings` method just to return the directory path.

src/Integration/Persistence/ISolutionBindingDataWriter.cs
```diff
@@ -29,7 +29,7 @@ internal interface ISolutionBindingDataWriter
         /// Writes the binding information
         /// </summary>
         /// <returns>Has file been saved</returns>
-        bool Write(string configFilePath, BoundSonarQubeProject binding, Action<string> onSuccessfulFileWrite);
```
duncanp-sonar, line 32: The action parameter was what used to stored in the `WriteSettings.OnSuccessfulFileWrite` field.
It wasn't being used so I removed it.

Issue #4147, opened 2023-04-27
[Binding] Add a class that creates a .globalconfig file with the rule settings.


bigfluffycookie:
Create a GlobalConfigGenerator that will be used instead of the RuleSetGenerator in the future.

PR #4158, opened 2023-05-02
Add GlobalConfigGenerator Class

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4147


src/ConnectedMode/Binding/GlobalConfigGenerator.cs
```diff
@@ -0,0 +1,117 @@
...
+            foreach (var rule in sortedRules)
+            {
+                var severityText = (rule.IsActive) ? GetActionText(GetVsSeverity(rule.Severity)) : inactiveRuleActionText;
+
+                sb.AppendLine($"dotnet_diagnostic.{rule.Key}.severity = {severityText}");
+            }
+
+            return sb.ToString();
+        }
+
```
bigfluffycookie, line 72: The below to methods were copied from RuleSetGenerator

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -1,7 +1,7 @@
 ﻿---
 ################################
 # Assembly references report
-# Report date/time: 2023-04-24T16:00:30.3059365Z
+# Report date/time: 2023-05-02T09:14:46.3435124Z
```
duncanp-sonar, line 4: It doesn't do any harm to check these files in, but I don't normally bother if the only change is the date/time and file path (if anything else has changed then we have to check them in otherwise the CI build will fail).

I wouldn't bother removing the files from this PR, though.

bigfluffycookie: My bad, in the individual commits it looked like more had changed, didnt realize the second commit reverted the first commits change.

src/ConnectedMode/Binding/GlobalConfigGenerator.cs
```diff
@@ -0,0 +1,117 @@
...
+                case SonarQubeIssueSeverity.Blocker:
+                    return environmentSettings.TreatBlockerSeverityAsError() ? RuleAction.Error : RuleAction.Warning;
+
+                default:
+                    throw new NotSupportedException($"Unsupported SonarQube issue severity: {sqSeverity}");
+            }
+        }
+
+        internal /* for testing */ static string GetActionText(RuleAction ruleAction)
```
duncanp-sonar, line 93: It wouldn't hurt to include a link to the MS docs page about severities: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-options#severity-level

src/ConnectedMode/Binding/GlobalConfigGenerator.cs
```diff
@@ -0,0 +1,117 @@
...
+                    return "Info";
+
+                case RuleAction.Warning:
+                    return "Warning";
+
+                case RuleAction.Error:
+                    return "Error";
+
+                case RuleAction.Hidden:
+                    return "Hidden";
```
duncanp-sonar, line 110: It looks from the docs that `editorconfig` uses different values from those used in `RuleSet`s.

* all of the names are lower-case in docs
* `Info` and `Hidden` don't exist. Perhaps should be `suggestion` and `silent` instead?

src/ConnectedMode/Binding/GlobalConfigGenerator.cs
duncanp-sonar: `git blame` says I wrote those methods originally. Looking at them now, I think I'd write them differently - they look more complicated than necessary. Now, I'd probably have a single translation method that handled the translation and active/inactive translation. However, I wouldn't change it at this point.

src/ConnectedMode.UnitTests/Binding/GlobalConfigGeneratorTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            act.Should().Throw<NotSupportedException>();
+        }
+
+        private static SonarQubeRule CreateRule(string ruleKey, string repoKey, bool isActive = true, SonarQubeIssueSeverity sqSeverity = SonarQubeIssueSeverity.Info) =>
+            new SonarQubeRule(ruleKey, repoKey, isActive, sqSeverity, new Dictionary<string, string>(), SonarQubeIssueType.Unknown, null, null, null, null, null, null);
+
+        private string GetRuleString(string expectedKey, string expectedSeverity) =>
+            $"dotnet_diagnostic.{expectedKey}.severity = {expectedSeverity}";
+
+        private string GetStartOfGlobalConfig() => $"is_global = true\nglobal_level=500000";
```
duncanp-sonar, line 141: Is this method used?

duncanp-sonar (commented):
See inline - I think a couple of the `ActionText` values need to be tweaked.

Issue #4149, opened 2023-04-27
Add ImportsBefore targets file content


duncanp-sonar:
### Description

Add the boilerplate content to the ImportBefore file.

The file will be imported into every build, so it should used conditions so that it is a no-op unless it is being used inside Visual Studio

#### TODO
 -[ ] detect the appropriate language (C# or VB.NET)
 -[ ] reference the global config if it exists
 -[ ] reference the additional file if it exists
 -[ ]

PR #4159, opened 2023-05-02
Add ImportBefore targets content

duncanp-sonar:
Part of #4149

### DIFF
--- a/src/ConnectedMode/Embedded/SonarLintTargets.xml
+++ b/src/ConnectedMode/Embedded/SonarLintTargets.xml
@@ -1,4 +1,59 @@
 ﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <ItemGroup>
+<!-- TODO 
+
+* decide the final root directory location
+* remove the debug Target
+* decide whether to include $(DesignTimeBuild) in the Conditions.
+
+-->
+
+  <!-- 
+      This targets file was added by SonarLint for Visual Studio.
+      If you are no longer using SonarLint for Visual Studio then you can safely delete it.
+      
+      It is only active for builds taking place inside Visual Studio, and does not impact the content of the assembly being built.
+ 
+      Its purpose is to ensure the correct configuration for the Sonar C# and Sonar VB.NET Roslyn rules is used
+      when a solution is in Connected Mode i.e. when it is bound to a SonarQube or SonarCloud project.
+      
+  -->
+  
+  <PropertyGroup Condition="$(BuildingInsideVisualStudio) == 'true'">
+    <!-- C#/VB.NET settings need to be in different folders -->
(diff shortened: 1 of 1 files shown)


duncanp-sonar:
This PR contains a initial version of the targets file that will allow us to move forward with getting an end-to-end build working.

There are a number of decisions we still need to make about e.g. where to drop the file, documented in this TODO. I've also added a separate task to the build for writing unit tests (past experience of writing tests using the MSBuild APIs suggests this might take a while).

For now, this initial version looks under `%LOCALAPPDATA%`. I've manually tested it inside and outside VS with C# and VB.NET and it seems to work as expected (see the screenshots of the MSBuild logs below).

### C#, not building inside VS -> no-op
image

### C#, building inside VS -> files exist and are detected
image

### VB.NET, building inside VS -> files exist and are detected
image

... and the correct settings are passed to the compiler (note that there is another target called `CompileDesignTime` - the same settings are passed to that target too)

image

src/ConnectedMode/Embedded/SonarLintTargets.xml
```diff
@@ -1,4 +1,59 @@
...
+      It is only active for builds taking place inside Visual Studio, and does not impact the content of the assembly being built.
+ 
+      Its purpose is to ensure the correct configuration for the Sonar C# and Sonar VB.NET Roslyn rules is used
+      when a solution is in Connected Mode i.e. when it is bound to a SonarQube or SonarCloud project.
+      
+  -->
+  
+  <PropertyGroup Condition="$(BuildingInsideVisualStudio) == 'true'">
+    <!-- C#/VB.NET settings need to be in different folders -->
+    <_SLVSRootFolder>$(LOCALAPPDATA)\SonarLint\VS\$(SolutionName)_$(SolutionDir.GetHashCode())\$(Language)</_SLVSRootFolder>
```
duncanp-sonar, line 23: Once we decide the final location for the files, we should just need to update this build property.

src/ConnectedMode/Embedded/SonarLintTargets.xml
```diff
@@ -1,4 +1,59 @@
...
   </ItemGroup>
-</Project>
\ No newline at end of file
+
+  
+  
+  
+  
+ 
+  <Target Name="SLVS_DebugOutput_TODO_Remove" BeforeTargets="CoreCompile" >
```
duncanp-sonar, line 39: This target can be removed, leaving only the comment and the Properties/ItemGroups.

src/ConnectedMode/Embedded/SonarLintTargets.xml
```diff
@@ -1,4 +1,59 @@
...
-  <ItemGroup>
+<!-- TODO 
+
+* decide the final root directory location
+* remove the debug Target
+* decide whether to include $(DesignTimeBuild) in the Conditions.
+
+-->
+
+  <!--
```
duncanp-sonar, line 10: The `TODO` comment above should be removed.

The comment below would remain so anyone looking at the file knows what it is for.
We might want to include a link to the wiki/docs.

bigfluffycookie (approved):

PR #4165, opened 2023-05-02
Drop use of ISourceControlledFileSystem from product code

duncanp-sonar:
- unintrusive Connected Mode does not save files in a source-controlled location

Part of #4120


src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -129,25 +121,18 @@ public void Prepare(CancellationToken token)
 
                 var info = keyValue.Value;
 
-                sourceControlledFileSystem.QueueFileWrites(info.SolutionLevelFilePaths, () =>
```
duncanp-sonar, line 132: This is the key part of the change - we're no longer queuing the changes, we're writing them directly.

Issue #4152, opened 2023-04-27
[Installation] Trigger install/update of ImportBefore targets 


duncanp-sonar:
### Description

We can't install our targets file under `MSBuild\...\ImportBefore` as part of the VSIX installation, so we'll have to do it from code.

Strictly speaking it only needs to be done:
* for C# or VB.NET projects,
* when using Connected Mode

For simplicity, we could decide to always install on startup

#### TODO
-[ ] decide when to trigger the check
-[ ] trigger the install component
-[ ] tests

PR #4167, opened 2023-05-03
Add ImportBeforeInstallTrigger to trigger ImportFileGenerator

bigfluffycookie:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4152


src/ConnectedMode/Binding/PreSolutionBoundUpdateHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        private bool disposed;
+
+        [ImportingConstructor]
+        public PreSolutionBoundUpdateHandler(IActiveSolutionBoundTracker activeSolutionBoundTracker, IImportBeforeFileGenerator importBeforeFileGenerator)
+        {
+            this.activeSolutionBoundTracker = activeSolutionBoundTracker;
+            this.importBeforeFileGenerator = importBeforeFileGenerator;
+
+            this.activeSolutionBoundTracker.PreSolutionBindingChanged += OnPreSolutionBindingChanged;
```
bigfluffycookie, line 42: I think my logic might be a bit flawed here, as the TriggerUpdate() will not be called when a user opens an already bound project. 

I am not sure this is the desired behaviour so please lmk

duncanp-sonar: > the TriggerUpdate() will not be called when a user opens an already bound project.

I thought it was called any time a solution is opened or closed, regardless of whether it is currently bound or not. Is it definitely not called when opening a bound solution?

bigfluffycookie: I have tried it a few times now and it seems that it does not get called

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
```diff
@@ -19,32 +19,39 @@
...
+using System.ComponentModel.Composition;
 using System.IO;
 using System.IO.Abstractions;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Binding
 {
-    internal class ImportBeforeFileGenerator
+    public interface IImportBeforeFileGenerator
```
duncanp-sonar, line 30: Does this need to be `public`? Could the interface and class both be `internal`?

src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -65,6 +67,7 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
             boundSolutionUpdateHandler = componentModel.GetService<BoundSolutionUpdateHandler>();
             timedUpdateHandler = componentModel.GetService<TimedUpdateHandler>();
             localSuppressionsChangedHandler = componentModel.GetService<LocalSuppressionsChangedHandler>();
+            preSolutionBoundUpdateHandler = componentModel.GetService<PreSolutionBoundUpdateHandler>();
```
duncanp-sonar, line 70: Seeing the name of the new class here (`PreSolutionBoundUpdateHandler`) made me realise it doesn't give any real idea what the class is for. It's named after the event it listens to, rather than telling us what it does. We followed a similar naming pattern in when doing the original `Education` (e.g. `BoundSolutionUpdateHandler`, `TimedUpdateHandler`), and they suffer from the same problem.

I think it would be better if the name reflected what the class does - the fact that it handles a particular event   is an implementation detail. The problem then is coming up with a good name... `ImportBeforeInstallTrigger`? `CheckImportBeforeInstalledTrigger`? `TriggerImportBeforeCheckListener`?

src/ConnectedMode/Binding/ImportBeforeFileGenerator.cs
duncanp-sonar: The interface should have a doc comment explaining what it does

src/ConnectedMode/Binding/PreSolutionBoundUpdateHandler.cs
bigfluffycookie: Yea that seems to be the problem, I just saw this in the connectedMode package ->
```
// Trigger an initial update of suppressions (we might have missed the solution binding
// event from the ActiveSolutionBoundTracker)
// See https://github.com/SonarSource/sonarlint-visualstudio/issues/3886
var updater = componentModel.GetService<ISuppressionIssueStoreUpdater>();
updater.UpdateAllServerSuppressionsAsync().Forget();
```

but I dont think in our case that makes much sense, unless we do a check here -> "are we in connected mode" wdyt

src/ConnectedMode/Binding/PreSolutionBoundUpdateHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            if (e.Configuration.Mode != SonarLintMode.Standalone)
+            {
+                TriggerUpdate();
+            }
+        }
+
+        private void OnPreSolutionBindingUpdated(object sender, EventArgs e) => TriggerUpdate();
+
+        private void TriggerUpdate()
```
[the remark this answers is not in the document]
bigfluffycookie: Should we switch to the background thread in here or in the ImportBeforeFileGenerator class?

src/ConnectedMode/Binding/PreSolutionBoundUpdateHandler.cs
duncanp-sonar: Yep. We could move the "is connected" check into the `TriggerUpdate` method of this class and then just have the package import the class and call `TriggerUpdate`. That would mean re-writing the `TriggerUpdate` method to look something like this:

```
public Task TriggerUpdateAsync()
{
    var config = configurationProvider.GetConfiguration();

    if (e.Configuration.Mode != SonarLintMode.Standalone)
    {
        await threadHandling.SwitchToBackGroundThread();

        // This should only happen once which is why it gets disposed after.
        importBeforeFileGenerator.WriteTargetsFileToDiskIfNotExists();
        Dispose();
    }
}
```

The simpler alternative would be to always install the targets, even when we are not in Connected Mode (which is what @rita-gorokhod originally suggested). If we did that then we could just have the package create and call `IImportBeforeFileGenerator` directly.

What do you think?

bigfluffycookie: I went with the first approach, simply because considering most of our users do not use connected made I feel like its better to not always put the files.

src/ConnectedMode.UnitTests/Binding/ImportBeforeInstallTriggerTests.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.VisualStudio.Threading;
+using SonarLint.VisualStudio.ConnectedMode.Binding;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.TestInfrastructure;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
+{
+    [TestClass]
+    public class ImportBeforeInstallTriggerTests
```
bigfluffycookie, line 31: Forgot to add a test to check if it switches to ui thread, my bad. Will add it standalone or with other pr fixes.

src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -47,7 +47,7 @@ public sealed class ConnectedModePackage : AsyncPackage
         private BoundSolutionUpdateHandler boundSolutionUpdateHandler;
         private TimedUpdateHandler timedUpdateHandler;
         private LocalSuppressionsChangedHandler localSuppressionsChangedHandler;
-        private PreSolutionBoundUpdateHandler preSolutionBoundUpdateHandler;
+        private ImportBeforeInstallTrigger importBeforeInstallTrigger;
```
bigfluffycookie, line 50: Also forgot to remove this

duncanp-sonar (approved):
Looks good!

PR #4175, opened 2023-05-03
Simplify SolutionBindingDataWriter construction

duncanp-sonar:
- class is not quite MEF-exportable, but it's close


src/Integration.UnitTests/Persistence/SolutionBindingDataWriterTests.cs
```diff
@@ -68,15 +68,15 @@ public void TestInitialize()
         [TestMethod]
         public void Ctor_NullSourceControlledFileSystem_Exception()
         {
-            Action act = () => new SolutionBindingDataWriter(null, null, null);
+            Action act = () => new SolutionBindingDataWriter(null, (ISolutionBindingFileLoader)null, null);
```
duncanp-sonar, line 71: The public and internal constructors have the same number of arguments, so we now need to include a cast to help the compiler decide which one to use.

Issue #4170, opened 2023-05-03
Add unintrusive path provider


duncanp-sonar:
### Description

Implement `UnintrusiveBindingPathProvider`. Returns a solution-specific location under the standard SLVS user-specific folder
i.e. `$(APPDATA)\SonarLint for Visual Studio\\Bindings\\$(SolutionName)_$(SolutionDir.GetHashCode())\binding.config`

See #4168 for more details

PR #4181, opened 2023-05-04
Add RunOnUIThreadSync method to IThreadHandling

duncanp-sonar:
Prep for #4170

### DIFF
--- a/src/Core/IThreadHandling.cs
+++ b/src/Core/IThreadHandling.cs
@@ -53,6 +53,16 @@ namespace SonarLint.VisualStudio.Core
         /// </summary>
         Task RunOnUIThread(Action op);
 
+        // TODO: renaming async methods to end with "Async", then rename this method to "RunOnUIThread"
+        // See https://github.com/SonarSource/sonarlint-visualstudio/issues/4180
+        /// <summary>
+        /// Executes the operation synchronously on the main thread.
+        /// If the caller is on the main thread already then the operation is executed directly.
+        /// If the caller is not on the main thread then the method will switch to the main thread,
+        /// then resume on the caller's thread when then the operation completes.
+        /// </summary>
+        void RunOnUIThreadSync(Action op);
+
         /// <summary>
         /// Executes the operation asynchronously on the background thread.
         /// If the caller is on a background thread already then the operation is executed directly.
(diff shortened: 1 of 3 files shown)


src/Core/IThreadHandling.cs
```diff
@@ -53,6 +53,16 @@ public interface IThreadHandling
...
 
+        // TODO: renaming async methods to end with "Async", then rename this method to "RunOnUIThread"
+        // See https://github.com/SonarSource/sonarlint-visualstudio/issues/4180
+        /// <summary>
+        /// Executes the operation synchronously on the main thread.
+        /// If the caller is on the main thread already then the operation is executed directly.
+        /// If the caller is not on the main thread then the method will switch to the main thread,
+        /// then resume on the caller's thread when then the operation completes.
+        /// </summary>
+        void RunOnUIThreadSync(Action op);
```
duncanp-sonar, line 64: Some of the existing code uses our old non-mockable `RunOnUIThread` class directly.
I've added a new abstraction to `IThreadHandling` so I don't have to do that.
This time, however, I created a couple of separate tickets for the additional work that is required, rather than trying to fix everything at once.

I will need the new method when initialising the new path provider.

src/Infrastructure.VS/ThreadHandling.cs
```diff
@@ -48,6 +48,10 @@ public class ThreadHandling : IThreadHandling
 
         public async Task RunOnUIThread(Action op) => await VS.RunOnUIThread.RunAsync(op);
 
+        // TODO: check the implementation of RunOnUIThread.Run is actually synchronous
+        // See https://github.com/SonarSource/sonarlint-visualstudio/issues/4179
+        public void RunOnUIThreadSync(Action op) => VS.RunOnUIThread.Run(op);
```
duncanp-sonar, line 53: This is a bit of a worry. However, fixing the method will change the runtime behaviour of all of the exiting callers and will need to be looked at carefully, so I created a separate ticket.

bigfluffycookie (approved):
Looks good! I think we had a ticket in the backlog somewhere about adding RunOnUIThreadSync because I remember we talked about it a few months ago (Just incase you go through the backlog at some point)

Issue #4162, opened 2023-05-02
[Binding] Change binding to call globalconfig generator


duncanp-sonar:
### Description

TODO
-[ ] Change C#/VB binder to call globalconfig generator
-[ ] Fix up tests
-[ ] Drop ruleset generator and tests

PR #4182, opened 2023-05-04
Replace RuleSetGenerator with GlobalConfigGenerator

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4162


src/Core/Language.cs
```diff
@@ -43,8 +43,8 @@ namespace SonarLint.VisualStudio.Core
     public sealed class Language : IEquatable<Language>
     {
         public readonly static Language Unknown = new Language();
```
bigfluffycookie, line 45: I feel like it makes sense to just change this here

src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
```diff
@@ -144,32 +132,29 @@ public async Task GetConfig_NoSupportedActiveRules_ReturnsNull()
...
-        public async Task GetConfig_ReturnsCorrectRuleset()
+        public async Task GetConfig_ReturnsCorrectGlobalConfig()
         {
             var builder = new TestEnvironmentBuilder(validQualityProfile, Language.VBNET)
             {
                 ActiveRulesResponse = validRules,
                 InactiveRulesResponse = emptyRules,
                 PropertiesResponse = anyProperties,
-                RuleSetGeneratorResponse = validRuleSet,
+                GlobalConfigGeneratorResponse = globalConfig,
```
duncanp-sonar, line 146: I'd set the `GlobalConfigGeneratorResponse` to a hard-coded string, and check the expected content is returned after line 174.

duncanp-sonar (commented):
Looks good - just a suggestion get rid of the class-level `private string globalConfig` from the provider tests.

PR #4185, opened 2023-05-04
Drop IErrorListInfoBarController

duncanp-sonar:
- coordinated displaying a gold bar if a binding update was required. No longer required since updates can be done without user interaction in the new Connected Mode

Part of #4120


src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -328,16 +327,6 @@ private void RegisterLocalServices()
 
                 return new ProjectSystemFilter(this, testProjectIndicator);
             }));
-            this.localServices.Add(typeof(IErrorListInfoBarController), new Lazy<ILocalService>(() =>
-                new ErrorListInfoBarController(this,
-                    new BindingChecker(
-                        new UnboundSolutionChecker(
-                            new ExclusionSettingsStorage(this.GetService<IConfigurationProviderService>(), Logger),
```
duncanp-sonar, line 335: Going forwards, `IConfigurationProvider` will only be available as a MEF service (i.e. when using the new Connected Mode settings/location).

However, I needed to do some more cleanup first.
This is part of it - dropping an unnecessary class that fetches `IConfigurationProviderService` as a service.

In the next PR I'll drop `IConfigurationProviderService` completely so that only the MEF-exported `IConfigurationProvider` is left.

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -75,9 +73,6 @@ public ActiveSolutionBoundTracker(IHost host, IActiveSolutionTracker activeSolut
             configurationProvider = extensionHost.GetService<IConfigurationProviderService>();
             configurationProvider.AssertLocalServiceIsNotNull();
 
-            errorListInfoBarController = extensionHost.GetService<IErrorListInfoBarController>();
-            errorListInfoBarController.AssertLocalServiceIsNotNull();
```
duncanp-sonar, line 79: Previously the `ActiveSolutionBoundTracker` knew about the `IErrorListInfoBarController` and called it directly when solution binding changes occurred.

This was always a hack; the `ErrorListInfoBarController` should have been listening to the events from the `ActiveSolutionBoundTracker`. I can't remember why the hack was necessary - perhaps an event-ordering problem (i.e. because we couldn't rely on the events being handled in a particular order, so the `ActiveSolutionBoundTracker` explicitly called the `ErrorListInfoBarController` after the events had been raised and handled).

Whatever the reason, it's gone now.

bigfluffycookie (approved):
Very good

PR #4186, opened 2023-05-04
Remove IConfigurationProviderService

duncanp-sonar:
- now IConfigurationProvider is only available as a MEF export
- fixed up consumers and tests

Prep for #4170


src/Integration/UnintrusiveBinding/UnintrusiveConfigurationProvider.cs
```diff
@@ -23,13 +23,9 @@
 
 namespace SonarLint.VisualStudio.Integration.NewConnectedMode
 {
-    internal interface IConfigurationProviderService : IConfigurationProvider, ILocalService
```
duncanp-sonar, line 26: This is the original of all of the rest of the changes: I've dropped the `IConfigurationProviderService` marker interface.

That meant I had to:
1) update the `VsSessionHost` so that it doesn't return it as a local service any more, and
2) changed the remaining few classes that still fetched it as a service so they now get it via MEF.

src/Integration/UnintrusiveBinding/UnintrusiveConfigurationProvider.cs
```diff
@@ -23,13 +23,9 @@
...
-    {
-    }
-
     [Export(typeof(IConfigurationProvider))]
-    [PartCreationPolicy(CreationPolicy.Any)]
-    internal class UnintrusiveConfigurationProvider : IConfigurationProviderService
+    [PartCreationPolicy(CreationPolicy.Shared)]
```
duncanp-sonar, line 27: Miscellaneous: this should have been `Shared` too

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -48,7 +47,6 @@ internal sealed class VsSessionHost : IHost, IProgressStepRunnerWrapper, IDispos
                 typeof(IProjectSystemHelper),
                 typeof(ISourceControlledFileSystem),
                 typeof(IProjectSystemFilter),
-                typeof(IConfigurationProviderService),
```
duncanp-sonar, line 51: This is no longer a local service. Instead, the host MEF-imports it now.

src/Integration.UnitTests/MefServices/VsSessionHostTests.cs
```diff
@@ -371,7 +383,7 @@ public void VsSessionHost_IServiceProvider_GetService()
 
         #region Helpers
 
-        private VsSessionHost CreateTestSubject(ConfigurableActiveSolutionTracker tracker)
+        private VsSessionHost CreateTestSubject(ConfigurableActiveSolutionTracker tracker = null)
```
duncanp-sonar, line 386: Miscellaneous: line 393 below handles the tracker being `null`, so we might as well make the parameter optional.

src/Integration.UnitTests/MefServices/VsSessionHostTests.cs
```diff
@@ -339,16 +353,14 @@ public void VsSessionHost_ResetBinding_ErrorInReadingBinding()
         public void VsSessionHost_IServiceProvider_GetService()
         {
             // Arrange
-            var testSubject = new VsSessionHost(this.serviceProvider, this.sonarQubeServiceMock.Object,
```
duncanp-sonar, line 342: Tidy up - all of the other tests in this class had previously been changed to use the `CreateTestSubject` method, but not this one; not sure why.

src/Integration.UnitTests/MefServices/VsSessionHostTests.cs
```diff
@@ -71,6 +71,20 @@ public void TestInitialize()
 
         #region Tests
 
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
```
duncanp-sonar, line 75: Adding missing test case. The `VsSessionHost` has always (?) been MEF-created, but we didn't have a test for it.

The `ActiveSolutionBoundTracker` is also MEF-creatable but doesn't have a `MefCtor_CheckIsExported` test. I looked at creating one, but it would have been more complicated since the constructor does other things e.g. fetches services, registers events etc. That test would have required a lot more setup, so I chose not to do it.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -96,7 +95,7 @@ public void ActiveSolutionBoundTracker_Initialisation_Unbound()
             host.VisualStateManager.ClearBoundProject();
 
             // Act
-            using (var testSubject = CreateTestSubject(this.host, this.activeSolutionTracker, loggerMock.Object))
+            using (var testSubject = CreateTestSubject(this.host, this.activeSolutionTracker, this.configProvider, loggerMock.Object))
```
duncanp-sonar, line 98: The setup for this test is ugly, but I took the easy way out and did the simplest fixup possible due to time constraints.

bigfluffycookie (approved):

PR #4190, opened 2023-05-05
Add triggers to update on SolutionBindingChanged/Update

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4189

### DIFF
--- a/src/ConnectedMode.UnitTests/Binding/ImportBeforeInstallTriggerTests.cs
+++ b/src/ConnectedMode.UnitTests/Binding/ImportBeforeInstallTriggerTests.cs
@@ -53,7 +53,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
         [DataRow(SonarLintMode.Connected, true)]
         [DataRow(SonarLintMode.LegacyConnected, true)]
         [DataRow(SonarLintMode.Standalone, false)]
-        public void Ctor__TriggersImportBeforeFileDependingOnMode(SonarLintMode mode, bool shouldTrigger)
+        public void Ctor_TriggersImportBeforeFileDependingOnMode(SonarLintMode mode, bool shouldTrigger)
         {
             var activeSolutionTracker = CreateActiveSolutionBoundTrackerWihtBindingConfig(mode);
             var importBeforeFileGenerator = new Mock<IImportBeforeFileGenerator>();
--- a/src/ConnectedMode.UnitTests/Binding/ImportBeforeInstallTriggerTests.cs
+++ b/src/ConnectedMode.UnitTests/Binding/ImportBeforeInstallTriggerTests.cs
@@ -129,7 +129,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
 
             var testSubject = CreateTestSubject(activeSolutionTracker.Object, Mock.Of<IImportBeforeFileGenerator>());
 
-            testSubject.Dispose();
+            ((IDisposable)testSubject).Dispose();
 
             activeSolutionTracker.VerifyRemove(x => x.PreSolutionBindingChanged -= It.IsAny<EventHandler<ActiveSolutionBindingEventArgs>>(), Times.Once);
             activeSolutionTracker.VerifyRemove(x => x.PreSolutionBindingUpdated -= It.IsAny<EventHandler>(), Times.Once);
--- a/src/ConnectedMode.UnitTests/Binding/ImportBeforeInstallTriggerTests.cs
+++ b/src/ConnectedMode.UnitTests/Binding/ImportBeforeInstallTriggerTests.cs
@@ -19,7 +19,6 @@
  */
 
 using System;
-using Microsoft.VisualStudio.Threading;
 using SonarLint.VisualStudio.ConnectedMode.Binding;
 using SonarLint.VisualStudio.Core;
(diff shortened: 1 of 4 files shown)


src/ConnectedMode/Binding/ImportBeforeInstallTrigger.cs
```diff
@@ -18,8 +18,10 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
```
bigfluffycookie, line 18: I have decided to go back to the original implementation and simply trigger it once from within the CTOR.

src/ConnectedMode/Binding/ImportBeforeInstallTrigger.cs
```diff
@@ -29,21 +31,38 @@ namespace SonarLint.VisualStudio.ConnectedMode.Binding
...
+
+        private void OnPreSolutionBindingChanged(object sender, ActiveSolutionBindingEventArgs e)
+        {
+            TriggerUpdate().Forget();
         }
 
-        public async Task TriggerUpdate()
+        private void OnPreSolutionBindingUpdated(object sender, EventArgs e) => TriggerUpdate().Forget();
+
+        private async Task TriggerUpdate()
```
bigfluffycookie, line 63: One question though, Do we want the file to only get written once or do we want it to get updated whenever the user does an update or binds? probably only once right? if so I'll add the dispose call back into this method.

duncanp-sonar: Yep, only once sounds right. We can always change it in a future release if necessary.

src/ConnectedMode.UnitTests/Binding/ImportBeforeInstallTriggerTests.cs
```diff
@@ -34,56 +33,130 @@ public class ImportBeforeInstallTriggerTests
...
+        [TestMethod]
+        public void Dispose_UnsubscribesToEvent()
+        {
+            var activeSolutionTracker = new Mock<IActiveSolutionBoundTracker>();
 
-            configProvider.Setup(x => x.GetConfiguration()).Returns(bindingConfig);
+            var testSubject = CreateTestSubject(activeSolutionTracker.Object, Mock.Of<IImportBeforeFileGenerator>());
 
-            return configProvider.Object;
+            testSubject.Dispose();
```
duncanp-sonar, line 132: The product class isn't actually marked as disposable, but this test passes because `Dispose` is being called directly.

You could tweak the test to check that the class is disposable by casting explicitly to `IDisposable`:
```
  ((IDisposable)testSubject).Dispose();
```

src/ConnectedMode/Binding/ImportBeforeInstallTrigger.cs
```diff
@@ -29,21 +31,38 @@ namespace SonarLint.VisualStudio.ConnectedMode.Binding
     [PartCreationPolicy(CreationPolicy.Shared)]
     internal class ImportBeforeInstallTrigger
     {
-        private readonly IConfigurationProvider configurationProvider;
+        private readonly IActiveSolutionBoundTracker activeSolutionBoundTracker;
```
duncanp-sonar, line 34: The class probable needs to be `sealed` to prevent a different code smell about the implementation of `IDisposable`.

duncanp-sonar (commented):
See inline - a few minor changes required.

rita-gorokhod (approved):

src/ConnectedMode.UnitTests/Binding/ImportBeforeInstallTriggerTests.cs
```diff
@@ -34,56 +33,130 @@ public class ImportBeforeInstallTriggerTests
...
 
-        private IConfigurationProvider SetUpConfigProvider(SonarLintMode mode)
+        [TestMethod]
+        public void InvokeBindingUpdated_ImportBeforeFileGeneratorIsCalled()
         {
-            var configProvider = new Mock<IConfigurationProvider>();
-            var bindingConfig = new BindingConfiguration(
-                                    new BoundSonarQubeProject(new Uri("http://localhost"), "test", ""),
-                                    mode, "");
+            var activeSolutionTracker = CreateActiveSolutionBoundTrackerWihtBindingConfig(SonarLintMode.Standalone);
```
rita-gorokhod, line 111: I think you can extract this and the Changed test into one method that receives RaiseBindingEvent method. But it's minor, your call.

src/ConnectedMode/Binding/ImportBeforeInstallTrigger.cs
```diff
@@ -18,38 +18,69 @@
...
+        private async Task TriggerUpdateAsync()
         {
-            var config = configurationProvider.GetConfiguration();
+            var config = activeSolutionBoundTracker.CurrentConfiguration;
 
             if (config.Mode != SonarLintMode.Standalone)
             {
                 await threadHandling.SwitchToBackgroundThread();
 
                 importBeforeFileGenerator.WriteTargetsFileToDiskIfNotExists();
```
rita-gorokhod, line 71: @bigfluffycookie @duncanp-sonar I took a look at importBeforeFileGenerator -- It's not thread-safe, and this event can call it multiple times. You need to add a lock there.

PR #4199, opened 2023-05-06
(2) Implement IConfigurationProvider

### LINKED ISSUE #4198: Implement UnintrusiveConfigurationProvider
### Description

Add the implementation for `UnintrusiveConfigurationProvider`.

This will be similar to the implementation of `ObsoleteConfigurationProvider`, except it will only import one path provider, `IUnintrusiveBindingPathProvider`.

duncanp-sonar:
Fixes #4198

### DIFF
--- a/src/Integration.UnitTests/UnintrusiveBinding/UnintrusiveConfigurationProviderTests.cs
+++ b/src/Integration.UnitTests/UnintrusiveBinding/UnintrusiveConfigurationProviderTests.cs
@@ -1,0 +1,128 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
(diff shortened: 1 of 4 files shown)


src/Integration/UnintrusiveBinding/UnintrusiveConfigurationProvider.cs
```diff
@@ -19,15 +19,54 @@
...
     [PartCreationPolicy(CreationPolicy.Shared)]
     internal class UnintrusiveConfigurationProvider : IConfigurationProvider
     {
-        // TODO - fetch settings from new location #4170
-        public BindingConfiguration GetConfiguration() => BindingConfiguration.Standalone;
+        private readonly IUnintrusiveBindingPathProvider pathProvider;
+        private readonly ISolutionBindingDataReader solutionBindingDataReader;
+
+        [ImportingConstructor]
+        public UnintrusiveConfigurationProvider(
```
duncanp-sonar, line 37: This is very similar to the old [ObsoleteConfigurationProvider](https://github.com/SonarSource/sonarlint-visualstudio/blob/173efdcd16e3af13cbf6394bc129bc81ce634af7/src/Integration/NewConnectedMode/ObsoleteConfigurationProvider.cs#L52), but simpler.
The old class handled two different locations for Connected Mode settings. This version only needs to handle one.

src/Integration.UnitTests/UnintrusiveBinding/UnintrusiveConfigurationProviderTests.cs
```diff
@@ -0,0 +1,128 @@
...
+using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Integration.NewConnectedMode;
+using SonarLint.VisualStudio.Integration.Persistence;
+using SonarLint.VisualStudio.Integration.UnintrusiveBinding;
+using SonarLint.VisualStudio.TestInfrastructure;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.UnintrusiveBinding
+{
+    [TestClass]
+    public class UnintrusiveConfigurationProviderTests
```
duncanp-sonar, line 39: The tests for the old class are [here](https://github.com/SonarSource/sonarlint-visualstudio/blob/173efdcd16e3af13cbf6394bc129bc81ce634af7/src/Integration.UnitTests/NewConnectedMode/ObsoleteConfigurationProviderTests.cs#L54) for reference.
However, the new class is simpler, and the old tests used `TestInitialize` so I didn't really copy anything from the old tests.

rita-gorokhod (approved):

PR #4208, opened 2023-05-10
Add support for span to RuleHelpXamlTranslator 

### LINKED ISSUE #4202: Add support for span to RuleHelpXamlTranslator

ugras-ergun-sonarsource:
fixes #4202

### DIFF
--- a/src/Education/XamlGenerator/RuleHelpXamlTranslator.cs
+++ b/src/Education/XamlGenerator/RuleHelpXamlTranslator.cs
@@ -232,12 +232,14 @@ namespace SonarLint.VisualStudio.Education.XamlGenerator
 
                         PushOutputElementInfo("h4", true);
                         break;
+
                     case "h5":
                         WriteBlockElementStart("Paragraph");
                         writer.ApplyStyleToElement(StyleResourceNames.Heading5_Paragraph);
 
                         PushOutputElementInfo("h5", true);
                         break;
+
                     case "h6":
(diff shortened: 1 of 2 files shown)


src/Education/XamlGenerator/RuleHelpXamlTranslator.cs
```diff
@@ -307,6 +309,11 @@ private void ProcessElement()
 
                         break;
 
+                    case "span":
+                        WriteInlineElementStart("Span");
```
georgii-borovinskikh-sonarsource, line 313: Does it display correctly? Also add a simple test for this, maybe even parse to make sure

georgii-borovinskikh-sonarsource (changes requested):
I think a test case is needed here

src/Education/XamlGenerator/RuleHelpXamlTranslator.cs
ugras-ergun-sonarsource: Tests are in place already, that's how we have found the issue. And I have manually checked for failing rules and they were displayed correctly.

georgii-borovinskikh-sonarsource: Those are smoke tests. We also have `RuleHelpXamlTranslatorTests`. It's not a complete set of test cases, but IMO it's good to add tests for new stuff we add, especially if it's small things

georgii-borovinskikh-sonarsource (approved):

Issue #4203, opened 2023-05-09
Fix Core.Language filesuffix property


bigfluffycookie:
Change fileSuffix in Core.Language for Csharp and VB to match the msbuild $(Language) property.

$(Language) for csharp is "C#" whereas in the Core.Language the filesuffix is "csharp".

Issue #4188, opened 2023-05-05
Change locations and names of settings files 


duncanp-sonar:
### Description

The old structure assumed that the settings would be in a `.sonarlint` folder under the folder containing the `.sln` file.
Since there could be multiple solutions in the same folder, the individual file names started with the Sonar project key to disambiguate between different Sonar projects.

The new structure assumes there is as a top-level folder named after the solution file, with all of the settings files being written under that i.e. there is no need to include the Sonar project key in the file name

#### Old structure
```
-- [.sonarlint]
---- SLVS_Bound_VS2019.slconfig
---- slvs_samples_bound_vs2019_cpp_settings.json
---- slvs_samples_bound_vs2019cSharp.ruleset
---- slvs_samples_bound_vs2019vb.ruleset
...

---- [slvs_samples_bound_vs2019]
------ [CSharp]
-------- SonarLint.xml
------ [VB]
-------- SonarLint.xml

```

#### Proposed new structure

```
-- [Bindings]

---- [slvs_samples_bound_vs2019_HASHCODE]
---- binding.config     <-- formerly xxx.slconfig
---- cpp_settings.json
---- cSharp.ruleset
---- vb.ruleset
...

------ [CSharp]
-------- csharp.globalconfig
-------- SonarLint.xml

------ [VB]
-------- SonarLint.xml
-------- vb.globalconfig
```

Issue #1068, opened 2019-11-06
Inconsistent casing of ruleset filenames causes build warnings on non-Windows machines


duncanp-sonar:
### Description

SLVS generates a solution-level ruleset from the Quality Profile and then references this ruleset as in `<Include>` in project-level rulesets.

The casing of the ruleset name in the `<Include>` is not preserved which can lead to build warnings when building .NET Core projects on Linux machines.

e.g.
Generated ruleset name: `sonarscanner-msbuildCSharp.ruleset`
Include element: `<Include Path="..\.sonarlint\sonarscanner-msbuildcsharp.ruleset" Action="Default" />`

See [this community thread](https://community.sonarsource.com/t/sonarlint-visual-studio-sync/11723).

### Repro steps

1. Create a SonarQube project
2. Bind a VS solution containing .NET Core projects to the SonarQube project
3. Build the solution on a non-Windows machine

### Expected behavior

The generated rulesets are resolved correctly during the build.

### Actual behavior

Build warnings that the ruleset cannot be resolved.

### Known workarounds

Manually rename the ruleset files so the cases match.

### Related information

* SonarLint for VS Version: all
* Visual Studio Version: n/a
* SonarQube server version: n/a
* Languages in your solution: C#, VB.NET

duncanp-sonar:
The constructor for `Microsoft.VisualStudio.CodeAnalysis.RuleSets.RuleSetInclude` (Microsoft.VisualStudio.CodeAnalysis.dll) calls `ToLowerInvariant` on the supplied path.

PR #4214, opened 2023-05-15
[2] Store the globalconfig and SonarLint.xml without projectKey

rita-gorokhod:
Fixes [#4203](https://github.com/SonarSource/sonarlint-visualstudio/issues/4203) and [#4188](https://github.com/SonarSource/sonarlint-visualstudio/issues/4188)

As we discussed, I've removed the projectKey distinction. It simplifies things, and we already broke the support of it when we added `sonar.settings.json`.

New folder structure on disk:

image
image
image

### DIFF
--- a/src/ConnectedMode/Embedded/SonarLintTargets.xml
+++ b/src/ConnectedMode/Embedded/SonarLintTargets.xml
@@ -20,11 +20,14 @@
 
     <PropertyGroup Condition="$(BuildingInsideVisualStudio) == 'true'">
         <SolutionDirAsInVs>$([System.IO.Path]::GetDirectoryName($(SolutionDir)))</SolutionDirAsInVs>
+        <SonarLanguage>$(Language)</SonarLanguage>
+        <SonarLanguage Condition=" $(Language) == 'C#' OR $(Language) == 'c#'">csharp</SonarLanguage>
+        <SonarLanguage Condition=" $(Language) == 'VISUALBASIC'">vb</SonarLanguage>
         <!-- C#/VB.NET settings need to be in different folders -->
-        <_SLVSRootFolder>$(APPDATA)\SonarLint for Visual Studio\Bindings\$(SolutionName)_$(SolutionDirAsInVs.GetHashCode())\$(Language)</_SLVSRootFolder>
+        <_SLVSRootFolder>$(APPDATA)\SonarLint for Visual Studio\Bindings\$(SolutionName)_$(SolutionDirAsInVs.GetHashCode())</_SLVSRootFolder>
 
-        <_SLVSGeneratedRoslynConfigFile>$(_SLVSRootFolder)\slvs_generated.globalconfig</_SLVSGeneratedRoslynConfigFile>
-        <_SLVSGeneratedAdditionalFile>$(_SLVSRootFolder)\SonarLint.xml</_SLVSGeneratedAdditionalFile>
+        <_SLVSGeneratedRoslynConfigFile>$(_SLVSRootFolder)\$(SonarLanguage).globalconfig</_SLVSGeneratedRoslynConfigFile>
+        <_SLVSGeneratedAdditionalFile>$(_SLVSRootFolder)\$(SonarLanguage)\SonarLint.xml</_SLVSGeneratedAdditionalFile>
     </PropertyGroup>
 
     <ItemGroup Condition="$(BuildingInsideVisualStudio) == 'true' AND Exists($(_SLVSGeneratedRoslynConfigFile))">
@@ -44,7 +47,7 @@
         </PropertyGroup>
 
         <Message Importance="high" Text="XXX: Building inside VS: $(BuildingInsideVisualStudio)" />
-        <Message Importance="high" Text="XXX: Language: $(Language)" />
+        <Message Importance="high" Text="XXX: SonarLanguage: $(SonarLanguage)" />
         <Message Importance="high" Text="XXX: SolutionDirAsInVs: $(SolutionDirAsInVs)" />
         <Message Importance="high" Text="XXX: SolutionDirAsInVs.GetHashCode: $(SolutionDirAsInVs.GetHashCode())" />
         <Message Importance="high" Text="XXX: SLVSRootFolder: $(_SLVSRootFolder)" />
--- a/src/Core/Binding/BindingConfiguration.cs
+++ b/src/Core/Binding/BindingConfiguration.cs
@@ -92,7 +92,7 @@ namespace SonarLint.VisualStudio.Core.Binding
 
         public string BuildPathUnderConfigDirectory(string fileSuffix = "")
         {
-            var escapedFileName = Helpers.PathHelper.EscapeFileName(Project.ProjectKey + fileSuffix).ToLowerInvariant(); // Must be lower case - see https://github.com/SonarSource/sonarlint-visualstudio/issues/1068;
+            var escapedFileName = Helpers.PathHelper.EscapeFileName(fileSuffix).ToLowerInvariant(); // Must be lower case - see https://github.com/SonarSource/sonarlint-visualstudio/issues/1068;
             
             return Path.Combine(BindingConfigDirectory, escapedFileName);
         }
--- a/src/Core/Language.cs
+++ b/src/Core/Language.cs
@@ -45,11 +45,11 @@ namespace SonarLint.VisualStudio.Core
         public readonly static Language Unknown = new Language();
         public readonly static Language CSharp = new Language("CSharp", CoreStrings.CSharpLanguageName, "csharp.globalconfig", SonarQubeLanguage.CSharp);
         public readonly static Language VBNET = new Language("VB", CoreStrings.VBNetLanguageName, "vb.globalconfig", SonarQubeLanguage.VbNet);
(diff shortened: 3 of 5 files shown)


src/ConnectedMode/Embedded/SonarLintTargets.xml
```diff
@@ -20,11 +20,14 @@
 
     <PropertyGroup Condition="$(BuildingInsideVisualStudio) == 'true'">
         <SolutionDirAsInVs>$([System.IO.Path]::GetDirectoryName($(SolutionDir)))</SolutionDirAsInVs>
+        <SonarLanguage>$(Language)</SonarLanguage>
+        <SonarLanguage Condition=" $(Language) == 'C#' OR $(Language) == 'c#'">csharp</SonarLanguage>
+        <SonarLanguage Condition=" $(Language) == 'VISUALBASIC'">vb</SonarLanguage>
```
rita-gorokhod, line 25: According to this website, apparently those are the possible values, I don't know if there is a better/official documentation: https://documentation.help/MSBuild/P_MSBuild_Community_Tasks_Script_Language.html

src/ConnectedMode/Embedded/SonarLintTargets.xml
```diff
@@ -20,11 +20,14 @@
...
         <!-- C#/VB.NET settings need to be in different folders -->
-        <_SLVSRootFolder>$(APPDATA)\SonarLint for Visual Studio\Bindings\$(SolutionName)_$(SolutionDirAsInVs.GetHashCode())\$(Language)</_SLVSRootFolder>
+        <_SLVSRootFolder>$(APPDATA)\SonarLint for Visual Studio\Bindings\$(SolutionName)_$(SolutionDirAsInVs.GetHashCode())</_SLVSRootFolder>
 
-        <_SLVSGeneratedRoslynConfigFile>$(_SLVSRootFolder)\slvs_generated.globalconfig</_SLVSGeneratedRoslynConfigFile>
-        <_SLVSGeneratedAdditionalFile>$(_SLVSRootFolder)\SonarLint.xml</_SLVSGeneratedAdditionalFile>
+        <_SLVSGeneratedRoslynConfigFile>$(_SLVSRootFolder)\$(SonarLanguage).globalconfig</_SLVSGeneratedRoslynConfigFile>
+        <_SLVSGeneratedAdditionalFile>$(_SLVSRootFolder)\$(SonarLanguage)\SonarLint.xml</_SLVSGeneratedAdditionalFile>
```
rita-gorokhod, line 30: Technically, I think that SonarLanguage can be CSHARP whereas the folder on disk will be csharp. We had a comment in the code about this issue https://github.com/SonarSource/sonarlint-visualstudio/issues/1068
But I don't think it's relevant here, since this code is not supposed to run in MSBuild builds.

src/Core/Binding/BindingConfiguration.cs
```diff
@@ -92,7 +92,7 @@ public override int GetHashCode()
 
         public string BuildPathUnderConfigDirectory(string fileSuffix = "")
         {
-            var escapedFileName = Helpers.PathHelper.EscapeFileName(Project.ProjectKey + fileSuffix).ToLowerInvariant(); // Must be lower case - see https://github.com/SonarSource/sonarlint-visualstudio/issues/1068;
+            var escapedFileName = Helpers.PathHelper.EscapeFileName(fileSuffix).ToLowerInvariant(); // Must be lower case - see https://github.com/SonarSource/sonarlint-visualstudio/issues/1068;
```
rita-gorokhod, line 95: Removal of project key.

src/Core/Language.cs
```diff
@@ -45,11 +45,11 @@ public sealed class Language : IEquatable<Language>
         public readonly static Language Unknown = new Language();
         public readonly static Language CSharp = new Language("CSharp", CoreStrings.CSharpLanguageName, "csharp.globalconfig", SonarQubeLanguage.CSharp);
         public readonly static Language VBNET = new Language("VB", CoreStrings.VBNetLanguageName, "vb.globalconfig", SonarQubeLanguage.VbNet);
-        public readonly static Language Cpp = new Language("C++", CoreStrings.CppLanguageName, "_cpp_settings.json", SonarQubeLanguage.Cpp);
-        public readonly static Language C = new Language("C", "C", "_c_settings.json", SonarQubeLanguage.C);
-        public readonly static Language Js = new Language("Js", "JavaScript", "_js_settings.json", SonarQubeLanguage.Js);
-        public readonly static Language Ts = new Language("Ts", "TypeScript", "_ts_settings.json", SonarQubeLanguage.Ts);
-        public readonly static Language Secrets = new Language("Secrets", "Secrets", "_secrets_settings.json", SonarQubeLanguage.Secrets);
+        public readonly static Language Cpp = new Language("C++", CoreStrings.CppLanguageName, "cpp_settings.json", SonarQubeLanguage.Cpp);
```
rita-gorokhod, line 48: Since we don't have the project key anymore, I removed the redundant `_`.

src/Integration.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
```diff
@@ -180,7 +180,7 @@ public async Task GetConfig_ReturnsCorrectAdditionalFile()
             };
             var testSubject = builder.CreateTestSubject();
 
-            var expectedAdditionalFilePath = builder.BindingConfiguration.BuildPathUnderConfigDirectory() + "\\VB\\SonarLint.xml";
+            var expectedAdditionalFilePath = builder.BindingConfiguration.BuildPathUnderConfigDirectory() + "VB\\SonarLint.xml";
```
rita-gorokhod, line 183: Minor: no project name anymore so BuildPathUnderConfigDirectory already returns something that ends with `\\`

src/ConnectedMode/Embedded/SonarLintTargets.xml
```diff
@@ -20,11 +20,14 @@
...
+        <SonarLanguage>$(Language)</SonarLanguage>
+        <SonarLanguage Condition=" $(Language) == 'C#' OR $(Language) == 'c#'">csharp</SonarLanguage>
+        <SonarLanguage Condition=" $(Language) == 'VISUALBASIC'">vb</SonarLanguage>
         <!-- C#/VB.NET settings need to be in different folders -->
-        <_SLVSRootFolder>$(APPDATA)\SonarLint for Visual Studio\Bindings\$(SolutionName)_$(SolutionDirAsInVs.GetHashCode())\$(Language)</_SLVSRootFolder>
+        <_SLVSRootFolder>$(APPDATA)\SonarLint for Visual Studio\Bindings\$(SolutionName)_$(SolutionDirAsInVs.GetHashCode())</_SLVSRootFolder>
 
-        <_SLVSGeneratedRoslynConfigFile>$(_SLVSRootFolder)\slvs_generated.globalconfig</_SLVSGeneratedRoslynConfigFile>
-        <_SLVSGeneratedAdditionalFile>$(_SLVSRootFolder)\SonarLint.xml</_SLVSGeneratedAdditionalFile>
+        <_SLVSGeneratedRoslynConfigFile>$(_SLVSRootFolder)\$(SonarLanguage).globalconfig</_SLVSGeneratedRoslynConfigFile>
```
rita-gorokhod, line 29: I thought about placing the file under the csharp/vb folders, next to SonarLint.xml, but it wouldn't be consistent with the other setting files. I don't have a preference either way, LMK what you think would be better/simpler.

bigfluffycookie: Hm I can see where you are coming from, but to me its more confusing that we have a folder called csharp and then a file for csharp outside of said folder but that could just be a personal preference

bigfluffycookie (approved):
Looks good to me, one comment about the globalconfig path but I'll leave the choice to you

src/ConnectedMode/Embedded/SonarLintTargets.xml
duncanp-sonar: `#` is a valid character in folder and file names on Windows so in theory we could just use `$(Language)` directly, although it does feel a bit weird. What do you think?

src/ConnectedMode/Embedded/SonarLintTargets.xml
duncanp-sonar: I'm in two minds about which I prefer. Having both settings files in the same folder feels simpler, rather than having to look in two places, and the Roslyn configuration is already inconsistent with the other languages since it contains the extra `SonarLint.xml` file.

However, the DotNet language team have looked at whether it would be possible to provide the rule configuration parameters in an `.editorconfig`/`.globalconfig` file in the future i.e. get rid of the `SonarLint.xml` file completely.  If they did that, we could get rid of the `csharp` and `vb` folders since we only need them for the `SonarLint.xml` file (this isn't something they are going to do immediately, but might in the future).

For me, there isn't much in it. Since we're not expecting users to have to look at these files directly, I would probably leave it as it is (i.e. only the `SonarLint.xml` file in a separate folder), but I don't have a strong preference for it.

duncanp-sonar (approved):
Looks good. A couple of comments/suggestions inline.

duncanp-sonar:
Superceded by https://github.com/SonarSource/sonarlint-visualstudio/pull/4239 (rebased on `branch-unintrusive2` and merge conflicts fixed)

Issue #4168, opened 2023-05-03
[Auto-migrate] Automatically convert old settings for new format when necessary


duncanp-sonar:
### Description

If a bound solution using one of the old Connected Modes is opened, the settings should be migrated to the new location/format if they do not already exist.

### Notes
This implies that we need to be able to load both the old and new settings, and to be able to distinguish between them.

More generally, the rest of the existing product code should only use the new Connected Mode settings location to check for the binding configuration i.e. in normal operation we ignore the existence of the old settings.

#### Proposed changes
* write a new `IConfigurationProvider` interface/implementation that only returns the unintrusive Connected Mode settings format/location

* rename the existing `IConfigurationProvider` interfaces/implementations by prefxing them with `Obsolete`.

i.e. we will provide two services:
* `IConfigurationProvider` - used by all of the existing code. Only supports the new settings format/location
* `IObsoleteConfigurationProvider` - only used the auto-migration and clean-up code

At some point in the future we will be able to drop the `Obsolete` types.

PR #4227, opened 2023-05-24
Add class that handles checking if auto migration is necessary

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4168


src/ConnectedMode/Migration/IUnintrusiveBindingController.cs
```diff
@@ -0,0 +1,36 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Migration
+{
+    internal interface IUnintrusiveBindingController
```
bigfluffycookie, line 25: I spent a good 15 minutes trying to think of a name for this class and eventually gave up, this is a bad name and does not represent what this class does at all however, I am not 100% sure about  1. all the responsibilities & 2. what a sensible name would be for the responsibilities I do know about

I need help thanks

src/TestInfrastructure/InternalsVisibleTo.cs
```diff
@@ -25,7 +25,8 @@
 [assembly: InternalsVisibleTo("SonarLint.VisualStudio.Integration.TeamExplorer.UnitTests,PublicKey=002400000480000094000000060200000024000052534131000400000100010081b4345a022cc0f4b42bdc795a5a7a1623c1e58dc2246645d751ad41ba98f2749dc5c4e0da3a9e09febcb2cd5b088a0f041f8ac24b20e736d8ae523061733782f9c4cd75b44f17a63714aced0b29a59cd1ce58d8e10ccdb6012c7098c39871043b7241ac4ab9f6b34f183db716082cd57c1ff648135bece256357ba735e67dc6")]
 [assembly: InternalsVisibleTo("SonarLint.VisualStudio.Integration.Vsix.UnitTests,PublicKey=002400000480000094000000060200000024000052534131000400000100010081b4345a022cc0f4b42bdc795a5a7a1623c1e58dc2246645d751ad41ba98f2749dc5c4e0da3a9e09febcb2cd5b088a0f041f8ac24b20e736d8ae523061733782f9c4cd75b44f17a63714aced0b29a59cd1ce58d8e10ccdb6012c7098c39871043b7241ac4ab9f6b34f183db716082cd57c1ff648135bece256357ba735e67dc6")]
 #else
+[assembly: InternalsVisibleTo("SonarLint.VisualStudio.ConnectedMode.UnitTests")]
```
bigfluffycookie, line 28: I just realized I need to remove this again, I had it in to reference the ConfigurableConfigurationProvider class but I could not make it work.

src/ConnectedMode/Migration/AutoMigrationHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        public AutoMigrationHandler(IUnintrusiveBindingController bindingController, IConfigurationProvider configurationProvider, IObsoleteConfigurationProvider obsoleteConfigurationProvider)
+        {
+            this.bindingController = bindingController;
+            this.configurationProvider = configurationProvider;
+            this.obsoleteConfigurationProvider = obsoleteConfigurationProvider;
+        }
+
+        internal /* for testing */ void Migrate()
+        {
+            var obsoleteConfiguration = obsoleteConfigurationProvider.GetConfiguration();
```
bigfluffycookie, line 44: I could have put this in the if() the same way I had the second half however, we will need to reuse the configuration so to me it makes sense to save it here.

src/ConnectedMode.UnitTests/Migration/AutoMigrationHandlerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+        [TestMethod]
+        [DataRow(SonarLintMode.Standalone, SonarLintMode.Standalone, false)]
+        [DataRow(SonarLintMode.Connected, SonarLintMode.Standalone, true)]
+        [DataRow(SonarLintMode.LegacyConnected, SonarLintMode.Standalone, true)]
+        [DataRow(SonarLintMode.Connected, SonarLintMode.Connected, false)]
+        [DataRow(SonarLintMode.LegacyConnected, SonarLintMode.Connected, false)]
+        public void Migrate_BindingGetsCalledWithCorrectCondition(SonarLintMode obsoleteMode, SonarLintMode mode, bool expectBindingToBeCalled)
+        {
+            var bindingController = new Mock<IUnintrusiveBindingController>();
+
```
bigfluffycookie, line 49: Here I wanted to use ConfigurableConfigurationProvider but I could not make it work as I need the obsoleteConfigration Provider to be of the type IObsoleteConfigurationProvider. I still think there might be a prettier way to do this, I just do not know how.

src/ConnectedMode.UnitTests/Migration/AutoMigrationHandlerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+                MefTestHelpers.CreateExport<IConfigurationProvider>(),
+                MefTestHelpers.CreateExport<IObsoleteConfigurationProvider>());
+        }
+
+        [TestMethod]
+        [DataRow(SonarLintMode.Standalone, SonarLintMode.Standalone, false)]
+        [DataRow(SonarLintMode.Connected, SonarLintMode.Standalone, true)]
+        [DataRow(SonarLintMode.LegacyConnected, SonarLintMode.Standalone, true)]
+        [DataRow(SonarLintMode.Connected, SonarLintMode.Connected, false)]
+        [DataRow(SonarLintMode.LegacyConnected, SonarLintMode.Connected, false)]
```
bigfluffycookie, line 45: I didn't bother checking the premutation of using SonarLintMode.LegacyConnected for the new configProvider as it will never be of that type.

src/ConnectedMode/Migration/IUnintrusiveBindingController.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        void Bind();
+    }
+
+    [Export(typeof(IUnintrusiveBindingController))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class UnintrusiveBindingController : IUnintrusiveBindingController
+    {
+        public void Bind() { }
```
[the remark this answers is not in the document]
duncanp-sonar: You could have, but it doesn't matter too much since we know we're going add the implementation shortly.

Normally, I'd add a TODO comment referencing the ticket that relates to creating the implementation (which in turn creates another code smell since it's a TODO. However, from a reader point of view it's clear that there is work planned to remove the TODO).

src/ConnectedMode/Migration/AutoMigrationHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            // If the user has the old files but not the new files it means they are bound and we can migrate over to the new binding.
+            if (obsoleteConfiguration.Mode != SonarLintMode.Standalone
+                && configurationProvider.GetConfiguration().Mode == SonarLintMode.Standalone)
+            {
+                bindingController.Bind();
```
duncanp-sonar, line 50: I think at this point we will need to pass some information from the obsolete config provider so the binding controller knows what binding to create.

I don't think we want to pass the whole `BindingConfiguration`, since that contains info that is not correct i.e. the `bindingConfigurationDirectory`. However, it looks like the `BindingConfiguration.Project` (i.e. the `BoundSonarQubeProject` object) has just the information about the server that we need.

src/ConnectedMode.UnitTests/Migration/AutoMigrationHandlerTests.cs
duncanp-sonar: To be honest, just using mocks here is probably simpler and better as we only need to mock one property.

The `ConfigurationConfigurationProvider` dates from a time before we started using `Moq` (i.e. more than five years ago), so all of the `ConfigurableXXX` helpers classes are legacy code. Some of them are unnecessarily complicated e.g. the `ConfigurableConfigurationProvider` also implements `IConfigurationPersister`, which we don't need here.

My first choice when writing new tests is to use `Moq`. I only fall back on the `ConfigurableXXX` helper classes if the behaviour is hard to mock from scratch (which in turn is probably an indication that there's a design/complexity problem in the product code itself).

src/ConnectedMode/Migration/AutoMigrationHandler.cs
duncanp-sonar: I was wondering whether we should add verbose logging here to help with debugging and user support.
However, it's probably a bit too early at the moment since we don't know what the overall user experience will be.

I'll add a note to the board so we don't forget to consider it later.

src/ConnectedMode/Migration/IUnintrusiveBindingController.cs
duncanp-sonar: I think it's ok to leave it as it is for now, effectively with a placeholder name. When we come to implement it, we'll have a better idea of what its responsibilities are so we can rename it then.

Later on when we come to implement the auto-update of QPs, its responsibilities might change or expand, at which point we might decide to rename it again.

duncanp-sonar (commented):
Looks good. Main observation is that the new `Bind` method probably needs a parameter containing the info about the server and project to bind to.

src/ConnectedMode.UnitTests/Migration/AutoMigrationHandlerTests.cs
duncanp-sonar: > we only care if its standalone or not

True, and the unit tests for `UnintrusiveConfigurationProvider` should check it returns the correct values, so there's no need for an assertion.

duncanp-sonar (approved):
Looks good.

Apologies, I forgot to review this earlier.

PR #4252, opened 2023-05-30
Add IHotspotAnalysisConfiguration integration to cfamily RulesConfigFixup.cs

georgii-borovinskikh-sonarsource:


src/CFamily.UnitTests/Rules/DynamicCFamilyRulesConfigTests.cs
```diff
@@ -318,15 +341,20 @@ public void EffectiveParameters_CustomSettingsOverrideDefaults()
 
         private static DynamicCFamilyRulesConfig CreateTestSubject(ICFamilyRulesConfig defaultConfig,
             RulesSettings customSettings,
-            IRulesConfigFixup fixup = null)
+            IRulesConfigFixup fixup = null,
+            IHotspotAnalysisConfiguration hotspotAnalysisConfiguration = null)
         {
             fixup ??= new NoOpRulesConfigFixup();
-            return new DynamicCFamilyRulesConfig(defaultConfig, customSettings, new TestLogger(), fixup);
+            var analysisConfiguration = new Mock<IHotspotAnalysisConfiguration>();
```
duncanp-sonar, line 348: Always creating an instance of `analysisConfiguration` and then not using it in some cases looks a bit messy and misleading. The normal approach in other tests is to only create an object with default behaviour if one is not supplied.

src/CFamily.UnitTests/Rules/RulesConfigFixupTests.cs
```diff
@@ -30,6 +33,19 @@ namespace SonarLint.VisualStudio.CFamily.Rules.UnitTests
...
+        private static readonly IHotspotAnalysisConfiguration HotspotsDisabled;
+        
+        static RulesConfigFixupTests()
+        {
+            var hotspotsEnabledMock = new Mock<IHotspotAnalysisConfiguration>();
+            hotspotsEnabledMock.Setup(x => x.IsEnabled()).Returns(true);
+            HotspotsEnabled = hotspotsEnabledMock.Object;
+            var hotspotsDisabledMock = new Mock<IHotspotAnalysisConfiguration>();
+            hotspotsDisabledMock.Setup(x => x.IsEnabled()).Returns(false);
+            HotspotsDisabled = hotspotsDisabledMock.Object;
```
duncanp-sonar, line 46: This would be easier to read and more consistent with other tests if it had a simple helper method rather than a `static` initializer i.e.

```
private static IHotspotAnalysisConfiguration CreateHotspotAnalysisConfig(bool isEnabled) 
{
  ...
}

```

duncanp-sonar (commented):
Just a couple of minor comments about the tests.

src/CFamily.UnitTests/Rules/DynamicCFamilyRulesConfigTests.cs
georgii-borovinskikh-sonarsource: My bad it was meant to be removed before commiting

PR #4255, opened 2023-05-30
Simplify IBindingProcess - drop solution-save code

duncanp-sonar:
- new binding method does not change any files in the solution

Part of #4120

### DIFF
--- a/src/Integration/Binding/IBindingProcess.cs
+++ b/src/Integration/Binding/IBindingProcess.cs
@@ -34,8 +34,6 @@ namespace SonarLint.VisualStudio.Integration.Binding
     /// making the changes to projects and files.</remarks>
     internal interface IBindingProcess
     {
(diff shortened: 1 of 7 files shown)


src/Integration/Binding/IBindingProcess.cs
```diff
@@ -34,8 +34,6 @@ namespace SonarLint.VisualStudio.Integration.Binding
     /// making the changes to projects and files.</remarks>
     internal interface IBindingProcess
     {
-        bool PromptSaveSolutionIfDirty();
```
duncanp-sonar, line 37: I delete the two solution-save related steps from the interface, and the associated implementation and tests.

duncanp-sonar:
I finally got my VS Exp instance working again, so I was able to manually testing that binding still works.

bigfluffycookie (approved):
Looks good!

Issue #4254, opened 2023-05-30
Implement Migration Checker


bigfluffycookie:
### Description

Check if migration is required. If so, display a top-level gold bar prompting the user to migrate.

Links on the gold bar:
* _Start migration_
* _What does migration do?_ / _Get help_
* ?_Don't show this again_  - no? We want to encourage users to upgrade -> nag

TODO 

- [x] Change AutoMigrationHandler to handle the above
- [x] Add a new IConnectedModeMigration interface

PR #4257, opened 2023-05-31
Amend InfoBarManager to allow for a gold bar to be attached to the main window

bigfluffycookie:
part of https://github.com/SonarSource/sonarlint-visualstudio/issues/4254


src/Core/InfoBar/IInfoBarManager.cs
```diff
@@ -54,6 +54,35 @@ public interface IInfoBarManager
         /// </remarks>
```
bigfluffycookie, line 54: Technically we could merge all three methods into one and replace the buttonTexts with params, the only reason I thought about that is because now this class has 6 methods which are almost identical. Changing it would require touching quite a few classes though so I didn't bother.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -303,6 +303,29 @@ public void AttachInfoBarWithButtons_ButtonClickedEvent_RaisedWithTheRightButton
             eventHandler.VerifyNoOtherCalls();
         }
 
+        /// <summary>
+        /// The other tests cover the implementation of gold bar functionality.
+        /// The methods AttachInfoBarWithMainWindow & AttachInfoBarWithMainWindowButtons are almost identical to their ToolWindow method counterparts
+        /// which is why a simple test to check if it gets the main window instead of a toolwindow and if it can detach correctly is enough.
+        /// </summary>
+        [TestMethod]
+        public void InfoBarManager_AttachInfoBarWithMainWindowAndDetach()
```
bigfluffycookie, line 312: I wasn't sure if it is appropriate to put a comment like this in a test class so lmk if I should remove it.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -303,6 +303,29 @@ public void AttachInfoBarWithButtons_ButtonClickedEvent_RaisedWithTheRightButton
             eventHandler.VerifyNoOtherCalls();
```
bigfluffycookie, line 303: Also while looking at this class, I found there are quite a few Configurable... test helper classes that we do not use. Maybe it would make sense to delete them simply to avoid anyone reusing them in the future. (For example the class  ConfigurableInfoBar.cs )

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -87,6 +87,46 @@ public IInfoBar AttachInfoBar(Guid toolWindowGuid, string message, SonarLintImag
             return AttachInfoBarImpl(toolWindowGuid, message, imageMoniker);
         }
 
+        public IInfoBar AttachInfoBarWithButtonMainWindow(string message, string buttonText, SonarLintImageMoniker imageMoniker)
```
bigfluffycookie, line 90: Almost identical to the already existing methods expect they dont need a guid and they call AttachInfoBarMainWindowImpl instead of AttachInfoBarmpl

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -102,8 +142,8 @@ public void DetachInfoBar(IInfoBar currentInfoBar)
             currentInfoBar.Close();
 
             ThreadHelper.ThrowIfNotOnUIThread();
-            IVsInfoBarHost host;
-            if (TryGetInfoBarHost(wrapper.Frame, out host))
+            IVsInfoBarHost host = wrapper.Host;
```
bigfluffycookie, line 145: If its a main window gold bar we already know the host, else try get it.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -132,6 +172,29 @@ private IInfoBar AttachInfoBarImpl(Guid toolWindowGuid, string message, SonarLin
             return null;
         }
 
+        private IInfoBar AttachInfoBarMainWindowImpl(string message, SonarLintImageMoniker imageMoniker, ButtonStyle buttonStyle = ButtonStyle.Button, params string[] buttonTexts)
```
bigfluffycookie, line 175: Pretty similar to the other implementation except for "shell.GetProperty((int)__VSSPROPID7.VSSPROPID_MainWindowInfoBarHost, out object value);" and I am also adding the info bar to the host directly. here instead of going through the method TryAddInfoBarToFrame()

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -132,6 +172,29 @@ private IInfoBar AttachInfoBarImpl(Guid toolWindowGuid, string message, SonarLin
...
+            Debug.Assert(shell != null);
+
+            shell.GetProperty((int)__VSSPROPID7.VSSPROPID_MainWindowInfoBarHost, out object value);
+
+            InfoBarModel model = CreateModel(message, buttonTexts, buttonStyle, imageMoniker);
+            IVsInfoBarUIFactory infoBarUIFactory = serviceProvider.GetService(typeof(SVsInfoBarUIFactory)) as IVsInfoBarUIFactory;
+            IVsInfoBarUIElement uiElement;
+
+            if (TryCreateInfoBarUI(infoBarUIFactory, model, out uiElement)
+                && value is IVsInfoBarHost host)
```
duncanp-sonar, line 189: I'd put the `value is IVsInfoBarHost host` condition first. That way the InfoBarUI won't be created if the condition evaluates to `false`.

It might also be simpler to do the check separately after line 182, and early-out if the returned value isn't a host.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
duncanp-sonar: This is a bit untidy because the calling code (i.e. this method) has to know to try the host property first, before trying the frame property.

It doesn't matter too much since `PrivateInforBarWrapper` is a private class so it can be only used within this class.

However, it would be easy to make it neater: move the code in lines 144 to 149 into the `PrivateInfroBarWrapper.Close` method i.e. make the private class responsible for clearing up the info bar UI element and the host.

That hides the special cleanup behaviour from the caller. It might also mean that the ``Frame` and `Host` properties in the private class can both be made `private` if nothing else in the parent class uses them.

It shouldn't affect the existing tests since they can't see the `PrivateInfoBarWrapper` class anyway.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -227,10 +290,23 @@ public PrivateInfoBarWrapper(IVsWindowFrame frame, IVsInfoBarUIElement uiElement
...
+                this.InfoBarUIElement = uiElement;
+
+                this.Advise();
+            }
+
             internal IVsInfoBarUIElement InfoBarUIElement { get; }
 
             internal IVsWindowFrame Frame { get; }
 
+            internal IVsInfoBarHost Host { get; }
```
duncanp-sonar, line 308: There's a design smell here: the way the class is constructed means only one of the fields will ever have a value, and the caller needs to be aware of this to use the class properly. The refactoring suggested above removes/hides this.

The wrapper class could probably be refactored so that it only accepts a `host`, since in the non-Main window case we still need to get the host via the frame. However, that would involve refactoring the existing code further.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
duncanp-sonar: It's a risky assumption because the tool window and main window method implementations are separate, and because they are separate they could be modified separately.

I think it's ok to skip testing the two overrides of  `AttachXXXMainWindow` because they are simple overloads that call down to the base method.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
duncanp-sonar: Update: I've deleted two of the unused `ConfigurableXXX` classes

src/Core/InfoBar/IInfoBarManager.cs
```diff
@@ -54,6 +54,15 @@ public interface IInfoBarManager
...
 
+        /// <summary>
+        /// Attach an info bar with a message and multiple buttons to the main window
+        /// </summary>
+        /// <remarks>
+        /// If there is one buttonText it will be rendered as a button on the info bar.
+        /// If there are multiple buttonTexts they will all be rendered as hyperlinks.
+        /// </remarks>
+        IInfoBar AttachInfoBarToMainWindow(string message, SonarLintImageMoniker imageMoniker, params string[] buttonTexts);
+
```
duncanp-sonar, line 65: As discussed offline, I removed the two overloads to just leave the one method we need for now.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -303,6 +303,62 @@ public void AttachInfoBarWithButtons_ButtonClickedEvent_RaisedWithTheRightButton
...
         }
 
+        #region MainWindow tests
+        // The ToolWindow tests above cover the common implementation of gold bar functionality.
+        // These MainWindow tests just cover a few extra main window specific cases.
```
duncanp-sonar, line 308: I added a couple more tests, and reworded the test comment you had added.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -303,6 +303,62 @@ public void AttachInfoBarWithButtons_ButtonClickedEvent_RaisedWithTheRightButton
...
+
+        [TestMethod]
+        public void InfoBarManager_AttachInfoBarToMainWindowAndDetach_ArgChecks()
```
duncanp-sonar, line 311: Low value tests to increase coverage.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -303,6 +303,62 @@ public void AttachInfoBarWithButtons_ButtonClickedEvent_RaisedWithTheRightButton
...
+
+            Action act = () => testSubject.AttachInfoBarToMainWindow(null, SonarLintImageMoniker.OfficialSonarLintMoniker);
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("message");
+
+            act = () => testSubject.AttachInfoBarToMainWindow("a message", SonarLintImageMoniker.OfficialSonarLintMoniker, null);
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("buttonTexts");
+        }
+
+        [TestMethod]
+        public void InfoBarManager_AttachInfoBarToMainWindowAndDetach_AttachAndDetach()
```
duncanp-sonar, line 323: This is the test from your first commit. I appended `_AttachAndDetach` to the name, and added the assertion on line 335 that the returned infobar isn't null.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -303,6 +303,62 @@ public void AttachInfoBarWithButtons_ButtonClickedEvent_RaisedWithTheRightButton
...
+
+            serviceProviderMock.Verify(x => x.GetService(typeof(SVsInfoBarUIFactory)), Times.Once);
+            host.Verify(x => x.AddInfoBar(It.IsAny<IVsInfoBarUIElement>()), Times.Once);
+
+            testSubject.DetachInfoBar(infoBar);
+            host.Verify(x => x.RemoveInfoBar(It.IsAny<IVsInfoBarUIElement>()), Times.Once);
+        }
+
+        [TestMethod]
+        public void InfoBarManager_AttachInfoBarToMainWindow_HostIsNotAvailable_ReturnsNull()
```
duncanp-sonar, line 345: New test - early out and return null if the host isn't available.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -327,6 +383,16 @@ private static ConfigurableVsInfoBarHost RegisterFrameInfoBarHost(ConfigurableVs
             return host;
         }
 
+        private static void RegisterMainWindowHostWithShell(Mock<IServiceProvider> serviceProvider, IVsInfoBarHost host)
```
duncanp-sonar, line 386: I tweaked this helper method so the caller passes in the host to return. Needed so that one of the new tests can simulate the service returning a null host.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -94,39 +118,49 @@ public void DetachInfoBar(IInfoBar currentInfoBar)
                 throw new ArgumentNullException(nameof(currentInfoBar));
             }
 
-            if (!(currentInfoBar is PrivateInfoBarWrapper wrapper))
```
duncanp-sonar, line 97: Misc: removed unused variable to fix code smell

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -176,18 +210,6 @@ private static InfoBarModel CreateModel(string message, IEnumerable<string> butt
...
-            IVsInfoBarHost infoBarHost;
-            if (TryGetInfoBarHost(frame, out infoBarHost))
-            {
-                infoBarHost.AddInfoBar(uiElement);
-                return true;
-            }
-
-            return false;
-        }
-
```
duncanp-sonar, line 190: No longer required. The `TryGetInfoBarHost` is still used, but the common low-level implementation method is now responsible for adding the info bar to the host.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -201,6 +223,18 @@ private static bool TryGetInfoBarHost(IVsWindowFrame frame, out IVsInfoBarHost i
...
+        private bool TryGetMainWindowInfoBarHost(out IVsInfoBarHost infoBarHost)
+        {
+            var shell = serviceProvider.GetService(typeof(SVsShell)) as IVsShell;
+            Debug.Assert(shell != null);
+
+            shell.GetProperty((int)__VSSPROPID7.VSSPROPID_MainWindowInfoBarHost, out object value);
+
+            infoBarHost = value as IVsInfoBarHost;
+            return infoBarHost != null;
+        }
```
duncanp-sonar, line 236: I moved some of your code into this helper method.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -214,39 +248,50 @@ private class PrivateInfoBarWrapper : IInfoBar, IVsInfoBarUIEvents
             private uint? cookie;
             private bool isClosed;
 
-            public PrivateInfoBarWrapper(IVsWindowFrame frame, IVsInfoBarUIElement uiElement)
+            public PrivateInfoBarWrapper(IVsInfoBarHost host, IVsInfoBarUIElement uiElement)
```
duncanp-sonar, line 251: Changed the wrapper so it works more generically with the `host`, so it can handle both the ToolWindow and main window cases.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -214,39 +248,50 @@ private class PrivateInfoBarWrapper : IInfoBar, IVsInfoBarUIEvents
...
+            public void Close()
+            {
+                CloseUIElement();
+
+                ThreadHelper.ThrowIfNotOnUIThread();
+                if (Host != null )
+                {
+                    Host.RemoveInfoBar(InfoBarUIElement);
+                }
+            }
```
duncanp-sonar, line 293: This is the method that is called by `InfoBarManager.Detach` in lines 102-109. It is functionally the same.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -303,6 +303,62 @@ public void AttachInfoBarWithButtons_ButtonClickedEvent_RaisedWithTheRightButton
...
+            serviceProviderMock.Setup(x => x.GetService(typeof(SVsInfoBarUIFactory))).Returns(new ConfigurableVsInfoBarUIFactory());
+
+            var host = new Mock<IVsInfoBarHost>();
+            RegisterMainWindowHostWithShell(serviceProviderMock, host.Object);
+
+            var testSubject = new InfoBarManager(serviceProviderMock.Object);
+            host.Verify(x => x.AddInfoBar(It.IsAny<IVsInfoBarUIElement>()), Times.Never);
+
+            var infoBar = testSubject.AttachInfoBarToMainWindow("message", default);
+            infoBar.Should().NotBeNull();
```
duncanp-sonar, line 335: I added this assertion to the existing test. Unfortunately, the test is failing on the CI machine for VS2019 but not VS2022. I'm trying to work out why, but debugging the VS2019 tests locally is not straightforward...

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -33,6 +33,12 @@
 
 namespace SonarLint.VisualStudio.Integration.UnitTests
 {
+    // NB we need to use the ConfigurableServiceProvider for these tests because it correctly handles
+    // COM type equivalance for embedded interop types such as SVsInfoBarUIFactory.
+    // Otherwise, we can get unexpected failures such as tests passing in the VS2022 build but not in
+    // the VS2019 build
+    // See https://learn.microsoft.com/en-us/dotnet/framework/interop/type-equivalence-and-embedded-interop-types
+
```
duncanp-sonar, line 41: This caused a lot of problems. I eventually fixed it by changing the new tests to use the old `ConfigurableXXX` classes, rather than using `Mock<x>`.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -303,6 +309,60 @@ public void AttachInfoBarWithButtons_ButtonClickedEvent_RaisedWithTheRightButton
...
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("message");
+
+            act = () => testSubject.AttachInfoBarToMainWindow("a message", SonarLintImageMoniker.OfficialSonarLintMoniker, null);
+            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("buttonTexts");
+        }
+
+        [TestMethod]
+        public void InfoBarManager_AttachInfoBarToMainWindowAndDetach_AttachAndDetach()
+        {
+            var sp = new ConfigurableServiceProvider();
```
duncanp-sonar, line 331: This test class is written in the "legacy" style i.e. class-instance level variables and a `TestInitialize` method. I wanted to make the new tests independent so they don't rely on the `TestInitialize. There is already an instance-level variable called `serviceProvider` so I used `sp` here.

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
```diff
@@ -303,6 +309,60 @@ public void AttachInfoBarWithButtons_ButtonClickedEvent_RaisedWithTheRightButton
...
+            sp.RegisterService(typeof(SVsInfoBarUIFactory), new ConfigurableVsInfoBarUIFactory());
+
+            var host = new ConfigurableVsInfoBarHost();
+            RegisterMainWindowHostWithShell(sp, host);
+
+            var testSubject = new InfoBarManager(sp);
```
duncanp-sonar, line 337: The test class does have a `CreateTestSubject` method, but it does extra work. I didn't want to refactor/clean-up the existing tests as part of this PR as that would get confusing.

(I actually started doing the cleanup separately this morning until I noticed the problem with the VS2019 CI build)

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
bigfluffycookie: Yea, it is a bit unfortunate that we have to deal with it like this but atleast it works now!

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
bigfluffycookie: Wouldn't it make more sense to have "_AttachesAndDetaches" instead? just because else reading Attach...Detach_AtachAndDetach is a bit confusing to me

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
bigfluffycookie: Makes sense, I think I had to do the same with "shell" -> shellMock somewhere here

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
bigfluffycookie: Yea agree, I thought about it too but I think its better to have a "clean up" ticket somewhere and put it in the backlog.

src/Integration.Vsix/InfoBar/InfoBarManager.cs
```diff
@@ -84,7 +84,31 @@ public IInfoBar AttachInfoBar(Guid toolWindowGuid, string message, SonarLintImag
...
+            {
+                throw new ArgumentNullException(nameof(message));
+            }
+
+            if (buttonTexts == null)
+            {
+                throw new ArgumentNullException(nameof(buttonTexts));
+            }
+
+            ThreadHelper.ThrowIfNotOnUIThread();
```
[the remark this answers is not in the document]
bigfluffycookie: Yea probably best to add a ticket to the other Threading clean up stuff

src/Progress.TestFramework/ConfigurableServiceProvider.cs
```diff
@@ -25,6 +25,16 @@
 
 namespace SonarLint.VisualStudio.Progress.UnitTests
 {
+    /// <summary>
+    /// ServiceProvider that correctly handles COM type equivalence for embedded interop types.
+    /// See https://learn.microsoft.com/en-us/dotnet/framework/interop/type-equivalence-and-embedded-interop-types
+    /// Mock<ISeviceProvider> does not handle this, which *can* lead to requests for services
```
bigfluffycookie, line 31: Mock<ISe*_r_*vice>

bigfluffycookie (commented):
Looks good!

src/Integration.Vsix.UnitTests/InfoBar/InfoBarManagerTests.cs
duncanp-sonar: Yep, that's the alternative. I was experimenting with a more "clean as you code" approach. I didn't want to pollute the existing PR with unrelated cleanup, but the trouble with creating a ticket is that we seldom get around to them. Also, since I've just been working on the tests I'm more familiar with them.

Rather than creating a ticket, I was going to spend a maximum of fifteen/twenty minutes doing some low-risk cleanup, and then send it for review. If that approach worked, we could try allocating a time budget for each sprint for miscellaneous cleanup e.g. two hours, to make sure we don't lose focus on the main purpose of the sprint. Something to discuss offline.

PR #4261, opened 2023-06-01
Add MigrationChecker class and GoldBarController skeleton. Also remove AutoMigrationHandler.

bigfluffycookie:
part of https://github.com/SonarSource/sonarlint-visualstudio/issues/4254


src/ConnectedMode/Migration/IGoldBarController.cs
```diff
@@ -0,0 +1,36 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Migration
+{
+    public interface IGoldBarController
```
bigfluffycookie, line 25: Empty goldbar controller so I could properly test the MigrationChecker

src/ConnectedMode/Migration/MigrationChecker.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            activeSolutionTracker.ActiveSolutionChanged += OnActiveSolutionChanged;
+
+            // Initial check incase event was fired before we registered.
+            DisplayGoldBarIfMigrationIsNeeded();
+        }
+
+        private void OnActiveSolutionChanged(object sender, ActiveSolutionChangedEventArgs args)
+        {
+            if (args.IsSolutionOpen)
```
bigfluffycookie, line 57: Do we need to close the goldbar if the solution closes?

src/ConnectedMode.UnitTests/Migration/MigrationCheckerTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<MigrationChecker, MigrationChecker>(
+                MefTestHelpers.CreateExport<IActiveSolutionTracker>(),
+                MefTestHelpers.CreateExport<IGoldBarController>(),
+                MefTestHelpers.CreateExport<IConfigurationProvider>(),
+                MefTestHelpers.CreateExport<IObsoleteConfigurationProvider>()); 
+        }
+
+        [TestMethod]
```
bigfluffycookie, line 42: This test is more or less the same as it was in the AutoMigrationHandlerTests.cs

src/ConnectedMode/Migration/MigrationChecker.cs
duncanp-sonar: Yes, I think we should.

src/ConnectedMode/Migration/IGoldBarController.cs
duncanp-sonar: 'IGoldBarController` is pretty generic. How about something more specific e.g. `IMigrationGoldBarController`.

Or perhaps something even more specific such as `IMigrationPrompt` with a method called `Show` - the fact that it is displaying a gold bar is an implementation detail from the point of view of the caller.

duncanp-sonar (approved):
Looks good

PR #4266, opened 2023-06-02
(2) Remove IProjectSystemHelper.GetFilteredSolutionProjects

duncanp-sonar:
- no longer called. With the new binding model, we don't need to check which projects contain supported languages since 1) we won't make changes to individual project files, and 2) we download QPs for all supported languages if a project does not have any supported languages.

Part of #4120

### DIFF
--- a/src/Integration/LocalServices/IProjectSystemHelper.cs
+++ b/src/Integration/LocalServices/IProjectSystemHelper.cs
@@ -78,14 +78,6 @@ namespace SonarLint.VisualStudio.Integration
         IEnumerable<Project> GetSolutionProjects();
 
         /// <summary>
-        /// Returns only the filtered project based on a common filter.
-        /// This should only be called after we connected to a SonarQube server,
-        /// since some of the filtering is SonarQube server instance specific.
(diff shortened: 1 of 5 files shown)


src/Integration/LocalServices/IProjectSystemHelper.cs
```diff
@@ -77,14 +77,6 @@ public interface IProjectSystemHelper : ILocalService, IVsHierarchyLocator
         /// </summary>
         IEnumerable<Project> GetSolutionProjects();
 
-        /// <summary>
-        /// Returns only the filtered project based on a common filter.
-        /// This should only be called after we connected to a SonarQube server,
-        /// since some of the filtering is SonarQube server instance specific.
-        /// <seealso cref="IProjectSystemFilter"/> which is used internally.
-        /// </summary>
-        IEnumerable<Project> GetFilteredSolutionProjects();
```
duncanp-sonar, line 86: This was previously used by the binding code, but I removed that call during earlier cleanup. Nothing else uses this method now, so I've deleted it from the interface.

That in turn means the IProjectSystemFilter is no longer used. I'll remove that in the next PR.

Issue #4269, opened 2023-06-02
Change INotificationService to not be a singleton


bigfluffycookie:
This allows us to reuse the notification service for multiple notifications at the same time. Now the Node service can show for example an UnsupportedNodeVersionNotification and a MigrationPrompt notification at the same time. This way the notification service is also only responsible for one type of notification at a time.

PR #4271, opened 2023-06-02
Change INotificationService to not be a singleton.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4269

### DIFF
--- a/src/Core.UnitTests/Notifications/NotificationServiceTests.cs
+++ b/src/Core.UnitTests/Notifications/NotificationServiceTests.cs
@@ -48,7 +48,7 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Notifications
         [TestMethod]
         public void CheckIsNonSharedMefComponent()
         {
-            MefTestHelpers.IsMefComponentNonShared(typeof(NotificationService));
+            MefTestHelpers.CheckIsNonSharedMefComponent<NotificationService>();
         }
 
         [TestMethod]
--- a/src/Core.UnitTests/Notifications/NotificationServiceTests.cs
+++ b/src/Core.UnitTests/Notifications/NotificationServiceTests.cs
@@ -20,6 +20,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.Composition;
 using System.Linq;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
@@ -42,6 +43,14 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Notifications
                 MefTestHelpers.CreateExport<IDisabledNotificationsStorage>(),
                 MefTestHelpers.CreateExport<IThreadHandling>(),
                 MefTestHelpers.CreateExport<ILogger>());
+        }
+
+        [TestMethod]
+        public void NotificationServiceIsNonShared()
+        {
+            var customAttributes = typeof(NotificationService).GetCustomAttributes(typeof(PartCreationPolicyAttribute), true);
+            var partCreationPolicyAttribute = (PartCreationPolicyAttribute)customAttributes[0];
+            partCreationPolicyAttribute.CreationPolicy.Should().Be(CreationPolicy.NonShared);
         }
 
         [TestMethod]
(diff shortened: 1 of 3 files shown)


src/Core/Notifications/INotificationService.cs
```diff
@@ -33,7 +33,7 @@ public interface INotificationService : IDisposable
     }
```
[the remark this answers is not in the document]
duncanp-sonar: We don't, but it would be better if we did. You could try adding one using reflection i.e. using `typeof(X).GetCustomAttributes(...)`

duncanp-sonar (commented):
The change looks good ;-)

Could you try to add a test please?

Also, could you add some comments to ticket #4269 giving a brief explanation of why we are changing the behaviour and how we now expect it to be used? i.e. document the design decision we made. It would be worth adding a comment to the service documenting how callers are expected to use it.

src/Core.UnitTests/Notifications/NotificationServiceTests.cs
```diff
@@ -44,6 +45,12 @@ public void MefCtor_CheckIsExported()
                 MefTestHelpers.CreateExport<ILogger>());
         }
 
+        [TestMethod]
+        public void CheckIsNonSharedMefComponent()
+        {
+            MefTestHelpers.IsMefComponentNonShared(typeof(NotificationService));
```
duncanp-sonar, line 51: The method is returning a bool and this test doesn't contain any assertions so this test will always pass if the type has a `PartCreationPolicy`.

It might be better to structure the helper method like `CheckTypeCanBeImported` e.g. `CheckIsSingletonMefComponent` / `CheckIsNonSharedMefComponent`.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -180,5 +180,27 @@ public static void Compose(Assembly exportingAssembly, params object[] importers
...
             container.Compose(batch);
         }
+
+        /// <summary>
+        /// Check if type has creation policy non shared.
+        /// </summary>
+        public static bool IsMefComponentNonShared(Type type)
+        {
+            var customAttributes = type.GetCustomAttributes(typeof(PartCreationPolicyAttribute), true);
+            var partCreationPolicyAttribute = (PartCreationPolicyAttribute)customAttributes[0];
```
duncanp-sonar, line 190: This will fail with an array out-of-bounds exception if it is called on a type that doesn't have the attribute. Test failures might be easier to debug with a more explicit assertion e.g. 
`customAttributes.Should().HaveCount(1)`.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -180,5 +180,27 @@ public static void Compose(Assembly exportingAssembly, params object[] importers
             using var container = new CompositionContainer(catalog);
             container.Compose(batch);
         }
+
+        /// <summary>
+        /// Check if type has creation policy non shared.
+        /// </summary>
+        public static bool IsMefComponentNonShared(Type type)
```
duncanp-sonar, line 187: If you make the method generic it would be syntactically slightly easier to call e.g.

```
public static void CheckIsSingletonMefComponent<T>()
{
  var customAttributes = type.GetCustomAttributes(typeof(T), true);
  ...
}
```

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -180,5 +180,27 @@ public static void Compose(Assembly exportingAssembly, params object[] importers
...
+        /// <summary>
+        /// Check if type has creation policy shared.
+        /// </summary>
+        public static bool IsMefComponentShared(Type type)
+        {
+            var customAttributes = type.GetCustomAttributes(typeof(PartCreationPolicyAttribute), true);
+            var partCreationPolicyAttribute = (PartCreationPolicyAttribute)customAttributes[0];
+
+            return partCreationPolicyAttribute.CreationPolicy == CreationPolicy.Shared;
+        }
```
duncanp-sonar, line 204: This method is a duplicate of the other, except for the enum value being compared against.
I'd move the code into a helper method.

duncanp-sonar (commented):
Nice: a few comments about the helper methods.

duncanp-sonar (approved):
Nice!

PR #4277, opened 2023-06-02
(2) Drop check of plugins available on the server

### LINKED ISSUE #4272: Drop check for supported plugins/project from ConnectionWorkflow
### Description

Simplify the connection code by removing the checks for used/supported languages. Users will always be able to connect regardless of the SonarQube version they are connected to or what languages they are using in their solution.

Note: dropping this check means users will be able to connect to a server that does not have any Sonar projects, so they wil be shown an empty project list.

#### Historical note
Originally, the connection did the following checks before connecting:
1) find the list of languages used in the solution
2) fetch the list of supported plugins installed on the server
3) check that there is at least one language used in the solution that has a supported plugin on the server.

The thinking was that there was no point in allowing the user to set up Connected Mode if it wasn't going to do anything i.e. there were no supported languages in the solution.

This was more likely in the past for a couple of reasons: originally only C#/VB.NET were supported, and users could delete the Sonar plugins from the server if they wanted to.

The situation is different now:
* deleting Sonar plugins from the server is no longer supported (since at least SQv8.9 i.e. LTS - 1).
* the Secrets language checks all text files, so it applies to all projects
* the Connected Mode settings files will no longer be visible to the users, so downloading QPs that will not be used will have no functional impact, other than slightly increasing the time taken to fetch the QPs.

duncanp-sonar:
Fixes #4272


src/Integration.UnitTests/Connection/ConnectionWorkflowTests.cs
```diff
@@ -65,14 +64,7 @@ public void TestInit()
...
             this.host.SetActiveSection(ConfigurableSectionController.CreateDefault());
             this.host.SonarQubeService = this.sonarQubeServiceMock.Object;
-            this.projectSystemHelper = new ConfigurableVsProjectSystemHelper(this.serviceProvider);
-
-            this.sonarQubeServiceMock.Setup(x => x.GetAllPluginsAsync(It.IsAny<CancellationToken>()))
-                .ReturnsAsync(new List<SonarQubePlugin>
-                {
-                    new SonarQubePlugin(MinimumSupportedSonarQubePlugin.CSharp.Key, MinimumSupportedSonarQubePlugin.CSharp.MinimumVersion),
-                    new SonarQubePlugin(MinimumSupportedSonarQubePlugin.VbNet.Key, MinimumSupportedSonarQubePlugin.VbNet.MinimumVersion)
-                });
```
duncanp-sonar, line 75: The product code no longer fetches the list of plugins from the server so the test setup is simpler.

src/Integration.UnitTests/Connection/ConnectionWorkflowTests.cs
```diff
@@ -227,77 +204,6 @@ public async Task ConnectionWorkflow_ConnectionStep_Credentials_Invalid()
             AssertCredentialsNotStored(); // Connection was rejected by SonarQube
         }
 
-        [TestMethod]
-        public async Task ConnectionWorkflow_ConnectionStep_WhenMissingCSharpPluginAndVBNetPlugin_AbortsWorkflowAndDisconnects()
```
duncanp-sonar, line 231: I deleted all of the tests that depended on the plugins available on the server.

src/Integration.UnitTests/Connection/ConnectionWorkflowTests.cs
```diff
@@ -423,7 +259,6 @@ public async Task ConnectionWorkflow_ConnectionStep_UnsuccessfulConnection()
                 c.Should().Be(connectionInfo, "Unexpected connection");
                 p.Should().BeNull("Not expecting any projects");
             };
-            this.projectSystemHelper.Projects = new[] { new ProjectMock("foo.csproj") { ProjectKind = ProjectSystemHelper.CSharpProjectKind } };
```
duncanp-sonar, line 426: The project system helper isn't required any more, so I cleaned up those bits from the tests.

src/Integration.UnitTests/Connection/ConnectionWorkflowTests.cs
```diff
@@ -544,7 +374,8 @@ public async Task ConnectionWorkflow_ConnectionStep_10KProjectsAndKeyIsNotPartOf
 
             AssertCredentialsStored(connectionInfo);
 
-            logger.AssertOutputStrings(4);
```
duncanp-sonar, line 547: Checking for a specific number of output messages is quite fragile and it's not something we usually do anymore.
I've replaced it with a more specific check in this case that a message about the expected project has been logged.

src/Integration.UnitTests/Connection/ConnectionWorkflowTests.cs
```diff
@@ -596,8 +425,6 @@ public async Task ConnectionWorkflow_ConnectionStep_10KProjectsAndKeyIsPartOfIt(
             ((ConfigurableUserNotification)this.host.ActiveSection.UserNotifications).AssertNoNotification(NotificationIds.FailedToConnectId);
 
             AssertCredentialsStored(connectionInfo);
-
-            logger.AssertOutputStrings(2);
```
duncanp-sonar, line 600: Both messages have been removed.

src/Integration.UnitTests/Connection/MinimumSupportedSonarQubePluginTests.cs
```diff
@@ -1,43 +0,0 @@
-﻿/*
```
duncanp-sonar, line 1: The `MinimumSupportedSonarQubePlugin` class has been dropped.

src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -163,13 +156,6 @@ private void OnProjectsChanged(ConnectionInformation connection, IEnumerable<Son
                         new Credential(connection.UserName, connection.Password.ToUnsecureString()));
                 }
 
-                var isCompatible = await this.AreSolutionProjectsAndSonarQubePluginsCompatibleAsync(controller, notifications,
```
duncanp-sonar, line 166: Dropped the compatibility check and all the methods and resources strings associated with it.

src/TestInfrastructure/Framework/ConfigurableHost.cs
```diff
@@ -106,8 +104,6 @@ public void SetActiveSection(ISectionController section)
             this.VisualStateManager.SyncCommandFromActiveSection();
         }
 
-        public ISet<Language> SupportedPluginLanguages { get; } = new HashSet<Language>();
-
```
duncanp-sonar, line 110: The property is no longer used anywhere so I removed it from the interfaces and implementations.

bigfluffycookie (approved):
Looks good

PR #4280, opened 2023-06-05
Add IServerHotspotStore interface

### LINKED ISSUE #4268: Get local rule descriptions for  Hotspots

ugras-ergun-sonarsource:
fixes #4268


src/IssueViz.Security/Hotspots/ServerHotspotStore.cs
```diff
@@ -0,0 +1,104 @@
...
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Hotspots
+{
+    internal interface IServerHotspotStore
+    {
+        Task UpdateAsync(string projectKey, string branch, CancellationToken token);
+
+        Task<IList<SonarQubeHotspotSearch>> GetServerHotspotsAsync(string projectKey, string branch, CancellationToken token);
+
+        event EventHandler<ServerHotspotStoreUpdatedEventArgs> ServerHotspotStoreUpdated;
```
georgii-borovinskikh-sonarsource, line 37: I think having an event here is overkill. We really only have one consumer of this info, so why not just call that directly? The GetServerHotspotsAsync method should stay for when the local issues change and need to be rematched

src/IssueViz.Security/Hotspots/ServerHotspotStore.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        public event EventHandler<ServerHotspotStoreUpdatedEventArgs> ServerHotspotStoreUpdated;
+
+        [ImportingConstructor]
+        public ServerHotspotStore(ISonarQubeService sonarQubeService)
+        {
+            this.sonarQubeService = sonarQubeService;
+        }
+
+        public async Task UpdateAsync(string projectKey, string branch, CancellationToken token)
```
georgii-borovinskikh-sonarsource, line 55: This method keeps info for all compositekeys. While it could be useful, we need to assess that versus the likelihood of this being unused (aka just a memory leak)

src/IssueViz.Security.UnitTests/Hotspots/ServerHotspotStoreTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            return new ServerHotspotStore(sonarQubeService);
+        }
+
+        [TestCleanup]
+        public void CleanUp()
+        {
+            ServerHotspotStore.serverHotspots.Clear();
+        }
+
+        private class EventCounter
```
georgii-borovinskikh-sonarsource, line 191: Removing the event would simplify these tests as well

src/IssueViz.Security/Hotspots/ServerHotspotStore.cs
```diff
@@ -0,0 +1,104 @@
...
+            RaiseServerHotspotStoreUpdated(projectKey, branch);
+        }
+
+        public async Task<IList<SonarQubeHotspotSearch>> GetServerHotspotsAsync(string projectKey, string branch, CancellationToken token)
+        {
+            var compositeKey = CreateCompositeKey(projectKey, branch);
+
+            if (!serverHotspots.ContainsKey(compositeKey))
+            {
+                await UpdateAsync(projectKey, branch, token);
```
georgii-borovinskikh-sonarsource, line 76: Shouldn't this already be updated? Or you're not planning to react to the branch changed event?

georgii-borovinskikh-sonarsource (changes requested):

georgii-borovinskikh-sonarsource (approved):

PR #4281, opened 2023-06-05
(3) Handle missing server plugins

### LINKED ISSUE #4272: Drop check for supported plugins/project from ConnectionWorkflow

duncanp-sonar:
Fixes #4272

### DIFF
--- a/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
@@ -166,53 +166,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
         }
 
         [TestMethod]
-        public async Task DownloadQualityProfile_Success()
-        {
-            var configPersister = new ConfigurableConfigurationProvider();
-            this.serviceProvider.RegisterService(typeof(IConfigurationPersister), configPersister);
-
-            // Arrange
-            const string QualityProfileName = "SQQualityProfileName";
-            const string ProjectName = "SQProjectName";
-
-            var notifications = new ConfigurableProgressStepExecutionEvents();
-            var progressAdapter = new FixedStepsProgressAdapter(notifications);
-
-            var bindingConfig = new Mock<IBindingConfig>().Object;
-
-            var language = Language.VBNET;
-            SonarQubeQualityProfile profile = this.ConfigureQualityProfile(language, QualityProfileName);
-
-            var configProviderMock = new Mock<IBindingConfigProvider>();
-            configProviderMock.Setup(x => x.GetConfigurationAsync(profile, language, BindingConfiguration.Standalone, CancellationToken.None))
-                .ReturnsAsync(bindingConfig);
-
-            var bindingArgs = CreateBindCommandArgs("key", ProjectName, new ConnectionInformation(new Uri("http://connected")));
-            var testSubject = this.CreateTestSubject(bindingArgs, configProviderMock.Object, languagesToBind: new[] { language });
-
-            List<Language> supportedLanguages = new List<Language>() { language };
-
-            // Act
-            var result = await testSubject.DownloadQualityProfileAsync(progressAdapter, CancellationToken.None);
-
-            // Assert
-            result.Should().BeTrue();
-            testSubject.InternalState.BindingConfigs.Should().ContainKey(language);
-            testSubject.InternalState.BindingConfigs[language].Should().Be(bindingConfig);
-            testSubject.InternalState.BindingConfigs.Count().Should().Be(1);
-
-            testSubject.InternalState.QualityProfiles[language].Should().Be(profile);
-
-            notifications.AssertProgress(0.0, 1.0);
-            notifications.AssertProgressMessages(Strings.DownloadingQualityProfileProgressMessage, string.Empty);
-
-            logger.AssertOutputStrings(1);
-            var expectedOutput = string.Format(Strings.SubTextPaddingFormat,
-                string.Format(Strings.QualityProfileDownloadSuccessfulMessageFormat, QualityProfileName, string.Empty, language.Name));
-            logger.AssertOutputStrings(expectedOutput);
-        }
-
-        [TestMethod]
         public async Task DownloadQualityProfile_SavesConfiguration()
         {
             var configPersister = new ConfigurableConfigurationProvider();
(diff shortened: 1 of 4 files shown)


src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -165,53 +165,6 @@ private static ServerExclusions CreateSettings()
             };
         }
 
-        [TestMethod]
-        public async Task DownloadQualityProfile_Success()
```
duncanp-sonar, line 169: This test is a subset of the modified test below i.e. it doesn't add anything in terms of scenario coverage, so I deleted it.

duncanp-sonar:
Previously the binding workflow would abort if a QP could not be downloaded. Now it logs a message and continues.
The screen shot below was taken using a Community edition i.e. one that does not have the CFamily plugin:

image


In addition, I've tweaked the messages in the Team Explorer progress dialogue. Previously the bit that says `Downloading quality profile: VB.NET would have been blank`.

image

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -97,52 +97,78 @@ internal class BindingProcessImpl : IBindingProcess
 
         public async Task<bool> DownloadQualityProfileAsync(IProgress<FixedStepsProgress> progress, CancellationToken cancellationToken)
         {
-            var languageList = this.GetBindingLanguages();
+            var languageList = GetBindingLanguages();
 
             var languageCount = languageList.Count();
             int currentLanguage = 0;
-            progress?.Report(new FixedStepsProgress(Strings.DownloadingQualityProfileProgressMessage, currentLanguage, languageCount));
```
duncanp-sonar, line 104: I removed this initial progress message. It only appeared briefly, until the first QP was downloaded. The new messaging is hopefully more useful.

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -97,52 +97,78 @@ internal class BindingProcessImpl : IBindingProcess
 
         public async Task<bool> DownloadQualityProfileAsync(IProgress<FixedStepsProgress> progress, CancellationToken cancellationToken)
         {
-            var languageList = this.GetBindingLanguages();
```
duncanp-sonar, line 100: Miscellaneous: removed `this.` from the changed method to get rid of some code smells. I didn't do it for the whole file in this PR as that would have been too noisy.

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -97,52 +97,78 @@ internal class BindingProcessImpl : IBindingProcess
...
+                    host.Logger.WriteLine(string.Format(Strings.SubTextPaddingFormat,
                         string.Format(Strings.FailedToCreateBindingConfigForLanguage, language.Name)));
                     return false;
                 }
 
-                this.InternalState.BindingConfigs[language] = bindingConfig;
+                InternalState.BindingConfigs[language] = bindingConfig;
 
-                currentLanguage++;
-                progress?.Report(new FixedStepsProgress(string.Empty, currentLanguage, languageCount));
```
duncanp-sonar, line 137: Progress reporting moved to to before the download starts. Previously the message was blank, which wasn't very helpful. Effectively it meant the progress message just updated the progress bar. Now it shows a message too.

src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -97,52 +97,78 @@ internal class BindingProcessImpl : IBindingProcess
...
+            // There are valid scenarios in which a language plugin will not be available on the server:
+            // 1) the CFamily plugin does not ship in Community edition (nor do any other commerical plugins)
+            // 2) a recently added language will not be available in older-but-still-supported SQ versions
+            //      e.g. the "secrets" language
+            // The unavailability of a language should not prevent binding from succeeding.
+
+            // Note: the historical check that plugins meet a minimum version was removed. 
+            // See https://github.com/SonarSource/sonarlint-visualstudio/issues/4272
+
+            var qualityProfileInfo = await WebServiceHelper.SafeServiceCallAsync(() =>
```
duncanp-sonar, line 156: This logic hasn't changed; it was just moved to a helper method.
The comment above is new.

src/Integration/Resources/Strings.resx
```diff
@@ -314,8 +314,8 @@ ocalization note: '[try again]()' is a special markup used to translate the text
     <comment>Server context menu command display text to show all projects (inc. unbound projects)</comment>
   </data>
   <data name="DownloadingQualityProfileProgressMessage" xml:space="preserve">
-    <value>Download quality profile(s):</value>
```
duncanp-sonar, line 317: The old message appeared once before the first QP was downloaded, then replaced with a blank message. Now each QP is listed as it is being downloaded.

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -308,9 +305,6 @@ public async Task DownloadQualityProfile_WhenBindingConfigIsNull_Fails()
             result.Should().BeFalse();
             testSubject.InternalState.QualityProfiles[language].Should().Be(profile);
 
-            notifications.AssertProgress(0.0);
-            notifications.AssertProgressMessages(Strings.DownloadingQualityProfileProgressMessage);
-
```
duncanp-sonar, line 313: This test is checking what happens when a binding config is `null`. It doesn't need to check the progress messages too.

src/Integration/Resources/Strings.resx
```diff
@@ -627,4 +627,7 @@ This is the general format to use when writing errors to output window or when t
   <data name="BindingUpdateFailedToCheckSettings" xml:space="preserve">
     <value>[Binding check] Error checking bound project settings file: {0}</value>
   </data>
+  <data name="String1" xml:space="preserve">
+    <value />
+  </data>
```
duncanp-sonar, line 632: Oops, this shouldn't be here. I'll delete it.

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -256,33 +205,77 @@ public async Task DownloadQualityProfile_SavesConfiguration()
...
+            var missingPluginMessageCFamily = string.Format(Strings.SubTextPaddingFormat,
+                string.Format(Strings.CannotDownloadQualityProfileForLanguage, Language.Cpp.Name));
+            var missingPluginMessageSecrets = string.Format(Strings.SubTextPaddingFormat,
+                string.Format(Strings.CannotDownloadQualityProfileForLanguage, Language.Secrets.Name));
+
+            logger.AssertOutputStringExists(missingPluginMessageCFamily);
+            logger.AssertOutputStringExists(missingPluginMessageSecrets);
+            logger.AssertPartialOutputStringDoesNotExist(Language.Css.Name);
+
+            void CheckProgressMessages(params Language[] languages)
```
duncanp-sonar, line 261: Checking the progress messages is done in two nested methods (this one and the one in line 267).
I'm not sure the second nested method helps readability or not, since it's only one line long.

bigfluffycookie: hmm I think it is okay this way, it did take me a tiny bit too understand but not a significant amount

bigfluffycookie (approved):
Looks good :)

PR #4290, opened 2023-06-06
Cleanup - save only in new format

duncanp-sonar:
Part of #4120

Removed code relating to saving in different Connected Mode formats - we'll only save in the new format.

### DIFF
--- a/src/Integration/Binding/BindingController.cs
+++ b/src/Integration/Binding/BindingController.cs
@@ -138,31 +138,17 @@ namespace SonarLint.VisualStudio.Integration.Binding
 
             var currentConfiguration = configProvider.GetConfiguration();
 
-            SonarLintMode modeToBind;
-
             // If we are currently in standalone then the project is being bound for the first time.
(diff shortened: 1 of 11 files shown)


src/Integration/Binding/BindingController.cs
```diff
@@ -138,31 +138,17 @@ void IBindingWorkflowExecutor.BindProject(BindCommandArgs bindingArgs)
 
             var currentConfiguration = configProvider.GetConfiguration();
 
-            SonarLintMode modeToBind;
-
             // If we are currently in standalone then the project is being bound for the first time.
             // Otherwise, we are updating an existing binding
             var isFirstBinding = currentConfiguration.Mode == SonarLintMode.Standalone;
 
-            // TODO: CM cleanup - will never be saving in legacy format
```
duncanp-sonar, line 147: Essentially the change is to remove this TODO.

I've manually tested that binding/updating still work.

bigfluffycookie (approved):

PR #4291, opened 2023-06-06
(2) Drop usage of ISourceControlledFileSystem

duncanp-sonar:
- new connected mode doesn't save files in a source-controlled location

Part of #4120

### DIFF
--- a/src/Integration/MefServices/VsSessionHost.cs
+++ b/src/Integration/MefServices/VsSessionHost.cs
@@ -45,7 +45,6 @@ namespace SonarLint.VisualStudio.Integration
     {
         internal /*for testing purposes*/ static readonly Type[] SupportedLocalServices = {
                 typeof(IProjectSystemHelper),
-                typeof(ISourceControlledFileSystem),
                 typeof(IRuleSetInspector),
                 typeof(IConfigurationPersister),
                 typeof(ICredentialStoreService),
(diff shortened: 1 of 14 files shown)


src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -308,18 +307,13 @@ private void RegisterLocalServices()
             this.localServices.Add(typeof(IProjectSystemHelper), new Lazy<ILocalService>(() => new ProjectSystemHelper(this, projectToLanguageMapper)));
             this.localServices.Add(typeof(IRuleSetInspector), new Lazy<ILocalService>(() => new RuleSetInspector(this, Logger)));
 
-            // Use Lazy<object> to avoid creating instances needlessly, since the interfaces are serviced by the same instance
-            var sccFs = new Lazy<ILocalService>(() => new SourceControlledFileSystem(this, Logger));
-            this.localServices.Add(typeof(ISourceControlledFileSystem), sccFs);
```
duncanp-sonar, line 313: `ISourceControllerFileSystem` was used because we wrote files under source-controlled locations, which meant we had to interface with the VS SCC APIs. The implementation also deferred writing anything until the end i.e. it queued up file-writes until `Commit` was called, so if the workflow aborted there would be nothing to undo.

We don't need that complexity now. Files are written immediately. If we do need to cleanup at the end we can just delete the files.

bigfluffycookie (approved):
Looks good, just a minor comment

PR #4292, opened 2023-06-06
(3) Drop unused SBO.GetBindingConfig method

duncanp-sonar:
- no longer used. Also means the class no longer needs an IProjectSystemHelper

Part of #4120

### DIFF
--- a/src/Integration.UnitTests/Binding/SolutionBindingOperationTests.cs
+++ b/src/Integration.UnitTests/Binding/SolutionBindingOperationTests.cs
@@ -64,13 +64,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Binding
         #region Tests
 
         [TestMethod]
-        public void SolutionBindingOperation_ArgChecks()
-        {
-            Action act = () => new SolutionBindingOperation(null);
-            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("serviceProvider");
-        }
-
-        [TestMethod]
         public void SolutionBindingOperation_RegisterKnownRuleSets_ArgChecks()
         {
             // Arrange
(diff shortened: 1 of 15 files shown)


src/Integration.UnitTests/Binding/SolutionBindingOperationTests.cs
```diff
@@ -215,7 +170,7 @@ public void SolutionBindingOperation_CommitSolutionBinding()
 
         private SolutionBindingOperation CreateTestSubject()
         {
-            return new SolutionBindingOperation(serviceProvider, fileSystem);
```
duncanp-sonar, line 218: The `SolutionBindingOperation` only needed the service provider so it could get a `IProjectSystemHelper`. However, the method that used it is no longer called, so the constructor can be simplified.

src/Integration/Binding/BindingController.cs
```diff
@@ -147,8 +147,7 @@ void IBindingWorkflowExecutor.BindProject(BindCommandArgs bindingArgs)
                     Strings.Bind_FirstTimeBinding :
                     Strings.Bind_UpdatingNewStyleBinding);
 
-
-            var solutionBindingOp = new SolutionBindingOperation(host);
+            var solutionBindingOp = new SolutionBindingOperation();
```
duncanp-sonar, line 150: Now that the `SolutionBindingOperation` has an empty constructor, it doesn't need to be created here.
I'll move it in a separate PR.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -36,33 +35,21 @@ namespace SonarLint.VisualStudio.Integration.Binding
...
-            serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
             this.fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
-
-            this.projectSystem = serviceProvider.GetService<IProjectSystemHelper>();
-            this.projectSystem.AssertLocalServiceIsNotNull();
         }
 
         #region State
 
-        internal /*for testing purposes*/ string SolutionFullPath
```
duncanp-sonar, line 60: Not used.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -85,29 +72,18 @@ public void RegisterKnownConfigFiles(IDictionary<Language, IBindingConfig> langu
             }
         }
 
-        public IBindingConfig GetBindingConfig(Language language)
```
duncanp-sonar, line 88: Not used.

src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -85,29 +72,18 @@ public void RegisterKnownConfigFiles(IDictionary<Language, IBindingConfig> langu
...
-
         #endregion
 
         #region Public API
 
         public void Initialize()
         {
-            this.SolutionFullPath = this.projectSystem.GetCurrentActiveSolution().FullName;
+            // TODO CM cleanup
+            // no-op
```
duncanp-sonar, line 82: This method will be removed separately.

In fact, the `ISolutionBindingOperation` interface has three methods, only one of which is required. I'll continue to clean it up.

src/Integration/LocalServices/ProjectSystemHelper.cs
```diff
@@ -22,7 +22,6 @@
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Diagnostics;
-using System.Linq;
```
duncanp-sonar, line 25: Miscellaneous: unused using.

bigfluffycookie (approved):

PR #4293, opened 2023-06-06
(4) Simplify SolutionBindingOperation.Prepare

duncanp-sonar:
Part of #4120

### DIFF
--- a/src/Integration/Binding/SolutionBindingOperation.cs
+++ b/src/Integration/Binding/SolutionBindingOperation.cs
@@ -21,7 +21,6 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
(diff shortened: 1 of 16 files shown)


src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -82,24 +82,22 @@ public void Initialize()
             // no-op
         }
 
-        public void Prepare(CancellationToken token)
+        public void Prepare(IEnumerable<IBindingConfig> bindingConfigs, CancellationToken token)
```
duncanp-sonar, line 85: Simplifying this method: it doesn't need to be passed the data in a previous method. It can just be passed as a parameter in this one.

I'll remove the unnecessary methods later.

PR #4295, opened 2023-06-07
Drop unused methods from ISolutionBindingOperation

duncanp-sonar:
TODO: rename interface and remaining method

Part of #4120


src/Integration/Binding/SolutionBindingOperation.cs
```diff
@@ -81,13 +75,6 @@ public void Prepare(IEnumerable<IBindingConfig> bindingConfigs, CancellationToke
                 config.Save();
             }
         }
-
-        public bool CommitSolutionBinding()
-        {
-            // TODO - CM cleanup
-            // no-op
-            return true;
-        }
```
duncanp-sonar, line 90: All of the other changes originate from the ones in this class: the implementations are now empty
-> I deleted the empty methods-> deleted the methods from the interface -> deleted the code that called those interface methods -> fixed up the tests.

bigfluffycookie (approved):
Very nice

PR #4299, opened 2023-06-07
(2) Move binding types from Core to ConnectedMode project

duncanp-sonar:
TODO - rename namespaces

Part of https://github.com/SonarSource/sonarlint-visualstudio/issues/4120


duncanp-sonar:
I moved binding-related files from Core to ConnectedMode, along with the corresponding test files.

I only moved files that could be moved without requiring code changes.

I'll fix up the namespaces separately (if I did it in this PR it would be very hard to review).

src/ConnectedMode/ConnectedMode.csproj
```diff
@@ -16,6 +16,7 @@
     <AssemblyName>SonarLint.VisualStudio.ConnectedMode</AssemblyName>
     <RootNamespace>SonarLint.VisualStudio.ConnectedMode</RootNamespace>
     <RequiresLibGit>true</RequiresLibGit>
+    <RequiresNewtonsoft>true</RequiresNewtonsoft>
```
duncanp-sonar, line 19: This tells our import targets to import the correct reference version of the NewtonSoft NuGet package.

The solution actually builds without this explicit reference, but that's just luck (ConnectedMode depends on something else that uses the NewtonSoft assemblies, so it will be copied during the build). It's better to explicit.

PR #4301, opened 2023-06-07
(3) Move binding types from Integration to ConnectedMode project

duncanp-sonar:
Part of #4120


src/ConnectedMode.UnitTests/Binding/CSharpVBBindingConfigProviderTests.cs
```diff
@@ -125,8 +125,8 @@ public async Task GetConfig_NoSupportedActiveRules_ReturnsNull()
             result.Should().BeNull();
 
             builder.Logger.AssertOutputStrings(1);
-            var expectedOutput = string.Format(Strings.SubTextPaddingFormat,
-                string.Format(Strings.NoSonarAnalyzerActiveRulesForQualityProfile, validQualityProfile.Name, Language.VBNET.Name));
+            var expectedOutput = string.Format(BindingStrings.SubTextPaddingFormat,
```
duncanp-sonar, line 128: I added a new resource file called `BindingStrings` to the `ConnectedMode` assembly and add the required strings to it (surprisingly only two).

src/ConnectedMode/ConnectedMode.csproj
```diff
@@ -42,6 +42,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Compile Update="Binding\BindingStrings.Designer.cs">
```
duncanp-sonar, line 45: Auto-added by VS

src/ConnectedMode/InternalsVisibleTo.cs
```diff
@@ -30,3 +30,14 @@
 [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
 
 #endif
+
+
+
+// TODO - CM cleanup. Remove once we've finished moving files to ConnectedMode and fixed up scopes
```
duncanp-sonar, line 36: Some of the moved types are `internal` but still currently used by types in `Integration`.

I decided to add this `TODO` and comment for now rather than making the necessary types `public` for now.

PR #4304, opened 2023-06-07
Fix up namespace after moving types to ConnectedMode

duncanp-sonar:
- no code changes

Part of #4120


duncanp-sonar:
No code changes, just fixing up the namespace and `using` statements.

Issue #4240, opened 2023-05-24
Implement IUnintrusiveBindingController


duncanp-sonar:
### Description

New component that triggers binding to a specific project.
This will be called when migrating from one of the "old" Connected Mode settings formats.

Logically, we will need similar functionality when implement non-intrusive auto-sync of QPs/settings in the background.

TODO
-[ ] work out how to trigger the binding workflow. It should not be necessary for the Team Explorer SonarQube tab to be visible, and no UI should be shown (logging to the output window is ok). It might be necessary to refactor/change the existing binding code to enable this.
-[ ] consider whether to rename the interface

PR #4311, opened 2023-06-08
Add IBindingProcessFactory

duncanp-sonar:
Prep for #4240

### DIFF
--- a/src/Integration/Binding/BindingController.cs
+++ b/src/Integration/Binding/BindingController.cs
@@ -22,10 +22,8 @@ using System;
 using System.Diagnostics;
 using System.Linq;
 using Microsoft.VisualStudio.OLE.Interop;
-using SonarLint.VisualStudio.ConnectedMode.Binding;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
-using SonarLint.VisualStudio.Integration.Exclusions;
 using SonarLint.VisualStudio.Integration.Progress;
 using SonarLint.VisualStudio.Integration.Resources;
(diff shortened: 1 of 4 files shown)


src/Integration/Binding/BindingController.cs
```diff
@@ -148,26 +148,7 @@ void IBindingWorkflowExecutor.BindProject(BindCommandArgs bindingArgs)
                     Strings.Bind_FirstTimeBinding :
                     Strings.Bind_UpdatingNewStyleBinding);
 
-            var solutionBindingOp = new SolutionBindingOperation();
```
duncanp-sonar, line 151: Moved this ugly code out of the `BindingContoller` into a separate factory class.
There are no other product changes.

This is part of a series of changes I'm planning to make it possible to create an `IBindingProcess` independently of the Team Explorer page and the `IHost`.

The idea is that we will be able to create and execute the binding steps in the background, whether Team Explorer page exists or not.

src/Integration/Binding/IBindingProcessFactory.cs
```diff
@@ -0,0 +1,83 @@
...
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class BindingProcessFactory : IBindingProcessFactory
+    {
+        private readonly IHost host;
+        private readonly ISonarQubeService sonarQubeService;
+        private readonly IExclusionSettingsStorage exclusionSettingsStorage;
+        private readonly ILogger logger;
+
+        [ImportingConstructor]
+        public BindingProcessFactory(IHost host,
```
duncanp-sonar, line 47: Extracted the code from the `BindingController` into a simple MEF-exportable factory.

Issue #4312, opened 2023-06-08
Remove UI and IHost dependencies from the binding process


duncanp-sonar:
### Description

We can't refresh a binding in the background because the current binding code is very tightly coupled to the `Team Explorer` page and the `IHost`. We could probably live with forcing the `Team Explorer` page to be visible as part of the migration process, but we will need to be able to re-bind / update the binding silently in the background to unobtrusively update the Quality Profiles so we will need to refactor it soon anyway. Refactoring the binding code to break the dependency on the host and the UI will make the code easier to reuse and modify.

Prep for #4240 

TODO
 - [x] move the creation of `BindingProcessImpl` out of `BindingController` to a factory
 - [x] make `IConfigurationPersister` MEF-creatable (includes make some dependent objects MEF-creatable)
 - [x] remove `IHost` dependency from `BindingProcessImpl`
 - [x] remove`IConfigurationPersister` as an `IService` (should not be required. There might be other `VSHost`-cleanup we can do at the same time)
 - [x] move non-VS specific code to `ConnectedMode` project

PR #4313, opened 2023-06-08
Make IConfigurationPersister MEF-exportable

duncanp-sonar:
Part of #4312


src/Integration.UnitTests/Persistence/SolutionBindingDataWriterTests.cs
```diff
@@ -58,19 +59,11 @@ public void TestInitialize()
         }
 
         [TestMethod]
-        public void Ctor_NullFileLoader_Exception()
+        public void MefCtor_CheckIsExported()
         {
-            Action act = () => new SolutionBindingDataWriter((ISolutionBindingFileLoader)null, null);
```
duncanp-sonar, line 63: MEF-imported components never have `null`s passed in the `ImportingConstructor` so I've removed the null-args  check and tests.

PR #4315, opened 2023-06-08
Remove IHost from BindingProcessImpl

duncanp-sonar:
Part of #4312

### DIFF
--- a/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
@@ -23,7 +23,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Windows.Threading;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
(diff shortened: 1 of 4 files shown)


src/Integration/Binding/BindingProcessImpl.cs
```diff
@@ -31,48 +31,61 @@
...
         private readonly BindCommandArgs bindingArgs;
         private readonly ISolutionBindingOperation solutionBindingOperation;
         private readonly IBindingConfigProvider bindingConfigProvider;
+        private readonly ISonarQubeService sonarQubeService;
+        private readonly IConfigurationPersister configurationPersister;
         private readonly IExclusionSettingsStorage exclusionSettingsStorage;
         private readonly IEnumerable<Language> languagesToBind;
+        private readonly ILogger logger;
 
-        public BindingProcessImpl(IHost host,
```
duncanp-sonar, line 48: Previously this class used `IHost` to get hold of `ISonarQubeService`, `ILogger` and `IConfigurationPersister`.
It now imports them directly so we can drop `IHost` from this class.

For now, `IConfigurationPersister` is still exposed as an `ILocalService` via `
In the next PR I'll remove it as a service.

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -43,61 +42,53 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
     [TestClass]
     public class BindingProcessImplTests
     {
-        private ConfigurableServiceProvider serviceProvider;
-        private Mock<ISonarQubeService> sonarQubeServiceMock;
-        private ConfigurableHost host;
-        private TestLogger logger;
-
-        [TestInitialize]
-        public void TestInitialize()
```
duncanp-sonar, line 52: The tests don't need the `host` any more so I took the opportunity to get rid of the `TestInitialize`.

src/Integration.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -43,61 +42,53 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
...
+            // 7. Null logger
+            act = () => new BindingProcessImpl(bindingArgs, slnBindOp, bindingConfigProvider, exclusionSettingsStorage, configurationPersister, sonarQubeService, null);
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("logger");
         }
 
         [TestMethod]
         public async Task SaveServerExclusionsAsync_ReturnsTrue()
         {
             var bindingArgs = CreateBindCommandArgs(projectKey: "projectKey");
+            var logger = new TestLogger(logToConsole: true);
```
duncanp-sonar, line 91: Fix up existing tests - `logger` and `configurationPersister` are no longer supplied as services so they need to be created explicitly in the tests that need them.

PR #4316, opened 2023-06-08
Remove IConfigurationPersister as an ILocalService

duncanp-sonar:
Part of #4312

### DIFF
--- a/src/Integration/MefServices/VsSessionHost.cs
+++ b/src/Integration/MefServices/VsSessionHost.cs
@@ -27,13 +27,10 @@ using System.Windows.Threading;
 using Microsoft.VisualStudio.Shell;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
-using SonarLint.VisualStudio.Integration.NewConnectedMode;
-using SonarLint.VisualStudio.Integration.Persistence;
 using SonarLint.VisualStudio.Integration.ProfileConflicts;
 using SonarLint.VisualStudio.Integration.Progress;
 using SonarLint.VisualStudio.Integration.State;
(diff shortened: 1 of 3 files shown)


src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -297,25 +288,16 @@ private BindingConfiguration SafeGetBindingConfig()
 
         private void RegisterLocalServices()
         {
-            this.localServices.Add(typeof(ICredentialStoreService), new Lazy<ILocalService>(() => credentialStoreService));
```
duncanp-sonar, line 300: Miscellaneous: I removed `this` from this method since I was making changes to the method. However, it does mess the GitHub diff up somewhat :-(

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -297,25 +288,16 @@ private BindingConfiguration SafeGetBindingConfig()
...
-
-            this.localServices.Add(typeof(IConfigurationPersister), new Lazy<ILocalService>(GetConfigurationPersister));
```
duncanp-sonar, line 302: This is the only functional change in the method - deleting the service registration.

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -81,7 +75,6 @@ internal sealed class VsSessionHost : IHost, IProgressStepRunnerWrapper, IDispos
                 credentialStoreService,
                 projectToLanguageMapper,
                 configurationProvider,
-                configFilePathProvider,
```
duncanp-sonar, line 84: The `configFilePathProvider` was only being imported because it was needed to construct the configuration persister -> removed.

PR #4317, opened 2023-06-08
Remove UI and IHost dependencies from the binding process

### LINKED ISSUE #4312: Remove UI and IHost dependencies from the binding process

duncanp-sonar:
Fixes #4312


duncanp-sonar:
Talked through the changes with @bigfluffycookie offline.

Issue #4320, opened 2023-06-09
Move CM types from Integration -> CM


duncanp-sonar:
### Description

The `ConnectedMode` project was added relatively recently, so all of the pre-existing Connected Mode-related classes are in the `Integration` project.

This complicates making changes to Connected Mode as new types that should be in the `ConnectedMode` need to reference existing types that are currently in `Integration`.

TODO
 - [x] add reference to ALM binary to `ConnectedMode`
 - [x] move persistence types and tests to `ConnectedMode`
 - [x] fix up namespaces

PR #4322, opened 2023-06-09
Change ICredentialStoreService to not be an ILocalService

duncanp-sonar:
- can now only be created using MEF
- prep for moving the implementation from Integration to ConnectedMode

Part of #4320

### DIFF
--- a/src/Integration/Connection/ConnectionWorkflow.cs
+++ b/src/Integration/Connection/ConnectionWorkflow.cs
@@ -55,7 +55,7 @@ namespace SonarLint.VisualStudio.Integration.Connection
             this.host = host ?? throw new ArgumentNullException(nameof(host));
             this.parentCommand = parentCommand ?? throw new ArgumentNullException(nameof(parentCommand));
 
-            this.credentialStore = this.host.GetService<ICredentialStoreService>();
+            this.credentialStore = this.host.GetMefService<ICredentialStoreService>();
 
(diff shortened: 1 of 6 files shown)


src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -55,7 +55,7 @@ public ConnectionWorkflow(IHost host, ICommand parentCommand)
             this.host = host ?? throw new ArgumentNullException(nameof(host));
             this.parentCommand = parentCommand ?? throw new ArgumentNullException(nameof(parentCommand));
 
-            this.credentialStore = this.host.GetService<ICredentialStoreService>();
```
duncanp-sonar, line 58: `ICredentialStoreService` was already MEF-creatable, and was MEF-created by most consumers. However, it was still exposed as an `ILocalService`.

The ConnectionWorkflow was the only consumer that still fetched it using the service provider mechanism. The consumer is easy to fix up - just request the service as a MEF service instead.

src/Integration/LocalServices/ICredentialStoreService.cs
```diff
@@ -22,7 +22,7 @@
 
 namespace SonarLint.VisualStudio.Integration
 {
-    public interface ICredentialStoreService : ILocalService
```
duncanp-sonar, line 25: I considered dropping the `Service` suffix from the interface name since it is no longer a local service, but that would have created a lot of renames. I might do it later in a separate PR.

PR #4323, opened 2023-06-09
Move persistence types from Integration -> ConnectedMode

duncanp-sonar:
TODO - fix namespaces

Part of #4320


duncanp-sonar:
Moved types: no change changes.

The existing tests pass and I manually tested that binding still works.
The namespaces will be fixed up in a separate commit.

PR #4324, opened 2023-06-09
Fix up namespaces of types move to ConnectedMode

duncanp-sonar:
Part of #4320


duncanp-sonar:
Fixed up namespaces. No code changes.

At this point, almost all of the non-VS specific binding types are now in the `ConnectedMode` assembly (although I missed moving the `CredentialStore` implementation, so I'll do that next).

Most of the binding types left in `Integration` are related to the Connection and Binding workflows and displaying progress in the UI.

PR #4327, opened 2023-06-09
Move persistence types from Integration -> ConnectedMode

duncanp-sonar:
TODO - fix namespaces

Part of #4320


duncanp-sonar:
Failing VS2019 test is a flaky test that passed on an earlier run -> not blocking.

PR #4332, opened 2023-06-12
Move targets types to separate "Install" folder

### LINKED ISSUE #4331: Move targets-related types to separate folder
### Description

Currently the targets types are in the same folder as a lot of the binding types. This makes it hard to tell which types are related to binding and which deal with the installation of the targets.

Logically they are separate features. Moving the targets types in a separate folder would make it easier to tell which types relate to which feature.

duncanp-sonar:
Fixes #4331


src/ConnectedMode.UnitTests/Install/ImportsBeforeFileGeneratorTests.cs
```diff
@@ -0,0 +1,163 @@
+﻿/*
```
duncanp-sonar, line 1: I don't know why GitHub is showing this a new set of tests; I just moved it and changed the namespace, exactly the same as the other test file i.e. no code changes.

src/ConnectedMode/ConnectedMode.csproj
```diff
@@ -34,8 +34,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <None Remove="Embedded\SonarLintTargets.xml" />
-    <EmbeddedResource Include="Embedded\SonarLintTargets.xml" />
+    <None Remove="Install\SonarLintTargets.xml" />
+    <EmbeddedResource Include="Install\SonarLintTargets.xml" />
```
duncanp-sonar, line 38: I moved the resource file so all of the "install" related files are in the same folder.

src/ConnectedMode/Install/ImportBeforeInstallTrigger.cs
```diff
@@ -78,8 +78,8 @@ public void Dispose()
         {
             if (!disposed)
             {
-                this.activeSolutionBoundTracker.PreSolutionBindingChanged -= OnPreSolutionBindingChanged;
-                this.activeSolutionBoundTracker.PreSolutionBindingUpdated -= OnPreSolutionBindingUpdated;
+                activeSolutionBoundTracker.PreSolutionBindingChanged -= OnPreSolutionBindingChanged;
```
duncanp-sonar, line 81: I didn't explicitly make these changes, so presumably VS did it automatically.

I'll check my VS settings if it happens again. We don't really want a whole load of unplanned cosmetic changes being applied automatically and obscuring the real changes in a PR.

duncanp-sonar:
Now that I've moved all of the binding types from `Integration` to `ConnectedMode`, the targets-related types are hard to see as they are in the same folder.

So, I moved the target-installation types to a new folder called `Install`. I think I've moved them all - let me know if I missed any.

No code changes, just namespaces.

bigfluffycookie (approved):

PR #4333, opened 2023-06-12
Add LocalHotspotStore for locally analyzed hotspots

georgii-borovinskikh-sonarsource:


src/Integration.Vsix/Analysis/IssueConsumerFactory.cs
```diff
@@ -51,17 +52,19 @@ internal partial class IssueConsumerFactory : IIssueConsumerFactory
     {
         private readonly ISuppressedIssueMatcher suppressedIssueMatcher;
         private readonly IAnalysisIssueVisualizationConverter converter;
+        private readonly ILocalHotspotsStoreUpdater localHotspotsStore;
 
         [ImportingConstructor]
-        internal IssueConsumerFactory(ISuppressedIssueMatcher suppressedIssueMatcher, IAnalysisIssueVisualizationConverter converter)
+        internal IssueConsumerFactory(ISuppressedIssueMatcher suppressedIssueMatcher, IAnalysisIssueVisualizationConverter converter, ILocalHotspotsStoreUpdater localHotspotsStore)
```
duncanp-sonar, line 58: Changing the issue consumer to use the new store is logically separate and should have been a separate ticket and done in a separate PR. It's not worth separating the changes out for this PR, but going forwards please create separate tickets for each logically separate tasks.

src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs
```diff
@@ -76,9 +82,20 @@ internal class IssueHandler
                 // The text buffer might have changed since the analysis was triggered, so translate
                 // all issues to the current snapshot.
                 // See bug #1487: https://github.com/SonarSource/sonarlint-visualstudio/issues/1487
-                var translatedIssues = translateSpans(issues, textDocument.TextBuffer.CurrentSnapshot);
-
-                var newSnapshot = new IssuesSnapshot(projectName, projectGuid, textDocument.FilePath, translatedIssues);
+                var translatedIssues = translateSpans(issues, textDocument.TextBuffer.CurrentSnapshot).ToList();
```
duncanp-sonar, line 85: Please add review comments to help reviewers understand why changes have been made - what seems obvious to you as the person who made the change might not be obvious to anyone else reading it.

Here, I'm guessing you added the `ToList()` because you are filtering the collection twice before. Using `ToArray()` would be cheaper as it's a lighter-weight data structure.

However, is it required? Doesn't `DoTranslateSpans` return an array already?

src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs
```diff
@@ -76,9 +82,20 @@ internal class IssueHandler
...
+                    translatedIssues
+                        .Where(issue => (issue.Issue as IAnalysisIssue)?.Type == AnalysisIssueType.SecurityHotspot));
+                
+                var newSnapshot = new IssuesSnapshot(projectName,
+                    projectGuid,
+                    textDocument.FilePath,
+                    translatedIssues.Where(issue =>
+                    {
+                        var analysisIssueType = (issue.Issue as IAnalysisIssue)?.Type;
+                        return analysisIssueType == AnalysisIssueType.Bug || analysisIssueType == AnalysisIssueType.CodeSmell;
```
duncanp-sonar, line 97: This would be clearer and shorter checking for `!= AnalysisIssueType.SecurityHotspot`.
That would make it clearer that there aren't any other issues types that we are deliberately filtering out.

src/Integration.Vsix.UnitTests/Analysis/IssueHandlerTests.cs
```diff
@@ -112,33 +122,44 @@ public void HandleNewIssues_IssuesGetMatchesIsCalled()
...
             {
-                CreateIssue("yyy", startLine: 2, endLine: 2)
+                CreateIssue("yyy", startLine: 2, endLine: 2),
+                CreateIssue("yyyy", startLine: 4, endLine: 4, true)
             };
 
             var notificationHandler = new SnapshotChangeHandler();
 
-            var textDocument = CreateValidTextDocument("any.txt");
+            var filePath = "any.txt";
```
duncanp-sonar, line 141: This could be a constant. Is there a warning/info in the IDE about it?

src/IssueViz.Security/Hotspots/LocalHotspot.cs
```diff
@@ -0,0 +1,37 @@
...
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Hotspots
+{
+    internal class LocalHotspot
+    {
+        public LocalHotspot(IAnalysisIssueVisualization visualization, SonarQubeHotspot serverHotspot = null)
+        {
+            Visualization = visualization;
+            ServerHotspot = serverHotspot;
```
duncanp-sonar, line 31: If I remember correctly, the `Visualization` should never be null but the `ServerHotspot` could be? If so, it's not obvious from the code.

Assuming that's the expected behaviour, I'd `throw` in the constructor if the visualization is null and add a comment explaining why the server hotspot could be null.

src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs
georgii-borovinskikh-sonarsource: Here ToList is because we enumerate twitce afterwards: one for hotspots, one for issues

src/Integration.Vsix.UnitTests/Analysis/IssueHandlerTests.cs
```diff
@@ -42,9 +44,14 @@ public class IssueHandlerTests
         [TestMethod]
         public void HandleNewIssues_UpdatedSnapshotHasExpectedValues()
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Yep, missed that

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -0,0 +1,208 @@
...
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarLint.VisualStudio.IssueVisualization.Security.IssuesStore;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Hotspots
+{
+    public interface ILocalHotspotsStoreUpdater
+    {
+        void AddOrUpdate(string filePath, IEnumerable<IAnalysisIssueVisualization> hotspots);
```
duncanp-sonar, line 34: We should always have doc comments on interfaces.

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -0,0 +1,208 @@
...
+        void RemoveForFile(string filePath);
+    }
+
+    internal interface ILocalHotspotsStore : ILocalHotspotsStoreUpdater, IIssuesStore
+    {
+        List<LocalHotspot> GetAllLocalHotspots();
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I've actually been following the opposite rule: accept abstract types where possible, return concrete types where possible. 

The idea with list here is that it's a copy each time (the elements are immutable), so that the list is not modified by the source after it's returned

src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs
georgii-borovinskikh-sonarsource: But yes, I missed that translatespans already returns an array. Will fix the method

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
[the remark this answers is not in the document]
duncanp-sonar: That's not the approach we follow anywhere else.

Effectively what you're saying is "because we're returning a copy we don't care if you change the list", which isn't quite the same thing as return a type that conveys "here is a read-only list".

We can still return a copy each time if we want to, but it's an implementation detail.

src/IssueViz.Security/Hotspots/IHotspotMatcher.cs
```diff
@@ -0,0 +1,30 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.IssueVisualization.Models;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Hotspots
+{
+    internal interface IHotspotMatcher
```
duncanp-sonar, line 26: Is there a no-op MEF-exported implementation of this interface? If not, anything that depends on it will fail at runtime. Since this now includes the `IssueConsumerFactory` it presumably means that analysis won't work.

This would probably be pretty confusing/frustrating for anyone else syncing the branch while working on the sprint.

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -0,0 +1,214 @@
...
+        private List<SonarQubeHotspot> serverHotspots = new List<SonarQubeHotspot>();
+        private HashSet<SonarQubeHotspot> unmatchedHotspots = new HashSet<SonarQubeHotspot>();
+
+        [ImportingConstructor]
+        public LocalHotspotsStore(IServerHotspotStore serverHotspotStore, IHotspotMatcher hotspotMatcher, IThreadHandling threadHandling)
+        {
+            this.serverHotspotStore = serverHotspotStore;
+            this.hotspotMatcher = hotspotMatcher;
+            this.threadHandling = threadHandling;
+            InitializeServerHotspots(this.serverHotspotStore.GetAll());
```
duncanp-sonar, line 73: This is problematic, based on what @bigfluffycookie pointed out this morning - the VS threading cookbook [says](https://github.com/microsoft/vs-threading/blob/main/doc/cookbook_vs.md#how-do-i-effectively-verify-that-my-code-is-fully-free-threaded) importing constructors should be free-threaded i.e. able to run completely on the caller's thread
-> we might be on the UI thread
-> we shouldn't be making server calls

Since importing constructors also have to be transitively free-thread there won't be a lot we can do. We'll need to come up with a policy on what do in importing constructors. The simplest policy would be to do nothing except register event handlers. 

For now, it might be easiest to just drop the initial refresh and revisit how we do it later.
There are a couple of options: one is just relying on the timed update to kick in at some point; another would be to move the responsibility to the package (i.e. have the package explicitly call for a refresh/trigger the update).

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -0,0 +1,214 @@
...
+                    oldIssueVisualizations = oldHotspots
+                        .Select(x => x.Visualization)
+                        .ToList();
+                }
+                else
+                {
+                    oldIssueVisualizations = EmptyList;
+                }
+
+                var hotspotsList = hotspots.ToList();
```
duncanp-sonar, line 119: Why are you making a copy here? Because you don't know if the enumeration is lazy or not? (if so, does it matter?)

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -0,0 +1,214 @@
...
+                
+                if (!hotspotsList.Any() && !oldIssueVisualizations.Any())
+                {
+                    return;
```
duncanp-sonar, line 123: `Any` should work efficiently with lazy enumerations, so you could check `hotspots` before making a copy of the list i.e. avoid making a copy unless you need to.

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
duncanp-sonar: Actually, what does `serverHotspotStore.GetAll()` do? Just it just fetch the issues currently known to the server hotspot store? If so it might be ok to call in in the constructor.

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -0,0 +1,214 @@
...
+                
+                fileToHotspotsMapping.Remove(filePath);
+                UnmatchServerHotspots(localHotspots);
+
+                NotifyIssuesChanged(new IssuesChangedEventArgs(localHotspots.Select(x => x.Visualization).ToList(),
+                    EmptyList));
+            }
+        }
+
+        private List<LocalHotspot> GetLocalHotspots(IEnumerable<IAnalysisIssueVisualization> hotspots)
```
duncanp-sonar, line 151: This is more like `CreateLocalHotpots`, isn't it?

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
```diff
@@ -0,0 +1,343 @@
...
+using SonarLint.VisualStudio.TestInfrastructure;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Hotspots;
+
+[TestClass]
+public class LocalHotspotStoreTests
+{
+    [TestMethod]
+    public void MefCtor_CheckExports_ILocalHotspotsStore()
```
duncanp-sonar, line 40: FYI there are new test helper methods in `MefTestHelpers` to check whether the type is exported as a singleton or not. Going forwards, we should routinely test that as well as the "can be exported".
I'd leave it for now, though, since the code isn't in the hotspots branch. We can go through the existing tests as some point and fix them up.

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
```diff
@@ -0,0 +1,343 @@
...
+        matcherMock.Setup(x => x.IsMatch(issueVis1, serverHotspot1)).Returns(true);
+        matcherMock.Setup(x => x.IsMatch(issueVis2, serverHotspot2)).Returns(true);
+        matcherMock.Setup(x => x.IsMatch(issueVis3, serverHotspot1))
+            .Returns(true); // 2 local hotspots match 1 server but only one pair should be created
+
+        var testSubject = CreateTestSubject(out _, serverStoreMock.Object, matcherMock.Object);
+
+        testSubject.AddOrUpdate("lalala", new[] { issueVis1, issueVis2, issueVis3 });
+
+        VerifyContent(testSubject, new LocalHotspot(issueVis1, serverHotspot1),
```
duncanp-sonar, line 151: Minor: listing the expected values on separate lines would make them easier to read and compare to the code above.

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
```diff
@@ -0,0 +1,343 @@
...
+        matcherMock.Setup(x => x.IsMatch(issueVis1, serverHotspot1)).Returns(true);
+        matcherMock.Setup(x => x.IsMatch(issueVis2, serverHotspot2)).Returns(true);
+        matcherMock.Setup(x => x.IsMatch(issueVis3, serverHotspot1)).Returns(true);
+        matcherMock.Setup(x => x.IsMatch(issueVis4, serverHotspot2)).Returns(true);
+
+        var testSubject = CreateTestSubject(out _, serverStoreMock.Object, matcherMock.Object);
+
+        testSubject.AddOrUpdate("lalala", new[] { issueVis1, issueVis2 });
+        testSubject.AddOrUpdate("lalala", new[] { issueVis3 });
+        testSubject.AddOrUpdate("lalala", new[] { issueVis3, issueVis4 });
```
duncanp-sonar, line 178: It's hard to follow what this is testing and why the final result is the expected one.
The test is making three state changes to the store but only checking the final one. It would be clearer and more robust if it validate the contents after each event.

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: OK, I see what you mean

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
georgii-borovinskikh-sonarsource: Yeah it should just look up the local value and not fetch from the server

duncanp-sonar (changes requested):
See comments/questions inline.

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -0,0 +1,214 @@
...
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Hotspots
+{
+    /// <summary>
+    /// Methods for modification of <see cref="ILocalHotspotsStore"/>
+    /// </summary>
+    public interface ILocalHotspotsStoreUpdater
+    {
+        void AddOrUpdate(string filePath, IEnumerable<IAnalysisIssueVisualization> hotspots);
```
duncanp-sonar, line 37: It's not clear from the method signature whether the add/update is happening at file-level or hotspot-level e.g. if I make the following two calls:

```
  updater.AddOrUpdate("file1.txt", new [] { hotspot1 });
  updater.AddOrUpdate("file1.txt", new [] { hotspot2 });
```

Does the second call completely replace the list of hotspots for the file, or does it add to it?

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -0,0 +1,214 @@
...
+                {
+                    oldIssueVisualizations = EmptyList;
+                }
+                
+                if (!hotspots.Any() && !oldIssueVisualizations.Any())
+                {
+                    return;
+                }
+                
+                var hotspotsList = hotspots.ToList();
```
duncanp-sonar, line 124: Do we need to copy the list? We only loop through it once.

duncanp-sonar (approved):
Looks good.

One minor comment about the `AddOrUpdate` being ambiguous.

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
georgii-borovinskikh-sonarsource: Maybe rename to UpdateHotspotsForFile?

duncanp-sonar: Yep, that's better.

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
georgii-borovinskikh-sonarsource: Yes, the event args below need IReadOnlyCollection

Issue #4335, opened 2023-06-12
Create skeleton of MigrationWizardController


bigfluffycookie:
Create a skeleton MigrationWizardController which eventually will be responsible for wiring up the WizardWindow and BindingController.

PR #4338, opened 2023-06-12
Add migration wizard controller

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4335


src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    /// <summary>
+    /// Starts a wizard which guides a user through the migration progress.
+    /// </summary>
+    public interface IMigrationWizardController
+    {
+        /// <summary>
+        /// Raised when the wizard is closed after successfully finishing the migration process.
+        /// </summary>
+        event EventHandler MigrationWizardFinished;
```
bigfluffycookie, line 38: Successfully refers to the wizard not being cancelled but finished.

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        void StartMigrationWizard();
+    }
+
+    [Export(typeof(IMigrationPrompt))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class MigrationWizardController : IMigrationWizardController
```
bigfluffycookie, line 45: I got a bit confused at some point about the life cycle of this class. Do we want it to be a singleton? How long does it live? Always? Whenever a solution is closed / opened?

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
```diff
@@ -0,0 +1,52 @@
...
+        void StartMigrationWizard();
+    }
+
+    [Export(typeof(IMigrationWizardController))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class MigrationWizardController : IMigrationWizardController
+    {
+        public event EventHandler MigrationWizardFinished;
+
+        public void StartMigrationWizard() { }
```
bigfluffycookie, line 45: Forgot to add a // TODO comment. Will do it with other changes.

src/ConnectedMode.UnitTests/Migration/MigrationPromptTests.cs
```diff
@@ -91,17 +102,70 @@ public async Task ShowAsync_ShowNotificationIsCalledOnMainThread()
...
+            notificationService.Verify(x => x.RemoveNotification(), Times.Once);
+        }
+
         [TestMethod]
         public void Clear_RemoveNotificationIsCalled()
         {
             var notificationService = new Mock<INotificationService>();
 
-            var testSubject = new MigrationPrompt(Mock.Of<IServiceProvider>(), notificationService.Object, new NoOpThreadHandler()); ;
+            var testSubject = new MigrationPrompt(Mock.Of<IServiceProvider>(), notificationService.Object, Mock.Of<IMigrationWizardController>(), new NoOpThreadHandler()); ;
```
bigfluffycookie, line 148: Will remove double semicolon with other changes

src/ConnectedMode/Migration/IMigrationPrompt.cs
```diff
@@ -31,7 +32,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
     /// <summary>
     /// In charge of showing users a prompt to migrate to a new connected mode.
     /// </summary>
-    public interface IMigrationPrompt
+    internal interface IMigrationPrompt : IDisposable
```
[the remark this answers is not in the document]
duncanp-sonar: We do it elsewhere in the product code, so yes.

FYI even if the interface does not implement `IDisposable` directly, the calling code could still safely call it as follows:
```
  (foo as IDisposable)?.Dispose();
```

... which will no-op if `foo` does not implement `IDisposable`.

src/ConnectedMode/Migration/IMigrationPrompt.cs
```diff
@@ -79,14 +88,16 @@ public async Task ShowAsync()
...
+        
         public void Clear()
         {
             notificationService.RemoveNotification();
         }
 
         private void OnMigrate()
         {
-            // TODO: Show migration wizard
+            migrationWizardController.StartMigrationWizard();
```
duncanp-sonar, line 100: With this implementation, the gold bar will remain visible until the migration wizard has finished.

Would it make much difference from a UX point of view if we removed it immediately i.e. as soon as the wizard has started? That would make the code simpler because we wouldn't need to listen for the "finished" event.

src/ConnectedMode/Migration/IMigrationPrompt.cs
```diff
@@ -101,5 +112,10 @@ private string GetSolutionPath()
 
             return fullSolutionName as string;
         }
+
+        public void Dispose()
+        {
+            migrationWizardController.MigrationWizardFinished -= OnMigrationWIzardFinished;
+        }
```
duncanp-sonar, line 119: It seems a bit odd to have separate `Clear` and `Dispose` methods (or perhaps it seems odd that `Dispose` doesn't call `Clear`).

If this class is `Shared` it won't be disposed until VS shuts down, but it's still neater to explicitly unregister the event handler.

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
duncanp-sonar: Ah: so if the wizard is cancelled, the event is not raised -> the gold bar will still be visible?

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
duncanp-sonar: Hhmmm. Good question. We only want one instance to be executing at a time, and we don't need to create it at all unless and until the user clicks on the `Migrate` button. We have two options:

* leave it as a singleton. If the user never clicks on `Migrate` we'll have created it unnecessarily, but that isn't a big problem. We'd also be relying on the fact that the migration uses a modal dialogue, which would prevent the user from triggering a second migration while the first one is active. If the user migrates several solutions one after another, they would all use the same instance.

* create it when needed i.e. create a new instance for each migration, and dispose of it when the migration finishes. That feels neater in some ways, but we'd need to make some changes to make it happen e.g. we'd need to create single factory to create the necessary migration objects when needed. Let's discuss it offline and work out what changes we would need to make.

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
bigfluffycookie: Yes :)

src/ConnectedMode/Migration/IMigrationPrompt.cs
bigfluffycookie: But what if the user closes the migration window before actively starting the migration process?

src/ConnectedMode.UnitTests/Migration/Wizard/MigrationWizardControllerTests.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration.Wizard
+{
+    [TestClass]
+    public class MigrationPromptTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<MigrationWizardController, IMigrationWizardController>();
+        }
```
duncanp-sonar, line 33: Could add a check for the `PartCreationPolicy`. We should probably do that as standard now when adding new types.

src/ConnectedMode/Migration/IMigrationPrompt.cs
duncanp-sonar: Ok: so the UX would be:

* solution opens, migration is required -> gold bar shown.

**Scenario 1 :** user closes the gold bar without clicking on any buttons.
The migration check will happen again next time the user opens the solution.

**Scenario 2:** users clicks on `Migrate` -> wizard shown. User clicks `Cancel`.
Gold bar is still visible. User can trigger the wizard again or close the gold bar.

**Scenario 3:** users clicks on `Migrate` -> wizard shown. User clicks `Start` but migration is not successful.
Same as (2) i.e. gold bar is still visible. User can trigger the wizard again or close the gold bar.

**Scenario 4:** users clicks on `Migrate` -> wizard shown. User clicks `Start`, migration is successful.
Gold bar is closed when the wizard closes.

Is that right? If so, seems ok to me.

src/ConnectedMode/Migration/IMigrationPrompt.cs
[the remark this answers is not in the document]
bigfluffycookie: But how could you not implement dispose? you have to no?

src/ConnectedMode/Migration/IMigrationPrompt.cs
bigfluffycookie: Yes, that is correct. The only problem would be Scenario 3 (depending on how far the migration got) but I think we need to deal with that seperately anyways.

src/ConnectedMode/Migration/IMigrationPrompt.cs
bigfluffycookie: Considering we will change this to be NonShared Is it okay that for now I just added Clear into the Dispose()?

duncanp-sonar:
> Also, I dont know what to do about the coverage. The methods are mostly empty so I cannot really test them

You could just raise the `Finished` event in `StartMigrationWizard`, so it starts and immediately signals that it has finished.
However, I wouldn't worry about it; the implementation is a work in progress so we can make up the coverage later. I wouldn't let it block the PR at this point.

duncanp-sonar (approved):

PR #4340, opened 2023-06-13
Add skeleton project cleaners

### LINKED ISSUE #4339: Create skeleton project cleaners
### Description

TODO
 - [ ] create skeleton IProjectCleaner implementations for rulesets and additional files
 - [ ] modify the migration class to import many cleaners
 - [ ] unit tests

duncanp-sonar:
Fixes #4339

### DIFF
--- a/src/ConnectedMode/Migration/ConnectedModeMigration.cs
+++ b/src/ConnectedMode/Migration/ConnectedModeMigration.cs
@@ -1,0 +1,73 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using EnvDTE;
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Migration
+{
+    [Export(typeof(IConnectedModeMigration))]
+    [PartCreationPolicy(CreationPolicy.NonShared)]
+    internal class ConnectedModeMigration : IConnectedModeMigration
+    {
+        private readonly ILogger logger;
+        private readonly IProjectCleaner[] projectCleaners;
+
+        internal /* for testing */ IEnumerable<IProjectCleaner> ProjectCleaners => projectCleaners?.ToList() ?? Enumerable.Empty<IProjectCleaner>();
+
+        [ImportingConstructor]
+        public ConnectedModeMigration(ILogger logger, [ImportMany]IProjectCleaner[] projectCleaners)
+        {
+            this.logger = logger;
+            this.projectCleaners = projectCleaners;
+        }
+
+        public async Task MigrateAsync(IProgress<MigrationProgress> progress, CancellationToken token)
+        {
+            logger.WriteLine(MigrationStrings.Starting);
+
+            // TODO: cleanup - dummy implementation to provide feedback for the UI
+            progress?.Report(new MigrationProgress(1, 2, "TODO 1", false));
+            progress?.Report(new MigrationProgress(2, 2, "TODO 2", true));
+
+            foreach (var project in GetProjects())
+            {
+                foreach (var cleaner in projectCleaners)
+                {
+                    await cleaner.CleanAsync(project, progress, token);
+                }
+            }
+
+            logger.WriteLine(MigrationStrings.Finished);
+        }
+
+        private IEnumerable<Project> GetProjects()
+        {
+            // TODO
+            return Array.Empty<Project>();
+        }
+    }
+}
--- a/src/ConnectedMode/Migration/IConnectedModeMigration.cs
+++ b/src/ConnectedMode/Migration/IConnectedModeMigration.cs
@@ -19,10 +19,9 @@
  */
 
 using System;
-using System.ComponentModel.Composition;
 using System.Threading;
 using System.Threading.Tasks;
-using SonarLint.VisualStudio.Integration;
+using EnvDTE;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Migration
 {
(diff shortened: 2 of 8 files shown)


src/ConnectedMode/Migration/IConnectedModeMigration.cs
```diff
@@ -19,10 +19,9 @@
  */
 
 using System;
-using System.ComponentModel.Composition;
 using System.Threading;
 using System.Threading.Tasks;
-using SonarLint.VisualStudio.Integration;
+using EnvDTE;
```
duncanp-sonar, line 24: The GitHub side-by-side diff looks a bit weird for this file, so you might want to ignore the "before".

I moved the "cleaner" and progress interfaces to this single file so we can see all of the related interface types together.

I moved the implementation of `ConnectedModeMigration` to a separate class. 

I think it's easier to see the bigger picture this way, but if not we can separate them out to individual files again.

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using EnvDTE;
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Migration
+{
+    [Export(typeof(IConnectedModeMigration))]
+    [PartCreationPolicy(CreationPolicy.NonShared)]
+    internal class ConnectedModeMigration : IConnectedModeMigration
```
duncanp-sonar, line 34: I moved this class to a separate file and modified it to import many `IProjectCleaner`s.

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -0,0 +1,73 @@
...
+            this.projectCleaners = projectCleaners;
+        }
+
+        public async Task MigrateAsync(IProgress<MigrationProgress> progress, CancellationToken token)
+        {
+            logger.WriteLine(MigrationStrings.Starting);
+
+            // TODO: cleanup - dummy implementation to provide feedback for the UI
+            progress?.Report(new MigrationProgress(1, 2, "TODO 1", false));
+            progress?.Report(new MigrationProgress(2, 2, "TODO 2", true));
```
duncanp-sonar, line 54: I left the dummy progress calls behind for now to make testing the UI easier. We can remove them later.

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -0,0 +1,73 @@
...
+                    await cleaner.CleanAsync(project, progress, token);
+                }
+            }
+
+            logger.WriteLine(MigrationStrings.Finished);
+        }
+
+        private IEnumerable<Project> GetProjects()
+        {
+            // TODO
```
duncanp-sonar, line 69: I'll create a separate ticket for this TODO.

src/ConnectedMode/Migration/RuleSetCleaner.cs
```diff
@@ -18,32 +18,36 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
```
duncanp-sonar, line 21: The GitHub side-by-side diff looks a bit weird for this file, so you might want to ignore the "before".

I moved the `MigrationProgress` class to another file and added this new `RuleSetCleaner` class. For some reason, GitHub is showing the unrelated new file as a modification of the deleted file.

src/ConnectedMode/Migration/RuleSetCleaner.cs
```diff
@@ -18,32 +18,36 @@
...
+    [Export(typeof(IProjectCleaner))]
+    [PartCreationPolicy(CreationPolicy.NonShared)]
     /// <summary>
-    /// Data class containing information about migration progress
+    /// Attempts to remove references to the generated ruleset from the project
     /// </summary>
-    internal class MigrationProgress
+    /// <remarks>NB this should only be called for C# or VB.NET projects - other project types
+    /// won't have references to a generated ruleset</remarks>
+    internal class RuleSetCleaner : IProjectCleaner
```
duncanp-sonar, line 37: Empty implementation, but has the required MEF exports and tests.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -54,6 +54,44 @@ public static Export CreateExport<T>(object exportInstance, string contractName
             return export;
         }
 
+        /// <summary>
+        /// Overload that imports using a SingleObjectImporter and checks that the expected
+        /// implementing type was imported, and that the expected CreationPolicy has been applied
+        /// </summary>
+        public static SingleObjectImporter<TImportType> CheckTypeCanBeImported<TTypeToCheck, TImportType>(
+            CreationPolicy expectedCreationPolicy,
```
duncanp-sonar, line 62: Added `expectedCreationPolicy` as a parameter so we can check whether the type is a singleton or not.

The way this method is written means none of the existing exports are being checked. We could assume that most of them are singletons, and tweak these test helpers so that if we don't specify a creation policy we check for a singleton. However, that would probably caused some tests failures that would required fixing up, so I'll do that separately.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -54,6 +54,44 @@ public static Export CreateExport<T>(object exportInstance, string contractName
...
+            params Export[] requiredExports)
+            where TImportType : class
+            where TTypeToCheck : class
+        {
+            if (expectedCreationPolicy == CreationPolicy.Any)
+            {
+                throw new ArgumentOutOfRangeException("Test setup error: CreationPolicy.Any is not supported (we don't use it anywhere currently)");
```
duncanp-sonar, line 69: We don't use it, so it was simpler just to make the method throw.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -54,6 +54,44 @@ public static Export CreateExport<T>(object exportInstance, string contractName
...
+            var importer1 = new SingleObjectImporter<TImportType>();
+            Compose(importer1, typeof(TTypeToCheck), requiredExports);
+            importer1.AssertImportIsNotNull();
+            importer1.AssertImportIsInstanceOf<TTypeToCheck>();
+
+            // Now import the type again, doing the full check that the requiredExports
+            // are actually required
+            var importer2 = CheckTypeCanBeImported<TTypeToCheck, TImportType>(requiredExports);
+
+            // Check the expected creation policy was used
```
duncanp-sonar, line 82: Checking the creation policy is simple - import the type twice, and see if we get the same instance or not.

georgii-borovinskikh-sonarsource (approved):

src/TestInfrastructure/Helpers/MefTestHelpers.cs
bigfluffycookie: Actually, Why do you not reuse the code we already have in this file? CheckIsNonSharedMefComponent() and CheckIsSingletonMefComponent() ?

bigfluffycookie (approved):
Looks good but one small point about the MefTestHelpers

src/TestInfrastructure/Helpers/MefTestHelpers.cs
duncanp-sonar: Good point. I remembered you'd added those recently, then forgot to go and look at them.
I'll delete my new test method since it is more complicated and use your existing helpers.

Issue #4344, opened 2023-06-13
Implement C#/VB.NET project walker


duncanp-sonar:
### Description

When migrating, it would be useful to identify the set of  C#/VB.NET projects in the solution, if any.
We do not need to visit any other project types, or any other "special" Solution Explorer folders.

Options:
1) use the existing-but-ugly `IProjectHelper` methods to walk the `SolutionExplorer` hierarchy and filter out the projects/folders we are not interested in, or 
2) walk the Roslyn workspace, which will only return the projects we care about.

Option(2) seems like a simpler option. In addition, having the Roslyn `Project` objects will allow us to check whether the project references any additional files or not.

TODO
- [x] add Roslyn project walker
- [x] add tests

PR #4345, opened 2023-06-13
Move LanguageServices package ref to import file

duncanp-sonar:
Prep for #4344

### DIFF
--- a/src/Integration.Vsix/Integration.Vsix.csproj
+++ b/src/Integration.Vsix/Integration.Vsix.csproj
@@ -76,6 +76,7 @@
     <RequiresSystemIOAbstractions>true</RequiresSystemIOAbstractions>
     <RequiresLibGit>true</RequiresLibGit>
     <RequiresChannels>true</RequiresChannels>
+    <RequiresLanguageServices>true</RequiresLanguageServices>
   </PropertyGroup>
   
   <!-- This project references the SonarLint TeamExplorer assembly which is VS-version specific -->
@@ -148,9 +149,6 @@
          files means it's simpler to always include it so the contents of the lock file don't change for
          signed builds. -->
     <PackageReference Include="Microsoft.VSSDK.Vsixsigntool" Version="16.2.29116.78" PrivateAssets="all" />
-
-    <!-- This package references Roslyn components used by VS. It isn't part of the VSSDK meta-package. -->
-    <PackageReference Include="Microsoft.VisualStudio.LanguageServices" Version="1.3.2" />
   </ItemGroup>
 
   <ItemGroup Label="Project references">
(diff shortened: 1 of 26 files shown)


duncanp-sonar:
No code changes.

The `ConnectedMode` project is going to need a reference to `Microsoft.VisualStudio.LanguageServices`, which previously was only used by `Integration.Vsix`. They need to use the same version so I have moved the `PackageReference` to a shared import file.

Adding the package reference to `ConnectedMode` caused all the `package.lock.json` files to be regenerated. It's a lot of noise, which is why I've create this as a separate PR. I'll make the code changes in a later PR.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -148,9 +149,6 @@
          files means it's simpler to always include it so the contents of the lock file don't change for
          signed builds. -->
     <PackageReference Include="Microsoft.VSSDK.Vsixsigntool" Version="16.2.29116.78" PrivateAssets="all" />
-
-    <!-- This package references Roslyn components used by VS. It isn't part of the VSSDK meta-package. -->
-    <PackageReference Include="Microsoft.VisualStudio.LanguageServices" Version="1.3.2" />
```
duncanp-sonar, line 153: Moved this reference to the imports file.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -76,6 +76,7 @@
     <RequiresSystemIOAbstractions>true</RequiresSystemIOAbstractions>
     <RequiresLibGit>true</RequiresLibGit>
     <RequiresChannels>true</RequiresChannels>
+    <RequiresLanguageServices>true</RequiresLanguageServices>
```
duncanp-sonar, line 79: Added a new MSBuild property so projects can indicate that they need the package.

src/SonarLint.VSSpecificAssemblies.props
```diff
@@ -111,4 +111,13 @@
...
   </ItemGroup>
 
+
+  <!-- ******************************************************************************* -->
+  <!-- MS.VS.LanguageServices package -->
+  <!-- ******************************************************************************* -->
+  <ItemGroup Label="MS.VS.LanguageServices" Condition="$(RequiresLanguageServices)=='true'">
+    <!-- This package references Roslyn components used by VS. It isn't part of the VSSDK meta-package. -->
+    <PackageReference Include="Microsoft.VisualStudio.LanguageServices" Version="1.3.2" />
+  </ItemGroup>
```
duncanp-sonar, line 121: Strictly speaking this import file is called `VSSpecificAssemblies`, but this package reference isn't version-specific i.e. we can use the same version in VS2019 and VS2022. However, we don't currently have a better location for this kind of shared reference, and the ALM reference above is not version-specific either.

duncanp-sonar:
There's a test failure but it's a known flaky test rather than a `package.lock.json` problem so I'm assume it will pass on re-run.

Issue #4307, opened 2023-06-07
[Infra] Test 'UpdateAll_CallInProgress_CallIsCancelled' is flaky on the CI agent


duncanp-sonar:
### Description

The test fails intermittently on the CI machine.

#### Assertion message

```
Expected this.OutputStrings {"[Suppressions] Fetching all suppressions...
", "[Suppressions] Fetching all suppressions...
", "[Suppressions] Finished fetching all suppressions
", "[Suppressions] Fetch operation cancelled
"} to contain "[Suppressions] Fetch operation cancelled
".
```

The failure looks weird because it seems from the output that the expected string is in the list of `OutputStrings` (unless there is a an issue with matching the trailing whitespace).

pavel-mikula-sonarsource:
Moved to [SLVS-2203](https://sonarsource.atlassian.net/browse/SLVS-2203)

[SLVS-2203]: https://sonarsource.atlassian.net/browse/SLVS-2203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4348, opened 2023-06-13
Ignore flaky test UpdateAll_CallInProgress_CallIsCancelled

duncanp-sonar:
See #4307


src/ConnectedMode.UnitTests/Suppressions/SuppressionIssueStoreUpdaterTests.cs
```diff
@@ -169,6 +169,7 @@ public void UpdateAll_OperationCancelledException_CancellationMessageLogged()
         }
 
         [TestMethod]
+        [Ignore] // Flaky. See https://github.com/SonarSource/sonarlint-visualstudio/issues/4307
```
duncanp-sonar, line 172: I tried increasing the timeout to see if that made the test less flaky... then I read the ticket I created about the task earlier this week (#4307). It seems from the test log that the required string is present, but the test is still failing -> it requires more investigation -> disable the test for now.

bigfluffycookie (approved):

PR #4350, opened 2023-06-13
Change migrationchecker to use mefFactory

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4346

### DIFF
--- a/src/ConnectedMode.UnitTests/Migration/MigrationCheckerTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/MigrationCheckerTests.cs
@@ -24,6 +24,8 @@ using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.TestInfrastructure;
 
+using Task = System.Threading.Tasks.Task;
+
 namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
 {
     [TestClass]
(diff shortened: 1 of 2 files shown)


src/ConnectedMode/Migration/MigrationChecker.cs
```diff
@@ -24,66 +24,71 @@
...
-            this.migrationPrompt = migrationPrompt;
+            this.mefFactory = mefFactory;
             this.configurationProvider = configurationProvider;
             this.obsoleteConfigurationProvider = obsoleteConfigurationProvider;
 
             activeSolutionTracker.ActiveSolutionChanged += OnActiveSolutionChanged;
 
             // Initial check incase event was fired before we registered.
-            DisplayMigrationPromptIfMigrationIsNeeded();
+            DisplayMigrationPromptIfMigrationIsNeededAsync().Forget();
```
bigfluffycookie, line 56: The threading cookbook says we should not switch threads in an ImportingConstructor however DisplayMigrationPromptIfMigrationIsNeededAsync() maybe does switch but not sure if the Forget() is good enough or not

duncanp-sonar: Good question. Might be worth trying the suggested [code snippet](https://github.com/microsoft/vs-threading/blob/main/doc/cookbook_vs.md#how-do-i-effectively-verify-that-my-code-is-fully-free-threaded) that would force deadlocks to surface.

src/ConnectedMode.UnitTests/Migration/MigrationCheckerTests.cs
```diff
@@ -48,53 +56,57 @@ public void MefCtor_CheckIsExported()
...
+            var mefFactory = CreateMefFactory(migrationPrompt.Object);
 
             var configurationProvider = new Mock<IConfigurationProvider>();
             configurationProvider.Setup(x => x.GetConfiguration()).Returns(CreateBindingConfiguration(mode));
 
             var obsoleteConfigurationProvider = new Mock<IObsoleteConfigurationProvider>();
             obsoleteConfigurationProvider.Setup(x => x.GetConfiguration()).Returns(CreateBindingConfiguration(obsoleteMode));
 
-            _ = new MigrationChecker(Mock.Of<IActiveSolutionTracker>(), migrationPrompt.Object, configurationProvider.Object, obsoleteConfigurationProvider.Object);
+            _ = new MigrationChecker(Mock.Of<IActiveSolutionTracker>(), mefFactory, configurationProvider.Object, obsoleteConfigurationProvider.Object);
```
duncanp-sonar, line 67: Adding a `CreateTestSubject` method in the standard form might reduce some of the duplication in the test setup

duncanp-sonar (approved):
Looks good.

It would be worth trying out the cookbook code snippet to try to force a deadlock if you have time.

src/ConnectedMode/Migration/MigrationChecker.cs
duncanp-sonar: Actually, one simple way around this would be to make the package that creates the migration checker responsible for triggering the initial check (i.e. add a public method to the interface and have the package call it).

Issue #4353, opened 2023-06-14
Add skeleton wizard with functioning buttons.


bigfluffycookie:
Functioning buttons are continue, close, cancel etc.. This task also includes correct migration prompt closing behavior.

PR #4354, opened 2023-06-14
Add skeleton implementation of Wizard with functioning buttons.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4353

After clicking on "Migrate" on the Migration prompt -> 
page1
Both cancel and the "x" button will close the window while the goldbar stays on.
After Clicking Continue -> 
page2pre
After clicking migrate all closing related buttons will be disabled (as in you cannot click on them) and the migrate button disappears. After migration is done -> 
page2af
At this point when the window isc losed the gold bar disappears


src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
```diff
@@ -43,13 +43,15 @@ internal class MigrationWizardController : IMigrationWizardController
...
         public void StartMigrationWizard()
-        { 
-            // TODO show the migration wizard window.
-        }
-
-        private void OnMigrationWizardFinished()
         {
-            MigrationWizardFinished?.Invoke(this, EventArgs.Empty);
+            var migrationWizardWindow = new MigrationWizardWindow();
+
```
bigfluffycookie, line 48: I can't really test this, I could have the window saved as an internal member variable but that feels a bit ugly. I don't think its okay to leave it untested though so I am not sure how to proceed.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public sealed partial class MigrationWizardWindow : DialogWindow
+    {
+        public event EventHandler StartMigration;
+
+        private bool dialogResult;
+
+        internal MigrationWizardWindow()
+        {
+            InitializeComponent();
```
bigfluffycookie, line 36: Technically I should set all the correct visibility states here (buttons and pages) to ensure that it is set up correctly. I chose against this for now because it is cluttering and considering we do not actually know how the ui will look like unnecessary.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        private void NavigateToMigrationProgressPage(object sender, RoutedEventArgs e)
+        {
+            StartWindow.Visibility = Visibility.Collapsed;
+            MigrationProgressWindow.Visibility = Visibility.Visible;
+        }
+
+        private void OnStartMigration(object sender, RoutedEventArgs e)
+        {
+            // Disables all closing / cancel buttons.
```
bigfluffycookie, line 49: My bad, comment is in the wrong place. Should have been one line down.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -0,0 +1,64 @@
...
+            this.migrateButton.Visibility = Visibility.Collapsed;
+            this.IsCloseButtonEnabled = false;
+            MigrationFinished();
```
bigfluffycookie, line 52: Line 52 will happen potentially in a Async.ContinueWith()

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
duncanp-sonar: We'd have to define an interface to wrap the dialog e.g.

```
    internal interface IMigrationDialog
    {
        DialogResult? ShowModal();
    }
```

... and then inject some way of creating into the controller e.g. by adding a factory interface. We use that approach with e.g. [IAnalysisStatusNotifierFactory](https://github.com/SonarSource/sonarlint-visualstudio/blob/03123f51a5f22aa2bd6e347f00a33ca01e9ee44b/src/CFamily/Analysis/CLangAnalyzer.cs#L58)

Instead of creating a factory interface, we could just add a factory method to the controller class (i.e. a method that just creates and returns a new `IMigrationDialog`), and give tests a way of injecting a different behaviour in the constructor e.g. like we do with [TranslateSpans](https://github.com/SonarSource/sonarlint-visualstudio/blob/03123f51a5f22aa2bd6e347f00a33ca01e9ee44b/src/Integration.Vsix/Analysis/IssueConsumerFactory_IssueHandler.cs#L69) in `IssueHandler`.

However, I wouldn't worry about it for the time being. Let's see if we need to make any more changes first,, and add a note to the board to remind us to look at testing this later.

duncanp-sonar (approved):
Looks good!

Issue #4357, opened 2023-06-15
Update skeleton migration interfaces


duncanp-sonar:
### Description

We've decided on XML-based file cleanup i.e. not using the VS project/build property APIs to clean up the files, so the APIs need to be updated.

We don't need separate cleanup types for MSBuild properties and MSBuild item groups.

However, we would still need special logic if we wanted to remove transitively-imported rulesets.

PR #4358, opened 2023-06-15
Update skeleton interfaces

duncanp-sonar:
Relates #4357

### DIFF
--- a/src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
@@ -20,8 +20,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel.Composition;
-using System.ComponentModel.Composition.Hosting;
 using System.Threading;
 using System.Threading.Tasks;
 using SonarLint.VisualStudio.ConnectedMode.Migration;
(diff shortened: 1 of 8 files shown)


src/ConnectedMode/Migration/IConnectedModeMigration.cs
```diff
@@ -71,4 +62,49 @@ public MigrationProgress(int currentProject, int totalProjects, string message,
         /// </summary>
         public bool IsWarning { get; }
     }
+
```
duncanp-sonar, line 65: I've made the interface changes we discussed offline. We might need to tweak them later when we work on the implementation.

src/ConnectedMode/Migration/IConnectedModeMigration.cs
```diff
@@ -71,4 +62,49 @@ public MigrationProgress(int currentProject, int totalProjects, string message,
...
+    /// <summary>
+    /// Finds the set of MSBuild files that need to be cleaned as part of the migration
+    /// </summary>
+    internal interface IFileProvider
+    {
+        /// <summary>
+        /// Returns a list of full paths to the set of MSBuild files to be cleaned
+        /// </summary>
+        /// <remarks>The files can be project files or imported props/targets etc</remarks>
+        Task<IEnumerable<string>> GetFilesAsync(CancellationToken token);
```
duncanp-sonar, line 75: I made a couple of tweaks: I removed the `IProgress` from the file provider and file cleaner.

I think `ConnectedModelMigration` class should be able to raise the higher-level UI progress events, and the cleaner and file provider could just log to the output window.

src/ConnectedMode/Migration/IConnectedModeMigration.cs
```diff
@@ -71,4 +62,49 @@ public MigrationProgress(int currentProject, int totalProjects, string message,
...
+    internal interface IFileCleaner
+    {
+        Task CleanAsync(string filePath, LegacySettings legacySettings, CancellationToken token);
+    }
+
+    /// <summary>
+    /// Data class containing information about the paths to the legacy generated files
+    /// </summary>
+    /// <remarks>Required by <see cref="IFileCleaner"/> to identify the references to remove</remarks>
+    internal class LegacySettings
```
duncanp-sonar, line 92: I added this data class - the cleaner is going to need to know what strings to look for.

src/ConnectedMode/Migration/MSBuildFileCleaner.cs
```diff
@@ -18,33 +18,26 @@
...
 using System.ComponentModel.Composition;
 using System.Threading;
 using System.Threading.Tasks;
-using EnvDTE;
 using SonarLint.VisualStudio.Integration;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Migration
 {
-    [Export(typeof(IProjectCleaner))]
+    [Export(typeof(IFileCleaner))]
```
duncanp-sonar, line 28: I dropped the `RuleSetCleaner` and `IAdditionalFilesCleaner` and added skeleton implementations of the two new interfaces instead.

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -34,15 +31,15 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
...
         private readonly ILogger logger;
-        private readonly IProjectCleaner[] projectCleaners;
-
-        internal /* for testing */ IEnumerable<IProjectCleaner> ProjectCleaners => projectCleaners?.ToList() ?? Enumerable.Empty<IProjectCleaner>();
+        private readonly IFileProvider fileProvider;
+        private readonly IFileCleaner fileCleaner;
 
         [ImportingConstructor]
-        public ConnectedModeMigration(ILogger logger, [ImportMany]IProjectCleaner[] projectCleaners)
+        public ConnectedModeMigration(IFileProvider fileProvider, IFileCleaner fileCleaner, ILogger logger)
```
duncanp-sonar, line 38: Updated to use the new interfaces.

src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
```diff
@@ -41,31 +39,28 @@ public void MefCtor_CheckTypeIsNonShared()
         public void MefCtor_CheckIsExported_NoProjectCleaners()
         {
             MefTestHelpers.CheckTypeCanBeImported<ConnectedModeMigration, IConnectedModeMigration>(
+                MefTestHelpers.CreateExport<IFileProvider>(),
+                MefTestHelpers.CreateExport<IFileCleaner>(),
                 MefTestHelpers.CreateExport<ILogger>());
         }
 
         [TestMethod]
-        public void MefCtor_CheckIsExported_MultipleProjectCleaners()
```
duncanp-sonar, line 48: We no longer need the old, more complicated test for `ImportMany` since we just import single instances now.

src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
```diff
@@ -41,31 +39,28 @@ public void MefCtor_CheckTypeIsNonShared()
...
+        public void Migrate_FilesExist_AllFilesCleaned()
```
duncanp-sonar, line 48: Simple test of the new happy-path.

duncanp-sonar:
I'm not worried about the low coverage. We'll add tests when we do the implementations.

src/ConnectedMode/Migration/IConnectedModeMigration.cs
bigfluffycookie: Yea I agree, makes sense that way to me.

bigfluffycookie (approved):
LGTM

Issue #4355, opened 2023-06-15
Implement IConnectedModeMigration in MigrationWizard


bigfluffycookie:
Simple implementation that updates the ui whenever new progress gets reported.

PR #4366, opened 2023-06-16
Implement IConnectedModeMigration in MigrationWizard and add simple ui

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4355

listview

### DIFF
--- a/src/ConnectedMode/ConnectedMode.csproj
+++ b/src/ConnectedMode/ConnectedMode.csproj
@@ -1,5 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  
+
   <Import Project="..\SonarLint.Product.props" />
   <Import Project="..\SonarLint.VsPkgRefs.props" />
 
--- a/src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
+++ b/src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
@@ -19,20 +19,31 @@
  */
 
 using System;
-using System.Windows;
-using Microsoft.VisualStudio.PlatformUI;
 using System.ComponentModel;
+using System.Threading;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Media;
+using Microsoft.VisualStudio.PlatformUI;
+using Microsoft.VisualStudio.Threading;
+using Task = System.Threading.Tasks.Task;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Migration.Wizard
 {
-    public sealed partial class MigrationWizardWindow : DialogWindow
+    public sealed partial class MigrationWizardWindow : DialogWindow, IProgress<MigrationProgress>
     {
         public event EventHandler StartMigration;
 
         private bool dialogResult;
 
-        internal MigrationWizardWindow()
+        private readonly IConnectedModeMigration connectedModeMigration;
+
+        private bool migrationInProgress;
+
+        internal MigrationWizardWindow(IConnectedModeMigration connectedModeMigration)
         {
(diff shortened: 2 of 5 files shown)


src/ConnectedMode/ConnectedMode.csproj
```diff
@@ -1,5 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  
+
```
bigfluffycookie, line 2: I do not know why this empty line is here however my bad for not spotting it before pushing

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -46,9 +57,19 @@ private void NavigateToMigrationProgressPage(object sender, RoutedEventArgs e)
...
             // Disables all closing / cancel buttons.
             this.migrateButton.Visibility = Visibility.Collapsed;
             this.IsCloseButtonEnabled = false;
+
+            MigrateAsync().Forget();
+        }
+
+        private async Task MigrateAsync()
+        {
+            await connectedModeMigration.MigrateAsync(this, CancellationToken.None);
```
bigfluffycookie, line 72: I checked out .ContinueWith but using that would make everything a bit ugly. I think it is okay to use await instead here. It is not the exact same functionality but it does not really matter too much here.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -46,9 +57,19 @@ private void NavigateToMigrationProgressPage(object sender, RoutedEventArgs e)
 
         private void OnStartMigration(object sender, RoutedEventArgs e)
         {
+            if (migrationInProgress) { return; }
```
bigfluffycookie, line 60: I added this simply to ensure that OnStartMigration can never be executed if it is already running

src/ConnectedMode/ConnectedMode.csproj
duncanp-sonar: It doesn't matter too much. _Not_ having whitespace is more of a problem as that is more likely to affect readability.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -59,6 +80,18 @@ private void MigrationFinished()
...
-        private void OnClosing(object sender, CancelEventArgs e) => this.DialogResult = dialogResult;
+        private void OnClosing(object sender, CancelEventArgs e)
+        {
+            migrationInProgress = false;
+            this.DialogResult = dialogResult;
+        }
+
+        void IProgress<MigrationProgress>.Report(MigrationProgress value)
+        {
+            ListBoxItem item = new ListBoxItem();
```
duncanp-sonar, line 91: How does this work from a threading point of view? Presumably this update needs to be on the UI thread.
Does the whole process take place on the UI thread i.e. does the migration start on the UI thread?
(not necessarily a problem now, but it might be when we start doing "real work" during the migration).

duncanp-sonar (approved):
Looks good!

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
bigfluffycookie: hmm yea good point, probably we have to add some thread handling. Should I do it in this pr already?

duncanp-sonar: No, just a new note on the board.

PR #4368, opened 2023-06-16
Add XmlDocumentHelper that simplifies XML handling

georgii-borovinskikh-sonarsource:


src/ConnectedMode.UnitTests/Migration/XmlDocumentHelperTests.cs
```diff
@@ -0,0 +1,133 @@
...
+        MefTestHelpers.CheckTypeCanBeImported<XmlDocumentHelper, IXmlDocumentHelper>();
+
+    [TestMethod]
+    public void MefCtor_CheckTypeIsShared() =>
+        MefTestHelpers.CheckIsSingletonMefComponent<XmlDocumentHelper>();
+    
+    [TestMethod]
+    public void LoadFromString_SaveToString_NoModification_ReturnsEquivalentString()
+    {
+        var input =
```
bigfluffycookie, line 40: Is it github formatting it like this? If not I'd consider fixing it a bit for readability

bigfluffycookie (approved):
Looks good, just a formatting question for the tests.

src/ConnectedMode.UnitTests/Migration/XmlDocumentHelperTests.cs
georgii-borovinskikh-sonarsource: I formatted it like this on purpose so that the multiline string indentation is consistent. However, in the other 2 tests something happened to the formatting

src/ConnectedMode/Migration/XmlDocumentHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <summary>
+        /// Loads <see cref="XmlDocument"/> from <see cref="string"/> with preserved original formatting
+        /// </summary>
+        /// <param name="content">XML document contents</param>
+        /// <returns><see cref="XmlDocument"/> object representation</returns>
+        XmlDocument LoadFromString(string content);
+        /// <summary>
+        /// Saves <see cref="XmlDocument"/> to <see cref="string"/>
+        /// </summary>
+        /// <remarks>Does not force add ?xml? header if the original XML didn't have it</remarks>
```
duncanp-sonar, line 41: Could mention that it preserves the existing formatting and whitespace.

... and add a new line at line 38.

src/ConnectedMode/Migration/XmlDocumentHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            var xmlDocument = new XmlDocument { PreserveWhitespace = true };
+            xmlDocument.LoadXml(content);
+            return xmlDocument;
+        }
+
+        public string SaveToString(XmlDocument document)
+        {
+            var memoryStream = new MemoryStream();
+            var streamReader = new StreamReader(memoryStream);
```
duncanp-sonar, line 62: Both MemoryStream and StreamReader should be disposed.
(I'm surprised there weren't code smells)

src/ConnectedMode/Migration/XmlDocumentHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+        public XmlDocument LoadFromString(string content)
+        {
+            var xmlDocument = new XmlDocument { PreserveWhitespace = true };
+            xmlDocument.LoadXml(content);
+            return xmlDocument;
+        }
+
+        public string SaveToString(XmlDocument document)
+        {
+            var memoryStream = new MemoryStream();
```
duncanp-sonar, line 61: Might be worth commenting explicitly that other ways of saving the XML can introduce unwanted changes.

src/ConnectedMode/Migration/XmlDocumentHelper.cs
georgii-borovinskikh-sonarsource: But they both don't have any unmanaged resources. I can add disposal so that it's not confusing to the person who looks at this code in the future, but it shouldn't change anything functionally

duncanp-sonar (commented):
Looks good.  A couple of minor comments inline.

PR #4388, opened 2023-06-19
Add CancellationToken

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4370

### DIFF
--- a/src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
+++ b/src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
@@ -27,6 +27,7 @@ using System.Windows.Media;
 using Microsoft.VisualStudio.PlatformUI;
 using Microsoft.VisualStudio.Threading;
 using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Integration;
 using Task = System.Threading.Tasks.Task;
 
@@ -36,6 +37,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration.Wizard
     {
         public event EventHandler StartMigration;
 
+        private readonly BoundSonarQubeProject oldBinding;
         private readonly IConnectedModeMigration connectedModeMigration;
         private readonly ILogger logger;
 
@@ -44,8 +46,9 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration.Wizard
 
         private CancellationTokenSource cancellationTokenSource;
 
-        internal MigrationWizardWindow(IConnectedModeMigration connectedModeMigration, ILogger logger)
+        internal MigrationWizardWindow(BoundSonarQubeProject oldBinding, IConnectedModeMigration connectedModeMigration, ILogger logger)
         {
+            this.oldBinding = oldBinding;
             this.connectedModeMigration = connectedModeMigration;
             this.logger = logger;
 
@@ -78,7 +81,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration.Wizard
         {
             try
             {
-                await connectedModeMigration.MigrateAsync(this, cancellationTokenSource.Token);
+                await connectedModeMigration.MigrateAsync(oldBinding, this, cancellationTokenSource.Token);
                 MigrationFinished();
             }
             catch (OperationCanceledException ex)
--- a/src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
@@ -23,27 +23,49 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using SonarLint.VisualStudio.ConnectedMode.Migration;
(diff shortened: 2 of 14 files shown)


src/ConnectedMode/Migration/MigrationStrings.resx
```diff
@@ -117,6 +117,12 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="CancelTokenFailure_NormalLog" xml:space="preserve">
+    <value>[Migration] Failed to close migration window correctly: {0}</value>
+  </data>
+  <data name="CancelTokenFailure_VerboseLog" xml:space="preserve">
+    <value>[MigrationWizardXaml] Failed to close migration windowe correctly: {0}</value>
```
bigfluffycookie, line 124: windowe

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -69,19 +75,29 @@ private void OnStartMigration(object sender, RoutedEventArgs e)
 
         private async Task MigrateAsync()
         {
-            await connectedModeMigration.MigrateAsync(this, CancellationToken.None);
-            MigrationFinished();
+            try
```
bigfluffycookie, line 78: You were right, we need to try catch in here.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -85,6 +85,11 @@ private async Task MigrateAsync()
                 logger.LogVerbose(MigrationStrings.CancelTokenFailure_VerboseLog, ex);
                 logger.WriteLine(MigrationStrings.CancelTokenFailure_NormalLog, ex.Message);
             }
+            catch(Exception ex)
+            {
+                logger.LogVerbose(MigrationStrings.ErrorDuringMigation_VerboseLog, ex);
```
bigfluffycookie, line 90: I really struggle with knowing when to log verbose and when just normal logging is fine so I usually log both since it covers the most scenarios.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -69,19 +75,29 @@ private void OnStartMigration(object sender, RoutedEventArgs e)
...
-            MigrationFinished();
+            try
+            {
+                await connectedModeMigration.MigrateAsync(this, cancellationTokenSource.Token);
+                MigrationFinished();
+            }
+            catch (OperationCanceledException ex)
+            {
+                logger.LogVerbose(MigrationStrings.CancelTokenFailure_VerboseLog, ex);
+                logger.WriteLine(MigrationStrings.CancelTokenFailure_NormalLog, ex.Message);
```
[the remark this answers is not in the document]
bigfluffycookie: I've never considered having multiple catch blocks but that makes sense, is it a common thing to do or not so much?

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
duncanp-sonar: I think it's a good pattern to follow for errors: the user will normally see the "normal" verbosity message if something goes wrong, and we can ask for the verbose message to help with support/troubleshooting.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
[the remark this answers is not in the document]
duncanp-sonar: No, it's not common. It only really makes sense if there are specific exceptions that you know can occur and want to handle differently, like here.

Here, the cancellation exception is really an implementation detail; it's how the code is signalling internally that the user has closed the dialogue. So we don't really want to log it as an error.

src/ConnectedMode/Migration/MigrationStrings.resx
```diff
@@ -117,6 +117,12 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="CancelTokenFailure_NormalLog" xml:space="preserve">
+    <value>[Migration] Migration cancelled: {0}</value>
```
duncanp-sonar, line 121: I would make this message `[Migration] Migration cancelled by user` - I don't think show the user the exception message will help, and might confuse the user.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -69,19 +75,34 @@ private void OnStartMigration(object sender, RoutedEventArgs e)
...
+            }
+            catch(Exception ex)
```
duncanp-sonar, line 88: Minor - to be consistent, don't catch critical errors.

PR #4389, opened 2023-06-19
(2) Implement basic file provider

### LINKED ISSUE #4377: Implement basic file provider
### Description

Implement a minimal version of `IFileProvider` that just looks under the solution folder.

More complex behaviour will be added in separate ticket e.g. directory build props, checking project files are in the list etc.

duncanp-sonar:
Fixes #4377

### DIFF
--- a/src/ConnectedMode.UnitTests/Migration/FileProviderTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/FileProviderTests.cs
@@ -112,6 +112,32 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
             actual.Should().BeEquivalentTo("should be included1", "should be included2\\obj", "should be included3");
         }
 
+        [TestMethod]
+        public async Task GetFiles_DuplicatesAreIgnored()
+        {
+            var solution = CreateIVsSolution("root dir");
+            var serviceProvider = CreateServiceProviderWithSolution(solution.Object);
+
+            var filesInFileSystem = new string[] {
+                // duplicates - should only appear once
+                "file_1",
+                "FILE_1",
+                "file_1",
+
+                // non-duplicates - should be included
+                "file_11",
+                "file_111"
+            };
+
+            var fileSystem = CreateFileSystem(filesInFileSystem);
+
+            var testSubject = CreateTestSubject(serviceProvider.Object, fileSystem.Object);
+
+            var actual = await testSubject.GetFilesAsync(CancellationToken.None);
+
+            actual.Should().BeEquivalentTo("file_1", "file_11", "file_111");
+        }
+
         private static MSBuildFileProvider CreateTestSubject(IServiceProvider serviceProvider = null,
             IFileSystem fileSystem = null, ILogger logger = null, IThreadHandling threadHandling = null)
         {
--- a/src/ConnectedMode.UnitTests/Migration/FileProviderTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/FileProviderTests.cs
@@ -88,6 +88,30 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
             fileSystem.Verify(x => x.Directory.GetFiles("root dir", "*.vbproj", SearchOption.AllDirectories), Times.Once);
         }
 
+        [TestMethod]
+        public async Task GetFiles_FilesInExcludedDirectoriesAreNotReturned()
+        {
+            var solution = CreateIVsSolution("root dir");
+            var serviceProvider = CreateServiceProviderWithSolution(solution.Object);
+
+            var filesInFileSystem = new string[] {
+                "should be included1",
+                "\\bin\\file1",
+                "aaa\\bin\\zzz",
+                "should be included2\\obj", // doesn't contain \obj\
+                "ccc\\obj\\xxx",
+                "should be included3"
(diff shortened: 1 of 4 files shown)


src/ConnectedMode/Migration/MSBuildFileProvider.cs
```diff
@@ -31,18 +38,111 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
...
         private readonly ILogger logger;
+        private readonly IThreadHandling threadHandling;
+        private readonly IFileSystem fileSystem;
+
+        internal static readonly string[] FileSearchPatterns = new string[] {
+            "*.ruleset",
+            "*.props",
+            "*.targets",
+            "*.csproj",
+            "*.vbproj"
```
duncanp-sonar, line 51: `Directories.GetFiles` only matches one pattern at a time, so we need to call it multiple times and aggregate the results, dropping out any duplicates.

I don't have test for de-duplication - I'll add one.

src/ConnectedMode/Migration/MSBuildFileProvider.cs
```diff
@@ -31,18 +38,111 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
...
+        };
+
+        internal static readonly string[] ExcludedDirectores = new string[]
+        {
+            "\\bin\\",
+            "\\obj\\"
+        };
```
duncanp-sonar, line 58: Testing this on SLVS itself was picking up lots of generated .props and .targets files in the `obj` directory. I've added simple filtering to exclude both `obj` and `bin` directories.

src/ConnectedMode/Migration/MSBuildFileProvider.cs
```diff
@@ -31,18 +38,111 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
...
+        {
+            var solutionDir = await GetSolutionDirectoryAsync();
+            if (solutionDir == null)
+            {
+                LogVerbose("Unable to locate the solution folder.");
+                return Enumerable.Empty<string>();
+            }
+
+            // Make sure we are doing the disk search on a background thread
+            await threadHandling.SwitchToBackgroundThread();
```
duncanp-sonar, line 91: We'll need to look at this as part of the end-to-end migration threading, starting from the wizard down.

src/ConnectedMode/Migration/MSBuildFileProvider.cs
```diff
@@ -31,18 +38,111 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
...
+                var files = fileSystem.Directory.GetFiles(rootFolder, pattern, SearchOption.AllDirectories);
+                var filesToInclude = files.Where(x => !IsInExcludedDirectory(x)).ToArray();
+
+                AddToMatches(allMatches, filesToInclude);
+                LogVerbose($"  Pattern: {pattern}, Number of matching files: {filesToInclude.Length}");
+            }
+
+            timer.Stop();
+            LogVerbose("Total number of matching files: " + allMatches.Count);
+            LogVerbose("Search time (ms): " + timer.ElapsedMilliseconds);
```
duncanp-sonar, line 131: Logging to give us some of idea of what the performance impact in the real-world is.

src/ConnectedMode/Migration/MSBuildFileProvider.cs
```diff
@@ -31,18 +38,111 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
...
+
+            return allMatches;
         }
 
-        public Task<IEnumerable<string>> GetFilesAsync(CancellationToken token)
+        private static bool IsInExcludedDirectory(string fullPath) => ExcludedDirectores.Any(x => fullPath.Contains(x));
+
+        private static void AddToMatches(HashSet<string> results, IEnumerable<string> matches)
```
duncanp-sonar, line 138: `HashSet<>` doesn't have an `AddRange` method (unlike `List<>`) so we need this helper method.

src/ConnectedMode.UnitTests/Migration/FileProviderTests.cs
```diff
@@ -31,11 +40,115 @@ public class MSBuildFileProviderTests
...
+            var actual = await testSubject.GetFilesAsync(CancellationToken.None);
+
+            actual.Should().BeEquivalentTo(filesToReturn);
+            solution.VerifyAll();
+
+            fileSystem.Verify(x => x.Directory.GetFiles("root dir", "*.ruleset", SearchOption.AllDirectories), Times.Once);
+            fileSystem.Verify(x => x.Directory.GetFiles("root dir", "*.props", SearchOption.AllDirectories), Times.Once);
+            fileSystem.Verify(x => x.Directory.GetFiles("root dir", "*.targets", SearchOption.AllDirectories), Times.Once);
+            fileSystem.Verify(x => x.Directory.GetFiles("root dir", "*.csproj", SearchOption.AllDirectories), Times.Once);
+            fileSystem.Verify(x => x.Directory.GetFiles("root dir", "*.vbproj", SearchOption.AllDirectories), Times.Once);
```
duncanp-sonar, line 88: If we have time, we might fetch the project files in a different way to make sure we only pick up the projects reference in the solution (the `RoslynProjectWalker` class gives us an easy way to find just the Roslyn projects).

Issue #4242, opened 2023-05-24
Implement ruleset cleanup logic


duncanp-sonar:
### Description

Attempt to remove the `<CodeAnalysisRuleSet>` entries that point at the ruleset file(s) generated by SonarLint.

This only applies to C# and VB.NET projects. There are separate ruleset files for C# and VB.NET.

It is not always possible to remove build properties using the VS APIs e.g. if the `CodeAnalysisRuleSet` property is set in an imported props file then VS will add an empty `<CodeAnalysisRuleSet></CodeAnalysisRuleSet>` entry to the importing project. This is functionally correct, but is probably now what the user expects.

Prototype: https://github.com/duncanp-sonar/slvs-prototypes/tree/main/BindingCleanup

#### Scenarios
* Generated ruleset imported referenced directly in a project file. _This is the simplest case and is what VS will add when binding in the simple case._
* Generated ruleset imported referenced indirectly in imported file e.g. Directory.Build.props / another imported file
* Generated ruleset imported referenced indirectly from custom ruleset file
e.g. project reference A.ruleset which Includes our generated ruleset
e.g. project reference A.ruleset which Includes B.ruleset which Includes our generated ruleset

Note: there are differences in behaviour between old-style and SDK-style MSBuild projects. We can use the same VS API (e.g. IVsBuildPropertyStorage.RemoveProperty), but the results differ because the underlying project system implementations are different.

#### TODO
 - [x] clean up `<CodeAnalysisRuleSet>` properties in MSBuild files
 - [x] clean up transitive `Includes` in ruleset files

PR #4393, opened 2023-06-19
(1) Clean references from .ruleset files

duncanp-sonar:
Clean <Include Path="..." /> references from `.ruleset` files

Part of #4242

### DIFF
--- a/src/ConnectedMode/Migration/MSBuildFileCleaner.ElementAndAttributeTailMatcher.cs
+++ b/src/ConnectedMode/Migration/MSBuildFileCleaner.ElementAndAttributeTailMatcher.cs
@@ -55,9 +55,9 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
                 return attributeCollection
                             .Cast<XmlAttribute>()
                             .Any(attribute =>
-                                valuesToTailMatch.Any(path =>
+                                valuesToTailMatch.Any(val =>
                                     attribute.Name == attributeName &&
-                                    attribute.Value.EndsWith(path, System.StringComparison.OrdinalIgnoreCase)));
+                                    attribute.Value.EndsWith(val, System.StringComparison.OrdinalIgnoreCase)));
             }
         }
     }


src/ConnectedMode.UnitTests/Migration/FileCleanerTestCases/Ruleset_NoIncludedRulesets_Input.ruleset.xml
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="My local ruleset" ToolsVersion="17.0">
+  <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.Features" RuleNamespace="Microsoft.CodeAnalysis.CSharp.Features">
```
duncanp-sonar, line 3: Nothing to clean up in this file

src/ConnectedMode.UnitTests/Migration/FileCleanerTestCases/Ruleset_XX-project-key_Input.ruleset.xml
```diff
@@ -0,0 +1,20 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="My local ruleset" ToolsVersion="17.0">
+
+  <!-- Both vb and csharp references should be removed -->
```
duncanp-sonar, line 4: A couple of references that should be removed

src/ConnectedMode/Migration/MSBuildFileCleaner.ElementAndAttributeTailMatcher.cs
```diff
@@ -0,0 +1,64 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Xml;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Migration
+{
+    internal partial class MSBuildFileCleaner
+    {
+        private static class ElementAndAttributeTailMatcher
```
duncanp-sonar, line 29: The AdditionalFile references look like this:
`<AdditionalFiles Include="xxx[partial value]" />`

... and the ruleset Include reference looks like this:
`<Include Path="xxx[partial value]" />`

Logically cleaning up the AdditionalFiles and ruleset Includes follow the same pattern:
- matching first an element name, then an attribute name, then a value in the attribute; so I factored it out into a helper class.

src/ConnectedMode/Migration/MSBuildFileCleaner.ElementAndAttributeTailMatcher.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            /// <summary>
+            /// Returns a list of nodes in the document matching the following criteria:
+            ///   <ElementName AttributeName="xxxx[value to tail match] ... />
+            /// e.g.
+            ///   <Include Path="..\..\.sonarlint\my-project-keyvb.ruleset" Action="Default" />
+            /// </summary>
+            public static IList<XmlNode> Find(XmlDocument document, string elementName, string attributeName,
```
duncanp-sonar, line 37: This is effectively the two methods moved from the cleaner class, but parameterised with the element and attribute names.

src/ConnectedMode/Migration/MSBuildFileCleaner.cs
```diff
@@ -30,15 +30,13 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
 {
     [Export(typeof(IFileCleaner))]
     [PartCreationPolicy(CreationPolicy.NonShared)]
-    internal class MSBuildFileCleaner : IFileCleaner
+    internal partial class MSBuildFileCleaner : IFileCleaner
```
bigfluffycookie, line 33: Why did you make this a partial class? (Not questioning the decision I've just never made a partial class)

bigfluffycookie (approved):
Looks good just one question

src/ConnectedMode/Migration/MSBuildFileCleaner.ElementAndAttributeTailMatcher.cs
```diff
@@ -0,0 +1,64 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Xml;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Migration
+{
+    internal partial class MSBuildFileCleaner
```
duncanp-sonar, line 27: @bigfluffycookie I wanted to make the `ElementAndAttributeTailMatcher` helper class a `private` class, since it's only ever going to be used by the cleaner class (we're not testing it independently, just via the cleaner class).

To make it `private` it has to be nested inside another class, in this case the cleaner class.

However, I also want to put the code in a separate file. So that meant making the cleaner class `partial` so I could split its definition across multiple files and put the new helper class as a private  class inside one of them.

src/ConnectedMode/Migration/MSBuildFileCleaner.cs
duncanp-sonar: (answered in separate comment)

PR #4394, opened 2023-06-19
(2) Clean ruleset property references from project files

duncanp-sonar:
Part of #4242


src/ConnectedMode.UnitTests/Migration/MSBuildFileCleanerTests.cs
```diff
@@ -91,7 +91,7 @@ public void Clean_AdditionalFileRefsExist_DifferentProjectKey_SettingsAreNotRemo
         }
 
         [TestMethod]
-        public void Clean_RulesetRefsExist_CorrectProjectKey_SettingsAreRemoved()
```
duncanp-sonar, line 94: I renamed the existing tests to try to make the differences between the ruleset tests clearer.

We have two sorts of ruleset tests: 
* one the cleans up `Include` elements in `.ruleset` files, and
* the new tests that check `CodeAnalysisRuleSet` properties inside MSBuild project files.

src/ConnectedMode/Migration/MSBuildFileCleaner.ElementAndValueTailMatcher.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace SonarLint.VisualStudio.ConnectedMode.Migration
+{
+    internal partial class MSBuildFileCleaner
+    {
+        private static class ElementAndValueTailMatcher
+        {
+            /// <summary>
+            /// Returns a list of nodes in the document matching the following criteria:
+            ///   <ElementName ... ></ElementName>"xxxx[value to tail match]</ElementName>
```
duncanp-sonar, line 33: This class follows a similar pattern to the other helper class. but the condition it is looking for is simpler since we just need to check the element name and the value it contains.

bigfluffycookie (approved):

PR #4395, opened 2023-06-20
Add migration logger output

duncanp-sonar:


src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -61,24 +61,25 @@ public async Task MigrateAsync(BoundSonarQubeProject oldBinding, IProgress<Migra
 
             logger.WriteLine(MigrationStrings.Starting);
 
-            // TODO: cleanup - dummy implementation to provide feedback for the UI
-            progress?.Report(new MigrationProgress(1, 2, "TODO 1", false));
-            progress?.Report(new MigrationProgress(2, 2, "TODO 2", true));
```
duncanp-sonar, line 66: I've added some more representative `progress?.Report` calls, but they are not supposed to be the final version.

src/ConnectedMode/Migration/MSBuildFileCleaner.cs
```diff
@@ -63,7 +63,7 @@ public string Clean(string content, LegacySettings legacySettings, CancellationT
 
             if (!nodesToRemove.Any())
             {
-                logger.LogVerbose("No settings to remove");
+                logger.LogVerbose("    No settings to remove");
```
duncanp-sonar, line 66: Cosmetic: a non-verbose message about the file being changed is written just before this message. Indenting this verbose message makes it easier to see the two messages are related.

bigfluffycookie (approved):

Issue #4396, opened 2023-06-20
Connect to SonarQubeService in ConnectedModeMigration


bigfluffycookie:
### Description

To trigger a new binding we need the `SonarQubeService` to be connected to the server so we can fetch data from it.

TODO
- if we're not already connected to the server when migration starts, establish the connection.
- if the migration fails and we established the server connection, we should disconnect it.

#### Background
There are a couple of scenarios:

1) First migration
In this case, we'll detect the old binding configuration and start the migration process, but the service will not already be connected. So, we'll need to connect to the server as part of the migration process. If the migration succeeds, we can leave the server connected. If it fails, we should disconnect it.

2) Subsequent migration attempt
i.e. an earlier migration attempt failed. In this case, the connection to the server might or might not be established, depending on how far through the migration process we got. If the migration process got as far as writing the new binding files we would expect the server to already be connected. In that case, we should not disconnect the binding even if the migration fails.
Otherwise, we're in the same situation as scenario (1).

PR #4398, opened 2023-06-20
Connect to SonarQubeService in ConnectedModeMigration

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4396

### DIFF
--- a/src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
@@ -51,6 +51,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
                 MefTestHelpers.CreateExport<IFileProvider>(),
                 MefTestHelpers.CreateExport<IFileCleaner>(),
                 MefTestHelpers.CreateExport<IVsAwareFileSystem>(),
+                MefTestHelpers.CreateExport<ISonarQubeService>(),
                 MefTestHelpers.CreateExport<ILogger>());
         }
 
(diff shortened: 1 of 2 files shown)


src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
```diff
@@ -189,21 +191,65 @@ public async Task Migrate_HasProgressListener_ReportsProgress()
             progressMessages.Should().NotBeEmpty();
         }
 
+        [TestMethod]
+        [DataRow(true, false)]
+        [DataRow(false, true)]
+        public async Task Migrate_ConnectToSonarQubeIfNeeded_DisconnectedNotCalled(bool isConnectedToNewConnectedMode, bool expectedResult)
```
bigfluffycookie, line 197: If we successfully migrate the connection stays as is.

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -42,19 +44,53 @@ private sealed class ChangedFiles : List<FilePathAndContent<string>> { }
         private readonly IFileProvider fileProvider;
         private readonly IFileCleaner fileCleaner;
         private readonly IVsAwareFileSystem fileSystem;
+        private readonly ISonarQubeService sonarQubeService;
         private readonly ILogger logger;
 
+        // The user can have both the legacy and new connected mode files. In this case the behaviour is slightly different.
```
duncanp-sonar, line 50: `In this case the behaviour is slightly different.` isn't very enlightening ;-)

Maybe something like `In that case, we expect the SonarQubeService to already be connected.` ?

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -42,19 +44,53 @@ private sealed class ChangedFiles : List<FilePathAndContent<string>> { }
...
+        private bool isConnectedToNewConnectedMode;
```
duncanp-sonar, line 51: Naming - `isSonarQubeServerAlreadyConnected`, `isAlreadyConnectedToServer` ?

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -42,19 +44,53 @@ private sealed class ChangedFiles : List<FilePathAndContent<string>> { }
...
+               await sonarQubeService.ConnectAsync(oldBinding.CreateConnectionInformation(), token);
+            }
+
+            try
+            {
+                await MigrateImplAsync(oldBinding, progress, token);
+            }
+            catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+            {
+                if (!isConnectedToNewConnectedMode)
```
duncanp-sonar, line 84: It's probably worth adding a comment here e.g. `If we establish the server connection during migration, disconnect it if the migration failed.`

src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
duncanp-sonar: This tests the `migration successful` case, doesn't it?

If so:
* it might help to make the method name even longer...
* do we need both parameters? i.e. is the second always the inverse of the first?


e.g.

```
public async Task Migrate_ConnectToSonarQubeIfNeeded_MigrationSucceeds_DisconnectedNotCalled(bool isAlreadyConnected)

```

src/ConnectedMode.UnitTests/Migration/ConnectedModeMigrationTests.cs
```diff
@@ -189,21 +192,65 @@ public async Task Migrate_HasProgressListener_ReportsProgress()
...
+            await testSubject.MigrateAsync(AnyBoundProject, null, CancellationToken.None);
+
+            sonarQubeService.Verify(x => x.ConnectAsync(It.IsAny<ConnectionInformation>(), CancellationToken.None), expectedResult ? Times.Once : Times.Never);
+            sonarQubeService.Verify(x => x.Disconnect(), Times.Never);
+        }
+
+        [TestMethod]
+        [DataRow(true, false)]
+        [DataRow(false, true)]
+        public async Task Migrate_DisconnectToSonarQubeIfNeeded_NonCritical_HandledAndThrown(bool isConnectedToNewConnectedMode, bool expectedResult)
```
duncanp-sonar, line 212: Same question about the second argument here - do we need both?

`isAlreadyConnected` might be better than `isConnectedToNewConnectedMode`.

duncanp-sonar (commented):
The code looks good. Just some comments about comments and naming.

duncanp-sonar (approved):

PR #4399, opened 2023-06-20
Implement UnintrusiveBindingController

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4240


src/ConnectedMode/Migration/IUnintrusiveBindingController.cs
```diff
@@ -19,19 +19,46 @@
...
-        public void Bind(BoundSonarQubeProject project) { }
+        private readonly IBindingProcessFactory bindingProcessFactory;
+
+        [ImportingConstructor]
+        public UnintrusiveBindingController(IBindingProcessFactory bindingProcessFactory)
+        {
+            this.bindingProcessFactory = bindingProcessFactory;
+        }
+
+        public async Task BindAsync(BoundSonarQubeProject project, CancellationToken token)
```
bigfluffycookie, line 47: For now this doesn't report any progress. The way I see it we could either directly do it in here or do it in the BindingProcess. Either way this is a different ticket problem

src/ConnectedMode/Migration/IUnintrusiveBindingController.cs
```diff
@@ -19,19 +19,46 @@
...
+        {
+            var bindingProcess = CreateBindingProcess(project);
+
+            await bindingProcess.DownloadQualityProfileAsync(null, token);
```
bigfluffycookie, line 51: Downloading Quality profile seems to take around 10 seconds.

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -125,10 +128,11 @@ private async Task MigrateImplAsync(BoundSonarQubeProject oldBinding, IProgress<
                 logger.WriteLine(MigrationStrings.Process_SkippingChecking);
             }
 
-            // TODO - trigger unintrusive binding process (will need the binding arguments)
             progress?.Report(new MigrationProgress(0, 1, "TODO: Create new binding", true));
             logger.WriteLine(MigrationStrings.Process_ProcessingNewBinding);
 
+            await unintrusiveBindingController.BindAsync(oldBinding, token);
```
bigfluffycookie, line 134: One thing we should discuss later is that right now the progress will not be updated in the UI until this gets hit. For me it just showed the first report -> Getting files.. and then nothing up until the code goes into this method at which point all the following logs show. Is this because of the threading or?

src/ConnectedMode/Migration/IUnintrusiveBindingController.cs
duncanp-sonar: Ouch. Well, it's in the background...

duncanp-sonar (commented):
Code looks good. One question about whether we can check the bound project arguments.

PR #4401, opened 2023-06-20
Rename file cleaner and provider types

duncanp-sonar:
- no code changes, just file renames


src/ConnectedMode.UnitTests/Migration/XmlFileCleanerTests.cs
```diff
@@ -27,20 +27,20 @@
 namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
 {
     [TestClass]
-    public class MSBuildFileCleanerTests
+    public class XmlFileCleanerTests
```
duncanp-sonar, line 30: The file provider and cleaner classes are not really MSBuild-specific as they handle rulesets as well as MSBuild files.
Both formats are XML, though, so this change just renames them to `XmlXXX`.

bigfluffycookie (approved):

Issue #4365, opened 2023-06-16
Goldbar if user is in new connected mode and old connected mode


bigfluffycookie:
There can be a situation where the user reverts our deletion but the new binding is already in place. In that case it would make sense to have a gold bar with a slightly different message that triggers the clean up.

PR #4402, opened 2023-06-20
Change Migration prompt to show if user has both old and new style connected mode.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4365

### DIFF
--- a/src/ConnectedMode/Migration/IMigrationPrompt.cs
+++ b/src/ConnectedMode/Migration/IMigrationPrompt.cs
@@ -36,6 +36,9 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
     /// </summary>
     internal interface IMigrationPrompt : IDisposable
     {
+        /// <summary>
+        /// Shows a prompt async with a different message depending on the value of hasNewBindingFiles.
+        /// </summary>
         Task ShowAsync(BoundSonarQubeProject oldBinding, bool hasNewBindingFiles);
 
         void Clear();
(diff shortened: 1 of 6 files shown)


src/ConnectedMode/Migration/IMigrationPrompt.cs
```diff
@@ -35,7 +36,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
     /// </summary>
     internal interface IMigrationPrompt : IDisposable
     {
-        Task ShowAsync(BoundSonarQubeProject oldBinding);
+        Task ShowAsync(BoundSonarQubeProject oldBinding, bool isAlreadyConnected);
```
duncanp-sonar, line 39: Probably worth adding a doc comment to this interface now.

duncanp-sonar: Could we call `isAlreadyConnected` something else to avoid confusion? (since we use it somewhere else to mean something subtly different)

e.g. `hasNewBinding`, `hasNewBindingFiles` ?

bigfluffycookie: Hmm what do you mean we use it to mean something different? Does it not mean the exact same?

duncanp-sonar (approved):
Looks good. One suggestion about a variable name.

src/ConnectedMode/Migration/IMigrationPrompt.cs
duncanp-sonar: Here it means "has connected mode settings".

In the other usage it means "SonarQubeService.Connected=true". 

They are subtly different.

Issue #4403, opened 2023-06-20
Add progress report to UnintrusiveBindingController


bigfluffycookie:
The normal binding process reports from within BindingProcessImpl in the form of a IProgress<FixedStepsProgress>. Add a converter class that maps it to IProgress<MigrationProgress>

PR #4404, opened 2023-06-20
Add progress report to UnintrusiveMigrationController

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4403

### DIFF
--- a/src/ConnectedMode.UnitTests/Migration/FixedStepsProgressToMigrationProgressConverterTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/FixedStepsProgressToMigrationProgressConverterTests.cs
@@ -1,0 +1,45 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using SonarLint.VisualStudio.ConnectedMode.Binding;
+using SonarLint.VisualStudio.ConnectedMode.Migration;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
+{
+    [TestClass]
+    public class FixedStepsProgressToMigrationProgressConverterTests
+    {
+        [TestMethod]
+        public void Report_ReportsMigrationProgress()
+        {
+            var migrationProgress = new Mock<IProgress<MigrationProgress>>();
+            MigrationProgress migrationProgressReport = null;
+            migrationProgress.Setup(x => x.Report(It.IsAny<MigrationProgress>())).Callback<MigrationProgress>(x => migrationProgressReport = x) ;
+
+
+            var testSubject = new FixedStepsProgressToMigrationProgressConverter(migrationProgress.Object);
+            testSubject.Report(new FixedStepsProgress("test message", 0, 1));
+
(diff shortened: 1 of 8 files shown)


src/ConnectedMode/Migration/FixedStepsProgressToMigrationProgressConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using SonarLint.VisualStudio.ConnectedMode.Binding;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Migration
+{
+    internal class FixedStepsProgressToMigrationProgressConverter : IProgress<FixedStepsProgress>
```
bigfluffycookie, line 26: I feel like the name is long enough to not write any documentation WDYT

src/ConnectedMode/Migration/FixedStepsProgressToMigrationProgressConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+        private readonly IProgress<MigrationProgress> progress;
+
+        public FixedStepsProgressToMigrationProgressConverter(IProgress<MigrationProgress> progress)
+        {
+            this.progress = progress;
+        }
+
+        public void Report(FixedStepsProgress value)
+        {
+            progress?.Report(new MigrationProgress(0, 1, value.Message, false));
```
bigfluffycookie, line 37: Right now we really only use the value.message part which makes this simple to convert.

src/ConnectedMode.UnitTests/Migration/FixedStepsProgressToMigrationProgressConverterTests.cs
```diff
@@ -0,0 +1,45 @@
...
+    [TestClass]
+    public class FixedStepsProgressToMigrationProgressConverterTests
+    {
+        [TestMethod]
+        public void Report_ReportsMigrationProgress()
+        {
+            var migrationProgress = new Mock<IProgress<MigrationProgress>>();
+            MigrationProgress migrationProgressReport = null;
+            migrationProgress.Setup(x => x.Report(It.IsAny<MigrationProgress>())).Callback<MigrationProgress>(x => migrationProgressReport = x) ;
+
```
bigfluffycookie, line 36: white space :(

duncanp-sonar: An extra blank line doesn't really matter. Personally, it isn't the sort of thing I would mention in a review.

src/ConnectedMode/Migration/FixedStepsProgressToMigrationProgressConverter.cs
duncanp-sonar: Yep, I don't think it needs a comment. However, I would change the suffix from `Converter` to `Adapter`, since it's an example of the [adapter pattern](https://en.wikipedia.org/wiki/Adapter_pattern). That would make the name even clearer.

duncanp-sonar:
Looks good! Is this enough to update the binding progress to the wizard UI?

src/ConnectedMode/Migration/IUnintrusiveBindingController.cs
```diff
@@ -18,18 +18,18 @@
...
 using SonarLint.VisualStudio.Core.Binding;
-
 using Task = System.Threading.Tasks.Task;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Migration
 {
     internal interface IUnintrusiveBindingController
     {
-        Task BindAsync(BoundSonarQubeProject project, CancellationToken token);
+        Task BindAsync(BoundSonarQubeProject project, IProgress<MigrationProgress> progress, CancellationToken token);
```
duncanp-sonar, line 32: Actually, this doesn't look quite right. The `UnintrusiveBindingController` is the general-purpose-update-silently-in-the-background that will be used for updating QP unobtrusively
i.e. it is not migration-specific, so it shouldn't know about `IMigrationProcess`.

Wouldn't it be better if:
* the `UnintrusiveBindingController` used the `FixedStepsProgress`
* the `ConnectedModeMigration` class created the adapter and passed it to the `UnobtrusiveBindingController` ?

bigfluffycookie: Yea you are right, I think I reasoned that it is not all purpose because it is in the migration folder however that is bad logic. I'll change it :)

duncanp-sonar: > I think I reasoned that it is not all purpose because it is in the migration folder

Good point. It should be in the `Binding` folder instead. I'll add a note to the board to remind us to move it.

duncanp-sonar (approved):
Loks good

PR #4405, opened 2023-06-20
Update wizard text

duncanp-sonar:
- tweak button behaviour. Show Finish button when migration starts, and enable it when migration finishes.

### DIFF
--- a/src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
+++ b/src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
@@ -71,6 +71,8 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration.Wizard
             migrationInProgress = true;
 
             this.migrateButton.Visibility = Visibility.Collapsed;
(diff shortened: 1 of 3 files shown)


src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -99,7 +101,7 @@ private async Task MigrateAsync()
         private void MigrationFinished()
         {
             migrationInProgress = false;
-            this.finishButton.Visibility = Visibility.Visible;
+            this.finishButton.IsEnabled = true;
```
bigfluffycookie, line 104: Ahh... thats how you grey it out haha, I knew it was possible but somehow didn't think of this

duncanp-sonar: Buttons (and other controls) can update their styles automatically based on the state of the button i.e. based on the value of various properties. Here, the button is changing its style based on the value of `IsEnabled`. Buttons appeared grey-ed out is the default styling behaviour, but we could change that if wanted.

Issue #4407, opened 2023-06-21
Wizard - size to content


duncanp-sonar:
### Description

We should avoid using fixed widths/hieghts in the wizard, otherwise it won't resize correctly if the VS font size changes.

PR #4408, opened 2023-06-21
Change wizard XAML to size to content

duncanp-sonar:
i.e. removed fixed sizes

Fixese #4407

### DIFF
--- a/src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml
+++ b/src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml
@@ -21,7 +21,7 @@
     </Window.Resources>
 
     <Grid TextBlock.FontSize="15">
-        <DockPanel x:Name="StartWindow" Margin="10" LastChildFill="True" Visibility="Hidden" >
+        <DockPanel x:Name="StartWindow" Margin="10" LastChildFill="True" Visibility="Visible" >
             <DockPanel DockPanel.Dock="Bottom">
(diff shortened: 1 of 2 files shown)


duncanp-sonar:
Page 1 now looks like this:
image

... and page 2 like this. Note that the close button is disabled - I think I inadvertently changed this in a previous commit. I'll look at it separately.

image

The dialog will now resize (within limits) as the font size changes e.g.

image

It still needs some tweaking but we can look at it as part of the theming ticket, as that is when we'll get the font size from VS.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml
```diff
@@ -4,51 +4,92 @@
...
-                <Bold><Run>Before starting:</Run></Bold><LineBreak />
-                <Run>* check that you have no unsaved files</Run><LineBreak />
-                <Run>* if the solution is under source control, check that the </Run><LineBreak/>
-                <Run>  you have no uncommitted changes.</Run><LineBreak />
+            <!-- Last child - fills the body of the panel -->
+            <StackPanel Orientation="Vertical" HorizontalAlignment="Left">
 
-                <LineBreak />
+                <!-- Intro -->
+                <TextBlock TextWrapping="WrapWithOverflow" >
```
duncanp-sonar, line 34: I split the text into separate logical sections i.e. separate `TextBlock`s. That gives us more control over the spacing/layout of each block.
However, we might not need it. We should revisit it when we look at theming and how the dialog looks at different themed sizes.

bigfluffycookie (approved):
Very good

Issue #4386, opened 2023-06-19
Implement full file provider


duncanp-sonar:
### Description

Implement a more comprehensive file provider. The basic version only looks for files under the solution folder. This might be ok in most cases, but there will be situations where the projects are not under the solution folder.

#### Files to include
* the C# and VB.NET project files in the solution. In theory, it shouldn't matter too much if we find projects that are not in the solution, unless they happen to be bound to the same Sonar project. In that case, they would be unbound to.
* rulesets, imported files and Directory.Build.prop/targets. Some of these could be at solution-level, or above the solution folder.

#### Notes
If there are no Roslyn projects then there are no additional files to check.

* vbproj, csproj - get from the Roslyn project workspace
* search for rulesets, .props and .targets under the project folders
* search for ruleset, .props and targets under the solution folder
* ? search for ruleset, .props and targets in the folder(s) above the solution folder? If so, how many? Up to the repo root if there is  one?
* ? ditto for the folders above the project folders?

List of files would be de-duplicated.

#### Risks:
1) finding too many files -> slow
Can test against a large repo to see if this is really an issue.

2) finding files not in the solution -> cleaning the wrong files. Some files like props files and Directory.Build.props don't need to be referenced in the solution, or even in a project file in the case of Directory.Build.props

We are looking for a set of Sonar-specific project files so we should only match settings for that specific Sonar project.
We would only "over-match" if there was a repo with more than one solution bound to the same Sonar project.
? can we mitigate this risk?

3) missing files -> not cleaning up fully
We have to handle this situation since it is unlikely we will be able to cover all possible scenarios automatically.
The best we can do in this case is help the user by identifying what still needs to be cleaned up.

PR #4412, opened 2023-06-21
Get list of project to check from RoslynProjectProvider

duncanp-sonar:
- if there are no projects then there is nothing to clean

Part of #4386

### DIFF
--- a/src/ConnectedMode.UnitTests/Migration/FileProviders/XmlFileProviderTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/FileProviders/XmlFileProviderTests.cs
@@ -19,9 +19,12 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.IO.Abstractions;
+using System.Linq;
 using System.Threading;
+using Microsoft.CodeAnalysis;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
@@ -37,11 +40,16 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration.FileProviders
     [TestClass]
     public class XmlFileProviderTests
     {
+        // Non-empty provider used by tests that don't care about the specific projects,
+        // just that there are some. CreateTestSubject will use this provider by default.
+        private static readonly IRoslynProjectProvider NonEmptyRoslynProjectProvider = CreateRoslynProjectProvider("c:\\any\\x.proj").Object;
+
         [TestMethod]
         public void MefCtor_CheckIsExported()
         {
             MefTestHelpers.CheckTypeCanBeImported<XmlFileProvider, IFileProvider>(
                 MefTestHelpers.CreateExport<SVsServiceProvider>(),
+                MefTestHelpers.CreateExport<IRoslynProjectProvider>(),
                 MefTestHelpers.CreateExport<ILogger>(),
                 MefTestHelpers.CreateExport<IThreadHandling>());
         }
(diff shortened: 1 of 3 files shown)


src/ConnectedMode/Migration/RoslynProjectProvider.cs
```diff
@@ -42,6 +42,7 @@ internal class RoslynProjectProvider : IRoslynProjectProvider
 
         [ImportingConstructor]
         public RoslynProjectProvider(VisualStudioWorkspace vsWorkspace)
+            : this(new VisualStudioWorkspaceWrapper(vsWorkspace))
```
duncanp-sonar, line 45: Oopsie - minor problem with the wrapper that I found when debugging...

src/ConnectedMode.UnitTests/Migration/FileProviders/XmlFileProviderTests.cs
```diff
@@ -177,5 +234,42 @@ private static Mock<IVsSolution> CreateIVsSolution(string pathToReturn)
...
+        {
+            var projectProvider = new Mock<IRoslynProjectProvider>();
+
+            var projects = CreateProjects(projectFilePaths);
+            projectProvider.Setup(x => x.Get()).Returns(projects);
+
+            return projectProvider;
+        }
+
+        private static IReadOnlyList<Project> CreateProjects(params string[] projectFilePaths)
```
duncanp-sonar, line 248: These helper methods aren't the same as the ones I wrote for the `RoslynProjectProvider` itself i.e. we can share them.

These tests need to mock the project file path only.

src/ConnectedMode/Migration/FileProviders/XmlFileProvider.cs
```diff
@@ -39,16 +39,19 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration.FileProviders
...
         private readonly IServiceProvider serviceProvider;
+        private readonly IRoslynProjectProvider projectProvider;
         private readonly ILogger logger;
         private readonly IThreadHandling threadHandling;
         private readonly IFileSystem fileSystem;
 
+        // We only search for non-project files - we get the Roslyn project file paths
+        // from the VSWorkspace.
+        // Corner case: this means that we won't clean projects that are in the solution
+        // but unloaded, since they won't appear in the VsWorkspace
```
duncanp-sonar, line 50: That way, there's no risk that we pick up any projects that are not part of the solution.

Also, it means we can skip looking for other types of file to clean up: no Roslyn projects = no files to check for settings.

src/ConnectedMode/Migration/FileProviders/XmlFileProvider.cs
```diff
@@ -115,14 +143,14 @@ private IEnumerable<string> FindFiles(string rootFolder)
 
             var timer = Stopwatch.StartNew();
 
-            var allMatches = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            var allMatches = new List<string>();
```
duncanp-sonar, line 146: I moved the de-duplication using a hashset to the top-level, so we can just use a List here and call List.AddRange.

duncanp-sonar:
I've manually tested the behaviour with Roslyn and no-Roslyn projects. In the no-Roslyn project case, the reset of the migration process still happens. 
We could update progress messages to make it clearer is nothing to clean in that case, although we do log this in the output window.

src/ConnectedMode/Migration/RoslynProjectProvider.cs
bigfluffycookie: Ah, yea also missed that

bigfluffycookie (approved):

Issue #4415, opened 2023-06-22
[Threading] Add synchronous "run on UI thread" method to IThreadHandling


duncanp-sonar:
### Description

The existing RunOnUIThreadSync isn`t synchronous.
Add a new method that is and make the old one obsolete.

FYI there is a separate ticket for fixing the existing method: #4179

PR #4416, opened 2023-06-22
Add an actually synchronous method to IThreadHandling

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4415

### DIFF
--- a/src/Core/IThreadHandling.cs
+++ b/src/Core/IThreadHandling.cs
@@ -61,7 +61,8 @@ namespace SonarLint.VisualStudio.Core
         /// If the caller is not on the main thread then the method will switch to the main thread,
         /// then resume on the caller's thread when then the operation completes.
         /// </summary>
-        [Obsolete]
+        [Obsolete("This method does not work correctly. Use RunOnUIThreadSync2 instead.")]
+        // See https://github.com/SonarSource/sonarlint-visualstudio/issues/4179 for more info.
         void RunOnUIThreadSync(Action op);
 
         /// <summary>
--- a/src/Core/IThreadHandling.cs
+++ b/src/Core/IThreadHandling.cs
@@ -101,6 +101,12 @@ namespace SonarLint.VisualStudio.Core
         /// </summary>
         /// <remarks>Wrapper that calls <see cref="TaskScheduler.Default"/></remarks>
         IAwaitableWrapper SwitchToBackgroundThread();
+        
+        /// <summary>
+        /// Switches to the main thread
+        /// </summary>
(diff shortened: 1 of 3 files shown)


src/TestInfrastructure/NoOpThreadHandler.cs
```diff
@@ -46,10 +46,14 @@ public Task RunOnUIThread(Action op)
 
         public void RunOnUIThreadSync(Action op) => op();
 
+        public void RunOnUIThreadSync2(Action op) => op();
+
         public Task<T> RunOnBackgroundThread<T>(Func<Task<T>> asyncMethod) => asyncMethod();
 
         public IAwaitableWrapper SwitchToBackgroundThread() => new NoOpAwaitable();
 
+        public Task SwitchToMainThreadAsync() => Task.CompletedTask;
```
bigfluffycookie, line 55: This was on a separate commit than the original SwitchToMainThreadAsync add-in was so it was simpler for me to just add in this line. (I made sure it is in the same place.)

bigfluffycookie:
One code smell is about not blocking the thread, I assume I can ignore that as we want to block? Then two of them are about the obsolete, should I have written a comment about why it is obsolete? (I could link to the but ticket or something)

src/Core/IThreadHandling.cs
```diff
@@ -61,8 +61,18 @@ public interface IThreadHandling
         /// If the caller is not on the main thread then the method will switch to the main thread,
         /// then resume on the caller's thread when then the operation completes.
         /// </summary>
+        [Obsolete]
```
duncanp-sonar, line 64: Could you add a comment explaining why this is obsolete and what the alternative is, please? e.g.

```
[Obsolete("This method does not work correctly. Use RunOnUIThreadSync2 instead.")]
// See https://github.com/SonarSource/sonarlint-visualstudio/issues/4179 for more info.
```

I think the message text in the attribute might show up in IntelliSense, but I'm not sure.

src/Infrastructure.VS/ThreadHandling.cs
```diff
@@ -52,6 +53,22 @@ public class ThreadHandling : IThreadHandling
...
+                op();
+                return;
+            }
+            ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                op();
+            });
+
+            Debug.Assert(!ThreadHelper.CheckAccess(), "Not expecting to returning on the UI thread");
```
duncanp-sonar, line 69: Minor: missing word  - "...be returning..."

(or just "...to be on...")

duncanp-sonar (approved):
Looks good. Just a suggestion about commenting the attribute

PR #4422, opened 2023-06-22
Update progress messages

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4421

### DIFF
--- a/src/ConnectedMode/Migration/ConnectedModeMigration.cs
+++ b/src/ConnectedMode/Migration/ConnectedModeMigration.cs
@@ -109,7 +109,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
             var legacySettings = await settingsProvider.GetAsync(oldBinding.ProjectKey);
 
             // TODO: add proper progress messages.
-            progress?.Report(new MigrationProgress(0, 1, "Getting files to clean...", false));
+            progress?.Report(new MigrationProgress(0, 1, "Finding files to clean ...", false));
 
             logger.WriteLine(MigrationStrings.Process_GettingFiles);
             var files = await fileProvider.GetFilesAsync(token);
@@ -139,7 +139,6 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
 
             var progressAdapter = new FixedStepsProgressToMigrationProgressAdapter(progress);
             await unintrusiveBindingController.BindAsync(oldBinding, progressAdapter, token);
-            progress?.Report(new MigrationProgress(0, 1, "Saving new binding files ...", false));
 
             // Note: SLVS will continue to detect the legacy binding mode until this step,
             // so if anything goes wrong during the migration and an exception occurs, the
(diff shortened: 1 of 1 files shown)


src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -130,14 +130,16 @@ private async Task MigrateImplAsync(BoundSonarQubeProject oldBinding, IProgress<
...
                 logger.WriteLine(MigrationStrings.Process_SkippingChecking);
             }
 
-            progress?.Report(new MigrationProgress(0, 1, "Create new binding files ...", false));
+            progress?.Report(new MigrationProgress(0, 1, "Creating new binding files ...", false));
             logger.WriteLine(MigrationStrings.Process_ProcessingNewBinding);
 
             var progressAdapter = new FixedStepsProgressToMigrationProgressAdapter(progress);
             await unintrusiveBindingController.BindAsync(oldBinding, progressAdapter, token);
+            progress?.Report(new MigrationProgress(0, 1, "Saving new binding files ...", false));
```
bigfluffycookie, line 142: To me it feels weird not having this here so I added it but lmk what you think

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -109,7 +109,7 @@ private async Task MigrateImplAsync(BoundSonarQubeProject oldBinding, IProgress<
             var legacySettings = await settingsProvider.GetAsync(oldBinding.ProjectKey);
 
             // TODO: add proper progress messages.
-            progress?.Report(new MigrationProgress(0, 1, "Getting files ...", false));
+            progress?.Report(new MigrationProgress(0, 1, "Getting files to clean...", false));
```
duncanp-sonar, line 112: Maybe `Finding files to clean ...`?

The other messages have a space before the `...` - it looks a bit on the screen, I think.

src/ConnectedMode/Migration/ConnectedModeMigration.cs
duncanp-sonar: I don't think it's needed, but it doesn't make a big difference.
Does it cause the vertical scroll bar to appear? If so, it means the users will always have to scroll to see the last message, which isn't ideal.

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -146,7 +148,7 @@ private async Task MigrateImplAsync(BoundSonarQubeProject oldBinding, IProgress<
             logger.WriteLine(MigrationStrings.Process_DeletingSonarLintFolder);
             await fileSystem.DeleteFolderAsync(legacySettings.LegacySonarLintFolderPath);
 
-            progress?.Report(new MigrationProgress(0, 1, "Finished", false));
+            progress?.Report(new MigrationProgress(0, 1, "Finished!", false));
```
duncanp-sonar, line 151: `Migration finished successfully` ? (with or without the !)

FYI the other possible final messages are:
* `Migration cancelled` and 
* `Error during migration. See the Output Window for more information`

duncanp-sonar (approved):
Looks good. A couple of minor suggestions.

PR #4428, opened 2023-06-23
Remove None ItemGroup that references the generated ruleset

### LINKED ISSUE #4427: [Bug] None ItemGroup references to the ruleset file are not removed
### Description

For legacy VB.NET projects, a `None` ItemGroup reference pointing at the ruleset is add by the legacy binding process, but not removed from during cleanup.

Example:
```
    <None Include="..\SolutionFolder\.sonarlint\duncanp_junkvb.ruleset">
      <Link>duncanp_junkvb.ruleset</Link>
    </None>
```

The Solution Explorer shows the file as missing post-cleanup:
image

duncanp-sonar:
Fix #4427

### DIFF
--- a/src/ConnectedMode.UnitTests/Migration/XmlFileCleanerTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/XmlFileCleanerTests.cs
@@ -29,8 +29,8 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
 {
     // NOTE: in the event of a test failure the actual and expected results are
     // written to files and added to the Test Results i.e. they should be
-    // accessible in the test results pane (and automatically capture by the
-    // Azure DevOps CI build).
+    // accessible at the bottom of the test results pane (and automatically
+    // captured by the Azure DevOps CI build).
 
     [TestClass]
     public class XmlFileCleanerTests
--- a/src/ConnectedMode.UnitTests/Migration/XmlFileCleanerTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/XmlFileCleanerTests.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.IO;
 using System.Threading;
 using SonarLint.VisualStudio.ConnectedMode.Migration;
@@ -26,6 +27,11 @@ using SonarLint.VisualStudio.TestInfrastructure;
 
 namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
 {
+    // NOTE: in the event of a test failure the actual and expected results are
+    // written to files and added to the Test Results i.e. they should be
+    // accessible in the test results pane (and automatically capture by the
+    // Azure DevOps CI build).
+
     [TestClass]
     public class XmlFileCleanerTests
     {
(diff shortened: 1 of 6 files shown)


src/ConnectedMode.UnitTests/Migration/XmlFileCleanerTests.cs
```diff
@@ -157,7 +158,44 @@ public void Clean_RulesetRefsExist_DifferentProjectKey_SettingsAreNotRemoved()
             var testSubject = CreateTestSubject();
 
             var actual = testSubject.Clean(input, settings, CancellationToken.None);
-            actual.Should().Be(XmlFileCleaner.Unchanged);
+            CheckIsUnchanged(actual);
+        }
+
+        [TestMethod]
+        public void Clean_None_CorrectProjectKey_SettingsAreRemoved()
```
duncanp-sonar, line 165: Two new tests, in the same pattern as the existing ones i.e.
1) the first checks the file is cleaned as expected if you use the correct project key, and
2) the second checks nothing is removed if the project key doesn't match.

src/ConnectedMode.UnitTests/Migration/XmlFileCleanerTests.cs
```diff
@@ -26,10 +27,16 @@
 
 namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
 {
+    // NOTE: in the event of a test failure the actual and expected results are
+    // written to files and added to the Test Results i.e. they should be
+    // accessible at the bottom of the test results pane (and automatically
+    // captured by the Azure DevOps CI build).
+
```
duncanp-sonar, line 34: Writing the fix was easy: writing the tests was a pain because the whitespace and indentation in the "cleaned" file have to be exactly right.

So, I wrote some helper methods to simplify seeing why the files are different.
The files are written to disk and added to the tests results, which means you can access the files from the test result window:

image

Clicking on the file name opens it. The `actual` and `expected` files are next to each other on disc so you can easily diff them.

src/ConnectedMode.UnitTests/Migration/XmlFileCleanerTests.cs
```diff
@@ -26,10 +27,17 @@
...
     [TestClass]
     public class XmlFileCleanerTests
     {
         private static readonly LegacySettings AnyLegacySettings = new LegacySettings("c:\\any\\root\\folder", "csharpruleset", "csharpXML", "vbruleset", "vbXML");
+     
+        public TestContext TestContext { get; set; }
```
duncanp-sonar, line 40: MSTest will set this property when running the test. It gives us access to e.g. the test-specific results directory created by MSTest at runtime. See [the docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.testcontext?view=visualstudiosdk-2022) for more info.

src/ConnectedMode.UnitTests/Migration/XmlFileCleanerTests.cs
```diff
@@ -173,5 +217,46 @@ private static string LoadEmbeddedTestCase(string testResourceName)
 
             return stream.ReadToEnd();
         }
+
+        private void CheckAreSame(string actual, string expected)
```
duncanp-sonar, line 221: New helper methods to conditionally dump the results to disc.

bigfluffycookie (approved):
Good

PR #4429, opened 2023-06-24
Don't suppress TaskCanceled exceptions

### LINKED ISSUE #4424: Cancelling the migration causes a null-ref exception
### Description

Cancelling the migration process when a Quality Profile is being downloaded causes the migration to stop, but with a null-ref exception rather than an `OperationCancelled` exception.

#### Partial call stack

The error is happening in `CSharpVBBindingConfigProvider`.

~~Initial guess: the SonarQubeClient service is checking the token but exiting rather than throwing an `OperationCancelledException`. The method that detected the cancellation is presumably returning a null object (rather than a list of results).~~

Cause: `WebServiceHelper.SafeServiceCallAsync` was suppressing the cancellation exception so the callers were not aware the method had failed -> continued processing -> null ref exception throw.

```
   at System.Linq.Enumerable.Where[TSource](IEnumerable`1 source, Func`2 predicate)
   at SonarLint.VisualStudio.ConnectedMode.Binding.CSharpVBBindingConfigProvider.<FetchSupportedRulesAsync>d__13.MoveNext() in D:\repos\sq\slvs\src\ConnectedMode\Binding\CSharpVBBindingConfigProvider.cs:line 145
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at SonarLint.VisualStudio.ConnectedMode.Binding.CSharpVBBindingConfigProvider.<DoGetConfigurationAsync>d__9.MoveNext() in D:\repos\sq\slvs\src\ConnectedMode\Binding\CSharpVBBindingConfigProvider.cs:line 86
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotif
```

duncanp-sonar:
Fixes #4424

### DIFF
--- a/src/Core.UnitTests/Helpers/WebServiceHelperTests.cs
+++ b/src/Core.UnitTests/Helpers/WebServiceHelperTests.cs
@@ -90,28 +90,30 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Helpers
         }
 
         [TestMethod]
-        public async Task TaskIsCancelled_ExceptionIsSuppressed()
+        public async Task TaskIsCanceled_ExceptionIsNotSuppressed()
         {
             // Arrange
             var testLogger = new TestLogger();
+            Func<Task> operation = () => WebServiceHelper.SafeServiceCallAsync(
+                () => throw new TaskCanceledException("should not be suppressed"), testLogger);
 
             // Act
-            await WebServiceHelper.SafeServiceCallAsync<bool>(() => throw new TaskCanceledException("dummy error message"),
-                testLogger);
+            await operation.Should().ThrowAsync<TaskCanceledException>();
 
             // Assert
             testLogger.AssertOutputStringExists(CoreStrings.SonarQubeRequestTimeoutOrCancelled);
         }
 
         [TestMethod]
-        public async Task OperationIsCancelled_ExceptionIsSuppressed()
+        public async Task OperationIsCanceled_ExceptionIsNotSuppressed()
         {
             // Arrange
             var testLogger = new TestLogger();
+            Func<Task> operation = () => WebServiceHelper.SafeServiceCallAsync(
+                () => throw new OperationCanceledException("should not be suppressed"), testLogger);
 
             // Act
-            await WebServiceHelper.SafeServiceCallAsync<bool>(() => throw new OperationCanceledException("dummy error message"),
-                testLogger);
+            await operation.Should().ThrowAsync<OperationCanceledException>();
 
             // Assert
             testLogger.AssertOutputStringExists(CoreStrings.SonarQubeRequestTimeoutOrCancelled);
--- a/src/Core.UnitTests/Helpers/WebServiceHelperTests.cs
+++ b/src/Core.UnitTests/Helpers/WebServiceHelperTests.cs
@@ -31,7 +31,7 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Helpers
     public class WebServiceHelperTests
     {
         [TestMethod]
-        async public Task CallSucceeds()
+        public async Task CallSucceeds()
         {
             // Arrange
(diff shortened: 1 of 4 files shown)


duncanp-sonar:
I suggest reviewing the two commits separately. The first is CAYC to clean up the tests, the second is the fix.

src/Core.UnitTests/Helpers/WebServiceHelperTests.cs
```diff
@@ -31,7 +31,7 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Helpers
     public class WebServiceHelperTests
     {
         [TestMethod]
-        async public Task CallSucceeds()
+        public async Task CallSucceeds()
```
duncanp-sonar, line 34: Minor: by convention we put `async` after the scope.

src/Core.UnitTests/Helpers/WebServiceHelperTests.cs
```diff
@@ -46,102 +46,103 @@ async public Task CallSucceeds()
...
         [TestMethod]
-        public void HttpRequestException_NoInnerException_IsHandledAndLogged()
+        public async Task HttpRequestException_NoInnerException_IsHandledAndLogged()
         {
             // Arrange
             var testLogger = new TestLogger();
 
             // Act
-            WebServiceHelper.SafeServiceCallAsync<bool>(() => throw new HttpRequestException("outer message"),
-                testLogger).Wait();
```
duncanp-sonar, line 56: We shouldn't need to use `Wait()` in tests - we can just make the test `async Task` and `await`.

src/Core/Helpers/WebServiceHelper.cs
```diff
@@ -40,15 +40,14 @@ public static async Task<T> SafeServiceCallAsync<T>(Func<Task<T>> call, ILogger
                 var innerException = e.InnerException as System.Net.WebException;
                 logger.WriteLine(CoreStrings.SonarQubeRequestFailed, e.Message, innerException?.Message);
             }
-            catch (TaskCanceledException)
```
duncanp-sonar, line 43: Miscellaneous: `TaskCanceledException` is a subclass of `OperationCanceledException` so there's no need to specify it separately - it will be caught by the `catch(OperationCanceledException)`.

src/Core/Helpers/WebServiceHelper.cs
```diff
@@ -40,15 +40,14 @@ public static async Task<T> SafeServiceCallAsync<T>(Func<Task<T>> call, ILogger
...
-                logger.WriteLine(CoreStrings.SonarQubeRequestTimeoutOrCancelled);
-            }
             catch (OperationCanceledException)
             {
                 // Canceled or timeout
                 logger.WriteLine(CoreStrings.SonarQubeRequestTimeoutOrCancelled);
+
+                // Re-throw the exception so that methods further up the stack know the task was cancelled
+                // See bug https://github.com/SonarSource/sonarlint-visualstudio/issues/4424
+                throw;
```
duncanp-sonar, line 50: This is the actual fix.

duncanp-sonar:
This fix has no effect on the UX of the cancellation a binding that was triggered from the Team Explorer. It still looks like this:
image

It fixes the migration binding cancellation so we get the correct error message:
image

duncanp-sonar:
I'm not sure why SonarCloud is complaining about test coverage. It's showing the whole `WebServiceHelper` class as not being covered in the PR, but there are clearly tests for it, and the coverage in the main branch is showing as 100% for that class.
Either way, I wouldn't let it prevent merging this change.

bigfluffycookie (approved):

PR #4431, opened 2023-06-24
Handle help hyperlink in wizard

### LINKED ISSUE #4430: Handle "Learn more" link in wizard
### Description

Navigate to wiki help page when the user clicks on `Learn more`

duncanp-sonar:
Fixes  #4430

### DIFF
--- a/src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
+++ b/src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
@@ -22,6 +22,7 @@ using System;
 using System.ComponentModel.Composition;
 using SonarLint.VisualStudio.Integration;
 using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Core;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Migration.Wizard
 {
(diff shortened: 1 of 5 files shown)


src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml
```diff
@@ -38,7 +38,7 @@
                         and the settings for this project need to be migrated to the new format.
                     </Run>
                     <LineBreak />
-                    <Run>See </Run> <Hyperlink> here</Hyperlink> <Run> for more information.</Run>
+                    <Run>See </Run> <Hyperlink NavigateUri="http://any" RequestNavigate="OnNavigateToHelp"> here</Hyperlink> <Run> for more information.</Run>
```
duncanp-sonar, line 41: We have to supply a URL in the XAML otherwise the `OnNavigateToHelp` method won't be called. However, we can't easily reference the `MigrationString.LearnMoreUrl` value from the resource file.

So, we can just use a dummy URL. All the wizard does is forward the event on to the controller to handle so it doesn't actually need the URL.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -182,5 +187,8 @@ private void btnPage2_Cancel_Click(object sender, RoutedEventArgs e)
             btnPage2_Cancel.IsEnabled = false;
             cancellationTokenSource.Cancel();
         }
+
+        private void OnNavigateToHelp(object sender, System.Windows.Navigation.RequestNavigateEventArgs e)
+            => onShowHelp?.Invoke();
```
duncanp-sonar, line 192: Just forward the caller to the controller.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -46,12 +47,16 @@ public sealed partial class MigrationWizardWindow : DialogWindow, IProgress<Migr
 
         private CancellationTokenSource cancellationTokenSource;
 
-        internal MigrationWizardWindow(BoundSonarQubeProject oldBinding, IConnectedModeMigration connectedModeMigration, ILogger logger)
+        internal MigrationWizardWindow(BoundSonarQubeProject oldBinding,
+            IConnectedModeMigration connectedModeMigration,
+            Action onShowHelp,
```
duncanp-sonar, line 52: I could have passed the `IBrowserService` to the wizard but it doesn't really need to know how to handle the request. Instead, an `Action` will do just as well, and we can put the logic for handling the call in the controller.

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
```diff
@@ -45,25 +46,31 @@ internal class MigrationWizardController : IMigrationWizardController
...
         {
-            var migrationWizardWindow = new MigrationWizardWindow(oldBinding, connectedModeMigration, logger);
+            var migrationWizardWindow = new MigrationWizardWindow(oldBinding, connectedModeMigration, OnShowHelp, logger);
 
             var finishedSuccessfully = migrationWizardWindow.ShowModal();
 
             if (finishedSuccessfully != null && finishedSuccessfully.Value)
             {
-               MigrationWizardFinished?.Invoke(this, EventArgs.Empty);
+                MigrationWizardFinished?.Invoke(this, EventArgs.Empty);
```
duncanp-sonar, line 70: ?Auto-formatting by VS? Not something I changed on purpose.

PR #4433, opened 2023-06-26
Fix spacing of bullets in wizard

duncanp-sonar:
Fit and finish: fix the spacing of the text next to the bullet points

### DIFF
--- a/src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml
+++ b/src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml
@@ -22,7 +22,7 @@
 
     <Grid TextBlock.FontSize="15">
         <Separator VerticalAlignment="Top" />
-        <DockPanel x:Name="StartWindow" Margin="10" LastChildFill="True" Visibility="Visible" >
+        <DockPanel x:Name="StartWindow" Margin="20" LastChildFill="True" Visibility="Visible" >
             <DockPanel DockPanel.Dock="Bottom">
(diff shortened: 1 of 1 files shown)


duncanp-sonar:
The updated UI looks like this:

image

The second line of text under the second bullet point is now aligned correctly.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml
```diff
@@ -49,22 +50,40 @@
...
+                        <Run> check that you have no unsaved files</Run>
+                    </TextBlock>
+                    
+                    <BulletDecorator Grid.Column="0" Grid.Row="1" VerticalAlignment="Center">
                         <BulletDecorator.Bullet>
                             <Ellipse Height="8" Width="8" Fill="Black" />
                         </BulletDecorator.Bullet>
                     </BulletDecorator>
-                                    
-                    <Run> if the solution is under source control, check that you have no uncommitted changes.</Run>
```
duncanp-sonar, line 66: Previously the bullets and text were next to each other in the same horizontal stack panel.
That meant that when the text wrapped it appear under the bullet, rather than aligned under the text.

To fix it, I created a grid with three rows and two columns, with the long line of text for the second bullet point spanning two rows.

If I had used a 2 x 2 grid, the alignment of the second bullet point would not have been correct - it would have been centred vertically between the two lines of text.

bigfluffycookie (approved):
Very nice! The ui is really coming together.

Issue #4387, opened 2023-06-19
Implement full IVsAwareFileSystem


duncanp-sonar:
### Description

The current implementation is not VS-aware, and just passes 
calls through the file system.

Three separate areas, in priority order:
* source-code control - notify VS source-controlled files are being modified. Important for a minority of users who are not using git for source control. The old binding method correctly handled this situation.
* reading file contents from in-memory: this would prevent conflicts/prompts when VS notices we've changed the file on disc outside the solution. However, the migration guidance to users is to start from a clean slate before migrating i.e. no unsaved files, and no outstanding source-control changes.
* notify VS that we have changed a file outside the solution so it does not prompt the user.

PR #4443, opened 2023-06-30
Preparation for adding SCC operations to IVsAwareFileSystem

duncanp-sonar:
Prep for #4387


src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -115,18 +115,12 @@ private async Task MigrateImplAsync(BoundSonarQubeProject oldBinding, IProgress<
             var files = await fileProvider.GetFilesAsync(token);
             logger.WriteLine(MigrationStrings.Process_CountOfFilesToCheck, files.Count());
 
+            ChangedFiles changedFiles = null;
```
duncanp-sonar, line 118: When we interact with the source control system via the VS APIs, we do it as follows:
* tell VS we are starting a batch of operations
* check out and modify one or more files
* tell VS we have finished the batch.

I've moved all of the file-changing operations to a single method to make it simpler to put them all inside a batch (the batching operations are added in a later commit).

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -194,9 +185,38 @@ private System.Threading.Tasks.Task<string> GetFileContentAsync(string filePath)
             return changedFiles;
         }
 
-        private async Task SaveChangedFilesAsync(ChangedFiles changedFiles)
+        private async Task MakeLegacyFileChangesAsync(LegacySettings legacySettings, ChangedFiles changedFiles,
```
duncanp-sonar, line 188: Logically the code is the same as before i.e. I haven't changed what it does, just the order it does it in.

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -188,11 +188,20 @@ private System.Threading.Tasks.Task<string> GetFileContentAsync(string filePath)
...
+                // Note: no files have been changed yet. Now we are going to start making changes
+                // to the user's projects and deleting files that might be under source control...
+                await SaveChangedFilesAsync(changedFiles, progress);
+
+                await DeleteSonarLintFolderAsync(legacySettings, progress);
+            }
+            finally
+            {
+                await fileSystem.EndChangeBatchAsync();
+            }
```
duncanp-sonar, line 204: Add the batch operations to the interface. They don't do anything at this point.
We've closing the batch in a `finally` block to make sure it happens even if there are exceptions.

src/ConnectedMode/Migration/VsAwareFileSystem.cs
```diff
@@ -30,18 +34,28 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
...
+        private readonly IThreadHandling threadHandling;
+        private IVsQueryEditQuerySave2 queryFileOperation;
 
         [ImportingConstructor]
-        public VsAwareFileSystem(ILogger logger)
-            : this(logger, new FileSystem())
+        public VsAwareFileSystem([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider,
+            ILogger logger,
+            IThreadHandling threadHandling)
+            : this(serviceProvider, logger, threadHandling, new FileSystem())
```
duncanp-sonar, line 47: Adding the imports that will be required to get source control service.

src/ConnectedMode/Migration/VsAwareFileSystem.cs
```diff
@@ -68,16 +82,28 @@ public Task SaveAsync(string filePath, string text)
             return Task.CompletedTask;
         }
 
-        public Task BeginChangeBatchAsync()
+        public async Task BeginChangeBatchAsync()
         {
             logger.LogMigrationVerbose("Beginning batch of file changes...");
-            return Task.CompletedTask;
+
+            queryFileOperation = await GetSccServiceAsync();
```
duncanp-sonar, line 89: Fetching the service, but not doing anything with it yet.

duncanp-sonar:
Tests will be added when the rest of the implementation is done.

duncanp-sonar:
As discussed offline, I'll merge this now and move on to the next part. I'll make any required changes later.

PR #4445, opened 2023-06-30
Add SCC operations to VsAwareFileSystem

### LINKED ISSUE #4387: Implement full IVsAwareFileSystem

duncanp-sonar:
Fixes #4387

### DIFF
--- a/src/ConnectedMode.UnitTests/Migration/VsAwareFileSystemTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/VsAwareFileSystemTests.cs
@@ -19,6 +19,7 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.IO.Abstractions;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
@@ -98,6 +99,24 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
         }
 
         [TestMethod]
+        public async Task Save_SccServiceIsCalledOnUIThread()
+        {
+            var calls = new List<string>();
+
+            var threadHandling = CreateThreadHandlingWithRunOnUICallback(() => calls.Add("RunOnUIThread"));
+            var testSubject = CreateTestSubject(out var _, out var sccService, threadHandling.Object);
+
+            sccService.SetupQueryEditResponse(tagVSQueryEditResult.QER_EditOK, new[] { "any" }, () => calls.Add("QueryEdit"));
+            await testSubject.BeginChangeBatchAsync();
+            calls.Clear();
+
+            // Act
+            await testSubject.SaveAsync("any", "any");
+
+            calls.Should().ContainInOrder("RunOnUIThread", "QueryEdit");
+        }
+
+        [TestMethod]
         [DataRow(tagVSQueryEditResult.QER_NoEdit_UserCanceled, true)]
         [DataRow(tagVSQueryEditResult.QER_EditNotOK, true)]
         [DataRow(tagVSQueryEditResult.QER_EditOK, false)]
@@ -106,7 +125,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
             var testSubject = CreateTestSubject(out var fileSystem, out var sccService);
 
             fileSystem.SetupFile("file1");
-            sccService.SetupQueryEditResponse(verdict, "file1");
+            sccService.SetupQueryEditResponse(verdict, new[] { "file1" });
 
             await testSubject.BeginChangeBatchAsync(); // all modifications are expected to be in a batch
             Func<Task> act = () => testSubject.SaveAsync("file1", "c1");
@@ -146,8 +165,8 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
             var testSubject = CreateTestSubject(out var fileSystem, out var sccService);
             fileSystem.SetupGetFilesInDir(dirName, "f1", "f2", "f3");
 
-            sccService.SetupQueryEditResponse(tagVSQueryEditResult.QER_EditOK, dirName);
-            sccService.SetupQuerySavesResponse(tagVSQuerySaveResult.QSR_SaveOK, "f1", "f2", "f3");
+            sccService.SetupQueryEditResponse(tagVSQueryEditResult.QER_EditOK, new[] { dirName });
+            sccService.SetupQuerySavesResponse(tagVSQuerySaveResult.QSR_SaveOK, new[] { "f1", "f2", "f3" });
 
             await testSubject.BeginChangeBatchAsync(); // all modifications are expected to be in a batch
             await testSubject.DeleteFolderAsync(dirName);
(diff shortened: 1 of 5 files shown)


src/ConnectedMode/Migration/VsAwareFileSystem.cs
```diff
@@ -58,52 +61,148 @@ internal class VsAwareFileSystem : IVsAwareFileSystem
...
+            }
+
+            if (!success)
+            {
+                var message = string.Format(MigrationStrings.VsFileSystem_Error_FailedToCheckOutFolderForDeletion, folderPath);
+                throw new InvalidOperationException(message);
+            }
+        }
+
+        private async Task<bool> CheckoutForEditAsync(params string[] fileNames)
```
duncanp-sonar, line 148: This was copied from the [equivalent method](https://github.com/SonarSource/sonarlint-visualstudio/blob/03123f51a5f22aa2bd6e347f00a33ca01e9ee44b/src/Integration/LocalServices/SourceControlledFileSystem.cs#L133) of the existing `SourceControlledFileSystem` class.

I've tweaked it slightly to make it `async` so we can force the SCC operation to happen on the UI thread (VS crashes if it is called on a background thread),

The legacy `SourceControlledFileSystem` implementation assumed that the caller would ensure that it was called on the main thread.

src/ConnectedMode/Migration/VsAwareFileSystem.cs
```diff
@@ -58,52 +61,148 @@ internal class VsAwareFileSystem : IVsAwareFileSystem
...
+        {
+            bool result = false;
+
+            await threadHandling.RunOnUIThread(() =>
+                result = DoCheckoutForSave(fileNames));
+
+            return result;
+        }
+
+        private bool DoCheckoutForSave(params string[] fileNames)
```
duncanp-sonar, line 185: Copied the existing [method](https://github.com/SonarSource/sonarlint-visualstudio/blob/03123f51a5f22aa2bd6e347f00a33ca01e9ee44b/src/Integration/LocalServices/SourceControlledFileSystem.cs#L158).

I didn't change this method to handle threading; it was simpler to add a thread-aware wrapper method instead.

src/ConnectedMode/Migration/VsAwareFileSystem.cs
```diff
@@ -58,52 +61,148 @@ internal class VsAwareFileSystem : IVsAwareFileSystem
...
 
-        public Task SaveAsync(string filePath, string text)
+        public async Task SaveAsync(string filePath, string text)
         {
-            // TODO - error handling
+            Debug.Assert(batchStarted, "Expecting the changes to happen in a batch");
+
             logger.LogMigrationVerbose($"Saving file: {filePath}");
+
+            bool success = await CheckoutForEditAsync(filePath);
```
duncanp-sonar, line 88: Before we save a file we need to tell VS we are going to check it out.
If we can't check it out we will throw, which will cause the migration to end.

src/ConnectedMode/Migration/VsAwareFileSystem.cs
```diff
@@ -58,52 +61,148 @@ internal class VsAwareFileSystem : IVsAwareFileSystem
...
 
         public async Task BeginChangeBatchAsync()
         {
+            Debug.Assert(!batchStarted, "Not expecting to already be in a batch");
+
             logger.LogMigrationVerbose("Beginning batch of file changes...");
 
-            queryFileOperation = await GetSccServiceAsync();
+            sccService = await GetSccServiceAsync();
+            sccService.BeginQuerySaveBatch();
```
duncanp-sonar, line 106: Simple pass-through to the VS API.

src/ConnectedMode/Migration/VsAwareFileSystem.cs
```diff
@@ -58,52 +61,148 @@ internal class VsAwareFileSystem : IVsAwareFileSystem
...
             });
 
             return sccService;
         }
+
+        private async Task CheckOutFilesToDeleteAsync(string folderPath)
+        {
+            // Can't check out the directory for saving - if you do, VS will pop up a dialogue,
+            // even if the "silent" flag is set.
+            bool success = await CheckoutForEditAsync(folderPath);
```
duncanp-sonar, line 133: Tell VS we are going to "change" the directory...

src/ConnectedMode/Migration/VsAwareFileSystem.cs
```diff
@@ -58,52 +61,148 @@ internal class VsAwareFileSystem : IVsAwareFileSystem
...
+
+            if (success)
+            {
+                var files = fileSystem.Directory.GetFiles(folderPath, "*.*", System.IO.SearchOption.AllDirectories);
+                success = await CheckoutForSaveAsync(files);
```
duncanp-sonar, line 138: Telling VS that we are going to edit/save the `.sonarlint` folder doesn't appear to be enough on its own - we have to tell VS we are going to change each file under the folder.

If we don't, the deleted files don't appear as `Pending changes` in the `Solution Explorer`.

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -136,7 +136,7 @@ private async Task MigrateImplAsync(BoundSonarQubeProject oldBinding, IProgress<
 
             // Now make all of the files changes required to remove the legacy settings
             // i.e. update project files and delete .sonarlint folder
-            await MakeLegacyFileChangesAsync(legacySettings, changedFiles, progress);
+            await MakeLegacyFileChangesAsync(legacySettings, changedFiles, progress, token);
```
duncanp-sonar, line 139: Passing the token down so we can check for cancellation.

duncanp-sonar:
I've already merged the first four commits since they were simple refactoring.
This PR covers the last two - adding the SCC operations to the VSAwareFileSystem class.

duncanp-sonar:
I'll fix the merge conflicts tomorrow.

src/ConnectedMode.UnitTests/Migration/VsAwareFileSystemTests.cs
```diff
@@ -18,11 +18,17 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
```
bigfluffycookie, line 20: We use threadhandling 3 times in the actual class and I think we should have some tests covering that it runs on the expected thread

bigfluffycookie (commented):
Looks good but maybe some missing test cases

src/ConnectedMode.UnitTests/Migration/VsAwareFileSystemTests.cs
duncanp-sonar: @bigfluffycookie I've added threading unit tests for the two cases where we use the VS interface (the calls that crash VS if we are not on the UI thread).

duncanp-sonar: (and I fixed the code smell)

PR #4454, opened 2023-07-03
Show Tfvc info in wizard conditionally

### LINKED ISSUE #4452: Only show Tfvc text on wizard page one if the user is not using git
### Description

Hide the text on the first wizard page if the user is not using git.

We don't have a component that detects if the user is using Tfvc. However, we do have a component that returns the current branch if the user is using git. We could use that to determine if the user is using git, and not show the Tfvc text in that case.

duncanp-sonar:
Fixes #4452

### DIFF
--- a/src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
+++ b/src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
@@ -48,23 +48,28 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration.Wizard
         private readonly IConnectedModeMigration connectedModeMigration;
         private readonly IBrowserService browserService;
         private readonly IOutputWindowService outputWindowService;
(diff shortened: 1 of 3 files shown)


src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
```diff
@@ -79,6 +84,29 @@ public void StartMigrationWizard(BoundSonarQubeProject oldBinding)
             }
         }
 
+        private Action GetTfVcHelpAction()
+        {
+            Action helpOp;
+
+            // We don't have an existing component that detects whether user is using Tfvc.
+            // However, we can tell if they are using git, which we expect to be the majority
+            // of users. So to reduce the noise, we won't show the Tfvc help for git users.
```
duncanp-sonar, line 93: This comment says it all: quick and dirty hack to reduce the noise for the majority of users.

bigfluffycookie (approved):

Issue #4173, opened 2023-05-03
Bump minimum supported VS version to v16.7/16.8


duncanp-sonar:
### Description

Global analyzer config files are only supported from VS16.7 (se [here](https://github.com/dotnet/roslyn/issues/47707)).

TODO
- [x] check the minimum version - 16.7 or 16.8? (Roslyn ticket above says 16.7, local experiments with `globalconfig` suggest 16.8) **UPDATE: 16.7. See comment below.**
- [x] update VS manifests (VS2019 ~~and VS2022~~)
- [x] check VS marketplace page - talk to PMM if it needs to be updated/updated the marketplace description file in the repo
- [ ] check with Docs team - any updates required?

Update: testing with a 16.7 VM shows that global configs are picked up

duncanp-sonar:
Global config files work with VS16.7:

![image](https://github.com/SonarSource/sonarlint-visualstudio/assets/10757559/2168a904-fea1-470d-84e7-416113d8cd74)

PR #4457, opened 2023-07-04
Increase minimum vs version in manifest

bigfluffycookie:
fixes part of https://github.com/SonarSource/sonarlint-visualstudio/issues/4173


bigfluffycookie:
Is this enough? In the 2022 manifest the min is already >17 and in previous updates of the manifest this was the only change made

duncanp-sonar:
> Is this enough? In the 2022 manifest the min is already >17 and in previous updates of the manifest this was the only change made

Yes, it should be. I'm not sure why I was thinking we'd need to update both manifests. I've edited the ticket to say it's just VS2019.

duncanp-sonar (approved):

PR #4460, opened 2023-07-04
Don't show generated SonarLint.xml in the Solution Explorer

### LINKED ISSUE #4459: Generated SonarLint.xml file is visible in the Solution Explorer
### Description

The generated file appears in the `Solution Explorer`. This is undesirable for a couple of reasons:
* it appears in all VB.NET/C# projects, adding clutter
* double-clicking on the file opens it for editing, which isn't ideal since it is a generated file that should not be manually edited.

Note: the generated global analyzer config files do not appear in the `Solution Explorer`; only the `SonarLInt.xml` files.

duncanp-sonar:
Fixes  #4459

### DIFF
--- a/src/ConnectedMode/Install/SonarLintTargets.xml
+++ b/src/ConnectedMode/Install/SonarLintTargets.xml
@@ -25,6 +25,9 @@
 
     <ItemGroup Condition="$(BuildingInsideVisualStudio) == 'true' AND Exists($(_SLVSGeneratedRoslynConfigFile))">
         <EditorConfigFiles Condition="Exists($(_SLVSGeneratedRoslynConfigFile))" Include="$(_SLVSGeneratedRoslynConfigFile)" />
-        <AdditionalFiles Condition="Exists($(_SLVSGeneratedAdditionalFile))" Include="$(_SLVSGeneratedAdditionalFile)" />
+        <AdditionalFiles Condition="Exists($(_SLVSGeneratedAdditionalFile))" Include="$(_SLVSGeneratedAdditionalFile)">
+            <!-- Don't show in the Solution Explorer -->  
+            <Visible>false</Visible>
+        </AdditionalFiles>
     </ItemGroup>
 </Project>


duncanp-sonar:
Before:
image

After:
image

src/ConnectedMode/Install/SonarLintTargets.xml
```diff
@@ -25,6 +25,9 @@
 
     <ItemGroup Condition="$(BuildingInsideVisualStudio) == 'true' AND Exists($(_SLVSGeneratedRoslynConfigFile))">
         <EditorConfigFiles Condition="Exists($(_SLVSGeneratedRoslynConfigFile))" Include="$(_SLVSGeneratedRoslynConfigFile)" />
-        <AdditionalFiles Condition="Exists($(_SLVSGeneratedAdditionalFile))" Include="$(_SLVSGeneratedAdditionalFile)" />
+        <AdditionalFiles Condition="Exists($(_SLVSGeneratedAdditionalFile))" Include="$(_SLVSGeneratedAdditionalFile)">
+            <!-- Don't show in the Solution Explorer -->  
+            <Visible>false</Visible>
```
duncanp-sonar, line 30: VS uses this metadata property to show/hide the element.

src/ConnectedMode/Install/SonarLintTargets.xml
```diff
@@ -25,6 +25,9 @@
 
     <ItemGroup Condition="$(BuildingInsideVisualStudio) == 'true' AND Exists($(_SLVSGeneratedRoslynConfigFile))">
         <EditorConfigFiles Condition="Exists($(_SLVSGeneratedRoslynConfigFile))" Include="$(_SLVSGeneratedRoslynConfigFile)" />
```
duncanp-sonar, line 27: FYI the analyzer config files don't seem to appear in the `SolutionExplorer` even if you add `<Visible>true</Visible>`.

bigfluffycookie (approved):

PR #4477, opened 2023-07-06
Add ref to ALM binary to tests

duncanp-sonar:
- fixes merge issue (builds ok locally but not on CI)

### DIFF
--- a/src/ConnectedMode.UnitTests/ConnectedMode.UnitTests.csproj
+++ b/src/ConnectedMode.UnitTests/ConnectedMode.UnitTests.csproj
@@ -9,7 +9,12 @@
 
     <RequiresLibGit>true</RequiresLibGit>
   </PropertyGroup>
-  
+
+  <PropertyGroup Label="Specify which external/VS-specific refs are required">
+    <RequiresAlmAuthentication>true</RequiresAlmAuthentication>
+  </PropertyGroup>
+  <Import Project="..\SonarLint.VSSpecificAssemblies.props" />
+
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


src/ConnectedMode.UnitTests/ConnectedMode.UnitTests.csproj
```diff
@@ -9,7 +9,12 @@
 
     <RequiresLibGit>true</RequiresLibGit>
   </PropertyGroup>
-  
+
+  <PropertyGroup Label="Specify which external/VS-specific refs are required">
+    <RequiresAlmAuthentication>true</RequiresAlmAuthentication>
+  </PropertyGroup>
+  <Import Project="..\SonarLint.VSSpecificAssemblies.props" />
```
duncanp-sonar, line 16: Weird; the merge PR build completed successfully and it built ok locally.
However, the post-merge build failed for this test project. Referencing the ALM project directly fixes the issue.

Thinking about it, it's possible that the post-merge build is failing because it's a signed build, whereas the local and pre-merge CI builds are unsigned builds.

The ALM assembly is treated slightly differently in release builds as it needs to be signed using StrongNamer. That might account for the difference.

src/ConnectedMode.UnitTests/packages.lock.json
```diff
@@ -139,14 +154,6 @@
         "resolved": "2.2.85",
         "contentHash": "YptRsDCQK35K5FhmZ0LojW4t8I6DpetLfK5KG8PVY2f6h7/gdyr8f4++xdSEK/xS6XX7/GPvEpqszKVPksCsiQ=="
       },
-      "Microsoft.Alm.Authentication": {
-        "type": "Transitive",
```
duncanp-sonar, line 143: These are now direct references rather than transitive references.

PR #4479, opened 2023-07-06
Change Tfvc migration help URL

duncanp-sonar:
- changed to a section on the migration page rather than a separate page.


duncanp-sonar:
I'm currently drafting the additional migration help for Tfvc users. It will be a section on the main migration wiki page, rather than a separate page:

image

bigfluffycookie (approved):

Issue #4480, opened 2023-07-06
C#/VB.NET issues are not suppressed in unintrusive Connected Mode


duncanp-sonar:
### Description

C# and VB.NET issues are not suppressed correctly in the new unintrusive Connected Mode.

Affects v7 pre-release.

### Repro
1. Open a bound solution
2. Open a file with suppressed C# issues

### Expected
The issues are not visible in the `Error List`.

### Actual
The issues are visible in the `Error List`.

#### Investigation
The problem is that naming and location of the binding files has changed - see #4188.

There are two parts to the fix:
1) the `SuppressionExecutionContext ` needs [to be updated](https://github.com/SonarSource/sonarlint-visualstudio/blob/5cd8033fbb2cde08c631cfaaf5851fabcfa11f16/src/Roslyn.Suppressions/Roslyn.Suppressions/SuppressionExecutionContext.cs#L34) to look in a different location

2) the `RoslynSettingsFileStorage` needs to be [updated](https://github.com/SonarSource/sonarlint-visualstudio/blob/5cd8033fbb2cde08c631cfaaf5851fabcfa11f16/src/Roslyn.Suppressions/Roslyn.Suppressions/SettingsFile/RoslynSettingsFileStorage.cs#L104) to write the file using the solution name rather than the `Sonar` project key.

Issue #4482, opened 2023-07-06
[Cleanup] Extend ISolutionInfoProvider to return the solution directory


duncanp-sonar:
### Description

Several existing classes use `IVsSolutionProvider` directly to get the current solution directory.

The `ISolutionInfoProvider` abstraction returns the current solution name; extend it to return the current solution directory, and refactor the existing code to use the new method.

duncanp-sonar:
Probably not really required - the existing method on the interface returns the full file path to the solution file, so extracting the solution directory is straightforward. It could perhaps be added as an extension method.

duncanp-sonar:
Obsolete - as described above, the interface returns the full file path, so adding an interface method to return just the directory part isn't really necessary.

Issue #4481, opened 2023-07-06
[Cleanup] Refactor existing code to use ISolutionInfoProvider


duncanp-sonar:
### Description

Several existing classes use `IVsSolution` directly to get the current solution name.
Refactor them to use the `ISolutionInfoProvider` abstraction instead.

TODO
 - [x] change `IMigrationPrompt` to use the async method
 - [x] add a non-async method for existing code that can't easily be made synchronous
 - [x] change legacy path providers to use `ISolutionInfoProvider`
 - [x] add interface methods to return the solution directory
 - [x] change remaining legacy components to use `ISolutionInfoProvider`: `XmlFileProvider`, `MigrationSettingsProvider`

PR #4483, opened 2023-07-06
Add ISolutionInfoProvider

duncanp-sonar:
Prep for fixing #4480

### DIFF
--- a/src/ConnectedMode/SolutionInfoProvider.cs
+++ b/src/ConnectedMode/SolutionInfoProvider.cs
@@ -1,0 +1,72 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.ComponentModel.Composition;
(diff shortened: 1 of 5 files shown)


src/ConnectedMode/SolutionInfoProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+    /// <summary>
+    /// Returns information about the current solution if there is one
+    /// </summary>
+    internal interface ISolutionInfoProvider
+    {
+        /// <summary>
+        /// Returns the name of the current solution, or null if there is not a current solution
+        /// </summary>
+        Task<string> GetSolutionNameAsync();
```
duncanp-sonar, line 38: We need the current solution name to save the Roslyn suppressions file in the correct place. This PR adds an abstraction to fetch it.

We have very similar code in several other places that uses `IVsSolution` e.g. [MigrationSettingsProvider](https://github.com/SonarSource/sonarlint-visualstudio/blob/5cd8033fbb2cde08c631cfaaf5851fabcfa11f16/src/ConnectedMode/Migration/MigrationSettingsProvider.cs#L40).

The longer-term aim would be to refactor the existing code to use this abstraction. However, it's out of scope for this release, so I've created a separate ticket for it.

In fact, I've created two: one to refactor existing code to use this method (#4481), and a second to extend the abstraction to return the directory name too (#4482), as that code is duplicated in several places too.

src/Infrastructure.VS/SolutionInfoProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        [ImportingConstructor]
+        public SolutionInfoProvider([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider,
+            IThreadHandling threadHandling)
+        {
+            this.serviceProvider = serviceProvider;
+            this.threadHandling = threadHandling;
+        }
+
+        public async Task<string> GetSolutionNameAsync()
```
duncanp-sonar, line 45: We need the current solution name to save the Roslyn suppressions file in the correct place. This PR adds an abstraction to fetch it.

We have very similar code in several other places that uses `IVsSolution` e.g. [MigrationSettingsProvider](https://github.com/SonarSource/sonarlint-visualstudio/blob/5cd8033fbb2cde08c631cfaaf5851fabcfa11f16/src/ConnectedMode/Migration/MigrationSettingsProvider.cs#L40).

The longer-term aim would be to refactor the existing code to use this abstraction. However, it's out of scope for this release, so I've created a separate ticket for it.

In fact, I've created two: one to refactor existing code to use this method (#4481), and a second to extend the abstraction to return the directory name too (#4482), as that code is duplicated in several places too.

bigfluffycookie (approved):

PR #4486, opened 2023-07-07
Trigger Roslyn suppressed issues update post-migration

### LINKED ISSUE #4485: [Migration] C#/VB.NET suppressions are not updated during migration
### Description

Bug #4480 occurred because the file location for the Roslyn suppressions file has changed. This means that during migration we need to re-save the suppressions file in the new location, otherwise our diagnostic suppressor won't pick it up.

Workarounds:
* close and re-open the solution -> suppressions file is written in the new location, or
* manually trigger an update from the `Team Explorer`

#### Issues are not suppressed immediately after migration

The correct way to fix this is probably to have the new unintrusive binding process class interact with the `ActiveSolutionBoundTracker` to trigger a "binding updated" event.

Note: the suppression files are written under `%LocalAppData%l\Temp\SLVS\Roslyn`

duncanp-sonar:
Fixes #4485


src/ConnectedMode/Binding/IRoslynSettingsFileSynchronizer.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+using System.Threading.Tasks;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Binding
+{
+    /// <summary>
+    /// Responsible for listening to <see cref="IServerIssuesStore.ServerIssuesChanged"/> and calling
+    /// <see cref="IRoslynSettingsFileStorage.Update"/> with the new suppressions.
+    /// </summary>
+    public interface IRoslynSettingsFileSynchronizer : IDisposable
```
duncanp-sonar, line 30: I had to move the interface out of `RoslynSuppressions` so that the migration code can import it.

Normally we'd put shared interfaces like this in `Core`. However, `Core` is starting to become a bit of a dumping ground for interfaces to multiple different unrelated features.

So, since the Roslyn suppressions file is only relevant in Connected Mode I decided to move it to the `ConnectedMode` assembly instead. `RoslynSuppressions` already has a project dependency on the `ConnectedMode` project.

If we need to move the interfaces in the future we can easily do so.

duncanp-sonar:
Triggering an explicit isn't enough on its own - needs more work.

PR #4490, opened 2023-07-07
Change RoslynSettingsFileSynch to use IConfigurationProvider

duncanp-sonar:
... instead of IActiveSolutionBoundTracker.

Prep for #4485

### DIFF
--- a/src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
+++ b/src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
@@ -51,19 +51,19 @@ namespace SonarLint.VisualStudio.Roslyn.Suppressions.InProcess
         private readonly IThreadHandling threadHandling;
         private readonly IServerIssuesStore serverIssuesStore;
         private readonly IRoslynSettingsFileStorage roslynSettingsFileStorage;
-        private readonly IActiveSolutionBoundTracker activeSolutionBoundTracker;
+        private readonly IConfigurationProvider configurationProvider;
         private readonly ISolutionInfoProvider solutionInfoProvider;
         private readonly ILogger logger;
 
         [ImportingConstructor]
         public RoslynSettingsFileSynchronizer(IServerIssuesStore serverIssuesStore,
             IRoslynSettingsFileStorage roslynSettingsFileStorage,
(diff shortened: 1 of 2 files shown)


src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -51,19 +51,19 @@ internal sealed class RoslynSettingsFileSynchronizer : IRoslynSettingsFileSynchr
         private readonly IThreadHandling threadHandling;
         private readonly IServerIssuesStore serverIssuesStore;
         private readonly IRoslynSettingsFileStorage roslynSettingsFileStorage;
-        private readonly IActiveSolutionBoundTracker activeSolutionBoundTracker;
```
duncanp-sonar, line 54: Swapped to use `IConfigurationProvider` rather than `IActiveSolutionBoundTracker` ("ASBT").

The ASBT uses the config provider internally and caches the value, which means it can be out of date when the binding state changes until the cached value is updated e.g. during migration.

The config provider re-calculates the value each time so it is always up to date i.e. as soon as the new format config files are written to disc the config provider will detect them.

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -115,7 +115,7 @@ public async Task UpdateFileStorageAsync()
             {
                 await threadHandling.SwitchToBackgroundThread();
 
-                var sonarProjectKey = activeSolutionBoundTracker.CurrentConfiguration.Project?.ProjectKey;
+                var sonarProjectKey = configurationProvider.GetConfiguration().Project?.ProjectKey;
```
duncanp-sonar, line 119: The config provider was a straightforward swap in the product and the tests.
This line is the only other change.

duncanp-sonar:
I've been trying a number of different ways to force the Roslyn suppressions to be updated at the end of the migration, and I've finally found one that works.

This is a preparatory change; make the Roslyn settings synchronizer pick up the "live" config info rather than the cached info (which won't be updated until after the migration wizard has finished i.e. too late).

bigfluffycookie (approved):
Very nice!

PR #4492, opened 2023-07-07
Implement HotspotMatcher

georgii-borovinskikh-sonarsource:


src/IssueViz.Security/Hotspots/HotspotMatcher.cs
```diff
@@ -30,11 +32,24 @@ internal interface IHotspotMatcher
...
+            }
+
+            if (!string.IsNullOrEmpty(serverHotspot.LineHash)
+                && StringComparer.Ordinal.Equals(localHotspotVisualization.LineHash, serverHotspot.LineHash))
+            {
+                return true;
+            }
+
+            return localHotspotVisualization.StartLine == serverHotspot.TextRange.StartLine 
+                   || StringComparer.Ordinal.Equals(localHotspotVisualization.Issue.PrimaryLocation.Message, serverHotspot.Message);
```
duncanp-sonar, line 52: The issue matching for non-hotspots doesn't look at the 'Message', does it?

What's the reason for using a different algorithm here?

Using a different algorithm is fine as long as there is a reason for it. However, we need to capture the reason for it, otherwise it will look like a bug to anyone looking at the code in the future.

src/IssueViz.Security.UnitTests/Hotspots/HotspotMatcherTests.cs
```diff
@@ -30,6 +36,188 @@ public class HotspotMatcherTests
     [TestMethod]
     public void MefCtor_CheckExports()
     {
-        MefTestHelpers.CheckTypeCanBeImported<NoOpHotspotMatcher, IHotspotMatcher>();
+        MefTestHelpers.CheckTypeCanBeImported<HotspotMatcher, IHotspotMatcher>();
+    }
```
duncanp-sonar, line 40: FYI we now have additional test helpers for [CheckIsNonSharedMefComponent](https://github.com/search?q=repo%3ASonarSource%2Fsonarlint-visualstudio+CheckIsNonSharedMefComponent&type=code) and [CheckIsSingletonMefComponent](https://github.com/search?q=repo%3ASonarSource%2Fsonarlint-visualstudio%20CheckIsSingletonMefComponent&type=code).

When you touch a test file that doesn't have a `CheckIsNonSharedMefComponent` / `CheckIsSingletonMefComponent`, could you add one please?

src/IssueViz.Security.UnitTests/Hotspots/HotspotMatcherTests.cs
```diff
@@ -30,6 +36,188 @@ public class HotspotMatcherTests
...
+        var testSubject = CreateTestSubject();
+
+        testSubject.IsMatch(CreateLocalHotspot(ruleId, "message1", "abchash", filePath, 10),
+                CreateServerHotspot(ruleId, "message2", null, serverPath, 10))
+            .Should()
+            .BeTrue();
+    }
+    
+    [TestMethod]
+    public void IsMatch_SameIssueDifferentHash_ReturnsTrue()
```
duncanp-sonar, line 72: A number of the tests have "SameIssue" in the name; it's not clear what that means, or what the test is checking.

src/IssueViz.Security.UnitTests/Hotspots/HotspotMatcherTests.cs
```diff
@@ -30,6 +36,188 @@ public class HotspotMatcherTests
...
+            .BeTrue();
+    }
+
+    [TestMethod]
+    public void IsMatch_DifferentRuleId_ReturnsFalse()
+    {
+        const string message = "some message";
+        const string filePath = "A:\\ny\\p\\ath";
+        const string serverPath = "p\\ath";
+        const string lineHash = null;
```
duncanp-sonar, line 134: It would be better to have matching non-null line hash value in this test since a null server line hash is treated differently.

src/IssueViz.Security.UnitTests/Hotspots/HotspotMatcherTests.cs
```diff
@@ -30,6 +36,188 @@ public class HotspotMatcherTests
     [TestMethod]
     public void MefCtor_CheckExports()
     {
-        MefTestHelpers.CheckTypeCanBeImported<NoOpHotspotMatcher, IHotspotMatcher>();
+        MefTestHelpers.CheckTypeCanBeImported<HotspotMatcher, IHotspotMatcher>();
+    }
+    
+    [TestMethod]
+    public void IsMatch_NoHashDifferentLineSameMessage_ReturnsTrue()
```
duncanp-sonar, line 43: It's difficult to tell whether all of the possible cases are covered or not from the test names.

Ideally most of the tests would be varying one value at a time and testing for both `true` and `false` cases e.g. this test would be stronger (and more isolated) if it had different inputs for the `message`, and tested the product code only returned `true` when the messages were identical and in the same case.

e.g.
```
    [TestMethod]
    [DataRow(null, null, false)] // null server has -> no match
    [DataRow(null, "any", false)] // null server hash -> no match
    [DataRow("", "", false)] // empty server -> no match
    [DataRow("", "any", false)] // empty server -> no match
    [DataRow("aaa", "AAA", false)] // different case - no match
    [DataRow("sameHash", "sameHash", true)]
    public void IsMatch_SameFileAndRuleId_VaryLineHashes(string serverHash, string localHash, bool shouldMatch)
    {
        // Different lines and messages.
        // RuleIds and paths do match.
        const string ruleId = "rule1";
        const string filePath = "A:\\ny\\p\\ath";
        const string serverPath = "p\\ath";

        testSubject.IsMatch(CreateLocalHotspot(ruleId, "message 1", localHash, filePath, 10),
        CreateServerHotspot(ruleId, "message 2", serverHash, serverPath, 20))
            .Should()
            .Be(shouldMatch);
    }

    [TestMethod]
    [DataRow(0, 1, false)]
    [DataRow(1, 0, false)]
    [DataRow(0, 0, true)]
    [DataRow(1, 1, true)]
    public void IsMatch_SameFileAndRuleId_VaryLineNumber(int serverLineNumber, int localLineNumber, bool shouldMatch)
    {
        // Different hashes and messages.
        // RuleIds and paths do match.
        const string ruleId = "rule1";
        const string filePath = "A:\\ny\\p\\ath";
        const string serverPath = "p\\ath";

        testSubject.IsMatch(CreateLocalHotspot(ruleId, "message 1", "localHash", filePath, serverLineNumber),
        CreateServerHotspot(ruleId, "message 2", "serverHash", serverPath, localLineNumber))
            .Should()
            .Be(shouldMatch);
    }

    [TestMethod]
    [DataRow(null, null, true)]
    [DataRow("", "", true)]
    [DataRow("aaa", "AAA", false)] // different case -> no match
    [DataRow("sameMessage", "sameMessage", true)]
    public void IsMatch_SameFileAndRuleId_VaryMessages(string serverMessage, string localMessage, bool shouldMatch)
    {
        // Different lines and hashes.
        // RuleIds and paths do match.
        const string ruleId = "rule1";
        const string filePath = "A:\\ny\\p\\ath";
        const string serverPath = "p\\ath";

        testSubject.IsMatch(CreateLocalHotspot(ruleId, localMessage, "localHash", filePath, 10),
        CreateServerHotspot(ruleId, serverMessage, "serverHash", serverPath, 20))
            .Should()
            .Be(shouldMatch);
    }
```

I think those example tests plus
`IsMatch_DifferentRuleId_ReturnsFalse` and `IsMatch_IdenticalIssuesExceptFilePath_ReturnsBasedOnFilePathMatch` cover all of the cases and are easier to follow.

src/IssueViz.Security/Hotspots/HotspotMatcher.cs
georgii-borovinskikh-sonarsource: I got this idea from talking to Damien. Since there's no hashes for hotspots in the API response, they use message as an additional feature to match based on it.

duncanp-sonar: Ok, makes sense (although that does beg the question that if we don't have hash for hotspots, why is the code above checking for them at all?).

This is a clear design/implementation decision that deserves to be captured. Otherwise, it's not obvious whether it's a mistake or not e.g.


```
// We don't currently expect to have hashes for hotspopts so we use the same approach as other SL flavours and use the exact message as an extra value to match on (we don't check the message for non-hotspot issues since we do have hashes).
```

duncanp-sonar (approved):
Looks good!

PR #4496, opened 2023-07-07
Catch any xml load errors

georgii-borovinskikh-sonarsource:


src/ConnectedMode/Migration/XmlFileCleaner.cs
```diff
@@ -53,7 +53,11 @@ public XmlFileCleaner(ILogger logger) : this(logger, new XmlDocumentHelper())
 
         public string Clean(string content, LegacySettings legacySettings, CancellationToken token)
         {
-            var document = xmlDocumentHelper.LoadFromString(content);
+            if (!xmlDocumentHelper.TryLoadFromString(content, out var document))
+            {
+                logger.WriteLine("Could not load xml");
```
duncanp-sonar, line 58: Normal verbosity string => should be in a resource file and should have a `[Migration]` prefix.
`The file could not be loaded as XML` might be a clearer message.

Could you fix the other string in the file at the same time please?

duncanp-sonar (commented):
Looks good; just a suggestion about updating the string.

PR #4503, opened 2023-07-10
Fix LocalMetadataProviderTests.GetRuleHelp_CheckAllRules to check tha…

georgii-borovinskikh-sonarsource:
…t all rules of rich format are loaded correctly and that all rules have at least one type of rule description

### DIFF
--- a/src/Rules.UnitTests/LocalRuleMetadataProviderTests.cs
+++ b/src/Rules.UnitTests/LocalRuleMetadataProviderTests.cs
@@ -88,12 +88,22 @@ namespace SonarLint.VisualStudio.Rules.UnitTests
                 Console.WriteLine("Checking " + fullResourceName);
 
                 SonarCompositeRuleId ruleId = GetCompositeRuleIdFromResourceName(fullResourceName);
-                var expectedDescription = GetEmbeddedRuleDescription(fullResourceName);
+                var expectedRuleInfo = GetEmbeddedRuleDescription(fullResourceName);
 
                 var actual = testSubject.GetRuleInfo(ruleId);
 
                 actual.FullRuleKey.Should().Be(ruleId.ToString());
-                actual.Description.Should().Be(expectedDescription);
+                actual.Description.Should().Be(expectedRuleInfo.Description);
+                
+                if (actual.DescriptionSections.Count > 0)
+                {
+                    actual.DescriptionSections.Should().BeEquivalentTo(expectedRuleInfo.DescriptionSections);
+                    actual.DescriptionSections.All(section => !string.IsNullOrWhiteSpace(section.HtmlContent)).Should().BeTrue();
+                }
+
+                (!string.IsNullOrWhiteSpace(actual.Description) || actual.DescriptionSections.Count > 0)
+                    .Should()
+                    .BeTrue();
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Rules.UnitTests/LocalRuleMetadataProviderTests.cs
```diff
@@ -112,11 +122,11 @@ private static SonarCompositeRuleId GetCompositeRuleIdFromResourceName(string fu
             return new SonarCompositeRuleId(repoKey, ruleKey);
         }
 
-        private static string GetEmbeddedRuleDescription(string fullResourceName)
+        private static IRuleInfo GetEmbeddedRuleDescription(string fullResourceName)
```
duncanp-sonar, line 125: Minor: the method name is no longer accurate. `GetEmbeddedRuleInfo`?

src/Rules.UnitTests/LocalRuleMetadataProviderTests.cs
```diff
@@ -88,12 +88,22 @@ void CheckRule(string fullResourceName)
...
-                actual.Description.Should().Be(expectedDescription);
+                actual.Description.Should().Be(expectedRuleInfo.Description);
+                
+                if (actual.DescriptionSections.Count > 0)
+                {
+                    actual.DescriptionSections.Should().BeEquivalentTo(expectedRuleInfo.DescriptionSections);
+                    actual.DescriptionSections.All(section => !string.IsNullOrWhiteSpace(section.HtmlContent)).Should().BeTrue();
+                }
+
+                (!string.IsNullOrWhiteSpace(actual.Description) || actual.DescriptionSections.Count > 0)
```
duncanp-sonar, line 104: Just to be clear, is it ok for a rule to have a non-null `Description` and multiple `DescriptionSections`?

src/Rules.UnitTests/LocalRuleMetadataProviderTests.cs
```diff
@@ -88,12 +88,22 @@ void CheckRule(string fullResourceName)
...
-                var expectedDescription = GetEmbeddedRuleDescription(fullResourceName);
+                var expectedRuleInfo = GetEmbeddedRuleDescription(fullResourceName);
 
                 var actual = testSubject.GetRuleInfo(ruleId);
 
                 actual.FullRuleKey.Should().Be(ruleId.ToString());
-                actual.Description.Should().Be(expectedDescription);
+                actual.Description.Should().Be(expectedRuleInfo.Description);
+                
+                if (actual.DescriptionSections.Count > 0)
```
duncanp-sonar, line 98: Shouldn't this be checking `expectedRuleInfo.DescriptionsSections.Count > 0` instead?

Currently, if the `actual.Descriptions.Count` incorrectly returns no sections when we are expecting some, the tests won't pick it up.

duncanp-sonar (commented):
Couple of small things: see comments.

src/Rules.UnitTests/LocalRuleMetadataProviderTests.cs
georgii-borovinskikh-sonarsource: I think the problem here is that `actual.DescriptionSections.Should().BeEquivalentTo(expectedRuleInfo.DescriptionSections);` should be outside of the IF

src/Rules.UnitTests/LocalRuleMetadataProviderTests.cs
georgii-borovinskikh-sonarsource: Yes. Some rule descriptions still have legacy rule description along with the rich

PR #4504, opened 2023-07-10
Exclude wizard XAML component from code coverage

duncanp-sonar:
- also exclude the thin controller that creates it


duncanp-sonar:
@bigfluffycookie FYI I've excluded the XAML wizard and controller from coverage; we've tested them extensively manually.

PR #4530, opened 2023-07-18
Add IDocumentEvents interface and implementation

### LINKED ISSUE #4509: Add document open/closing tracker
### Description

Add an interface that provides a notification of when a document is closed (only need close at the moment).

Prerequisite for #4508

#### Notes

There are various VS APIs that might be of use:

[IVsRunningDocTableEventsX](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsrunningdoctableevents2.onbeforelastdocumentunlock?view=visualstudiosdk-2022): there isn't a specific "doc closed/closing event". [This page](https://learn.microsoft.com/en-us/visualstudio/extensibility/internals/running-document-table?view=vs-2022#edit-locks-and-read-locks) suggests that the lock count will be zero when the document is about to be closed.

The downside with this approach is that we want the document tracker to be MEF-importable. However, the VS threading rules mean the MEF constructor has to be free-threaded so we can't call `serviceProvider.GetService` to get the running document table i.e. it's going to be difficult to initialize the component safely in such a way that we don't miss any notifications.

[DTE.Events.DocumentEvents](https://learn.microsoft.com/en-us/dotnet/api/envdte.documentevents?view=visualstudiosdk-2022): there is a `DocumentClosing` event, but it doesn't seem to fire consistently i.e. when testing, it didn't fire for all documents that were closed.

Alternatively, we could rely on our `TaggerProvider` implementation:
Pros:
* it's already MEF-imported (although not strictly free-threaded)
* it tracks the lifetime of documents that we analyse -> it knows when an analyzed document has been closed
* simple: minimal amount of new code

Cons:
* the `TaggerProvider` constructor has its own existing free-threading issues. However, it has worked correctly for years so far.
* it relies on the rather complicated tagger counting code to know that a file has been closed. Again, this has worked correctly for years.

duncanp-sonar:
Fixes #4509

### DIFF
--- a/src/Core/DocumentEvents.cs
+++ b/src/Core/DocumentEvents.cs
@@ -20,7 +20,7 @@
 
 using System;
 
-namespace SonarLint.VisualStudio.Infrastructure.VS.DocumentEvents
+namespace SonarLint.VisualStudio.Core
 {
     public class DocumentClosedEventArgs : EventArgs
     {
--- a/src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
+++ b/src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
@@ -20,12 +20,14 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.Composition.Primitives;
 using System.Linq;
 using System.Text;
 using System.Threading;
 using EnvDTE;
 using EnvDTE80;
 using FluentAssertions;
+using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.Text;
@@ -34,6 +36,7 @@ using Microsoft.VisualStudio.Utilities;
 using Moq;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Infrastructure.VS.DocumentEvents;
 using SonarLint.VisualStudio.Integration.Vsix;
 using SonarLint.VisualStudio.Integration.Vsix.Analysis;
 using SonarLint.VisualStudio.Integration.Vsix.ErrorList;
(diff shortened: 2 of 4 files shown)


src/Core/DocumentEvents.cs
```diff
@@ -0,0 +1,48 @@
...
+        public DocumentClosedEventArgs(string fullPath)
+        {
+            FullPath = fullPath;
+        }
+    }
+
+    /// <summary>
+    /// Raises notifications about document events
+    /// </summary>
+    public interface IDocumentEvents
```
duncanp-sonar, line 41: Simple new interface that currently just notifies that a document has been closed.

This interface is in Core as it doesn't currently expose any VS-specific interfaces, and it's implemented by a type in Integration.VSIX but will be consumed by ConnectedMode.

If we need to add VS-specific types to the API later we would move the interface declaration to VSInfrastructure.

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
```diff
@@ -117,6 +120,57 @@ public void SetUp()
...
+        [TestMethod]
+        public void CheckIsSingletonMefComponent()
+            => MefTestHelpers.CheckIsSingletonMefComponent<TaggerProvider>();
+
+        [TestMethod]
+        public void MefCtor_CheckIsExported_TaggerProvider()
+            => MefTestHelpers.CheckTypeCanBeImported<TaggerProvider, ITaggerProvider>(GetRequiredExports());
+
+        [TestMethod]
+        public void MefCtor_CheckIsExported_DocumentEvents()
```
duncanp-sonar, line 135: ... and a new test for the new export.

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
```diff
@@ -117,6 +120,57 @@ public void SetUp()
...
+            => MefTestHelpers.CheckTypeCanBeImported<TaggerProvider, IDocumentEvents>(GetRequiredExports());
+
+        [TestMethod]
+        public void MefCtor_Check_SameInstanceExported()
```
duncanp-sonar, line 139: ... and a test that the exports return the same instance, which is what we need to happen.

This is based on the [equivalent test](https://github.com/SonarSource/sonarlint-visualstudio/blob/7543d596f851c0d09b910da7996162d6c781531b/src/ConnectedMode.UnitTests/Suppressions/ServerIssuesStoreTests.cs#L45) for `ServerIssuesStore`.

At the time the `ServerIssuesStoreTest` was written we didn't have the test helper `MefTestHelpers.CheckIsSingletonMefComponent`. In theory we could just rely on the fact that the export is marked as a singleton and assume we'd get the same instance for each export. However, this test is more explicit and not particularly complicated.

If we had more types with multiple exports it would be worth adding a new test helper.

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
```diff
@@ -117,6 +120,57 @@ public void SetUp()
...
+            MefTestHelpers.CreateExport<ITextDocumentFactoryService>(),
+            MefTestHelpers.CreateExport<IAnalyzerController>(),
+            MefTestHelpers.CreateExport<SVsServiceProvider>(),
+            MefTestHelpers.CreateExport<ISonarLanguageRecognizer>(),
+            MefTestHelpers.CreateExport<IAnalysisRequester>(),
+            MefTestHelpers.CreateExport<ITaggableBufferIndicator>(),
+            MefTestHelpers.CreateExport<IIssueConsumerFactory>(),
+            MefTestHelpers.CreateExport<ILogger>(),
+            MefTestHelpers.CreateExport<IScheduler>(),
+            MefTestHelpers.CreateExport<IThreadHandling>()
```
duncanp-sonar, line 169: This class is problematic in that it has too many dependencies. However, fixing that is a separate issue. We're not making it any worse here.

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
```diff
@@ -210,10 +264,15 @@ public void CreateTagger_should_return_new_tagger_for_already_tracked_file()
         }
 
         [TestMethod]
-        public void CreateTagger_close_last_tagger_should_unregister_tracker()
+        public void CreateTagger_CloseLastTagger_ShouldUnregisterTrackerAndRaiseEvent()
         {
             var doc1 = CreateMockedDocument("doc1.js");
 
+            object actualSender = null;
```
duncanp-sonar, line 271: Updated the existing test to check the event is only raised when the last tagger is removed.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -47,10 +48,12 @@ namespace SonarLint.VisualStudio.Integration.Vsix
     /// See the README.md in this folder for more information
     /// </remarks>
     [Export(typeof(ITaggerProvider))]
+    [Export(typeof(IDocumentEvents))]
```
duncanp-sonar, line 51: There are a number of VS APIs that look like they should tell us when a file has been closed.
However, using them would involve a lot more work, particularly to make them follow the VS threading rules for MEF components.

The `TaggerProvider` already tracks all of the documents that we can analyse, and knows when a document has been closed. The simplest implementation that works is the moment is just to raise the event from this class for now.

I've manually tested this by modifying the ConnectedMode package to import `IDocumentEvents` and log when documents are closed, and it works as expected (if there are two or more instances of a window for the same file, the event is not raised until the last window is closed).

See the notes in #4509 for more information.

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
```diff
@@ -223,10 +282,34 @@ public void CreateTagger_close_last_tagger_should_unregister_tracker()
             // Remove one tagger -> tracker should still be registered
```
georgii-borovinskikh-sonarsource, line 282: why is provider.ActiveTrackersForTesting.Count never 2 in this test? even though we add 2 trackers to it?

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -244,7 +247,17 @@ public void RemoveIssueTracker(IIssueTracker issueTracker)
             lock (issueTrackers)
             {
                 issueTrackers.Remove(issueTracker);
+
+                // The lifetime of an issue tracker is tied to a single document. A tracker is removed when
+                // it is no longer needed i.e. the document has been closed.
```
georgii-borovinskikh-sonarsource, line 252: Can we have multiple taggers for the same document? What will happen then? Is this event raised multiple times?

georgii-borovinskikh-sonarsource (approved):

src/Integration.Vsix.UnitTests/SonarLintTagger/TaggerProviderTests.cs
duncanp-sonar: We don't - we add two taggers.

There is one _tracker_ per document. Each tracker will have one or more _taggers_.
When all of the taggers have been removed, the tracker will unregister itself from the TaggerProvider.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
duncanp-sonar: 1. Yes.
3. No.

See [ReadMe.md](https://github.com/SonarSource/sonarlint-visualstudio/blob/7543d596f851c0d09b910da7996162d6c781531b/src/Integration.Vsix/SonarLintTagger/README.md?plain=1#L4) for more information.

PR #4540, opened 2023-07-19
Trigger local hotspot store cleanup when a file is closed

duncanp-sonar:
Part of #4509

### DIFF
--- a/src/IssueViz.Security/Hotspots/DocumentClosedHandler.cs
+++ b/src/IssueViz.Security/Hotspots/DocumentClosedHandler.cs
@@ -31,16 +31,16 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Hotspots
     internal sealed class DocumentClosedHandler : IDisposable
     {
         private readonly IDocumentEvents documentEvents;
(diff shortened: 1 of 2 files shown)


src/IssueViz.Security/Hotspots/DocumentClosedHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Hotspots
+{
+    [Export(typeof(DocumentClosedHandler))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal sealed class DocumentClosedHandler : IDisposable
```
duncanp-sonar, line 31: Simple class that does the cleanup.

It isn't being created anywhere yet; I'll do that in the next PR. I'll either have to move some classes around so we can create it at the point we need it, or do some other MEF tweaks.

georgii-borovinskikh-sonarsource (approved):

Issue #4508, opened 2023-07-11
Remove hotspots from store when file is closed


duncanp-sonar:
### Description

Hotspots are only shown for open files i.e.
* when a file is opened -> trigger analysis
* when a file is closed -> remove any open issues from the store.

? What happens when a document becomes inactive? i.e. still open but not top-most?

PR #4546, opened 2023-07-20
Create doc closed handler in ConnectedMode package initialization

duncanp-sonar:
Part of #4508

### DIFF
--- a/src/ConnectedMode/ConnectedModePackage.cs
+++ b/src/ConnectedMode/ConnectedModePackage.cs
@@ -32,6 +32,7 @@ using SonarLint.VisualStudio.ConnectedMode.Install;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Integration;
 using Task = System.Threading.Tasks.Task;
+using SonarLint.VisualStudio.ConnectedMode.Hotspots;
 
 namespace SonarLint.VisualStudio.ConnectedMode
 {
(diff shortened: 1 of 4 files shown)


src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -67,9 +69,11 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
             boundSolutionUpdateHandler = componentModel.GetService<BoundSolutionUpdateHandler>();
             timedUpdateHandler = componentModel.GetService<TimedUpdateHandler>();
             localSuppressionsChangedHandler = componentModel.GetService<LocalSuppressionsChangedHandler>();
-            
+
             importBeforeInstallTrigger = componentModel.GetService<ImportBeforeInstallTrigger>();
 
+            hotspotDocumentClosedHandler = componentModel.GetService<IHotspotDocumentClosedHandler>();
```
duncanp-sonar, line 75: Create an instance of the doc closing handler when the ConnectedMode package loads.

Note: there _might_ be a corner case where this is too late e.g. the user starts VS by opening a bound solution that has an open file that contains hotspots. In that case, there will be a delay before VS loads the package in the background, and the user might close the file in that gap.
I haven't been able to test whether this corner case exists yet as we don't have a working end-to-end. I'll add a note to the Testing ticket.

I tried importing the class in different places e.g. in the `LocalHotspotStore` itself (since we want the lifetime of the doc manager to be the same). However, that creates MEF circular references. I've gone with this solution for now as we need to move forward.

src/ConnectedMode/Hotspots/IHotspotDocumentClosedHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.ConnectedMode.Hotspots
+{
+    /// <summary>
+    /// Marker interface for the component that listens for documents being closed
+    /// and triggers the removal of issues from the local hotsp
+    /// </summary>
+    public interface IHotspotDocumentClosedHandler { }
```
duncanp-sonar, line 27: New interface in the ConnectedMode project.

I couldn't move the hotspot store etc to the ConnectedMode project without moving a whole load of other classes, so I created this interface instead.

georgii-borovinskikh-sonarsource (approved):

Issue #4548, opened 2023-07-20
Create rule key to review priority mapping


georgii-borovinskikh-sonarsource:
Needed for #4532

PR #4551, opened 2023-07-20
Add HotspotReviewPriorityProvider

georgii-borovinskikh-sonarsource:
Implements #4548

### DIFF
--- a/src/IssueViz.Security/Hotspots/HotspotReviewPriorityProvider.cs
+++ b/src/IssueViz.Security/Hotspots/HotspotReviewPriorityProvider.cs
@@ -1,0 +1,222 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.IssueVisualization.Security.Hotspots.Models;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.Hotspots
+{
+    /// <summary>
+    /// Provides a mapping for hotspots to their review priority
+    /// </summary>
+    public interface IHotspotReviewPriorityProvider
+    {
+        /// <summary>
+        /// Supported languages:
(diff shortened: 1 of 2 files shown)


src/IssueViz.Security/Hotspots/HotspotReviewPriorityProvider.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// </list>
+        /// </summary>
+        /// <param name="hotspotKey">Hotspot Rule Key</param>
+        /// <returns>Associated review priority or null if not mapped</returns>
+        HotspotPriority? GetPriority(string hotspotKey);
+    }
+
+    [Export(typeof(IHotspotReviewPriorityProvider))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class HotspotReviewPriorityProvider : IHotspotReviewPriorityProvider
```
duncanp-sonar, line 48: I think this needs a comment explaining why we are using a hard-coded list, as it isn't something we would normally do.

src/IssueViz.Security/Hotspots/HotspotReviewPriorityProvider.cs
```diff
@@ -0,0 +1,222 @@
...
+            { "typescript:S5852", HotspotPriority.Medium },
+            { "typescript:S6268", HotspotPriority.High },
+            { "typescript:S5734", HotspotPriority.Low },
+            { "typescript:S5732", HotspotPriority.Low },
+            { "typescript:S6350", HotspotPriority.High },
+        };
+
+        public HotspotPriority? GetPriority(string hotspotKey)
+        {
+            return hotspotKey == null || !KeyToPriorityMap.TryGetValue(hotspotKey, out var priority) ? (HotspotPriority?)null : priority;
```
duncanp-sonar, line 219: Calling this method with `null` would be an error in our code, wouldn't it?

I wouldn't code defensively in this case; I'd throw an exception.

src/IssueViz.Security/Hotspots/HotspotReviewPriorityProvider.cs
```diff
@@ -0,0 +1,222 @@
...
+        HotspotPriority? GetPriority(string hotspotKey);
+    }
+
+    [Export(typeof(IHotspotReviewPriorityProvider))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class HotspotReviewPriorityProvider : IHotspotReviewPriorityProvider
+    {
+        private static readonly Dictionary<string, HotspotPriority> KeyToPriorityMap = new Dictionary<string, HotspotPriority>
+        {
+            { "cpp:S5443", HotspotPriority.Low },
```
duncanp-sonar, line 53: It would be useful if we had a test that checked that no new hotspots have been added.
However, we don't need it for this sprint. Could you add an Infra ticket to the backlog please(/

src/IssueViz.Security/Hotspots/HotspotReviewPriorityProvider.cs
georgii-borovinskikh-sonarsource: added

src/IssueViz.Security/Hotspots/HotspotReviewPriorityProvider.cs
georgii-borovinskikh-sonarsource: Added null check

src/IssueViz.Security/Hotspots/HotspotReviewPriorityProvider.cs
georgii-borovinskikh-sonarsource: https://github.com/SonarSource/sonarlint-visualstudio/issues/4554

duncanp-sonar (approved):

PR #4553, opened 2023-07-20
Raise config changed when solution binding changes

### LINKED ISSUE #4547: [Bug] Local hotspot analysis is not executed
### Description

`HotspotAnalysisConfiguration` gets called only once - before we actually connect to SQ - which leaves Hotspot analysis disabled

#### Repro
1. start VS by open a bound solution that has a JS/TS file with a hotspot open

#### Expected
The file is analyzed and the hotspot is reported in the hotspots view

#### Actual
The hotspot is not reported, even when the file is closed and opened again.

#### Notes
The problem occurs if the file is analyzed before the connection to the server has been established. If that happens, the rule config is calculated with "enabled hotspots = false" and not recalculated.

The problem doesn't occur if the connection to the server has been established before any JS/TS file is opened.

TODO - is the CFamily analyzer affected?
Yes, but not as badly. Initially, an open file will be analyzed with hotspots disabled.
However, the rule settings are not cached in Connected Mode (?), so when the file is re-analyzed the hotspot rules will be enabled.

TODO - it is possible the same issue exists in the release version when the the user updates a binding e.g.
1. open a bound solution, open a JS/TS file with an issue for rule X.
2. update the QP on the server to disable rule X
3. update the binding
4. re-analyze the file
Expected: issue for rule X is not reported.
Actual: ?

duncanp-sonar:
Fixes #4547

### DIFF
--- a/src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
+++ b/src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
@@ -24,6 +24,7 @@ using System.Threading.Tasks;
 using Microsoft.VisualStudio.Threading;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Infrastructure.VS;
 
 namespace SonarLint.VisualStudio.Integration.Vsix.Analysis
@@ -38,6 +39,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis
     {
         private readonly IAnalysisRequester analysisRequester;
         private readonly IUserSettingsProvider userSettingsProvider;
+        private readonly IActiveSolutionBoundTracker activeSolutionBoundTracker;
         private readonly ILogger logger;
         private readonly IThreadHandling threadHandling;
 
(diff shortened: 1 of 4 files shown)


src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -46,20 +48,30 @@ internal sealed class AnalysisConfigMonitor : IAnalysisConfigMonitor, IDisposabl
         [ImportingConstructor]
         public AnalysisConfigMonitor(IAnalysisRequester analysisRequester,
             IUserSettingsProvider userSettingsProvider, // reports changes to user settings.json
```
duncanp-sonar, line 50: This existing class monitors changes that can impact the analysis results.
The ESLintBridge analyzer classes listen to the `ConfigChanged` event and flag that the ESLintBridge needs to be re-initialised i.e. the rules config re-calculated.

The solution binding changing (e.g. from unbound to bound) does affect the analysis, so I've changed the monitor to listen for this event too.

There is a performance impact to this change, as it will also trigger the re-analysis of all open documents. However, that's probably the behaviour we want.

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -46,20 +48,30 @@ internal sealed class AnalysisConfigMonitor : IAnalysisConfigMonitor, IDisposabl
...
             IThreadHandling threadHandling)
         {
             this.analysisRequester = analysisRequester;
             this.userSettingsProvider = userSettingsProvider;
+            this.activeSolutionBoundTracker = activeSolutionBoundTracker;
             this.logger = logger;
             this.threadHandling = threadHandling;
 
             userSettingsProvider.SettingsChanged += OnUserSettingsChanged;
+            activeSolutionBoundTracker.SolutionBindingChanged += OnSolutionBindingChanged;
```
duncanp-sonar, line 68: Perhaps we should listen to the `SolutionBindingUpdated` event too. However, I'll create a separate ticket for that as it would be a separate bug.

georgii-borovinskikh-sonarsource (approved):

PR #4564, opened 2023-07-20
Disable "Remove" hotspot view context menu command

### LINKED ISSUE #4552: Disable Remove from hotspots list
Old hotspots list that was only used with Open in IDE allowed to remove hotspots from the list. It doesn't make sense to have this option now that we have locally analyzed hotspots in the list

duncanp-sonar:
Fixes #4552

### DIFF
--- a/src/IssueViz.Security/Hotspots/HotspotsList/ViewModels/HotspotsControlViewModel.cs
+++ b/src/IssueViz.Security/Hotspots/HotspotsList/ViewModels/HotspotsControlViewModel.cs
@@ -23,7 +23,6 @@ using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.Linq;
 using System.Runtime.CompilerServices;
-using System.Threading.Tasks;
 using System.Windows.Data;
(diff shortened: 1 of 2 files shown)


src/IssueViz.Security/Hotspots/HotspotsList/ViewModels/HotspotsControlViewModel.cs
```diff
@@ -121,11 +116,6 @@ private void SetCommands(ILocationNavigator locationNavigator)
                 var hotspot = (IHotspotViewModel)parameter;
                 locationNavigator.TryNavigate(hotspot.Hotspot);
             }, parameter => parameter is IHotspotViewModel);
-
-            RemoveCommand = new DelegateCommand(parameter =>
-            {
-                // todo fix in separate PR
-            }, parameter => parameter is IHotspotViewModel);
```
duncanp-sonar, line 128: Some of the code and tests have already been removed. This seems to be all that is left.

georgii-borovinskikh-sonarsource (approved):

Issue #4532, opened 2023-07-19
Populate Priority column in Hotspots list


georgii-borovinskikh-sonarsource:
Currently, we don't have data for Priority column (aka vulnerabilityProbability) extracted from rule definition json. For example, it's there in typescript-rules.json, but not when we unpack it into single-rule jsons

In order to save time and considering the fact that hotspots are no longer being updated, we decided to make a hardcoded list of priorities by rule key

PR #4570, opened 2023-07-21
Add data binding for priority column

georgii-borovinskikh-sonarsource:
* LocalHotspot now has Priority
* LocalHotspotsStore uses PriorityProvider to create LocalHotspot
* HotspotViewModel uses Priority from LocalHotspot
* HotspotsControl displays the new Priority of HotspotViewModel

Implements #4532

### DIFF
--- a/src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
+++ b/src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
@@ -132,10 +132,32 @@ public class LocalHotspotStoreTests
         eventListener.Events.Should().HaveCount(2).And.Subject.Last().Should()
             .BeEquivalentTo(new IssuesChangedEventArgs(Array.Empty<IAnalysisIssueVisualization>(), newHotspots));
     }
+    
+    [TestMethod]
+    public void UpdateForFile_NoServerHotspots_UsesDefaultReviewPriorityWhenUnmapped()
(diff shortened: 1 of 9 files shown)


src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
```diff
@@ -122,13 +126,44 @@ public void UpdateForFile_NoServerHotspots_AddsForDifferentFile()
...
-            new LocalHotspot(issueVis2));
+        VerifyContent(testSubject, 
+            new LocalHotspot(issueVis1, default),
+            new LocalHotspot(issueVis2, default));
         eventListener.Events.Should().HaveCount(2).And.Subject.Last().Should()
             .BeEquivalentTo(new IssuesChangedEventArgs(Array.Empty<IAnalysisIssueVisualization>(), newHotspots));
     }
 
+    [TestMethod]
+    public void UpdateForFile_NoServerHotspots_UsesReviewPriority()
```
duncanp-sonar, line 137: As discussed offline, if this test is just checking the correct review priority is being used then it can be much simpler: it doesn't need to call `RemoveForFile` twice, and it doesn't need to check the events that were raised. Same for the other new test.

duncanp-sonar (commented):
Looks good, just the comment about the test complexity.

duncanp-sonar (approved):

PR #4574, opened 2023-07-21
Show only unreviewed hotspots

### LINKED ISSUE #4550: Show only unreviewed hotspots
LocalHotspotStore should filter out suppressed hotspots so that they're not underlined by squiggly lines and are not shown in the hotspot list

ugras-ergun-sonarsource:
fixes #4550

### DIFF
--- a/src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
+++ b/src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
@@ -126,13 +126,13 @@ public class LocalHotspotStoreTests
 
         testSubject.UpdateForFile("file2", newHotspots);
 
-        VerifyContent(testSubject, 
+        VerifyContent(testSubject,
             new LocalHotspot(issueVis1, default),
             new LocalHotspot(issueVis2, default));
         eventListener.Events.Should().HaveCount(2).And.Subject.Last().Should()
             .BeEquivalentTo(new IssuesChangedEventArgs(Array.Empty<IAnalysisIssueVisualization>(), newHotspots));
     }
-    
+
     [TestMethod]
     public void UpdateForFile_NoServerHotspots_UsesDefaultReviewPriorityWhenUnmapped()
     {
@@ -157,7 +157,7 @@ public class LocalHotspotStoreTests
          * issue2 -> rule2 -> Medium
          * issue3 -> rule1 -> Low
          */
-        
+
         const string rule1 = "rule:s1";
         const string rule2 = "rule:s2";
         var issueVis1 = new Mock<IAnalysisIssueVisualization>();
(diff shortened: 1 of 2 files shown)


src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -172,12 +173,12 @@ private void RematchAll()
                     kvp => CreateLocalHotspots(kvp.Value.Select(localHotspot => localHotspot.Visualization)));
 
                 var visualizations = GetAll();
-                
+
                 if (!visualizations.Any())
```
georgii-borovinskikh-sonarsource, line 177: The issue could have been unreviewed before and became reviewed, but it won't get added to the event. TBF I think we don't use the payload of the event and always call GetAll, however I would still fix it so that it's correct or at least log a bug

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
```diff
@@ -336,6 +336,35 @@ public void Refresh_NewServerHotspots_ExistingHotspotsRematched()
             .BeEquivalentTo(new IssuesChangedEventArgs(issueVisualizations, issueVisualizations));
     }
 
+    [TestMethod]
+    public void GetAll_ReviewedServerHotspots_Filters()
+    {
+        var serverStoreMock = new Mock<IServerHotspotStore>();
+        var serverHotspot1 = CreateEmptyServerHotspot(status: "TO_REVIEW");
+        var serverHotspot2 = CreateEmptyServerHotspot(status: "REVIEWED", resolution: "ACKNOWLEDGED");
+        var serverHotspot3 = CreateEmptyServerHotspot(status: "REVIEWED", resolution: "FIXED"); //Expected to be filtered out
```
georgii-borovinskikh-sonarsource, line 345: It could also be marked as False Positive. Could be useful to add that to the test

georgii-borovinskikh-sonarsource (approved):

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -214,8 +215,8 @@ private LocalHotspot MatchAndConvert(IAnalysisIssueVisualization visualization)
             return new LocalHotspot(visualization, hotspotReviewPriorityProvider.GetPriority(visualization.RuleId) ?? DefaultPriority);
         }
 
-        private IEnumerable<LocalHotspot> FlattenMapping() => 
-            fileToHotspotsMapping.SelectMany(kvp => kvp.Value);
+        private IEnumerable<LocalHotspot> FlattenMapping() =>
```
duncanp-sonar, line 218: Probably worth renaming the method to make it clear it is doing more than just flattening
e.g. `FlattenAndFilterMapping`, or even `GetOpenHotspots` / `GetUnreviewedHotspots`.

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
georgii-borovinskikh-sonarsource: Or won't fix, can't remember

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
georgii-borovinskikh-sonarsource: + 1

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
duncanp-sonar: The UI options look like this - not sure what the stored values are though:

image

duncanp-sonar: [API docs](https://next.sonarqube.com/sonarqube/web_api/api/hotspots/search):

Status: REVIEWED or TO_REVIEW
Resolution: FIXED, SAFE, ACKNOWLEDGED.

So "SAFE" is the only resolution that isn't covered by tests.

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
```diff
@@ -336,6 +336,35 @@ public void Refresh_NewServerHotspots_ExistingHotspotsRematched()
             .BeEquivalentTo(new IssuesChangedEventArgs(issueVisualizations, issueVisualizations));
     }
 
+    [TestMethod]
```
duncanp-sonar, line 339: Would this be simpler as a data-driven test with parameters `status`, `resolution`, `shouldBeIncluded` ?

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
ugras-ergun-sonarsource: I think Open makes more sense because we're getting the ones reviewed and acknowledged.

duncanp-sonar (commented):
A couple of suggestions.

src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
georgii-borovinskikh-sonarsource: Actually, the event won't raise at all so this is an important issue

PR #4576, opened 2023-07-21
Show/hide hotspot list based on existence of issues

### LINKED ISSUE #4557: Local hotspot tool window should appear when local hotspots exist
### Description

Currently, the user has to manually open the local hotspot window to see any issues, so they have to know that it even exists.

The behaviour should like the taint tool window: if there are local hotspots issues, the window should be visible but not active i.e. not top-most.

**Taint window:**
image

duncanp-sonar:
Fixes #4557

### DIFF
--- a/src/ConnectedMode/ConnectedModePackage.cs
+++ b/src/ConnectedMode/ConnectedModePackage.cs
@@ -26,14 +26,13 @@ using Microsoft.VisualStudio.ComponentModelHost;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Threading;
 using SonarLint.VisualStudio.ConnectedMode.Hotspots;
+using SonarLint.VisualStudio.ConnectedMode.Install;
 using SonarLint.VisualStudio.ConnectedMode.ServerSentEvents;
 using SonarLint.VisualStudio.ConnectedMode.ServerSentEvents.Issue;
 using SonarLint.VisualStudio.ConnectedMode.Suppressions;
-using SonarLint.VisualStudio.ConnectedMode.Install;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Integration;
 using Task = System.Threading.Tasks.Task;
-using SonarLint.VisualStudio.ConnectedMode.Hotspots;
 
 namespace SonarLint.VisualStudio.ConnectedMode
 {
@@ -51,6 +50,7 @@ namespace SonarLint.VisualStudio.ConnectedMode
         private LocalSuppressionsChangedHandler localSuppressionsChangedHandler;
         private ImportBeforeInstallTrigger importBeforeInstallTrigger;
         private IHotspotDocumentClosedHandler hotspotDocumentClosedHandler;
+        private ILocalHotspotStoreMonitor hotspotStoreMonitor;
 
         protected override async Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
         {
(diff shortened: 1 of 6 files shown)


src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -75,6 +75,9 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
 
             hotspotDocumentClosedHandler = componentModel.GetService<IHotspotDocumentClosedHandler>();
 
+            hotspotStoreMonitor = componentModel.GetService<ILocalHotspotStoreMonitor>();
+            await hotspotStoreMonitor.InitializeAsync();
```
duncanp-sonar, line 79: Create the monitor in the ConnectedMode package, same as for the hotspot document handler.

src/ConnectedMode/Hotspots/ILocalHotspotStoreMonitor.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace SonarLint.VisualStudio.ConnectedMode.Hotspots
+{
+    /// <summary>
+    /// Marker interface for the component that listens for changes to the local hotspots
+    /// and triggers showing/hiding the local hotspots tool window
+    /// </summary>
+    public interface ILocalHotspotStoreMonitor
+    {
+        System.Threading.Tasks.Task InitializeAsync();
```
duncanp-sonar, line 29: The implementation has to be in `IssueViz.Security`because of the types used, but the class is created by the ConnectedMode package, so this interface has to be in the ConnectedMode project.

src/IssueViz.Security/Hotspots/LocalHotspotStoreMonitor.cs
```diff
@@ -0,0 +1,93 @@
...
+            this.threadHandling = threadHandling;
+        }
+
+        public async Task InitializeAsync()
+        {
+            localHotspotsStore.IssuesChanged += OnIssuesChanged;
+
+            await threadHandling.RunOnUIThread(() =>
+            {
+                vsMonitorSelection = (IVsMonitorSelection)serviceProvider.GetService(typeof(SVsShellMonitorSelection));
```
duncanp-sonar, line 65: This code is based on the implementation used to [show/hide the taint window](https://github.com/SonarSource/sonarlint-visualstudio/blob/7543d596f851c0d09b910da7996162d6c781531b/src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs#L81), although the thread handling is more correct in this version.

src/IssueViz.Security/IssueVizSecurityPackage.cs
```diff
@@ -48,8 +49,13 @@ tool window.
     [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid("D7D54E08-45E1-49A6-AA53-AF1CFAA6EBDC")]
     [ProvideMenuResource("Menus.ctmenu", 1)]
-    [ProvideToolWindow(typeof(HotspotsToolWindow), MultiInstances = false, Transient = true, Style = VsDockStyle.Tabbed, Window = VsWindowKindErrorList, Width = 700, Height = 250)]
+
+    [ProvideToolWindow(typeof(HotspotsToolWindow), MultiInstances = false, Transient = false, // Note: Transient must false when using ProvideToolWindowVisibility
+        Style = VsDockStyle.Tabbed, Window = VsWindowKindErrorList, Width = 700, Height = 250)]
+    [ProvideToolWindowVisibility(typeof(HotspotsToolWindow), LocalHotspotIssuesExistUIContext.GuidString)]
```
duncanp-sonar, line 55: Declare to VS that the visibility of the tool window is tied to the specific context.
Then we only need to set/clear the context and VS will do the rest.

georgii-borovinskikh-sonarsource (approved):

PR #4589, opened 2023-07-24
CAYC - tests

duncanp-sonar:
No product code changes

### DIFF
--- a/src/Education.UnitTests/Layout/Logical/RuleInfoTranslatorTests.cs
+++ b/src/Education.UnitTests/Layout/Logical/RuleInfoTranslatorTests.cs
@@ -84,7 +84,7 @@ namespace SonarLint.VisualStudio.Education.UnitTests.Layout.Logical
                         GetRuleInfo(new[] { new DescriptionSection("root_cause", null) }, issueType), null)
                     .ToList();
 
-            descriptionSections.Should().HaveCount(1);
+            descriptionSections.Should().ContainSingle();
             descriptionSections.Single().Should().BeOfType<RootCauseSection>().Which.Title.Should().Be(expectedTitle);
         }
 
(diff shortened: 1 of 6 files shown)


duncanp-sonar:
Low-value changes to get rid of [code smells in tests](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=4585&id=sonarlint-visualstudio&open=AYmHrNzNvmXtN3y4uMos) in the feature branch.

src/Education.UnitTests/Layout/Logical/RuleInfoTranslatorTests.cs
```diff
@@ -84,7 +84,7 @@ public void GetRuleDescriptionSections_RootCauseSectionSetsCorrectTitle(RuleIssu
                         GetRuleInfo(new[] { new DescriptionSection("root_cause", null) }, issueType), null)
                     .ToList();
 
-            descriptionSections.Should().HaveCount(1);
```
duncanp-sonar, line 87: The Fluent test analyzer thinks `HaveCount` should be used except if there is only one, in which case `ContainSingle` should be used. Doesn't seem particularly consistent...

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotTests.cs
```diff
@@ -50,6 +50,6 @@ public void Ctor_VisualizationIsNull_Throws()
                 null,
                 null));
 
-        test.Should().Throw<ArgumentNullException>().Which.Message.Should().Contain("visualization");
+        test.Should().Throw<ArgumentNullException>().WithMessage("*visualization*");
```
duncanp-sonar, line 53: Complains that `WithMessage` should be used, but that means we have to add wildcards to the string to get the same effect as `Contain`...

georgii-borovinskikh-sonarsource (approved):
Should we care about code smells in tests? Some of these are subjective as well. For example, I had the same issue about async methods raised by the IDE, but chose to ignore it as I think this Task.CompletedTask copy paste looks worse

PR #4590, opened 2023-07-24
CAYC - low-risk product cleanup

duncanp-sonar:


src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -186,8 +185,7 @@ private void RematchAll()
 
         private void InitializeServerHotspots(IList<SonarQubeHotspot> sonarQubeHotspots)
         {
-            serverHotspots = sonarQubeHotspots?.ToList() ?? new List<SonarQubeHotspot>();
-            unmatchedHotspots = serverHotspots.ToHashSet();
```
duncanp-sonar, line 190: The issue said to remove the field and define it in the methods that used it.
It was only used in one method, and it didn't actually need to be defined as a separate variable.

georgii-borovinskikh-sonarsource (approved):

Issue #4336, opened 2023-06-12
Update embedded JS/TS analyzer to 10.3.2.22047


alexander-kamushkin-sonarsource:
10.2.0 -> [10.3.0](https://github.com/SonarSource/SonarJS/releases/tag/10.3.0.21893) -> [10.3.1](https://github.com/SonarSource/SonarJS/releases/tag/10.3.1.21905) -> [10.3.2](https://github.com/SonarSource/SonarJS/releases/tag/10.3.2.22047)

duncanp-sonar:
Note: upgrade is currently blocked. Changes have been made to the eslintbridge API.

saberduck:
@duncanp-sonar We did a bugfix which rollbacks the breaking changes due to performance issues we encountered. I think you should be able to upgrade now

PR #4606, opened 2023-07-26
Update SonarJS to version 10.3.2.22047

georgii-borovinskikh-sonarsource:
Implements #4336

### DIFF
--- a/src/Education.UnitTests/SimpleRuleHelpXamlBuilderTests.cs
+++ b/src/Education.UnitTests/SimpleRuleHelpXamlBuilderTests.cs
@@ -151,7 +151,6 @@ namespace SonarLint.VisualStudio.Education.UnitTests
                     // Note: this is a quick way of getting the size of the document. Serializing the doc to a string
                     // and checking the length takes much longer (around 25 seconds)
                     var docLength = doc.ContentStart.DocumentStart.GetOffsetToPosition(doc.ContentEnd.DocumentEnd);
-                    // Console.WriteLine($"{jsonRuleInfo.FullRuleKey}: size = {docLength}");
                     docLength.Should().BeGreaterThan(30);
                 }
                 return true;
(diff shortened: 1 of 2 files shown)


georgii-borovinskikh-sonarsource:
Manually tested, but pending confirmation from JS team that this version is backwards compatible (as a precaution)

src/Education.UnitTests/SimpleRuleHelpXamlBuilderTests.cs
```diff
@@ -148,7 +151,7 @@ private static bool ProcessResource(string fullResourceName)
                     // Note: this is a quick way of getting the size of the document. Serializing the doc to a string
                     // and checking the length takes much longer (around 25 seconds)
                     var docLength = doc.ContentStart.DocumentStart.GetOffsetToPosition(doc.ContentEnd.DocumentEnd);
-                    Console.WriteLine($"{jsonRuleInfo.FullRuleKey}: size = {docLength}");
+                    // Console.WriteLine($"{jsonRuleInfo.FullRuleKey}: size = {docLength}");
```
duncanp-sonar, line 154: Does this need to be commented out?

duncanp-sonar (commented):
Looks good, just a question about the commented-out line in the test

src/Education.UnitTests/SimpleRuleHelpXamlBuilderTests.cs
georgii-borovinskikh-sonarsource: Forgot to revert it before the commit. TBH, this line is pretty useless and was preventing me from seeing the actual errors so at one point I was considering removing it. What do you think?

duncanp-sonar: If it's not helping, remove it.

PR #4610, opened 2023-07-27
Make local hotspot matching deterministic

### LINKED ISSUE #4602: Local hotspot suppression matching is not deterministic
### Repro
1. Open the sample bound project `SLVS_Bound_VS2019`
2. Open `Cpp_OldStyleProject_ConsoleApp\Hotspots\Hotspot1.js`
3. Mark one of the hotspots in the file as `Safe` on SonarCloud
4. Update the binding (to refresh the suppressions)
5. Check the file only shows one local hotspot, noting the line number of the reported issue.
6. Save the file to trigger re-analysis.

### Expected
The set of results does not change.

### Actual
The other issue is suppressed.
Each time the file is saved, the line number of the reported issue changes i.e. the issue being suppressed changes each time.

#### Screenshot - after first save
image

#### Screenshot - after second save
image

duncanp-sonar:
Fixes #4602

### DIFF
--- a/src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
+++ b/src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
@@ -288,6 +288,91 @@ public class LocalHotspotStoreTests
     }
 
     [TestMethod]
+    // Regression test for https://github.com/SonarSource/sonarlint-visualstudio/issues/4602
+    public void UpdateForFile_ServerHotspots_SameFileUpdate_ServerHotspotMatchingIsDeterministic()
+    {
+        // Note: the original bug was caused because *in some cases* removing items from a HashSet<>
+        // then adding them back in again in a different order changes the order in which the
+        // items in the HashSet<> are enumerated -> we get different matches next time.
+
+        var serverStoreMock = new Mock<IServerHotspotStore>();
+        var unusedButEssential = CreateEmptyServerHotspot(hotspotKey: "unused1");
+        var matchingServerHotspot1 = CreateEmptyServerHotspot(status: "REVIEWED", resolution: "SAFE", hotspotKey: "111");
+        var matchingServerHotspot2 = CreateEmptyServerHotspot(status: "TO_REVIEW", hotspotKey: "222");
+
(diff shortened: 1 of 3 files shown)


src/IssueViz.Security/Hotspots/LocalHotspotsStore.cs
```diff
@@ -235,5 +236,38 @@ private void NotifyIssuesChanged(IssuesChangedEventArgs args)
...
+        private static ISet<SonarQubeHotspot> CreateServerHotspotSet(IEnumerable<SonarQubeHotspot> serverHotspots = null)
+        {
+            return new SortedSet<SonarQubeHotspot>(
+                serverHotspots ?? Enumerable.Empty<SonarQubeHotspot>(),
+                new ServerHotspotComparer());
+        }
+
+        /// <summary>
+        /// Comparer to return server hotspots in a deterministic order.
+        /// Ordered by: StartLine, StartLineOffset, HotspotKey
```
duncanp-sonar, line 249: As discussed, this doesn't use the file name.

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
```diff
@@ -287,6 +287,91 @@ public void UpdateForFile_ServerHotspots_SameFileUpdate_MakesUnmatchedServerHots
         serverStoreMock.Verify(store => store.GetAll(), Times.Once);
     }
 
+    [TestMethod]
+    // Regression test for https://github.com/SonarSource/sonarlint-visualstudio/issues/4602
```
duncanp-sonar, line 291: The regression test was difficult to write and is quite fragile.

The other new test `UpdateForFile_MatchingDoesNotDependOnOrderServerHotspotsAreReturnedByTheServer()` is simpler. It's not an exact repro of the bug, but it also failed before I added the comparer fix.

I'm tempted to delete this regression test and rely on the other new test. What do you think?

duncanp-sonar:
I manually tested that this stops the flip-flopping of the suppressed issues when saving repeatedly.

georgii-borovinskikh-sonarsource (approved):

PR #4616, opened 2023-07-28
Fix flaky hotspot test

### LINKED ISSUE #4615: [Tests] Refresh_NewServerHotspots_ExistingHotspotsRematched is flaky
### Description

`Refresh_NewServerHotspots_ExistingHotspotsRematched ` is flaky following the fix for #4602.

As a result of fixing #4602, the order in which server hotspots are matched depends now depends on the data in the hotspot.

This means that tests that depend on the ordering of the server hotspots but that create multiple hotspots using code like `var serverHotspot1 = CreateEmptyServerHotspot();` will pass/fail depending on the random GUID that is used as the `hotspotKey`.

`Refresh_NewServerHotspots_ExistingHotspotsRematched` is the only test that has exhibited any flakiness.

duncanp-sonar:
Fixes #4615

### DIFF
--- a/src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
+++ b/src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
@@ -394,9 +394,12 @@ public class LocalHotspotStoreTests
     public void Refresh_NewServerHotspots_ExistingHotspotsRematched()
     {
         var serverStoreMock = new Mock<IServerHotspotStore>();
-        var serverHotspot1 = CreateEmptyServerHotspot();
-        var serverHotspot2 = CreateEmptyServerHotspot();
-        var serverHotspot3 = CreateEmptyServerHotspot();
+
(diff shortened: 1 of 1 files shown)


georgii-borovinskikh-sonarsource (approved):
LGTM. Just one question. Could this issue appear in the other tests as well? Would it be worth making at least one parameter, that is used by the comparer, non-optional when creating a mock server hotspot?

duncanp-sonar:
> LGTM. Just one question. Could this issue appear in the other tests as well? 

I checked the other tests and none of them were affected. This test was unusual in that (1) has multiple server hotspots and (2) has a mock matcher that matches any issue to any server hotspot.
Also, we've run the other tests multiple times and none of them have failed, so probably not. The failure rate of this test when I ran it in a loop is about 50%, so we'd probably have seen if any of the other tests are flaky by now.

> Would it be worth making at least one parameter, that is used by the comparer, non-optional when creating a mock server hotspot?

Good idea. That's much safer; we couldn't accidentally hit the same issue in the future. I'll make the `hotspotKey` required.

Issue #4621, opened 2023-07-31
Add new style resources


bigfluffycookie:
Add style resources for compliant and non compliant code diffs

PR #4622, opened 2023-07-31
Add new style resources for diff highlighting.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4621


src/Education/Commands/RuleHelpUserControl.xaml
```diff
@@ -233,6 +233,12 @@
                 <Setter Property="FontFamily" Value="Courier New" />
                 <Setter Property="FontSize" Value="{DynamicResource {x:Static vsshell:VsFonts.Environment90PercentFontSizeKey}}"/>
             </Style>
+            <Style x:Key="NonCompliant_Diff" TargetType="Span">
```
bigfluffycookie, line 236: I can't test out the colors yet so I just chose these. Good change they have to be changed but we will have to see. (Added a ticket as a reminder)

Issue #4542, opened 2023-07-19
Implement diffplex into code post translation


bigfluffycookie:
Get diff from non-compliant and compliant code snippets by using DiffPlex SideBySideDiffBuilder.Diff() function and add the resulting text back into the xaml

PR #4629, opened 2023-08-02
Implement diffplex in DiffTranslator

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4542


src/Education/XamlGenerator/DiffTranslator.cs
```diff
@@ -18,10 +18,81 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Text;
+using DiffPlex.DiffBuilder;
+using DiffPlex.DiffBuilder.Model;
+
 namespace SonarLint.VisualStudio.Education.XamlGenerator
 {
-    internal interface IDiffTranslator
+    public interface IDiffTranslator
```
bigfluffycookie, line 29: Needs to be public now as its a mef export

src/Education/XamlGenerator/DiffTranslator.cs
```diff
@@ -18,10 +18,81 @@
...
     }
+
+    [Export(typeof(IDiffTranslator))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class DiffTranslator : IDiffTranslator
+    {
+        private readonly IXamlWriterFactory xamlWriterFactory;
+
+        [ImportingConstructor]
+        public DiffTranslator(IXamlWriterFactory xamlWriterFactory)
```
bigfluffycookie, line 47: We could just make a new xamlWriter ourselves but I think it makes sense to just reuse this class

src/Education/XamlGenerator/DiffTranslator.cs
```diff
@@ -18,10 +18,81 @@
...
+                    writer.WriteEndElement();
+                }
+                else
+                {
+                    writer.WriteString(line.Text);
+                }
+
+                // The differ removes the next line element as it returns a list of lines.
+                // It needs to be added back for the xaml to render as intended.
+                if (i < lines.Count - 1)
```
bigfluffycookie, line 85: Not a huge fan of this but I cant remember how else to do this

src/Education/XamlGenerator/DiffTranslator.cs
```diff
@@ -18,10 +18,81 @@
...
+                {
+                    writer.WriteRaw("\n");
+                }
+            }
+
+            writer.Close();
+
+            // The xml writer converts \n to \r\n which is not needed here.
```
bigfluffycookie, line 93: There would be a setting in the XmlWriter during creation but I am unsure what other implications it would have so I think it is better to have this.

bigfluffycookie:
The pipeline failures have nothing to do with my changes, not sure why it is complaining.

ugras-ergun-sonarsource (approved):

PR #4634, opened 2023-08-02
Add IQualityProfileEvent

georgii-borovinskikh-sonarsource:
Add IQualityProfileEvent interface and an empty implementation
Add RuleSetEvent to SSEStreamReader

Implements #4633


duncanp-sonar:
Hhmmm - the Mend-only pipeline shouldn't be triggered for PRs. I'll disable it.

PR #4642, opened 2023-08-02
Add IQualityProfileUpdater definition

### LINKED ISSUE #4640: Add IQualityProfileUpdater interface
### Description

Add the updater interface.

To start with this will only need a single "update" method as we have decided to start with a coarse-grained refresh of all out-of-date Quality Profiles, rather than updating individual rules.

duncanp-sonar:
Fixes #4640

### DIFF
--- a/src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
+++ b/src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
@@ -1,0 +1,46 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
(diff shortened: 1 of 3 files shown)


src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <summary>
+        /// When in Connected Mode, ensures that all of the Quality Profiles
+        /// are up to date
+        /// </summary>
+        Task UpdateAsync(CancellationToken cancellationToken);
+    }
+
+    [Export(typeof(IQualityProfileUpdater))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class QualityProfileUpdater : IQualityProfileUpdater
```
duncanp-sonar, line 38: Add interface and skeleton implementation so we can work in parallel.

src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace SonarLint.VisualStudio.ConnectedMode.QualityProfiles
+{
+    internal interface IQualityProfileUpdater
+    {
+        /// <summary>
+        /// When in Connected Mode, ensures that all of the Quality Profiles
+        /// are up to date
+        /// </summary>
+        Task UpdateAsync(CancellationToken cancellationToken);
```
georgii-borovinskikh-sonarsource, line 33: Why do we pass a cancellation token here? It's not clear where it should come from in the case where we call Update based on SSE QP events.

BTW, did you look at the ICancellableActionRunner? If the point of this CancellationToken is to cancel an ongoing update before starting a new one, the ICancellableActionRunner is the abstraction that handles that.

duncanp-sonar: I put the cancellation token there mainly to remind us to think about cancellation.
I was thinking that we would use the `ICancellableActionRunner` so that if we get multiple events in a short space of time we reduce the number of calls to the server.

At this point, I'm not clear whether we'd handle cancelling in-progress requests inside this component or outside it. We can remove the cancellation token later if it turns out we don't need it.

georgii-borovinskikh-sonarsource: None of our updaters use cancellation tokens. I would prefer to do the other way around, add it only when we need it. But it's up to you.

PR #4643, opened 2023-08-02
[CAYC] Delete unused state

duncanp-sonar:
Delete unused internal state from `SolutionBindingOperation`
Add missing test method

### DIFF
--- a/src/ConnectedMode.UnitTests/Binding/BindingProcessImplTests.cs
+++ b/src/ConnectedMode.UnitTests/Binding/BindingProcessImplTests.cs
@@ -352,6 +352,19 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
             testSubject.InternalState.BindingOperationSucceeded.Should().BeTrue($"Initial state of {nameof(BindingProcessImpl.InternalState.BindingOperationSucceeded)} should be true");
         }
 
+        [TestMethod]
+        public void SaveRuleConfiguration_CallsSolutionBindingOperation()
+        {
+            var solutionBindingOperation = new Mock<ISolutionBindingOperation>();
+
+            var testSubject = CreateTestSubject(solutionBindingOperation: solutionBindingOperation.Object);
+
+            testSubject.SaveRuleConfiguration(CancellationToken.None);
+
+            solutionBindingOperation.Verify(x => x.SaveRuleConfiguration(testSubject.InternalState.BindingConfigs.Values, It.IsAny<CancellationToken>()),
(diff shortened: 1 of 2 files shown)


src/ConnectedMode/Binding/SolutionBindingOperation.cs
```diff
@@ -47,13 +44,6 @@ internal SolutionBindingOperation(IFileSystem fileSystem)
             this.fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
         }
 
-        #region State
-
-        internal /*for testing purposes*/ IReadOnlyDictionary<Language, IBindingConfig> RuleSetsInformationMap => 
-            new ReadOnlyDictionary<Language, IBindingConfig>(bindingConfigInformationMap);
-
-        #endregion
```
duncanp-sonar, line 55: This dictionary is no longer used so I deleted it.

src/ConnectedMode.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -371,10 +385,9 @@ public void EmitBindingCompleteMessage()
             exclusionSettingsStorage ??= Mock.Of<IExclusionSettingsStorage>();
             logger ??= new TestLogger(logToConsole: true);
             languagesToBind ??= Language.KnownLanguages;
+            solutionBindingOperation ??= Mock.Of<ISolutionBindingOperation>();
 
-            var slnBindOperation = new SolutionBindingOperation();
```
duncanp-sonar, line 375: The test was using the real type rather than a mock so I changed it to a mock.
Having done that, none of tests failed, so I realised there must be a missing test case.

src/ConnectedMode.UnitTests/Binding/BindingProcessImplTests.cs
```diff
@@ -352,6 +352,19 @@ public void EmitBindingCompleteMessage()
             testSubject.InternalState.BindingOperationSucceeded.Should().BeTrue($"Initial state of {nameof(BindingProcessImpl.InternalState.BindingOperationSucceeded)} should be true");
         }
 
+        [TestMethod]
+        public void SaveRuleConfiguration_CallsSolutionBindingOperation()
```
duncanp-sonar, line 356: Previously there were no tests for `SaveRuleConfiguration`.
It's a very simple method so it only needs this one test.

georgii-borovinskikh-sonarsource (approved):

Issue #4646, opened 2023-08-03
Add QualityProfileServerEventListener


georgii-borovinskikh-sonarsource:
Should listen for QP events and call the updater

PR #4650, opened 2023-08-03
Add QualityProfileServerEventsListener

georgii-borovinskikh-sonarsource:
Implements #4646

### DIFF
--- a/src/ConnectedMode.UnitTests/ServerSentEvents/QualityProfile/QualityProfileServerEventsListenerTests.cs
+++ b/src/ConnectedMode.UnitTests/ServerSentEvents/QualityProfile/QualityProfileServerEventsListenerTests.cs
@@ -1,0 +1,81 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using SonarLint.VisualStudio.ConnectedMode.QualityProfiles;
+using SonarLint.VisualStudio.ConnectedMode.ServerSentEvents.QualityProfile;
+using SonarLint.VisualStudio.TestInfrastructure;
+using SonarQube.Client.Models.ServerSentEvents.ClientContract;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.ServerSentEvents.QualityProfile;
+
+[TestClass]
+public class QualityProfileServerEventsListenerTests
+{
+    [TestMethod]
+    public void MefCtor_CheckExports()
+    {
+        MefTestHelpers.CheckTypeCanBeImported<QualityProfileServerEventsListener, IQualityProfileServerEventsListener>(
+            MefTestHelpers.CreateExport<IQualityProfileServerEventSource>(),
+            MefTestHelpers.CreateExport<IQualityProfileUpdater>());
+    }
+
+    [TestMethod]
+    public void MefCtor_CheckIsSingleton()
+    {
+        MefTestHelpers.CheckIsSingletonMefComponent<QualityProfileServerEventsListener>();
+    }
+
+    [TestMethod]
+    public void ListenAsync_CallsUpdaterForEachEventUntilNullEvent()
+    {
+        var events = new[]
+        {
+            Mock.Of<IQualityProfileEvent>(),
+            Mock.Of<IQualityProfileEvent>(),
+            Mock.Of<IQualityProfileEvent>(),
+            null
+        };
+
+        var tcs = Enumerable.Range(0, events.Length).Select(_ => new TaskCompletionSource<IQualityProfileEvent>()).ToArray();
+
+        var eventSourceMock = new Mock<IQualityProfileServerEventSource>();
+        eventSourceMock.Setup(x => x.GetNextEventOrNullAsync()).Returns(new Queue<Task<IQualityProfileEvent>>(tcs.Select(x => x.Task)).Dequeue);
+        
+        var updaterMock = new Mock<IQualityProfileUpdater>();
+
+        var testSubject = new QualityProfileServerEventsListener(eventSourceMock.Object, updaterMock.Object);
+        var task = testSubject.ListenAsync();
+
+        for (var index = 0; index < tcs.Length; index++)
+        {
+            var taskCompletionSource = tcs[index];
+            taskCompletionSource.SetResult(events[index]);
+            task.IsCompleted.Should().Be(index == tcs.Length - 1);
+        }
+        
+        task.GetAwaiter().GetResult();
+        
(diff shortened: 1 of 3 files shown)


src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -67,6 +69,9 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
             issueServerEventsListener = componentModel.GetService<IIssueServerEventsListener>();
             issueServerEventsListener.ListenAsync().Forget();
 
+            qualityProfileServerEventsListener = componentModel.GetService<IQualityProfileServerEventsListener>();
+            qualityProfileServerEventsListener.ListenAsync().Forget();
```
duncanp-sonar, line 73: This instance isn't disposed (unlike the equivalent `IssueServerEventsListener`).
Does it need to be?

src/ConnectedMode/ServerSentEvents/QualityProfile/QualityProfileServerEventsListener.cs
```diff
@@ -0,0 +1,57 @@
...
+        [ImportingConstructor]
+        public QualityProfileServerEventsListener(IQualityProfileServerEventSource eventSource, IQualityProfileUpdater updater)
+        {
+            this.eventSource = eventSource;
+            this.updater = updater;
+        }
+
+        public async Task ListenAsync()
+        {
+            while (await eventSource.GetNextEventOrNullAsync() != null)
```
duncanp-sonar, line 51: Probably worth commenting that the eventSource will return null when it is disposed (same as in the `IssueServerEventListener`).

georgii-borovinskikh-sonarsource: added the comment

src/ConnectedMode/ConnectedModePackage.cs
duncanp-sonar: That's not a helpful reply. Why not?

src/ConnectedMode.UnitTests/ServerSentEvents/QualityProfile/QualityProfileServerEventsListenerTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        var task = testSubject.ListenAsync();
+
+        for (var index = 0; index < tcs.Length; index++)
+        {
+            var taskCompletionSource = tcs[index];
+            taskCompletionSource.SetResult(events[index]);
+            task.IsCompleted.Should().Be(index == tcs.Length - 1);
+        }
+        
+        task.GetAwaiter().GetResult();
```
duncanp-sonar, line 76: Is this needed? Either the task will have completed by the time we get here, or the test will have asserted.

src/ConnectedMode/ConnectedModePackage.cs
georgii-borovinskikh-sonarsource: There's nothing to dispose there. It's not responsible for disposing it's dependencies.

src/ConnectedMode.UnitTests/ServerSentEvents/QualityProfile/QualityProfileServerEventsListenerTests.cs
georgii-borovinskikh-sonarsource: What if it throws? We won't see the error otherwise

georgii-borovinskikh-sonarsource: I could  add another assertion, that it didn't complete with an exception, but it's essentially the same thing as awaiting the task

src/ConnectedMode.UnitTests/ServerSentEvents/QualityProfile/QualityProfileServerEventsListenerTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        var testSubject = new QualityProfileServerEventsListener(eventSourceMock.Object, updaterMock.Object);
+        var task = testSubject.ListenAsync();
+
+        for (var index = 0; index < tcs.Length; index++)
+        {
+            var taskCompletionSource = tcs[index];
+            taskCompletionSource.SetResult(events[index]);
+            task.IsCompleted.Should().Be(index == tcs.Length - 1);
+        }
```
duncanp-sonar, line 74: Reading this test made my brain hurt so I debugged through it so I could understand it better, and the debugging experience was weird too.

It feels like there are more data structures than are really needed, and that there isn't a clean separation between the setup code and the execution (i.e. setting the `taskCompletionSource` result in the `for` loop after execution has started).

I had a go at writing it in an alternative way and came up with the following:

```
    [TestMethod]
    public async Task ListenAsync_CallsUpdaterForEachEventUntilNullEvent()
    {
        var currentEventIndex = 0;
        var eventTasks = new Task<IQualityProfileEvent>[]
        {
            Task.FromResult(Mock.Of<IQualityProfileEvent>()),
            Task.FromResult(Mock.Of<IQualityProfileEvent>()),
            Task.FromResult(Mock.Of<IQualityProfileEvent>()),
            Task.FromResult<IQualityProfileEvent>(null),
        };

        var eventSourceMock = new Mock<IQualityProfileServerEventSource>();
        eventSourceMock.Setup(x => x.GetNextEventOrNullAsync()).Returns(() => eventTasks[currentEventIndex++]);

        var updaterMock = new Mock<IQualityProfileUpdater>();

        var testSubject = new QualityProfileServerEventsListener(eventSourceMock.Object, updaterMock.Object);
        await testSubject.ListenAsync();

        eventSourceMock.Verify(x => x.GetNextEventOrNullAsync(), Times.Exactly(eventTasks.Length));
        updaterMock.Verify(x => x.UpdateAsync(), Times.Exactly(eventTasks.Length - 1));
    }
```

This version doesn't have an equivalent of the check `task.IsCompleted.Should().Be(...)`,
However, I don't think it needs it: checking the correct number of server and updater calls should be enough.

What do you think?

src/ConnectedMode/ConnectedModePackage.cs
duncanp-sonar: This class is logically similar to `IssueServerEventsListener`.
The `IssueServerEventsListener` isn't responsible for disposing its dependencies either, but we do dispose it.

Given that the classes are logically similar, why do we dispose one and not the other?

georgii-borovinskikh-sonarsource: It only disposes of it's internal cancellation token. This class doesn't need that.

duncanp-sonar: (it's fine for them to be different, but there should be a reason, and we should record it somewhere e.g. a short comment in the code).

src/ConnectedMode.UnitTests/ServerSentEvents/QualityProfile/QualityProfileServerEventsListenerTests.cs
georgii-borovinskikh-sonarsource: The point of checking IsCompleted was to indicate/verify that it's a long running action. If you think that it's not worth making the test more complicated to test that, then your test is better. I'm personally not sure if it's important. But tbf, I also not fully understand why my test is confusing

src/ConnectedMode/ConnectedModePackage.cs
georgii-borovinskikh-sonarsource: But isn't it clear from just looking at the class that there's nothing to dispose?

src/ConnectedMode.UnitTests/ServerSentEvents/QualityProfile/QualityProfileServerEventsListenerTests.cs
georgii-borovinskikh-sonarsource: Found a way to do it with only one assertion

georgii-borovinskikh-sonarsource: >The whole test is run on a single thread since it's using completed tasks.

georgii-borovinskikh-sonarsource: But I already changed it

duncanp-sonar (approved):

PR #4652, opened 2023-08-04
Add OutOfDateQPFinder

### LINKED ISSUE #4651: Implement IOutOfDateQPFinder
### Description

Component to detect which profiles need to be updated

Prep for #4641

duncanp-sonar:
Fixes #4651

### DIFF
--- a/src/ConnectedMode.UnitTests/QualityProfiles/OutOfDateQPFinderTests.cs
+++ b/src/ConnectedMode.UnitTests/QualityProfiles/OutOfDateQPFinderTests.cs
@@ -1,0 +1,193 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using SonarLint.VisualStudio.ConnectedMode.QualityProfiles;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.TestInfrastructure;
+using SonarQube.Client;
+using SonarQube.Client.Models;
+using Task = System.Threading.Tasks.Task;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.QualityProfiles
+{
+    [TestClass]
+    public class OutOfDateQPFinderTests
+    {
+        private static readonly Uri ValidUri = new Uri("http://server");
+        private static readonly SonarQubeQualityProfile ValidServerResponse = new SonarQubeQualityProfile("any key", "any name", "any language", true, DateTime.UtcNow);
+        private static readonly ApplicableQualityProfile ValidApplicableQualityProfile = new ApplicableQualityProfile
+        {
+            ProfileKey = "any", ProfileTimestamp = DateTime.UtcNow,
+        };
+
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+            => MefTestHelpers.CheckTypeCanBeImported<OutOfDateQPFinder, IOutOfDateQPFinder>(
+                MefTestHelpers.CreateExport<IConfigurationProvider>(),
+                MefTestHelpers.CreateExport<ISonarQubeService>(),
+                MefTestHelpers.CreateExport<ILogger>());
+
+        [TestMethod]
+        public void MefCtor_CheckIsSingleton()
+            => MefTestHelpers.CheckIsSingletonMefComponent<OutOfDateQPFinder>();
+
+        [TestMethod]
+        [DataRow("project1", null, "ts")] // Null org key should be handled correctly
+        [DataRow("project2", "my org", "secrets")] // Null org key should be handled correctly
+        public async Task GetAsync_CorrectParametersPassedToServer(string projectKey, string orgKey, string languageKey)
+        {
+            var expectedLanguage = Language.GetLanguageFromLanguageKey(languageKey);
+            var org = orgKey == null ? null :  new SonarQubeOrganization(orgKey, "org name");
+
+            var boundProject = new BoundSonarQubeProject(ValidUri, projectKey, "any name", null, org)
+            {
+                Profiles = new Dictionary<Language, ApplicableQualityProfile>
+                {
+                    { expectedLanguage, ValidApplicableQualityProfile}
+                }
+            };
+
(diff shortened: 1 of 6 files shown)


src/ConnectedMode/QualityProfiles/OutOfDateQPFinder.cs
```diff
@@ -0,0 +1,115 @@
...
+
+namespace SonarLint.VisualStudio.ConnectedMode.QualityProfiles
+{
+    internal interface IOutOfDateQPFinder
+    {
+        /// <summary>
+        /// When in Connected Mode, returns the languages for which the Quality Profile
+        /// needs to be updated
+        /// </summary>
+        Task<IEnumerable<Language>> GetAsync(CancellationToken cancellationToken);
```
duncanp-sonar, line 40: Separate component to find the list of QPs that are out of date

src/ConnectedMode/QualityProfiles/OutOfDateQPFinder.cs
```diff
@@ -0,0 +1,115 @@
...
+            Debug.Assert(profile.Value != null, "Not expecting profile to be null. Language: " + profile.Key);
+
+            var serverProfileInfo = await sonarQubeService.GetQualityProfileAsync(config.Project.ProjectKey,
+                config.Project.Organization?.Key,
+                profile.Key.ServerLanguage, CancellationToken.None);
+
+            return HasProfileChanged(serverProfileInfo, profile.Value);
+        }
+
+        private bool HasProfileChanged(SonarQubeQualityProfile newProfileInfo, ApplicableQualityProfile oldProfileInfo)
```
duncanp-sonar, line 94: There used to be a component that did a similar thing that was part of the now-deleted `ErrorListInfoBarController`. This method uses the same logic as [the old code](https://github.com/SonarSource/sonarlint-visualstudio/blob/75f3826e72a584818070801a1d5c7808cb4a8aa0/src/Integration/LocalServices/ErrorListInfoBarController.cs#L741).

src/Integration/Resources/Strings.resx
```diff
@@ -447,34 +447,6 @@ This is the general format to use when writing errors to output window or when t
 {0} - service
 {1} - http error code</comment>
   </data>
-  <data name="SonarLintProfileCheck" xml:space="preserve">
```
duncanp-sonar, line 450: Miscellaneous: deleted redundant strings that were used by the old QP comparison code, but that weren't cleaned up when that code was deleted.

src/ConnectedMode/QualityProfiles/QualityProfileExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.ConnectedMode.QualityProfiles
+{
+    internal static class QualityProfileExtensions
+    {
+        public static void LogQPVerbose(this ILogger logger, Language language, string text) => logger.LogVerbose($"[QualityProfile][{language.Name}] {text}");
+
+        public static void LogQPVerbose(this ILogger logger, string text) => logger.LogVerbose("[QualityProfile] " + text);
```
duncanp-sonar, line 30: Miscellaneous: helper methods so we get consistent prefixes across different QP classes.
Same as for the [MigrationExtensions](https://github.com/SonarSource/sonarlint-visualstudio/blob/154183e913b6a4f2740b601e171835035f20b641/src/ConnectedMode/Migration/MigrationExtensions.cs#L27).

src/ConnectedMode/QualityProfiles/OutOfDateQPFinder.cs
duncanp-sonar: Currently, it returns the list of languages that need to be updated.
I was debating whether to return the server timestamp too (since we'll want to update the contents of the `binding.config` file with the updated timestamp). However, I decided to leave it for now and add it later if required.

src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
```diff
@@ -37,10 +39,27 @@ internal interface IQualityProfileUpdater
...
-            // TODO - implement
-            return Task.CompletedTask;
+            var languagesToUpdate = await outOfDateQPFinder.GetAsync(CancellationToken.None);
+
+            if (!languagesToUpdate.Any())
+            {
+                logger.LogQPVerbose("No updates required");
+            }
+
+            // TODO - update and save changed profiles
```
duncanp-sonar, line 62: The code that does the "fetch and save" during binding is in [BindingProcessImpl](https://github.com/SonarSource/sonarlint-visualstudio/blob/154183e913b6a4f2740b601e171835035f20b641/src/ConnectedMode/Binding/BindingProcessImpl.cs#L101). I'm looking at how to refactor/reuse that code so it can be used both for a new binding and for updating QPs.

src/ConnectedMode/QualityProfiles/OutOfDateQPFinder.cs
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            return languagesToUpdate;
+        }
+
+        private async Task<bool> IsUpdateRequiredAsync(BindingConfiguration config, KeyValuePair<Language, ApplicableQualityProfile> profile, CancellationToken cancellationToken)
+        {
+            Debug.Assert(profile.Value != null, "Not expecting profile to be null. Language: " + profile.Key);
+
+            var serverProfileInfo = await sonarQubeService.GetQualityProfileAsync(config.Project.ProjectKey,
```
georgii-borovinskikh-sonarsource, line 87: There's a search method that returns the list of all QP for the project. Using that would mean that if we find out we actually want to update a QP, we would still have to call Get for that specific QP, but considering the fact that QPs don't change often, on average we would make 1 request per timed update, rather than making 1 for each language

src/ConnectedMode/QualityProfiles/QualityProfileExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Integration;
+
+namespace SonarLint.VisualStudio.ConnectedMode.QualityProfiles
+{
+    internal static class QualityProfileExtensions
```
georgii-borovinskikh-sonarsource, line 26: These are ILogger extensions, so the class name is misleading

src/ConnectedMode/QualityProfiles/OutOfDateQPFinder.cs
georgii-borovinskikh-sonarsource: Actually, since we're not using the info we fetched here and instead change the QP later (presumably making another Get for each changed QP?) using the search method makes even more sense

georgii-borovinskikh-sonarsource: https://next.sonarqube.com/sonarqube/web_api/api/qualityprofiles/search

src/ConnectedMode.UnitTests/QualityProfiles/OutOfDateQPFinderTests.cs
```diff
@@ -0,0 +1,193 @@
...
+                MefTestHelpers.CreateExport<ISonarQubeService>(),
+                MefTestHelpers.CreateExport<ILogger>());
+
+        [TestMethod]
+        public void MefCtor_CheckIsSingleton()
+            => MefTestHelpers.CheckIsSingletonMefComponent<OutOfDateQPFinder>();
+
+        [TestMethod]
+        [DataRow("project1", null, "ts")] // Null org key should be handled correctly
+        [DataRow("project2", "my org", "secrets")] // Null org key should be handled correctly
```
georgii-borovinskikh-sonarsource, line 58: Non null here?

georgii-borovinskikh-sonarsource (changes requested):
There's potentially a more efficient way to fetch the state of all QPs from the server

Issue #4665, opened 2023-08-07
Fix subpiece issue


bigfluffycookie:
There can be a case where a line has no subpieces even though it has text. In that case line text still needs to be written.

PR #4666, opened 2023-08-07
Fix problem with subpieces

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4665

### DIFF
--- a/src/Education/XamlGenerator/DiffTranslator.cs
+++ b/src/Education/XamlGenerator/DiffTranslator.cs
@@ -21,6 +21,7 @@
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Text;
+using System.Xml;
 using DiffPlex.DiffBuilder;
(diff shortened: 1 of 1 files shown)


src/Education/XamlGenerator/DiffTranslator.cs
```diff
@@ -109,5 +97,30 @@ private string HighlightLines(List<DiffPiece> lines, StyleResourceNames style, S
...
+        {
+            if (line.SubPieces.Count == 0)
+            {
+                writer.WriteString(line.Text);
+                return;
+            }
+
+            foreach (var subPiece in line.SubPieces)
+            {
+                if (subPiece.Type == ChangeType.Imaginary) { continue; }
```
bigfluffycookie, line 111: Imaginary changes are a bit strange, I can't quite explain what it means but whenever a subpiece is imaginary its text is null in which case we should ignore it.

PR #4677, opened 2023-08-08
(1) Modify QualityProfileDownloader to be MEF-creatable

### LINKED ISSUE #4676: Make IQualityProfileDownloader MEF-importable
### Description

`QualityProfileDownloader` is currently created directly by the `BindingProcessFactory`. It needs to be MEF-importable so it can be used by the timed update trigger too.

However, that means a couple of other types needs to be made MEF-creatable too e.g. `IBindingConfigProvider`

duncanp-sonar:
Fixes #4676

### DIFF
--- a/src/ConnectedMode.UnitTests/Binding/CompositeBindingConfigProviderTests.cs
+++ b/src/ConnectedMode.UnitTests/Binding/CompositeBindingConfigProviderTests.cs
@@ -26,6 +26,9 @@ using System.Threading.Tasks;
 using SonarLint.VisualStudio.ConnectedMode.Binding;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Integration;
+using SonarLint.VisualStudio.TestInfrastructure;
+using SonarQube.Client;
 using SonarQube.Client.Models;
 
 namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
@@ -34,34 +37,27 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
     public class CompositeBindingConfigProviderTests
     {
         [TestMethod]
-        public void Ctor_InvalidArgs()
-        {
-            // 1. No config providers supplied
-            Action act = () => new CompositeBindingConfigProvider();
-            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("providers");
+        public void MefCtor_CheckIsExported()
+            => MefTestHelpers.CheckTypeCanBeImported<CompositeBindingConfigProvider, IBindingConfigProvider>(
+                MefTestHelpers.CreateExport<ISonarQubeService>(),
+                MefTestHelpers.CreateExport<ILogger>());
 
-            // 2. Null provider supplied
-            var providerMock = new Mock<IBindingConfigProvider>();
-            act = () => new CompositeBindingConfigProvider(providerMock.Object, null);
-            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("providers");
-        }
+        [TestMethod]
+        public void MefCtor_CheckIsSingleton()
+            => MefTestHelpers.CheckIsSingletonMefComponent<CompositeBindingConfigProvider>();
 
 
         [TestMethod]
-        public void Ctor_ValidArgs()
+        public void Ctor_PublicConstructor_ContainsExpectedBindingConfigProviders()
         {
-            // Arrange
-            var providerMock1 = new Mock<IBindingConfigProvider>();
-            var providerMock2 = new Mock<IBindingConfigProvider>();
-
             // Act
-            var testSubject = new CompositeBindingConfigProvider(
-                providerMock1.Object,
-                providerMock2.Object, providerMock2.Object); // duplicate should be ignored
+            var testSubject = new CompositeBindingConfigProvider(Mock.Of<ISonarQubeService>(), Mock.Of<ILogger>());
 
             // Assert
             testSubject.Providers.Count().Should().Be(2);
-            testSubject.Providers.Should().BeEquivalentTo(providerMock1.Object, providerMock2.Object);
+            testSubject.Providers.Select(x => x.GetType()).Should().BeEquivalentTo(
+                typeof(NonRoslynBindingConfigProvider),
+                typeof(CSharpVBBindingConfigProvider));
         }
 
         [TestMethod]
--- a/src/ConnectedMode.UnitTests/Binding/NonRoslynBindingConfigProviderTests.cs
+++ b/src/ConnectedMode.UnitTests/Binding/NonRoslynBindingConfigProviderTests.cs
@@ -38,17 +38,9 @@ namespace SonarLint.VisualStudio.ConnectedMode.Binding.UnitTests
         private static readonly Language[] SupportedLanguages = { Language.C, Language.Cpp };
 
         [TestMethod]
-        public void Ctor_NullSupportedLanguages_ArgumentNullException()
-        {
-            Action act = () => new NonRoslynBindingConfigProvider(null, Mock.Of<ISonarQubeService>(), Mock.Of<ILogger>());
-
-            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("supportedLanguages");
-        }
-
-        [TestMethod]
         public void Ctor_NullService_ArgumentNullException()
         {
-            Action act = () => new NonRoslynBindingConfigProvider(SupportedLanguages, null, Mock.Of<ILogger>());
+            Action act = () => new NonRoslynBindingConfigProvider(null, Mock.Of<ILogger>());
 
             act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("sonarQubeService");
         }
@@ -56,7 +48,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Binding.UnitTests
         [TestMethod]
         public void Ctor_NullLogger_ArgumentNullException()
         {
-            Action act = () => new NonRoslynBindingConfigProvider(SupportedLanguages, Mock.Of<ISonarQubeService>(), null);
+            Action act = () => new NonRoslynBindingConfigProvider(Mock.Of<ISonarQubeService>(), null);
 
             act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("logger");
         }
@@ -266,7 +258,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Binding.UnitTests
 
         private static NonRoslynBindingConfigProvider CreateTestSubject(Mock<ISonarQubeService> serviceMock, TestLogger testLogger)
         {
-            return new NonRoslynBindingConfigProvider(SupportedLanguages, serviceMock.Object, testLogger);
+            return new NonRoslynBindingConfigProvider(serviceMock.Object, testLogger, SupportedLanguages);
         }
     }
 }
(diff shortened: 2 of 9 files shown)


duncanp-sonar:
Previously the BindingProcessFactory directly created the `QualityProfileDownloader`, which meant it also needed to manually create the `CompositeBindingProvider`, the `CSharpVBBindingConfigProvider` and the `NonBindingConfigProvider`.

I've made the `QualityProfileDownloader` and `CompositeBindingProvider` MEF-creatable and simplified the construction of the other classes.


Historical note: the `BindingProcessFactory` was only introduced recently as part of the first "unintrusive Connected Mode" work for v7.0. Before that, the binding process directly created the classes. The factory was introduced as part of a series of refactorings to make the binding process more unit-testable.

src/ConnectedMode.UnitTests/Binding/CompositeBindingConfigProviderTests.cs
```diff
@@ -34,34 +37,27 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
     public class CompositeBindingConfigProviderTests
     {
         [TestMethod]
-        public void Ctor_InvalidArgs()
```
duncanp-sonar, line 37: The class is now MEF-created, the arguments will never be null -> deleted the redundant tests.

src/ConnectedMode.UnitTests/Binding/CompositeBindingConfigProviderTests.cs
```diff
@@ -34,34 +37,27 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
...
-            // 1. No config providers supplied
-            Action act = () => new CompositeBindingConfigProvider();
-            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("providers");
-
-            // 2. Null provider supplied
-            var providerMock = new Mock<IBindingConfigProvider>();
-            act = () => new CompositeBindingConfigProvider(providerMock.Object, null);
-            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("providers");
-        }
+        public void MefCtor_CheckIsExported()
```
duncanp-sonar, line 40: Added standard new MEF tests

src/ConnectedMode.UnitTests/Binding/CompositeBindingConfigProviderTests.cs
```diff
@@ -34,34 +37,27 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
...
+                MefTestHelpers.CreateExport<ILogger>());
+
+        [TestMethod]
+        public void MefCtor_CheckIsSingleton()
+            => MefTestHelpers.CheckIsSingletonMefComponent<CompositeBindingConfigProvider>();
 
 
         [TestMethod]
-        public void Ctor_ValidArgs()
+        public void Ctor_PublicConstructor_ContainsExpectedBindingConfigProviders()
```
duncanp-sonar, line 51: New test that the composite uses the expected config providers.

src/ConnectedMode.UnitTests/Binding/NonRoslynBindingConfigProviderTests.cs
```diff
@@ -37,26 +37,18 @@ public class NonRoslynBindingConfigProviderTests
     {
         private static readonly Language[] SupportedLanguages = { Language.C, Language.Cpp };
 
-        [TestMethod]
-        public void Ctor_NullSupportedLanguages_ArgumentNullException()
```
duncanp-sonar, line 41: The supported languages are no longer set in the public constructor, so I removed the null check

src/ConnectedMode.UnitTests/Binding/NonRoslynBindingConfigProviderTests.cs
```diff
@@ -37,26 +37,18 @@ public class NonRoslynBindingConfigProviderTests
...
-        {
-            Action act = () => new NonRoslynBindingConfigProvider(null, Mock.Of<ISonarQubeService>(), Mock.Of<ILogger>());
-
-            act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("supportedLanguages");
-        }
-
         [TestMethod]
         public void Ctor_NullService_ArgumentNullException()
         {
-            Action act = () => new NonRoslynBindingConfigProvider(SupportedLanguages, null, Mock.Of<ILogger>());
```
duncanp-sonar, line 51: Removed the `SupportedLanguages` argument since it is no longer required in the public constructor.

src/ConnectedMode.UnitTests/QualityProfiles/QualityProfileDownloaderTests.cs
```diff
@@ -255,8 +254,8 @@ public async Task UpdateAsync_SavesConfiguration()
                 sonarQubeService ?? Mock.Of<ISonarQubeService>(),
                 bindingConfigProvider ?? Mock.Of<IBindingConfigProvider>(),
                 configurationPersister ?? new DummyConfigPersister(),
-                solutionBindingOperation ?? Mock.Of<ISolutionBindingOperation>(),
                 logger ?? new TestLogger(logToConsole: true),
+                solutionBindingOperation ?? Mock.Of<ISolutionBindingOperation>(),
```
duncanp-sonar, line 258: Re-ordered the parameters: the parameters in the public constructor come first, with the extra ones that are only set in the internal constructor set last.

The `ISolutionBindingOperation` is now a non-public parameter so it is set at the end.

src/ConnectedMode/Binding/CompositeBindingConfigProvider.cs
```diff
@@ -20,31 +20,39 @@
...
 
 namespace SonarLint.VisualStudio.ConnectedMode.Binding
 {
-    public class CompositeBindingConfigProvider : IBindingConfigProvider
+    /// <summary>
+    /// Composite that coordinates turning server Quality Profile information into a set of config files
+    /// for a particular language
+    /// </summary>
+    [Export(typeof(IBindingConfigProvider))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
```
duncanp-sonar, line 41: Now MEF-creatable so it can be imported by the `QualityProfileDownloader`.

src/ConnectedMode/Binding/CompositeBindingConfigProvider.cs
```diff
@@ -20,31 +20,39 @@
...
-            }
-            if (providers.Any(p => p == null))
-            {
-                throw new ArgumentNullException(nameof(providers));
-            }
+        [ImportingConstructor]
+        public CompositeBindingConfigProvider(ISonarQubeService sonarQubeService, ILogger logger)
+            : this(
+                  new CSharpVBBindingConfigProvider(sonarQubeService, logger),
+                  new NonRoslynBindingConfigProvider(sonarQubeService, logger))
```
duncanp-sonar, line 50: I could have made the `CSharpVBBindingConfigProvider` and `NonRoslynBindingConfigProvider` providers MEF-creatable and done an `ImportMany` here, but it didn't seem worth it.

It would have meant defining a new interface and writing more MEF tests for no real gain.
The list of config providers we support is fixed, the `CompositeBindingConfigProvider` is the only thing that needs to use them, and they are easy to create. It was simpler just to make the composite provider responsible for creating them.

If we do need to add other types of config provider in the future it would be easy enough to change later.

src/ConnectedMode/Binding/CompositeBindingConfigProvider.cs
```diff
@@ -20,31 +20,39 @@
...
 using System.Threading.Tasks;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Integration;
+using SonarQube.Client;
 using SonarQube.Client.Models;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Binding
 {
-    public class CompositeBindingConfigProvider : IBindingConfigProvider
```
duncanp-sonar, line 32: Miscellaneous - class does not need to be `public`

src/ConnectedMode/Binding/CompositeBindingConfigProvider.cs
```diff
@@ -20,31 +20,39 @@
...
+    [Export(typeof(IBindingConfigProvider))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    internal class CompositeBindingConfigProvider : IBindingConfigProvider
     {
         private readonly HashSet<IBindingConfigProvider> providers;
 
-        public CompositeBindingConfigProvider(params IBindingConfigProvider[] providers)
-        {
-            // params args can't be null - will be an empty array
-            if (providers.Length == 0)
```
duncanp-sonar, line 39: The providers are no longer set in the public constructor, so I change the "throw" to an "assert".
There is a new unit test that checks that the public constructor creates the expected provider types.

src/ConnectedMode/Binding/IBindingProcessFactory.cs
```diff
@@ -40,57 +39,31 @@ internal interface IBindingProcessFactory
...
-                bindingConfigProvider,
-                configurationPersister,
-                solutionBindingOp,
-                logger);
-
             return new BindingProcessImpl(bindingArgs,
                 exclusionSettingsStorage,
                 sonarQubeService,
-                qpUpdater,
+                qualityProfileDownloader,
```
duncanp-sonar, line 64: The factory is now much simpler; it just imports the `IQualityProfileDownloader` and uses it. It doesn't need to create any additional objects.

src/ConnectedMode/Binding/NonRoslynBindingConfigProvider.cs
```diff
@@ -31,13 +31,28 @@
 
 namespace SonarLint.VisualStudio.ConnectedMode.Binding
 {
-    public class NonRoslynBindingConfigProvider : IBindingConfigProvider
+    internal class NonRoslynBindingConfigProvider : IBindingConfigProvider
```
duncanp-sonar, line 34: Miscellaneous - no need to be `public`.

src/ConnectedMode/Binding/NonRoslynBindingConfigProvider.cs
```diff
@@ -31,13 +31,28 @@
...
     {
+        // List of languages that use this type of configuration file (all non-Roslyn languages)
+        internal static readonly IEnumerable<Language> SupportedLanguages = new []
```
duncanp-sonar, line 37: The list of languages supported by this class is an implementation detail of the class i.e. it shouldn't be set from outside. So, I added an `internal` constructor that allows the set of languages to be set when testing, and a `public` constructor that uses the default set of supported languages.

src/ConnectedMode/Binding/SolutionBindingOperation.cs
```diff
@@ -27,8 +27,10 @@
 
 namespace SonarLint.VisualStudio.ConnectedMode.Binding
 {
+    // TODO - rename. The current name is misleading. IBindingConfigWriter?
```
duncanp-sonar, line 30: This class now does very little (although if you look at the history it used to do an awful lot more e.g. [here](https://github.com/SonarSource/sonarlint-visualstudio/blob/19c6a0fc4e307d9226c5f8bca9b4ebcb6ff5f045/src/Integration/Binding/SolutionBindingOperation.cs)).

The name isn't helpful, but changing it in this PR would make the PR harder to review.

src/ConnectedMode/QualityProfiles/QualityProfileDownloader.cs
```diff
@@ -52,18 +54,18 @@ internal class QualityProfileDownloader : IQualityProfileDownloader
...
             IConfigurationPersister configurationPersister,
-            ISolutionBindingOperation solutionBindingOperation,
             ILogger logger) :
             this(
                 sonarQubeService,
                 bindingConfigProvider,
                 configurationPersister,
-                solutionBindingOperation,
                 logger,
+                new SolutionBindingOperation(),
```
duncanp-sonar, line 68: The `SolutionBindingOperation` is now very simple and doesn't have any dependencies itself, so the fact that it is being used here is an implementation detail that can be hidden
-> removed from the `public` constructor and added to the `internal` constructor for testing.

src/ConnectedMode/Binding/NonRoslynBindingConfigProvider.cs
georgii-borovinskikh-sonarsource: I understand that it wasn't there before, but why is c# not in the list?

PR #4678, opened 2023-08-08
"&lt;" and "&gt;" is replaced by "<" and ">"

### LINKED ISSUE #4675: "&lt;" and "&gt;" is replaced by "<" and ">"

ugras-ergun-sonarsource:
fixes #4675


ugras-ergun-sonarsource:
This will fix the "<>" issue but the build will fail. I will exclude remaining rules in a separate PR.

src/Education/XamlGenerator/RuleHelpXamlTranslator.cs
```diff
@@ -24,6 +24,7 @@
 using System.Diagnostics;
 using System.IO;
 using System.Text;
+using System.Web;
```
ugras-ergun-sonarsource, line 27: I've added this to be able to use HttpUtility.HtmlEncode

bigfluffycookie (approved):
LGTM

PR #4686, opened 2023-08-08
Introduce ICancellableJobRunner interface

duncanp-sonar:
[CAYC] add an abstraction for testing
Minimal product changes.

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/StatusBarReanalysisProgressHandlerTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/StatusBarReanalysisProgressHandlerTests.cs
@@ -23,6 +23,7 @@ using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarLint.VisualStudio.TestInfrastructure;
 using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.Integration.Vsix.CancellableJob;
 
 namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
 {
@@ -46,19 +47,19 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
             var progressHandler = new StatusBarReanalysisProgressHandler(dummyStatusBar, logger);
 
             // 1. Initial request
-            ReportProgress(progressHandler, CancellableJobRunner.RunnerState.Running, 0, 2);
+            ReportProgress(progressHandler, RunnerState.Running, 0, 2);
             dummyStatusBar.CheckLastCallWasSetupCall(0, 2);
 
             // 2a. Progress updates
-            ReportProgress(progressHandler, CancellableJobRunner.RunnerState.Running, 1, 2);
+            ReportProgress(progressHandler, RunnerState.Running, 1, 2);
             dummyStatusBar.CheckLastCallWasInProgressCall(1, 2);
 
             // 2b. Progress updates
-            ReportProgress(progressHandler, CancellableJobRunner.RunnerState.Running, 2, 2);
+            ReportProgress(progressHandler, RunnerState.Running, 2, 2);
             dummyStatusBar.CheckLastCallWasInProgressCall(2, 2);
 
             // 3. Finished -> clean up and reset the statusbar
-            ReportProgress(progressHandler, CancellableJobRunner.RunnerState.Finished, 111, 222);
+            ReportProgress(progressHandler, RunnerState.Finished, 111, 222);
 
             dummyStatusBar.CheckLastCallWasCleanup();
             dummyStatusBar.ProgressCallCount.Should().Be(4);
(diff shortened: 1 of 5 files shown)


src/Integration.Vsix.UnitTests/Analysis/StatusBarReanalysisProgressHandlerTests.cs
```diff
@@ -23,6 +23,7 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarLint.VisualStudio.TestInfrastructure;
 using SonarLint.VisualStudio.Integration.UnitTests;
+using SonarLint.VisualStudio.Integration.Vsix.CancellableJob;
```
duncanp-sonar, line 26: No code changes, just fixing up namespaces.

src/Integration.Vsix.UnitTests/SonarLintTagger/CancellableJobRunnerTests.cs
```diff
@@ -23,14 +23,23 @@
...
+using SonarLint.VisualStudio.Integration.Vsix.CancellableJob;
 using SonarLint.VisualStudio.TestInfrastructure;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests.SonarLintTagger
 {
     [TestClass]
     public class CancellableJobRunnerTests
     {
+        [TestMethod]
+        public void StaticInstance_ReturnsSameProperty()
```
duncanp-sonar, line 35: New test for singleton

src/Integration.Vsix/SonarLintTagger/CancellableJobRunner.cs
```diff
@@ -26,161 +26,180 @@
...
 
-namespace SonarLint.VisualStudio.Integration.Vsix
+namespace SonarLint.VisualStudio.Integration.Vsix.CancellableJob
 {
-    /// <summary>
-    /// Runs a series of operations on a separate background thread.
-    /// Can be cancelled.
-    /// </summary>
-    internal sealed class CancellableJobRunner
+    public interface ICancellableJobRunner
```
duncanp-sonar, line 31: Introduce an interface so classes don't need to reference the implementing type directly.

Previously the associated types (e.g. the status enum) where nested inside the class. Since they are no longer inside the class, I added a more specific namespace (line 29 above) so the types don't pollute the `SonarLint.VisualStudio.Integration` namespace.

src/Integration.Vsix/SonarLintTagger/CancellableJobRunner.cs
```diff
@@ -26,161 +26,180 @@
...
-            {
-                CurrentState = currentState;
-                CompletedOperations = completed;
-                TotalOperations = totalOperations;
-            }
+        /// <summary>
+        /// Runs a series of operations on a separate background thread.
+        /// Can be cancelled.
+        /// </summary>
+        ICancellableJob Start(string jobDescription, IEnumerable<Action> operations, IProgress<JobRunnerProgress> progress, ILogger logger);
```
duncanp-sonar, line 37: The GitHub diff is confusing and not particularly helpful.
Only the construction code has changed; everything else is the same.
I'll call out the changes in comments.

src/Integration.Vsix/SonarLintTagger/CancellableJobRunner.cs
```diff
@@ -26,161 +26,180 @@
...
-        }
+    public interface ICancellableJob
+    {
+        void Cancel();
+    }
 
-        public enum RunnerState
+    // Not overriding Equals etc - this type is only used for progress reporting.
+    // It's not intended to be stored or compared
+    public struct JobRunnerProgress
```
duncanp-sonar, line 47: Unchanged, just moved so it is no longer a nested class.

src/Integration.Vsix/SonarLintTagger/CancellableJobRunner.cs
```diff
@@ -26,161 +26,180 @@
...
-            {
-                // NB this will throw if the cancellation source has
-                // been disposed but not yet nulled out
-                return cancellationSource?.Token.WaitHandle;
-            }
-        }
-        internal /* for testing */ RunnerState State { get; private set; }
+    internal sealed class CancellableJobRunner : ICancellableJobRunner
+    {
+        public static ICancellableJobRunner Instance { get; } = new CancellableJobRunner();
```
duncanp-sonar, line 68: Previously there was a static method that initialized and returned a new instance of the class. It's replaced by this singleton.

src/Integration.Vsix/SonarLintTagger/CancellableJobRunner.cs
```diff
@@ -26,161 +26,180 @@
...
 
-        public static CancellableJobRunner Start(string jobDescription, IEnumerable<Action> operations, IProgress<JobRunnerProgress> progress, ILogger logger)
+        public ICancellableJob Start(string jobDescription, IEnumerable<Action> operations, IProgress<JobRunnerProgress> progress, ILogger logger)
```
duncanp-sonar, line 70: Method is now non-static, and it returns the new `ICancellableJob` - the caller doesn't need to know anything about the returned type except that it can call `Cancel` on it.

src/Integration.Vsix/SonarLintTagger/CancellableJobRunner.cs
```diff
@@ -26,161 +26,180 @@
...
 
             return runner;
         }
 
-        private CancellableJobRunner(string jobDescription, IEnumerable<Action> operations, IProgress<JobRunnerProgress> progress, ILogger logger)
+        /// <summary>
+        /// Runs a series of operations on a separate background thread.
+        /// Can be cancelled.
+        /// </summary>
+        internal sealed class CancellableJob : ICancellableJob
```
duncanp-sonar, line 84: No code changes below this point...

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -131,7 +131,7 @@ private void OnAnalysisRequested(object sender, AnalysisRequestEventArgs args)
                 reanalysisProgressHandler = new StatusBarReanalysisProgressHandler(vsStatusBar, logger);
 
                 var message = string.Format(CultureInfo.CurrentCulture, Strings.JobRunner_JobDescription_ReaanalyzeDocs, operations.Length);
-                reanalysisJob = CancellableJobRunner.Start(message, operations,
+                reanalysisJob = CancellableJobRunner.Instance.Start(message, operations,
```
duncanp-sonar, line 134: This is the only product code that uses the cancellable job runner. I didn't change this class so the runner could be injected as part of this PR, but now that the new abstractions exist it would be possible to do so.

PR #4692, opened 2023-08-09
Cancel previous QP update jobs when triggering a new one

### LINKED ISSUE #4684: Add CancellableJobRunner to QP updating
### Description

Each server-sent notification of a QP update will trigger a new download of the Quality Profiles. We might receive a series of notifications in a short space if time.

If that happens, we don't want multiple QP updates to be running concurrently - there should only be one active update at a time, and the previous update should be cancelled.

TODO
 - [ ] decide whether the QP update on binding should be cancellable by a SSE update or not (?could we get server-sent events if the original binding hasn't finished?)
 - [ ] use the `CancellableJobRunner` to cancel previous jobs.

duncanp-sonar:
Fixes #4684

### DIFF
--- a/src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
+++ b/src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
@@ -62,11 +62,10 @@ namespace SonarLint.VisualStudio.ConnectedMode.QualityProfiles
             if (config.Mode != SonarLintMode.Connected)
             {
                 logger.LogVerbose($"[QualityProfiles] Skipping Quality Profile update. Solution is not bound. Mode: {config.Mode}");
+                return;
             }
-            else
-            {
-                await runner.RunAsync(token => qualityProfileDownloader.UpdateAsync(config.Project, null, CancellationToken.None));
-            }
+            
+            await runner.RunAsync(token => qualityProfileDownloader.UpdateAsync(config.Project, null, CancellationToken.None));
         }
     }
 }
--- a/src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
+++ b/src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using System.ComponentModel.Composition;
 using System.Threading;
 using SonarLint.VisualStudio.ConnectedMode.Helpers;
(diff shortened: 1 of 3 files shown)


src/ConnectedMode/Suppressions/TimedUpdateHandler.cs
```diff
@@ -44,11 +44,11 @@ internal sealed class TimedUpdateHandler : IDisposable
         [ImportingConstructor]
         public TimedUpdateHandler(ISuppressionIssueStoreUpdater suppressionIssueStoreUpdater,
             IServerHotspotStoreUpdater serverHotspotStoreUpdater,
-            IQualityProfileUpdater qualityProfileUpdater)
```
duncanp-sonar, line 47: I accidentally committed some debug-only changes, then undid them. However, VS has auto-formatted the file to remove trailing whitespace.

src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
```diff
@@ -58,8 +64,8 @@ public async Task UpdateAsync()
                 logger.LogVerbose($"[QualityProfiles] Skipping Quality Profile update. Solution is not bound. Mode: {config.Mode}");
                 return;
             }
-
-            await qualityProfileDownloader.UpdateAsync(config.Project, null, CancellationToken.None);
+            
+            await runner.RunAsync(token => qualityProfileDownloader.UpdateAsync(config.Project, null, CancellationToken.None));
```
duncanp-sonar, line 68: Simple change to call the downloader via the runner rather than directly.

src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
```diff
@@ -40,13 +41,18 @@ internal class QualityProfileUpdater : IQualityProfileUpdater
...
         [ImportingConstructor]
-        public QualityProfileUpdater(IConfigurationProvider configProvider, IQualityProfileDownloader qualityProfileDownloader, ILogger logger)
+        public QualityProfileUpdater(IConfigurationProvider configProvider,
+            IQualityProfileDownloader qualityProfileDownloader,
+            ICancellableActionRunner runner,
+            ILogger logger)
         {
             this.configProvider = configProvider;
             this.qualityProfileDownloader = qualityProfileDownloader;
+            this.runner = runner;
```
georgii-borovinskikh-sonarsource, line 55: Runner is not disposed. Not a big deal, but I think it's better to add that

georgii-borovinskikh-sonarsource (approved):
One comment about Dispose

PR #4694, opened 2023-08-09
Add OutOfDateQualityProfileFinder

georgii-borovinskikh-sonarsource:
Implements #4651

### DIFF
--- a/src/ConnectedMode/QualityProfiles/OutOfDateQualityProfileFinder.cs
+++ b/src/ConnectedMode/QualityProfiles/OutOfDateQualityProfileFinder.cs
@@ -21,6 +21,7 @@
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using SonarLint.VisualStudio.Core;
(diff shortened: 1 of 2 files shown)


src/ConnectedMode/QualityProfiles/OutOfDateQualityProfileFinder.cs
```diff
@@ -0,0 +1,93 @@
...
+                
+                // if we know the language, it's in the dictionary
+                Debug.Assert(sonarQubeProject.Profiles.ContainsKey(language));
+
+                var localQualityProfile = sonarQubeProject.Profiles[language];
+                
+                if (!serverQualityProfile.Key.Equals(localQualityProfile.ProfileKey) 
+                    || serverQualityProfile.TimeStamp > localQualityProfile.ProfileTimestamp)
+                {
+                    yield return (language, serverQualityProfile);
```
duncanp-sonar, line 88: What's the reason for using `yield` here? We've already done the web call which is the expensive part of the operation so we're not really saving anything and are just adding unnecessary overhead at runtime.

duncanp-sonar (commented):
Looks good. Just the question about whether the `yield` is adding any value.

src/ConnectedMode/QualityProfiles/OutOfDateQualityProfileFinder.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// <summary>
+        /// Gives the list of outdated quality profiles based on the existing ones from <see cref="BoundSonarQubeProject.Profiles"/>
+        /// </summary>
+        Task<IEnumerable<(Language language, SonarQubeQualityProfile qualityProfile)>> GetAsync(
+            BoundSonarQubeProject sonarQubeProject,
+            CancellationToken cancellationToken);
+    }
+
+    [Export(typeof(IOutOfDateQualityProfileFinder))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Since this particular class is stateless, It could be non shared. But at the same time, it's an implementation detail. I don't know.

src/ConnectedMode/QualityProfiles/OutOfDateQualityProfileFinder.cs
```diff
@@ -0,0 +1,89 @@
...
+            var sonarQubeQualityProfiles =
+                await sonarQubeService.GetAllQualityProfilesAsync(sonarQubeProject.ProjectKey,
+                    sonarQubeProject.Organization.Key,
+                    cancellationToken);
+
+            return sonarQubeQualityProfiles
+                .Select(serverQualityProfile => 
+                    (language: Language.GetLanguageFromLanguageKey(serverQualityProfile.Language),
+                    qualityProfile: serverQualityProfile))
+                .Where(languageAndQp =>
```
duncanp-sonar, line 70: Looks good.
Would it be even more readable with the contents of the `Where` in a helper function?

e.g.
```
.Where(languageAndQp => IsLocalQPOutOfDate(languageAndQp, sonarQubeProject)
```

duncanp-sonar (approved):

PR #4696, opened 2023-08-09
Trigger re-analysis of open docs when QPs are updated

### LINKED ISSUE #4693: Trigger re-analysis of open documents after QP update
### Description

Open documents should be reanalyzed when a QP has been updated

duncanp-sonar:
Fixes #4693

### DIFF
--- a/src/Core/Analysis/INotifyQualityProfilesChanged.cs
+++ b/src/Core/Analysis/INotifyQualityProfilesChanged.cs
@@ -1,0 +1,31 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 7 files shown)


src/Core/Analysis/INotifyQualityProfilesChanged.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+
+namespace SonarLint.VisualStudio.Core.Analysis
+{
+    public interface INotifyQualityProfilesChanged
+    {
+        /// <summary>
+        /// Notification that one or more Quality Profiles have changed
+        /// </summary>
+        event EventHandler QualityProfilesChanged;
```
duncanp-sonar, line 29: The `IAnalysisConfigMonitor` implementation will listen for this event.

The implementation of `AnalysisConfigMonitor` is in currently in `Integration.VSIX`. However, all of its dependencies are in `Core` so the implementation can be moved into `Core`. I was thinking about doing that first, but I'll do it separately after this PR.

src/Integration.Vsix/Analysis/AnalysisStrings.resx
```diff
@@ -187,4 +187,7 @@
   <data name="ConfigMonitor_BindingChanged" xml:space="preserve">
     <value>Binding has changed. Open documents will be re-analysed.</value>
   </data>
+  <data name="ConfigMonitor_QualityProfilesChanged" xml:space="preserve">
+    <value>Quality Profiles have changed. Open documents will be re-analyzed.</value>
```
duncanp-sonar, line 191: Equivalent message to the existing ones for e.g. user settings changed.

georgii-borovinskikh-sonarsource (approved):

Issue #4700, opened 2023-08-10
Quality profile downloader should only update QP for outdated languages


georgii-borovinskikh-sonarsource:
Use IOutOfDateQualityProfileFinder

PR #4704, opened 2023-08-10
Make QualityProfileDownloader download only outdated Quality Profiles

georgii-borovinskikh-sonarsource:
Implements #4700

### DIFF
--- a/src/ConnectedMode.UnitTests/QualityProfiles/QualityProfileDownloaderTests.cs
+++ b/src/ConnectedMode.UnitTests/QualityProfiles/QualityProfileDownloaderTests.cs
@@ -71,9 +71,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.QualityProfiles
 
             SetupAvailableLanguages(out var outOfDateQualityProfileFinderMock,
                 boundProject,
-                languagesToBind
-                    .Select(x => (x, CreateQualityProfile()))
-                    .ToArray());
+                languagesToBind);
 
             var bindingConfigProviderMock = new Mock<IBindingConfigProvider>();
             bindingConfigProviderMock.Setup(x =>
@@ -131,10 +129,9 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.QualityProfiles
 
             // Configure available languages on the server
             SetupAvailableLanguages(out var outOfDateQualityProfileFinderMock,
-                boundProject,
-                new[] { Language.CSharp, Language.VBNET }
-                    .Select(x => (x, CreateQualityProfile()))
-                    .ToArray());
+                boundProject, 
+                Language.CSharp,
+                Language.VBNET);
 
             var configProvider = new Mock<IBindingConfigProvider>();
             var cppConfig = SetupConfigProvider(configProvider, Language.Cpp);
@@ -273,6 +270,16 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.QualityProfiles
         private static SonarQubeQualityProfile CreateQualityProfile(string key = "key", DateTime timestamp = default)
         {
             return new SonarQubeQualityProfile(key, default, default, default, timestamp);
+        }
+
+        private static void SetupAvailableLanguages(
+            out Mock<IOutOfDateQualityProfileFinder> outOfDateQualityProfileFinderMock,
+            BoundSonarQubeProject boundProject,
+            params Language[] languages)
+        {
+            SetupAvailableLanguages(out outOfDateQualityProfileFinderMock, 
+                boundProject,
+                languages.Select(x => (x, CreateQualityProfile())).ToArray());
         }
 
         private static void SetupAvailableLanguages(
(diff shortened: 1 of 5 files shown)


src/ConnectedMode/QualityProfiles/QualityProfileDownloader.cs
```diff
@@ -94,19 +93,19 @@ public async Task<bool> UpdateAsync(BoundSonarQubeProject boundProject, IProgres
...
-            foreach (var language in languagesToBind)
+            foreach (var (language, qualityProfileInfo) in qpToUpdate)
             {
                 currentLanguage++;
 
                 var progressMessage = string.Format(BindingStrings.DownloadingQualityProfileProgressMessage, language.Name);
-                progress?.Report(new FixedStepsProgress(progressMessage, currentLanguage, languageCount));
-
-                var qualityProfileInfo = await TryDownloadQualityProfileAsync(boundProject, language, cancellationToken);
+                progress?.Report(new FixedStepsProgress(progressMessage, currentLanguage, totalLanguages));
```
duncanp-sonar, line 108: The last parameter in `progress.Report` is the total number of steps that will be performed. We're only processing the out of date QPs now - shouldn't the total be `qpToUpdate.Count()`?

src/ConnectedMode/QualityProfiles/QualityProfileDownloader.cs
```diff
@@ -94,19 +93,19 @@ public async Task<bool> UpdateAsync(BoundSonarQubeProject boundProject, IProgres
 
             EnsureProfilesExistForAllSupportedLanguages(boundProject);
 
+            var qpToUpdate = await outOfDateQualityProfileFinder.GetAsync(boundProject, cancellationToken);
```
duncanp-sonar, line 96: Minor: the variable name `qpToUpdate` suggests a single QP, not a list.

Maybe something like `profilesToUpdate` / `outOfDateProfiles`  instead?

src/ConnectedMode/QualityProfiles/QualityProfileDownloader.cs
```diff
@@ -94,19 +93,19 @@ public async Task<bool> UpdateAsync(BoundSonarQubeProject boundProject, IProgres
...
             {
                 currentLanguage++;
 
                 var progressMessage = string.Format(BindingStrings.DownloadingQualityProfileProgressMessage, language.Name);
-                progress?.Report(new FixedStepsProgress(progressMessage, currentLanguage, languageCount));
-
-                var qualityProfileInfo = await TryDownloadQualityProfileAsync(boundProject, language, cancellationToken);
+                progress?.Report(new FixedStepsProgress(progressMessage, currentLanguage, totalLanguages));
 
                 if (qualityProfileInfo == null)
```
duncanp-sonar, line 110: Can `qualityProfileInfo` ever be `null` now? Would that be a bug in the `OutOfDateQualityProfileFinder`?

If so, then this check and the associated test could be removed, and perhaps replaced with `Debug.Assert` instead (or by an assertion earlier that checks that none of the entries in `qpToUpdate` are `null`).

src/ConnectedMode.UnitTests/QualityProfiles/QualityProfileDownloaderTests.cs
```diff
@@ -40,51 +39,61 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.QualityProfiles
...
                 Language.Cpp,
                 Language.CSharp,
                 Language.Secrets,
                 Language.VBNET
             };
 
+            SetupAvailableLanguages(out var outOfDateQualityProfileFinderMock,
+                boundProject,
+                languagesToBind
+                    .Select(x => (x, CreateQualityProfile()))
```
duncanp-sonar, line 75: Multiple tests have this `.Select(x => (x, CreateQualityProfile()` snippet repeated to create the tuples.

It would simplify the test setup to have a version of `SetupAvailableLanguages` that just accepted a list of languages and created the tuples internally.

src/ConnectedMode/QualityProfiles/QualityProfileDownloader.cs
georgii-borovinskikh-sonarsource: Yes, but I got warnings about double enumeration, so I decided to just take the worst case scenario in terms of number of steps. Sure, sometimes it will stop before that, but nothing would break.

duncanp-sonar: > I got warnings about double enumeration

What do you mean? Couldn't you just store the value once in line 101?


BTW this is the sort of thing that you should call out explicitly in code review comments, and not wait to see if the reviewer notices it. It's not obvious, easy to miss, and means the UX won't be what the user expects. You've made a clear decision to handle the problem in a particular way, so call it out in a review comment.

src/ConnectedMode/QualityProfiles/QualityProfileDownloader.cs
georgii-borovinskikh-sonarsource: You're right, it should never be null. Debug.Assert is not needed. Also, there were no tests that covered this path.

duncanp-sonar (approved):
Looks good!

Issue #4707, opened 2023-08-11
Eliminate ISolutionBindingOperation


duncanp-sonar:
### Description

`ISolutionBindingOperation` used to be much more complicated. Now it only contains a single method and adds very little value. The class ensures the directories for language-specific settings exist, then tells the `IBindingConfig` to save itself.

It unnecessarily complicates how the language-specific rules configurations are saved, and makes fixing #4699 more complicated.

Proposal: make the individual `IBindingConfig` implementations responsible for ensure any necessary sub-directories exist, and remove this class.

Prep for #4699

PR #4709, opened 2023-08-11
Move responsibility for directory creation to IBindingConfig

duncanp-sonar:
... out of ISolutionBindingOperation

Prep for #4707

### DIFF
--- a/src/ConnectedMode.UnitTests/Binding/CSharpVbBindingConfigTests.cs
+++ b/src/ConnectedMode.UnitTests/Binding/CSharpVbBindingConfigTests.cs
@@ -20,7 +20,7 @@
 
 using System;
 using System.IO;
-using System.Linq;
+using System.IO.Abstractions.TestingHelpers;
 using SonarLint.VisualStudio.Core.CSharpVB;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Binding.UnitTests
@@ -28,8 +28,6 @@ namespace SonarLint.VisualStudio.ConnectedMode.Binding.UnitTests
     [TestClass]
     public class CSharpVbBindingConfigTests
     {
-        public TestContext TestContext { get; set; }
-
         [TestMethod]
         public void Ctor_InvalidArgs()
         {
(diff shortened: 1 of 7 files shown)


src/ConnectedMode/Binding/SolutionBindingOperation.cs
```diff
@@ -27,7 +27,7 @@
 
 namespace SonarLint.VisualStudio.ConnectedMode.Binding
 {
-    // TODO - rename. The current name is misleading. IBindingConfigWriter?
+    // TODO - remove. See #4707.
```
duncanp-sonar, line 30: This file will be completely removed in the next PR.

src/ConnectedMode/Binding/SolutionBindingOperation.cs
```diff
@@ -57,12 +57,6 @@ public void SaveRuleConfiguration(IEnumerable<IBindingConfig> bindingConfigs, Ca
                     return;
                 }
 
-                foreach (var solutionItem in config.SolutionLevelFilePaths)
-                {
-                    var ruleSetDirectoryPath = Path.GetDirectoryName(solutionItem);
-                    fileSystem.Directory.CreateDirectory(ruleSetDirectoryPath); // will no-op if exists
```
duncanp-sonar, line 63: Previously, the only real value left in the `SolutionBindingOperation` class was that it ensured a directory exists before writing a rule file. This PR moves that responsibility to the individual binding config implementations, so this class can then be deleted.

src/ConnectedMode/Binding/IBindingConfigProvider.cs
```diff
@@ -47,14 +47,10 @@ public interface IBindingConfigProvider
     /// For C++ it will be in a json file in a Sonar-specific format</remarks>
     public interface IBindingConfig
     {
-        /// <summary>
-        /// Returns a list of all of the solution-level files that should exist if the solution is correctly bound. The files might not exist on disk. List cannot be null or empty.
-        /// </summary>
-        IEnumerable<string> SolutionLevelFilePaths { get; }
```
duncanp-sonar, line 53: The property existed so the `SolutionBindingOperation` could work out which directories to create -> no longer required.

src/ConnectedMode/Binding/CSharpVBBindingConfig.cs
```diff
@@ -49,10 +47,18 @@ internal CSharpVBBindingConfig(FilePathAndContent<string> globalConfig, FilePath
...
         {
+            EnsureParentDirectoryExists(GlobalConfig.Path);
             fileSystem.File.WriteAllText(GlobalConfig.Path, GlobalConfig.Content);
 
             var serializedAdditionalFile = Serializer.ToString(AdditionalFile.Content);
+            EnsureParentDirectoryExists(AdditionalFile.Path);
             fileSystem.File.WriteAllText(AdditionalFile.Path, serializedAdditionalFile);
         }
+
+        private void EnsureParentDirectoryExists(string filePath)
```
duncanp-sonar, line 58: This method is duplicated in the two binding config implementations. However, it's so small I didn't think it was worth factoring out elsewhere.

src/ConnectedMode.UnitTests/Binding/CSharpVbBindingConfigTests.cs
```diff
@@ -47,42 +45,38 @@ public void Ctor_InvalidArgs()
         }
 
         [TestMethod]
-        public void GetSolutionLevelFilePaths_ReturnFilePaths()
```
duncanp-sonar, line 50: The property not longer exists on the interface -> this test is no longer required.

src/ConnectedMode.UnitTests/Binding/CSharpVbBindingConfigTests.cs
```diff
@@ -47,42 +45,38 @@ public void Ctor_InvalidArgs()
...
-            testSubject.SolutionLevelFilePaths.Count().Should().Be(2);
-            testSubject.SolutionLevelFilePaths.First().Should().Be(globalConfig.Path);
-            testSubject.SolutionLevelFilePaths.Last().Should().Be(additionalFile.Path);
-        }
-
-        [TestMethod]
-        public void Save_FilesSaved()
-        {
-            // We can't mock the RuleSet class so we're testing Save by actually
-            // writing to disk.
```
duncanp-sonar, line 65: I changed this test so it doesn't write to the real file system.

src/ConnectedMode.UnitTests/Binding/CSharpVbBindingConfigTests.cs
```diff
@@ -20,16 +20,14 @@
...
-using System.Linq;
+using System.IO.Abstractions.TestingHelpers;
 using SonarLint.VisualStudio.Core.CSharpVB;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Binding.UnitTests
 {
     [TestClass]
     public class CSharpVbBindingConfigTests
     {
-        public TestContext TestContext { get; set; }
```
duncanp-sonar, line 31: No longer required. It was only needed before because a test accessed the real file system, and used `TestContext` to work out where to write a file.

src/ConnectedMode.UnitTests/Binding/NonRoslynBindingConfigFileTests.cs
```diff
@@ -55,15 +54,6 @@ public void Ctor_ValidArgs()
             testSubject.FilePath.Should().BeEquivalentTo("c:\\test");
         }
 
-        [TestMethod]
-        public void GetSolutionLevelFilePaths_ReturnPathToSettingsFile()
```
duncanp-sonar, line 59: Same as above - this test is no longer required.

src/ConnectedMode.UnitTests/Binding/NonRoslynBindingConfigFileTests.cs
```diff
@@ -85,21 +75,28 @@ public void Save_SettingsAreSerializedAndSaved()
                 }
             };
 
-            string actualPath = null;
-            string actualText = null;
+            string filePath = "c:\\full\\path\\file.txt";
 
-            var fileSystemMock = new Mock<IFileSystem>();
```
duncanp-sonar, line 91: Using `MockFileSystem` instead simplifies the test.

duncanp-sonar:
The same set of directories and files are created:

image

Issue #4705, opened 2023-08-11
Make QP downloader return whether any QPs were updated


georgii-borovinskikh-sonarsource:
Could then avoid raising an event if nothing changed

PR #4710, opened 2023-08-11
Make QualityProfileDownloader return if there were changes

georgii-borovinskikh-sonarsource:
Implements #4705

* QualityProfileDownloader returns false when no QPs updated, true if at least one

* QualityProfileDownloader throws InvalidOperationException if binding config not found

* QualityProfileUpdater only raises the event if QualityProfileDownloader returns true

* QualityProfileUpdater handles InvalidOperationException from QualityProfileDownloader

* BindingProcessImpl handles InvalidOperationException from QualityProfileDownloader

### DIFF
--- a/src/ConnectedMode.UnitTests/QualityProfiles/QualityProfileDownloaderTests.cs
+++ b/src/ConnectedMode.UnitTests/QualityProfiles/QualityProfileDownloaderTests.cs
@@ -28,7 +28,6 @@ using SonarLint.VisualStudio.ConnectedMode.QualityProfiles;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Integration;
-using SonarLint.VisualStudio.Integration.Resources;
 using SonarLint.VisualStudio.Progress.Controller;
 using SonarLint.VisualStudio.TestInfrastructure;
 using SonarQube.Client.Models;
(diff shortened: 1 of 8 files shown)


src/ConnectedMode/QualityProfiles/QualityProfileUpdater.cs
```diff
@@ -72,13 +72,18 @@ public async Task UpdateAsync()
...
+                    {
+                        QualityProfilesChanged?.Invoke(this, EventArgs.Empty);
+                    }
+                });
             }
-            catch (OperationCanceledException)
+            catch (Exception e) when (e is OperationCanceledException || e is InvalidOperationException)
             {
                 // no-op - job was cancelled
+                logger.LogVerbose(e.ToString());
```
duncanp-sonar, line 86: Prefix the string with the feature name to give a bit more context: `[QualityProfiles] `

src/ConnectedMode.UnitTests/QualityProfiles/QualityProfileDownloaderTests.cs
```diff
@@ -53,6 +52,31 @@ public void MefCtor_CheckIsSingleton()
...
+
+            var testSubject = CreateTestSubject(outOfDateQualityProfileFinderMock.Object,
+                logger: logger,
+                solutionBindingOperation: solutionBindingOperationMock.Object,
+                languagesToBind: Language.KnownLanguages.ToArray());
+
+            var result = await testSubject.UpdateAsync(boundSonarQubeProject, null, CancellationToken.None);
+
+            result.Should().BeFalse();
+            solutionBindingOperationMock.Verify(
```
duncanp-sonar, line 74: The check here could be simpler and easier to read e.g. `solutionBindingOperationMock.Invocations.Should().BeEmpty()`.

But I wouldn't bother changing it since I am going to delete the class completely in one of my PRs.

duncanp-sonar (approved):
Looks good!

PR #4713, opened 2023-08-11
Save language-specific rule config files ASAP

### LINKED ISSUE #4699: Cancelling QP update can leave the rule config in an inconsistent state
### Description

The `binding.config` and rule config files can be left in an inconsistent state if the QP update is cancelled.

* for each language, the QP data is fetched from the server, and the `binding.config` file is updated and saved with the updated QP timestamp for that language.
* the language-specific rule config files are not written until all of the languages have been processed

i.e. if the process is cancelled between the two stages then the profile/timestamp information in `binding.config` will be wrong.

It will look as if the QPs are up to date, when in fact the updated rule config files have not been saved -> the rule config files will not be updated next time the QPs are updated.

Historical note: the downloader is a new component, but it was created when refactoring the binding code for unintrusive Connected Mode (v7.0). The logic dates from a time when file writes were queued rather than committed to disc immediately, so the order didn't matter at that point.
File writes were queued because they were written to a source-controlled location. Queuing meant all the file changes were written once at the end of the binding if everything else had succeeded.

duncanp-sonar:
- remove ISolutionBindingOperation

Fixes #4699 and #4707


src/ConnectedMode.UnitTests/QualityProfiles/QualityProfileDownloaderTests.cs
```diff
@@ -61,20 +61,19 @@ public async Task UpdateAsync_NothingToUpdate_ReturnsFalse()
...
                 languagesToBind: Language.KnownLanguages.ToArray());
 
             var result = await testSubject.UpdateAsync(boundSonarQubeProject, null, CancellationToken.None);
 
             result.Should().BeFalse();
-            solutionBindingOperationMock.Verify(
-                x =>
-                    x.SaveRuleConfiguration(It.IsAny<IEnumerable<IBindingConfig>>(), It.IsAny<CancellationToken>()),
-                Times.Never);
+            bindingConfigProvider.Invocations.Should().BeEmpty();
```
duncanp-sonar, line 75: The `ISolutionBindingConfiguration` class doesn't exist anymore, so instead we're checking the config provider is never called as an indicator that the method exited.

georgii-borovinskikh-sonarsource (approved):
LGTM

PR #4741, opened 2023-08-18
Refactor MigrationPrompt to use ISolutionInfoProvider

duncanp-sonar:
Part of #4481

### DIFF
--- a/src/ConnectedMode.UnitTests/Migration/MigrationPromptTests.cs
+++ b/src/ConnectedMode.UnitTests/Migration/MigrationPromptTests.cs
@@ -21,7 +21,6 @@
 using System;
 using System.Linq;
 using Microsoft.VisualStudio.Shell;
-using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Threading;
 using SonarLint.VisualStudio.ConnectedMode.Migration;
 using SonarLint.VisualStudio.ConnectedMode.Migration.Wizard;
@@ -42,11 +41,10 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Migration
         public void MefCtor_CheckIsExported()
         {
             MefTestHelpers.CheckTypeCanBeImported<MigrationPrompt, IMigrationPrompt>(
-                MefTestHelpers.CreateExport<SVsServiceProvider>(),
+                MefTestHelpers.CreateExport<ISolutionInfoProvider>(),
                 MefTestHelpers.CreateExport<INotificationService>(),
                 MefTestHelpers.CreateExport<IMigrationWizardController>(),
-                MefTestHelpers.CreateExport<IBrowserService>(),
-                MefTestHelpers.CreateExport<IThreadHandling>());
+                MefTestHelpers.CreateExport<IBrowserService>());
         }
 
         [TestMethod]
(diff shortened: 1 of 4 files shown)


src/ConnectedMode.UnitTests/Migration/MigrationPromptTests.cs
```diff
@@ -89,31 +87,6 @@ public async Task ShowAsync_VerifyNotificationIsCreatedWithExpectedParameters(bo
             notification.Actions.Last().CommandText.Should().Be(MigrationStrings.MigrationPrompt_LearnMoreButton);
         }
 
-        [TestMethod]
-        public async Task ShowAsync_ShowNotificationIsCalledOnMainThread()
```
duncanp-sonar, line 93: The migration prompt class no longer needs to do any explicit thread handling, so this thread is redundant.

src/ConnectedMode.UnitTests/Migration/MigrationPromptTests.cs
```diff
@@ -42,11 +41,10 @@ public class MigrationPromptTests
         public void MefCtor_CheckIsExported()
         {
             MefTestHelpers.CheckTypeCanBeImported<MigrationPrompt, IMigrationPrompt>(
-                MefTestHelpers.CreateExport<SVsServiceProvider>(),
```
duncanp-sonar, line 45: The class no longer uses `IServiceProvider` directly - it uses it indirectly via `ISolutionInfoProvider`.

As a consequence of this, it no longer needs to manage thread handling explicitly, since the implementation of `ISolutionInfoProvider` handles switching to the UI thread when necessary.

src/ConnectedMode.UnitTests/Migration/MigrationPromptTests.cs
```diff
@@ -125,11 +98,9 @@ public async Task MigrateButtonClicked_StartMigrationWizard()
 
             var migrationWizardController = new Mock<IMigrationWizardController>();
 
-            var serviceProvider = SetUpServiceProviderWithSolution();
```
duncanp-sonar, line 128: Most of the tests don't use the `ISolutionInfoProvider`, so I just removed it (previously, all of the tests needed a service provider that returned a VS service otherwise we'd get a null-ref exception).

src/ConnectedMode/Migration/IMigrationPrompt.cs
```diff
@@ -34,6 +32,10 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
     /// <summary>
     /// In charge of showing users a prompt to migrate to a new connected mode.
     /// </summary>
+    /// <remarks>The caller can assume that the component follows VS threading rules
+    /// i.e. the implementing class is responsible for switching to the UI thread if necessary.
+    /// The caller doesn't need to worry about it.
+    /// </remarks>
```
duncanp-sonar, line 38: I've added this comment to all three of the interfaces I've touched as part of this PR, since the implementations all handle VS threading correctly.

src/ConnectedMode/Migration/IMigrationPrompt.cs
```diff
@@ -78,25 +77,24 @@ public async Task ShowAsync(BoundSonarQubeProject oldBinding, bool hasNewBinding
         {
             this.oldBinding = oldBinding;
 
-            await threadHandling.RunOnUIThreadAsync(() =>
```
duncanp-sonar, line 81: The `ISolutionInfoProvider` and `INotificationService` both handle threading correctly, so this class doesn't need to do anything extra at all.

(As it happens, it will be called on the UI thread, but that's down to the caller. This class no longer cares.)

bigfluffycookie (approved):
Very nice!

PR #4755, opened 2023-08-21
Add "GetSolutionDirectory" methods to ISolutionInfoProvider

duncanp-sonar:
Part of #4481

### DIFF
--- a/src/ConnectedMode.UnitTests/Persistence/LegacySolutionBindingPathProviderTests.cs
+++ b/src/ConnectedMode.UnitTests/Persistence/LegacySolutionBindingPathProviderTests.cs
@@ -49,21 +49,21 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Persistence
 
         [TestMethod]
         [DataRow(null, null)] // null solution -> solution not open -> return null
-        [DataRow(@"c:\test\any.sln", @"c:\test\SonarQube\SolutionBinding.sqconfig")]
-        [DataRow(@"c:\aaa\bbb\x.y.x.any.sln", @"c:\aaa\bbb\SonarQube\SolutionBinding.sqconfig")]
-        public void Get_ReturnsExpectedValue(string solutionPathToReturn, string expectedResult)
+        [DataRow(@"c:\test", @"c:\test\SonarQube\SolutionBinding.sqconfig")]
+        [DataRow(@"c:\aaa\bbb", @"c:\aaa\bbb\SonarQube\SolutionBinding.sqconfig")]
+        public void Get_ReturnsExpectedValue(string solutionDirectoryToReturn, string expectedResult)
         {
-            var solutionInfoProvider = CreateSolutionInfoProvider(solutionPathToReturn);
+            var solutionInfoProvider = CreateSolutionInfoProvider(solutionDirectoryToReturn);
             var testSubject = CreateTestSubject(solutionInfoProvider.Object);
 
             var actual = testSubject.Get();
             actual.Should().Be(expectedResult);
         }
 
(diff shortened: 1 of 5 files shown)


src/Core/ISolutionInfoProvider.cs
```diff
@@ -31,6 +31,18 @@ namespace SonarLint.VisualStudio.Core
     /// </remarks>
     public interface ISolutionInfoProvider
     {
+        /// <summary>
+        /// Returns the full directory path for the current solution file, or null if there is not a current solution
+        /// </summary>
+        Task<string> GetSolutionDirectoryAsync();
```
duncanp-sonar, line 37: There are several classes that need to know the solution directory, rather than the full solution file path.

I considered a number of options:
1. change each of those places to use `ISolutionInfoProvider` and then call `Path.GetDirectory`;
2. add an extension method to the interface that did the same thing
3. add the methods to the interface itself.

I went for option (3) - no duplication in the calling code, and it's immediately obvious that the caller wants the directory.

I would have had to write the same amount of code for (2), and it would have added an unnecessary layer of complexity.

src/Infrastructure.VS.UnitTests/SolutionInfoProviderTests.cs
```diff
@@ -65,7 +65,7 @@ public void Ctor_DoesNotCallServices()
         [TestMethod]
         [DataRow(null)]
         [DataRow("a value")]
-        public async Task GetAsync_ReturnsExpectedValue(string solutionNameToReturn)
+        public async Task GetSlnFilePathAsync_ReturnsExpectedValue(string solutionNameToReturn)
```
duncanp-sonar, line 68: Renamed the existing test methods to make it clear which methods they relate to.

src/Infrastructure.VS.UnitTests/SolutionInfoProviderTests.cs
```diff
@@ -91,7 +91,35 @@ public void Get_ReturnsExpectedValue(string solutionNameToReturn)
         }
 
         [TestMethod]
-        public async Task GetAsync_ServiceCalledOnUIThread()
+        [DataRow(null, null)]
+        [DataRow("c:\\aaa\\bbb\\mysolution.sln", "c:\\aaa\\bbb")]
+        public async Task GetDirectoryAsync_ReturnsExpectedValue(string solutionNameToReturn, string expectedResult)
```
duncanp-sonar, line 96: Copies of the existing tests, modified to call the new directory methods.

src/ConnectedMode.UnitTests/Persistence/LegacySolutionBindingPathProviderTests.cs
```diff
@@ -49,21 +49,21 @@ public void Ctor_DoesNotCallServices()
 
         [TestMethod]
         [DataRow(null, null)] // null solution -> solution not open -> return null
-        [DataRow(@"c:\test\any.sln", @"c:\test\SonarQube\SolutionBinding.sqconfig")]
-        [DataRow(@"c:\aaa\bbb\x.y.x.any.sln", @"c:\aaa\bbb\SonarQube\SolutionBinding.sqconfig")]
-        public void Get_ReturnsExpectedValue(string solutionPathToReturn, string expectedResult)
+        [DataRow(@"c:\test", @"c:\test\SonarQube\SolutionBinding.sqconfig")]
+        [DataRow(@"c:\aaa\bbb", @"c:\aaa\bbb\SonarQube\SolutionBinding.sqconfig")]
+        public void Get_ReturnsExpectedValue(string solutionDirectoryToReturn, string expectedResult)
```
duncanp-sonar, line 54: Fixed up the existing test.
Previously, the product code called "get solution file name" and used `Path.GetDirectoryName` to calculate the directory. Now it just calls the "get directory" method.

PR #4759, opened 2023-08-22
[Infra] Tweak cleanup script

duncanp-sonar:
... so it doesn't matter which directory it is run from

### DIFF
--- a/tools/CleanRebuild.cmd
+++ b/tools/CleanRebuild.cmd
@@ -1,8 +1,8 @@
 @REM this script cleans all of the artifact locations and rebuilds the project
-call cd ..
-(for %%a in (".\build\ProcessJarFiles\tmp" ".\src\Rules\Embedded" ".\src\Integration.Vsix\lib" "%LOCALAPPDATA%\SLVS_Build_DownloadedJars" "%LOCALAPPDATA%\SLVS_CFamily_Build" "%LOCALAPPDATA%\SLVS_TypeScript_Build") do rd /s /q "%%~a")
-call cd ".\build\ProcessJarFiles"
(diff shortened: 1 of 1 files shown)


tools/CleanRebuild.cmd
```diff
@@ -1,8 +1,8 @@
 @REM this script cleans all of the artifact locations and rebuilds the project
-call cd ..
```
duncanp-sonar, line 2: I modified the script so it doesn't need to changed directory, and will work correctly regardless of what the current directory is.

tools/CleanRebuild.cmd
```diff
@@ -1,8 +1,8 @@
...
-(for %%a in (".\build\ProcessJarFiles\tmp" ".\src\Rules\Embedded" ".\src\Integration.Vsix\lib" "%LOCALAPPDATA%\SLVS_Build_DownloadedJars" "%LOCALAPPDATA%\SLVS_CFamily_Build" "%LOCALAPPDATA%\SLVS_TypeScript_Build") do rd /s /q "%%~a")
-call cd ".\build\ProcessJarFiles"
-call msbuild.exe -t:Rebuild
-call cd "..\..\"
-call msbuild.exe "SonarLint.VisualStudio.Integration.sln" -t:Rebuild
-call cd "..\tools"
\ No newline at end of file
+
+@REM %~dp0 gives the directory containing this batch file
```
duncanp-sonar, line 3: The paths used in the script are now all effectively absolute: they are either relative to the location of the script file, or absolute paths like `%APPDATA%`.

georgii-borovinskikh-sonarsource (approved):

PR #4766, opened 2023-08-22
[CAYC] Drop unused interface ISourceControlledFileSystem

duncanp-sonar:


duncanp-sonar:
The product implementation and tests had already been dropped, but the interface and a test helper class had been left behind.

georgii-borovinskikh-sonarsource (approved):

PR #4771, opened 2023-08-23
Remove IHost.UIDispatcher

### LINKED ISSUE #3165: [Threading] Remove use of IHost.UIDispatcher
### Description

The `IHost` has a `UIDispatcher` property that directly exposes a `System.Windows.Threading.Dispatcher`.
This should be replaced with `IThreadHandling`.

It's possible that `IThreadHandling` will need to be extended to provide additional methods e.g. the `ErrorListInfoBarController` specifies a `DispatcherPriority` when launching some tasks. The VS `ThreadHelper` has a `StartOnIdle` method that might be an appropriate alternative.

Update: following the refactorings and code deletion for v7.0, there is only remaining use of `IHost.Dispatcher`, which can easily be replaced with `IThreadHandling`.

duncanp-sonar:
Fixes #3165

### DIFF
--- a/src/Integration.TeamExplorer.UnitTests/SectionControllerTests.cs
+++ b/src/Integration.TeamExplorer.UnitTests/SectionControllerTests.cs
@@ -51,11 +51,10 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.TeamExplorer
         [TestInitialize]
         public void TestInitialize()
         {
-            ThreadHelper.SetCurrentThreadAsUIThread();
             this.serviceProvider = new ConfigurableServiceProvider(assertOnUnexpectedServiceRequest: false);
 
             this.sonarQubeServiceMock = new Mock<ISonarQubeService>();
-            this.host = new ConfigurableHost(this.serviceProvider, Dispatcher.CurrentDispatcher)
+            this.host = new ConfigurableHost(this.serviceProvider)
             {
                 SonarQubeService = this.sonarQubeServiceMock.Object
             };
--- a/src/Integration.UnitTests/State/StateManagerTests.cs
+++ b/src/Integration.UnitTests/State/StateManagerTests.cs
@@ -36,12 +36,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.State
     [TestClass]
     public class StateManagerTests
     {
-        [TestInitialize]
-        public void TestInitialize()
-        {
-            ThreadHelper.SetCurrentThreadAsUIThread();
-        }
-
         #region Tests
 
         [TestMethod]
@@ -506,7 +500,8 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.State
 
         private StateManager CreateTestSubject(ConfigurableHost host, ConfigurableSectionController section = null)
         {
-            var testSubject = new StateManager(host, new TransferableVisualState());
+            var threadHandling = new NoOpThreadHandler();
+            var testSubject = new StateManager(host, new TransferableVisualState(threadHandling), threadHandling);
 
             if (section != null)
             {
--- a/src/Integration/MefServices/IHost.cs
+++ b/src/Integration/MefServices/IHost.cs
@@ -19,7 +19,6 @@
  */
 
 using System;
-using System.Windows.Threading;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration.State;
 using SonarLint.VisualStudio.Integration.TeamExplorer;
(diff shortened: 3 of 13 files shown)


src/Integration.TeamExplorer.UnitTests/SectionControllerTests.cs
```diff
@@ -51,11 +51,10 @@ public class SectionControllerTests
         [TestInitialize]
         public void TestInitialize()
         {
-            ThreadHelper.SetCurrentThreadAsUIThread();
             this.serviceProvider = new ConfigurableServiceProvider(assertOnUnexpectedServiceRequest: false);
 
             this.sonarQubeServiceMock = new Mock<ISonarQubeService>();
-            this.host = new ConfigurableHost(this.serviceProvider, Dispatcher.CurrentDispatcher)
```
duncanp-sonar, line 58: Most of the changes are to fix up the tests since the host no longer needs a dispatcher.
The product code changes are very small.

src/Integration.UnitTests/State/StateManagerTests.cs
```diff
@@ -36,12 +36,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.State
     [TestClass]
     public class StateManagerTests
     {
-        [TestInitialize]
-        public void TestInitialize()
-        {
-            ThreadHelper.SetCurrentThreadAsUIThread();
```
duncanp-sonar, line 42: The VS `ThreadHelper` is no longer used in the modified product code, so I've removed this call from any of the tests that had it.

src/Integration/MefServices/IHost.cs
```diff
@@ -29,11 +28,6 @@ namespace SonarLint.VisualStudio.Integration
 {
     internal interface IHost : IServiceProvider
     {
-        /// <summary>
-        /// The UI thread dispatcher. Not null.
-        /// </summary>
-        Dispatcher UIDispatcher { get; }
-
```
duncanp-sonar, line 36: The host didn't do anything with the dispatcher except expose it as a property.
The only consumer left was the `StateManager`.

(there used to be more consumers of the property before all the simplifications we did in v7, so removing the dispatcher would have involved a lot more work)

src/Integration/State/StateManager.cs
```diff
@@ -109,14 +108,7 @@ public IEnumerable<ConnectionInformation> GetConnectedServers()
...
         public void SetProjects(ConnectionInformation connection, IEnumerable<SonarQubeProject> projects)
         {
-            if (this.Host.UIDispatcher.CheckAccess())
-            {
-                this.SetProjectsUIThread(connection, projects);
-            }
-            else
-            {
-                this.Host.UIDispatcher.BeginInvoke(new Action(() => this.SetProjectsUIThread(connection, projects)));
-            }
```
duncanp-sonar, line 119: This is the only material product change.

There weren't any tests that checked that the Dispatcher was used.
I looked at adding one to check that the thread handler is called but the `StateManager` is old code that wasn't written with testing in mind, and I couldn't see a straightforward way to do it.

Instead, I've added a thread check in line 167 below. Also, I've manually debugged through the code to check what happens when setting up a new binding. This method is called on the UI thread and from a background thread during binding, and from a UI thread when a solution is closed. In all cases the `SetProjectsUIThread` is called on the UI thread.

src/Integration/State/StateManager.cs
```diff
@@ -172,6 +164,7 @@ private void OnStatePropertyChanged(object sender, System.ComponentModel.Propert
 
         private void SetProjectsUIThread(ConnectionInformation connection, IEnumerable<SonarQubeProject> projects)
         {
+            threadHandling.ThrowIfNotOnUIThread();
```
duncanp-sonar, line 167: Additional check that we are on the correct thread.

src/Integration/State/TransferableVisualState.cs
```diff
@@ -29,14 +30,24 @@ namespace SonarLint.VisualStudio.Integration.State
...
+        internal /* for testing */ TransferableVisualState(IThreadHandling threadHandling)
+        {
+            this.threadHandling = threadHandling;
+        }
+
         public ObservableCollection<ServerViewModel> ConnectedServers
         {
             get
             {
-                Debug.Assert(ThreadHelper.CheckAccess(), $"{nameof(ConnectedServers)} should only be accessed from the UI thread");
```
duncanp-sonar, line 39: This class had lots of assertions using `ThreadHelper`. I've replaced them with `IThreadHandling`, but to do that I had to add some constructors so the tests can supply the component to use.

georgii-borovinskikh-sonarsource (approved):

georgii-borovinskikh-sonarsource:
There's a "Major" code smell. False positive?

Issue #4773, opened 2023-08-23
[Cleanup] Remove ILocalService


duncanp-sonar:
### Description

`ILocalService` is a legacy concept that is no longer used in the product; we use MEF composition instead.

TODO
 - [x] convert/remove any remaining `ILocalService` implementations
 - [x] remove local service handling from `IHost`
 - [x] remove `ILocalService`

PR #4776, opened 2023-08-23
[CAYC] Don't download server settings in ConnectionWorkflow

### LINKED ISSUE #4775: [Cleanup] Simplify ConnectionWorkflow to not download server settings
### Description

The old Connected Mode logic downloaded server settings when establishing the connection. It did this so that before starting the binding process it could use one of the server settings to help work out which projects in the solutions were test projects, as test projects are excluded from various operations e.g. binding, and checking for unbound projects.

As part of setting up the project-filtering logic, the `ConnectionWorkflow` fetched `sonar.cs.msbuild.testProjectPattern` setting from the server and passed it to the `ITestProjectRegexSetter`.

The new unintrusive Connected Mode doesn't change any project files
-> it no longer requires the filtering code
-> `ConnectionWorkflow` no longer needs to fetch the setting.

This filtering code has already been deleted in [this commit](https://github.com/SonarSource/sonarlint-visualstudio/commit/588f04b8414db35eff793291140ed550edeafdc2). However, the `ConnectionWorkflow` is still fetching the property from the server.

Note: server settings are still fetched by the binding workflow and some are saved as part of the binding configuration. However, that is entirely separate.

TODO
 - [x] simplify `ConnectionWorkflow` to remove fetching/setting the test project regular expression

#### Background
`ITestProjectRegexSetter` was add in [this commit](https://github.com/SonarSource/sonarlint-visualstudio/commit/aff79ca8#diff-df4be3282534355a4457bba086fb3e51ecd64390bc4f313f0439e75fea3e1c86) as a level of indirection over the top of the `IProjectSystemFilter` (related to fixing bug #787)

Related: see #4773. `ITestProjectRegexSetter` is an `ILocalService`; fixing this ticket will enable #4773.

duncanp-sonar:
Fixes #4775. Prep for #4773.

### DIFF
--- a/src/Integration/Connection/ConnectionWorkflow.cs
+++ b/src/Integration/Connection/ConnectionWorkflow.cs
@@ -49,17 +49,13 @@ namespace SonarLint.VisualStudio.Integration.Connection
         private readonly IHost host;
         private readonly ICommand parentCommand;
         private readonly ICredentialStoreService credentialStore;
-        private readonly ITestProjectRegexSetter testProjectRegexSetter;
 
         public ConnectionWorkflow(IHost host, ICommand parentCommand)
         {
             this.host = host ?? throw new ArgumentNullException(nameof(host));
             this.parentCommand = parentCommand ?? throw new ArgumentNullException(nameof(parentCommand));
 
-            this.credentialStore = this.host.GetMefService<ICredentialStoreService>();
-
-            this.testProjectRegexSetter = this.host.GetService<ITestProjectRegexSetter>();
-            testProjectRegexSetter.AssertLocalServiceIsNotNull();
+            credentialStore = this.host.GetMefService<ICredentialStoreService>();
         }
 
         internal /*for testing purposes*/ ConnectionInformation ConnectedServer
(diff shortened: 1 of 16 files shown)


src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -49,17 +49,13 @@ internal class ConnectionWorkflow
...
-        private readonly ITestProjectRegexSetter testProjectRegexSetter;
 
         public ConnectionWorkflow(IHost host, ICommand parentCommand)
         {
             this.host = host ?? throw new ArgumentNullException(nameof(host));
             this.parentCommand = parentCommand ?? throw new ArgumentNullException(nameof(parentCommand));
 
-            this.credentialStore = this.host.GetMefService<ICredentialStoreService>();
-
-            this.testProjectRegexSetter = this.host.GetService<ITestProjectRegexSetter>();
```
duncanp-sonar, line 61: Previously, the connection workflow had a step that downloaded server settings so it could pass one of them to `ITestProjectRegexSetter`, which was then later used to filter out test projects.

The new unintrusive binding doesn't do any filtering since no project files are changed 
-> the `ITestProjectRegexSetter` doesn't need to be set 
-> the server settings don't need to be downloaded.

src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -228,30 +219,6 @@ private SonarQubeOrganization AskUserToSelectOrganizationOnUIThread(IEnumerable<
             });
         }
 
-        internal /*for testing purposes*/ async Task DownloadServiceParametersAsync(IProgressController controller,
```
duncanp-sonar, line 231: The only thing this step did was to pass one of the downloaded settings to the `ITestProjectRegexSetter` -> the whole method can now be deleted.

ugras-ergun-sonarsource (approved):

PR #4777, opened 2023-08-23
Remove ILocalService

### LINKED ISSUE #4773: [Cleanup] Remove ILocalService

duncanp-sonar:
Fixes #4773


src/Integration.UnitTests/MefServices/VsSessionHostTests.cs
```diff
@@ -335,20 +335,7 @@ public void VsSessionHost_ResetBinding_ErrorInReadingBinding()
         [TestMethod]
         public void VsSessionHost_IServiceProvider_GetService()
         {
-            // Arrange
```
duncanp-sonar, line 338: The code that is left in this test checks non-local services are returned.

I'm not sure what all the setup in the old `Arrange` lines 339-342 was for; my guess is that at one point the host fetched some VS services in the constructor.

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -269,36 +260,8 @@ private BindingConfiguration SafeGetBindingConfig()
...
-        public object GetService(Type serviceType)
-        {
-            // We don't expect COM types, otherwise the dictionary would have to use a custom comparer
-            Lazy<ILocalService> instanceFactory;
-            if (typeof(ILocalService).IsAssignableFrom(serviceType) &&
-                this.localServices.TryGetValue(serviceType, out instanceFactory))
-            {
-                return instanceFactory.Value;
-            }
+        public object GetService(Type serviceType) => serviceProvider.GetService(serviceType);
```
duncanp-sonar, line 263: Taking out all of the special-case code for local services just leaves this as a pass-through.

Logically, `IHost` doesn't need to be an `IServiceProvider` anymore since it isn't providing any services. However, I suspect that would create a lot of churn.

georgii-borovinskikh-sonarsource (approved):

PR #4780, opened 2023-08-23
Change IHost to not implement IServiceProvider

### LINKED ISSUE #4779: [Cleanup] IHost does not need to be a IServiceProvider
### Description

Once `ILocalService` has been removed (#4773) there is no reason for `IHost` to implement `IServiceProvider`.

This changed will generate some churn, but the the changes will be straightfoward and it will make future refactorings simpler.

duncanp-sonar:
Fixes #4779

### DIFF
--- a/src/Integration.TeamExplorer.UnitTests/SectionControllerTests.cs
+++ b/src/Integration.TeamExplorer.UnitTests/SectionControllerTests.cs
@@ -19,24 +19,21 @@
  */
 
 using System;
-using System.ComponentModel.Composition.Primitives;
 using System.ComponentModel.Design;
-using System.Windows.Threading;
 using FluentAssertions;
 using Microsoft.TeamFoundation.Controls;
-using Microsoft.VisualStudio.ComponentModelHost;
+using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Integration.TeamExplorer;
 using SonarLint.VisualStudio.Integration.WPF;
 using SonarLint.VisualStudio.TestInfrastructure;
-using SonarQube.Client.Models;
 using SonarQube.Client;
-
-using VS_OLECMD = Microsoft.VisualStudio.OLE.Interop.OLECMD;
+using SonarQube.Client.Models;
+using TF_IOleCommandTarget = Microsoft.TeamFoundation.Client.CommandTarget.IOleCommandTarget;
 using TF_OLECMD = Microsoft.TeamFoundation.Client.CommandTarget.OLECMD;
 using VS_IOleCommandTarget = Microsoft.VisualStudio.OLE.Interop.IOleCommandTarget;
-using TF_IOleCommandTarget = Microsoft.TeamFoundation.Client.CommandTarget.IOleCommandTarget;
+using VS_OLECMD = Microsoft.VisualStudio.OLE.Interop.OLECMD;
 using VS_OLEConstants = Microsoft.VisualStudio.OLE.Interop.Constants;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests.TeamExplorer
@@ -54,7 +51,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.TeamExplorer
             this.serviceProvider = new ConfigurableServiceProvider(assertOnUnexpectedServiceRequest: false);
 
             this.sonarQubeServiceMock = new Mock<ISonarQubeService>();
-            this.host = new ConfigurableHost(this.serviceProvider)
+            this.host = new ConfigurableHost()
             {
                 SonarQubeService = this.sonarQubeServiceMock.Object
             };
@@ -64,10 +61,14 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.TeamExplorer
         #region Tests
 
         [TestMethod]
-        public void SectionController_Ctor_ArgChecks()
+        public void MefCtor_CheckIsExported()
         {
-            Exceptions.Expect<ArgumentNullException>(() => new SectionController(null, new WebBrowser()));
-            Exceptions.Expect<ArgumentNullException>(() => new SectionController(this.host, null));
+            // Note: we're supplying a ConfigurableHost instance because an empty Mock host won't work here
+            // - we'd get an exception when the SectionController is disposed
+            MefTestHelpers.CheckTypeCanBeImported<SectionController, ITeamExplorerSection>(
+                MefTestHelpers.CreateExport<SVsServiceProvider>(),
+                MefTestHelpers.CreateExport<IHost>(new ConfigurableHost()),
+                MefTestHelpers.CreateExport<IWebBrowser>());
         }
 
         [TestMethod]
@@ -472,7 +473,7 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.TeamExplorer
 
         private SectionController CreateTestSubject(IWebBrowser webBrowser = null)
         {
-            var controller = new SectionController(host, webBrowser ?? new ConfigurableWebBrowser());
+            var controller = new SectionController(serviceProvider, host, webBrowser ?? new ConfigurableWebBrowser());
             controller.Initialize(null, new SectionInitializeEventArgs(new ServiceContainer(), null));
             return controller;
         }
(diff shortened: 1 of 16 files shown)


src/Integration.TeamExplorer.UnitTests/SectionControllerTests.cs
```diff
@@ -19,24 +19,21 @@
  */
 
 using System;
-using System.ComponentModel.Composition.Primitives;
 using System.ComponentModel.Design;
-using System.Windows.Threading;
 using FluentAssertions;
 using Microsoft.TeamFoundation.Controls;
-using Microsoft.VisualStudio.ComponentModelHost;
+using Microsoft.VisualStudio.Shell;
```
duncanp-sonar, line 25: These changes were made by the VS "Remove and sort usings" command.

src/Integration/MefServices/IHost.cs
```diff
@@ -26,7 +26,7 @@
 
 namespace SonarLint.VisualStudio.Integration
 {
-    internal interface IHost : IServiceProvider
```
duncanp-sonar, line 29: This is the root of all of the other changes: `IHost` no longer acts as a pass-through for `IServiceProvider`, so some components will now need to be given an `IServiceProvider` as well as an `IHost`.

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -70,7 +74,9 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
             this.gitEventsMonitor = gitEventsMonitor;
             this.logger = logger;
 
-            vsMonitorSelection = host.GetService<SVsShellMonitorSelection, IVsMonitorSelection>();
+            // TODO - MEF-ctor should be free-threaded -> we might be on a background thread ->
+            // calling serviceProvider.GetService is dangerous.
+            vsMonitorSelection = serviceProvider.GetService<SVsShellMonitorSelection, IVsMonitorSelection>();
```
duncanp-sonar, line 79: I've manually debugged to check this class and the other modified ones are created correctly at runtime.

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -70,7 +74,9 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
             this.gitEventsMonitor = gitEventsMonitor;
             this.logger = logger;
 
-            vsMonitorSelection = host.GetService<SVsShellMonitorSelection, IVsMonitorSelection>();
+            // TODO - MEF-ctor should be free-threaded -> we might be on a background thread ->
+            // calling serviceProvider.GetService is dangerous.
```
duncanp-sonar, line 78: Not sure how we'd fix this not-free-threaded issue, but I added this comment to call out the problem.

src/Integration/Binding/BindingWorkflow.cs
```diff
@@ -37,24 +37,17 @@ namespace SonarLint.VisualStudio.Integration.Binding
...
         private readonly IHost host;
-
         private readonly IBindingProcess bindingProcess;
 
-        public BindingWorkflow(IHost host,
+        public BindingWorkflow(IServiceProvider serviceProvider,
+            IHost host,
             IBindingProcess bindingProcess)
         {
-            if (host == null)
```
duncanp-sonar, line 47: MIscellaneous: I switched the argument checks to use ` ... ?? throw ...`

src/Integration.UnitTests/MefServices/VsSessionHostTests.cs
```diff
@@ -332,22 +329,6 @@ public void VsSessionHost_ResetBinding_ErrorInReadingBinding()
             this.stateManager.AssignedProjectKey.Should().BeNull();
         }
 
-        [TestMethod]
```
duncanp-sonar, line 335: `IHost` is no longer an `IServiceProvider` so this test is redundant.

src/Integration.UnitTests/MefServices/ActiveSolutionBoundTrackerTests.cs
```diff
@@ -563,15 +565,17 @@ public void ActiveSolutionChanged_GitEventsMonitorRefreshInvoked()
...
             IConfigurationProvider configurationProvider,
             ILogger logger = null,
-            IBoundSolutionGitMonitor gitEvents = null)
+            IBoundSolutionGitMonitor gitEvents = null,
+            IServiceProvider serviceProvider = null)
         {
             logger ??= new TestLogger(logToConsole: true);
             gitEvents ??= Mock.Of<IBoundSolutionGitMonitor>();
-            return new ActiveSolutionBoundTracker(host, solutionTracker, logger, gitEvents, configurationProvider);
+            serviceProvider ??= this.serviceProvider;
```
duncanp-sonar, line 577: This test class still has a `[TestInitialize]` that sets up most of the mocks used. Removing the `[TestInitialize]` would have created too much churn. Instead, I changed this method to allow tests to supply a custom `IServiceProvider` if required. By default, the one configured in `TestInitialize` is used.

src/Integration.UnitTests/Connection/ConnectionWorkflowTests.cs
```diff
@@ -86,11 +86,20 @@ public void TestInit()
 
         #region Tests
 
+        [TestMethod]
+        public void ConnectionWorkflow_ConnectionStep_WhenGivenANullServiceProvider_ThrowsArgumentNullException()
```
duncanp-sonar, line 90: New low-value test, but I added it since there were already tests for the other constructor arguments.

src/Integration.UnitTests/Connection/ConnectControllerTests.cs
```diff
@@ -55,11 +54,10 @@ public void TestInit()
             serviceProvider.RegisterService(typeof(ILogger), logger);
 
             this.sonarQubeServiceMock = new Mock<ISonarQubeService>();
-            this.connectionWorkflowMock = new Mock<IConnectionWorkflowExecutor>();
```
duncanp-sonar, line 58: Miscellaneous: most of the tests don't need the workflow executor, so I removed it from `TestInitialize` and added it to the tests that needed it.

src/Integration.TeamExplorer/SectionController.cs
```diff
@@ -49,21 +50,16 @@ internal class SectionController : TeamExplorerSectionBase, ISectionController
 
         internal const int CommandNotHandled = (int)Microsoft.VisualStudio.OLE.Interop.Constants.OLECMDERR_E_UNKNOWNGROUP;
 
+        private readonly IServiceProvider serviceProvider;
         private readonly IWebBrowser webBrowser;
 
         [ImportingConstructor]
-        public SectionController(IHost host, IWebBrowser webBrowser)
+        public SectionController(
+            [Import(typeof(SVsServiceProvider))]IServiceProvider serviceProvider,
```
duncanp-sonar, line 58: Now import an `IServiceProvider` as well as the host.

src/Integration.TeamExplorer/SectionController.cs
```diff
@@ -248,8 +244,8 @@ private void InitializeControllerCommands()
         {
             // Due to complexity of connect and bind we "outsource" the controlling part
             // to separate controllers which just expose commands
-            var connectionController = new Connection.ConnectionController(this.Host);
-            var bindingController = new Binding.BindingController(this.Host);
+            var connectionController = new Connection.ConnectionController(serviceProvider, Host);
+            var bindingController = new Binding.BindingController(serviceProvider, Host);
```
duncanp-sonar, line 248: TODO - we could probably now easily MEF-export the `ConnectionController` and `BindingController`, which would simplify the construction of the objects.

src/Integration.TeamExplorer.UnitTests/SectionControllerTests.cs
```diff
@@ -64,10 +61,14 @@ public void TestInitialize()
         #region Tests
 
         [TestMethod]
-        public void SectionController_Ctor_ArgChecks()
```
duncanp-sonar, line 67: MEF-created, so no point in checking for null arguments.

src/Integration.TeamExplorer.UnitTests/SectionControllerTests.cs
```diff
@@ -64,10 +61,14 @@ public void TestInitialize()
...
+        public void MefCtor_CheckIsExported()
```
duncanp-sonar, line 64: I didn't add the "check is singleton" or "check is non-shared" tests since we don't currently specify a `CreationPolicy` attribute on the class, and I wasn't sure what the correct value should be.

georgii-borovinskikh-sonarsource (approved):

PR #4789, opened 2023-08-24
Remove remaining uses of Dispatcher

### LINKED ISSUE #4774: [Threading] Remove direct use of Application.Current.Dispatcher
### Description

`IHost` exposed a `Dispatcher` property (to be removed in #3165).

However, `ConnectionWorkflow` directly accesses `Application.Current.Dispatcher.

TODO
- [ ] remove any uses of `Application.Current.Dispatcher`

duncanp-sonar:
Fixes #4774


duncanp-sonar:
This change gets rid of the last direct uses of `System.Windows.Threading.Dispatcher`. VS extensions shouldn't use this directly; they should use one of the VS threading helper classes (or, in our case, our `IThreadHandling` wrapper).

src/Integration.UnitTests/Connection/ConnectionWorkflowTests.cs
```diff
@@ -118,9 +119,7 @@ public void ConnectionWorkflow_ConnectionStep_ConnectedServerCanBeGetAndSet()
         {
             // Arrange
             var connectionInfo = new ConnectionInformation(new Uri("http://server"));
-#pragma warning disable IDE0017 // Simplify object initialization
-            ConnectionWorkflow testSubject = new ConnectionWorkflow(serviceProvider, host, new RelayCommand(() => { }));
-#pragma warning restore IDE0017 // Simplify object initialization
+            ConnectionWorkflow testSubject = CreateTestSubject(serviceProvider, host, new RelayCommand(() => { }));
```
duncanp-sonar, line 122: Changing the tests to use a `CreateTestSubject` helper made it easier to pass a dummy `IThreadHandling` instance to the constructor.

src/Integration.UnitTests/Connection/ConnectionWorkflowTests.cs
```diff
@@ -477,6 +468,11 @@ private void AssertCredentialsNotStored()
                 Times.Never());
         }
 
+        private static ConnectionWorkflow CreateTestSubject(IServiceProvider serviceProvider, IHost host, ICommand parentCommand)
+        {
+            return new ConnectionWorkflow(serviceProvider, host, parentCommand, new NoOpThreadHandler());
```
duncanp-sonar, line 473: I've hard-coded using a no-op thead handler here. There were no existing tests of the threading, and it would have been difficult to add any.

I manually debugged the product code and confirmed it is switching to/from the UI thread as expected.

src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -49,13 +50,22 @@ internal class ConnectionWorkflow
...
         private readonly ICredentialStoreService credentialStore;
 
         public ConnectionWorkflow(IServiceProvider serviceProvider, IHost host, ICommand parentCommand)
+            : this(serviceProvider, host, parentCommand, ThreadHandling.Instance)
+        { }
+
+        internal /* for testing */ ConnectionWorkflow(IServiceProvider serviceProvider,
+            IHost host,
+            ICommand parentCommand,
+            IThreadHandling threadHandling)
```
duncanp-sonar, line 63: New constructor for tests.

src/Integration/Connection/ConnectionWorkflow.cs
```diff
@@ -212,13 +222,16 @@ private void OnProjectsChanged(ConnectionInformation connection, IEnumerable<Son
...
+
+            threadHandling.RunOnUIThread2(() =>
             {
                 var organizationDialog = new OrganizationSelectionWindow(organizations) { Owner = Application.Current.MainWindow };
                 var hasUserClickedOk = organizationDialog.ShowDialog().GetValueOrDefault();
 
-                return hasUserClickedOk ? organizationDialog.Organization : null;
+                organization = hasUserClickedOk? organizationDialog.Organization : null;
             });
+            return organization;
```
duncanp-sonar, line 234: I manually debugged the product code and confirmed it is switching to/from the UI thread as expected.

src/Integration/MefServices/VsSessionHost.cs
```diff
@@ -90,8 +90,6 @@ void IProgressStepRunnerWrapper.ChangeHost(IProgressControlHost host)
         #region IHost
         public event EventHandler ActiveSectionChanged;
 
-        public Dispatcher UIDispatcher { get; }
```
duncanp-sonar, line 93: Miscellaneous: should have been removed in a previous PR, when I removed the property from the `IHost` interface.

src/Progress/MVVM/ViewModelBase.cs
```diff
@@ -20,154 +20,18 @@
...
-            this.stackPrint = new StackTrace().ToString();
-#endif
-        }
-
         /// <summary>
         ///  Occurs when a property value changes.
         /// </summary>
         public event PropertyChangedEventHandler PropertyChanged;
 
-        /// <summary>
```
duncanp-sonar, line 52: I wasn't looking forward to cleaning this code up. Had it been difficult to clean up I would have left it alone since it is only in the `Progress` library, which is only used by our UI in the `Team Explorer`. When the binding UI is rewritten, the UI won't be hosted in the `Team Explorer`, and this project can then be dropped.

src/Progress/MVVM/ViewModelBase.cs
```diff
@@ -177,7 +41,6 @@ protected void VerifyAccess()
         [SuppressMessage("Microsoft.Design", "CA1026:DefaultParametersShouldNotBeUsed", Justification = "Set by [CallerMemberName]")]
         protected virtual void RaisePropertyChanged([CallerMemberName]string propertyName = null)
         {
-            this.VerifyAccess();
```
duncanp-sonar, line 180: This call to `VerifyAccess()` was the only actual call to any of the methods in this class.

I considered refactoring the class to pass in an `IThreadHandling` instance to do the equivalent check, but decided it wasn't worth it. The existing code has not been changed for years, and shouldn't be changed before it is dropped.

Also, we don't bother checking the thread being used in the implementation of [ViewModelBase](https://github.com/SonarSource/sonarlint-visualstudio/blob/753f8b7275d3f9d94355603ee373b79e37d40e48/src/Integration/WPF/ViewModelBase.cs#L27) in `Integration`.

georgii-borovinskikh-sonarsource (approved):

Issue #4690, opened 2023-08-09
TimedUpdateHandler should only trigger updates when in ConnectedMode


duncanp-sonar:
### Description

Once the `TimedUpdateHandler` has been created it continues raising timed events, even if the current solution is in Standalone Mode, or even if there is no longer an open solution. This triggers updates for suppressions, hotspots and Quality Profiles.

It would be more efficient if the updates were only triggered when the solution is in Connected Mode. Even better, the timer wouldn't run unless in Connected Mode.

PR #4795, opened 2023-08-29
Add ActiveSolutionBoundTracker to TimedUpdateHandler

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4690

### DIFF
--- a/src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
+++ b/src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
@@ -68,6 +68,20 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
         }
 
         [TestMethod]
+        [DataRow(SonarLintMode.Standalone, false)]
+        public void Ctor_DependingOnBindingConfig_InitialTimeStateSetCorrectly(SonarLintMode mode, bool start)
+        {
+            var refreshTimer = new Mock<ITimer>();
+            var timerFactory = CreateTimerFactory(refreshTimer.Object);
+            var activeSolutionBoundTracker = CreateActiveSolutionBoundTrackerWihtBindingConfig(SonarLintMode.Standalone);
+
+            _ = CreateTestSubject(activeSolutionBoundTracker: activeSolutionBoundTracker.Object, timerFactory: timerFactory);
+
+            refreshTimer.Verify(x => x.Start(), start ? Times.Once : Times.Never);
+            refreshTimer.Verify(x => x.Stop(), start ? Times.Never : Times.Once );
+        }
+
+        [TestMethod]
         public void InvokeEvent_TimerElapsed_StoreUpdatersAreCalled()
         {
             var refreshTimer = new Mock<ITimer>();
@@ -117,16 +131,15 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
 
             _ = CreateTestSubject(activeSolutionBoundTracker:activeSolutionBoundTracker.Object, timerFactory:timerFactory);
 
-            refreshTimer.Verify(x => x.Start(), Times.Never);
-            refreshTimer.Verify(x => x.Stop(), Times.Once);
+            refreshTimer.Reset();
 
             activeSolutionBoundTracker.Raise(x => x.SolutionBindingChanged += null, new ActiveSolutionBindingEventArgs(CreateBindingConfiguration(SonarLintMode.Connected)));
             refreshTimer.Verify(x => x.Start(), Times.Once);
-            refreshTimer.Verify(x => x.Stop(), Times.Once);
+            refreshTimer.Verify(x => x.Stop(), Times.Never);
 
             activeSolutionBoundTracker.Raise(x => x.SolutionBindingChanged += null, new ActiveSolutionBindingEventArgs(CreateBindingConfiguration(SonarLintMode.Standalone)));
             refreshTimer.Verify(x => x.Start(), Times.Once);
-            refreshTimer.Verify(x => x.Stop(), Times.Exactly(2));
+            refreshTimer.Verify(x => x.Stop(), Times.Once);
         }
 
         private static ITimerFactory CreateTimerFactory(ITimer timer)
@@ -148,8 +161,8 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
                 suppressionIssueStoreUpdater ?? Mock.Of<ISuppressionIssueStoreUpdater>(),
                 serverHotspotStoreUpdater ?? Mock.Of<IServerHotspotStoreUpdater>(),
                 qualityProfileUpdater ?? Mock.Of<IQualityProfileUpdater>(),
-                new TestLogger(logToConsole: true),
                 activeSolutionBoundTracker,
+                new TestLogger(logToConsole: true),
                 timerFactory ?? Mock.Of<ITimerFactory>());
         }
 
--- a/src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
+++ b/src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
@@ -69,11 +69,13 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Suppressions
 
         [TestMethod]
         [DataRow(SonarLintMode.Standalone, false)]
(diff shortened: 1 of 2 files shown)


src/ConnectedMode/Suppressions/TimedUpdateHandler.cs
```diff
@@ -39,33 +40,56 @@ internal sealed class TimedUpdateHandler : IDisposable
...
+        }
+
+        private void SolutionBindingChanged(object sender, ActiveSolutionBindingEventArgs activeSolutionBindingEventArgs)
+        {
+            CreateTimerIfInConnectedMode(activeSolutionBindingEventArgs.Configuration);
+        }
+
+        private void CreateTimerIfInConnectedMode(BindingConfiguration bindingConfiguration)
+        {
+            if (bindingConfiguration?.Mode != SonarLintMode.Standalone)
```
bigfluffycookie, line 85: Instead of this we could also have a solution in which we create a new instance of the refreshTimer but I don't think it really makes sense considering we can just stop and start it

src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
```diff
@@ -112,7 +138,8 @@ private static ITimerFactory CreateTimerFactory(ITimer timer)
             return timerFactory.Object;
         }
 
-        private static TimedUpdateHandler CreateTestSubject(ISuppressionIssueStoreUpdater suppressionIssueStoreUpdater = null,
+        private static TimedUpdateHandler CreateTestSubject(IActiveSolutionBoundTracker activeSolutionBoundTracker,
```
bigfluffycookie, line 141: I was not sure if we should allow activeSolutionBoundTracker to be null here, personally, I am against it because I think neither standalone nor connected mode should be the default test state and that each test should define which state it is in.

src/ConnectedMode/Suppressions/TimedUpdateHandler.cs
```diff
@@ -39,33 +40,56 @@ internal sealed class TimedUpdateHandler : IDisposable
...
 
         private bool disposed;
 
         [ImportingConstructor]
         public TimedUpdateHandler(ISuppressionIssueStoreUpdater suppressionIssueStoreUpdater,
             IServerHotspotStoreUpdater serverHotspotStoreUpdater,
             IQualityProfileUpdater qualityProfileUpdater,
-            ILogger logger)
-            : this(suppressionIssueStoreUpdater, serverHotspotStoreUpdater, qualityProfileUpdater, logger, new TimerFactory()) { }
+            ILogger logger,
```
duncanp-sonar, line 51: Minor: the convention is to put `ILogger` last in the non-optional parameter list (the reasoning being that it is a boilerplate/utility class, and the other imports are more important from the point of view of understanding what the importing class so they should come first).

src/ConnectedMode/Suppressions/TimedUpdateHandler.cs
duncanp-sonar: Agreed; but the name of the method should change - `CreateTimer...` is a bit misleading.
It's not obvious what to call the method, though: `SetTimerStatus`, `EnableTimerInConnectedMode`, `StartStopTimer`?

src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
```diff
@@ -38,7 +39,8 @@ public void MefCtor_CheckIsExported()
                 MefTestHelpers.CreateExport<ISuppressionIssueStoreUpdater>(),
                 MefTestHelpers.CreateExport<IServerHotspotStoreUpdater>(),
                 MefTestHelpers.CreateExport<IQualityProfileUpdater>(),
-                MefTestHelpers.CreateExport<ILogger>());
+                MefTestHelpers.CreateExport<ILogger>(),
+                MefTestHelpers.CreateExport<IActiveSolutionBoundTracker>());
```
duncanp-sonar, line 43: Miscellaneous: could you add a "check is MEF singleton" test please?

Whenever I've touched an existing test file recently I've been checking it contains both of the "standard" MEF tests ("check is exported" and "check is/is not singleton") and adding them if necessary.

src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
duncanp-sonar: From the point of view of the tests we have to supply a non-null value, otherwise we'll get a null-ref exception in the constructor when it tries to register the event handler.

src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
```diff
@@ -103,6 +108,27 @@ public void Dispose_RefreshTimerDisposed_RaisingEventDoesNothing()
...
+        public void InvokeBindingChanged_TimerStartsStopsOnActiveSolutionBoundChange()
+        {
+            var refreshTimer = new Mock<ITimer>();
+            var timerFactory = CreateTimerFactory(refreshTimer.Object);
+            var activeSolutionBoundTracker = CreateActiveSolutionBoundTrackerWihtBindingConfig(SonarLintMode.Standalone);
+
+            _ = CreateTestSubject(activeSolutionBoundTracker:activeSolutionBoundTracker.Object, timerFactory:timerFactory);
+
+            refreshTimer.Verify(x => x.Start(), Times.Never);
+            refreshTimer.Verify(x => x.Stop(), Times.Once);
```
duncanp-sonar, line 121: Lines 120 and 121 are effectively checking that the constructor starts/stops the timer based on the initial binding state. However, it's only testing what happens if the solution is initially in standalone mode.

This perhaps should be a separate test i.e. that the constructor starts/stops the time based on the initial binding state. It could be a data-driven test.

duncanp-sonar (commented):
Some minor comments, and possibly a missing test case.

src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
bigfluffycookie: The test is right below this one haha

src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
```diff
@@ -55,25 +57,41 @@ public void Ctor_TimerIsSetupAsExpected()
...
             refreshTimer.Verify(x => x.Start(), Times.Once);
         }
 
+        [TestMethod]
+        [DataRow(SonarLintMode.Standalone, false)]
+        public void Ctor_DependingOnBindingConfig_InitialTimeStateSetCorrectly(SonarLintMode mode, bool start)
+        {
+            var refreshTimer = new Mock<ITimer>();
+            var timerFactory = CreateTimerFactory(refreshTimer.Object);
+            var activeSolutionBoundTracker = CreateActiveSolutionBoundTrackerWihtBindingConfig(SonarLintMode.Standalone);
```
duncanp-sonar, line 76: This line should be using the `mode` parameter.

src/ConnectedMode.UnitTests/Suppressions/TimedUpdateHandlerTests.cs
```diff
@@ -55,25 +57,41 @@ public void Ctor_TimerIsSetupAsExpected()
...
+            _ = CreateTestSubject(activeSolutionBoundTracker: activeSolutionBoundTracker.Object, timerFactory: timerFactory);
 
             refreshTimer.Object.AutoReset.Should().BeTrue();
             refreshTimer.Object.Interval.Should().Be(1000 * 60 * 10);
             refreshTimer.VerifyAdd(x => x.Elapsed += It.IsAny<EventHandler<TimerEventArgs>>(), Times.Once);
             refreshTimer.Verify(x => x.Start(), Times.Once);
         }
 
+        [TestMethod]
+        [DataRow(SonarLintMode.Standalone, false)]
```
duncanp-sonar, line 71: Needs a `DataRow` for the connected mode setting.

src/ConnectedMode/Suppressions/TimedUpdateHandler.cs
duncanp-sonar: `ILogger` should be the last argument in this constructor too.

duncanp-sonar (commented):
Minor issue in the new data-driven test...

duncanp-sonar (approved):

Issue #4785, opened 2023-08-24
[Cleanup] Remove references to VS RuleSet


duncanp-sonar:
### Description

We no longer generate or read legacy VS rulesets so the projects no longer need to reference those VS assemblies.

TODO
 - [ ] clean up test and product project files

PR #4801, opened 2023-08-29
Remove references to legacy RuleSet binaries

duncanp-sonar:
Fixees #4785


src/Integration.TeamExplorer.UnitTests/Integration.TeamExplorer.UnitTests.csproj
```diff
@@ -5,7 +5,6 @@
 
   <PropertyGroup Label="Specify which external/VS-specific refs are required">
     <RequiresTeamFoundation>true</RequiresTeamFoundation>
-    <RequiresRuleset>true</RequiresRuleset>
```
duncanp-sonar, line 8: We don't generate or update `.ruleset` files any more, so we don't need to reference those VS assemblies.

src/Integration.Vsix.UnitTests/ReferencesTests.cs
```diff
@@ -30,10 +29,11 @@ namespace SonarLint.VisualStudio.Integration.UnitTests
     public class ReferencesTests
     {
         [TestMethod]
-        public void MicrosoftVisualStudioCodeAnalysis_EnsureCorrectVersion()
+        public void MicrosoftVisualStudioVCProjectEngine_EnsureCorrectVersion()
         {
             var codeAnalysisAssemblyVersion = AssemblyHelper.GetVersionOfReferencedAssembly(
-                    typeof(SonarLint.VisualStudio.Integration.Binding.BindingController), "Microsoft.VisualStudio.CodeAnalysis");
```
duncanp-sonar, line 36: It isn't obvious from the GitHub diff, but I moved this test from `Integration.UnitTests` to `Integration.VSIX.UnitTests` and changed it to test a different assembly.

The VS ruleset assemblies are not included in the VS SDK. Also, we needed different versions for VS2019 and VS2022. So, we had checked-in binaries under the `third-parties` folder (deleted below) that we referenced in the projects.

This test checked that SLVS 2019 referenced the VS2019 RuleSet assembly and that SLVS 2022 referenced the VS2022 RuleSet assembly.

We no longer need this test to check the ruleset assembly. However, there is another third-party assembly similar to the RuleSet assembly that we didn't have a test for, namely `Microsoft.VisualStudio.VCProjectEngine`. So this test now validates that the correct version of that assembly is being used instead.

src/SonarLint.VSSpecificAssemblies.props
```diff
@@ -46,17 +46,6 @@
...
-  <!-- ******************************************************************************* -->
-  <!-- Legacy VS static analysis ruleset -->
-  <!-- ******************************************************************************* -->
-
-  <ItemGroup Label="Legacy VS CodeAnalysis" Condition=" $(RequiresRuleset)=='true' " >
-    <Reference Include="Microsoft.VisualStudio.CodeAnalysis">
-      <HintPath>$(ThirdPartyPath)\Microsoft.VisualStudio.CodeAnalysis.dll</HintPath>
-      <Private>$(CopyVSAssembliesToOutput)</Private>
-    </Reference>
-  </ItemGroup>
```
duncanp-sonar, line 58: This is what actually added the correct assembly reference to the project. No longer required.

(if you expand the file you'll see there are a couple of other third-party binaries that are still handled the same way).

third-parties/ReadMe.txt
```diff
@@ -0,0 +1,35 @@
+SLVS - third-party assemblies folder
```
duncanp-sonar, line 1: Having explained how this all works in review comments, I realised it would make sense to make a more permanent note of it somewhere nearer the code, so I added this read-me file.

bigfluffycookie (approved):
lgtm

Issue #4804, opened 2023-08-30
Map CCT severity to Vs Severity


georgii-borovinskikh-sonarsource:
Create the mapping from cct to vs severity, analogous to the old mapping

PR #4805, opened 2023-08-30
Add CCT severity to Vs Severity mapping

georgii-borovinskikh-sonarsource:
Implements #4804

### DIFF
--- a/src/IssueViz/Helpers/IAnalysisSeverityToVsSeverityConverter.cs
+++ b/src/IssueViz/Helpers/IAnalysisSeverityToVsSeverityConverter.cs
@@ -27,6 +27,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Helpers
     public interface IAnalysisSeverityToVsSeverityConverter
     {
         __VSERRORCATEGORY Convert(AnalysisIssueSeverity severity);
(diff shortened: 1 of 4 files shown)


src/IssueViz/Helpers/IAnalysisSeverityToVsSeverityConverter.cs
```diff
@@ -66,4 +86,14 @@ public __VSERRORCATEGORY Convert(AnalysisIssueSeverity severity)
             }
         }
     }
+
+    public static class AnalysisSeverityToVsSeverityConverterExtensions
```
ugras-ergun-sonarsource, line 90: Instead of adding this as a extension method wouldn't it make more sense to add it as public method to the IAnalysisSeverityToVsSeverityConverter and make the other two private?

georgii-borovinskikh-sonarsource: I can see your point, but I would need to change all of the tests. If you think it's worth it I'll do it

ugras-ergun-sonarsource: let's keep it for now but create a hardening ticket for the future.

ugras-ergun-sonarsource (approved):

PR #4812, opened 2023-08-31
Clean up references to NewtonSoft.json

### LINKED ISSUE #3098: [Infra] Clean up references to Newtonsoft.json
### Description

The project files contain out-of-date, inaccurate and redundant logic to handle references to the `NewtonSoft` NuGet package. See e.g. [here](https://github.com/SonarSource/sonarlint-visualstudio/blob/874168dd39ce7659efe325aa771ac8e786401944/Directory.Build.props#L31).

#### What's changed/inaccurate
* the Sonar analyzers no longer reference the `NewtonSoft` NuGet package so we're free to reference any version we want
* we no longer support VS2017 so we can reference a newer version of the package if required
* the comment in the props file says the `Newtonsoft` dll is embedded in the VSIX. This no longer appears to be the case.
* Mend/WhiteSource complains about the versions of `NewtonSoft` that we are referencing. The vulnerability it complains about apparently only affects IIS applications so is irrelevant. However, updating the package version would reduce the number of false positives that we need to deal with.

#### TODO
* decide which version of `NewtonSoft` to reference for VS2019 and VS2022.
* check whether we should be embedding a version of the `NewtonSoft` assembly in the VSIX, and if necessary include it. Note: we were embedding a version in old versions of SLVS (e.g. v4.7) but are not doing so now. Do we need to? Currently we are relying on VS having a compatible version installed.

duncanp-sonar:
Fixes #3098

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -20,26 +20,26 @@
   <!-- ******************************************************************************* -->
   <!-- Newtonsoft -->
   <!-- ******************************************************************************* -->
-  <!-- NOTE: we are referencing a version that is installed with VS2017 RTM. Newer versions  of VS ship with newer versions
-         of the Newtonsoft assembly and have binding redirects. This means that we don't need to ship the Newtonsoft dll
-         ourselves. 
-         However, the SonarAnalyzers from v4.25+ also depend on a different version of the Newtonsoft assembly which is
-         newer version than the one that ships in VS. This means the VSIX needs to contain that version of the Newtonsoft
-         assembly so that some of the analyzer rules can work (Integration.Vsix.csproj handles this).
-         
-         See #2474 for more info.
-   -->
-  <PropertyGroup>
-    <NewtonsoftVersion>8.0.3</NewtonsoftVersion>
-    <NewtonsoftVersion Condition=" $(VsTargetVersion) == '2022' ">13.0.1</NewtonsoftVersion>
-  </PropertyGroup>
+  <!-- NOTE: We're specifying the version we want to use and are embedding it in the VSIX.
+         However, VS ships with multiple verions of NewtonSoft.Json.dll and has binding redirects
+         in devenv.exe.config that can change the version is loaded at runtime. However, this
+         should not cause a problem as we will end up with a compatible version.
 
+         Also, VS2019 ships with NewtonSoft v12.x i.e. we are using a newer version. This
+         should not cause a problem since VS can load both versions.
+         
+         Finally, NewtonSoft.Json v12 has vulnerabilities reported against it which Mend
+         will complain about. By referencing v13.x we avoid those warnings.
+         
+         See #3098 for more info.
+   -->
+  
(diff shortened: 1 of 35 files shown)


Directory.Build.props
```diff
@@ -20,26 +20,26 @@
...
+         should not cause a problem as we will end up with a compatible version.
+
+         Also, VS2019 ships with NewtonSoft v12.x i.e. we are using a newer version. This
+         should not cause a problem since VS can load both versions.
+         
+         Finally, NewtonSoft.Json v12 has vulnerabilities reported against it which Mend
+         will complain about. By referencing v13.x we avoid those warnings.
          
-         See #2474 for more info.
+         See #3098 for more info.
```
duncanp-sonar, line 34: Updated comment: we're now embedding a specific version of the NewtonSoft dll, and we're using the same version for both VS2019 and VS2022.

duncanp-sonar: VS2019 ships with NewtonSoft.Json.dll v12.x. We're installing v13 on the same machine. When VS runs it will load both versions:

image

duncanp-sonar: With VS2022, it will only load one version:
image

src/CFamily.UnitTests/packages.lock.json
```diff
@@ -1019,8 +1019,8 @@
       },
       "Newtonsoft.Json": {
         "type": "Transitive",
-        "resolved": "13.0.1",
-        "contentHash": "ppPFpBcvxdsfUonNcvITKqLl3bqxWbDCZIzDWHzjpdAHRFfZe0Dw9HmA0+za13IdyrgJwpkDTDA9fHaxOrt20A=="
+        "resolved": "13.0.3",
```
duncanp-sonar, line 1022: Most of the changes are updates to `package.lock.json`. I'll add review comments for all of the non-package lock changes.

I've checked the `Integration.Vsix.UnitTests` package lock file has been updated correctly i.e. it doesn't have the extra "GraphModel" version changes that cause problems.

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -452,6 +452,7 @@
 
   <Target Name="IncludeNuGetPackageReferences" AfterTargets="GetVsixSourceItems">
     <ItemGroup>
+      <VSIXSourceItem Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' == 'NewtonSoft.Json'" />
```
duncanp-sonar, line 455: Added this line so the dll gets included in the VSIX. I checked both 2019 and 2022 VSIXes to make sure it is there.

duncanp-sonar:
I've installed the artefacts from the CI on the VS2019 and VS2022 VMs and checked that the user settings are correctly loaded (which means the NewtonSoft serialization code is being called).

bigfluffycookie (approved):

Issue #4748, opened 2023-08-21
Change links from wiki pages to the appropriate page on the new docs site


duncanp-sonar:
### Description

The new SonarLint documentation site is live: https://docs.sonarsource.com/sonarlint/visual-studio/

Existing in-product links to the old wiki pages should be changed to point to a page on the new site.

TODO:
* for each link, check for the best "new" page to link too. Some of the old links just pointed to the wiki home page as we didn't have a good story about how to handle redirection/renaming.

Issue #3665, opened 2023-02-08
Set minimal supported SonarQube version to 8.9


duncanp-sonar:
### Description

In line with the new policy on dropping support for older versions of SonarQube, we should stop supporting versions <= v7.9

Related SonarLint Core issse: [SLCORE-446](https://sonarsource.atlassian.net/browse/SLCORE-446)

TODO - what is the UX going to be?
- disallow new connections?
- refuse to connect existing connections that are too old?
- gold bar notification to warn users?
- any grace period?

pavel-mikula-sonarsource:
Moved to [SLVS-2225](https://sonarsource.atlassian.net/browse/SLVS-2225)

[SLVS-2225]: https://sonarsource.atlassian.net/browse/SLVS-2225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4822, opened 2023-08-31
Change all doc linkes to the new documentation website.

bigfluffycookie:
fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/4748


src/CFamily/CMake/Resources.resx
```diff
@@ -143,7 +143,7 @@
   </data>
   <data name="NoCompilationDatabaseFile" xml:space="preserve">
     <value>[CMake] Could not locate compilation database at '{0}'. Make sure that your project is configured correctly. 
-    See https://github.com/SonarSource/sonarlint-visualstudio/wiki for more information.</value>
+    See https://docs.sonarsource.com/sonarlint/visual-studio/getting-started/requirements/#language-specific-requirements for more information..</value>
```
bigfluffycookie, line 146: For this one there was a sublink -> https://docs.sonarsource.com/sonarlint/visual-studio/getting-started/requirements/#getting-started-vs2019-and-later however because it is version specific it felt wrong so I decided on using the parent link for this. LMK what you think would be better

src/Integration.Vsix.UnitTests/Settings/GeneralOptionsDialogPageTests.cs
```diff
@@ -130,7 +130,7 @@ public void ClickHyperlink_ShowWikiCommandIsCalled()
             page.Control.ShowWikiHyperLink.Command.Execute(null);
 
             // Assert
-            browserService.Verify(x => x.Navigate("https://github.com/SonarSource/sonarlint-visualstudio/wiki"), Times.Once);
+            browserService.Verify(x => x.Navigate("https://docs.sonarsource.com/sonarlint/visual-studio/team-features/connected-mode/"), Times.Once);
```
bigfluffycookie, line 133: This link might change as I had to ask kevin for a more detailed sub link. It is mentioned in our Options window with the text "Note: per-user settings are ignored for C/C++ for solutions using connected mode. See the  *wiki link* for more information." so if there is nothing more precise from kevin I think linking to the connected-mode page makes most sense.

src/Integration.Vsix/Settings/GeneralOptionsDialogPage.cs
```diff
@@ -33,7 +33,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix
     internal class GeneralOptionsDialogPage : UIElementDialogPage
     {
         public const string PageName = "General";
-        private const string WikiUrl = "https://github.com/SonarSource/sonarlint-visualstudio/wiki";
+        private const string WikiUrl = "https://docs.sonarsource.com/sonarlint/visual-studio/team-features/connected-mode/";
```
bigfluffycookie, line 36: This link might change as I had to ask kevin for a more detailed sub link. It is mentioned in our Options window with the text "Note: per-user settings are ignored for C/C++ for solutions using connected mode. See the wiki link for more information." so if there is nothing more precise from kevin I think linking to the connected-mode page makes most sense.

src/IssueViz.Security/Taint/TaintResources.resx
```diff
@@ -137,6 +137,6 @@
   </data>
   <data name="Synchronizer_UnsupportedSQVersion" xml:space="preserve">
     <value>[Taint] Displaying taint vulnerabilities in the IDE requires SonarQube v8.6 or later, or SonarCloud. Connected SonarQube version: v{0}
-    Visit https://github.com/SonarSource/sonarlint-visualstudio/wiki to find out more about this and other SonarLint features.</value>
+    Visit https://docs.sonarsource.com/sonarlint/visual-studio/using-sonarlint/taint-vulnerabilities/ to find out more about this and other SonarLint features.</value>
```
bigfluffycookie, line 140: I got a bit confused here as in the text it says we need SonarQube v8.6 or later however in the documentation here -> https://docs.sonarsource.com/sonarlint/visual-studio/using-sonarlint/taint-vulnerabilities/ it says v8.9 or higher. Is that a mistake on our side or?

src/CFamily/CMake/Resources.resx
duncanp-sonar: Yep, I agree that the parent link makes more sense.

src/IssueViz.Security/Taint/TaintResources.resx
duncanp-sonar: Hhmm. It looks from the code that we support the API call from [v8.6](https://github.com/SonarSource/sonarlint-visualstudio/blob/753f8b7275d3f9d94355603ee373b79e37d40e48/src/SonarQube.Client/Api/DefaultConfiguration.cs#L56) onwards.
I had a quick look in the code but couldn't find any checks against 8.9, so it looks like it should be 8.6.

duncanp-sonar (commented):
Looks good.

Last week @kevin-hinz pointed out an approach one of the SL squads had used to centralise the links: create a class just containing constants for the help URLs. That makes it easier to find out where the doc links are used, and easier to modify them centrally.
What do you think?

duncanp-sonar (approved):
Nice!

src/IssueViz.Security/Taint/TaintResources.resx
kevin-hinz: For future reference, we have [a version support policy](https://sonar-documentations-preview.netlify.app/sonarlint/visual-studio/team-features/connected-mode/#sonarlint-sonarqube-version-support-policy) which states we support v8.9+ until sometime around Dec 2023, then only back to the latest LTS.

duncanp-sonar: FYI SLVS currently works with older SQ releases. There is a ticket to remove support for older versions: #3665.

kevin-hinz:
The following comment was added (in the CMS) to the URL destinations updated in this PR: _The URL for this article/page is an in-product link. Please advise SLVS team before changing the URL slug._

Issue #4770, opened 2023-08-22
[Cleanup] [Test] Fix unit tests that don't correctly mock IProjectPropertyManager


duncanp-sonar:
### Description

Three sets of units tests that have `IProjectPropertyManager` as a dependency create an actual instance of `ProjectPropertyManager` rather than using a mock. This makes the tests far more complicated than necessary.

The test classes are:

- [x] * ProjectExcludePropertyToggleCommandTests
- [x] * ProjectSonarLintMenuCommandTests
- [x] * ProjectTestPropertySetCommandTests

PR #4828, opened 2023-09-01
Use mock of IProjectPropertyManager rather than an actual instance in ProjectSonarLintMenuCommandTests

bigfluffycookie:
fixes part of https://github.com/SonarSource/sonarlint-visualstudio/issues/4770

### DIFF
--- a/src/Integration.Vsix.UnitTests/Commands/ProjectSonarLintMenuCommandTests.cs
+++ b/src/Integration.Vsix.UnitTests/Commands/ProjectSonarLintMenuCommandTests.cs
@@ -31,26 +31,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Commands
     [TestClass]
     public class ProjectSonarLintMenuCommandTests
     {
-        #region Test boilerplate
-
-        private ConfigurableVsProjectSystemHelper projectSystem;
-
-        // TODO - cleanup. These unit tests are supposed to be testing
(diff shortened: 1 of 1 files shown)


src/Integration.Vsix.UnitTests/Commands/ProjectSonarLintMenuCommandTests.cs
```diff
@@ -31,26 +31,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Commands
     [TestClass]
     public class ProjectSonarLintMenuCommandTests
     {
-        #region Test boilerplate
```
bigfluffycookie, line 34: Removed TestInitialize because it was not really necessary

src/Integration.Vsix.UnitTests/Commands/ProjectSonarLintMenuCommandTests.cs
```diff
@@ -88,16 +68,23 @@ public void ProjectSonarLintMenuCommand_QueryStatus_NoProjects_IsDisableIsHidden
         public void ProjectSonarLintMenuCommand_QueryStatus_HasUnsupportedProject_IsDisabledIsHidden()
```
bigfluffycookie, line 68: I rearranged both tests a bit so it is a bit more readable (imo)

duncanp-sonar (approved):

PR #4832, opened 2023-09-01
Use mock of IProjectPropertyManager rather than an actual instance in ProjectExcludePropertyToggleCommandTests

bigfluffycookie:
fixes part of https://github.com/SonarSource/sonarlint-visualstudio/issues/4770

### DIFF
--- a/src/Integration.Vsix.UnitTests/Commands/ProjectExcludePropertyToggleCommandTests.cs
+++ b/src/Integration.Vsix.UnitTests/Commands/ProjectExcludePropertyToggleCommandTests.cs
@@ -19,7 +19,9 @@
  */
 
 using System;
+using EnvDTE;
 using FluentAssertions;
+using FluentAssertions.Common;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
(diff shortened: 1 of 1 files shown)


src/Integration.Vsix.UnitTests/Commands/ProjectExcludePropertyToggleCommandTests.cs
```diff
@@ -31,141 +33,79 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Commands
...
-            act = () => new ProjectExcludePropertyToggleCommand(propertyManager, null);
+            act = () => new ProjectExcludePropertyToggleCommand(Mock.Of<IProjectPropertyManager>(), null);
             act.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("projectToLanguageMapper");
         }
 
         [TestMethod]
-        public void ProjectExcludePropertyToggleCommand_Invoke_SingleProject_TogglesValue()
+        [DataRow(true, null)]
+        [DataRow(false, true)]
+        public void Invoke_SingleProject_TogglesValue(bool initialExcludeValue, bool? expectedSetValue)
```
bigfluffycookie, line 51: This test is the same as discussed offline.

src/Integration.Vsix.UnitTests/Commands/ProjectExcludePropertyToggleCommandTests.cs
```diff
@@ -31,141 +33,79 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.Commands
...
         }
 
         [TestMethod]
-        public void ProjectExcludePropertyToggleCommand_Invoke_MultipleProjects_MixedPropValues_SetIsExcludedTrue()
+        [DataRow(true, true, null, null, 2)]
+        [DataRow(false, false, true, true, 2)]
+        [DataRow(null, false, true, true, 1)]
+        [DataRow(null, true, true, true, 1)]
+        public void Invoke_MultipleProjects_TogglesValues(bool? initialExcludeValue1, bool? initialExcludeValue2,
+            bool? expectedSetValue1, bool? expectedSetValue2, int nrOfCallsToGetBoolean)
```
bigfluffycookie, line 80: Here I merge two tests into one, 1. MultipleProjects same properties 2. Multiple Projects Different properties.

The only thing I don't really like is the nrOfCallsToGetBoolean but my brain is fried. If you have a better suggesting lmk!

src/Integration.Vsix.UnitTests/Commands/ProjectExcludePropertyToggleCommandTests.cs
```diff
@@ -181,221 +121,145 @@ public void ProjectExcludePropertyToggleCommand_QueryStatus_MissingPropertyManag
...
-            command.Checked.Should().BeTrue("Expected command to be checked");
+            command.Checked.Should().IsSameOrEqualTo(expectedCheckedValue);
         }
 
         [TestMethod]
-        public void ProjectExcludePropertyToggleCommand_QueryStatus_MultipleProjects_MixedPropValues_IsUnchecked()
+        [DataRow(true, true, true)]
+        [DataRow(false, false, false)]
+        [DataRow(true, false, false)]
+        public void QueryStatus_MultipleProjects_AllSupported_MixedSupport_CorrectCommandState(bool supportedValue1, bool supportedValue2, bool expectedCommandState)
```
bigfluffycookie, line 207: This is also 2 merged tests the same as the previous one

src/Integration.Vsix.UnitTests/Commands/ProjectExcludePropertyToggleCommandTests.cs
duncanp-sonar: My brain hurt trying to work out what the product code was supposed to be doing...

Perhaps the test is being over-specific i.e. it is testing too much. The most important bit of the functionality here is whether the properties are set correctly, not how many times they are read.
Also, it looks like all of the properties should be set to the same value (either `null` or `true`), so `expectedSetValue2` shouldn't be necessary.


If I understand the intention of the code correctly, it's the following:
* if the initial property values are inconsistent then all properties should be set to `true`.
* if the initial property values are identical then they should be toggled: (`true` -> `null`, `false`/`null` -> true)


Assuming all of the above is correct, the test could be simplified to three input values e.g.

```
        [TestMethod]
        [DataRow(true, true, null)]   // Same across projects -> toggle
        [DataRow(false, false, true)] // Same across projects -> toggle
        [DataRow(null, null, true)]   // Same across projects (null === false) -> set to true
        [DataRow(null, false, true)]  // Same across projects (null === false) -> set to true
        [DataRow(null, true, true)]   // Different across projects -> set to true
        [DataRow(false, true, true)]  // Different across projects -> set to true
        public void Invoke_MultipleProjects_ExpectedValueIsSet(bool? initialExcludeValue1, bool? initialExcludeValue2,
            bool? expectedSetValue)
        {
           ...
        }
```

duncanp-sonar (commented):
Looks good. One suggestion about simplifying the "toggle" test.

duncanp-sonar (approved):

Issue #4842, opened 2023-09-05
Fix MEF importing constructors - Definitely not ok, but probably straightforward to fix


bigfluffycookie:
First sub group of classes that need to be fixed in ticket https://github.com/SonarSource/sonarlint-visualstudio/issues/4512

- [x] * **FolderWorkspaceService**: calls `GetService`. Could be deferred until needed. Or could refactor a bit more and make `ISolutionInfoProvider` a wrapper round the `IVsSolution` and make it handle all of the threading correctly.
- [x] * ~~**WorkspaceService**: calls `GetService`. Could be dropped entirely in favour of `ISolutionInfoProvider` instead, which is thread-aware.~~
- [x] * **TelemetryPayloadCreator**: makes an indirect service call that could be deferred via `vsVersion = vsVersionProvider.Version;`
- [x] * **UserSettingsProvider**: loads file from disc. Doesn't seem to call VS services.

PR #4852, opened 2023-09-06
Fix UserSettingsProvider constructor to be free threaded

bigfluffycookie:
Part of https://github.com/SonarSource/sonarlint-visualstudio/issues/4842

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/UserSettingsProviderTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/UserSettingsProviderTests.cs
@@ -161,7 +161,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.Analysis.UnitTests
             testSubject.EnsureFileExists();
 
             // Assert
-            fileSystemMock.Verify(x => x.File.Exists(fileName), Times.Exactly(2));
+            fileSystemMock.Verify(x => x.File.Exists(fileName), Times.Exactly(1));
             fileSystemMock.Verify(x => x.File.WriteAllText(It.IsAny<string>(), It.IsAny<string>()), Times.Never);
         }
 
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/Analysis/UserSettingsProvider.cs
```diff
@@ -70,7 +69,15 @@ private void OnFileChanged(object sender, EventArgs e)
 
         #region IUserSettingsProvider implementation
 
-        public UserSettings UserSettings { get; private set; }
+        public UserSettings UserSettings
+        {
+            get
```
bigfluffycookie, line 74: I get this -> https://rules.sonarsource.com/csharp/RSPEC-2190/ for both the set and get (Add a way to break out of this property) and I don't fully understand considering this is not like, a loop or anything

duncanp-sonar: `if (UserSettings == null)...` in line 76 will call the `get` method from inside the `get` method -> recursion. Have you tried debugging it?

bigfluffycookie: No, for some reason on my machine it wont run the tests for this class (I click run and something compiles and then nothing happens) so I wanted to upload it to see if it runs there. And yes your right, In my head I had added another private variable to use for checking and assigning but clearly that did not happen

src/Integration.Vsix.UnitTests/Analysis/UserSettingsProviderTests.cs
```diff
@@ -161,7 +161,7 @@ public void EnsureFileExists_NotCreatedIfExists()
             testSubject.EnsureFileExists();
 
             // Assert
-            fileSystemMock.Verify(x => x.File.Exists(fileName), Times.Exactly(2));
+            fileSystemMock.Verify(x => x.File.Exists(fileName), Times.Exactly(1));
```
duncanp-sonar, line 164: It might be worth adding a unit test that checks that the constructor doesn't call any of the services we know are not free-threaded e.g. logger. Alternatively, it might be simpler just to check that it doesn't call methods on the supplied services.

duncanp-sonar (approved):
Looks good. Just a suggestion about adding a test.

PR #4855, opened 2023-09-06
[Infra] Add helper method to check if UI thread is required

### LINKED ISSUE #4854: [Infra] Add helper class to check whether the UI thread is required
### Description

The VS threading cookbook describes scenarios that should be  free-threaded and includes some sample code to check a particular set of operations is free-threaded or not.

TODO - add a helper class with documentation to make it easier to test whether code is free-threaded during development.

duncanp-sonar:
Fixes #4854

### DIFF
--- a/src/Infrastructure.VS/DEBUG_ONLY_VsCookbookThreadingChecker.cs
+++ b/src/Infrastructure.VS/DEBUG_ONLY_VsCookbookThreadingChecker.cs
@@ -1,0 +1,154 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/Infrastructure.VS/DEBUG_ONLY_VsCookbookThreadingChecker.cs
```diff
@@ -0,0 +1,154 @@
...
+    // **
+    // **  Not be to used in shipping production code !
+    // **
+    // **  Only use for testing whether code is free-threaded
+    // **  in local dev environments.
+    // **
+    // ******************************************************
+
+
+    /* From the VS threading cookbook: https://github.com/microsoft/vs-threading/blob/main/doc/cookbook_vs.md
```
duncanp-sonar, line 38: Added helper methods based on the sample code in the VS thread cookbook.

src/Infrastructure.VS/DEBUG_ONLY_VsCookbookThreadingChecker.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// <summary>
+        /// If this method is called on the UI thread it will check
+        /// whether the operation _requires_ the UI thread.
+        /// If the method is called from a background thread it will just execute the operation.
+        /// </summary>
+        public static void IfOnUIThreadCheckIfRequiresMainThread(Action operationToTest)
+        {
+            if (ThreadHelper.CheckAccess())
+            {
+                Debug.WriteLine("On UI thread: testing operation...");
```
duncanp-sonar, line 73: I'm deliberately not using any of our other classes e.g. `ILogger` as they might skew the results.

src/Infrastructure.VS/DEBUG_ONLY_VsCookbookThreadingChecker.cs
```diff
@@ -0,0 +1,154 @@
...
+
+    */
+    public static class DEBUG_ONLY_VsCookbookThreadingChecker
+    {
+        /// <summary>
+        /// If this method is called on the UI thread it will check
+        /// whether the operation _requires_ the UI thread.
+        /// If the method is called from a background thread it will just execute the operation.
+        /// </summary>
+        public static void IfOnUIThreadCheckIfRequiresMainThread(Action operationToTest)
```
duncanp-sonar, line 69: There are four methods:

* _IfOnUIThreadCheckIfRequiresMainThread(Action)_  : only does the check if the caller is on the UI thread.
* _CheckIfRequiresMainThread(Action)_  : does the check

... and equivalent methods for `Func<T>`.

bigfluffycookie (approved):
Looks good! we need to remember to fill in the others about this class so they can use it in the future! :)

Issue #4856, opened 2023-09-07
Fix MEF importing constructors - Possibly not ok


bigfluffycookie:
Sub Group of Ticket https://github.com/SonarSource/sonarlint-visualstudio/issues/4512

- [ ] **LocalHotspotsStore**: calls InitializeServerHotspots(this.serverHotspotStore.GetAll()); What does this do?
- [ ] **CssEslintBridgeClient**: calls eslintBridgeProcessFactory.Create() and creates several other objects.  ditto JavaScriptEslintBridgeClient and TypeScriptEslintBridgeClient
- [ ] **SonarErrorListDataSource**: calls var errorTableManager = tableManagerProvider.GetTableManager(StandardTables.ErrorsTable);. Check what the VS Extensibility sample in GitHub does - it might be ok.
- [x] **SecretsAnalyzer**: calls ruleSettingsProvider = ruleSettingsProviderFactory.Get(Language.Secrets)
- [x] **ImportBeforeInstallTrigger**: calls TriggerUpdateAsync().Forget(); // easy to fix - directly create by a package -> make the package call it
- [x] **SSESessionManager**: calls CreateSessionIfInConnectedMode(activeSolutionBoundTracker.CurrentConfiguration); // does this switch threads?
- [x] **SonarQubeNavigationItem**: makes calls to fetch resources. Needs a bit more investigation.
- [x] **TypeScript.Analyzer.AnalyzerBase**: calls rulesProviderFactory and eslintBridgeAnalyzerFactory. What do these do, and could they be deferred?
- [x] **NodeLocationsProvider**: creates various INodeLocationProvider instances - need to check them.

pavel-mikula-sonarsource:
Moved to [SLVS-2187](https://sonarsource.atlassian.net/browse/SLVS-2187)

[SLVS-2187]: https://sonarsource.atlassian.net/browse/SLVS-2187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4857, opened 2023-09-07
Make ImportingConstructor in ImportBeforeInstallTrigger free threaded

bigfluffycookie:

### DIFF
--- a/src/ConnectedMode/Install/ImportBeforeInstallTrigger.cs
+++ b/src/ConnectedMode/Install/ImportBeforeInstallTrigger.cs
@@ -55,7 +55,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Install
 
         private void OnPreSolutionBindingUpdated(object sender, EventArgs e) => TriggerUpdateAsync().Forget();
 
-        internal async Task TriggerUpdateAsync()
+        public async Task TriggerUpdateAsync()
         {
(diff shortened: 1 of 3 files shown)


duncanp-sonar (approved):
Nice

src/ConnectedMode/Install/ImportBeforeInstallTrigger.cs
```diff
@@ -60,7 +55,7 @@ private void OnPreSolutionBindingChanged(object sender, ActiveSolutionBindingEve
 
         private void OnPreSolutionBindingUpdated(object sender, EventArgs e) => TriggerUpdateAsync().Forget();
 
-        private async Task TriggerUpdateAsync()
+        internal async Task TriggerUpdateAsync()
```
duncanp-sonar, line 58: By convention, this should be `public` rather than `internal` since it is part of the public API that we expect other product classes to use.

We use `internal` on methods/properties to indicate the method/property should only be used by the declaring type and from test code.

duncanp-sonar (approved):
Looks good; one minor comment about the scope of the public method.

src/ConnectedMode/Install/ImportBeforeInstallTrigger.cs
bigfluffycookie: Ah good catch, thank you!

PR #4860, opened 2023-09-07
[Infra] Add helper class to encapsulate using a VS service

### LINKED ISSUE #4859: [Infra] Add helper class to wrap fetching and using a VS service
### Description

Fetching a VS service and performing an operation on it is a common task. 

Handling the threading correctly requires the calling code to use `IServiceProvider` and `IThreadHandling`. Repeating this code in multiple classes is time-consuming and error-prone, as is adding the appropriate tests. It would be simpler to encapsulate the logic in a helper class.

That would simplify fixing the many instances where the code doesn't currently do the right think e.g. many of the non-free-threaded MEF constructors - see #4512

duncanp-sonar:
Fixes #4859


src/Infrastructure.VS/IVsUIServiceOperation.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// <summary>
+    /// Helper to simplify fetching and using VS services that need to
+    /// be called on the UI thread
+    /// </summary>
+    /// <remarks>The class handles thread switching when required. All of the methods
+    /// return on the type of thread that they were called on i.e.
+    /// * returns on the UI thread if called from the UI thread
+    /// * returns on a background thread if called from the background thread
+    /// </remarks>
+    public interface IVsUIServiceOperation
```
duncanp-sonar, line 39: New helper class. I use the same casing format as `IVsUIShell`.

src/Infrastructure.VS/VsInfoService.cs
```diff
@@ -41,18 +40,18 @@ public interface IVsInfoService
     [PartCreationPolicy(CreationPolicy.Shared)]
     internal class VsInfoService : IVsInfoService
     {
+        private readonly Lazy<string> installRootDir;
+
         [ImportingConstructor]
-        public VsInfoService([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider)
+        public VsInfoService(IVsUIServiceOperation vSServiceOperation)
```
duncanp-sonar, line 46: Converted the service to use the new service operation rather than using `IServiceProvider` directly. This class previously didn't handle threading at all; using the new service operation helper fixes that.

src/Infrastructure.VS.UnitTests/VsInfoServiceTests.cs
```diff
@@ -34,18 +33,32 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests
     public class VsInfoServiceTests
     {
         [TestMethod]
-        public void MefCtor_CheckIsExported()
+        public void MefCtor_CheckIsExported()
```
duncanp-sonar, line 36: Added the three standard MEF tests for the existing class.

src/Infrastructure.VS.UnitTests/VsInfoServiceTests.cs
```diff
@@ -54,22 +67,34 @@ public void Create_VsShellCallSucceeds_ReturnsExpectedPath()
...
+            var serviceOp = CreateConfiguredServiceOperation("c:\\test\\", shellHrResult: -123);
 
-            Action act = () => new VsInfoService(serviceProvider.Object);
+            var testSubject = new VsInfoService(serviceOp);
+
+            Action act = () => _ = testSubject.InstallRootDir;
             act.Should().ThrowExactly<COMException>();
         }
 
-        private Mock<IServiceProvider> CreateConfiguredServiceProvider(string installDirectory, int shellHrResult = VSConstants.S_OK)
```
duncanp-sonar, line 63: Fix up the existing tests to use the new service.
I've also manually debugged through the code in VS to check the threading works as expected.

src/Infrastructure.VS.UnitTests/VsUIServiceOperationTests.cs
```diff
@@ -0,0 +1,319 @@
...
+
+            // The MEF constructor should be free-threaded, which it will be if
+            // it doesn't make any external calls.
+            serviceProvider.Invocations.Should().BeEmpty();
+            threadHandling.Invocations.Should().BeEmpty();
+        }
+
+        #endregion
+
+        #region RunOnUIThread tests
```
duncanp-sonar, line 71: As discussed offline, there are four methods on the interface, all very similar, to give cover all of the combinations `[sync / async] + [ action / function ]`.

I've grouped the tests by scenario.

src/Infrastructure.VS.UnitTests/VsUIServiceOperationTests.cs
```diff
@@ -0,0 +1,319 @@
...
+                "get service",
+                "callback");
+        }
+
+        [TestMethod]
+        public void Execute_Func_RunOnUIThread()
+        {
+            (var calls, var testSubject) = SetupUpThreadHandlingTests<SMyInterface>();
+
+            string callback(IMyInterface x)
```
duncanp-sonar, line 94: FYI I had originally defined this as a `Func<IMyInterface>` variable, but an IDE quick fix suggested changing it to a local function instead.

src/Infrastructure.VS.UnitTests/VsUIServiceOperationTests.cs
```diff
@@ -0,0 +1,319 @@
...
+
+            return (callback, testSubject);
+        }
+
+        #endregion
+
+        #region Multiple calls, same service tests
+
+        // NOTE - we're only testing this scenario for Execute(Action)
+        // Adding tests for all of the other methods doesn't add enough value to justify the effort at this point.
```
duncanp-sonar, line 220: As discussed offline, I couldn't work out an easy way to avoid a lot of duplication in this test and the next one, so there are only tests for one entry point.

src/Infrastructure.VS.UnitTests/VsUIServiceOperationTests.cs
```diff
@@ -0,0 +1,320 @@
...
+            // Act
+            _ = await testSubject.ExecuteAsync<SMyInterface, IMyInterface, string>(callback);
+
+            calls.Should().ContainInOrder(
+                "change thread - async",
+                "get service",
+                "callback");
+        }
+
+        private static (IList<string> calls, VsUIServiceOperation testSubject) SetupUpThreadHandlingTests<S>()
```
bigfluffycookie, line 145: Minor: Setup**Up**

src/Infrastructure.VS.UnitTests/VsUIServiceOperationTests.cs
```diff
@@ -0,0 +1,320 @@
...
+            (var callback, var testSubject) = SetupUpFuncReturnsValueTests("expected async result");
+
+            // Act
+            var result = await testSubject.ExecuteAsync<SMyInterface, IMyInterface, string>(callback.Object);
+
+            result.Should().Be("expected async result");
+            callback.Verify(x => x.Invoke(It.IsAny<IMyInterface>()), Times.Once);
+        }
+
+        private static (Mock<Func<IMyInterface, string>> callback, VsUIServiceOperation vsServiceOperation) SetupUpFuncReturnsValueTests(string callbackResultToReturn)
```
bigfluffycookie, line 203: Same typo here

src/Infrastructure.VS.UnitTests/VsUIServiceOperationTests.cs
```diff
@@ -0,0 +1,320 @@
...
+            await testSubject.ExecuteAsync<SVsShellMonitorSelection, IVsMonitorSelection, int>(callback2.Object);
+
+            serviceProvider.Verify(x => x.GetService(typeof(SVsShellMonitorSelection)), Times.Once);
+            serviceProvider.Invocations.Should().HaveCount(2);
+            callback2.Verify(x => x.Invoke(service2), Times.Once);
+        }
+
+        #endregion
+
+        private static Mock<IServiceProvider> CreateConfiguredServiceProvider<S>(object serviceToReturn)
```
bigfluffycookie, line 308: Question, couldn't you do the following so you could reuse this in the test method above here? 

```
private static Mock<IServiceProvider> CreateConfiguredServiceProvider<S>(param object[] servicesToReturn)
{
var serviceProvider = new Mock<IServiceProvider>();
foreach(var service in servicesToReturn)
{
            serviceProvider.Setup(x => x.GetService(typeof(S))).Returns(service );
}
return serviceProvider;
}
```

bigfluffycookie (approved):
Looks good, just some small stuff :)

src/Infrastructure.VS.UnitTests/VsUIServiceOperationTests.cs
duncanp-sonar: Nice idea; I didn't think of that. It wouldn't quite work in that form since we want to set up different service types i.e. `<S>` changes each time. We could pass in an array of `(Type, Object)` tuples e.g.


```
        private static Mock<IServiceProvider> CreateConfiguredServiceProvider(params (Type serviceType, object serviceToReturn)[] serviceTuple)
        {
            var serviceProvider = new Mock<IServiceProvider>();
            foreach((var serviceType, var serviceToReturn) in serviceTuple)
            { 
                serviceProvider.Setup(x => x.GetService(serviceType)).Returns(serviceToReturn);
            }
            return serviceProvider;
        }
``` 


From the callers point of view, it works well if there are multiple services but it's a bit clunkier for a single service:
```
            // Multiple services
            var serviceProvider1 = CreateConfiguredServiceProvider(
                (typeof(SVsUIShell), service1),
                (typeof(SVsShellMonitorSelection), service2));

            // Single service
            var serviceProvider2 = CreateConfiguredServiceProvider(
                (typeof(SMyInterface), new object()));

            var serviceProvider3 = CreateConfiguredServiceProvider<SMyInterface>(new object());

```

I'll add it as an overload.

PR #4862, opened 2023-09-08
Make ctor in NodeLocationsProvider free threaded.

bigfluffycookie:


src/TypeScript.UnitTests/NodeJSLocator/NodeLocationsProviderTests.cs
```diff
@@ -41,15 +42,38 @@ public void MefCtor_CheckIsExported()
                 MefTestHelpers.CreateExport<ILogger>());
         }
 
+        [TestMethod]
```
bigfluffycookie, line 45: The I added are the same as always except for the ProvidersListIsLazy test.

duncanp-sonar (approved):
Nice

PR #4863, opened 2023-09-08
Make SSESessionManager ctor free threaded

bigfluffycookie:

### DIFF
--- a/src/ConnectedMode/ConnectedModePackage.cs
+++ b/src/ConnectedMode/ConnectedModePackage.cs
@@ -43,7 +43,7 @@ namespace SonarLint.VisualStudio.ConnectedMode
     [Guid("dd3427e0-7bb2-4a51-b00a-ddae2c32c7ef")]
     public sealed class ConnectedModePackage : AsyncPackage
     {
-        private ISSESessionManager sseSessionManager;
+        private SSESessionManager sseSessionManager;
         private IIssueServerEventsListener issueServerEventsListener;
         private IQualityProfileServerEventsListener qualityProfileServerEventsListener;
         private ServerSuppressionsChangedHandler serverSuppressionsChangedHandler;
@@ -64,7 +64,7 @@ namespace SonarLint.VisualStudio.ConnectedMode
 
             logger.WriteLine(Resources.Package_Initializing);
 
-            sseSessionManager = componentModel.GetService<ISSESessionManager>();
+            LoadServicesAndDoInitialUpdates(componentModel);
 
             issueServerEventsListener = componentModel.GetService<IIssueServerEventsListener>();
             issueServerEventsListener.ListenAsync().Forget();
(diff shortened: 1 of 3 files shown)


src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
```diff
@@ -27,16 +27,9 @@
...
 {
-    /// <summary>
-    /// Reacts to project changes and opens/closes Server Sent Events sessions
-    /// </summary>
-    internal interface ISSESessionManager : IDisposable
-    {
-    }
-
-    [Export(typeof(ISSESessionManager))]
+    [Export(typeof(SSESessionManager))]
```
bigfluffycookie, line 30: We dont use this anywhere else so I removed the interface so it is the same as the TriggerImportInstall class

src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -84,17 +84,24 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
...
-            // See https://github.com/SonarSource/sonarlint-visualstudio/issues/3886
+            logger.WriteLine(Resources.Package_Initialized);
+        }
+
+        /// <summary>
+        /// Trigger an initial update of suppressions (These classes might have missed the initial solution binding
+        /// event from the ActiveSolutionBoundTracker)
+        /// See https://github.com/SonarSource/sonarlint-visualstudio/issues/3886
+        /// </summary>
+        private void LoadServicesAndDoInitialUpdates(IComponentModel componentModel)
```
bigfluffycookie, line 95: With how many there are now I thought it made sense to put it in a seperate method.

src/ConnectedMode/ConnectedModePackage.cs
```diff
@@ -64,7 +64,7 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
 
             logger.WriteLine(Resources.Package_Initializing);
 
-            sseSessionManager = componentModel.GetService<ISSESessionManager>();
+            LoadServicesAndDoInitialUpdates(componentModel);
```
bigfluffycookie, line 67: I moved it up as on line 85 there is an await and this way it doesnt disturb the load order as much

bigfluffycookie:
A little confused about the dispose code smell as I didnt change anything on that, should I fix it?

src/ConnectedMode/ServerSentEvents/SSESessionManager.cs
```diff
@@ -27,16 +27,9 @@
 
 namespace SonarLint.VisualStudio.ConnectedMode.ServerSentEvents
 {
-    /// <summary>
-    /// Reacts to project changes and opens/closes Server Sent Events sessions
-    /// </summary>
-    internal interface ISSESessionManager : IDisposable
```
duncanp-sonar, line 33: I think you need to make `SSESessionManager : IDisposable` now (one of the code smells).

src/ConnectedMode/ConnectedModePackage.cs
duncanp-sonar: The doc comment on the method needs to updated.

duncanp-sonar (commented):
Looks good. A couple of minor comments.

duncanp-sonar (approved):

Issue #4861, opened 2023-09-08
Fix MEF importing constructors - calls GetService


duncanp-sonar:
Sub Group of Ticket https://github.com/SonarSource/sonarlint-visualstudio/issues/4512

Depends on #4859

TODO
 - [x] ActiveDocumentLocator
- [x] VsInfoService
- [x] AbsoluteFilePathLocator
- [x] ToolWindowService
- [x] TeamExplorerController
- [x] VcxRequestFactory
- [x] StatusBarNotifier
- [x] InfoBarManager -> Only needs a test to confirm it is free threaded 
- [x] IssueLocationActionsSourceProvider
- [x] StatusRequestHandler 
- [x] ErrorListHelper 
- [x] TaintIssuesSynchronizer
- [x] VsVersionProvider (in review)

PR #4867, opened 2023-09-08
Make ctor in ToolWindowService free-threaded

duncanp-sonar:
Part of #4861

### DIFF
--- a/src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
+++ b/src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
@@ -22,12 +22,11 @@ using System;
 using System.Runtime.InteropServices;
 using FluentAssertions;
 using Microsoft.VisualStudio;
-
-using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Infrastructure.VS;
 using SonarLint.VisualStudio.Integration.Helpers;
 using SonarLint.VisualStudio.TestInfrastructure;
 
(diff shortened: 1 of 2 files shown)


src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
```diff
@@ -40,29 +39,35 @@ public class ToolWindowServiceTests
 
         [TestMethod]
         public void MefCtor_CheckIsExported()
-        {
-            // Set up a service provider with the required service
-            var serviceProviderMock = new Mock<IServiceProvider>();
-            serviceProviderMock.Setup(x => x.GetService(typeof(SVsUIShell))).Returns(Mock.Of<IVsUIShell>());
```
duncanp-sonar, line 46: Another example where the old MEF test was ugly.

Replaced with the three standard tests.

src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
```diff
@@ -40,29 +39,35 @@ public class ToolWindowServiceTests
...
         {
-            var serviceProviderMock = new Mock<IServiceProvider>();
             var uiShellMock = new Mock<IVsUIShell>();
             var frameMock = new Mock<IVsWindowFrame>();
+            SetupFindToolWindow(uiShellMock, VSConstants.E_OUTOFMEMORY, ValidToolWindowId, frameMock.Object);
 
-            SetupFindToolWindow(serviceProviderMock, uiShellMock, VSConstants.E_OUTOFMEMORY, ValidToolWindowId, frameMock.Object);
+            var serviceOp = CreateServiceOperation(uiShellMock.Object);
 
-            var testSubject = new ToolWindowService(serviceProviderMock.Object);
```
duncanp-sonar, line 65: Again, I added the conventional `CreateTestSubject` method, even though there is only one parameter to the constructor at the moment.

src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
```diff
@@ -40,29 +39,35 @@ public class ToolWindowServiceTests
...
 
         [TestMethod]
         public void Show_FindWindowFailed_NoExceptionAndFrameNotShown()
         {
-            var serviceProviderMock = new Mock<IServiceProvider>();
             var uiShellMock = new Mock<IVsUIShell>();
             var frameMock = new Mock<IVsWindowFrame>();
+            SetupFindToolWindow(uiShellMock, VSConstants.E_OUTOFMEMORY, ValidToolWindowId, frameMock.Object);
 
-            SetupFindToolWindow(serviceProviderMock, uiShellMock, VSConstants.E_OUTOFMEMORY, ValidToolWindowId, frameMock.Object);
```
duncanp-sonar, line 63: Previously the `SetupFindToolWindow` method was doing two separate tasks:
1. setting up the `IServiceProvider` and 
2. configure the mock shell service.

I've removed step (1) from `SetupFindToolWindow` so that set up for the new service operation is separate.

We actually have two different setup methods for the service operation, one that sets up a sync `Action`, and another that sets up a sync `Func`.

src/Integration.UnitTests/Helpers/ToolWindowServiceTests.cs
```diff
@@ -157,13 +164,37 @@ private class MyDummyToolWindow : IMyDummyToolWindow
...
+            var serviceOp = new Mock<IVsUIServiceOperation>();
+
+            // Set up the mock to invoke the operation with the supplied VS service
+            serviceOp.Setup(x => x.Execute<SVsUIShell, IVsUIShell>(It.IsAny<Action<IVsUIShell>>()))
+                .Callback<Action<IVsUIShell>>(op => op(svcToPassToCallback));
+
+            return serviceOp.Object;
+        }
+
+        private IVsUIServiceOperation CreateServiceOperation<V>(IVsUIShell svcToPassToCallback)
```
duncanp-sonar, line 185: This overload sets up the sync `Func` overload.

I'm thinking about adding some test helper methods in `TestInfrastructure` to help with setting up the various `Execute/ExecuteAsync` overrides, but I'm not sure if it's worth it yet. I'll see when I've fixed a few more of the constructors.

src/Integration/Helpers/ToolWindowService.cs
```diff
@@ -21,45 +21,59 @@
...
 
-            return (V)docView;
+                    return (V)docviewObject;
+                });
+
+            return docView;
         }
 
-        private IVsWindowFrame GetOrCreateWindowFrame(Guid toolWindowId)
+        private static IVsWindowFrame GetOrCreateWindowFrame(IVsUIShell shell, Guid toolWindowId)
```
duncanp-sonar, line 76: Miscelleanous: can now be static.

src/Integration/Helpers/ToolWindowService.cs
```diff
@@ -21,45 +21,59 @@
...
         }
 
         public void Show(Guid toolWindowId)
         {
-            var frame = GetOrCreateWindowFrame(toolWindowId);
-            frame?.Show();
+            vsUIServiceOperation.Execute<SVsUIShell, IVsUIShell>(
+                shell =>
+                {
+                    var frame = GetOrCreateWindowFrame(shell, toolWindowId);
```
duncanp-sonar, line 48: I looked at refactoring the code to see if there was a way to avoid putting the  `vsUIServiceOperation` call into every interface method implementation, but couldn't see one.

bigfluffycookie (approved):
Nice

PR #4875, opened 2023-09-11
Make ctor in TeamExplorerController free threaded

bigfluffycookie:


src/Integration.TeamExplorer.UnitTests/TeamExplorerControllerTests.cs
```diff
@@ -31,52 +33,56 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.TeamExplorer
     public class TeamExplorerControllerTests
     {
         [TestMethod]
-        public void TeamExplorerController_Ctor_NullArgChecks()
-        {
-            Exceptions.Expect<ArgumentNullException>(() => new TeamExplorerController(null));
-        }
+        public void MefCtor_CheckIsExported()
```
bigfluffycookie, line 36: Added all the Mef tests

src/Integration.TeamExplorer/TeamExplorerController.cs
```diff
@@ -21,40 +21,32 @@
...
-        private readonly ITeamExplorer teamExplorer;
-
-        internal /* testing purposes */ ITeamExplorer TeamExplorer => this.teamExplorer;
+        private readonly IVsUIServiceOperation vSServiceOperation;
 
         [ImportingConstructor]
-        public TeamExplorerController([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider)
+        public TeamExplorerController(IVsUIServiceOperation vSServiceOperation)
         {
-            if (serviceProvider == null)
```
bigfluffycookie, line 42: I removed this instead of checking if vsserviceOperation is null because we have a mef test to check that the import is successful

duncanp-sonar (commented):
Looks good. One question about the resource string.

duncanp-sonar (approved):

src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -298,13 +298,14 @@ Referenced assemblies:
 - 'PresentationCore, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
 - 'PresentationFramework, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
 - 'SonarLint.VisualStudio.ConnectedMode, Version=7.3.0.0, Culture=neutral, PublicKeyToken=c5b62af9de6d7244'
+- 'SonarLint.VisualStudio.Infrastructure.VS, Version=7.3.0.0, Culture=neutral, PublicKeyToken=c5b62af9de6d7244'
```
duncanp-sonar, line 301: @bigfluffycookie FYI the CI build was failing because there was a difference between the baseline version of the AsmRef file and the version generated on the CI machine during the build.

I checked out your branch locally and rebuilt, which automatically updated the baseline file to include this change. `SL.VS.TeamExplorer` now uses the service operation class from the `SL.VS.Infrastructure` assembly so it is expected that it will reference the `SL.VS.Infrastructure` assembly.

PR #4878, opened 2023-09-12
Make ctor in VcxrequestFactory free threaded

bigfluffycookie:

### DIFF
--- a/src/Integration.Vsix/CFamily/VcxProject/VcxRequestFactory.cs
+++ b/src/Integration.Vsix/CFamily/VcxProject/VcxRequestFactory.cs
@@ -44,33 +44,39 @@ namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject
     {
         private readonly ICFamilyRulesConfigProvider cFamilyRulesConfigProvider;
         private readonly IThreadHandling threadHandling;
-        private readonly IFileConfigProvider fileConfigProvider;
+        private readonly Lazy<IFileConfigProvider> fileConfigProvider;
         private readonly ILogger logger;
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/CFamily/VcxProject/VcxRequestFactory.cs
```diff
@@ -44,33 +44,39 @@ internal class VcxRequestFactory : IRequestFactory
...
             ICFamilyRulesConfigProvider cFamilyRulesConfigProvider,
-            ILogger logger)
+            ILogger logger,
+            IThreadHandling threadHandling)
             : this(serviceProvider,
                 cFamilyRulesConfigProvider,
-                new FileConfigProvider(logger),
+                new Lazy<IFileConfigProvider>(() => new FileConfigProvider(logger)),
                 logger,
-                ThreadHandling.Instance)
```
bigfluffycookie, line 59: Moved this up because it is mef importable, this way its more consistent to all the other classes

bigfluffycookie:
The code smell is confusing to me

duncanp-sonar (approved):
Nice

duncanp-sonar:
> The code smell is confusing to me

Hhm. It does seem to be in the wrong place, and there is a `ThrowIfNotOnUIThread` call. You could just mark it as a `False Positive`; that, or use the VS `Suppress in file` context menu action to add a pragma to suppress it in the code.

I thought the CAYC team said that they would not turn on the "0 new defects" policy for projects that import external Roslyn issues; presumably they've changed their minds, or didn't check to see whether SLVS imported external Roslyn issues.

PR #4884, opened 2023-09-13
Make ctor of StatusRequestHandler free threaded

bigfluffycookie:

### DIFF
--- a/src/IssueViz.Security/OpenInIDE/Api/StatusRequestHandler.cs
+++ b/src/IssueViz.Security/OpenInIDE/Api/StatusRequestHandler.cs
@@ -18,13 +18,11 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System;
 using System.ComponentModel.Composition;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Shell.Interop;
 using SonarLint.VisualStudio.Infrastructure.VS;
 using SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Contract;
-using Task = System.Threading.Tasks.Task;
 
 namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Api
 {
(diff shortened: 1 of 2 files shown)


src/IssueViz.Security/OpenInIDE/Api/StatusRequestHandler.cs
```diff
@@ -31,22 +31,34 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Api
...
         {
-            vsShell = serviceProvider.GetService(typeof(SVsShell)) as IVsShell;
-            vsSolution = serviceProvider.GetService(typeof(SVsSolution)) as IVsSolution;
+            this.vSServiceOperation = vSServiceOperation;
         }
 
         Task<IStatusResponse> IStatusRequestHandler.GetStatusAsync()
         {
-            vsShell.GetProperty((int) __VSSPROPID5.VSSPROPID_AppBrandName, out var ideName);
+            var result = vSServiceOperation.Execute<SVsShell, IVsShell, Task<IStatusResponse>>(
```
bigfluffycookie, line 44: I mean, this is not the ugliest thing ever considering theres only 2 services we need however I am unsure how I'd l^solve this if there were more than 2

duncanp-sonar: It's a bit surprising we haven't hit an example that uses multiple services before...

One option would be to change the `IVsUIServiceOperation` interface (or create a separate interface) that is designed to work with multiple services
e.g. by passing `IServiceProvider` into the callback instead of a specific service, so the consuming could request any services it likes. The service operation class would be responsible for handling the thread transitions, and the consuming code would still be simpler as it would still only import the `IVsUIServiceOperation` interface.

As it is, the current code could be structured differently to make it more readable by avoiding the nesting e.g.

```
        async Task<IStatusResponse> IStatusRequestHandler.GetStatusAsync()
        {
            var ideName = await GetIdeNameAsync();
            var ideInstance = await GetIdeInstanceAsync();

            return new StatusResponse(ideName, ideInstance);
        }

        private async Task<string> GetIdeNameAsync()
            => await vSServiceOperation.ExecuteAsync<SVsShell, IVsShell, string>(
                shell =>
                {
                    shell.GetProperty((int)__VSSPROPID5.VSSPROPID_AppBrandName, out var ideName);
                    return ideName as string ?? "Microsoft Visual Studio";
                });

        private async Task<string> GetIdeInstanceAsync()
            => await vSServiceOperation.ExecuteAsync<SVsSolution, IVsSolution, string>(
                solution =>
                {
                    solution.GetProperty((int)__VSPROPID.VSPROPID_SolutionBaseName, out var ideInstance);
                    return ideInstance as string ?? "";
                });
```

This version is slightly less efficient as it will have to acquire the UI thread twice, but it won't be noticeable in practice.

src/IssueViz.Security/OpenInIDE/Api/StatusRequestHandler.cs
```diff
@@ -31,22 +31,34 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.OpenInIDE.Api
...
+        private Task<IStatusResponse> DoGetStatusAsync(IVsShell shell, IVsSolution solution)
+        {
+            shell.GetProperty((int)__VSSPROPID5.VSSPROPID_AppBrandName, out var ideName);
             ideName = ideName ?? "Microsoft Visual Studio";
 
-            vsSolution.GetProperty((int) __VSPROPID.VSPROPID_SolutionBaseName, out var ideInstance);
+            solution.GetProperty((int)__VSPROPID.VSPROPID_SolutionBaseName, out var ideInstance);
             ideInstance = ideInstance ?? "";
 
             return Task.FromResult<IStatusResponse>(new StatusResponse(ideName.ToString(), ideInstance.ToString()));
```
duncanp-sonar, line 64: There's something a bit weird about this; this method returns a `Task` and service operation has an `ExecuteAsync` method, so it shouldn't be necessary to use `Task.FromResult<...>`.

duncanp-sonar (commented):
Looks good. A couple of suggestions on how to simplify/remove the nested lambdas.

duncanp-sonar (approved):

PR #4888, opened 2023-09-13
[Infra] Fix MefTestHelper to not fail on exceptions in Dispose

### LINKED ISSUE #4886: [Infra] Fix MEF compose tests to not fail on Dispose errors
### Description

The standard MEF helper test `CheckTypeCanBeImported` currently disposes the imported objects when the composition container goes out of scope.

Most of the time this isn't a problem. However, some of our classes will throw exceptions because they have been set up with mocked implementations that behave as expected when disposing.

The goal of the test helper is to check that a type can be imported, so exceptions during disposal should not prevent the test from passing.

duncanp-sonar:
Fixes #4886

### DIFF
--- a/src/TestInfrastructure/Helpers/MefTestHelpers.cs
+++ b/src/TestInfrastructure/Helpers/MefTestHelpers.cs
@@ -152,8 +152,7 @@ namespace SonarLint.VisualStudio.TestInfrastructure
                 batch.AddExport(item);
             }
 
-            using var container = new CompositionContainer(catalog);
-            container.Compose(batch);
+            DoCompose(catalog, batch);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -152,8 +152,7 @@ public static void Compose(object[] importers, Type[] typesToExport, params Expo
                 batch.AddExport(item);
             }
 
-            using var container = new CompositionContainer(catalog);
```
duncanp-sonar, line 155: The `using` meant the container was disposed as soon as the method ending, which could then throw exceptions when the imported objects are disposed.

src/TestInfrastructure/Helpers/MefTestHelpers.cs
```diff
@@ -177,13 +176,46 @@ public static void Compose(Assembly exportingAssembly, params object[] importers
...
+            try
+            {
+                container.Compose(batch);
+
+            }
+            finally
+            {
+                SafeDispose(container);
+            }
+        }
```
duncanp-sonar, line 195: Long-hand version of a `using` statement, with a clean-up step that ignores errors during disposal.
If the composition failed then the original `CompositionException` will be re-raised when the `finally` block has completed.

bigfluffycookie (approved):

PR #4893, opened 2023-09-15
Make TaintIssuesSynchronizer ctor free threaded

bigfluffycookie:

### DIFF
--- a/src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
+++ b/src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
@@ -63,6 +63,33 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.UnitTests.Taint
         }
 
         [TestMethod]
+        public void MefCtor_DoesNotCallAnyServices_BesidesExpected()
+        {
+            var taintStore = new Mock<ITaintStore>();
+            var sonarQubeService = new Mock<ISonarQubeService>();
+            var taintIssueToIssueVisualizationConverter = new Mock<ITaintIssueToIssueVisualizationConverter>();
+            var configurationProvider = new Mock<IConfigurationProvider>();
+            var statefulServerBranchProvider = new Mock<IStatefulServerBranchProvider>();
+            var vsUIServiceOperation = new Mock<IVsUIServiceOperation>();
+            var toolWindowService = new Mock<IToolWindowService>();
+            var logger = new Mock<ILogger>();
+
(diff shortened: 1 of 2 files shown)


src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
```diff
@@ -190,7 +185,16 @@ private void ClearStore()
             taintStore.Set(Enumerable.Empty<IAnalysisIssueVisualization>(), null);
         }
 
-        private void UpdateTaintIssuesUIContext(bool hasTaintIssues) =>
-            vsMonitorSelection.SetCmdUIContext(contextCookie, hasTaintIssues ? 1 : 0);
+        private void UpdateTaintIssuesUIContext(bool hasTaintIssues)
```
bigfluffycookie, line 188: Was able to move everything in here, I don't think calling getcmdcookie multiple times should be a problem

duncanp-sonar: Probably not... does VS return the same value for the cookie on each call to `GetCmdUIContextCookie`? If so, then it should be ok.

src/IssueViz.Security.UnitTests/Taint/TaintIssuesSynchronizerTests.cs
```diff
@@ -54,11 +57,38 @@ public void MefCtor_CheckIsExported()
...
                 MefTestHelpers.CreateExport<IStatefulServerBranchProvider>(),
                 // The constructor calls the service provider so we need to pass a correctly-configured one
-                MefTestHelpers.CreateExport<SVsServiceProvider>(CreateServiceProvider()),
+                MefTestHelpers.CreateExport<IVsUIServiceOperation>(),
                 MefTestHelpers.CreateExport<IToolWindowService>(),
                 MefTestHelpers.CreateExport<ILogger>());
         }
 
+        [TestMethod]
+        public void MefCtor_DoesNotCallAnyServices_BesidesExpected()
```
duncanp-sonar, line 66: `_BesidesExpected` - the constructor doesn't call any services, does it?

duncanp-sonar (commented):
Looks good. Just a couple of questions...

src/IssueViz.Security/Taint/TaintIssuesSynchronizer.cs
bigfluffycookie: Here it says "Note Once a GUID is registered through this method, the GUID remains registered until the IDE exits." So from my understanding yes it always returns the same one

https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsmonitorselection.getcmduicontextcookie?view=visualstudiosdk-2022

Issue #4892, opened 2023-09-14
[Editor perf] Investigate whether we can avoid creating taggers by specifying TextViewRoles for tagger providers


duncanp-sonar:
### Description

Our `IssueLocationAdornmenTaggerProvider` tagger provider has the attribute `[TextViewRole(PredefinedTextViewRoles.PrimaryDocument)]` to limit when it is created by VS.

None of our other taggers providers do.

TODO - investigate whether we could apply `TextViewRole` to limit when other tagger providers are created.
Are some of our tagger providers currently being created for e.g. diff windows, peek windows etc, and would setting the attribute prevent it?

georgii-borovinskikh-sonarsource:
Doesn't affect buffer taggers. View taggers in our solution already have that.

PR #4894, opened 2023-09-15
Make Tagger Providers only support PrimaryDocument TextViewRole

georgii-borovinskikh-sonarsource:
Issue #4892


georgii-borovinskikh-sonarsource:
No impact on buffer taggers

PR #4897, opened 2023-09-18
Fix part of ctor in ActiveSolutionBoundTracker

bigfluffycookie:

### DIFF
--- a/src/Integration/MefServices/ActiveSolutionBoundTracker.cs
+++ b/src/Integration/MefServices/ActiveSolutionBoundTracker.cs
@@ -22,10 +22,10 @@ using System;
 using System.ComponentModel.Composition;
 using System.Diagnostics;
 using System.Threading;
-using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Infrastructure.VS;
 using SonarLint.VisualStudio.Integration.State;
 using SonarQube.Client;
 using Task = System.Threading.Tasks.Task;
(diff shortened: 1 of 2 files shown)


src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -88,8 +85,6 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
 
             CurrentConfiguration = configurationProvider.GetConfiguration();
```
bigfluffycookie, line 86: I really tried to get rid of this by adding another private variable currentConfiguration and if its null call GetConfiguration. 

However I ran into a variety of issues with the tests and considering how complicated the tests are I dont think it makes sense for me to change it. The main problem came with the RaiseAnalyzersChangedIfBindingChanged() method and the tests expecting it to be setup in a certain way. The whole test class needs a rewrite I think

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -88,8 +85,6 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
...
 
-            SetBoundSolutionUIContext();
```
bigfluffycookie, line 91: I am pretty confident that removing this should be fine but maybe Im completely off

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -47,21 +47,23 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
...
         private readonly IActiveSolutionTracker solutionTracker;
         private readonly IConfigurationProvider configurationProvider;
-        private readonly IVsMonitorSelection vsMonitorSelection;
         private readonly IBoundSolutionGitMonitor gitEventsMonitor;
+        private readonly IVsUIServiceOperation vSServiceOperation;
         private readonly ILogger logger;
-        private readonly uint boundSolutionContextCookie;
 
         public event EventHandler<ActiveSolutionBindingEventArgs> PreSolutionBindingChanged;
+
```
bigfluffycookie, line 55: My auto format removed the white spaces but I'll leave it considering we usually format it like this

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
duncanp-sonar: I think it will cause a problem, unfortunately. The `BoundSolutionUIContext` is used to trigger the loading of the `ConnectedModePackage` and the `TaintSyncPackage`  (they both have `[ProvideAutoLoad(BoundSolutionUIContext.GuidString, PackageAutoLoadFlags.BackgroundLoad)]`).

A problem scenario would be the following:
1. user opens VS
 - none of our packages are loaded yet
2. user opens a bound solution
  - VS will start loading some of our packages in the background, but it will take some time and probably won't finish until after the solution has finished loading.

In particular, the `ActiveSolutionBoundTracker` could be loaded after the bound solution has finished opening. If `SetBoundSolutionUIContext()` isn't called in the constructor, the other packages won't load until the user opens/closes another solution.

bigfluffycookie:
More complicated than anticipated. Moved the task back to the original list it was in.

Issue #4905, opened 2023-09-20
Remove unused binding code (BindingProcessState)


georgii-borovinskikh-sonarsource:
The class is not used (except some default values), remove it and all of the parameters passed into it

PR #4906, opened 2023-09-20
Remove BindingProcessState & related unused code

georgii-borovinskikh-sonarsource:
Implements #4905


src/ConnectedMode/Binding/BindingProcessImpl.cs
```diff
@@ -114,39 +108,6 @@ public async Task<bool> SaveServerExclusionsAsync(CancellationToken cancellation
             return true;
         }
 
-        public bool BindOperationSucceeded => InternalState.BindingOperationSucceeded;
```
georgii-borovinskikh-sonarsource, line 117: This is the only usage of this class and it's always the default value

src/ConnectedMode/Binding/BindingProcessImpl.cs
```diff
@@ -47,8 +47,7 @@ internal class BindingProcessImpl : IBindingProcess
             IExclusionSettingsStorage exclusionSettingsStorage,
             ISonarQubeService sonarQubeService,
             IQualityProfileDownloader qualityProfileDownloader,
-            ILogger logger,
-            bool isFirstBinding = false)
```
georgii-borovinskikh-sonarsource, line 51: Only used by BindingProcessState which is unused

duncanp-sonar (approved):
Looks good. There are probably some other bits of binding code that are now obsolete following the massive simplification done in v7.

PR #4912, opened 2023-09-28
[Infra] Add debug-only package to simplify local debugging of SLVS2019

duncanp-sonar:


src/Integration.Vsix/SlvsDebugPackage.cs
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+using System;
+using System.Runtime.InteropServices;
+using System.Threading;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell;
+using Task = System.Threading.Tasks.Task;
+
+#if DEBUG
```
duncanp-sonar, line 63: -> won't exist in CI builds since they are RELEASE by default.
If you want a DEBUG build on the CI machine you can manually trigger a build and override the default `BuildConfiguration` value:

image

duncanp-sonar:
As discussed offline. I've built and tested this locally, and it works as documented in the code.

georgii-borovinskikh-sonarsource (approved):

PR #4923, opened 2023-10-05
[CAYC] Fix code smells

duncanp-sonar:


duncanp-sonar:
Fixes 6 out of the 7 code smells in new code.

src/Integration.Vsix/CFamily/VcxProject/VcxRequestFactory.cs
```diff
@@ -53,11 +51,15 @@ internal class VcxRequestFactory : IRequestFactory
             ICFamilyRulesConfigProvider cFamilyRulesConfigProvider,
             ILogger logger,
             IThreadHandling threadHandling)
+// Suppress FP. The call to threadHandling.ThrowIfNotOnUIThread() inside the Lazy<> causes the
+// incorrect identify this constructor as needing to do a threading check.
+#pragma warning disable VSTHRD010 // Invoke single-threaded types on Main thread
```
duncanp-sonar, line 56: The code in this class was a bit more complicated, so it was simpler to just mark this as a false positive (which in this case it really is a false positive).

src/IssueViz.Security.UnitTests/Taint/TaintList/TaintIssuesControlViewModelTests.cs
```diff
@@ -882,7 +876,8 @@ public void OnIssueVizPropertyChanged_SuppressedProperty_FromSuppressedToNotSupp
             IShowInBrowserService showInBrowserService = null,
             IIssueSelectionService selectionService = null,
             IMenuCommandService menuCommandService = null,
-            ISonarQubeService sonarQubeService = null)
+            ISonarQubeService sonarQubeService = null,
+            IThreadHandling threadHandling = null)
```
duncanp-sonar, line 880: Miscellaneous: add `threadHandling` to the test so we don't use the real class by default.

src/IssueViz.Security/Hotspots/HotspotsList/ViewModels/HotspotsControlViewModel.cs
```diff
@@ -111,10 +109,7 @@ public async System.Threading.Tasks.Task UpdateHotspotsListAsync()
         /// Allow the observable collection <see cref="Hotspots"/> to be modified from non-UI thread. 
         /// </summary>
         private void AllowMultiThreadedAccessToHotspotsList()
-        {
-            threadHandling.ThrowIfNotOnUIThread();
-            BindingOperations.EnableCollectionSynchronization(Hotspots, Lock);
-        }
+            => threadHandling.RunOnUIThread(() => BindingOperations.EnableCollectionSynchronization(Hotspots, Lock));
```
duncanp-sonar, line 112: As discussed offline, we decided to change the code so that it doesn't matter what thread it is called on; this class will do the right thing.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -245,8 +246,9 @@ private bool NotSuppressedIssuesInCurrentDocumentFilter(object viewModel)
         /// </summary>
         private void AllowMultiThreadedAccessToIssuesCollection()
         {
-            threadHandling.ThrowIfNotOnUIThread();
-            BindingOperations.EnableCollectionSynchronization(unfilteredIssues, Lock);
+            Debug.Assert(unfilteredIssues != null, "unfiltered issues must be set before calling AllowMultiThreadedAccessToIssuesCollection");
```
duncanp-sonar, line 249: Added the assert as a reminder that the collection needs to be non-null.

src/TestInfrastructure/Framework/ConfigurableVsProjectSystemHelper.cs
```diff
@@ -31,7 +31,6 @@ internal class ConfigurableVsProjectSystemHelper : IProjectSystemHelper
     {
         private readonly IServiceProvider serviceProvider;
         private bool isSolutionFullyOpened;
-        private bool isLegacyProjectSystem = true; // assume is legacy by default
```
duncanp-sonar, line 34: Fix real code smell: variable was not used.

ugras-ergun-sonarsource (approved):

PR #4940, opened 2023-10-18
Added get Solution Name to IsolutionInfoProvider

ugras-ergun-sonarsource:


src/Core/SharedFolderProvider.cs
```diff
@@ -30,7 +30,7 @@ public interface ISharedFolderProvider
         /// Returns the file path to the current Shared Folder
         /// Or null if there's not a Shared Folder
         /// </summary>
-        string GetSharedFolderPathOrNull();
+        string GetSharedFolderPath();
```
ugras-ergun-sonarsource, line 33: Added this change to this PR because it was already a too small PR and it didn't make sense to create another PR just for this..

src/Core/ISolutionInfoProvider.cs
```diff
@@ -77,5 +77,16 @@ public interface ISolutionInfoProvider
         /// </summary>
         /// <remarks>to be used only by existing non-thread-aware code that cannot easily be refactored.</remarks>
         bool IsFolderWorkspace();
+
+        /// <summary>
+        /// Returns the Solution Name, or null if there is not a current solution
+        /// </summary>
+        Task<string> GetSolutionNameAsync();
```
georgii-borovinskikh-sonarsource, line 84: UnintrusiveBindingPathProvider.GetConnectionFilePath can be changed to use this method as well

georgii-borovinskikh-sonarsource (approved):

Issue #4944, opened 2023-10-19
Add call to credentials dialog on auto binding with shared config but no credentials 


georgii-borovinskikh-sonarsource:
Currently, auto binding only works when the user has credentials stored. 

We should prompt them with the credentials dialog and not default to manual bind.

PR #4946, opened 2023-10-20
Add call to credentials dialog on auto binding with shared config but no credentials

georgii-borovinskikh-sonarsource:
Implements #4944

### DIFF
--- a/src/Integration/Connection/ConnectionController.cs
+++ b/src/Integration/Connection/ConnectionController.cs
@@ -65,7 +65,7 @@ namespace SonarLint.VisualStudio.Integration.Connection
             : base(serviceProvider)
         {
             this.host = host ?? throw new ArgumentNullException(nameof(host));
-            this.autoBindTrigger = workflowExecutor == null ? autoBindTrigger : null; 
+            this.autoBindTrigger = workflowExecutor == null ? autoBindTrigger : null;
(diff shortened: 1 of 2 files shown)


src/Integration/Connection/ConnectionController.cs
```diff
@@ -109,37 +97,43 @@ internal bool IsConnectionInProgress
...
                 connectionInfo = new ConnectionInformation(sharedConfig.ServerUri,
-                    credentials.Username,
-                    credentials.Password.ToSecureString())
+                    credentials?.Username,
+                    credentials?.Password?.ToSecureString())
                 {
                     Organization = new SonarQubeOrganization(sharedConfig.Organization, string.Empty)
                 };
+
+                if (credentials == null)
```
ugras-ergun-sonarsource, line 133: Maybe move this up to prevent creating connectioninfo unnecessarily

ugras-ergun-sonarsource (approved):
Just one change on order of the things.

src/Integration/Connection/ConnectionController.cs
georgii-borovinskikh-sonarsource: It's always used, no? sometimes in line 135, and always in line 145

georgii-borovinskikh-sonarsource: sometimes = conditionally

PR #4952, opened 2023-10-23
Implement SharedBindingConfigFileProvider

### LINKED ISSUE #4943: Implement SharedBindingConfigFileProvider

ugras-ergun-sonarsource:
Fixes #4943

### DIFF
--- a/src/ConnectedMode.UnitTests/Shared/SharedBindingConfigFileProviderTests.cs
+++ b/src/ConnectedMode.UnitTests/Shared/SharedBindingConfigFileProviderTests.cs
@@ -1,0 +1,211 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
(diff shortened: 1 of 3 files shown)


src/ConnectedMode/Shared/SharedBindingConfigFileProvider.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        internal SharedBindingConfigFileProvider(ILogger logger, IFileSystem fileSystem)
+        {
+            this.fileSystem = fileSystem;
+            this.logger = logger;
+        }
+
+        public SharedBindingConfigModel ReadSharedBindingConfigFile(string filePath)
+        {
+            SharedBindingConfigModel result = null;
```
georgii-borovinskikh-sonarsource, line 55: can replace with 2 returns instead, would be cleaner IMO

src/ConnectedMode/Shared/SharedBindingConfigFileProvider.cs
```diff
@@ -0,0 +1,85 @@
...
+            {
+                logger.LogVerbose($"[SharedBindingConfigFileProvider] Unable to read shared sonarlint config file: {ex.Message}");
+            }
+
+            return result;
+        }
+
+        public bool WriteSharedBindingConfigFile(string filePath, SharedBindingConfigModel sharedBindingConfigModel)
+        {
+            bool result = false;
```
georgii-borovinskikh-sonarsource, line 71: can replace with 2 returns instead, would be cleaner IMO

src/ConnectedMode.UnitTests/Shared/SharedBindingConfigFileProviderTests.cs
```diff
@@ -0,0 +1,211 @@
...
+            var file = CreateFile(filePath, configFileContent);
+            var fileSystem = GetFileSystem(file.Object);
+
+            var testSubject = CreateTestSubject(fileSystem.Object);
+
+            var result = testSubject.ReadSharedBindingConfigFile(filePath);
+
+            result.Organization.Should().Be("Some Organisation");
+            result.ProjectKey.Should().Be("projectKey");
+            result.Uri.Should().BeNull();
```
georgii-borovinskikh-sonarsource, line 76: Setting the default Uri for SC would make code that uses this a bit less complicated. Not sure if it should be set by the file provider or by the config provider later.

georgii-borovinskikh-sonarsource (approved):

PR #4977, opened 2023-10-27
Add SharedConfig Checkbox to Migration Wizard

### LINKED ISSUE #4976: Add SharedConfig Checkbox to Migration Wizard
This is only adding the checkbox to UI

ugras-ergun-sonarsource:
fixes #4976

### DIFF
--- a/src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
+++ b/src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
@@ -84,7 +84,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration.Wizard
 
         private bool IsUnderGit
         {
-            get { return !(gitWorkspaceService.GetRepoRoot() == null); }
+            get { return gitWorkspaceService.GetRepoRoot() != null; }
         }
 
         private void OnShowHelp() => browserService.Navigate(DocumentationLinks.MigrateToConnectedModeV7);
--- a/src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
+++ b/src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
@@ -20,8 +20,8 @@
 
 using System;
 using System.ComponentModel.Composition;
-using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Binding;
 
 namespace SonarLint.VisualStudio.ConnectedMode.Migration.Wizard
 {
@@ -67,9 +67,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration.Wizard
 
         public void StartMigrationWizard(BoundSonarQubeProject oldBinding)
         {
-            var showTfvcHelpOp = GetTfVcHelpAction(); 
-
-            var migrationWizardWindow = new MigrationWizardWindow(oldBinding, connectedModeMigration, OnShowHelp, showTfvcHelpOp, logger);
+            var migrationWizardWindow = new MigrationWizardWindow(oldBinding, connectedModeMigration, OnShowHelp, OnShowTfvcHelp, OnShowSharedBinding, IsUnderGit, logger);
 
             var finishedSuccessfully = migrationWizardWindow.ShowModal();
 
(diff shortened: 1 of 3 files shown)


src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
```diff
@@ -84,31 +82,15 @@ public void StartMigrationWizard(BoundSonarQubeProject oldBinding)
             }
         }
 
-        private Action GetTfVcHelpAction()
```
ugras-ergun-sonarsource, line 87: I didn't want to use tfsHelpAction existence to determine if I should show the checkbox.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -53,30 +54,32 @@ public sealed partial class MigrationWizardWindow : DialogWindow, IProgress<Migr
...
-        
-        private void SetTfvcInfoVisibility()
-        {
-            bool shouldShowTfvcText = onShowTfvcHelp != null;
 
-            tfvcInfo.Visibility = shouldShowTfvcText ? Visibility.Visible : Visibility.Hidden;
+        private void SetVisibilities(bool isUnderGit)
+        {
+            tfvcInfo.Visibility = isUnderGit ? Visibility.Hidden : Visibility.Visible;
+            chk_SaveSharedBinding.Visibility = isUnderGit ? Visibility.Visible : Visibility.Hidden;
```
ugras-ergun-sonarsource, line 82: While migrating it should only work with git so it's better to hide when it's not in git

ugras-ergun-sonarsource:
here's an example view of the migration wizard

image

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
georgii-borovinskikh-sonarsource: You also removed logging which could be useful. Was that intentional?

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
```diff
@@ -84,31 +82,15 @@ public void StartMigrationWizard(BoundSonarQubeProject oldBinding)
...
+        private bool IsUnderGit
```
georgii-borovinskikh-sonarsource, line 85: property is only used once

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml
```diff
@@ -119,6 +129,8 @@
                     <LineBreak />
                 </TextBlock>
 
+
+
```
georgii-borovinskikh-sonarsource, line 133: empty lines

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
georgii-borovinskikh-sonarsource, line 82: The `chk_` prefix doesn't look great and adds little value

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml
```diff
@@ -111,6 +111,16 @@
                     </TextBlock>
                 </Border>
 
+                <CheckBox x:Name="chk_SaveSharedBinding">
```
georgii-borovinskikh-sonarsource, line 114: This PR doesn't have any checkbox logic. Will you add that later or?

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
ugras-ergun-sonarsource: yes, It looks cleaner on the method.

georgii-borovinskikh-sonarsource: It is confusing tho. First of all, it's just on ==. Secondly, semanticaly this is a method, not a property. If it's a property I assume that it's part of the data stored by object, but you're using it instead of local variable/method.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml
ugras-ergun-sonarsource: yes

src/ConnectedMode/Migration/Wizard/MigrationWizardController.cs
ugras-ergun-sonarsource: Yes we already log if git is not found. I do not believe these logs add much value so I didn't add those in the new method.

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
ugras-ergun-sonarsource: This is a convention using winforms it relays that it's a control it would be easier when reading the code. But for the rest of the window convention is, not having the "_" (e.g. btnPage2) so I'll remove it here also.

georgii-borovinskikh-sonarsource (approved):

PR #4986, opened 2023-10-30
Add shared config save to the migration

### LINKED ISSUE #4974: Add shared config save to the migration

ugras-ergun-sonarsource:
Fixes #4974

### DIFF
--- a/src/ConnectedMode/Migration/ConnectedModeMigration.cs
+++ b/src/ConnectedMode/Migration/ConnectedModeMigration.cs
@@ -155,9 +155,9 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
             logger.WriteLine(MigrationStrings.Process_Finished);
         }
 
-        private void SaveSharedBinding(BoundSonarQubeProject oldBinding)
+        private void SaveSharedBinding(BoundSonarQubeProject binding)
         {
-            var sharedBindingConfigModel = new SharedBindingConfigModel { Organization = oldBinding.Organization?.Key, ProjectKey = oldBinding.ProjectKey, Uri = oldBinding.ServerUri.ToString() };
+            var sharedBindingConfigModel = new SharedBindingConfigModel { Organization = binding.Organization?.Key, ProjectKey = binding.ProjectKey, Uri = binding.ServerUri.ToString() };
             _ = sharedBindingConfigProvider.SaveSharedBinding(sharedBindingConfigModel);
         }
 
--- a/src/ConnectedMode/Migration/ConnectedModeMigration.cs
+++ b/src/ConnectedMode/Migration/ConnectedModeMigration.cs
@@ -148,17 +148,21 @@ namespace SonarLint.VisualStudio.ConnectedMode.Migration
             if (shareBinding)
             {
                 progress?.Report(new MigrationProgress(0, 1, "Saving Shared Binding Config...", false));
-                SaveSharedBinding(oldBinding);
+                var saveSuccess = SaveSharedBinding(oldBinding);
+                if (saveSuccess == false)
+                {
+                    progress?.Report(new MigrationProgress(0, 1, "... Failed to save Binding Config. Skippig the step", false));
+                }
             }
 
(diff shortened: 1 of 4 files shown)


src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
```diff
@@ -169,6 +169,11 @@ private async Task MigrateAsync()
             }
         }
 
+        private bool shareBinding()
```
georgii-borovinskikh-sonarsource: Method name not UpperCamelCase

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -141,10 +145,22 @@ private async Task MigrateImplAsync(BoundSonarQubeProject oldBinding, IProgress<
...
+            {
+                progress?.Report(new MigrationProgress(0, 1, "Saving Shared Binding Config...", false));
+                SaveSharedBinding(oldBinding);
+            }
+
             progress?.Report(new MigrationProgress(0, 1, "Migration finished successfully!", false));
             logger.WriteLine(MigrationStrings.Process_Finished);
         }
 
+        private void SaveSharedBinding(BoundSonarQubeProject oldBinding)
```
georgii-borovinskikh-sonarsource, line 158: why is it oldBinding? it's still the up-to-date data?

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -141,10 +145,22 @@ private async Task MigrateImplAsync(BoundSonarQubeProject oldBinding, IProgress<
...
+        {
+            var sharedBindingConfigModel = new SharedBindingConfigModel { Organization = oldBinding.Organization?.Key, ProjectKey = oldBinding.ProjectKey, Uri = oldBinding.ServerUri.ToString() };
```
georgii-borovinskikh-sonarsource, line 160: NOTE: I don't like how we designed this SharedBindingModel.Uri thing, but it's too late now

src/ConnectedMode/Migration/ConnectedModeMigration.cs
```diff
@@ -141,10 +145,22 @@ private async Task MigrateImplAsync(BoundSonarQubeProject oldBinding, IProgress<
...
+            _ = sharedBindingConfigProvider.SaveSharedBinding(sharedBindingConfigModel);
```
georgii-borovinskikh-sonarsource, line 161: Why ignore the result? Pretty sure Migration Window has progress output, could be one of the steps in that

georgii-borovinskikh-sonarsource (approved):

src/ConnectedMode/Migration/ConnectedModeMigration.cs
ugras-ergun-sonarsource: Extract Method used this because of the name of the original let me fix :D

src/ConnectedMode/Migration/ConnectedModeMigration.cs
ugras-ergun-sonarsource: I thought about that but the rest of the code do not use this for error reporting and I don't think it would make sense. In case of an IO Error...etc we already log the issue so logging here would be redundant

src/ConnectedMode/Migration/Wizard/MigrationWizardWindow.xaml.cs
ugras-ergun-sonarsource: already fixed

src/ConnectedMode/Migration/ConnectedModeMigration.cs
georgii-borovinskikh-sonarsource: How would the user know they had an issue?

ugras-ergun-sonarsource: Same as any other step: logs.

PR #5012, opened 2023-11-03
EnsureHtmlIsXml column cleaner regexp mistakenly matches too much

### LINKED ISSUE #4605: EnsureHtmlIsXml column cleaner regexp mistakenly matches too much
This was discovered while doing #4497 

One of the regexps is defined as 
```
private static readonly Regex CleanCol = new Regex("(?<element>(<col\\s*)|(col\\s+[^/^>]*))>",
            RegexOptions.Compiled,
            Core.RegexConstants.DefaultTimeout);
```

the 2nd optional parameter does not have the opening bracket. This leads to overmatching; `Blablabla(col <someothertag>)` (<- braces denote match group) gets transformed to `Blablablacol <someothertag/>`

This happened to C# rule description 4433

`<p>Lightweight Directory Access Protocol (LDAP) servers provide two main authentication methods: the <em>SASL</em>`

^ the regexp matches from the end of Proto_**col**_ until the first `<em>`

`<p>Lightweight Directory Access Protocol (LDAP) servers provide two main authentication methods: the <em/>SASL</em>`

The transformed string is not a valid html and is not transformable to xaml

ugras-ergun-sonarsource:
Fixes #4605

### DIFF
--- a/src/Rules/HtmlXmlCompatibilityHelper.cs
+++ b/src/Rules/HtmlXmlCompatibilityHelper.cs
@@ -30,11 +30,11 @@ namespace SonarLint.VisualStudio.Rules
         // the empty elements and replace them with elements with closing tags
         // e.g. <br>  =>  <br/>
         // e.g. <col span="123">  =>  <col span="123"/>
-        private static readonly Regex CleanCol = new Regex("(?<element>(<col\\s*)|(col\\s+[^/^>]*))>",
+        private static readonly Regex CleanCol = new Regex("(?<element>(<\\s*col\\s*)|(<\\s*col\\s+[^/^>]*))>",
             RegexOptions.Compiled,
             Core.RegexConstants.DefaultTimeout);
 
-        private static readonly Regex CleanBr = new Regex("(?<element>(<br\\s*)|(br\\s+[^/^>]*))>",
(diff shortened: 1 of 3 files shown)


src/Rules/HtmlXmlCompatibilityHelper.cs
```diff
@@ -30,11 +30,11 @@ internal static class HtmlXmlCompatibilityHelper
         // the empty elements and replace them with elements with closing tags
         // e.g. <br>  =>  <br/>
         // e.g. <col span="123">  =>  <col span="123"/>
-        private static readonly Regex CleanCol = new Regex("(?<element>(<col\\s*)|(col\\s+[^/^>]*))>",
+        private static readonly Regex CleanCol = new Regex("(?<element>(<\\s*col\\s*)|(<\\s*col\\s+[^/^>]*))>",
```
georgii-borovinskikh-sonarsource, line 33: Is the `\\s*` needed here?

georgii-borovinskikh-sonarsource (approved):

src/Rules/HtmlXmlCompatibilityHelper.cs
ugras-ergun-sonarsource: it would work in case of something like < col>. I would say it would work for now without it but I can't think of any scenario that this would be harmful.

PR #5019, opened 2023-11-15
Fix the Threading Issue in GitEventsMonitor

### LINKED ISSUE #5020: Fix the Threading Issue in GitEventsMonitor
### Description 

When disposing of GitEventsMonitor and a head change triggered at the same time It would cause a race condition on underlying FileSystemWatcher and causes VS to freeze. 

### Reproduction

If you have a solution which checking out a earlier version triggers reloading of the solution also (this happens if the solution file changed between versions) will cause Visual Studio to freeze and killing the "devenv.exe" would be the only way to move forward.

ugras-ergun-sonarsource:
Fixes #5020

### DIFF
--- a/src/Core.UnitTests/GitEventsMonitorTests.cs
+++ b/src/Core.UnitTests/GitEventsMonitorTests.cs
@@ -47,7 +47,7 @@ namespace SonarLint.VisualStudio.Core.UnitTests
 
             fileSystemWatcher.Raise(w => w.Renamed += null, null, null);
 
-            Task.Delay(1000).Wait();
+            Task.Delay(10).Wait();
 
             counter.Should().Be(1);
         }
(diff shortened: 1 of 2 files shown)


src/Core.UnitTests/GitEventsMonitorTests.cs
```diff
@@ -46,6 +47,8 @@ public void HeadChanged_EventInvoked()
 
             fileSystemWatcher.Raise(w => w.Renamed += null, null, null);
 
+            Task.Delay(10).Wait();
```
ugras-ergun-sonarsource, line 50: A short wait is needed since we are now firing and forgetting.

pavel-mikula-sonarsource (approved):
LGTM
Please try those two other ways of solving the wait in UTs. I can't test it myself.

src/Core.UnitTests/GitEventsMonitorTests.cs
ugras-ergun-sonarsource: yes they work I pushed the changes

Issue #5040, opened 2023-11-23
Add direct mapping to SonarQubeIssue from IAnalysisIssueVisualization


georgii-borovinskikh-sonarsource:
Needs refactoring of multiple classes:

- [x] SuppressedIssueMatcher should be converted to IssueMatcher, that would return the first matched server issue 
- [x] Move synchronization-related files to a separate namespace 
- [ ] IAnalysisIssueVisualization should have a property of type SonarQubeIssue
- [ ] ClientSuppressionSynchronizer should set the ServerIssue property of IAnalysisIssueVisualization
- [ ] Start fetching all issues instead of only suppressed issues

PR #5045, opened 2023-11-23
Move & rename the files from Suppression to Synchronization folder

### LINKED ISSUE #5044: Move synchronization-related files to a separate namespace
Since we move away from suppression-only model to supporting unsuppressed server issues, we should move the files to SonarLint.VisualStudio.ConnectedMode.Synchronization

georgii-borovinskikh-sonarsource:
Preparation step before replacing IsSuppressed with ServerIssue in IAnalysisIssueVisualization

Fixes #5044 
Part of #5040

### DIFF
--- a/src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
+++ b/src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
@@ -24,7 +24,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Threading;
-using SonarLint.VisualStudio.ConnectedMode.Suppressions;
+using SonarLint.VisualStudio.ConnectedMode.Synchronization;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Core.ETW;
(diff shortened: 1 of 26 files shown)


src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -122,13 +122,12 @@ public async Task UpdateFileStorageAsync()
                     Debug.Assert(fullSolutionFilePath != null, "Not expecting the solution name to be null in Connected Mode");
                     var solnNameWithoutExtension = System.IO.Path.GetFileNameWithoutExtension(fullSolutionFilePath);
 
-                    var allSuppressedIssues = serverIssuesStore.Get();
+                    var allSuppressedIssues = serverIssuesStore.Get().Where(x => x.IsResolved);
```
ugras-ergun-sonarsource, line 125: here's a logic change I think this change requires some tests.

ugras-ergun-sonarsource (commented):
Some testing might be needed

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
georgii-borovinskikh-sonarsource: It's covered by the existing tests. I just moved this LINQ call from one place to the other.

georgii-borovinskikh-sonarsource:
We decided to go with a different approach

PR #5046, opened 2023-11-24
Add popup with choice of Accept or Won't Fix 

### LINKED ISSUE #5032: Add popup with choice of Accept or Won't Fix
The popup should be supplied with server issue id (done by matching against the fetched server issue) and will set the status of that issue. After successfully muting the issue on the server, it should also trigger `suppressionIssueStoreUpdater.UpdateSuppressedIssuesAsync`

ugras-ergun-sonarsource:
Fixes #5032


ugras-ergun-sonarsource:
Functional logic will be added in another ticket.

src/Integration/Muting/MuteWindowDialog.xaml
```diff
@@ -0,0 +1,60 @@
+﻿<vsui:DialogWindow  x:Class="SonarLint.VisualStudio.Integration.Muting.MuteWindowDialog"
```
georgii-borovinskikh-sonarsource, line 1: A screenshot would be nice to attach

src/Integration/Muting/MuteWindowDialog.xaml
```diff
@@ -0,0 +1,60 @@
...
+        <Border BorderBrush="Black" Name="BorderWontFix" BorderThickness="1" HorizontalAlignment="Left" Height="75" Margin="25,25,0,0" VerticalAlignment="Top" Width="450" Visibility="{Binding WontFixVisibility}">
+            <Grid>
+                <RadioButton Name="rbWontFix" GroupName="Transitions" Content="Won't Fix" VerticalAlignment="Top" Margin="10,16,-10,0" FontWeight="Bold" Grid.ColumnSpan="2" Checked="RadioButton_Checked"/>
+                <Label Content="The issue is valid but does not need fixing. It represents accepted technical debt." HorizontalAlignment="Left" Margin="10,36,0,0" VerticalAlignment="Top" Grid.ColumnSpan="2"/>
+            </Grid>
+        </Border>
+
+        <Border BorderBrush="Black" Name="BorderAccept" BorderThickness="1" HorizontalAlignment="Left" Height="75" Margin="25,25,0,0" VerticalAlignment="Top" Width="450" Visibility="{Binding AcceptVisibility}">
+            <Grid>
+                <RadioButton Name="rbAccept" GroupName="Transitions" Content="Accept" VerticalAlignment="Top" Margin="10,16,-10,0" FontWeight="Bold" Grid.ColumnSpan="2" Checked="RadioButton_Checked"/>
```
georgii-borovinskikh-sonarsource, line 20: Changing the text of the button, instead of creating hidden buttons is less confusing and more maintainable.

src/Integration/Muting/MuteWindowDialog.xaml
```diff
@@ -0,0 +1,60 @@
...
+                <Label Content="The issue is valid but does not need fixing. It represents accepted technical debt." HorizontalAlignment="Left" Margin="10,36,0,0" VerticalAlignment="Top" Grid.ColumnSpan="2"/>
```
georgii-borovinskikh-sonarsource, line 21: The content should be in the resource file, IMO

src/Integration/Muting/MuteWindowDialog.xaml
```diff
@@ -0,0 +1,60 @@
...
+        </Border>
+
+        <Border BorderBrush="Black" BorderThickness="1" HorizontalAlignment="Left" Height="75" Margin="25
+                ,115,0,0" VerticalAlignment="Top" Width="450">
+            <Grid>
+                <RadioButton x:Name="rbFalsePositive" GroupName="Transitions" Content="False Positive" VerticalAlignment="Top" Margin="10,16,-10,0" FontWeight="Bold" Grid.ColumnSpan="2" Checked="RadioButton_Checked"/>
+                <Label Content="The issue is raised unxpectedly on code that should not trigger an issue." HorizontalAlignment="Left" Margin="10,36,0,0" VerticalAlignment="Top" Grid.ColumnSpan="2"/>
+            </Grid>
+        </Border>
+        <Label Content="Add a comment (optional):" HorizontalAlignment="Left" Margin="25,200,0,0" VerticalAlignment="Top"/>
```
georgii-borovinskikh-sonarsource, line 32: SLI window had some formatting info in below the text box. We should at least add a ticket for that feature.

src/Integration/Muting/MuteWindowDialog.xaml.cs
```diff
@@ -0,0 +1,82 @@
...
+        public MuteWindowDialog(bool showAccept)
+        {
+            InitializeComponent();
+
+            NewMethod(showAccept);
+        }
+
+        private void NewMethod(bool showAccept)
+        {
+            BorderWontFix.Visibility = showAccept ? Visibility.Hidden : Visibility.Visible;
```
georgii-borovinskikh-sonarsource, line 43: See my point about hidden buttons above

src/Integration/Muting/MuteWindowDialog.xaml.cs
```diff
@@ -0,0 +1,82 @@
...
+    public partial class MuteWindowDialog : DialogWindow
+    {
+        public MuteWindowDialog(bool showAccept)
+        {
+            InitializeComponent();
+
+            NewMethod(showAccept);
+        }
+
+        private void NewMethod(bool showAccept)
```
georgii-borovinskikh-sonarsource, line 41: NewMethod

georgii-borovinskikh-sonarsource (changes requested):

src/Integration/Muting/MuteWindowDialog.xaml
ugras-ergun-sonarsource: I disagree, I believe this way it's more obvious what's happening. We create a logical separation of different transitions. Using same control for two different condition would be more confusing IMHO.

src/Integration/Transition/MuteWindowDialog.xaml.cs
```diff
@@ -0,0 +1,81 @@
...
+using Microsoft.VisualStudio.PlatformUI;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.Integration.Transition
+{
+    /// <summary>
+    /// Interaction logic for UserControl1.xaml
+    /// </summary>
+    [ContentProperty(nameof(MuteWindowDialog))]
+    [ExcludeFromCodeCoverage]
```
ugras-ergun-sonarsource, line 33: UI Code not testable

ugras-ergun-sonarsource:
image

src/Integration/Resources/Strings.resx
```diff
@@ -574,4 +574,34 @@ This is the general format to use when writing errors to output window or when t
...
+    <value>False Positive</value>
+  </data>
+  <data name="MuteWindow_SubmitButton" xml:space="preserve">
+    <value>Mark Issue as...</value>
+  </data>
+  <data name="MuteWindow_Title" xml:space="preserve">
+    <value>Mark Issue as Resolved on SonarQube</value>
+  </data>
+  <data name="MuteWindow_WontFixContent" xml:space="preserve">
+    <value>The issue is valid but does not need fixing. It represents accepted technical debt.</value>
```
georgii-borovinskikh-sonarsource, line 602: Same content as Accept 🤔

src/Integration/Transition/MuteWindowDialog.xaml
```diff
@@ -0,0 +1,62 @@
...
+        <Border BorderBrush="Black" Name="BorderWontFix" BorderThickness="1" HorizontalAlignment="Left" Height="75" Margin="25,25,0,0" VerticalAlignment="Top" Width="450" Visibility="{Binding WontFixVisibility}">
+            <Grid>
+                <RadioButton Name="rbWontFix" GroupName="Transitions" Content="{x:Static resx:Strings.MuteWindow_WontFixTitle}" VerticalAlignment="Top" Margin="10,16,-10,0" FontWeight="Bold" Grid.ColumnSpan="2" Checked="RadioButton_Checked"/>
+                <Label Content="{x:Static resx:Strings.MuteWindow_WontFixContent}" HorizontalAlignment="Left" Margin="10,36,0,0" VerticalAlignment="Top" Grid.ColumnSpan="2"/>
+            </Grid>
+        </Border>
+
+        <Border BorderBrush="Black" Name="BorderAccept" BorderThickness="1" HorizontalAlignment="Left" Height="75" Margin="25,25,0,0" VerticalAlignment="Top" Width="450" Visibility="{Binding AcceptVisibility}">
+            <Grid>
+                <RadioButton Name="rbAccept" GroupName="Transitions" Content="{x:Static resx:Strings.MuteWindow_AcceptTitle}" VerticalAlignment="Top" Margin="10,16,-10,0" FontWeight="Bold" Grid.ColumnSpan="2" Checked="RadioButton_Checked"/>
```
georgii-borovinskikh-sonarsource, line 22: Radio button stye (and possibly others) is duplicated.

src/Integration/Transition/MuteWindowDialog.xaml
```diff
@@ -0,0 +1,62 @@
...
+                <Label Content="{x:Static resx:Strings.MuteWindow_AcceptContent}" HorizontalAlignment="Left" Margin="10,36,0,0" VerticalAlignment="Top" Grid.ColumnSpan="2"/>
+            </Grid>
+        </Border>
+
+        <Border BorderBrush="Black" BorderThickness="1" HorizontalAlignment="Left" Height="75" Margin="25
+                ,115,0,0" VerticalAlignment="Top" Width="450">
+            <Grid>
+                <RadioButton x:Name="rbFalsePositive" GroupName="Transitions" Content="{x:Static resx:Strings.MuteWindow_FalsePositiveTitle}" VerticalAlignment="Top" Margin="10,16,-10,0" FontWeight="Bold" Grid.ColumnSpan="2" Checked="RadioButton_Checked"/>
+                <Label Content="{x:Static resx:Strings.MuteWindow_FalsePositiveContent}" HorizontalAlignment="Left" Margin="10,36,0,0" VerticalAlignment="Top" Grid.ColumnSpan="2"/>
```
georgii-borovinskikh-sonarsource, line 31: Style duplication

src/Integration/Transition/MuteWindowDialog.xaml.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public SonarQubeIssueTransition SelectedIssueTransition { get; private set; }
+
+        public string Comment => txtComment.Text;
+
+        private void RadioButton_Checked(object sender, RoutedEventArgs e)
+        {
+            switch (sender)
+            {
+                case var value when value == rbWontFix:
```
georgii-borovinskikh-sonarsource, line 67: Non-blocker: Could be replaced with a dictionary that would look nicer

src/Integration/Transition/MuteWindowDialog.xaml
```diff
@@ -0,0 +1,62 @@
...
+            </Grid>
+        </Border>
+        <Label Content="{x:Static resx:Strings.MuteWindow_CommentLabel}" HorizontalAlignment="Left" Margin="25,200,0,0" VerticalAlignment="Top"/>
+        <TextBox Name="txtComment" HorizontalAlignment="Center" Height="60" Margin="0,225,0,0" TextWrapping="Wrap" VerticalAlignment="Top" Width="450"/>
+        <Button Name="Cancel" Content="{x:Static resx:Strings.MuteWindow_CancelButton}" Padding="5,1" HorizontalAlignment="Left" Margin="300,300,0,0" VerticalAlignment="Top" Click="Cancel_Click"/>
+        <Button Name="Submit" Content="{x:Static resx:Strings.MuteWindow_SubmitButton}" Padding="5,1" HorizontalAlignment="Left" Margin="360,300,0,0" VerticalAlignment="Top" Click="Submit_Click">
```
georgii-borovinskikh-sonarsource, line 37: This style is not duplicated, however, it's hard to read

georgii-borovinskikh-sonarsource: Ah, actually it is duplicated from the Cancel button

georgii-borovinskikh-sonarsource (changes requested):

src/Integration/Transition/MuteWindowDialog.xaml
```diff
@@ -9,39 +9,64 @@
...
+                <Label Content="{x:Static resx:Strings.MuteWindow_FalsePositiveContent}" Style="{StaticResource TransitionLabelStyle}" />
             </Grid>
         </Border>
-        <Label Content="{x:Static resx:Strings.MuteWindow_CommentLabel}" HorizontalAlignment="Left" Margin="25,200,0,0" VerticalAlignment="Top"/>
+        <Label Content="{x:Static resx:Strings.MuteWindow_CommentLabel}" Style="{StaticResource LabelStyle}" Margin="25,200,0,0"/>
         <TextBox Name="txtComment" HorizontalAlignment="Center" Height="60" Margin="0,225,0,0" TextWrapping="Wrap" VerticalAlignment="Top" Width="450"/>
-        <Button Name="Cancel" Content="{x:Static resx:Strings.MuteWindow_CancelButton}" Padding="5,1" HorizontalAlignment="Left" Margin="300,300,0,0" VerticalAlignment="Top" Click="Cancel_Click"/>
-        <Button Name="Submit" Content="{x:Static resx:Strings.MuteWindow_SubmitButton}" Padding="5,1" HorizontalAlignment="Left" Margin="360,300,0,0" VerticalAlignment="Top" Click="Submit_Click">
+        <Button Name="Cancel" Content="{x:Static resx:Strings.MuteWindow_CancelButton}" Style="{StaticResource ButtonStyle}" Margin="300,300,0,0" Click="Cancel_Click"/>
```
georgii-borovinskikh-sonarsource, line 63: Why are these margins not part of the style definition? Btw you can also inherit styles and override certain properties in style definitions

georgii-borovinskikh-sonarsource (approved):

src/Integration/Resources/Strings.resx
ugras-ergun-sonarsource: I know I'll talk about it with Marco during the meet

src/Integration/Transition/MuteWindowDialog.xaml
ugras-ergun-sonarsource: margins are different in each component

georgii-borovinskikh-sonarsource: You can define style as (I don't remember wpf syntax, so I will explain this with C# objects :-) )
```
style AcceptButton : GenericButton
{
override Margin Margin = "1, 2, 3, 4"
}
```

ugras-ergun-sonarsource: Putting a property in a common place for only one component does not make sense.

georgii-borovinskikh-sonarsource: It does. You separate styling from layout. Easier to read, easier to maintain.

ugras-ergun-sonarsource: Adding a property and overriding it later create more noise IMO. Another solution would be creating a common style and child styles for each specific component based on the common one. Which would be an overkill in this situation just makes reading code harder.

georgii-borovinskikh-sonarsource: Yeah but now you have styling split between the style resources and the layout elements, which is IMO more confusing and harder to read

ugras-ergun-sonarsource: Discussing this takes more time than changing and I do not feel strongly about it so done.

src/Integration/Transition/MuteWindowDialog.xaml
```diff
@@ -0,0 +1,86 @@
...
+            </Grid>
+        </Border>
+
+        <Border Style="{StaticResource BorderStyle}" Margin="25,115,0,0">
+            <Grid>
+                <RadioButton x:Name="rbFalsePositive" Content="{x:Static resx:Strings.MuteWindow_FalsePositiveTitle}" Style="{StaticResource RadioButtonStyle}" Checked="RadioButton_Checked"/>
+                <Label Content="{x:Static resx:Strings.MuteWindow_FalsePositiveContent}" Style="{StaticResource TransitionLabelStyle}" />
+            </Grid>
+        </Border>
+        <Label Content="{x:Static resx:Strings.MuteWindow_CommentLabel}" Style="{StaticResource LabelStyle}" Margin="25,200,0,0"/>
```
georgii-borovinskikh-sonarsource, line 81: LabelStyle is only used once, margin can be moved there

src/Integration/Transition/MuteWindowDialog.xaml
```diff
@@ -0,0 +1,86 @@
...
+        <TextBox Name="txtComment" HorizontalAlignment="Center" Height="60" Margin="0,225,0,0" TextWrapping="Wrap" VerticalAlignment="Top" Width="450"/>
```
georgii-borovinskikh-sonarsource, line 82: Styles in the layout

src/Integration/Transition/MuteWindowDialog.xaml
```diff
@@ -0,0 +1,86 @@
...
+    </vsui:DialogWindow.Resources>
+    <Grid Height="350" Width="500">
+        <Border Name="BorderWontFix" Style="{StaticResource BorderStyle}" Margin="25,25,0,0" Visibility="{Binding WontFixVisibility}">
+            <Grid>
+                <RadioButton Name="rbWontFix" Content="{x:Static resx:Strings.MuteWindow_WontFixTitle}" Style="{StaticResource RadioButtonStyle}" Checked="RadioButton_Checked"/>
+                <Label Content="{x:Static resx:Strings.MuteWindow_WontFixContent}" Style="{StaticResource TransitionLabelStyle}" />
+            </Grid>
+        </Border>
+
+        <Border Name="BorderAccept" Style="{StaticResource BorderStyle}" Margin="25,25,0,0" Visibility="{Binding AcceptVisibility}">
```
georgii-borovinskikh-sonarsource, line 68: This margin value is used twice. There's only one BorderStyle instance with a different Mragin value. You can set this one as the default, and override it once at line 75

PR #5060, opened 2023-11-28
Add ProjectRootCalculator, cleanup PathHelper

### LINKED ISSUE #5058: Add IProjectRootCalculator
This class should use the supplied local file location to search the server for the files with the same name, find the longest tail-match and calculate project root based on that

georgii-borovinskikh-sonarsource:
Fixes #5058

### DIFF
--- a/src/Core/Helpers/PathHelper.cs
+++ b/src/Core/Helpers/PathHelper.cs
@@ -153,8 +153,7 @@ namespace SonarLint.VisualStudio.Core.Helpers
             }
 
             return localFilePath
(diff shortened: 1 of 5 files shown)


src/Core/Helpers/PathHelper.cs
```diff
@@ -53,83 +53,6 @@ public static string EscapeFileName(string unescapedName)
             return escapedName.ToString();
         }
 
-        /// <summary>
```
georgii-borovinskikh-sonarsource, line 56: Removed unused methods

ugras-ergun-sonarsource (approved):

PR #5067, opened 2023-11-29
Add IssueMatcher

### LINKED ISSUE #5066: Add IIssueMatcher
Should use existing matching criteria from ISuppressedIssueMatcher and find the matching server issue from the list returned by GetIssues method

georgii-borovinskikh-sonarsource:
Fixes #5066

### DIFF
--- a/src/ConnectedMode.UnitTests/IssueMatcherTests.cs
+++ b/src/ConnectedMode.UnitTests/IssueMatcherTests.cs
@@ -1,0 +1,162 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using SonarLint.VisualStudio.Core.Suppressions;
+using SonarLint.VisualStudio.TestInfrastructure;
+using SonarQube.Client.Models;
+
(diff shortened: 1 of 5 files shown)


src/ConnectedMode/IssueMatcher.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Linq;
+using SonarLint.VisualStudio.Core.Helpers;
+using SonarLint.VisualStudio.Core.Suppressions;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.ConnectedMode
+{
+    internal interface IIssueMatcher
+    {
+        bool IsGoodMatch(IFilterableIssue issue, SonarQubeIssue serverIssue);
```
ugras-ergun-sonarsource, line 33: A name which describes better should be used IMO. When I read the name of the method I do not understand what is a "good match" I need to look into the method to understand it. Something like "IsMatchWithFilePath" would be better.

src/ConnectedMode.UnitTests/IssueMatcherTests.cs
ugras-ergun-sonarsource, line 1: There are no tests with multiple issues to match.

ugras-ergun-sonarsource (changes requested):
Missing test case and also a small comment about the method name.

src/ConnectedMode.UnitTests/IssueMatcherTests.cs
georgii-borovinskikh-sonarsource: WDYM? Isn't FindMatchOrDefault_FindsFirstMatch that?

src/ConnectedMode/IssueMatcher.cs
georgii-borovinskikh-sonarsource: My idea was to indicate that it's not a 100% certain match (ie it can have both FPs and FNs). Maybe you have other suggestions considering my initial intention?

ugras-ergun-sonarsource: then IsProbableMatch maybe? Good is too vague for me.

ugras-ergun-sonarsource: Or at least add some documentation comments

src/ConnectedMode/IssueMatcher.cs
```diff
@@ -0,0 +1,103 @@
...
+using SonarLint.VisualStudio.Core.Suppressions;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.ConnectedMode
+{
+    internal interface IIssueMatcher
+    {
+        /// <summary>
+        /// This method attempts to match <paramref name="issue"/> with <paramref name="serverIssue"/>.
+        /// There's a possibility of False Positive matches: in case false tail-match of file paths, as the project root is not taken into account,
```
ugras-ergun-sonarsource, line 35: Remark could be better place for this sentence IMO

ugras-ergun-sonarsource (approved):

PR #5085, opened 2023-12-06
Add MuteIssueCommand logic for non-roslyn case

### LINKED ISSUE #5083: Implement MuteIssueCommand logic for non-roslyn case

georgii-borovinskikh-sonarsource:
Fixes #5083


src/Integration.Vsix.UnitTests/Analysis/MuteIssueCommandTests.cs
```diff
@@ -0,0 +1,277 @@
...
+        oleStatus.Should().Be(InvisibleAndDisabled);
+        testSubject.Visible.Should().BeFalse();
+        testSubject.Enabled.Should().BeFalse();
+    }
+
+    [TestMethod] 
+    public void QueryStatus_NotSupportedIssue_Invisible()
+    {
+        var testSubject = CreateTestSubject(out var errorListHelperMock, out _, out _, out var activeSolutionBoundTrackerMock, out _, out _);
+        SonarCompositeRuleId.TryParse("csharpsquid:S333", out SonarCompositeRuleId ruleId);
```
ugras-ergun-sonarsource, line 82: using csharp for not supported rule does not make sense. We plan to support it at the end.

src/Integration.Vsix.UnitTests/Analysis/MuteIssueCommandTests.cs
```diff
@@ -0,0 +1,277 @@
...
+        var issue = Mock.Of<IFilterableIssue>();
+        errorListHelperMock.Setup(x => x.TryGetIssueFromSelectedRow(out issue)).Returns(false);
+
+        testSubject.Invoke();
+        
+        errorListHelperMock.Verify(x => x.TryGetIssueFromSelectedRow(out issue), Times.Once);
+    }
+    
+    [TestMethod]
+    public void Execute_DoesNotFindServerIssue_TODO()
```
ugras-ergun-sonarsource, line 152: Do we need this test? I don't believe it should be responsibility of the command.

georgii-borovinskikh-sonarsource: Well, the command is the entity that has the knowledge that the issue was not found. As to what it's supposed to do with that info, there's a todo in the test to add that check

src/Integration.Vsix.UnitTests/Analysis/MuteIssueCommandTests.cs
georgii-borovinskikh-sonarsource: It's not supported as of this commit, so I just used it for now. When we support C# this test will need to be modified, but I think it's fine. If c# is implemented properly, it will fail and remind us to modify it.

PR #5098, opened 2023-12-08
Make radio button selected when clicked inside border

### LINKED ISSUE #5093: Make radio button selected when clicked inside border

ugras-ergun-sonarsource:
Fixes #5093

### DIFF
--- a/src/Integration/Transition/MuteWindowDialog.xaml
+++ b/src/Integration/Transition/MuteWindowDialog.xaml
@@ -23,6 +23,8 @@
             <Setter Property="Height" Value="75" />
             <Setter Property="VerticalAlignment" Value="Top" />
             <Setter Property="Width" Value="450" />
+            <Setter Property="Background" Value="Transparent"/>
+            <EventSetter Event="MouseLeftButtonDown" Handler="Border_MouseDown" />
         </Style>
         <Style x:Key="LabelStyle" TargetType="Label">
             <Setter Property="HorizontalAlignment" Value="Left" />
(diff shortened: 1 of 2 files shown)


src/Integration/Transition/MuteWindowDialog.xaml
```diff
@@ -23,6 +23,8 @@
             <Setter Property="Height" Value="75" />
             <Setter Property="VerticalAlignment" Value="Top" />
             <Setter Property="Width" Value="450" />
+            <Setter Property="Background" Value="Transparent"/>
```
ugras-ergun-sonarsource, line 26: This is needed so if we click an empty place in Border it would be recognized.

src/Integration/Transition/MuteWindowDialog.xaml
```diff
@@ -87,7 +89,7 @@
             </Grid>
         </Border>
 
-        <Border Style="{StaticResource BottomBorderStyle}">
+        <Border Name="BorderFalsePositive" Style="{StaticResource BottomBorderStyle}">
```
georgii-borovinskikh-sonarsource, line 92: Is this name needed?

src/Integration/Transition/MuteWindowDialog.xaml.cs
```diff
@@ -44,6 +46,7 @@ public MuteWindowDialog(bool showAccept)
             SetVisibility(showAccept);
 
             Transitions = InitializeTransitions();
+            Borders = InitializeBorders();
```
georgii-borovinskikh-sonarsource, line 49: IDK if borders is a good name. It doesn't tell what this is used for

georgii-borovinskikh-sonarsource (approved):

src/Integration/Transition/MuteWindowDialog.xaml
ugras-ergun-sonarsource: Yes, to be able to recognize which border was clicked in the code behind. The alternative would be creating different event for each border.

src/Integration/Transition/MuteWindowDialog.xaml.cs
ugras-ergun-sonarsource: Maybe BorderRadioButtons?

src/Integration/Transition/MuteWindowDialog.xaml
georgii-borovinskikh-sonarsource: This was really confusing because the diff only had one Name change

ugras-ergun-sonarsource: Other borders had already names setup due to being used for visibility.

PR #5101, opened 2023-12-08
Make SonarCloud use won't fix

### LINKED ISSUE #5100: Make SonarCloud use won't fix

ugras-ergun-sonarsource:
Fixes #5100


src/Core/Configuration/ConnectedModeFeaturesConfiguration.cs
```diff
@@ -88,7 +88,7 @@ public bool IsAcceptTransitionAvailable()
         {
             var serverInfo = sonarQubeService.GetServerInfo();
```
georgii-borovinskikh-sonarsource, line 89: This is not enough. You need to change the request configuration as well

georgii-borovinskikh-sonarsource (changes requested):

PR #5135, opened 2023-12-26
Add SLCoreServiceProvider, modify ISLCoreJsonRpc

### LINKED ISSUE #5130: Add ISLCoreServiceFactory

georgii-borovinskikh-sonarsource:
Fixes #5130


src/SLCore/Core/ISLCoreServiceProvider.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <returns>True if the underlying connection is alive, False if the connection is unavailable at the moment</returns>
+        bool TryGetTransientService<TService>(out TService service) where TService : ISLCoreService;
+    }
+
+    public interface ISLCoreServiceProviderWriter 
+    {
+        /// <summary>
+        /// Resets the state with a new <see cref="ISLCoreJsonRpc"/> instance and clears the cache
+        /// </summary>
+        void Reset(ISLCoreJsonRpc newRpcInstance);
```
ugras-ergun-sonarsource, line 43: Since we use this for also first time initialisation maybe we should choose a name such as "Initialize". Reset implies an existing instance every time.

ugras-ergun-sonarsource (approved):

src/SLCore/Core/ISLCoreServiceProvider.cs
georgii-borovinskikh-sonarsource: Eh. I see your point, but then Initialize implies to be only called once. `SetCurrentConnection`?

ugras-ergun-sonarsource: Initialize also includes resetting, it means setting it up to starting  state, but SetCurrentConnection works better more explicit.

PR #5139, opened 2023-12-27
Add IProgressListener

### LINKED ISSUE #5134: Add IProgressListener

ugras-ergun-sonarsource:
Fixes #5134


src/SLCore.UnitTests/Core/SLCoreListenerSetUpTests.cs
```diff
@@ -21,7 +21,7 @@
 using SonarLint.VisualStudio.SLCore.Core;
 using SonarLint.VisualStudio.TestInfrastructure;
 
-namespace SonarLint.VisualStudio.SLCore.UnitTests
+namespace SonarLint.VisualStudio.SLCore.UnitTests.Core
```
ugras-ergun-sonarsource, line 24: moved tests to core folder so it would have same structure with the main project. It's 2 tasks in 1 PR but since PR is so small I think it's OK.

src/SLCore.UnitTests/Core/SLCoreServiceProviderTests.cs
```diff
@@ -22,7 +22,7 @@
 using SonarLint.VisualStudio.SLCore.Core;
 using SonarLint.VisualStudio.TestInfrastructure;
 
-namespace SonarLint.VisualStudio.SLCore.UnitTests;
+namespace SonarLint.VisualStudio.SLCore.UnitTests.Core;
```
ugras-ergun-sonarsource, line 25: see comment on SLCoreListenerSetUpTests. (This also had some whitespace fixes done automatically)

src/SLCore/Listener/IProgressListener.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+using SonarLint.VisualStudio.SLCore.Core;
+using StreamJsonRpc;
+
+namespace SonarLint.VisualStudio.SLCore.Listener
+{
+    public interface IProgressListener : ISLCoreListener
+    {
+        [JsonRpcMethod(UseSingleObjectParameterDeserialization = true)]
```
georgii-borovinskikh-sonarsource, line 30: 1. I don't know if this is going to work when you put it on the interface instead of putting it on the class. 2. We can set this property in JsonRpcTargetOptions. It didn't work for me in the PoC, but I have a suspicion I did something wrong

src/SLCore/Listener/IProgressListener.cs
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+using System.ComponentModel.Composition;
+using System.Threading.Tasks;
+using SonarLint.VisualStudio.SLCore.Core;
+using StreamJsonRpc;
+
+namespace SonarLint.VisualStudio.SLCore.Listener
+{
+    public interface IProgressListener : ISLCoreListener
```
georgii-borovinskikh-sonarsource, line 28: We don't need this interface.

src/SLCore/Listener/IProgressListener.cs
georgii-borovinskikh-sonarsource: I will try to make the PoC work. Will get back to you.

src/SLCore/Listener/IProgressListener.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        [JsonRpcMethod(UseSingleObjectParameterDeserialization = true)]
+        Task ReportProgress(object parameters);
+    }
+
+    [Export(typeof(IProgressListener))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class ProgressListener : IProgressListener
+    {
+        public Task StartProgress(object parameters)
```
georgii-borovinskikh-sonarsource, line 41: You need to add a comment with why this is an object and not some real model ("we don't support this method")

src/SLCore/Listener/IProgressListener.cs
```diff
@@ -0,0 +1,51 @@
...
+    public interface IProgressListener : ISLCoreListener
+    {
+        [JsonRpcMethod(UseSingleObjectParameterDeserialization = true)]
+        Task StartProgress(object parameters);
+
+        [JsonRpcMethod(UseSingleObjectParameterDeserialization = true)]
+        Task ReportProgress(object parameters);
+    }
+
+    [Export(typeof(IProgressListener))]
```
georgii-borovinskikh-sonarsource, line 37: Wrong export contract

src/SLCore.UnitTests/Listener/ProgressListenerTests.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace SonarLint.VisualStudio.SLCore.UnitTests.Listener
+{
+    [TestClass]
+    public class ProgressListenerTests
+    {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<ProgressListener, IProgressListener>();
```
georgii-borovinskikh-sonarsource, line 33: Wrong export contract

georgii-borovinskikh-sonarsource (changes requested):
Give me some time to make the parameter deserialization work without annotations. Meanwhile you can remove the interface as it is not needed and change the export contract.

src/SLCore/Listener/IProgressListener.cs
georgii-borovinskikh-sonarsource: Discussed on slack

georgii-borovinskikh-sonarsource (approved):

PR #5182, opened 2024-01-18
Fix muting of incorrect issues when bound to SonarCloud

### LINKED ISSUE #5181: Fix muting of incorrect issues when bound to SonarCloud
On version 7.5 and release candidate 7.6.0.82886 when a solution is bound to SC, the mark as resolved button incorrectly mutes a different issue with the same rule id in a different file.

The root cause is that when we match the issues, we expect them to be from the single file. However, because the request to SC was malformed, it actually returned issues from different files. The issue is that the same filter key is named differently on different servers: `components` for SQ and `componentKeys` for SC, and we only support the 1st one.

georgii-borovinskikh-sonarsource:
Fixes #5181

### DIFF
--- a/src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestTests.cs
+++ b/src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestTests.cs
@@ -448,7 +448,7 @@ namespace SonarQube.Client.Tests.Requests.Api.V7_20
         }
 
         [TestMethod]
-        public async Task InvokeAsync_ComponentKeySpecified_ComponentsAreIncludedInQueryString()
+        public async Task InvokeAsync_ComponentKeySpecified_ComponentsAreNotIncludedInQueryString()
         {
             var testSubject = CreateTestSubject("any", "any", componentKey: "project1");
 
@@ -462,7 +462,7 @@ namespace SonarQube.Client.Tests.Requests.Api.V7_20
             _ = await testSubject.InvokeAsync(httpClient, CancellationToken.None);
 
             var actualQueryString = GetSingleActualQueryString(handlerMock);
-            actualQueryString.Contains("components=project1").Should().BeTrue();
+            actualQueryString.Contains("components=project1").Should().BeFalse();
         }
 
         private static GetIssuesRequest CreateTestSubject(string projectKey, string statusesToRequest, string branch = null, string[] issueKeys = null, string ruleId = null, string componentKey = null)
@@ -475,7 +475,6 @@ namespace SonarQube.Client.Tests.Requests.Api.V7_20
                 Branch = branch,
                 IssueKeys = issueKeys,
                 RuleId = ruleId,
-                ComponentKey = componentKey
             };
 
             return testSubject;
--- a/src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestWrapperTests.cs
+++ b/src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestWrapperTests.cs
@@ -25,6 +25,7 @@ using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
+using SonarQube.Client.Api;
 using SonarQube.Client.Api.V7_20;
 using SonarQube.Client.Tests.Infra;
 using static SonarQube.Client.Tests.Infra.MocksHelper;
@@ -34,10 +35,15 @@ namespace SonarQube.Client.Tests.Requests.Api.V7_20
     [TestClass]
     public class GetIssuesRequestWrapperTests
     {
-        [TestMethod]
-        public async Task InvokeAsync_NoIssueKeys_ExpectedPropertiesArePassedInMultipleRequests()
+        private const string SonarQubeComponentPropertyName = "components";
+        private const string SonarCloudComponentPropertyName = "componentKeys";
+        
+        [DataTestMethod]
+        [DataRow(SonarQubeComponentPropertyName, DisplayName = "SonarQube")]
+        [DataRow(SonarCloudComponentPropertyName, DisplayName = "SonarCloud")]
+        public async Task InvokeAsync_SonarQube_NoIssueKeys_ExpectedPropertiesArePassedInMultipleRequests(string componentPropertyName)
         {
-            var testSubject = CreateTestSubject("aaaProject", "xStatus", "yBranch", null, "rule1", "project1");
+            var testSubject = CreateTestSubject(componentPropertyName, "aaaProject", "xStatus", "yBranch", null, "rule1", "project1");
 
             var handlerMock = new Mock<HttpMessageHandler>(MockBehavior.Strict);
             var httpClient = new HttpClient(handlerMock.Object)
@@ -51,15 +57,18 @@ namespace SonarQube.Client.Tests.Requests.Api.V7_20
 
             // The wrapper is expected to make three calls, for code smells, bugs, then vulnerabilities
             handlerMock.Invocations.Count.Should().Be(3);
-            CheckExpectedQueryStringsParameters(handlerMock, 0, "aaaProject", "xStatus", "yBranch", "CODE_SMELL", "rule1", "project1");
-            CheckExpectedQueryStringsParameters(handlerMock, 1, "aaaProject", "xStatus", "yBranch", "BUG", "rule1", "project1");
-            CheckExpectedQueryStringsParameters(handlerMock, 2, "aaaProject", "xStatus", "yBranch", "VULNERABILITY", "rule1", "project1");
+            CheckExpectedQueryStringsParameters(componentPropertyName, handlerMock, 0, "aaaProject", "xStatus", "yBranch", "rule1", "project1", expectedTypes: "CODE_SMELL");
+            CheckExpectedQueryStringsParameters(componentPropertyName, handlerMock, 1, "aaaProject", "xStatus", "yBranch", "rule1", "project1", expectedTypes: "BUG");
+            CheckExpectedQueryStringsParameters(componentPropertyName, handlerMock, 2, "aaaProject", "xStatus", "yBranch", "rule1", "project1", expectedTypes: "VULNERABILITY");
         }
 
-        [TestMethod]
-        public async Task InvokeAsync_HasIssueKeys_ExpectedPropertiesArePassedInASingleRequest()
+        [DataTestMethod]
+        [DataRow(SonarQubeComponentPropertyName, DisplayName = "SonarQube")]
+        [DataRow(SonarCloudComponentPropertyName, DisplayName = "SonarCloud")]
+        public async Task InvokeAsync_HasIssueKeys_ExpectedPropertiesArePassedInASingleRequest(string componentPropertyName)
         {
-            var testSubject = CreateTestSubject("aaaProject", "xStatus", "yBranch", new[] { "issue1", "issue2" }, "rule1", "project1");
+            var issueKeys = new[] { "issue1", "issue2" };
+            var testSubject = CreateTestSubject(componentPropertyName,"aaaProject", "xStatus", "yBranch", issueKeys, "rule1", "project1");
 
             var handlerMock = new Mock<HttpMessageHandler>(MockBehavior.Strict);
             var httpClient = new HttpClient(handlerMock.Object)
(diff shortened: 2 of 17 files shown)


src/SonarQube.Client/Requests/RequestFactory.cs
```diff
@@ -61,7 +61,7 @@ public RequestFactory(ILogger logger)
             return RegisterRequest<TRequest, TRequestImpl>(version, () => new TRequestImpl());
         }
 
-        private RequestFactory RegisterRequest<TRequest, TRequestImpl>(string version, Func<TRequestImpl> factory)
+        public RequestFactory RegisterRequest<TRequest, TRequestImpl>(string version, Func<TRequestImpl> factory)
```
pavel-mikula-sonarsource, line 64: Was this intentional? It's only used inside the file

src/SonarQube.Client/Api/DefaultConfiguration.cs
```diff
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using SonarQube.Client.Api.V7_20;
```
pavel-mikula-sonarsource, line 21: I don't think this is desired in this file. As it's meant to work with various ranges of versions.
```suggestion
```

src/SonarQube.Client/Api/V7_20/GetIssuesWithComponentRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace SonarQube.Client.Api.V7_20
+{
+    internal class GetIssuesWithComponentRequest : GetIssuesRequest
+    {
+        // this class should not be used on it's own
+        private protected GetIssuesWithComponentRequest(){}
+        
+        public virtual string ComponentKey { get; set; }
+    }
+
```
pavel-mikula-sonarsource, line 32: You have some useless whitespace here. I'm not sure if you usually cleanup those or not.
`Trailing Whitespace Visualizer` extensions helps to spot those
```suggestion

        public virtual string ComponentKey { get; set; }
    }

```

src/SonarQube.Client/Api/V7_20/GetIssuesWithComponentRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+using Newtonsoft.Json;
+
+namespace SonarQube.Client.Api.V7_20
+{
+    internal class GetIssuesWithComponentRequest : GetIssuesRequest
+    {
+        // this class should not be used on it's own
+        private protected GetIssuesWithComponentRequest(){}
```
pavel-mikula-sonarsource, line 28: If I autoformat this file, it does a small update (and also remove the whitespace from the other comment)
```suggestion
        private protected GetIssuesWithComponentRequest() { }
```

pavel-mikula-sonarsource, line 28: I think abstract class with abstract property would be slightly better solution here. You'd get rid of the artificial constructor.

pavel-mikula-sonarsource, line 28: Is there a strong reason to have `private protected` instead of `protected` here? The class is already `internal`, so the `private` doesn't really do anything

src/SonarQube.Client.Tests/Requests/DefaultConfiguration_Configure_Tests.cs
```diff
@@ -20,6 +20,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Linq;
```
pavel-mikula-sonarsource, line 23: This doesn't seem intentional
```suggestion
```

src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestTests.cs
```diff
@@ -462,7 +462,7 @@ public async Task InvokeAsync_ComponentKeySpecified_ComponentsAreIncludedInQuery
             _ = await testSubject.InvokeAsync(httpClient, CancellationToken.None);
 
             var actualQueryString = GetSingleActualQueryString(handlerMock);
-            actualQueryString.Contains("components=project1").Should().BeTrue();
+            actualQueryString.Contains("components=project1").Should().BeFalse();
```
pavel-mikula-sonarsource, line 465: I would reduce the risk for miss-asserting like this (even removing the "s")
```suggestion
            actualQueryString.Contains("component").Should().BeFalse();
```

pavel-mikula-sonarsource (changes requested):
Some minor nitpicking.
I'll take a closer look at UTs in parallel

src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesSonarCloudRequestTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        var httpClient = new HttpClient(handlerMock.Object)
+        {
+            BaseAddress = new Uri(ValidBaseAddress)
+        };
+
+        SetupHttpRequest(handlerMock, EmptyGetIssuesResponse);
+        _ = await testSubject.InvokeAsync(httpClient, CancellationToken.None);
+
+        var actualQueryString = GetSingleActualQueryString(handlerMock);
+        actualQueryString.Contains("components").Should().BeFalse();
```
pavel-mikula-sonarsource, line 52: This is miss-asserting, because the real value would be `componentKeys`
```suggestion
        actualQueryString.Contains("component").Should().BeFalse();
```

src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesSonarQubeRequestTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        var httpClient = new HttpClient(handlerMock.Object)
+        {
+            BaseAddress = new Uri(ValidBaseAddress)
+        };
+
+        SetupHttpRequest(handlerMock, EmptyGetIssuesResponse);
+        _ = await testSubject.InvokeAsync(httpClient, CancellationToken.None);
+
+        var actualQueryString = GetSingleActualQueryString(handlerMock);
+        actualQueryString.Contains("components").Should().BeFalse();
```
pavel-mikula-sonarsource, line 52: The assertion is correct. I would remove "s" to play a little more safe
```suggestion
        actualQueryString.Contains("component").Should().BeFalse();
```

pavel-mikula-sonarsource, line 52: Better assertion style - this should be applied everywhere in this PR
```suggestion
        actualQueryString.Should().Contain("components=project1");
```

src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestWrapperTests.cs
```diff
@@ -34,10 +35,15 @@ namespace SonarQube.Client.Tests.Requests.Api.V7_20
     [TestClass]
     public class GetIssuesRequestWrapperTests
     {
-        [TestMethod]
-        public async Task InvokeAsync_NoIssueKeys_ExpectedPropertiesArePassedInMultipleRequests()
+        private const string SonarQubeComponentPropertyName = "components";
+        private const string SonarCloudComponentPropertyName = "componentKeys";
```
pavel-mikula-sonarsource, line 39: Personal preference: I would name them like this. The common prefix makes it more table-layouted and easier to read (easier to read what's the common part)
```suggestion
        private const string ComponentPropertyNameSonarQube = "components";
        private const string ComponentPropertyNameSonarCloud = "componentKeys";
```

src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestWrapperTests.cs
```diff
@@ -34,10 +35,15 @@ namespace SonarQube.Client.Tests.Requests.Api.V7_20
...
+        
+        [DataTestMethod]
+        [DataRow(SonarQubeComponentPropertyName, DisplayName = "SonarQube")]
+        [DataRow(SonarCloudComponentPropertyName, DisplayName = "SonarCloud")]
+        public async Task InvokeAsync_SonarQube_NoIssueKeys_ExpectedPropertiesArePassedInMultipleRequests(string componentPropertyName)
```
pavel-mikula-sonarsource, line 44: This UT is for SC and SQ, so SQ likely shouldn't be in the name
```suggestion
        public async Task InvokeAsync_NoIssueKeys_ExpectedPropertiesArePassedInMultipleRequests(string componentPropertyName)
```

src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestWrapperTests.cs
```diff
@@ -51,15 +57,18 @@ public async Task InvokeAsync_NoIssueKeys_ExpectedPropertiesArePassedInMultipleR
 
             // The wrapper is expected to make three calls, for code smells, bugs, then vulnerabilities
             handlerMock.Invocations.Count.Should().Be(3);
-            CheckExpectedQueryStringsParameters(handlerMock, 0, "aaaProject", "xStatus", "yBranch", "CODE_SMELL", "rule1", "project1");
-            CheckExpectedQueryStringsParameters(handlerMock, 1, "aaaProject", "xStatus", "yBranch", "BUG", "rule1", "project1");
-            CheckExpectedQueryStringsParameters(handlerMock, 2, "aaaProject", "xStatus", "yBranch", "VULNERABILITY", "rule1", "project1");
+            CheckExpectedQueryStringsParameters(componentPropertyName, handlerMock, 0, "aaaProject", "xStatus", "yBranch", "rule1", "project1", expectedTypes: "CODE_SMELL");
+            CheckExpectedQueryStringsParameters(componentPropertyName, handlerMock, 1, "aaaProject", "xStatus", "yBranch", "rule1", "project1", expectedTypes: "BUG");
+            CheckExpectedQueryStringsParameters(componentPropertyName, handlerMock, 2, "aaaProject", "xStatus", "yBranch", "rule1", "project1", expectedTypes: "VULNERABILITY");
```
pavel-mikula-sonarsource, line 62: This table-like layout is so much better than the old version

src/SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestWrapperTests.cs
```diff
@@ -74,44 +83,76 @@ public async Task InvokeAsync_HasIssueKeys_ExpectedPropertiesArePassedInASingleR
...
-        private static void CheckExpectedQueryStringsParameters(Mock<HttpMessageHandler> handlerMock, int invocationIndex,
-            string expectedProject, string expectedStatues, string expectedBranch, string expectedTypes, string expectedRule, string expectedComponent)
+        private static void CheckExpectedQueryStringsParameters(string componentKeyName,
+            Mock<HttpMessageHandler> handlerMock,
+            int invocationIndex,
+            string expectedProject,
+            string expectedStatues,
+            string expectedBranch,
+            string expectedRule,
+            string expectedComponent,
```
pavel-mikula-sonarsource, line 124: For sake of simplicity, do you need to bother with these parameters? I think values for most of them can be just hardcoded in the assertion. As the test is small, and individual UTs can be aligned on "&projects=project1"

pavel-mikula-sonarsource (changes requested):
Round for UTs

src/SonarQube.Client/Requests/RequestFactory.cs
georgii-borovinskikh-sonarsource: Good catch! I was experimenting with using factory instead of new() constraint for generics and forgot to remove it

src/SonarQube.Client/Api/DefaultConfiguration.cs
georgii-borovinskikh-sonarsource: This is due to the fact that I didn't add `V7_20.` prefix to `GetIssuesWithComponentSonarQubeRequest`. Compare it with `V7_20.GetIssuesRequestWrapper`, for example. I can fix it, if you think it would make it clearer

src/SonarQube.Client/Api/V7_20/GetIssuesWithComponentRequest.cs
georgii-borovinskikh-sonarsource: I think one of the tools (don't remember which one 😁 ) suggested me to change from abstract to concrete class with protected constructor. It seemed like a good idea to me, but maybe the tool was too afraid of abstract classes

georgii-borovinskikh-sonarsource: Not really, since as you said it doesn't do anything. Ideally I would love to make it explicit that this class shouldn't be inherited from without a good reason, but this doesn't really achieve that

src/SonarQube.Client.Tests/Requests/DefaultConfiguration_Configure_Tests.cs
georgii-borovinskikh-sonarsource: Yep, It was added during one of the iterations of writing the tests and is not needed anymore

src/SonarQube.Client/Api/V7_20/GetIssuesWithComponentRequest.cs
pavel-mikula-sonarsource: It is our rule. It raises when there's no abstract member on the class. The sole purpose of this one is to override the property, so having the property abstract to enforce the action would not make it raise again

src/SonarQube.Client/Api/V7_20/GetIssuesWithComponentRequest.cs
georgii-borovinskikh-sonarsource: Usually we ignore trailing spaces, but I will fix it since I'm touching the file anyways

src/SonarQube.Client/Api/DefaultConfiguration.cs
pavel-mikula-sonarsource: I'd say that every class usage in this file should have it's version. So files from a globally imported specific version V7_20 do not leak to unexpected lines.

pavel-mikula-sonarsource (approved):
LGTM

Issue #5179, opened 2024-01-16
Add Repox Artifactory authentication credentials to the CI Pipeline


georgii-borovinskikh-sonarsource:
Could be done similarly to this https://github.com/SonarSource/sonar-dotnet/pull/8514/files

PR #5185, opened 2024-01-19
Update pipeline to fix Repox authentication for CycloneDX tool install

georgii-borovinskikh-sonarsource:
Part of #5179


pavel-mikula-sonarsource (approved):
LGTM

PR #5197, opened 2024-01-26
Add GetEffectiveRuleDetailsResponse & EffectiveRuleDetailsDto

### LINKED ISSUE #5193: Create model for the rule descriptions
#5162 has a stub model instead of actual model implementation

georgii-borovinskikh-sonarsource:
Fixes #5193

### DIFF
--- a/src/SLCore.UnitTests/SLCore.UnitTests.csproj
+++ b/src/SLCore.UnitTests/SLCore.UnitTests.csproj
@@ -13,4 +13,8 @@
     <ProjectReference Include="..\TestInfrastructure\TestInfrastructure.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Update="FluentAssertions" Version="6.12.0" />
+  </ItemGroup>
+
 </Project>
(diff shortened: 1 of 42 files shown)


src/SLCore.UnitTests/SLCore.UnitTests.csproj
```diff
@@ -13,4 +13,8 @@
     <ProjectReference Include="..\TestInfrastructure\TestInfrastructure.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Update="FluentAssertions" Version="6.12.0" />
```
georgii-borovinskikh-sonarsource, line 17: I updated it for only one project, but maybe I should have done it for https://github.com/SonarSource/sonarlint-visualstudio/blob/86d1c450a50aa31bb3390f9a124390f01fa73337/src/SonarLint.Test.props

ugras-ergun-sonarsource: Yes, you should do it there instead. We don't want to maintain in many places. Also why did you need to change version?

georgii-borovinskikh-sonarsource: I needed it for a feature that was not available in the old one

ugras-ergun-sonarsource (approved):
Please move the new version of Fluent Assertions to SonarLint.Test.props

PR #5200, opened 2024-01-29
Update Repox nuget URL and protocol version

georgii-borovinskikh-sonarsource:


nuget.config
```diff
@@ -12,7 +12,7 @@
   <packageSources>
     <clear />
     <!-- SonarSecrets package published by the Language Team -->
-    <add key="repox_releases" value="https://repox.jfrog.io/repox/api/nuget/sonarsource-nuget-releases" />
+    <add key="repox_releases" value="https://repox.jfrog.io/repox/api/nuget/v3/sonarsource-nuget-releases/index.json" protocolVersion="3" />
```
pavel-mikula-sonarsource, line 15: Let's use this one. It's the one recomended by REs. I see the same content there, so it should work the same.

"artifactory" is a name of the technology, while "repox" is an old name of our old provider (that was replaced with jfrog) that is already in the hostname.
```suggestion
    <add key="repox_releases" value="https://repox.jfrog.io/artifactory/api/nuget/v3/sonarsource-nuget-releases/index.json" protocolVersion="3" />
```

pavel-mikula-sonarsource (changes requested):
Let's try a little different URL

pavel-mikula-sonarsource (approved):
LGTM

PR #5225, opened 2024-02-13
Merge classes into ISolutionBindingRepository

### LINKED ISSUE #5210: Merge classes into ISolutionBindingRepository

ugras-ergun-sonarsource:
Fixes #5210


src/Integration.Vsix/AsmRef_Integration.Vsix_Baseline_WithStrongNames.txt
```diff
@@ -1,7 +1,7 @@
 ﻿---
 ################################
 # Assembly references report
-# Report date/time: 2024-02-12T15:03:48.2745900Z
```
ugras-ergun-sonarsource, line 4: for some reason git created a conflict and didn't fixed itself even after merge from master. I had to manually resolve conflict and this happened.

src/ConnectedMode.UnitTests/Binding/BindingInfoProviderTests.cs
```diff
@@ -32,6 +32,19 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
     [TestClass]
     public class BindingInfoProviderTests
     {
+        [TestMethod]
+        public void MefCtor_CheckIsExported()
+        {
+            MefTestHelpers.CheckTypeCanBeImported<BoundConnectionInfoProvider, IBoundConnectionInfoProvider>(
+                MefTestHelpers.CreateExport<ISolutionBindingRepository>());
```
pavel-mikula-sonarsource, line 39: Out of curiosity: What's the line length that you try to follow?

I saw that S103 is not active on SL. Should activate it?

src/ConnectedMode/Persistence/ISolutionBindingRepository.cs
```diff
@@ -70,5 +86,45 @@ public BoundSonarQubeProject Read(string configFilePath)
 
             return bound;
         }
+
+        public bool Write(string configFilePath, BoundSonarQubeProject binding)
+        {
+            if (binding == null)
```
pavel-mikula-sonarsource, line 92: `is null` and `is not null` are semantically better way of checking for nullability.
```suggestion
            if (binding is null)
```

It reads more fluently. And it never calls `.Equals` for records, improving the performance a tiny bit.

src/ConnectedMode.UnitTests/Binding/BindingInfoProviderTests.cs
ugras-ergun-sonarsource: We don't have a rule for that. We decide what is more readable case by case. And I don't think it should be rule-based.

src/ConnectedMode/Binding/BoundConnectionInfoProvider.cs
```diff
@@ -28,24 +28,22 @@
...
         [ImportingConstructor]
-        public BoundConnectionInfoProvider(IUnintrusiveBindingPathProvider unintrusiveBindingPathProvider, ISolutionBindingFileLoader solutionBindingFileLoader)
-            : this(unintrusiveBindingPathProvider, solutionBindingFileLoader, ThreadHandling.Instance)
+        public BoundConnectionInfoProvider(ISolutionBindingRepository solutionBindingRepository)
+            : this(solutionBindingRepository, ThreadHandling.Instance)
         {
         }
 
-        internal BoundConnectionInfoProvider(IUnintrusiveBindingPathProvider unintrusiveBindingPathProvider, ISolutionBindingFileLoader solutionBindingFileLoader, IThreadHandling threadHandling)
+        internal BoundConnectionInfoProvider(ISolutionBindingRepository solutionBindingRepository, IThreadHandling threadHandling)
```
georgii-borovinskikh-sonarsource, line 44: This class could have been a method of the repo

src/ConnectedMode/Binding/BoundConnectionInfoProvider.cs
```diff
@@ -55,15 +53,11 @@ public IEnumerable<BoundConnectionInfo> GetExistingBindings()
...
+            var bindings = solutionBindingRepository.List();
 
             foreach (var binding in bindings)
             {
-                var boundSonarQubeProject = solutionBindingFileLoader.Load(binding);
-
-                if (boundSonarQubeProject == null) { continue; }
-
-                result.Add(ConvertToBindingInfo(boundSonarQubeProject));
+                result.Add(ConvertToBindingInfo(binding));
```
georgii-borovinskikh-sonarsource, line 60: You don't need to have the conversion if you can directly read this class from json

src/ConnectedMode/Persistence/ISolutionBindingRepository.cs
```diff
@@ -70,5 +86,42 @@ public BoundSonarQubeProject Read(string configFilePath)
...
+            }
+
+            credentialsLoader.Save(binding.Credentials, binding.ServerUri);
+
+            return true;
+        }
+
+        public IEnumerable<BoundSonarQubeProject> List()
+        {
+            var result = new List<BoundSonarQubeProject>();
```
georgii-borovinskikh-sonarsource, line 111: there's no need for list if you return IEnumerable. Also, for now we don't need to list the full list of bindings with QPs, we only need the connections. So, to simplify things, I would directly parse the connection info and return just that, unless you see a future need to return the full info.

src/ConnectedMode/Persistence/ISolutionBindingRepository.cs
```diff
@@ -70,5 +86,42 @@ public BoundSonarQubeProject Read(string configFilePath)
...
+
+            var bindingConfigPaths = unintrusiveBindingPathProvider.GetBindingPaths();
+
+            foreach (var bindingConfigPath in bindingConfigPaths)
+            {
+                var boundSonarQubeProject = Read(bindingConfigPath);
```
georgii-borovinskikh-sonarsource, line 117: Reusing the read method is good, but we might not need to load the credentials for each connection. You could create a private read method with a flag to load credentials

georgii-borovinskikh-sonarsource (changes requested):

src/ConnectedMode/Binding/BoundConnectionInfoProvider.cs
ugras-ergun-sonarsource: Repo is about read/write operations and more generic, but this class is about getting the unique connections for sloop. I would argue they are logically different things.

src/ConnectedMode/Persistence/ISolutionBindingRepository.cs
ugras-ergun-sonarsource: I don't see a future use but to keep things consistent using the same object with the rest of the methods makes more sense for me.

src/ConnectedMode/Binding/BoundConnectionInfoProvider.cs
georgii-borovinskikh-sonarsource: Still, this could be an extension method even if you don't want to make it a part of the main interface. This piece of functionality is too small to warrant a separate class

src/ConnectedMode/Persistence/ISolutionBindingRepository.cs
georgii-borovinskikh-sonarsource: I don't think it's worth it to parse unnecessary data just to keep the interface slightly more clean. There are ways to minimize the difference between models and the methods used to parse them (generics, inheritance hierarchy of classes, etc.)

src/ConnectedMode/Binding/BoundConnectionInfoProvider.cs
ugras-ergun-sonarsource: I don't see having a small class as a problem. This class tackles a separate problem, I might merge the functionality here with converting to SLOOP Dto's (not sure yet how it'll be)

src/ConnectedMode/Persistence/ISolutionBindingRepository.cs
ugras-ergun-sonarsource: Using something like Generics... etc, would be an overkill. 99% of the time, this collection will have only one member and  even if there are multiple connections, it won't be hundreds or even tens of connections.

src/ConnectedMode/Binding/BoundConnectionInfoProvider.cs
georgii-borovinskikh-sonarsource: Merging with dtos makes more sense

src/ConnectedMode/Persistence/ISolutionBindingRepository.cs
georgii-borovinskikh-sonarsource: I don't see generics as something that is too complicated to implement. And anyways, I was only using it as an example.

src/ConnectedMode.UnitTests/Binding/BindingInfoProviderTests.cs
pavel-mikula-sonarsource: Are you fine with long lines too? For example, we have S103 configured to 200 on sonar-dotnet, to have some guidance. And we ~~violate~~accept it where it makes sense.

georgii-borovinskikh-sonarsource (approved):

georgii-borovinskikh-sonarsource (approved):

PR #5231, opened 2024-02-15
Add GetConnectionIdFromUri Method to ConnectionIdHelper

### LINKED ISSUE #5230: Add GetConnectionIdFromUri Method to ConnectionIdHelper

ugras-ergun-sonarsource:
Fixes #5230

### DIFF
--- a/src/SLCore.UnitTests/Common/Helpers/ConnectionIdHelperTests.cs
+++ b/src/SLCore.UnitTests/Common/Helpers/ConnectionIdHelperTests.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
 using SonarLint.VisualStudio.SLCore.Common.Helpers;
 
 namespace SonarLint.VisualStudio.SLCore.UnitTests.Common.Helpers;
@@ -48,20 +49,29 @@ public class ConnectionIdHelperTests
     }
 
     [TestMethod]
-    [DataRow(null, null, null)]
-    [DataRow("", "something", null)]
-    [DataRow("not a uri", "something", null)]
-    [DataRow("http://someuri.com", null, "sq|http://someuri.com")]
-    [DataRow("http://someuri.com", "something", "sq|http://someuri.com")]
+    [DataRow("http://someuri.com", null, "sq|http://someuri.com/")]
+    [DataRow("http://someuri.com", "something", "sq|http://someuri.com/")]
     [DataRow("https://sonarcloud.io", "something", "sc|something")]
     [DataRow("https://sonarcloud.io", "", null)]
     [DataRow("https://sonarcloud.io", null, null)]
     public void GetConnectionIdFromUri_PassUri_ReturnsAsExpected(string uriString, string organisation, string expectedConnectionId)
     {
+        var uri = new Uri(uriString);
+
         var testSubject = new ConnectionIdHelper();
 
-        var actualConnectionId = testSubject.GetConnectionIdFromUri(uriString, organisation);
+        var actualConnectionId = testSubject.GetConnectionIdFromUri(uri, organisation);
 
         actualConnectionId.Should().Be(expectedConnectionId);
+    }
+
+    [TestMethod]
+    public void GetConnectionIdFromUri_UriIsNull_ReturnsNull()
+    {
+        var testSubject = new ConnectionIdHelper();
+
+        var actualConnectionId = testSubject.GetConnectionIdFromUri(null, "something");
+
+        actualConnectionId.Should().BeNull();
     }
 }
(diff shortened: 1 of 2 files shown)


src/SLCore/Common/Helpers/ConnectionIdHelper.cs
```diff
@@ -33,6 +35,22 @@ internal class ConnectionIdHelper : IConnectionIdHelper
         private const string SonarQubePrefix = "sq|";
         private static readonly Uri SonarCloudUri = new Uri("https://sonarcloud.io");
 
+        public string GetConnectionIdFromUri(string uriString, string organisation)
```
georgii-borovinskikh-sonarsource, line 38: BoundSonarQubeProject has `Uri` property, why do you use `string` here? The other method of this class also returns Uri and not string

georgii-borovinskikh-sonarsource (changes requested):

src/SLCore/Common/Helpers/ConnectionIdHelper.cs
```diff
@@ -33,6 +35,22 @@ internal class ConnectionIdHelper : IConnectionIdHelper
         private const string SonarQubePrefix = "sq|";
         private static readonly Uri SonarCloudUri = new Uri("https://sonarcloud.io");
 
+        public string GetConnectionIdFromUri(Uri uri, string organisation)
+        {
+            if (uri is not null)
```
georgii-borovinskikh-sonarsource, line 40: invert if to reduce nesting

src/SLCore/Common/Helpers/ConnectionIdHelper.cs
```diff
@@ -33,6 +35,22 @@ internal class ConnectionIdHelper : IConnectionIdHelper
...
+            {
+                if (uri == SonarCloudUri)
+                {
+                    if (!string.IsNullOrWhiteSpace(organisation)) { return SonarCloudPrefix + organisation; }
```
georgii-borovinskikh-sonarsource, line 44: newlines missing in if body

src/SLCore.UnitTests/Common/Helpers/ConnectionIdHelperTests.cs
```diff
@@ -38,12 +39,39 @@ public class ConnectionIdHelperTests
...
+
+    [TestMethod]
+    public void GetConnectionIdFromUri_UriIsNull_ReturnsNull()
+    {
+        var testSubject = new ConnectionIdHelper();
+
+        var actualConnectionId = testSubject.GetConnectionIdFromUri(null, "something");
+
+        actualConnectionId.Should().BeNull();
+    }
```
georgii-borovinskikh-sonarsource, line 76: add test that `uri == GetUriFromConnectionId(GetConnectionIdFromUri(uri))`

georgii-borovinskikh-sonarsource (approved):

georgii-borovinskikh-sonarsource (approved):
a couple of small suggestions, otherwise LGTM

src/SLCore/Common/Helpers/ConnectionIdHelper.cs
ugras-ergun-sonarsource: it will not be that useful because then we'll need to have more return points (since this way it returns null when organisation is missing)

src/SLCore/Common/Helpers/ConnectionIdHelper.cs
ugras-ergun-sonarsource: it's to reduce clutter I think the expression is simple enough to read easily.

georgii-borovinskikh-sonarsource: I think you're treating a symptom here. This method is hard to read, I completely missed the branch where organization is null (is that even a possibility?)

src/SLCore/Common/Helpers/ConnectionIdHelper.cs
```diff
@@ -33,6 +35,28 @@ internal class ConnectionIdHelper : IConnectionIdHelper
...
+            {
+                return null;
+            }
+
+            if (uri == SonarCloudUri)
+            {
+                if (!string.IsNullOrWhiteSpace(organisation))
+                {
+                    return SonarCloudPrefix + organisation;
+                }
```
pavel-mikula-sonarsource, line 50: Maybe for a refactoring change in the next PR: 
This would be easier to understand like this:
```
return string.IsNullOrWhiteSpace(organization)
    ? null
    : SonarCloudPrefix + organization;
```
and then removing the least `return null`

src/SLCore/Common/Helpers/ConnectionIdHelper.cs
```diff
@@ -33,6 +35,28 @@ internal class ConnectionIdHelper : IConnectionIdHelper
         private const string SonarQubePrefix = "sq|";
         private static readonly Uri SonarCloudUri = new Uri("https://sonarcloud.io");
```
pavel-mikula-sonarsource, line 36: C#9 target-typed null can be useful here (it's not useful ultimately everywhere, this is a place where it is)
```
private static readonly Uri SonarCloudUri = new("https://sonarcloud.io");
```

PR #5237, opened 2024-02-20
Convert Binding to SLOOP Dto

### LINKED ISSUE #5229: Convert Binding to SLOOP Dto

ugras-ergun-sonarsource:
Fixes #5229


ugras-ergun-sonarsource:
Most of the files are packages changes they're in different commits. Look at the 2nd commit for the real changes.

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+        internal ServerConnectionConfigurationProvider(ISolutionBindingRepository solutionBindingRepository, IThreadHandling threadHandling, IConnectionIdHelper connectionIdHelper)
+        {
+            this.solutionBindingRepository = solutionBindingRepository;
+            this.threadHandling = threadHandling;
+            this.connectionIdHelper = connectionIdHelper;
+        }
+
+        public IEnumerable<T> GetServerConnectionConfiguration<T>() where T : ServerConnectionConfiguration
+        {
+            if (bindingList == null) { InitBindingList(); }
```
georgii-borovinskikh-sonarsource, line 60: How is this going to handle new connections added?

ugras-ergun-sonarsource: I am going to handle that while doing the handle new binding ticket.

georgii-borovinskikh-sonarsource: But why make this class stateful in the first place then? You can make it always get the most relevant data, and then add caching later if you need to.

ugras-ergun-sonarsource: I was planning to have an add method later and add to list (if the conn does not exist already). I don't want to do IO operation everytime we call this class, I know probably I won't be that much but still.

georgii-borovinskikh-sonarsource: Why add that method here and not create a caching wrapper for IBindingRepository?

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+        private sealed class BoundSonarQubeProjectUriComparer : IEqualityComparer<BoundSonarQubeProject>
+        {
+            public bool Equals(BoundSonarQubeProject x, BoundSonarQubeProject y)
+            {
+                if (ReferenceEquals(x, y))
+                {
+                    return true;
+                }
+
+                if (x == null && y == null) { return true; }
```
pavel-mikula-sonarsource, line 97: `is null` everywhere around

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+
+                if (x == null ^ y == null)
```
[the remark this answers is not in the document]
ugras-ergun-sonarsource: but they are not the same I want to XOR them it should be true when they're both null.

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+                if (ReferenceEquals(x, y))
+                {
+                    return true;
+                }
+
+                if (x is null && y is null)
+                {
+                    return true;
+                }
+                //XOR it's true only when one of them is true
```
georgii-borovinskikh-sonarsource, line 94: why xor when you check that both are true in the condition above?

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+                if (x is null ^ y is null)
+                {
+                    return false;
+                }
+
+                return x.ServerUri == y.ServerUri;
```
georgii-borovinskikh-sonarsource, line 100: I would use Equals instead just to be safe

ugras-ergun-sonarsource: == is already overridden for Uri, so it wouldn't be safer also if the ServerUri of the first Uri were null somehow (It shouldn't be), it would create an exception.

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
georgii-borovinskikh-sonarsource: No, it's not misleading. You don't just forget the condition above when you read code. XOR is so infrequently used that it's actually more confusing to see it.

ugras-ergun-sonarsource: I do not agree but it's a so trivial issue that doesn't deserve any more discussion about it.

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
georgii-borovinskikh-sonarsource: The fact that == is overriden is an additional knowledge that is not apparent at the first glance. It's more natural to see Equals delegation inside Equals. Also, the uri is never null there. There's a check in the constructor

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    bindingList.Add(new SonarCloudConnectionConfigurationDto(connectionID, true, binding.Organization?.Key));
+                }
+                else
+                {
+                    bindingList.Add(new SonarQubeConnectionConfigurationDto(connectionID, true, binding.ServerUri.ToString()));
+                }
+            }
+
+            return bindingList.OfType<T>();
```
georgii-borovinskikh-sonarsource, line 78: This doesn't make much sense after you removed caching. I would return the list of ServerConnectionConfiguration and let the caller filter that instead

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    return false;
+                }
+
+                return x.ServerUri.Equals(y.ServerUri);
+            }
+
+            public int GetHashCode(BoundSonarQubeProject obj)
+            {
+                return obj.ServerUri.GetHashCode();
```
georgii-borovinskikh-sonarsource, line 105: You check for null in Equals, but not here

src/ConnectedMode.UnitTests/Binding/ServerConnectionConfigurationProviderTests.cs
```diff
@@ -0,0 +1,154 @@
...
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.SLCore.Common.Helpers;
+using SonarLint.VisualStudio.SLCore.Service.Connection.Models;
+using SonarQube.Client.Models;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.Binding
+{
+    [TestClass]
+    public class ServerConnectionConfigurationProviderTests
```
georgii-borovinskikh-sonarsource, line 34: Missing MEF tests

georgii-borovinskikh-sonarsource (changes requested):

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
ugras-ergun-sonarsource: caller can still do it if they pass ServerConnectionConfiguration as T.

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
ugras-ergun-sonarsource: We don't need to here it's the expected behaviour.

src/ConnectedMode/Binding/ServerConnectionConfigurationProvider.cs
georgii-borovinskikh-sonarsource: But if they don't, they will do IO twice. There's no need for this generic now.

georgii-borovinskikh-sonarsource (approved):

Issue #5240, opened 2024-02-21
Implement config scope change


georgii-borovinskikh-sonarsource:
- [] add handling of binding update to ActiveConfigScopeTracker
- [] add ActiveConfigScopeTracker integration to ActiveSolutionBoundTracker

PR #5241, opened 2024-02-21
Update config scope when binding changes

georgii-borovinskikh-sonarsource:
Part of #5240

### DIFF
--- a/src/SLCore/State/ActiveConfigScopeTracker.cs
+++ b/src/SLCore/State/ActiveConfigScopeTracker.cs
@@ -59,17 +59,14 @@ internal sealed class ActiveConfigScopeTracker : IActiveConfigScopeTracker
     private readonly IAsyncLock asyncLock;
 
     private readonly ISLCoreServiceProvider serviceProvider;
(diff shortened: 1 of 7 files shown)


src/SLCore/State/ActiveConfigScopeTracker.cs
```diff
@@ -104,10 +104,17 @@ public async Task SetCurrentConfigScopeAsync(string id, string connectionId = nu
 
         using (await asyncLock.AcquireAsync())
         {
-            Debug.Assert(currentConfigScope == null);
-
-            await configurationScopeService.DidAddConfigurationScopesAsync(
-                new DidAddConfigurationScopesParams(new List<ConfigurationScopeDto> { configurationScopeDto }));
+            if (currentConfigScope != null && currentConfigScope.id == id)
```
ugras-ergun-sonarsource, line 107: You can simplify if condition 
```suggestion
            if (currentConfigScope?.id == id)
```

ugras-ergun-sonarsource (approved):
one small suggestion and change we talked offline

PR #5242, opened 2024-02-22
Add config scope integration to ActiveSolutionBoundTracker

georgii-borovinskikh-sonarsource:
Part of #5240


src/Integration.UnitTests/MefServices/ConfigScopeUpdaterTests.cs
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
```
pavel-mikula-sonarsource, line 24: You desperately need global usings :)
https://github.com/SonarSource/sonar-dotnet/blob/72e14ee23ad494c38c1f6d80c3155991ee21f6eb/analyzers/tests/SonarAnalyzer.Test/SonarAnalyzer.Test.csproj#L76C3-L81C46

What can be added globally, add it to your targets for tests. What can be added only locally to individual CSPRoj, add it only there (maybe Moq is not everywhere, I don't know)

pavel-mikula-sonarsource: ..not in this PR

src/Integration/MefServices/ConfigScopeUpdater.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    private async Task HandleConfigScopeUpdateInternalAsync(string solutionName, string connectionId, string projectKey)
+    {
+        if (solutionName is null)
+        {
+            await activeConfigScopeTracker.RemoveCurrentConfigScopeAsync();
+            return;
+        }
+
+        await activeConfigScopeTracker.SetCurrentConfigScopeAsync(solutionName, connectionId, projectKey);
```
pavel-mikula-sonarsource, line 76: I have a strong preference for `if ... else ...` in this case

src/SLCore.UnitTests/Common/Helpers/ConnectionIdHelperTests.cs
```diff
@@ -20,12 +20,26 @@
...
 namespace SonarLint.VisualStudio.SLCore.UnitTests.Common.Helpers;
 
 [TestClass]
 public class ConnectionIdHelperTests
 {
+    [TestMethod]
+    public void MefCtor_CheckIsExported()
+    {
+        MefTestHelpers.CheckTypeCanBeImported<ConnectionIdHelper, IConnectionIdHelper>();
+    }
```
pavel-mikula-sonarsource, line 34: What's your opinion on arrow methods and their general adoption in the codebase?

georgii-borovinskikh-sonarsource: I am pro-arrow methods in general, but in the case of tests I am against it

src/Integration.UnitTests/MefServices/ConfigScopeUpdaterTests.cs
ugras-ergun-sonarsource: I think we have them in newer libraries....

ugras-ergun-sonarsource (approved):

PR #5259, opened 2024-02-29
Add integration test infrastructure

georgii-borovinskikh-sonarsource:
Part of #5265


src/SLCore.IntegrationTests/SLCore.IntegrationTests.csproj
```diff
@@ -13,4 +13,11 @@
     <ProjectReference Include="..\TestInfrastructure\TestInfrastructure.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Content Include="..\EmbeddedSonarAnalyzer.props">
+      <Link>EmbeddedSonarAnalyzer.props</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
georgii-borovinskikh-sonarsource, line 19: Not sure this is the correct option. WDYT?

src/SLCore.IntegrationTests/PluginInformationLoader.cs
```diff
@@ -0,0 +1,66 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.IO;
+using System.Linq;
+using System.Xml;
+
+namespace SonarLint.VisualStudio.SLCore.IntegrationTests;
+
+public static class PluginInformationLoader // this might be reused in the product code in the future
```
ugras-ergun-sonarsource, line 27: Then why not put it in Product code?

georgii-borovinskikh-sonarsource: Not sure where

georgii-borovinskikh-sonarsource: Also, same could be said about the process runner (partially)

src/SLCore.IntegrationTests/SLCore.IntegrationTests.csproj
ugras-ergun-sonarsource: I prefer to have it always. It's a tiny file, anyway. I don't think copying all the time is much more expensive than checking if the file is newer.

src/SLCore.IntegrationTests/SLCoreTestProcessRunner.cs
```diff
@@ -0,0 +1,132 @@
...
+        var token = cancellationTokenSource.Token;
+        var fileLoggingEnabled = stdErrLogFilePath != null;
+        var prefix = fileLoggingEnabled ? string.Empty : "ERR: ";
+        errorFileStream = fileLoggingEnabled ? new StreamWriter(File.OpenWrite(stdErrLogFilePath)){AutoFlush = enableAutoFlush} : null;
+
+
+        while (!token.IsCancellationRequested)
+        {
+            var line = process.StandardError.ReadLine();
+            if (string.IsNullOrEmpty(line)) // potential problem here if it returns too often?
```
ugras-ergun-sonarsource, line 107: Since this is test code I don't see as a big problem, we can focus if it ends up to be a problem.

ugras-ergun-sonarsource (approved):

PR #5281, opened 2024-03-06
Cleanup Rule Extraction

### LINKED ISSUE #5279: Cleanup Rule Extraction

ugras-ergun-sonarsource:
Fixes #5279


build/DownloadDependencies/JarProcessing.targets
```diff
@@ -19,15 +19,13 @@
     -->
   <ItemGroup>
     <InputFile Include="..\..\src\EmbeddedSonarAnalyzer.props" Visible="false" />
-    <OutputFile Include="$(_WorkingDirectory)\lastrun.txt" Visible="false" />
+    <OutputFile Include="$(SloopBinaryDownloadDir)\lastrun.txt" Visible="false" />
```
ugras-ergun-sonarsource, line 22: I wanted to drop WriteOutputFile step also but it caused Build the fail so instead I've changed the folder only. To drop it we need the change how this project is run and I think it's not a can of worms we want to open now.

georgii-borovinskikh-sonarsource (approved):

PR #5282, opened 2024-03-06
Add sloop locator + config scope initialization test

georgii-borovinskikh-sonarsource:


src/SLCore.IntegrationTests/DependencyLocator.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        var sloopVersion = GetDependencyVersion("EmbeddedSloopVersion", dependencyProps);
+        var sloopPath = Path.Combine(localAppData,
+            "SLVS_Build_Sloop",
+            $"sonarlint-backend-cli-{sloopVersion}-windows_x64",
+            "bin",
+            "sonarlint-backend.bat");
+        if (!File.Exists(sloopPath))
+        {
+            throw new InvalidOperationException();
```
ugras-ergun-sonarsource, line 71: Add a message to the exception so we can see what the problem is without debugging.

src/SLCore.IntegrationTests/DependencyLocator.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    private static string GetAnalyzerPath(string analyzerFileName, string analyzerVersion, string[] analyzerJars) => 
+        analyzerJars.First(x => Path.GetFileName(x) == $"{analyzerFileName}-{analyzerVersion}.jar");
+
+    private static string GetDependencyVersion(string tagName, XmlDocument dependencyProps)
+    {
+        var elementsByTagName = dependencyProps.GetElementsByTagName(tagName);
+        if (elementsByTagName.Count != 1)
+        {
+            throw new InvalidOperationException();
```
ugras-ergun-sonarsource, line 85: same as above

Issue #5291, opened 2024-03-11
Start - Stop SLOOP


georgii-borovinskikh-sonarsource:
- [x] Create process setup
- [x] Create Rpc setup
- [x] Create initialization code (send init, disable telemetry, set current config scope, initialize connections)
- [x] Refactor integration test infrastructure to reuse rpc setup as much as possible
- [x] Implement sloop & plugin jar locators
- [x] Implement Configurators
- [x] Start slcore from SLDaemonPackage & stop/dispose it there

PR #5294, opened 2024-03-12
Add SLCoreProcessFactory & related classes

georgii-borovinskikh-sonarsource:
Part of #5291

### DIFF
--- a/src/SLCore.IntegrationTests/SLCoreTestProcessRunner.cs
+++ b/src/SLCore.IntegrationTests/SLCoreTestProcessRunner.cs
@@ -24,6 +24,7 @@ using System.IO;
 using System.Threading;
 using Microsoft.VisualStudio.Threading;
 using SonarLint.VisualStudio.SLCore.Core;
+using SonarLint.VisualStudio.SLCore.Core.Process;
 using SonarLint.VisualStudio.SLCore.Protocol;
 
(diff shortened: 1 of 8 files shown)


src/SLCore.IntegrationTests/SLCoreTestProcessRunner.cs
```diff
@@ -51,7 +53,7 @@ public sealed class SLCoreTestProcessRunner : IDisposable
         bool enableVerboseLogs = false,
         bool enableAutoFlush = false)
     {
-        this.pathToBat = pathToBat;
+        launchParameters = new SLCoreLaunchParameters("cmd.exe", $"/c {pathToBat}");
```
ugras-ergun-sonarsource, line 56: It seems some users of IntelliJ have their access to cmd.exe disabled by group policy we might need to think about that. For now it's OK but please create a ticket for this so we won't forget.

ugras-ergun-sonarsource (approved):

src/SLCore.IntegrationTests/SLCoreTestProcessRunner.cs
georgii-borovinskikh-sonarsource: As discussed offline, this is only in tests code. It will be fixed when for product code later

PR #5297, opened 2024-03-13
Add SLCoreService with SLOOP initialization & shutdown

georgii-borovinskikh-sonarsource:
Part of #5291


src/SLCore/Configuration/SLCoreLaunchParameters.cs
```diff
@@ -18,6 +18,6 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-namespace SonarLint.VisualStudio.SLCore.Core.Process;
+namespace SonarLint.VisualStudio.SLCore.Configuration;
 
-internal record SLCoreLaunchParameters(string PathToExecutable, string LaunchArguments);
+public record SLCoreLaunchParameters(string PathToExecutable, string LaunchArguments);
```
georgii-borovinskikh-sonarsource, line 23: This might have become an irrelevant change after the files are moved. I will check and replace with internal later

src/SLCore/SLCoreHandle.cs
```diff
@@ -0,0 +1,125 @@
...
+        await lifecycleManagementSlCoreService.InitializeAsync(new InitializeParams(
+            constantsProvider.ClientConstants,
+            new HttpConfigurationDto(new SslConfigurationDto()),
+            constantsProvider.FeatureFlags,
+            storageRoot,
+            workDir,
+            embeddedPluginPaths: slCoreEmbeddedPluginJarProvider.ListJarFiles(),
+            connectedModeEmbeddedPluginPathsByKey: new Dictionary<string, string>(),
+            enabledLanguagesInStandaloneMode: new List<Language>
+            {
```
georgii-borovinskikh-sonarsource, line 88: should this be made configurable?

ugras-ergun-sonarsource (approved):

src/SLCore/Configuration/SLCoreLaunchParameters.cs
georgii-borovinskikh-sonarsource: The interface is public

src/SLCore/SLCoreHandle.cs
georgii-borovinskikh-sonarsource: @ugras-ergun-sonarsource

ugras-ergun-sonarsource: for now, it's OK I think it's easy to maintain at this stage. We can reconsider if this causes pain.

PR #5300, opened 2024-03-15
Move ITs to use SLCoreHandle, fix bug with double StartListening call

georgii-borovinskikh-sonarsource:
Part of #5291


src/SLCore/ISLCoreRpcFactory.cs
```diff
@@ -87,7 +87,6 @@ internal sealed class SlCoreRpc : ISLCoreRpc
         var slCoreJsonRpc = slCoreJsonRpcFactory.CreateSLCoreJsonRpc(jsonRpc);
         slCoreServiceProvider.SetCurrentConnection(slCoreJsonRpc);
         listenerSetUp.Setup(slCoreJsonRpc);
-        slCoreJsonRpc.StartListening();
```
georgii-borovinskikh-sonarsource, line 90: This is already called by listener set up

ugras-ergun-sonarsource (approved):

Issue #5299, opened 2024-03-15
Add Sloop logging


georgii-borovinskikh-sonarsource:
- [x] Add rpc tracing infrastructure
- [x] Add logging of sloop stderr logs to SL Output Window
- [x] Add optional logging of RPC tracing to file

PR #5307, opened 2024-03-19
Add ISLCoreErrorLogger

georgii-borovinskikh-sonarsource:
Part of #5299


src/SLCore/ISLCoreErrorLoggerFactory.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            while (!cancellationTokenSource.IsCancellationRequested)
+            {
+                var line = await errorStream.ReadLineAsync();
+                if (string.IsNullOrEmpty(line))
+                {
+                    continue;
+                }
+
+                logger.LogVerbose("[SLCORE-ERR]" + line);
```
pavel-mikula-sonarsource, line 91: `continue` should be avoided where possible, it makes the logic hard to follow
```suggestion
                if (!string.IsNullOrEmpty(line))
                {
                    logger.LogVerbose("[SLCORE-ERR]" + line);
                }
```

georgii-borovinskikh-sonarsource: I disagree. It is the case when you have many statements inside of the If body and there's a lot of branching, but here it actually makes the intent of skipping the iteration explicit. if I saw the suggested variant, I would think that someone forgot to finish the method. The good alternative would be to move the IsNullOrEmpty condition to the while statement, but it won't look great in this particular case due to variable assignment.

ugras-ergun-sonarsource: I agree that in this case having "continue" makes it more obvious of the intentionality of the skipping.

ugras-ergun-sonarsource (approved):

PR #5308, opened 2024-03-20
Add Rpc logging to File

### LINKED ISSUE #5299: Add Sloop logging

ugras-ergun-sonarsource:
Part of #5299


georgii-borovinskikh-sonarsource (changes requested):

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
```diff
@@ -90,7 +90,7 @@ public async Task Start()
 
         slCoreHandle = new SLCoreHandle(new SLCoreRpcFactory(slCoreTestProcessFactory, slCoreLocator,
                 new SLCoreJsonRpcFactory(new RpcMethodNameTransformer()),
-                new RpcDebugger { LogFilePath = Path.Combine(privateFolder, "logrpc.txt") },
```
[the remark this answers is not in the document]
ugras-ergun-sonarsource: Because it's not standard behaviour and we do not check for that anyway in integration tests. Also, it would create lots of files in the developer machines.

src/SLCore/Core/IRpcDebugger.cs
```diff
@@ -36,31 +36,29 @@ public interface IRpcDebugger
 internal class RpcDebugger : IRpcDebugger
 {
     private readonly IFileSystem fileSystem;
-    public string LogFilePath { get; set; } = null;
+    private readonly string logFilePath;
 
     [ImportingConstructor]
-    public RpcDebugger() : this(new FileSystem())
+    public RpcDebugger() : this(new FileSystem(), DateTime.Now)
```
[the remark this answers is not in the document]
ugras-ergun-sonarsource: Why?

src/SLCore/Core/IRpcDebugger.cs
```diff
@@ -36,31 +36,29 @@ public interface IRpcDebugger
...
+        logFilePath = $"{dateTime.ToString("yyyy-MM-dd_HHmmssffff")}.txt";
     }
-    
+
     public void SetUpDebugger(IJsonRpc jsonRpc)
     {
-        if (LogFilePath is null)
+        /* IMPORTANT!!!
+         * Enable this environment variable only if you want to collect rpc debug logs
+         * If you ask customer to enable this their SQ/SC user and token info will be leaked and inform them to clear sensitive data */
```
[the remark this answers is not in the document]
ugras-ergun-sonarsource: We can do it in a separate ticket if needed I'll add a card we can decide later.

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: you could override the file name to always be the same

src/SLCore/Core/IRpcDebugger.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Timezones can change. However unlikely it is, it's better to use UTC since it's stable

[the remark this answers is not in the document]
ugras-ergun-sonarsource: We don't care about timezones, we do not care about exact time this happened.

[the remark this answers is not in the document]
ugras-ergun-sonarsource: Actually because of timezones it could be confusing people would get logs with different date than they are in.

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: You can also add UTC at the end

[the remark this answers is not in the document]
ugras-ergun-sonarsource: It would be very seldom that a developer moves timezones and develop in the same day that would affect the order of the files. Only scenario I can think is if somebody crosses international date line and have change time zone automatically set up. 
Thinking having different dates with UTC would be around midnight is a very eurocentric approach, it's probable to have this scenario for a developer in California, Japan or New Zealand.

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
```diff
@@ -65,42 +65,51 @@ public void AddListener(ISLCoreListener listener)
...
-                new SLCoreListenerSetUp(listenersToSetUp)),
-            constantsProvider,
-            foldersProvider,
-            connectionProvider,
-            jarProvider,
-            noOpActiveSolutionBoundTracker,
-            noOpConfigScopeUpdater,
-            new NoOpThreadHandler());
-        await slCoreHandle.InitializeAsync();
+        try
```
ugras-ergun-sonarsource, line 68: I've just added try finally for environment variables rest is the same

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
```diff
@@ -65,42 +65,51 @@ public void AddListener(ISLCoreListener listener)
...
+            var jarProvider = Substitute.For<ISLCoreEmbeddedPluginJarLocator>();
+            jarProvider.ListJarFiles().Returns(DependencyLocator.AnalyzerPlugins);
+
+            var noOpActiveSolutionBoundTracker = Substitute.For<IActiveSolutionBoundTracker>();
+            noOpActiveSolutionBoundTracker.CurrentConfiguration.Returns(BindingConfiguration.Standalone);
+            var noOpConfigScopeUpdater = Substitute.For<IConfigScopeUpdater>();
+
+            slCoreHandle = new SLCoreHandle(new SLCoreRpcFactory(slCoreTestProcessFactory, slCoreLocator,
+                    new SLCoreJsonRpcFactory(new RpcMethodNameTransformer()),
+                    new RpcDebugger(new FileSystem(), new DateTime(0)), //we pass datetime so we do not create extra file for each run
```
ugras-ergun-sonarsource, line 97: only line changed

georgii-borovinskikh-sonarsource: why not create a constructor override instead of this hack with datetime?

src/SLCore/Core/IRpcDebugger.cs
```diff
@@ -36,31 +36,29 @@ public interface IRpcDebugger
...
+    public RpcDebugger() : this(new FileSystem(), DateTime.Now)
     {
     }
 
-    internal /* for testing */ RpcDebugger(IFileSystem fileSystem)
+    internal /* for testing */ RpcDebugger(IFileSystem fileSystem, DateTime dateTime)
     {
         this.fileSystem = fileSystem;
+
+        logFilePath = $"{dateTime.ToString("yyyy-MM-dd_HHmmssffff")}.txt";
```
georgii-borovinskikh-sonarsource, line 50: what is the base directory for this? is it going to be where the vsix is installed? It would be better if it was a constant folder (appdata, temp, etc.)

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
georgii-borovinskikh-sonarsource: Also, the log file originally was saved in the folder with the test name, now it's overridden by each test, which is not good

src/SLCore/Core/IRpcDebugger.cs
```diff
@@ -43,11 +43,17 @@ public RpcDebugger() : this(new FileSystem(), DateTime.Now)
     {
     }
 
-    internal /* for testing */ RpcDebugger(IFileSystem fileSystem, DateTime dateTime)
+    internal /* for testing */ RpcDebugger(string fileNameOverride) : this(new FileSystem(), DateTime.Now, fileNameOverride)
```
georgii-borovinskikh-sonarsource, line 46: You did it kind of backwards

```
RpcDebugger() : this(new FileSystem(), DateTime.Now)
RpcDebugger(IFileSystem fileSystem, DateTime fileDate) : this(fileSystem, $"{dateTime.ToString("yyyy-MM-dd_HHmmssffff")}.log")
RpcDebugger(IFileSystem fileSystem, string fileName)
```

but at the same time, the file will end up in the common folder for integration tests, which is not what we want, so:


```
RpcDebugger() : this(new FileSystem(), DateTime.Now)
RpcDebugger(IFileSystem fileSystem, DateTime fileDate) : this(fileSystem, $"{dateTime.ToString("yyyy-MM-dd_HHmmssffff")}.log", Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "SonarLint for Visual Studio", "Rpc Logs"))
RpcDebugger(IFileSystem fileSystem, string fileName, string baseFolder)
```

The integration tests then would initialize it as `new(new FileSystem(), "logrpc.txt", "privateFolder")`

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
```diff
@@ -94,7 +96,7 @@ public async Task Start()
 
             slCoreHandle = new SLCoreHandle(new SLCoreRpcFactory(slCoreTestProcessFactory, slCoreLocator,
                     new SLCoreJsonRpcFactory(new RpcMethodNameTransformer()),
-                    new RpcDebugger(new FileSystem(), new DateTime(0)), //we pass datetime so we do not create extra file for each run
+                    new RpcDebugger($"{testName}.log"),
```
georgii-borovinskikh-sonarsource, line 101: This will put integration test runs in the same folder as the vsix logs, and that's inconvenient (folder switching) and confusing

georgii-borovinskikh-sonarsource (changes requested):

src/SLCore/Core/IRpcDebugger.cs
ugras-ergun-sonarsource: I think integration test should mimic the behaviour as much as possible and having them on the common folder is not an issue. Only when we run the tests we would create these files and it won't create tons of file that would make it harder when we really run the logs.

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
ugras-ergun-sonarsource: wdym

georgii-borovinskikh-sonarsource: Before, the logs were saved at `bin/.../slcore/<test_name>` folder along with sloop private files

georgii-borovinskikh-sonarsource (approved):

PR #5312, opened 2024-03-20
Fix removing already removed config scope

### LINKED ISSUE #5310: Do not send remove config scope when it's not set

georgii-borovinskikh-sonarsource:
Fixes #5310


src/SLCore/State/ActiveConfigScopeTracker.cs
```diff
@@ -131,7 +131,10 @@ public void RemoveCurrentConfigScope()
 
         using (asyncLock.Acquire())
         {
-            Debug.Assert(currentConfigScope != null);
+            if (currentConfigScope is null)
```
georgii-borovinskikh-sonarsource, line 134: I can't do anything about remove being called multiple times, so it's best to just ignore it

ugras-ergun-sonarsource (approved):

Issue #5317, opened 2024-03-21
Restart SLOOP


georgii-borovinskikh-sonarsource:
- [x] Add restart infrastructure
- [x] Make RPC debugger not overwrite the file between restarts
- [x] Specify restart trigger
- [x] Add restart trigger
- [x] Add Goldbar for manual restart

PR #5320, opened 2024-03-25
SloopRestartFailedNotificationService Implemented

ugras-ergun-sonarsource:
Part of #5317


src/SLCore/Notification/ISloopRestartFailedNotificationService.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        [ImportingConstructor]
+        public SloopRestartFailedNotificationService(INotificationService notificationService)
+        {
+            this.notificationService = notificationService;
+        }
+
+        public void Show(Action act)
+        {
+            var notification = new VisualStudio.Core.Notifications.Notification(
```
georgii-borovinskikh-sonarsource, line 48: I don't remember the exact way of doing this, but we need to make sure this notification can not be muted forever. If it's already like that, then you can skip this note.

georgii-borovinskikh-sonarsource (approved):
Only one question, otherwise LGTM

PR #5322, opened 2024-03-25
Add Sloop restart initiation logic + renaming

georgii-borovinskikh-sonarsource:
Part of #5317


src/SLCore/ISLCoreHandler.cs
```diff
@@ -20,116 +20,81 @@
...
+            throw new ObjectDisposedException(nameof(SLCoreHandler));
         }
-        finally
+        
+        threadHandling.RunOnBackgroundThread(async () =>
         {
-            logger.WriteLine(SLCoreStrings.SLCoreHandler_InstanceDied);
-            HandleInstanceDeath();
-        }
+            while (!disposed && slCoreInstanceHandler.CurrentStartNumber - initiatedStartAtCount < maxAutoStartsBeforeManual)
```
ugras-ergun-sonarsource, line 78: Should be <= I believe run number should be "restartnumber + 1"

georgii-borovinskikh-sonarsource: the initial start is not counted as a restart, there are tests for that as well

ugras-ergun-sonarsource (approved):

PR #5329, opened 2024-03-26
Trim paths loaded from VS in solution mode

### LINKED ISSUE #5328: Paths coming from VS solution should be trimmed
While CMake and the command line don't trim those paths, it seems that VS automatically trims them.
Outside of the VS solution, they aren't trimmed because a space at the beginning or end of a directory on Windows, while not recommended, is allowed.

The fact that we currently don't trim paths means that the analyzer might not find some paths when the user inserts an extra space by mistake in the VS project settings.

AbbasNS:
Fixes #5328


AbbasNS:
Hi @georgii-borovinskikh-sonarsource,
This PR should fix the issue you reported: https://sonarsource.slack.com/archives/CFTDKT6RW/p1711371329608009.
Feel free to adjust/take over the PR.

georgii-borovinskikh-sonarsource (approved):

Issue #5092, opened 2023-12-07
Make branch matching algorithm aligned with other SonarLint flavours. 


ugras-ergun-sonarsource:
There are two differences between SLVS and other flavours. 

- [x] We do not ignore short-lived branches. (only applicable to SC; in SQ there's no distinction of short-lived vs long-lived branches) 
- [x] We need to prioritise main branch when there's a tie between branches. 

Based on slack discussion: https://sonarsource.slack.com/archives/C03PE8P8TK6/p1701336633925369

kevin-hinz:
🙏 Please notify the Documentation Squad when this issue is started. 
DOC-1043 is waiting to implement the changes to the docs.

ugras-ergun-sonarsource:
We haven't done a proper hardening sprint since this was created it'll be the top ticket when we do a hardening.

PR #5331, opened 2024-03-27
Short lived branches ignored on branch matching

ugras-ergun-sonarsource:
Part of #5092

### DIFF
--- a/src/ConnectedMode/BranchMatcher.cs
+++ b/src/ConnectedMode/BranchMatcher.cs
@@ -51,6 +51,8 @@ namespace SonarLint.VisualStudio.ConnectedMode
     [PartCreationPolicy(CreationPolicy.Shared)]
     internal class BranchMatcher : IBranchMatcher
     {
+        private const string shortBranchType = "SHORT";
+
         private readonly ISonarQubeService sonarQubeService;
         private readonly ILogger logger;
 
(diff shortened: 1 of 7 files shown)


src/SonarQube.Client/Models/SonarQubeProjectBranch.cs
```diff
@@ -27,12 +27,14 @@ public class SonarQubeProjectBranch
         public string Name { get; }
         public bool IsMain { get; }
         public DateTimeOffset LastAnalysisTimestamp { get; }
+        public string Type { get; set; }
```
georgii-borovinskikh-sonarsource, line 30: Why not enum?

src/ConnectedMode/BranchMatcher.cs
```diff
@@ -93,7 +95,7 @@ private async Task<string> DoGetMatchingBranch(string projectKey, IRepository gi
                 return null;
             }
 
-            var remoteBranches = await sonarQubeService.GetProjectBranchesAsync(projectKey, token);
+            var remoteBranches = (await sonarQubeService.GetProjectBranchesAsync(projectKey, token)).Where(b => b.Type != shortBranchType);
```
georgii-borovinskikh-sonarsource, line 98: Do we still want to ignore it even if it's the exact match?

georgii-borovinskikh-sonarsource (approved):

src/SonarQube.Client/Models/SonarQubeProjectBranch.cs
ugras-ergun-sonarsource: We do not have the list of all possible values. Also we do not need to maintain the values since we do not care other than short.

src/ConnectedMode/BranchMatcher.cs
ugras-ergun-sonarsource: Yes. https://sonarsource.slack.com/archives/C02E6L5C01H/p1711613419949759?thread_ts=1711552233.646179&cid=C02E6L5C01H

PR #5367, opened 2024-04-18
Gb/open issue prechecks

georgii-borovinskikh-sonarsource:


georgii-borovinskikh-sonarsource:
@ugras-ergun-sonarsource ignore the commit list, it's the rebase issue. The diff is correct, except for the infobarmanager. It got moved, no actual changes

src/IssueViz/OpenInIde/OpenInIdeConfigScopeValidator.cs
```diff
@@ -0,0 +1,75 @@
...
+        this.logger = logger;
+        this.threadHandling = threadHandling;
+    }
+
+    public bool TryGetConfigurationScopeRoot(string issueConfigurationScope, out string configurationScopeRoot)
+    {
+        threadHandling.ThrowIfOnUIThread();
+
+        var configScope = activeConfigScopeTracker.Current;
+        configurationScopeRoot = configScope?.RootPath;
```
ugras-ergun-sonarsource, line 53: You set it first and return false but this would set the out parameter even if it's false. Are you sure it's expected behaviour?

ugras-ergun-sonarsource: Yes, but I think you shouldn't also set the value when it's false.

georgii-borovinskikh-sonarsource: You have to set some value to the out parameter. There's no difference which value you set if the caller ignores the fact that the result is false.

ugras-ergun-sonarsource: You should return the default value when it's false, setting something other than default is wrong behaviour.

georgii-borovinskikh-sonarsource: Why?

ugras-ergun-sonarsource: Because this is the expected behaviour, If you tried to get some value and failed then you shouldn't have any value on the field. For instance how would you feel if TryGetValue of a dictionary returned a value from the dictionary even the result was false? You are creating dirty data.

georgii-borovinskikh-sonarsource: I would never use the value if the result is false

ugras-ergun-sonarsource: The issue is not whether you use it or not; the issue is leaking dirty data. We should avoid this.

ugras-ergun-sonarsource (approved):

PR #5391, opened 2024-04-26
Add IListFilesListener implementation

### LINKED ISSUE #5387: wire up list files to sloop
- [x]  Implement IListFilesListener
- [x] Make SolutionWorkspaceService can be run from any thread
- [x] Implement Update Root on ConfigScope

ugras-ergun-sonarsource:
Part of #5387


src/SLCore.Listeners/Implementation/ListFilesListener.cs
```diff
@@ -0,0 +1,68 @@
...
+                clientFileDtos.AddRange(fullFilePathList.Select(fp => new ClientFileDto(new Uri(fp), GetRelativePath(fp), parameters.configScopeId, null, Encoding.UTF8.WebName, fp)));
+            }
+            return Task.FromResult(new ListFilesResponse(clientFileDtos));
+        }
+
+        private string GetRelativePath(string fullPath)
+        {
+            var root = folderWorkspaceService.FindRootDirectory();
+
+            return fullPath.Substring(root is not null ? root.Length : 2);
```
georgii-borovinskikh-sonarsource, line 65: `2` is a magic number and could be unreliable, though I'm not sure on the 2nd point

src/SLCore.Listeners/Implementation/ListFilesListener.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            this.folderWorkspaceService = folderWorkspaceService;
+            this.solutionWorkspaceService = solutionWorkspaceService;
+            this.activeConfigScopeTracker = activeConfigScopeTracker;
+        }
+
+        public Task<ListFilesResponse> ListFilesAsync(ListFilesParams parameters)
+        {
+            var clientFileDtos = new List<ClientFileDto>();
+            if (activeConfigScopeTracker.Current.Id == parameters.configScopeId)
```
georgii-borovinskikh-sonarsource, line 53: current configuration scope RootPath should be updated, otherwise showIssue showHotspot implementations won't work

src/SLCore.Listeners/Implementation/ListFilesListener.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                var fullFilePathList = folderWorkspaceService.IsFolderWorkspace() ? folderWorkspaceService.ListFiles() : solutionWorkspaceService.ListFiles();
```
georgii-borovinskikh-sonarsource, line 55: ListFiles for solution service will throw since this is not the UI thread. Consider changing the implementation of SolutionWorkspaceService to accept calls from any thread (add RunOnUIThread instead of ThrowIfNotOnUIThread)

georgii-borovinskikh-sonarsource (changes requested):

src/SLCore.Listeners/Implementation/ListFilesListener.cs
ugras-ergun-sonarsource: Unless how windows disk system work changes it'll be always 2 but we can use a more meaningful variable to make it easier to understand.

src/SLCore.Listeners/Implementation/ListFilesListener.cs
ugras-ergun-sonarsource: I'll handle this in a separate pr

src/SLCore/State/ActiveConfigScopeTracker.cs
```diff
@@ -34,9 +34,14 @@ namespace SonarLint.VisualStudio.SLCore.State;
...
 {
     ConfigurationScope Current { get; }
+
     void SetCurrentConfigScope(string id, string connectionId = null, string sonarProjectKey = null);
+
     void Reset();
+
     void RemoveCurrentConfigScope();
+
+    void UpdateRootOnCurrentConfigScope(string root);
```
georgii-borovinskikh-sonarsource, line 44: Not thread safe. You need to pass the configScopeId in case it has changed. Also it would be useful to make it return bool in that case.

georgii-borovinskikh-sonarsource (changes requested):

georgii-borovinskikh-sonarsource (approved):

PR #5392, opened 2024-04-26
SolutionWorkspaceService should not throw on background thread

ugras-ergun-sonarsource:
Part of #5387

### DIFF
--- a/src/Integration.UnitTests/LocalServices/SolutionWorkspaceServiceTests.cs
+++ b/src/Integration.UnitTests/LocalServices/SolutionWorkspaceServiceTests.cs
@@ -55,7 +55,9 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.LocalServices
 
             var serviceProvider = Substitute.For<IServiceProvider>();
 
-            var testSubject = new SolutionWorkspaceService(solutionInfoProvider, serviceProvider);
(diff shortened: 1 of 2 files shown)


src/Integration.UnitTests/LocalServices/SolutionWorkspaceServiceTests.cs
```diff
@@ -55,7 +55,9 @@ public void IsSolutionWorkSpace_ShouldBeOppsiteOfFolderWorkSpace(bool isFolderSp
 
             var serviceProvider = Substitute.For<IServiceProvider>();
 
-            var testSubject = new SolutionWorkspaceService(solutionInfoProvider, serviceProvider);
```
georgii-borovinskikh-sonarsource, line 58: You need to verify UI thread is called

georgii-borovinskikh-sonarsource (changes requested):
UI thread test

src/Integration.UnitTests/LocalServices/SolutionWorkspaceServiceTests.cs
georgii-borovinskikh-sonarsource: You can make a separate test for that

georgii-borovinskikh-sonarsource (approved):

PR #5395, opened 2024-05-01
Open in IDE: various bug fixes

georgii-borovinskikh-sonarsource:
Fixes:

- [x] `SolutionWorkspaceService` & `FolderWorkspaceService`: change `IEnumerable` to `ICollection` in file listing to prevent delaying collection enumeration due to laziness
- [x] `SolutionWorkspaceService`: deduplicate files when listing solution items
- [x] `SLCoreConstantsProvider`: enable managing of the Open in IDE server by SLOOP
- [x] Add comment pointing to https://sonarsource.atlassian.net/browse/SLCORE-783 ticket for ide file path fix
- [x] Remove `TextRangeWithHashDto` as it is not what is being used by the core for Open in IDE
- [x] `LocationDto`: add `codeSnippet` property
- [x] `LocationDto`: change `textRange` property type to `TextRangeDto`
- [x] Fix `LocationDto.ideFilePath` property name
- [x] `OpenInIdeHandlerImplementation`: add initial Open in IDE request validation
- [x] `OpenInIdeHandlerImplementation`: change error list tool window  to be shown before navigating to the issue to prevent location hiding
- [x]  `ClientFileDto`: replace `Uri` with `string`, as `Uri` is fragile when it comes to paths with whitespaces and special characters
- [x] `ListFilesListener` now exports the correct `ISLCoreListener` interface
- [x] `ListFilesListener` now uses `file://` schema with `Uri.Encode(filePath)` as the location
- [x] `ListFilesListener` now appends directory separator to all roots to avoid `"C:" + "file\path"` -> `"C:file\path"` instead of expected `"C:\file\path\"`
- [x] `ListFilesListener` now uses `Path.GetPathRoot` as the root provider instead of custom logic (this allows to support UNC paths)
- [x] Fix `ShowHotspotParams.hotspotDetails` property name
- [x] Add `datetime` and `threadid` to `RpcDebugger`
- [x] Fix typo in `DidRemoveConfigurationScopeParams.removedId`
- [x] Fix `OpenInIdeMessageBox` visibility


src/IssueViz/OpenInIde/OpenInIdeHandlerImplementation.cs
```diff
@@ -108,17 +109,40 @@ internal class OpenInIdeHandlerImplementation : IOpenInIdeHandlerImplementation
...
+    {
+        failureReason = default;
+
+        if (issueDetails is { Key: not null })
+        {
+            return true;
+        }
+
+        failureReason = OpenInIdeResources.ValidationReason_MalformedRequest;
+        return false;
```
pavel-mikula-sonarsource, line 136: This is pretty hard to follow, the logic is somewhat jumping all over the place.
It should either do this, to clearly say what happens on happy path and what happens on invalid path
```
       if (issueDetails is { Key: not null })
       {
           failureReason = default;
            return true;
        }
        else
        {
            failureReason = OpenInIdeResources.ValidationReason_MalformedRequest;
            return false;
        }

```

or, make the failureReson coupled with the result value
```
      failureReason = issueDetails is { Key: not null } ? default : OpenInIdeResources.ValidationReason_MalformedRequest;
      return failureReason == default;
```

and that actually shows that the `bool` is redundant to the failureReason being null or not. So this should just return a string and the invocation should be something like
```
DetailsFailureReason(issueDetails) is { } failureReason
```

src/IssueViz/OpenInIde/OpenInIdeHandlerImplementation.cs
```diff
@@ -108,17 +109,40 @@ internal class OpenInIdeHandlerImplementation : IOpenInIdeHandlerImplementation
             visualization = visualizationProcessor.HandleConvertedIssue(visualization) ;
         }
         issueSelectionService.SelectedIssue = visualization;
-        
+
+        toolWindowService.Show(toolWindowId);
+
         var navigationResult = navigator.TryNavigatePartial(visualization);
-        HandleNavigationResult(navigationResult, toolWindowId, visualization);
+        if (navigationResult != NavigationResult.OpenedLocation)
```
pavel-mikula-sonarsource, line 116: Positive conditions should be used to make it fluent to read => swap if&else

src/Core/IFolderWorkspaceService.cs
```diff
@@ -39,6 +39,6 @@ public interface IFolderWorkspaceService
         /// <summary>
         /// Returns the list of files in a folder view
         /// </summary>
-        IEnumerable<string> ListFiles();
+        ICollection<string> ListFiles();
```
pavel-mikula-sonarsource, line 42: Does the consumer need to be able to Add and remove items from the list? If not, this should be an array instead

src/Core/ISolutionWorkspaceService.cs
```diff
@@ -33,6 +33,6 @@ public interface ISolutionWorkspaceService
         /// List Files in a solution
         /// </summary>
         /// <returns>Returns empty if it's not in solution Mode</returns>
-        IEnumerable<string> ListFiles();
+        ICollection<string> ListFiles();
```
pavel-mikula-sonarsource, line 36: Sam eas IFolderWorkspaceSErver: Probably an array

src/Infrastructure.VS/FolderWorkspaceService.cs
```diff
@@ -63,12 +63,12 @@ public string FindRootDirectory()
             return solutionInfoProvider.GetSolutionDirectory();
         }
 
-        public IEnumerable<string> ListFiles()
+        public ICollection<string> ListFiles()
         {
             var root = FindRootDirectory();
 
             return root is not null ?
```
pavel-mikula-sonarsource, line 70: We need to fix the coding style of this statement (not now)

src/Infrastructure.VS/FolderWorkspaceService.cs
```diff
@@ -63,12 +63,12 @@ public string FindRootDirectory()
...
-                fileSystem.Directory.EnumerateFiles(root, "*", SearchOption.AllDirectories).Where(x => !x.Contains("\\node_modules\\")) :
+                fileSystem.Directory.EnumerateFiles(root, "*", SearchOption.AllDirectories).Where(x => !x.Contains("\\node_modules\\")).ToList() :
                 Array.Empty<string>();
```
pavel-mikula-sonarsource, line 72:
```suggestion
                [ ];
```

src/Integration/LocalServices/SolutionWorkspaceService.cs
```diff
@@ -65,17 +65,18 @@ public IEnumerable<string> ListFiles()
...
             threadHandling.RunOnUIThread(() => result = GetLoadedProjects(solution)
                 .SelectMany(AllItemsInProject)
                 .Where(x => x != null)
                 .Where(x => x.Contains("\\"))
                 .Where(x => !x.EndsWith("\\"))
                 .Where(x => !x.Contains("\\.nuget\\"))
-                .Where(x => !x.Contains("\\node_modules\\"))); // move filtering closer to path extraction to avoid processing unnecessary items)
-
+                .Where(x => !x.Contains("\\node_modules\\"))
+                .ToHashSet(StringComparer.InvariantCultureIgnoreCase)); // move filtering closer to path extraction to avoid processing unnecessary items)
```
pavel-mikula-sonarsource, line 78: This will be better as `.Distinct(...).ToArray` once this returns an array

pavel-mikula-sonarsource, line 78: I think an UT is missing for this deduplication logic.
It should show or explain by a comment the scenario under which duplicate files could be found.

src/SLCore.Listeners/Implementation/ListFilesListener.cs
```diff
@@ -21,17 +21,19 @@
...
 using SonarLint.VisualStudio.SLCore.Listener.Files;
 using SonarLint.VisualStudio.SLCore.Listener.Files.Models;
 using SonarLint.VisualStudio.SLCore.State;
 
 namespace SonarLint.VisualStudio.SLCore.Listeners.Implementation
 {
-    [Export(typeof(IListFilesListener))]
+    [Export(typeof(ISLCoreListener))]
     [PartCreationPolicy(CreationPolicy.Shared)]
     public class ListFilesListener : IListFilesListener
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: It is still useful as the contract declaration, IMO

src/SLCore.Listeners/Implementation/ListFilesListener.cs
```diff
@@ -60,16 +62,21 @@ public Task<ListFilesResponse> ListFilesAsync(ListFilesParams parameters)
                     var root = GetRoot(fullFilePathList.First());
                     if (activeConfigScopeTracker.TryUpdateRootOnCurrentConfigScope(parameters.configScopeId, root))
                     {
-                        clientFileDtos.AddRange(fullFilePathList.Select(fp => new ClientFileDto(new Uri(fp), GetRelativePath(root, fp), parameters.configScopeId, null, Encoding.UTF8.WebName, fp)));
+                        clientFileDtos.AddRange(fullFilePathList.Select(fp => new ClientFileDto(CreateUri(fp), GetRelativePath(root, fp), parameters.configScopeId, null, Encoding.UTF8.WebName, fp)));
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Filepaths can contain whitespaces and it breaks System.Uri. Also, backslashes didn't work well too. Base64 was just an easy way to get rid of both problems

src/SLCore.Listeners/Implementation/ListFilesListener.cs
```diff
@@ -60,16 +62,21 @@ public Task<ListFilesResponse> ListFilesAsync(ListFilesParams parameters)
...
                     }
                 }
             }
             return Task.FromResult(new ListFilesResponse(clientFileDtos));
         }
 
+        private static string CreateUri(string fp)
+        {
+            return "file://" + Convert.ToBase64String(Encoding.UTF8.GetBytes(fp));
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: whitespaces need to be encoded. urlencode was too complicated to use (dependency to a new assembly needed)

src/Core/IFolderWorkspaceService.cs
georgii-borovinskikh-sonarsource: It would constrain the implementation too much

src/Infrastructure.VS/FolderWorkspaceService.cs
georgii-borovinskikh-sonarsource: You mean the `?` at the end of the first line rather than at the beginning of the second? Ugras needs to fix his editor settings ;-)

src/Integration/LocalServices/SolutionWorkspaceService.cs
georgii-borovinskikh-sonarsource: The UTs are missing for this whole class. I could refactor it to isolate the IVsHierarchy logic. Do you want me to do that in this PR or create a ticket?

src/IssueViz/OpenInIde/OpenInIdeHandlerImplementation.cs
georgii-borovinskikh-sonarsource: > and that actually shows that the bool is redundant to the failureReason being null or not.

I did it in the TryGet way to be consistent with other validation conditions. While it's redundant, I think it reads better (especially since you can use `out var failureReason`)

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Actually, there are still possible problems with the base64 as it is not url-safe by default. I will look into fixing it some other way

src/SLCore.Listeners/Implementation/ListFilesListener.cs
```diff
@@ -78,6 +85,11 @@ private string GetRoot(string filePath)
 
             root ??= filePath.Substring(0, DiskRootSize);
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Done

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Changed to uri encode

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Removed base64

src/IssueViz/OpenInIde/OpenInIdeHandlerImplementation.cs
georgii-borovinskikh-sonarsource: Changed the method structure, but did not change the signature

src/IssueViz/OpenInIde/OpenInIdeHandlerImplementation.cs
georgii-borovinskikh-sonarsource: Done

src/Core/IFolderWorkspaceService.cs
georgii-borovinskikh-sonarsource: Changed to IReadOnlyCollection

src/Core/ISolutionWorkspaceService.cs
georgii-borovinskikh-sonarsource: Changed to IReadOnlyCollection

src/Infrastructure.VS/FolderWorkspaceService.cs
georgii-borovinskikh-sonarsource: Done

src/Infrastructure.VS/FolderWorkspaceService.cs
georgii-borovinskikh-sonarsource: Done

src/Integration/LocalServices/SolutionWorkspaceService.cs
georgii-borovinskikh-sonarsource: See IReadOnlyCollection change

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: It's not used => it's useless and should not be there. Dead code is not a clean code.

Another question is, if `ListFilesListener` should be actually renamed to `SLCoreListener`, because that is what it really implements now

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Please add this to the list of testing scenarios for later

src/Core/IFolderWorkspaceService.cs
pavel-mikula-sonarsource: IReadOnlyCollection is fine. 
We're free to change any API anytime we want. So having `string[]` doesn't introduce any problems on the long run.

src/Integration/LocalServices/SolutionWorkspaceService.cs
pavel-mikula-sonarsource: Ticket

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: It's a parameter required by slcore. Kind of a file id. It's not directly used now, but it could be used later

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: You mean end to end?

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: > Another question is, if ListFilesListener should be actually renamed to SLCoreListener, because that is what it really implements now

No, ISLCoreListener is an empty marker interface used for exporting the listeners to be registered with rpc.

IListFilesListener, as well as other listener interfaces, are used to declare an external contract. While they're not used directly, they serve as the reminder that the class should not be modified without any reason.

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: In the future, these interfaces may be used for painless mocking in some ITs

src/Integration/LocalServices/SolutionWorkspaceService.cs
georgii-borovinskikh-sonarsource: [#5405](https://github.com/SonarSource/sonarlint-visualstudio/issues/5405)
Opened in [SonarSource/sonarlint-visualstudio](https://github.com/SonarSource/sonarlint-visualstudio)

pavel-mikula-sonarsource (approved):
LGTM

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: It's easy to add such a thing in the future in case it will be needed in the future. For now, I can't explain why it is needed => it's confusing dead code

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
pavel-mikula-sonarsource: Yes, testing before release

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: [#5408](https://github.com/SonarSource/sonarlint-visualstudio/issues/5408)

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: [#5407](https://github.com/SonarSource/sonarlint-visualstudio/issues/5407)

PR #5419, opened 2024-05-08
Fix ERR_NoCompatibleVersion message

### LINKED ISSUE #5357: Fix ERR_NoCompatibleVersion message.
ERR_NoCompatibleVersion string uses hardcoded version information which is outdated. We should update it and make it parameterized.

vnaskos-sonar:
Fixes #5357

### DIFF
--- a/src/TypeScript.UnitTests/NodeJSLocator/CompatibleNodeLocatorTests.cs
+++ b/src/TypeScript.UnitTests/NodeJSLocator/CompatibleNodeLocatorTests.cs
@@ -22,6 +22,7 @@ using System;
 using System.Collections.Generic;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.TestTools.UnitTesting.Logging;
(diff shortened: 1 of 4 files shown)


georgii-borovinskikh-sonarsource (approved):

ite-klass:
I see this was merged May 10, and v8.0 was released on may 29.

In VS I see a "could not find supported Node.js runtime" error. May this change be the cause of it?

image

The (C#) analysis seems to work despite that.

image

But the error popping up every VS start is irritating.

ite-klass:
Or is that expected behavior and shows up even if I don't open a js or ts file?

georgii-borovinskikh-sonarsource:
Hi @ite-klass 

This change was a simple in product text change, not changes to the notification logic

We use node for CSS analysis as well as for JS and TS. But if your project doesn't have that, analysis for node languages should not trigger and the notification should not appear. If it is indeed the case, we would kindly ask you to submit the bug report with a minimal repro to https://community.sonarsource.com/c/sl/visual-studio/35

PR #5423, opened 2024-05-13
Make LogVerbose in TestLogger assertable

### LINKED ISSUE #5064: Make LogVerbose in TestLogger assertable
For some reason LogVerbose method of Testlogger class is not asserted let's make it assertable so we can test that also.

ugras-ergun-sonarsource:
Fixes #5064

### DIFF
--- a/src/SLCore.Listeners.UnitTests/Logging/LoggerListenerTests.cs
+++ b/src/SLCore.Listeners.UnitTests/Logging/LoggerListenerTests.cs
@@ -40,12 +40,12 @@ namespace SonarLint.VisualStudio.SLCore.Listeners.UnitTests.Logging
         }
 
         [TestMethod]
-        [DataRow(LogLevel.ERROR, true)]
-        [DataRow(LogLevel.WARN, true)]
-        [DataRow(LogLevel.INFO, false)]
-        [DataRow(LogLevel.TRACE, false)]
-        [DataRow(LogLevel.DEBUG, false)]
-        public void Log_LogsOnlyErrorAndWarning(LogLevel logLevel, bool logs)
+        //[DataRow(LogLevel.ERROR, false)]
+        //[DataRow(LogLevel.WARN, false)]
+        //[DataRow(LogLevel.INFO, true)]
+        //[DataRow(LogLevel.TRACE, true)]
+        [DataRow(LogLevel.DEBUG, true)]
+        public void Log_LogsOnlyErrorAndWarning(LogLevel logLevel, bool verboseLogs)
         {
             var param = new LogParams { level = logLevel, message = "some Message" };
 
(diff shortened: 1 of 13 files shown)


src/TestInfrastructure/Framework/TestLogger.cs
```diff
@@ -113,13 +113,12 @@ public void WriteLine(string messageFormat, params object[] args)
 
         public void LogVerbose(string message, params object[] args)
         {
-            // Note: verbose messages are not stored so we can't write tests against them.
-            // However, they can be logged to test output window to help with checking the output
-            // that is produced.
+            var logText = "[Verbose] " + GetFormattedMessage(string.Format(System.Globalization.CultureInfo.CurrentCulture, message, args));
```
georgii-borovinskikh-sonarsource, line 116: Just make the debug log thing a constructor flag, it would fix the existing tests. But then the question is, is this ticket even worth it? Why was it prioritized?

ugras-ergun-sonarsource: it's worth it I believe because we do not test some part of the functionality. We still have lots of places that we do not test verbose logs but at least this is a start.

src/SLCore.Listeners.UnitTests/Logging/LoggerListenerTests.cs
```diff
@@ -40,12 +40,12 @@ public void Mef_CheckIsSingleton()
...
-        [DataRow(LogLevel.ERROR, true)]
-        [DataRow(LogLevel.WARN, true)]
-        [DataRow(LogLevel.INFO, false)]
-        [DataRow(LogLevel.TRACE, false)]
-        [DataRow(LogLevel.DEBUG, false)]
-        public void Log_LogsOnlyErrorAndWarning(LogLevel logLevel, bool logs)
+        //[DataRow(LogLevel.ERROR, false)]
+        //[DataRow(LogLevel.WARN, false)]
+        //[DataRow(LogLevel.INFO, true)]
+        //[DataRow(LogLevel.TRACE, true)]
```
vnaskos-sonar, line 46: This probably shouldn't be left commented out.

src/SLCore.Listeners.UnitTests/Logging/LoggerListenerTests.cs
```diff
@@ -40,12 +40,12 @@ public void Mef_CheckIsSingleton()
...
+        [DataRow(LogLevel.DEBUG, true)]
+        public void Log_LogsOnlyErrorAndWarning(LogLevel logLevel, bool verboseLogs)
```
vnaskos-sonar, line 48: Changing from `bool logs` to `bool verboseLogs,` also changes the scope of the test.

Basically, it was checking if it logs only on `ERROR` and `WARN`, so on the `INFO`, `TRACE` and `DEBUG` I don't expect to see anything on the logs.

However, now it always logs something (either a regular log or a verbose log), which is confusing. I am not sure whether the test's name is the issue or if this should split into 2 different tests or if it's the intended TestLogger's behavior. 🤔

vnaskos-sonar (changes requested):
some changes

src/SLCore.Listeners.UnitTests/Logging/LoggerListenerTests.cs
ugras-ergun-sonarsource: 😅

src/SLCore.Listeners.UnitTests/Logging/LoggerListenerTests.cs
ugras-ergun-sonarsource: Old test did not make sense, for info, trace and debug it should call the logverbose and based on log level it's shown or not but on the loggerlistener test this is not important. I've changed the name of the test to remove confusion.

vnaskos-sonar (approved):
👍

Issue #5345, opened 2024-04-03
Drop VS 2019 from code


ugras-ergun-sonarsource:
Clean all references to VS 2019 from code

PR #5437, opened 2024-05-22
Bump to version 7.8.1

ugras-ergun-sonarsource:
Part of  #5345


ugras-ergun-sonarsource:
We need to a 2019 release because we need to change md files to add Deprecation announcement like in [VS 2017](https://marketplace.visualstudio.com/items?itemName=SonarSource.SonarLintforVisualStudio2017) this is just version bump. This branch will not be released to 2022 Marketplace or Github.

ugras-ergun-sonarsource:
There are some files about 2022 but it was not worth it to change it for a branch that live for couple of months and most likely won't be touched again.

PR #5438, opened 2024-05-22
Remove RPC debug logs from production build

### LINKED ISSUE #5436: Remove RPC debug logs from production build
RpcDebugger

It's a security concern if we write credentials to the disk. Until we can come up with a way to hide user secrets, we need to disable the logs.

vnaskos-sonar:
Fixes #5436


ugras-ergun-sonarsource (approved):
To be more explicit, the PR should be titled "Remove Rpc debug logs from production build" This would make the changelog easier to understand in the future.

ugras-ergun-sonarsource (approved):

PR #5453, opened 2024-05-27
Handle Open in IDE on standalone mode

### LINKED ISSUE #5441: Handle `Open in IDE` on standalone mode
When in standalone mode, Open in IDE fails, which is indicated by sending `noBindingSuggestionFound` from sloop to slvs, but we don't handle it

vnaskos-sonar:
Fixes #5441

### DIFF
--- a/src/Integration.UnitTests/Binding/BindingSuggestionHandlerTests.cs
+++ b/src/Integration.UnitTests/Binding/BindingSuggestionHandlerTests.cs
@@ -1,0 +1,96 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2024 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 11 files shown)


src/Integration.UnitTests/Binding/BindingSuggestionHandlerTests.cs
```diff
@@ -0,0 +1,96 @@
...
+    [DataRow(SonarLintMode.Standalone, true)]
+    [DataRow(SonarLintMode.Connected, false)]
+    public void Notify_ShowsMessage(SonarLintMode sonarLintMode, bool promptToConnect)
+    {
+        var notificationService = Substitute.For<INotificationService>();
+
+        var testSubject = CreateTestSubject(sonarLintMode: sonarLintMode, notificationService: notificationService);
+        testSubject.Notify();
+
+        if (promptToConnect)
```
georgii-borovinskikh-sonarsource, line 71: the assert looks kind of complicated because of the parameter, IMO it would be better to split the test in two

georgii-borovinskikh-sonarsource (approved):
LGTM, just one minor comment

PR #5482, opened 2024-06-09
SLVS-1176 Create RaiseIssuesParams to IAnalysisIssue Converter

ugras-ergun-sonarsource:
Fixes #

### DIFF
--- a/src/Integration.Vsix/SLCore/RaiseIssueParamsToAnalysisIssueConverter.cs
+++ b/src/Integration.Vsix/SLCore/RaiseIssueParamsToAnalysisIssueConverter.cs
@@ -98,7 +98,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix.SLCore
 
         private IQuickFix GetQuickFix(Uri fileURi, QuickFixDto quickFixDto)
         {
-            var fileEdit = quickFixDto.inputFileEdits.FirstOrDefault(e => e.target == fileURi);
+            var fileEdit = quickFixDto.inputFileEdits.Find(e => e.target == fileURi);
 
             if (fileEdit is not null)
             {
--- a/src/Integration.Vsix/SLCore/RaiseIssueParamsToAnalysisIssueConverter.cs
+++ b/src/Integration.Vsix/SLCore/RaiseIssueParamsToAnalysisIssueConverter.cs
@@ -1,0 +1,112 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2024 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Utilities;
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.Infrastructure.VS.Editor;
+using SonarLint.VisualStudio.SLCore.Common.Helpers;
+using SonarLint.VisualStudio.SLCore.Common.Models;
+using SonarLint.VisualStudio.SLCore.Listener.Analysis;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.SLCore
+{
+    [Export(typeof(IRaiseIssueParamsToAnalysisIssueConverter))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class RaiseIssueParamsToAnalysisIssueConverter : IRaiseIssueParamsToAnalysisIssueConverter
+    {
+        private readonly ITextDocumentFactoryService textDocumentFactoryService;
+        private readonly IContentTypeRegistryService contentTypeRegistryService;
+        private readonly ILineHashCalculator lineHashCalculator;
+
+        [ImportingConstructor]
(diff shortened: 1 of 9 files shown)


src/SLCore/Common/Helpers/ModelExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+using SonarLint.VisualStudio.Core.Analysis;
+using SonarLint.VisualStudio.SLCore.Common.Models;
+
+namespace SonarLint.VisualStudio.SLCore.Common.Helpers
+{
+    public static class ModelExtensions
+    {
+        public static AnalysisIssueSeverity ToAnalysisIssueSeverity(this IssueSeverity issueSeverity)
+        {
+            switch (issueSeverity)
```
georgii-borovinskikh-sonarsource, line 30: Convert to switch expression. Should be one click with ReSharper.

src/SLCore/Common/Helpers/ModelExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+                case IssueSeverity.INFO:
+                    return AnalysisIssueSeverity.Info;
+                default:
+                    throw new ArgumentException(SLCoreStrings.ModelExtensions_UnexpectedValue, nameof(issueSeverity));
+            }
+        }
+
+        public static AnalysisIssueType ToAnalysisIssueType(this RuleType ruleType)
+        {
+            switch (ruleType)
```
georgii-borovinskikh-sonarsource, line 49: Convert to switch expression. Should be one click with ReSharper.

src/SLCore/Common/Helpers/ModelExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+                case RuleType.SECURITY_HOTSPOT:
+                    return AnalysisIssueType.SecurityHotspot;
+                default:
+                    throw new ArgumentException(SLCoreStrings.ModelExtensions_UnexpectedValue, nameof(ruleType));
+            }
+        }
+
+        public static SoftwareQualitySeverity ToSoftwareQualitySeverity(this ImpactSeverity impactSeverity)
+        {
+            switch (impactSeverity)
```
georgii-borovinskikh-sonarsource, line 66: Convert to switch expression. Should be one click with ReSharper.

src/Integration.Vsix/SLCore/RaiseIssueParamsToAnalysisIssueConverter.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+            return null;
+        }
+
+        private IAnalysisIssueLocation GetAnalysisIssueLocation(string filePath, string message, TextRangeDto textRangeDto)
+        {
+            var textDocument = textDocumentFactoryService.CreateAndLoadTextDocument(filePath, contentTypeRegistryService.UnknownContentType); // load the document from disc
+            var currentSnapshot = textDocument.TextBuffer.CurrentSnapshot;
+
+            var lineHash = lineHashCalculator.Calculate(currentSnapshot, textRangeDto.startLine + 1);
```
georgii-borovinskikh-sonarsource, line 89: This is incorrect. The range hash should come from Sloop. If it doesn't, we should use ranges without the hash.

Range hash calculation is also done incorrectly here. You're calculating based on current snapshot, not based on the snapshot at the time of the analysis. This was part of the reason to introduce IssueConsumerStorage, to map to correct snapshot and translate to current.

So, remove references to VS-aware interfaces, simplify this class + tests and move it to the SLCore assembly.

georgii-borovinskikh-sonarsource (changes requested):

src/Integration.Vsix/SLCore/RaiseIssueParamsToAnalysisIssueConverter.cs
ugras-ergun-sonarsource: This the logic we currently use, I have added this so we can have the logic since we can get it from SLOOP.

georgii-borovinskikh-sonarsource: Can get it from sloop or cannot?

Also, it doesn't matter that it's the current logic. It's incorrect. You're calculating line hashes based on a potentially different snapshot, without translating the position. What you end up with is potentially incorrect line hashes that can lead to false-negative location matching. Don't make your life harder by writing extra code that doesn't even work properly.

ugras-ergun-sonarsource: https://github.com/SonarSource/sonarlint-visualstudio/blob/6c0113fa0422aa08edd71533dcfe3ea162701b84/src/CloudSecrets/SecretsAnalyzer.cs#L102

georgii-borovinskikh-sonarsource: It's not even the same scenario. There the content is used for analysis

src/SLCore/Common/Helpers/ModelExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                RuleType.CODE_SMELL => AnalysisIssueType.CodeSmell,
+                RuleType.BUG => AnalysisIssueType.Bug,
+                RuleType.VULNERABILITY => AnalysisIssueType.Vulnerability,
+                RuleType.SECURITY_HOTSPOT => AnalysisIssueType.SecurityHotspot,
+                _ => throw new ArgumentException(SLCoreStrings.ModelExtensions_UnexpectedValue, nameof(ruleType)),
+            };
+        }
+
+        public static SoftwareQualitySeverity ToSoftwareQualitySeverity(this ImpactSeverity impactSeverity)
```
georgii-borovinskikh-sonarsource, line 53: Could this method be reused [here](https://github.com/SonarSource/sonarlint-visualstudio/blob/e014d334e8ad8d2147a8b8c056bcf3c4be19de55/src/Education/Rule/SLCoreRuleMetaDataProvider.cs#L114)? If so, can you make that change as part of CaYC?

georgii-borovinskikh-sonarsource (approved):
1 small comment

src/SLCore/Common/Helpers/ModelExtensions.cs
ugras-ergun-sonarsource: No, because the changed types are not the same; we have so many of the same enums everywhere that we should unify them sometime.

georgii-borovinskikh-sonarsource: https://github.com/SonarSource/sonarlint-visualstudio/blob/e014d334e8ad8d2147a8b8c056bcf3c4be19de55/src/Education/Rule/SLCoreRuleMetaDataProvider.cs#L123

georgii-borovinskikh-sonarsource: Same types, no?

PR #5488, opened 2024-06-11
SLVS-1177 - Implement RaiseIssuesListener

ugras-ergun-sonarsource:


src/SLCore.Listeners/Implementation/AnalysisListener.cs
```diff
@@ -29,18 +29,42 @@ namespace SonarLint.VisualStudio.SLCore.Listeners.Implementation;
...
+        }
+    }
+
+    private IEnumerable<IAnalysisIssue> GetSupportedLanguageIssues(IEnumerable<IAnalysisIssue> issues)
+    {
+        return issues.Where(i => i.RuleKey.StartsWith("secrets:", StringComparison.InvariantCultureIgnoreCase));
     }
 
     public void RaiseHotspots(RaiseHotspotsParams parameters)
     {
```
georgii-borovinskikh-sonarsource, line 68: Why is this empty? We support hotspots in the issue consumer. This should either be implemented, or have a comment with explanation + a ticket to implement later

src/SLCore.Listeners/Implementation/AnalysisListener.cs
```diff
@@ -29,18 +29,42 @@ namespace SonarLint.VisualStudio.SLCore.Listeners.Implementation;
...
         return Task.CompletedTask;
     }
 
     public void RaiseIssues(RaiseIssuesParams parameters)
     {
-        throw new NotImplementedException();
+        if (parameters.analysisId.HasValue)
+        {
+            var filepath = parameters.issuesByFileUri.Single().Key.LocalPath;
```
georgii-borovinskikh-sonarsource, line 51: This will throw for empty lists. IMO it's better to explicitly check that there's not 0, not 2, but only 1 file path and handle other cases with logging

src/SLCore.Listeners/Implementation/AnalysisListener.cs
```diff
@@ -29,18 +29,42 @@ namespace SonarLint.VisualStudio.SLCore.Listeners.Implementation;
...
+            if (issues.Any())
+            {
+                analysisService.PublishIssues(filepath, parameters.analysisId.Value, issues);
+            }
+        }
+    }
+
+    private IEnumerable<IAnalysisIssue> GetSupportedLanguageIssues(IEnumerable<IAnalysisIssue> issues)
+    {
+        return issues.Where(i => i.RuleKey.StartsWith("secrets:", StringComparison.InvariantCultureIgnoreCase));
```
georgii-borovinskikh-sonarsource, line 64: Hardcoded magic string. Use Core.Language or add the mapping to SLCore.Common.Models.Language and use that

src/SLCore.Listeners/Implementation/AnalysisListener.cs
```diff
@@ -29,18 +29,42 @@ namespace SonarLint.VisualStudio.SLCore.Listeners.Implementation;
...
     }
 
     public void RaiseIssues(RaiseIssuesParams parameters)
     {
-        throw new NotImplementedException();
+        if (parameters.analysisId.HasValue)
+        {
+            var filepath = parameters.issuesByFileUri.Single().Key.LocalPath;
+
+            var issues = GetSupportedLanguageIssues(raiseIssueParamsToAnalysisIssueConverter.GetAnalysisIssues(parameters));
```
georgii-borovinskikh-sonarsource, line 53: You can filter before conversion

src/SLCore.Listeners/Implementation/AnalysisListener.cs
```diff
@@ -29,18 +29,42 @@ namespace SonarLint.VisualStudio.SLCore.Listeners.Implementation;
...
+
+            if (issues.Any())
+            {
```
georgii-borovinskikh-sonarsource, line 56: Clear on final raise issues? Not sure how it will work with raiseIssues + raiseHotspots, tho.

georgii-borovinskikh-sonarsource (changes requested):

src/SLCore.Listeners/Implementation/AnalysisListener.cs
georgii-borovinskikh-sonarsource: Sorry, not clear, AnalysisService.CancelForFile

Also, we might want to do this as well `analysisStatusNotifierFactory.Create(nameof(SLCoreAnalyzer), path).AnalysisFinished()`

georgii-borovinskikh-sonarsource: Maybe CancelForFile is not the best idea, but consider the notifier factory

src/SLCore.Listeners/Implementation/AnalysisListener.cs
ugras-ergun-sonarsource: We do not have hotspots in secrets. A comment would make sense here.

src/SLCore.Listeners/Implementation/AnalysisListener.cs
ugras-ergun-sonarsource: We should not have any scenarios in which this would happen. We would trigger analysis for a single file, and it would always have one item. I am not sure defensive approach is the best option here.

src/SLCore.Listeners/Implementation/AnalysisListener.cs
ugras-ergun-sonarsource: Analysis finished should not be called from this class but from analysis service. We might need to pass `isIntermediatePublication` though.

src/SLCore.Listeners/Implementation/AnalysisListener.cs
ugras-ergun-sonarsource: To do that we also need to change the `raiseIssueParamsToAnalysisIssueConverter`. This change makes sense but I'll open a separate ticket for that.

src/SLCore.Listeners/Implementation/AnalysisListener.cs
georgii-borovinskikh-sonarsource: Nah. AnalysisService is not aware of sloop analysis. This is finalizing that. However, we might be able to call it from Sloop analyzer if the response to AnalyzeAndTrack is consistent

src/SLCore.Listeners/Implementation/AnalysisListener.cs
georgii-borovinskikh-sonarsource: Why? The dto has the same rule key value

ugras-ergun-sonarsource: we pass the params object to the converter I do not want to manipulate the object itself or create a copy of it. It would be a quick change but deserves a separate ticket IMO.

src/SLCore.Listeners/Implementation/AnalysisListener.cs
ugras-ergun-sonarsource: AnalyzeAndTrack response can come before RaiseIssues so we can't do it there.

georgii-borovinskikh-sonarsource: Yeah so logically it should be done here, as it's the 2nd part of the Sloop analyzer

src/SLCore.Listeners/Implementation/AnalysisListener.cs
```diff
@@ -29,18 +30,64 @@ namespace SonarLint.VisualStudio.SLCore.Listeners.Implementation;
...
+                analysisService.PublishIssues(filepath, parameters.analysisId.Value, issues);
+            }
+            if (!parameters.isIntermediatePublication)
+            {
+                var analysisStatusNotifier = analysisStatusNotifierFactory.Create("SLCoreAnalyzer", filepath);
```
ugras-ergun-sonarsource, line 72: this magic string will need to be changed after analyzer task is done with `nameof`

georgii-borovinskikh-sonarsource: yep

georgii-borovinskikh-sonarsource (approved):

PR #5505, opened 2024-06-20
SLVS-1218 Improve `raiseIssues` handling and `analysisId` logging

georgii-borovinskikh-sonarsource:
Fixes #

### DIFF
--- a/src/Integration.Vsix/Analysis/AnalysisStrings.resx
+++ b/src/Integration.Vsix/Analysis/AnalysisStrings.resx
@@ -170,16 +170,16 @@
     <value>SonarLint: Analyzing {0}</value>
   </data>
   <data name="MSG_AnalysisComplete" xml:space="preserve">
-    <value>Finished analyzing {0}, analysis time: {1}s</value>
+    <value>Finished analyzing {0} with id {1}, analysis time: {2}s</value>
   </data>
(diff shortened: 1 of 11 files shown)


src/Integration.Vsix/Analysis/AnalysisStrings.resx
```diff
@@ -170,16 +170,16 @@
     <value>SonarLint: Analyzing {0}</value>
   </data>
   <data name="MSG_AnalysisComplete" xml:space="preserve">
-    <value>Finished analyzing {0}, analysis time: {1}s</value>
+    <value>Finished analyzing {0} with id {1}, analysis time: {2}s</value>
```
vnaskos-sonar, line 173: The analysisId is Nullable so there is a chance that the message will be `...with id null...` which feels a bit weird

vnaskos-sonar (approved):
Just a comment, but I don't have any strong opinion on it, it's just something to be aware of.

src/Integration.Vsix/Analysis/AnalysisStrings.resx
georgii-borovinskikh-sonarsource: That's just for legacy analyzers. It would be too much work doing separate strings for the 2 cases

PR #5507, opened 2024-06-24
SLVS-1215 Send Sloop shutdown on background thread, Initialize refactoring

georgii-borovinskikh-sonarsource:


vnaskos-sonar (approved):
👍

PR #5512, opened 2024-06-26
SLVS-1207 Add integration test that runs analysis for one file and finds a secret

vnaskos-sonar:


georgii-borovinskikh-sonarsource (approved):
LGTM. Add the file to exclusions before merging

PR #5515, opened 2024-07-01
SLVS-1219 Pass the correct file encoding to SLCore

vnaskos-sonar:


georgii-borovinskikh-sonarsource (changes requested):
Encoding can be changed on file save as well

georgii-borovinskikh-sonarsource (approved):

PR #5537, opened 2024-07-15
SLVS-1252 FileTracker is accessing ConfigScope without null check

gabriela-trutan-sonarsource:
Fix FileTracker is accessing ActiveConfigScopeTracker without null check


georgii-borovinskikh-sonarsource (approved):
LGTM

PR #5545, opened 2024-07-16
SLVS-1268 Enable JavaScript, TypeScript and CSS analysis in SLCore

georgii-borovinskikh-sonarsource:
+disable legacy analyzers


gabriela-trutan-sonarsource:
@georgii-borovinskikh-sonarsource Can you please double check the quick fixes? They seem to be off by 1. E.g. replacing `var c` -> `vletc` when replacing "var" with "let".

PR #5554, opened 2024-07-19
SLVS-1273 Send standalone Rule Configuration updates to SLCore

gabriela-trutan-sonarsource:


src/Integration.Vsix/Analysis/RuleSettingsMapper.cs
```diff
@@ -0,0 +1,37 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.SLCore.Service.Rules.Models;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.Analysis
+{
+    internal static class RuleSettingsMapper
```
georgii-borovinskikh-sonarsource, line 26: Could this be an interface? So that you can mock it in the UserSettingsProviderTests
Or keep it as private method of that class

src/Integration.Vsix/Analysis/RuleSettingsMapper.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        internal static Dictionary<string, StandaloneRuleConfigDto> MapRuleSettingsToSlCoreSettings(RulesSettings rulesSettings)
+        { 
+            return rulesSettings.Rules.ToDictionary(kvp => kvp.Key, kvp => MapStandaloneRuleConfigDto(kvp.Value));
+        }
+
+        private static StandaloneRuleConfigDto MapStandaloneRuleConfigDto(RuleConfig ruleConfig) => new(MapIsActive(ruleConfig.Level), MapParameters(ruleConfig.Parameters));
+        private static bool MapIsActive(RuleLevel ruleLevel) => ruleLevel == RuleLevel.On;
```
georgii-borovinskikh-sonarsource, line 34: I think these separate methods are a bit redundant

georgii-borovinskikh-sonarsource (approved):
Some minor comments.

src/Integration.Vsix/Analysis/RuleSettingsMapper.cs
gabriela-trutan-sonarsource: Making it private would make it less testable. Addin an interface seems unnecessary, because it does not need to be mocked in UserSettingsProviderTests

georgii-borovinskikh-sonarsource: From just looking at the UserSettingsProviderTests, how can you be sure that UserSettingsProvider uses the correct RuleSettingsMapper without having all of the same tests there (checking by value) or using a mock call (checking by delegation)?

src/Integration.Vsix/Analysis/RuleSettingsMapper.cs
georgii-borovinskikh-sonarsource: But they are one expression helpers. IMO that's a bit too much

src/Integration.Vsix/Analysis/UserSettingsProvider.cs
```diff
@@ -120,6 +124,24 @@ private UserSettings SafeLoadUserSettings(string filePath, ILogger logger)
             return new UserSettings(settings);
         }
 
+        private void UpdateStandaloneRulesConfiguration()
+        {
+            if (!slCoreServiceProvider.TryGetTransientService(out IRulesSLCoreService rulesSlCoreService))
+            {
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: The purpose is to be able to tell from the debug logs that a rule configuration update was discarded. Just in case we encounter a situation where the user sees the wrong configuration is being applied. This doesn't need to have a lot of info, just some static string would be fine. You can look at `FileTracker` for an example

src/Integration.Vsix.UnitTests/Analysis/RuleSettingsMapperTests.cs
```diff
@@ -0,0 +1,102 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.Core;
+using static SonarLint.VisualStudio.Integration.Vsix.Analysis.RuleSettingsMapper;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.Analysis
+{
```
georgii-borovinskikh-sonarsource, line 25: Minor: you can convert the nested namespaces to file-scoped ones for less indentation

gabriela-trutan-sonarsource: Done

src/Integration.Vsix/Analysis/UserSettingsProvider.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Done

src/Integration.Vsix/Analysis/RuleSettingsMapper.cs
gabriela-trutan-sonarsource: I think it is a matter of opinion. We can discuss it in the coding style.

src/Integration.Vsix/Analysis/RuleSettingsMapper.cs
gabriela-trutan-sonarsource: Added interface

src/Integration.Vsix/Analysis/RuleSettingsMapper.cs
```diff
@@ -0,0 +1,44 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.SLCore.Service.Rules.Models;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.Analysis;
+
+public interface IRuleSettingsMapper
```
georgii-borovinskikh-sonarsource, line 27: can be internal

src/Integration.Vsix/Analysis/RuleSettingsMapper.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    Dictionary<string, StandaloneRuleConfigDto> MapRuleSettingsToSlCoreSettings(RulesSettings rulesSettings);
+}
+
+[Export(typeof(IRuleSettingsMapper))]
+[PartCreationPolicy(CreationPolicy.Shared)]
+public class RuleSettingsMapper : IRuleSettingsMapper
```
georgii-borovinskikh-sonarsource, line 34: can be internal

src/Integration.Vsix/Analysis/UserSettingsProvider.cs
```diff
@@ -18,112 +18,137 @@
...
+using SonarLint.VisualStudio.SLCore;
+using SonarLint.VisualStudio.SLCore.Core;
+using SonarLint.VisualStudio.SLCore.Service.Rules;
 
-namespace SonarLint.VisualStudio.Integration.Vsix.Analysis
+namespace SonarLint.VisualStudio.Integration.Vsix.Analysis;
+
+[Export(typeof(IUserSettingsProvider))]
+internal sealed class UserSettingsProvider : IUserSettingsProvider, IDisposable
 {
```
georgii-borovinskikh-sonarsource, line 33: Git diff is so broken because of the namespace change...

src/Integration.Vsix/Analysis/RuleSettingsMapper.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    public Dictionary<string, StandaloneRuleConfigDto> MapRuleSettingsToSlCoreSettings(RulesSettings rulesSettings)
```
georgii-borovinskikh-sonarsource, line 36: Would it make sense to move the SLCoreServiceProvider dependency from the UserSettingsProvider to this class instead? So the slcore integration is fully contained within this class?

georgii-borovinskikh-sonarsource (commented):
Some comments. I will approve depending on the conclusion

src/Integration.Vsix/Analysis/UserSettingsProvider.cs
gabriela-trutan-sonarsource: Sorry for that! I wanted to apply CaYC and use file scoped namespace for the classes that I touch. Maybe the commit diff is more helpful?

src/Integration.Vsix/Analysis/RuleSettingsMapper.cs
gabriela-trutan-sonarsource: I agree. That task would require quite some refactoring so I would rather leave it as it is for now.

PR #5556, opened 2024-07-23
SLVS-1288 Provide the rules configuration when initializing SlCore

gabriela-trutan-sonarsource:


src/Core/UserRuleSettings/UserSettingsConstants.cs
```diff
@@ -0,0 +1,31 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.IO;
+
+namespace SonarLint.VisualStudio.Core.UserRuleSettings;
+
+public static class UserSettingsConstants
```
georgii-borovinskikh-sonarsource, line 25: This should be a property of IUserSettingsProvider

src/SLCore/Analysis/SLCoreRuleSettings.cs
```diff
@@ -0,0 +1,81 @@
...
+using SonarLint.VisualStudio.Core.UserRuleSettings;
+using SonarLint.VisualStudio.SLCore.Core;
+using SonarLint.VisualStudio.SLCore.Service.Rules;
+using SonarLint.VisualStudio.SLCore.Service.Rules.Models;
+
+namespace SonarLint.VisualStudio.SLCore.Analysis;
+
+public interface ISLCoreRuleSettings
+{
+    Dictionary<string, StandaloneRuleConfigDto> RulesSettings { get; }
```
georgii-borovinskikh-sonarsource, line 32: Convert to Get.. method as this property does some calculation inside and does not cache the value

src/SLCore/Analysis/SLCoreRuleSettings.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    [ImportingConstructor]
+    public SlCoreRuleSettings(ILogger logger, IUserSettingsProvider userSettingsProvider, ISLCoreServiceProvider slCoreServiceProvider)
+    {
+        this.logger = logger;
+        this.userSettingsProvider = userSettingsProvider;
+        this.slCoreServiceProvider = slCoreServiceProvider;
+    }
+
+    public Dictionary<string, StandaloneRuleConfigDto> RulesSettings => MapRuleSettingsToSlCoreSettings(userSettingsProvider.UserSettings.RulesSettings);
```
georgii-borovinskikh-sonarsource, line 52: public -> private -> public -> private is hard to read

src/SLCore/Analysis/SLCoreRuleSettings.cs
```diff
@@ -0,0 +1,81 @@
...
+using SonarLint.VisualStudio.SLCore.Core;
+using SonarLint.VisualStudio.SLCore.Service.Rules;
+using SonarLint.VisualStudio.SLCore.Service.Rules.Models;
+
+namespace SonarLint.VisualStudio.SLCore.Analysis;
+
+public interface ISLCoreRuleSettings
+{
+    Dictionary<string, StandaloneRuleConfigDto> RulesSettings { get; }
+    void UpdateStandaloneRulesConfiguration();
```
georgii-borovinskikh-sonarsource, line 33: It feels like these 2 things don't really belong together. We have UserSettingsUpdater and UserSettingsProvider split, but SLCoreSettings provider and updater are in the same class together

src/SLCore.UnitTests/SLCoreInstanceHandleTests.cs
```diff
@@ -477,7 +483,42 @@ private void SetUpThreadHandling(IThreadHandling threadHandling)
             compatibleNodeLocator,
             activeSolutionBoundTracker,
             configScopeUpdater,
-            threadHandling);
+            threadHandling, 
+            Substitute.For<ISLCoreRuleSettings>());
+    }
+
+    private SLCoreInstanceHandle CreateTestSubject(out ISLCoreRpcFactory slCoreRpcFactory,
```
georgii-borovinskikh-sonarsource, line 490: this feels like a copy-paste just to add 1 parameter, why?

src/SLCore.UnitTests/SLCoreInstanceHandleTests.cs
```diff
@@ -405,6 +410,7 @@ public void Dispose_NotInitialized_DoesNothing()
         });
         jarLocator.ListJarFiles().Returns(JarList);
         activeSolutionBoundTracker.CurrentConfiguration.Returns(new BindingConfiguration(Binding, SonarLintMode.Connected, "dir"));
+        slCoreRuleSettings?.RulesSettings.Returns(new Dictionary<string, StandaloneRuleConfigDto>());
```
georgii-borovinskikh-sonarsource, line 413: No unit tests where you actually verify that this value is ever used anywhere

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
```diff
@@ -148,12 +157,17 @@ private void SetUpLocalFolders()
...
 
         Directory.CreateDirectory(privateFolder);
         Directory.CreateDirectory(storageRoot);
         Directory.CreateDirectory(workDir);
         Directory.CreateDirectory(userHome);
     }
+
+    private static void DeleteDirectoryForLongPaths(string dirToDelete)
+    {
+        Directory.Delete(@"\\?\" + dirToDelete, true);
```
georgii-borovinskikh-sonarsource, line 171: Ideally should be a separate PR as it is an unrelated change. And how does it work?

src/SLCore.IntegrationTests/FileAnalysisTests.cs
```diff
@@ -109,25 +110,77 @@ public async Task StandaloneRuleConfig_CtorParamsAboveThreshold_JavaScriptActive
...
+        var issuesByFileUri = await RunFileAnalysis(ThreeSecretsIssuesPath, initialRulesConfig:secretsRuleConfig, updatedRulesConfig:null);
+
+        issuesByFileUri.Should().HaveCount(1);
+        issuesByFileUri[new FileUri(GetFullPath(ThreeSecretsIssuesPath))].Should().HaveCount(1);
+    }
+
     private async Task<Dictionary<FileUri, List<RaisedIssueDto>>> RunFileAnalysis(string fileToAnalyzeRelativePath, Dictionary<string, StandaloneRuleConfigDto> ruleConfigByKey = null)
     {
-        const string configScopeId = "ConfigScope1";
+        return await RunFileAnalysis(fileToAnalyzeRelativePath, null, ruleConfigByKey);
```
georgii-borovinskikh-sonarsource, line 137: why overload and not optional parameters?

src/Integration.Vsix/Analysis/UserSettingsUpdater.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    /// <summary>
+    /// Notification that one or more settings have changed
+    /// </summary>
+    event EventHandler SettingsChanged;
+
+}
+
+[Export(typeof(IUserSettingsUpdater))]
+internal sealed class UserSettingsUpdater : IUserSettingsUpdater, IDisposable
```
georgii-borovinskikh-sonarsource, line 38: This refactoring should have been a separate PR before the SLCore integration

src/Integration.Vsix/Analysis/UserSettingsUpdater.cs
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.UserRuleSettings;
+using SonarLint.VisualStudio.SLCore.Analysis;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.Analysis;
+
+public interface IUserSettingsUpdater
```
georgii-borovinskikh-sonarsource, line 28: The name is kind of misleading. It does the updates but it's also a source of events about updates. I get why it's a separate class, so it's just a naming question

src/SLCore/Analysis/SLCoreRuleSettings.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    public Dictionary<string, StandaloneRuleConfigDto> RulesSettings => MapRuleSettingsToSlCoreSettings(userSettingsProvider.UserSettings.RulesSettings);
+
+    private Dictionary<string, StandaloneRuleConfigDto> MapRuleSettingsToSlCoreSettings(RulesSettings rulesSettings)
+    { 
+        return rulesSettings.Rules.ToDictionary(kvp => kvp.Key, kvp => MapStandaloneRuleConfigDto(kvp.Value));
+    }
+
+    public void UpdateStandaloneRulesConfiguration()
+    {
```
georgii-borovinskikh-sonarsource, line 60: This method is not expected to be called on the UI thread, right?

georgii-borovinskikh-sonarsource (changes requested):

src/Core/UserRuleSettings/UserSettingsConstants.cs
gabriela-trutan-sonarsource: It can not be a property, because then I have to provide it in the UserSettingUpdater constructor to the SingleFileMonitor constructor creation. This makes the importing MEF constructor test to fail (because the property is null at the point when the constructor is called) So it has to be a static, so I thought a separate class makes more sense.

src/SLCore/Analysis/SLCoreRuleSettings.cs
gabriela-trutan-sonarsource: Can you be more explicit what are you referring to, please?

src/Integration.Vsix/Analysis/UserSettingsUpdater.cs
gabriela-trutan-sonarsource: Any proposal for a better name?

src/Integration.Vsix/Analysis/UserSettingsUpdater.cs
gabriela-trutan-sonarsource: Sorry. I am not really used with PRs. I shall try to keep in mind next time.

src/SLCore.IntegrationTests/FileAnalysisTests.cs
gabriela-trutan-sonarsource: Because they are more speaking. Optional parameters some disadvantages, like the fact  the number of optional parameters tend to increase (I have seen that a lot of unit tests methods have > 7 optional parameter), which is makes code very ugly and difficult to maintain.

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
gabriela-trutan-sonarsource: It is related to the integration test, that was failing for me locally. 

The @"\\?\" prefix is a special syntax used to handle long paths in Windows. Normally, Windows paths are limited to 260 characters, but using the @"\\?\" prefix allows you to work with paths up to 32,767 characters.

https://learn.microsoft.com/en-gb/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN

src/SLCore/Analysis/SLCoreRuleSettings.cs
gabriela-trutan-sonarsource: Yes, because the UserSettingsUpdater, which is in the Integration.Vsix assembly, is responsible to update the all the RuleSettings. 

UserSettingsProvider, which is in the Core assembly takes care of the settings.json file.

Finally, the SlCoreSettings takes care of sending the settings to SlCore. So I think it makes sense that the settings and sending the update to SlCore should be together. Would be more confusing for the consumers if I split the logic into multiple classes.

src/SLCore.UnitTests/SLCoreInstanceHandleTests.cs
gabriela-trutan-sonarsource: Good point.

src/SLCore.UnitTests/SLCoreInstanceHandleTests.cs
gabriela-trutan-sonarsource: Otherwise, I would have to update 9 unit tests just because one more parameter was added, when those unit tests do not care at all about the rules settings. That's why this coding style with out parameters is so ugly.

But yes, doing that leads to duplicate code, so I have to find a better way.

src/SLCore/Analysis/SLCoreRuleSettings.cs
gabriela-trutan-sonarsource: Right

src/Core/UserRuleSettings/UserSettingsConstants.cs
georgii-borovinskikh-sonarsource: Create export supports providing mocks with custom behavior. Ideally, the file watcher should be wrapped with a factory interface and mocked, so that we don't actually create the file watcher in tests.

src/SLCore/Analysis/SLCoreRuleSettings.cs
georgii-borovinskikh-sonarsource: To me it's just a bit of a misuse of the property. It's a fetch + convert, which is a complicated operation, and that should be a method. https://stackoverflow.com/questions/1374273/when-to-use-properties-instead-of-functions#:~:text=Use%20a%20property%20when%20the,ToString.

src/SLCore/Analysis/SLCoreRuleSettings.cs
georgii-borovinskikh-sonarsource: The method ordering. You have a public method, followed by a private method, followed by a public method and then a private method again

src/SLCore/Analysis/SLCoreRuleSettings.cs
georgii-borovinskikh-sonarsource: For example, `SLCoreInstanceHandle` would never need to access `UpdateStandaloneRulesConfiguration`, so, technically, the interface is not minimal. To me it still feels like a provider and a updater got merged together.

src/SLCore.UnitTests/SLCoreInstanceHandleTests.cs
georgii-borovinskikh-sonarsource: Having this out-parameter pattern is not ideal, so if we have more time I would refactor this, but as it is now, it's better to just add the new parameter to the existing method and update the usages.

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
georgii-borovinskikh-sonarsource: Have you tried this? https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=powershell#registry-setting-to-enable-long-paths

src/SLCore.IntegrationTests/FileAnalysisTests.cs
georgii-borovinskikh-sonarsource: But they're not more speaking if you use the same name for them. Rename the first one to `RunFileAnalysisWithDefaultRuleConfiguration`.

src/Integration.Vsix/Analysis/UserSettingsUpdater.cs
georgii-borovinskikh-sonarsource: 🤷

src/SLCore/Analysis/SLCoreRuleSettings.cs
georgii-borovinskikh-sonarsource: What do you think about adding ThrowIfOnUIThread call here? I'm not sure if we should

src/SLCore/Analysis/SLCoreRuleSettings.cs
gabriela-trutan-sonarsource: We should talk about in the weekly

src/Core/UserRuleSettings/UserSettingsConstants.cs
gabriela-trutan-sonarsource: Will be done as part of the architecture refactoring, in https://sonarsource.atlassian.net/browse/SLVS-1301

src/SLCore/Analysis/SLCoreRuleSettings.cs
gabriela-trutan-sonarsource: Will be done as part of the architecture refactoring, in https://sonarsource.atlassian.net/browse/SLVS-1301

src/SLCore/Analysis/SLCoreRuleSettings.cs
gabriela-trutan-sonarsource: Will be done as part of the architecture refactoring, in https://sonarsource.atlassian.net/browse/SLVS-1301

src/SLCore/Analysis/SLCoreRuleSettings.cs
gabriela-trutan-sonarsource: Added it to the weekly discussion

src/SLCore.IntegrationTests/FileAnalysisTests.cs
gabriela-trutan-sonarsource: Done, but the renaming is a little bit different.

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
gabriela-trutan-sonarsource: Yes. Locally I could delete the folders, but the unit tests were failing

src/SLCore.UnitTests/SLCoreInstanceHandleTests.cs
gabriela-trutan-sonarsource: Simplified the discussion, by removing both overloads and using TestInitialize. It was a quick refactoring.

src/SLCore.UnitTests/SLCoreInstanceHandleTests.cs
gabriela-trutan-sonarsource: Done.

georgii-borovinskikh-sonarsource (approved):

PR #5573, opened 2024-07-31
SLVS-1292 Use vulnerabilityProbability from SLCore instead of the hardcoded mapping

gabriela-trutan-sonarsource:


src/IssueViz.Security/Hotspots/HotspotReviewPriorityProvider.cs
```diff
@@ -1,223 +0,0 @@
...
-
-    [Export(typeof(IHotspotReviewPriorityProvider))]
-    [PartCreationPolicy(CreationPolicy.Shared)]
-    public class HotspotReviewPriorityProvider : IHotspotReviewPriorityProvider
-    {
-        // In order to save time and considering the fact that hotspots are no longer being updated, we decided to make a hardcoded list of priorities by rule key
-        // See (https://github.com/SonarSource/sonarlint-visualstudio/issues/4532)
-        private static readonly Dictionary<string, HotspotPriority> KeyToPriorityMap = new Dictionary<string, HotspotPriority>
-        {
-            { "cpp:S5443", HotspotPriority.Low },
```
georgii-borovinskikh-sonarsource, line 55: We've not migrated CFamily hotspots to slcore yet, so this shouldn't have been deleted for those.

src/SLCore/Listener/Analysis/RaiseFindingToAnalysisIssueConverter.cs
```diff
@@ -82,5 +100,14 @@ private static IAnalysisIssueFlow[] GetFlows(List<IssueFlowDto> issueFlows)
...
             new Edit(textEdit.newText,
                 new TextRange(textEdit.range.startLine, textEdit.range.endLine, textEdit.range.startLineOffset, textEdit.range.endLineOffset, null));
+
+        private static HotspotPriority? GetHotspotPriority(RaisedHotspotDto hotspotDto)
+        {
+            if(hotspotDto.vulnerabilityProbability == null)
+            {
+                return null;
+            }
+            return (HotspotPriority)Enum.Parse(typeof(HotspotPriority), hotspotDto.vulnerabilityProbability.ToString(), ignoreCase:true);
```
georgii-borovinskikh-sonarsource, line 110: convert to switch-case & add it to ModelConversionExtensions

src/SLCore/Listener/Analysis/RaiseFindingToAnalysisIssueConverter.cs
```diff
@@ -32,15 +32,33 @@ public class RaiseFindingToAnalysisIssueConverter : IRaiseFindingToAnalysisIssue
...
                     GetFlows(item.flows),
                     item.quickFixes?.Select(qf => GetQuickFix(fileUri, qf)).Where(qf => qf is not null).ToList(),
-                    item.ruleDescriptionContextKey))
-                .ToList();
+                    item.ruleDescriptionContextKey,
+                    GetHotspotPriority(raisedHotspotDto));
+            }
+
+            return new AnalysisIssue(item.ruleKey,
+                item.severity.ToAnalysisIssueSeverity(),
```
georgii-borovinskikh-sonarsource, line 54: create variables for repeated values

georgii-borovinskikh-sonarsource (changes requested):

src/IssueViz.Security/Hotspots/HotspotReviewPriorityProvider.cs
gabriela-trutan-sonarsource: I am not sure I understand this comment. The CFamily assembly is not referencing at all this class. So it should not be impacted.

georgii-borovinskikh-sonarsource: Yes but how do CFamily hotspots get this priority if the class is deleted? They are using the same old model without the priority

src/SLCore/Listener/Analysis/RaiseFindingToAnalysisIssueConverter.cs
gabriela-trutan-sonarsource: I agree to move it to ModelConversionExtensions.

But why a manual switch is better than a Parse converter? Using Parse has the advantage that any new value will be automatically supported. And if the value is not found in the expected enum, it will throw an exception

georgii-borovinskikh-sonarsource: Summary:
* Not any new value, they have to match. If there would be such case, the method will need to be rewritten...
* But it won't happen since hotspots are pretty much deprecated now. So the declarative switch approach has the advantage of being explicit and clear.
* This will make sure that the decoupling of dto and model is respected
* Does not introduce intermediary data structure

vnaskos-sonar (approved):
Works great, some comments on the tests 😉

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
```diff
@@ -181,6 +177,25 @@ public void UpdateForFile_NoServerHotspots_UsesReviewPriority()
...
+        var issueVis1 = new Mock<IAnalysisIssueVisualization>();
+        issueVis1.SetupGet(x => x.RuleId).Returns(rule1);
+        MockHotspotPriority(issueVis1, priority);
+        var reviewPriorityProviderMock = new Mock<IHotspotReviewPriorityProvider>();
+        var testSubject = CreateTestSubject(out _, reviewPriorityProvider: reviewPriorityProviderMock.Object);
+
+        testSubject.UpdateForFile("file1", new[] { issueVis1.Object });
+
+        VerifyContent(testSubject, new LocalHotspot(issueVis1.Object, priority));
+        reviewPriorityProviderMock.Verify(mock => mock.GetPriority(It.IsAny<string>()), Times.Never);
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: I am not sure what you mean. But the reviewPriorityProviderMock is important to check that it was really not called, never, in the case that the SlCore call was executed.

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
```diff
@@ -181,6 +177,25 @@ public void UpdateForFile_NoServerHotspots_UsesReviewPriority()
             new LocalHotspot(issueVis3.Object, HotspotPriority.Low));
     }
 
+    [TestMethod]
+    [DataRow(HotspotPriority.High)]
+    [DataRow(HotspotPriority.Medium)]
+    [DataRow(HotspotPriority.Low)]
+    public void UpdateForFile_NoServerHotspots_UsesHotspotsPriorityForNonCFamily(HotspotPriority priority)
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: There is a test that tests exactly that: UpdateForFile_NoServerHotspots_UsesDefaultReviewPriorityWhenUnmapped

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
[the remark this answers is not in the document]
vnaskos-sonar: If we remove this mock, and the code tries to call reviewPriorityProvider, the test will still fail. The `reviewPriorityProviderMock.Verify` is not necessary.

vnaskos-sonar (approved):
👍

src/IssueViz.Security.UnitTests/Hotspots/LocalHotspotStoreTests.cs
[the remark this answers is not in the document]
vnaskos-sonar: We keep the reviewPriorityProviderMock.Verify to make the test fail for the right reasons

georgii-borovinskikh-sonarsource (approved):

PR #5580, opened 2024-08-06
SLVS-1336 Fix: NullReferenceException in SlCoreTelemetryMigrationProvider when no Telemetry.xml is on disk

georgii-borovinskikh-sonarsource:


vnaskos-sonar (approved):
👍

PR #5587, opened 2024-08-08
SLVS-1332 Create Project Selection window

vnaskos-sonar:


src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs
```diff
@@ -0,0 +1,79 @@
...
+            RaisePropertyChanged(nameof(IsProjectSelected));
+        }
+    }
+
+    public bool IsProjectSelected => SelectedProject != null;
+
+    private string projectSearchTerm;
+    private ServerProject selectedProject;
+
+    public void InitProjects(List<ServerProject> projects)
```
gabriela-trutan-sonarsource, line 57: Maybe initialize projects?

gabriela-trutan-sonarsource (changes requested):
The implementation looks good, although I have some feedback on it.

The main points that should be improved are:

- do not trigger a search for every key stroke
- no way to clear the search filter
- merge the two windows as per the mockup
- add an indeterminate progress bar when clicking bind

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            selectedProject = value;
+            RaisePropertyChanged();
+            RaisePropertyChanged(nameof(IsProjectSelected));
+        }
+    }
+
+    public bool IsProjectSelected => SelectedProject != null;
+
+    private string projectSearchTerm;
```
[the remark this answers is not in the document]
vnaskos-sonar: They should be last after `public` `internal` `protected`

https://github.com/SonarSource/sonar-dotnet/blob/master/docs/coding-style.md#class-members

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    public void InitProjects(List<ServerProject> projects)
+    {
+        ProjectResults.Clear();
+        projects.ForEach(x => ProjectResults.Add(x));
+    }
+
+    private void SearchForProject()
+    {
+        if (string.IsNullOrEmpty(ProjectSearchTerm))
```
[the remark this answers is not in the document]
vnaskos-sonar: OnInitialize the ViewModel will be responsible for fetching the projects using SLCore. It will be simpler, in this case, to re-fetch the projects when the filter is cleared. I didn't implement that in this PR because it will require the SLCore logic or else I have to add some caching mechanism (a List probably) that will be removed on the next sprint.

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionWindow.xaml
```diff
@@ -0,0 +1,90 @@
...
+            <RowDefinition Height="*" />
+            <RowDefinition Height="Auto" />
+        </Grid.RowDefinitions>
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition Width="*" />
+            <ColumnDefinition Width="Auto" />
+            <ColumnDefinition Width="Auto" />
+        </Grid.ColumnDefinitions>
+
+        <TextBox Grid.Column="0" Grid.ColumnSpan="3" Grid.Row="0"
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Created a task to not forget this point: https://sonarsource.atlassian.net/browse/SLVS-1350

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Created a task to not forget this point: https://sonarsource.atlassian.net/browse/SLVS-1350

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionWindow.xaml.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    public ProjectSelectionViewModel ViewModel { get; }
+
+    public ProjectSelectionWindow(ConnectionInfo.Connection connection)
+    {
+        ViewModel = new ProjectSelectionViewModel(connection);
+        
+        InitializeComponent();
+
+        ViewModel.BindingFinished += (_, _) => Close();
```
gabriela-trutan-sonarsource, line 37: Please make sure to unsubscribe from the ViewModel's event to prepare memory leaks.

You should create a CloseWindow() event handler, register to it in the constructor and then unregister from it when the window is unloaded, for example (Wpf components have an Unloaded event), or any other appropriate event.

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionWindow.xaml
```diff
@@ -0,0 +1,130 @@
...
+                    VerticalAlignment="Bottom"
+                    Click="CancelButton_OnClick" />
+        </Grid>
+        <DockPanel Grid.Row="1" LastChildFill="True">
+            <StatusBar DockPanel.Dock="Bottom">
+                <StatusBarItem>
+                    <TextBlock Text="{Binding BindingStatusText}"></TextBlock>
+                </StatusBarItem>
+                <StatusBarItem HorizontalContentAlignment="Right">
+                    <ProgressBar Height="15" Width="100" IsIndeterminate="{Binding BindClicked}" />
```
gabriela-trutan-sonarsource, line 125: The progresBar should always be indeterminate. So IsIndeterminate=true. Instead it should be hidden by default

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionWindow.xaml
```diff
@@ -0,0 +1,130 @@
...
+                    VerticalAlignment="Bottom"
+                    Margin="5, 0"
+                    IsEnabled="{Binding IsBindButtonEnabled}"
+                    Click="BindButton_OnClick" />
+            <Button Grid.Column="2" Grid.Row="3" Content="{x:Static ui:UiResources.CancelBtn}"
+                    HorizontalAlignment="Right"
+                    VerticalAlignment="Bottom"
+                    Click="CancelButton_OnClick" />
+        </Grid>
+        <DockPanel Grid.Row="1" LastChildFill="True">
```
gabriela-trutan-sonarsource, line 119: Please make sure this DockPanel is hidden by default. Use the Visibility converter (see CredentialsWindow for example)

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs
```diff
@@ -0,0 +1,124 @@
...
+
+    public void InitializeProjects(List<ServerProject> projects)
+    {
+        ProjectResults.Clear();
+        projects.ForEach(x => ProjectResults.Add(x));
+    }
+
+    public void Bind()
+    {
+        BindClicked = true;
```
gabriela-trutan-sonarsource, line 96: Please make sure to set the BindClicked to false. I know it is a dummy component, but if we want to demo it we should really be able to show how it would work.

You could add a delay, for example, as well

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionWindow.xaml
gabriela-trutan-sonarsource, line 119: Consider showing an overlay to the whole window with the progress bar after the "Bind" is clicked.

gabriela-trutan-sonarsource (changes requested):

gabriela-trutan-sonarsource (changes requested):
The progress bar doesn't really work. Please make sure to show make it work for demo purposes.

gabriela-trutan-sonarsource (approved):

PR #5588, opened 2024-08-09
SLVS-1329 Create Credentials dialog for new connected mode

gabriela-trutan-sonarsource:


src/ConnectedMode/UI/Credentials/CredentialsViewModel.cs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public string Username
+    {
+        get => username;
+        set
+        {
+            username = value;
+            RaisePropertyChanged();
+            RaisePropertyChanged(nameof(IsConfirmationBtnEnabled));
```
georgii-borovinskikh-sonarsource, line 74: What do you think about having a full property instead of the computed one for `IsConfirmationBtnEnabled` , setting the value here and updating `RaisePropertyChanged` in the `IsConfirmationBtnEnabled` setter? That would remove the need to keep track of all of the property changed and delegate that.

src/ConnectedMode/UI/Credentials/CredentialsViewModel.cs
```diff
@@ -0,0 +1,108 @@
...
+            RaisePropertyChanged(nameof(ShouldPasswordBeFilled));
+        }
+    }
+
+    public bool IsTokenAuthentication => SelectedAuthenticationType == UiResources.Token;
+    public bool IsCredentialsAuthentication => SelectedAuthenticationType == UiResources.Credentials;
+    public bool ShouldTokenBeFilled => IsTokenAuthentication && !IsTokenProvided;
+    public bool ShouldUsernameBeFilled => IsCredentialsAuthentication && !IsUsernameProvided;
+    public bool ShouldPasswordBeFilled => IsCredentialsAuthentication && !IsPasswordProvided;
+    public bool IsConfirmationBtnEnabled => (IsTokenAuthentication && IsTokenProvided) || (IsCredentialsAuthentication && AreCredentialsProvided);
```
georgii-borovinskikh-sonarsource, line 96: I would not use contractions like Btn. I think in this case you may be able to just call it `IsConfirmationEnabled` since it's obvious from the context that there's only one type of confirmation

src/ConnectedMode/UI/Credentials/CredentialsViewModel.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    private bool IsTokenProvided => IsValueFilled(Token);
+    private bool AreCredentialsProvided => IsPasswordProvided && IsUsernameProvided;
+    private bool IsUsernameProvided => IsValueFilled(Username);
+    private bool IsPasswordProvided => IsValueFilled(Password);
+    public string GenerateTokenUrl => Connection.serverType == ServerType.SonarCloud ? SonarCloudGenerateTokenUrl : Path.Combine(Connection.id, SonarQubeGenerateTokenUrl);
```
georgii-borovinskikh-sonarsource, line 102: Could we disable this easily **if** we cut the branch during the next sprint?

georgii-borovinskikh-sonarsource: Ah wait this is just a hyperlink to the security page? Then I think we might need to change the name, because token generation is a bit more than that and involves communication between the ide (through SLCore) with the server to automatically exchange the token

georgii-borovinskikh-sonarsource (commented):
Some comments/questions. Didn't review the UI part yet

src/ConnectedMode/UI/Credentials/CredentialsViewModel.cs
gabriela-trutan-sonarsource: I wouldn't do that, because I think it would violate the single responsibility principle. Also the IsConfirmationBtnEnabled needs to be set in all those places, where we trigger the RaisePropertyChanged: when the username changes, the password, token etc

src/ConnectedMode/UI/Credentials/CredentialsViewModel.cs
gabriela-trutan-sonarsource: Maybe I can name it "AccountSecurityUrl" as it is a link to that page, where the generate token button in the server is present.

src/ConnectedMode/UI/Credentials/CredentialsWnd.xaml.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Windows.Input;
+using System.Windows.Navigation;
+using SonarLint.VisualStudio.ConnectedMode.UI.Resources;
+using SonarLint.VisualStudio.Core;
+using static SonarLint.VisualStudio.ConnectedMode.ConnectionInfo;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UI.Credentials
+{
+    [ExcludeFromCodeCoverage] // UI, not really unit-testable
+    public partial class CredentialsWnd : Window
```
georgii-borovinskikh-sonarsource, line 32:
```suggestion
    public partial class CredentialsDialog : Window
```
WDYT?

From https://learn.microsoft.com/en-us/windows/apps/design/controls/dialogs-and-flyouts/
```
Dialogs are modal UI overlays that provide contextual app information. Dialogs block interactions with the app window until being explicitly dismissed. They often request some kind of action from the user
```

gabriela-trutan-sonarsource: For now, our windows in UI folder all have the suffix "Window". But we could rename all of them. I would then propose to add this as a point to the task https://sonarsource.atlassian.net/browse/SLVS-1340

src/ConnectedMode/UI/Credentials/CredentialsWnd.xaml
```diff
@@ -0,0 +1,158 @@
...
+                    <Style TargetType="Border">
+                        <Style.Triggers>
+                            <DataTrigger Binding="{Binding ShouldPasswordBeFilled}" Value="True">
+                                <Setter Property="BorderBrush" Value="{StaticResource InvalidText}"></Setter>
+                                <Setter Property="BorderThickness" Value="1"></Setter>
+                            </DataTrigger>
+                        </Style.Triggers>
+                    </Style>
+                </Border.Style>
+                <TextBox Text="{Binding Path=Password, UpdateSourceTrigger=PropertyChanged}" Margin="0"
```
georgii-borovinskikh-sonarsource, line 128: Use PasswordBox

georgii-borovinskikh-sonarsource (changes requested):

src/ConnectedMode/UI/Credentials/CredentialsWnd.xaml
georgii-borovinskikh-sonarsource: https://github.com/SonarSource/sonarlint-visualstudio/blob/546208681c669cc11747194c4aa6e9d66c120dd9/src/Integration/Connection/UI/ConnectionInfoDialogView.xaml#L190

georgii-borovinskikh-sonarsource:
I think we should also clear the previous text form when switching to the other authentication type. 

Type something in the token field -> switch -> type something in the password field -> switch -> token has value -> switch password has value -> confused

src/ConnectedMode/UI/Resources/UiResources.resx
```diff
@@ -117,6 +117,12 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="AuthenticationDescriptionLbl" xml:space="preserve">
```
georgii-borovinskikh-sonarsource, line 120: We should not use contractions like `Lbl`, it's hard to read.

src/ConnectedMode/UI/Credentials/CredentialsWnd.xaml.cs
georgii-borovinskikh-sonarsource: I've created a separate task for this. https://sonarsource.atlassian.net/browse/SLVS-1343
We can probably make it a subtask of https://sonarsource.atlassian.net/browse/SLVS-1340

src/ConnectedMode/UI/Resources/UiResources.resx
georgii-borovinskikh-sonarsource: https://sonarsource.atlassian.net/browse/SLVS-1343

src/ConnectedMode/UI/Credentials/CredentialsWnd.xaml
gabriela-trutan-sonarsource: Makes sense to use PasswordBox also for Token

src/ConnectedMode/UI/Credentials/CredentialsWnd.xaml.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        private readonly IBrowserService browserService;
+
+        public CredentialsWnd(IBrowserService browserService, Connection connection, bool isWizardMode)
```
georgii-borovinskikh-sonarsource, line 39: Nor sure what wizard means in this context

src/ConnectedMode/UI/Credentials/CredentialsWnd.xaml.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        private void PasswordBox_OnPasswordChanged(object sender, RoutedEventArgs e)
+        {
+            ViewModel.Password = Password.Password;
+        }
+
+        private void Token_OnPasswordChanged(object sender, RoutedEventArgs e)
+        {
+            ViewModel.Token = Token.Password;
```
georgii-borovinskikh-sonarsource, line 72: `Token` -> `TokenBox`? Kind of confusing to see `Token.Password`. Could do the same with `Password.Password`

src/ConnectedMode/UI/Resources/Styles.xaml
```diff
@@ -7,6 +7,16 @@
         <Setter Property="MinWidth" Value="75" />
     </Style>
 
+    <Style TargetType="TextBox">
```
georgii-borovinskikh-sonarsource, line 10: Should we add theming here? Later maybe?

georgii-borovinskikh-sonarsource (approved):
Some minor comments. Can be merged without another review

src/ConnectedMode/UI/Resources/Styles.xaml
gabriela-trutan-sonarsource: I would say as part of: https://sonarsource.atlassian.net/browse/SLVS-1340

PR #5590, opened 2024-08-09
SLVS-1339 Create connections window

gabriela-trutan-sonarsource:


src/ConnectedMode/ConnectionInfo.cs
```diff
@@ -28,5 +28,5 @@ public enum ServerType
         SonarCloud
     }
 
-    public record Connection(string id, ServerType serverType);
+    public record Connection(string id, ServerType serverType, bool hasSmartNotifications);
```
vnaskos-sonar, line 31: I am not sure about the `has`. SLCore is using `bool RetrieveSmartNotifications`, and on our side, it is currently `bool DisableNotifications`.

vnaskos-sonar, line 31: On the caller side of this record
```
new Connection("https://sonarcloud.io/myOrg", ServerType.SonarCloud, false)
```
we pass obviously boolean values 😅 

However, I noticed that it was hard to understand what this `true` or `false` represents without looking at the record's definition. For code readability, it would be nice to replace it with an enum `SmartNotifications.Enabled`.

For comparison:
```
new Connection("https://sonarcloud.io/myOrg", ServerType.SonarCloud, SmartNotifications.Enabled)
```

vnaskos-sonar (changes requested):

src/ConnectedMode/ConnectionInfo.cs
gabriela-trutan-sonarsource: Morning! Regarding your feedback bellow. You mentioned that on our side is "DisableNotifications", but I couldn't find this string in the whole solution. I named the property hasSmartNotifications, because in our mockup it is called "Smart Notifications". I think it would be nice if the property in the view model reflects the property in the UI. But what would be a better name "SmartNotificationsEnable" or just "SmartNotifications"? The problem with just "SmartNotifications" is that it is not speaking that is a bool value.

gabriela-trutan-sonarsource: I renamed it RetrieveSmartNotifications.

gabriela-trutan-sonarsource: I understand your point, but I think it is not worth it to introduce an enum because it brings more complexity: we need to convert it to bool for the UI and then again convert it when passing it to SlCore.

src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
```diff
@@ -0,0 +1,46 @@
...
+using SonarLint.VisualStudio.Core.WPF;
+using static SonarLint.VisualStudio.ConnectedMode.ConnectionInfo;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UI.ManageConnections
+{
+    public class ManageConnectionsViewModel : ViewModelBase
+    {
+        private readonly ObservableCollection<ConnectionViewModel> connectionViewModels;
+
+        public ManageConnectionsViewModel(IEnumerable<Connection> connections)
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: The MVVM doesn't have a say in this as long as the View only communicates with the ViewModel and the ViewModel only communicates with the Model (in our case, the services). 

I passed the list to the constructor, because it is a dummy window and it was the easiest. You are probably right that, when having the real implementation, we would probably need to move the initialization in another method, to allow it to run asynchronously. So I might as well do it now.

src/ConnectedMode/ConnectionInfo.cs
vnaskos-sonar: The `DisableNotifications` is in the `ServerConnectionConfiguration` class. It would be better to check that with Giorgii. I believe we still receive notifications, but we discard them or do not show them.

src/ConnectedMode/UI/ManageConnections/ManageConnectionsWindow.xaml
```diff
@@ -0,0 +1,112 @@
...
+                         xmlns:vsShell="clr-namespace:Microsoft.VisualStudio.Shell;assembly=Microsoft.VisualStudio.Shell.15.0"
+                         xmlns:ui="clr-namespace:SonarLint.VisualStudio.ConnectedMode.UI.Resources"
+                         xmlns:imaging="clr-namespace:Microsoft.VisualStudio.Imaging;assembly=Microsoft.VisualStudio.Imaging"
+                         xmlns:vsimagecatalog="clr-namespace:Microsoft.VisualStudio.Imaging;assembly=Microsoft.VisualStudio.ImageCatalog"
+                         Title="{x:Static ui:UiResources.ManageConnectionsWndTitle}" 
+                         HorizontalAlignment="Center"
+                         SizeToContent="WidthAndHeight"
+                         ResizeMode="NoResize"
+                         MinWidth="750"  MaxWidth="750"
+                         MinHeight="500" MaxHeight="500"
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: When you add both Min and Max width to the same value, they are no longer resizable. I would let the decision about making all or no window resizable as part of the task https://sonarsource.atlassian.net/browse/SLVS-1340 

From my point of view, they should not be resizable. But maybe this is a point to clarify with Joseph.

vnaskos-sonar (approved):
👍

PR #5593, opened 2024-08-12
SLVS-1343 Clean up name contractions

gabriela-trutan-sonarsource:


georgii-borovinskikh-sonarsource:
I think we should still keep the prefixes like Label, Button, etc. Currently it's a bit hard to see what some resource strings are used for

src/ConnectedMode/UI/Resources/UiResources.resx
```diff
@@ -228,6 +228,12 @@
   <data name="ManualOrganizationSelectionLabel" xml:space="preserve">
     <value>Enter organization key:</value>
   </data>
+  <data name="AuthenticationType_Credentials" xml:space="preserve">
+    <value>Credentials</value>
+  </data>
+  <data name="AuthenticationType_Token" xml:space="preserve">
```
vnaskos-sonar, line 234: If this naming format is a new suggestion, I would suggest bringing it up first in one of our standups or slack, so we are all on the same page. At the moment, I suggest being consistent across the file and continuing by removing the underscore `_`.

vnaskos-sonar (changes requested):

src/ConnectedMode/UI/Resources/UiResources.resx
gabriela-trutan-sonarsource: Georgi showed us in the daily one .resx file that uses this convention. Take a look at MigrationStrings.resx.

I would not postpone merging this PR, because it will bring a lot of conflicts to all other PRs. Could you, please, clarify maybe now so we can merge this ASAP?

vnaskos-sonar (changes requested):
👍

vnaskos-sonar (approved):

PR #5596, opened 2024-08-14
SLVS-1349 Create project binding window

gabriela-trutan-sonarsource:


src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
```diff
@@ -0,0 +1,286 @@
...
+    [TestMethod]
+    public void BoundProject_Set_RaisesEvents()
+    {
+        var eventHandler = Substitute.For<PropertyChangedEventHandler>();
+        testSubject.PropertyChanged += eventHandler;
+        eventHandler.ReceivedCalls().Should().BeEmpty();
+
+        testSubject.BoundProject = serverProject;
+
+        eventHandler.Received().Invoke(testSubject,
```
georgii-borovinskikh-sonarsource, line 68: There needs to be a check that he event was raised for `BoundProject` alongside this existing check

georgii-borovinskikh-sonarsource (changes requested):
* add Export Shared Binding button when solution is bound
* change `Bind` and `Cancel` buttons to `OK` when solution is bound
* Disable the connection dropdown and the project selection button when solution is bound

src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
gabriela-trutan-sonarsource: I am not sure I understand what you mean.

georgii-borovinskikh-sonarsource (approved):

PR #5598, opened 2024-08-14
SLVS-1344 Unify styling for project and organization list

georgii-borovinskikh-sonarsource:


src/ConnectedMode/UI/Resources/Styles.xaml
```diff
@@ -87,4 +87,18 @@
         <Setter Property="BaselineAlignment" Value="TextBottom" />
     </Style>
 
+    <DataTemplate x:Key="NameAndKeyListBoxItem">
```
vnaskos-sonar, line 90: Not sure if the `DataTemplate` belongs to the `Styles.xaml` it feels irrelevant, can it be extracted to another file?

src/ConnectedMode/UI/Resources/Styles.xaml
```diff
@@ -87,4 +87,18 @@
...
+        <StackPanel Orientation="Horizontal">
+            <TextBlock Text="{Binding Name}" FontWeight="Bold" Margin="0,0,5,0" />
+            <TextBlock>
```
vnaskos-sonar, line 93: As you also mentioned, we can make this text gray.

```suggestion
            <TextBlock Foreground="{DynamicResource {x:Static vsShell:VsBrushes.GrayTextKey}}">
```

src/ConnectedMode/UI/Resources/Styles.xaml
georgii-borovinskikh-sonarsource: I think it's fine because this template is used for display (ie styling)

vnaskos-sonar: Not exactly, because it contains details about the layout and the data. I am not sure if this will have an impact on the other xaml files that import `Styles.xaml`. @gabriela-trutan-sonarsource ?

gabriela-trutan-sonarsource: I think it is safe to add DataTemplate to Styles.xaml. It should not impact any other file, because it has a key defined. So as long as no control references that key, it will have no impact.

vnaskos-sonar (approved):

PR #5599, opened 2024-08-14
SLVS-1348 Integrate ConnectionInfo component with Project Selection window

georgii-borovinskikh-sonarsource:


georgii-borovinskikh-sonarsource:
@gabriela-trutan-sonarsource this was the only place that was not using the new component so I renamed the task to just mention this window, rather than "integrate to all"

gabriela-trutan-sonarsource (approved):

PR #5611, opened 2024-08-20
SLVS-1371 Add listUserOrganizations to ConnectionRpcService

gabriela-trutan-sonarsource:


src/SLCore/Service/Connection/ListUserOrganizationsResponse.cs
```diff
@@ -0,0 +1,27 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Diagnostics.CodeAnalysis;
+using SonarLint.VisualStudio.SLCore.Service.Connection.Models;
+
+namespace SonarLint.VisualStudio.SLCore.Service.Connection;
+
+[ExcludeFromCodeCoverage]
```
vnaskos-sonar, line 26: We should have a test for this one as well. I believe it's kinda a convention, to check if it searilizes correctly.

src/SLCore/Service/Connection/Models/OrganizationDto.cs
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace SonarLint.VisualStudio.SLCore.Service.Connection.Models;
+
+[ExcludeFromCodeCoverage]
```
vnaskos-sonar, line 25: You can safely remove the exclusion once the `ListUserOrganizationsResponse` has a test

vnaskos-sonar (changes requested):

src/SLCore/Service/Connection/ListUserOrganizationsResponse.cs
gabriela-trutan-sonarsource: I think we should discuss this point. It feels pointless to test the JsonConvert from Newtonsoft over and over again, especially for records with simple structure.

gabriela-trutan-sonarsource: I added it. We might add a more generic test using reflection, which will make these kind of tests obsolete, but it should probably be a separate PR

vnaskos-sonar (approved):
👍

PR #5613, opened 2024-08-20
SLVS-1372 Add SLCore ValidateConnection method to ConnectionRpcService

gabriela-trutan-sonarsource:


georgii-borovinskikh-sonarsource (approved):
Looks good

PR #5614, opened 2024-08-20
SLVS-1397 Add ServerConnection model

georgii-borovinskikh-sonarsource:


src/Core.UnitTests/Binding/ServerConnectionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+[TestClass]
+public class ServerConnectionTests
+{
+    [TestMethod]
+    public void Ctor_SonarCloudOrg_HasCorrectTypeAndDefaultSettings()
+    {
+        var serverConnection = new ServerConnection("Sonar Cloud Org");
+        
+        serverConnection.Type.Should().Be(ServerConnectionType.SonarCloud);
```
gabriela-trutan-sonarsource, line 35: Why are we not checking that the id was set? Same feedback for all the other tests

src/Core/Binding/ServerConnection.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        Id = id ?? throw new ArgumentNullException(nameof(id));
+        Type = type;
+        Settings = settings ?? throw new ArgumentNullException(nameof(settings));
+    }
+}
+
+public static class ServerConnectionExtensions
+{
+    public static Uri GetSonarQubeUri(this ServerConnection connection) =>
```
gabriela-trutan-sonarsource, line 63: Why have these methods as extension methods and not as static methods? There is no gain. The serialization will simply ignore them.

gabriela-trutan-sonarsource (approved):

gabriela-trutan-sonarsource (commented):
I do have some minor feedback

src/Core/Binding/ServerConnection.cs
georgii-borovinskikh-sonarsource: You mean not as instance methods? I can move them there, sure

gabriela-trutan-sonarsource: Yes, I meant instance methods.

gabriela-trutan-sonarsource (approved):

PR #5616, opened 2024-08-21
SLVS-1374   Validate connection in CredentialsDialog

gabriela-trutan-sonarsource:
Showing in the UI the validation status is part of another PR


src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    public async Task<ValidateConnectionResponse> ValidateConnectionAsync(ConnectionInfo connectionInfo, string username, string password)
+    {
+        var validateConnectionParams = GetValidateConnectionParams(connectionInfo, GetEitherForUsernamePassword(username, password));
+        return await ValidateConnectionAsync(validateConnectionParams);
+    }
+
+    private async Task<ValidateConnectionResponse> ValidateConnectionAsync(ValidateConnectionParams validateConnectionParams)
+    {
+        await threadHandling.SwitchToBackgroundThread();
```
georgii-borovinskikh-sonarsource, line 67: use RunOnBackgroundThread

src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -0,0 +1,106 @@
...
+    private static ValidateConnectionParams GetValidateConnectionParams(ConnectionInfo connectionInfo, Either<TokenDto, UsernamePasswordDto> credentials)
+    {
+        return new ValidateConnectionParams(GetTransientConnectionDto(connectionInfo, credentials));
+    }
+
+    private static Either<TransientSonarQubeConnectionDto, TransientSonarCloudConnectionDto> GetTransientConnectionDto(ConnectionInfo connectionInfo, Either<TokenDto, UsernamePasswordDto> credentials)
+    {
+        return connectionInfo.ServerType == ConnectionServerType.SonarQube
+            ? Either<TransientSonarQubeConnectionDto, TransientSonarCloudConnectionDto>.CreateLeft(new TransientSonarQubeConnectionDto(connectionInfo.Id, credentials))
+            : Either<TransientSonarQubeConnectionDto, TransientSonarCloudConnectionDto>.CreateRight(new TransientSonarCloudConnectionDto(connectionInfo.Id, credentials));
```
georgii-borovinskikh-sonarsource, line 94: if there's no organization, we should provide `null` here. Is that the case? Or do we provide `https://sonarcloud.io` here then?

src/ConnectedMode/Resources.resx
```diff
@@ -301,4 +301,7 @@
   <data name="MuteIssuesService_Error_InsufficientPermissions" xml:space="preserve">
     <value>Credentials you have provided do not have enough permission to resolve issues. It requires the permission 'Administer Issues'.</value>
   </data>
+  <data name="ValidateCredentials_Fails" xml:space="preserve">
+    <value>[ConnectedMode] Validating credentials failed</value>
```
georgii-borovinskikh-sonarsource, line 305: Credential validation failed + maybe add more info why, as I think there's the error message in the dto

georgii-borovinskikh-sonarsource (changes requested):

src/ConnectedMode/SlCoreConnectionAdapter.cs
vnaskos-sonar: Yes this is actually an issue, it must be `null`

src/ConnectedMode/UI/Credentials/CredentialsDialog.xaml.cs
```diff
@@ -82,10 +83,27 @@ private void AuthenticationTypeCombobox_OnSourceUpdated(object sender, DataTrans
             }
         }
 
-        private void OkButton_OnClick(object sender, RoutedEventArgs e)
+        private async void OkButton_OnClick(object sender, RoutedEventArgs e)
         {
+            var isConnectionValid = await IsConnectionValidAsync();
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: IsConnectionValidAsync is already wrapped inside a try/catch. I don't think it's needed to do it here as well.

src/ConnectedMode/SlCoreConnectionAdapter.cs
gabriela-trutan-sonarsource: Here we provide https://sonarcloud.io. I have to change it to null, but in another part of the code.

src/ConnectedMode/UI/Credentials/CredentialsDialog.xaml.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: The point was about the other statements you have in this method. Setting DialogResult might throw. Not sure if we need to actually defend against that, try to assess the probability of that ever happening.

Side note: Looking at the implementation of the DialogResult setter, it seems that it already calls the Window.Close() automatically

src/ConnectedMode/Resources.resx
georgii-borovinskikh-sonarsource: Sure. But it would be still nice to see the reason in the logs as well, unless we already receive a log from slcore, in which case you can leave this log line as-is

src/ConnectedMode/SlCoreConnectionAdapter.cs
gabriela-trutan-sonarsource: I will do this on the UI PR, because it is also possible to test. Additionally, I would have to adapt the ConnectedInfoComponent to show the sonarcloud.io when the organization key is NULL

src/ConnectedMode/UI/Credentials/CredentialsDialog.xaml.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: I do not think we have to defend against that. Therefore I did not add another try/catch.

src/ConnectedMode/UI/ConnectedModeServices.cs
```diff
@@ -18,15 +18,37 @@
...
+    public IBrowserService BrowserService { get; } 
+    public IThreadHandling ThreadHandling { get; } 
+    public ILogger Logger { get; } 
+    public ISlCoreConnectionAdapter SlCoreConnectionAdapter { get; } 
+    public IConfigurationProvider ConfigurationProvider { get; } 
+}
+
+[Export(typeof(IConnectedModeServices))]
+[method: ImportingConstructor]
+public class ConnectedModeServices(
```
georgii-borovinskikh-sonarsource, line 40: Add PartCreationPolicy attribute and the Mef tests

src/ConnectedMode/Resources.resx
gabriela-trutan-sonarsource: I was missing a logging statement, which I did add it in the latest PR.

georgii-borovinskikh-sonarsource (approved):

src/ConnectedMode/UI/ConnectedModeServices.cs
georgii-borovinskikh-sonarsource: It effectively is. You only import it once. Plus, it's immutable and I think all of the property types are singletons as well. But anyway, even if a class is not single, it's still better to explicitly add PartCreationPolicy (there's NonShared value of the enum)

gabriela-trutan-sonarsource: I can make it a singleton. I have seen classes that do not declare the PartCreationPolicy at all, so I was following that style.

georgii-borovinskikh-sonarsource: We can discuss that, but IMO it's wrong. We can also add that to the xtranet page afterwards. I will create a trello card for this

gabriela-trutan-sonarsource: I made it a singleton

PR #5618, opened 2024-08-21
SLVS-1413   Show connection validation status in UI

gabriela-trutan-sonarsource:
This is PR a draft, because it is based on this [PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/5616), but it can be reviewed
Its purpose it to show the status of the connection validation in the UI. 
The quality gate issues are from the previous PR, which will be fixed.


vnaskos-sonar:
When I create a new connection and pass the correct SonarCloud token I get the following error
image
I believe we still pass the Connection.Id which is the SonarCloud URL instead of null organization to the `TransientSonarCloudConnectionDto`

vnaskos-sonar (approved):
👍

PR #5621, opened 2024-08-22
SLVS-1398 Add IServerConnectionRepository interface

georgii-borovinskikh-sonarsource:


src/Core/Binding/IServerConnectionsRepository.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Diagnostics.CodeAnalysis;
+
+namespace SonarLint.VisualStudio.Core.Binding;
+
+public interface IServerConnectionsRepository
+{
+    bool TryGet(string connectionId, out ServerConnection serverConnection);
+    List<ServerConnection> GetAll();
+    bool TryAdd(ServerConnection connection);
+    void Delete(string connectionId);
```
gabriela-trutan-sonarsource, line 31: Shouldn't the delete also be TryDelete? I think it would be nice to know if the connection was deleted successfully or not.

src/Core/Binding/IServerConnectionsRepository.cs
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+using System.ComponentModel.Composition;
+using System.Diagnostics.CodeAnalysis;
+
+namespace SonarLint.VisualStudio.Core.Binding;
+
+public interface IServerConnectionsRepository
+{
+    bool TryGet(string connectionId, out ServerConnection serverConnection);
```
gabriela-trutan-sonarsource, line 28: I think all methods should return Task

gabriela-trutan-sonarsource (approved):

src/Core/Binding/IServerConnectionsRepository.cs
georgii-borovinskikh-sonarsource: I'm not sure. In ISolutionBindingRepository all operations are synchronous. It doesn't mean that we should not change that, but is there a good reason to do so? Why do you think having async here would be worth it? Anything that touches this will also have to be async due to that

georgii-borovinskikh-sonarsource: Good point about the UI. Also agree that we can change that later if needed.

PR #5624, opened 2024-08-22
SLVS-1401 Add new BoundProject model

georgii-borovinskikh-sonarsource:


src/Core/Binding/BoundServerProject.cs
```diff
@@ -0,0 +1,39 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Core.Binding;
+
+public class BoundServerProject
+{
+    public string ProjectKey { get; }
```
gabriela-trutan-sonarsource, line 25: Is this the projectKey as in the VisualStudio solution name or the projectKey from the server? 
Should we also not add the solution name and its type, e.g., see ConnectedModeProject for reference?

src/Core/Binding/BoundServerProject.cs
```diff
@@ -0,0 +1,39 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Core.Binding;
+
+public class BoundServerProject
```
gabriela-trutan-sonarsource, line 23: This model will not be used for serialization?

gabriela-trutan-sonarsource (changes requested):

src/Core/Binding/BoundServerProject.cs
georgii-borovinskikh-sonarsource: Yes

src/Core/Binding/BoundServerProject.cs
georgii-borovinskikh-sonarsource: We don't have that in the old binding model, but it's a good point to add it. I will do that.

gabriela-trutan-sonarsource (approved):

PR #5629, opened 2024-08-23
SLVS-1394 Load organizations

gabriela-trutan-sonarsource:
[SLVS-1394](https://sonarsource.atlassian.net/browse/SLVS-1394)

This PR updates the ISlCoreConnectionAdapter to load the organizations for SonarCloud when creating a new connection.
Showing it in the UI will be a different PR

[SLVS-1394]: https://sonarsource.atlassian.net/browse/SLVS-1394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode/SlCoreConnectionAdapter.cs
+++ b/src/ConnectedMode/SlCoreConnectionAdapter.cs
@@ -47,6 +47,7 @@ public class SlCoreConnectionAdapter : ISlCoreConnectionAdapter
     private readonly ISLCoreServiceProvider serviceProvider;
     private readonly IThreadHandling threadHandling;
     private readonly ILogger logger;
+    private static readonly AdapterResponse<List<OrganizationDisplay>> FailedResponse = new(false, []);
 
     [ImportingConstructor]
     public SlCoreConnectionAdapter(ISLCoreServiceProvider serviceProvider, IThreadHandling threadHandling, ILogger logger)
@@ -70,13 +71,11 @@ public class SlCoreConnectionAdapter : ISlCoreConnectionAdapter
 
     public async Task<AdapterResponse<List<OrganizationDisplay>>> GetOrganizationsAsync(ICredentialsModel credentialsModel)
     {
-        var failedResponse = new AdapterResponse<List<OrganizationDisplay>>(false, []);
-
         return await threadHandling.RunOnBackgroundThread(async () =>
         {
-            if (TryGetConnectionConfigurationSlCoreService() is not { } connectionConfigurationSlCoreService)
+            if (!TryGetConnectionConfigurationSlCoreService(out var connectionConfigurationSlCoreService))
             {
-                return failedResponse;
+                return FailedResponse;
             }
 
             try
@@ -90,7 +89,7 @@ public class SlCoreConnectionAdapter : ISlCoreConnectionAdapter
             catch (Exception ex)
             {
                 logger.LogVerbose($"{Resources.ListUserOrganizations_Fails}: {ex.Message}");
-                return failedResponse;
+                return FailedResponse;
             }
         });
     }
@@ -99,7 +98,7 @@ public class SlCoreConnectionAdapter : ISlCoreConnectionAdapter
     {
         return await threadHandling.RunOnBackgroundThread(async () =>
         {
-            if (TryGetConnectionConfigurationSlCoreService() is not {} connectionConfigurationSlCoreService)
+            if (!TryGetConnectionConfigurationSlCoreService(out var connectionConfigurationSlCoreService))
             {
                 return new ValidateConnectionResponse(false, UiResources.ValidatingConnectionFailedText);
             }
@@ -116,15 +115,17 @@ public class SlCoreConnectionAdapter : ISlCoreConnectionAdapter
         });
     }
 
-    private IConnectionConfigurationSLCoreService TryGetConnectionConfigurationSlCoreService()
+    private bool TryGetConnectionConfigurationSlCoreService(out IConnectionConfigurationSLCoreService connectionConfigurationSlCoreService)
     {
-        if (serviceProvider.TryGetTransientService(out IConnectionConfigurationSLCoreService connectionConfigurationSlCoreService))
+        if (serviceProvider.TryGetTransientService(out IConnectionConfigurationSLCoreService slCoreService))
         {
-            return connectionConfigurationSlCoreService;
+            connectionConfigurationSlCoreService = slCoreService;
+            return true;
         }
 
+        connectionConfigurationSlCoreService = null;
         logger.LogVerbose($"[{nameof(IConnectionConfigurationSLCoreService)}] {SLCoreStrings.ServiceProviderNotInitialized}");
-        return null;
+        return false;
     }
 
     private static ValidateConnectionParams GetValidateConnectionParams(ConnectionInfo connectionInfo, Either<TokenDto, UsernamePasswordDto> credentials)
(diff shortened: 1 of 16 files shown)


src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionDialog.xaml.cs
```diff
@@ -20,15 +20,16 @@
...
 using System.Windows;
+using SonarLint.VisualStudio.ConnectedMode.UI.Credentials;
 
 namespace SonarLint.VisualStudio.ConnectedMode.UI.OrganizationSelection;
 
 [ExcludeFromCodeCoverage]
 public partial class OrganizationSelectionDialog : Window
 {
-    public OrganizationSelectionDialog(IReadOnlyList<OrganizationDisplay> organizations)
+    public OrganizationSelectionDialog(IConnectedModeServices connectedModeServices, ICredentialsModel credentialsModel)
```
georgii-borovinskikh-sonarsource, line 30: I think it was simpler when it was just accepting a static list of connections, but I also get the point of moving the responsibility here

src/ConnectedMode/UI/ManageConnections/ManageConnectionsDialog.xaml.cs
```diff
@@ -52,7 +52,8 @@ private void EditConnection_Clicked(object sender, RoutedEventArgs e)
 
         private void NewConnection_Clicked(object sender, RoutedEventArgs e)
         {
-            if (GetTransientConnection() is {} transientConnection && CredentialsDialogSucceeded(transientConnection) && GetCompleteConnection(transientConnection) is {} completeConnection)
+            if (GetTransientConnection() is {} partialConnection && GetCredentialsDialog(partialConnection) is {} credentialsDialog &&
```
georgii-borovinskikh-sonarsource, line 55: 1) formatting is incorrect here. 1 part of boolean expression per line, line breaks before operators.
2) GetCredentialsDialog can never return null, so the condition feels a bit hacky. At this point I would convert this `if (a && b && c && d)` into
```
if(!a) return;
var dialog = b();
if(!c) return;
if(connection.Type == SonarCloud && !d) return;
viewModel.add()
```

georgii-borovinskikh-sonarsource (changes requested):

src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -63,13 +68,39 @@ public async Task<ValidateConnectionResponse> ValidateConnectionAsync(Connection
         return await ValidateConnectionAsync(validateConnectionParams);
     }
 
+    public async Task<AdapterResponse<List<OrganizationDisplay>>> GetOrganizationsAsync(ICredentialsModel credentialsModel)
+    {
+        var failedResponse = new AdapterResponse<List<OrganizationDisplay>>(false, []);
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: What can be static readonly?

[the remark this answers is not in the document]
gabriela-trutan-sonarsource: The variable failedResponse, you mean?

src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionDialog.xaml.cs
gabriela-trutan-sonarsource: Simpler, maybe. But that means we would have to load the connections from the dialog, which would violate the MVVM pattern: the view should have no knowledge about the Model (slCore in our case)

src/ConnectedMode/SlCoreConnectionAdapter.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Done

src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -63,13 +68,39 @@ public async Task<ValidateConnectionResponse> ValidateConnectionAsync(Connection
...
+
+        return await threadHandling.RunOnBackgroundThread(async () =>
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: I think it is better to be explicit.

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Eh, not really. You have the async lambda inside and you could add the Async suffix to the method name. Not a big deal in the end, but it would technically be more efficient to return just the task and not await it's result here

src/ConnectedMode/UI/ManageConnections/ManageConnectionsDialog.xaml.cs
```diff
@@ -52,11 +52,18 @@ private void EditConnection_Clicked(object sender, RoutedEventArgs e)
...
-            if (GetTransientConnection() is {} partialConnection && GetCredentialsDialog(partialConnection) is {} credentialsDialog && 
-                CredentialsDialogSucceeded(credentialsDialog) && GetCompleteConnection(partialConnection, credentialsDialog) is {} completeConnection)
+            if (GetTransientConnection() is not {} transientConnection)
             {
-               ViewModel.AddConnection(new Connection(completeConnection));
+                return;
+            }
+
+            var credentialsDialog = GetCredentialsDialog(transientConnection);
+            if (!CredentialsDialogSucceeded(credentialsDialog) || GetCompleteConnection(transientConnection, credentialsDialog) is not { } completeConnection)
```
georgii-borovinskikh-sonarsource, line 61: at this point I would extract the server type check from GetCompleteConnection and rename it to FinalizeSonarCloudConnection or smth like that

georgii-borovinskikh-sonarsource (approved):

src/ConnectedMode/UI/ManageConnections/ManageConnectionsDialog.xaml.cs
gabriela-trutan-sonarsource: I would rather leave it like this. I think it is more readable and less cognitive complexity. It is, in the end, related to the final connection.

Would you want me to rename the GetCompleteConnection to FinalizeConnection? I see that SlCore also has the Transient connection vs full connection. So for me seems fine to stay as it is.

src/ConnectedMode/SlCoreConnectionAdapter.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: I removed the async

src/ConnectedMode/UI/ManageConnections/ManageConnectionsDialog.xaml.cs
gabriela-trutan-sonarsource: I renamed the method

georgii-borovinskikh-sonarsource: 1) it hides the fact that this method is only really needed to be called for SC connections, which just makes it less clear why we call it for SQ
2) both of these connections are transitive, as they need to be saved in the repository first and then synced with SLCore to be non-transitive

gabriela-trutan-sonarsource: I don't think the caller needs to know that is only needed for SC. It simply needs the final connection. Adding another (a || (b&&c) ) makes the if more complex, therefore I think the logic should be part of another method

PR #5634, opened 2024-08-26
SLVS-1373  Validate credentials in OrganizationSelectionDialog

gabriela-trutan-sonarsource:


src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionDialog.xaml.cs
```diff
@@ -20,39 +20,65 @@
...
     {
         ViewModel.SelectedOrganization = null;
         var manualOrganizationSelectionDialog = new ManualOrganizationSelectionDialog();
-        manualOrganizationSelectionDialog.Owner = this;
-        var manualSelection = manualOrganizationSelectionDialog.ShowDialog(this);
-        ViewModel.SelectedOrganization =new OrganizationDisplay(manualOrganizationSelectionDialog.ViewModel.OrganizationKey, null);
-        if (manualSelection is true)
+        var isSelectedManualOrganizationValid = await ValidateManualOrganizationAsync(manualOrganizationSelectionDialog);
+
+        if (isSelectedManualOrganizationValid)
```
georgii-borovinskikh-sonarsource, line 58: I think this code can be simplified if you remove this check and put the old check back. We would assign SelectedOrganization whichever value is selected (list or manual), and then always validate the SelectedOrganization in ViewModel in a uniform way. The error text can be made generic to not worry about mentioning manual organization selection in it. This will be even more relevant once we add the duplication check.

src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionViewModel.cs
```diff
@@ -76,4 +80,24 @@ internal void UpdateOrganizations(AdapterResponseWithData<List<OrganizationDispl
         responseWithData.ResponseData.ForEach(AddOrganization);
         RaisePropertyChanged(nameof(NoOrganizationExists));
     }
+
+    internal async Task<bool> ValidateConnectionForOrganizationAsync(string organizationKey, string warningText)
```
georgii-borovinskikh-sonarsource, line 84: Can be made private, see other comment. The warning text does not need to be parametrized as well..

georgii-borovinskikh-sonarsource (changes requested):

src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionDialog.xaml.cs
gabriela-trutan-sonarsource: I removed the assigning of the SelectedOrganization from the manual selection for two reasons:

- Once the SelectedOrganization is assigned, the OK button becomes enabled. This should not be the case, though, because the organization is not valid.  Therefore not assigning the SelectedOrganization in the first place, makes it more consistent.

- The SelectedOrganization refers to the organization selected from the list. But the manual one that was chosen is not part of the list. This brings a lot of confusion: the view model property (hence also the Combobox.SelectedItem) is set to a string, but that string is not part of the list. You will notice that WPF does not add a combobox item that is selected, if you bind its view model to a string (as long as it's not part of the list, then it should not be a selected one)

- It is more understanding for the user to have separate error messages for the manual selection or for the list selection. Imagine that the list of organizations is long. It would be very difficult for the user to know which organization was wrong, without scrolling like crazy and trying to find out if the organization was from the list or a manual one that he entered.

src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionViewModel.cs
gabriela-trutan-sonarsource: I think it does. For usability purposes.

src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionDialog.xaml.cs
georgii-borovinskikh-sonarsource: We can split the list selection from the value that is validated (the organization value then can be passed as a parameter to ViewModel.Validate). SelectedOrganization can be reset after validation fails. The error text can include the organization key. There will be no confusion if we reset selection and show the key in the error.

gabriela-trutan-sonarsource: What is the gain of setting the SelectedOrganization if we will anyway pass the organization value to the Validate method?

georgii-borovinskikh-sonarsource: So the thing is. We will need to do a few common things for both scenarious
0) catch exception for both manual and list selection 😉 
1) show duplication errors with connection id
2) set some sort of final value for the manage connections dialog to use. currently we create connectioninfo here, discard it, and then create another connecioninfo in manage connections dialog

Try to tick all 3 of those boxes with refactoring. You don't have to do it in the way I initially suggested (I though it was a bit more simple to fix)

gabriela-trutan-sonarsource: Point 1 and 2 are solved by the fact that we use the same method in the ViewModel: ValidateConnectionForOrganizationAsync(

I could work on point 3 and find a different approach

src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionDialog.xaml.cs
```diff
@@ -20,42 +20,75 @@
...
     }
 
     private async void OrganizationSelectionDialog_OnLoaded(object sender, RoutedEventArgs e)
     {
         await ViewModel.LoadOrganizationsAsync();
     }
+
+    private void OnConnectionValidationSucceeded(string organizationKey)
+    {
+        ViewModel.UpdateConnectionInfo(organizationKey);
```
georgii-borovinskikh-sonarsource, line 90: why not call this inside the validation method and use the ConnectionInfo object that was already created there?

georgii-borovinskikh-sonarsource (approved):
You can automatically assign ConnectionInfo on adapterResponse.Success. I would also keep the default value as null rather than (null, SonarCloud), to make it more clear that it hasn't been set

src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionDialog.xaml.cs
gabriela-trutan-sonarsource: Because it should only be done once the validation succeeded and the dialog is closed.

src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionViewModel.cs
```diff
@@ -35,6 +40,7 @@ public OrganizationDisplay SelectedOrganization
         set
         {
             selectedOrganization = value;
+            UpdateConnectionInfo(selectedOrganization?.Key);
```
gabriela-trutan-sonarsource, line 43: Here

georgii-borovinskikh-sonarsource (approved):

PR #5640, opened 2024-08-28
SLVS-1426 Add generic JsonSerializer

gabriela-trutan-sonarsource:
[SLVS-1426](https://sonarsource.atlassian.net/browse/SLVS-1426)

This is a preparation for the [ServerConnectionRepository task ](https://sonarsource.atlassian.net/browse/SLVS-1399)

[SLVS-1426]: https://sonarsource.atlassian.net/browse/SLVS-1426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5641, opened 2024-08-28
SLVS-1427 Add generic json file handler

gabriela-trutan-sonarsource:
[SLVS-1427](https://sonarsource.atlassian.net/browse/SLVS-1427)

This PR targets the json [serializer PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/5640) because it depends on its implementation. Both of them are preparation for the [ServerConnectionRepository task ](https://sonarsource.atlassian.net/browse/SLVS-1399)

[SLVS-1427]: https://sonarsource.atlassian.net/browse/SLVS-1427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Persistence/JsonFileHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        this.fileSystem = fileSystem;
+        this.jsonSerializer = jsonSerializer;
+        this.logger = logger;
+    }
+
+    public bool TryReadFile<T>(string filePath, out T content) where T: class
+    {
+        content = null;
+        if (!fileSystem.File.Exists(filePath))
```
georgii-borovinskikh-sonarsource, line 56: it might be more performant to just try to read regardless of if the file exists or not and handle the exception in the odd case that the file does not exist, because generally the file should exist and you will have an unnecessary file system call here

src/Core/Persistence/JsonFileHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+                {
+                    return false;
+                }
+
+                fileSystem.File.WriteAllText(filePath, serializedObj);
+
+                return true;
+
+            }
+            catch (Exception e)
```
georgii-borovinskikh-sonarsource, line 100: as this class is general purpose, we shouldn't hide critical exceptions here

src/Core/Persistence/JsonFileHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        try
+        {
+            var jsonContent = fileSystem.File.ReadAllText(filePath);
+            var wasContentDeserialized = jsonSerializer.TryDeserialize(jsonContent, out T deserializedObj);
+            content = deserializedObj;
+            return wasContentDeserialized;
+          
+        }
+        catch (Exception e)
```
georgii-borovinskikh-sonarsource, line 69: Same as the other method

src/Core/Persistence/JsonFileHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+        if (!fileSystem.File.Exists(filePath))
+        {
+            return false;
+        }
+
+        try
+        {
+            var jsonContent = fileSystem.File.ReadAllText(filePath);
+            var wasContentDeserialized = jsonSerializer.TryDeserialize(jsonContent, out T deserializedObj);
+            content = deserializedObj;
```
georgii-borovinskikh-sonarsource, line 65: Can't directly pass content?

src/Core/Persistence/JsonFileHandler.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                var directoryName = Path.GetDirectoryName(filePath);
+
+                if (!fileSystem.Directory.Exists(directoryName))
+                {
+                    fileSystem.Directory.CreateDirectory(directoryName);
+                }
+
+                var wasContentDeserialized = jsonSerializer.TrySerialize(model, out string serializedObj, Formatting.Indented);
+                if (!wasContentDeserialized)
```
georgii-borovinskikh-sonarsource, line 90: if you reverse this condition, you can put `return false` once at the end of the method, instead of having it both inside try and inside catch

georgii-borovinskikh-sonarsource (approved):
Minor comments

src/Core/Persistence/JsonFileHandler.cs
gabriela-trutan-sonarsource: I understand your point, but I don't think it's worth changing it for the following reasons:

- This generic handler might be used in more places, so even if we expect the connections.json to exist, it might not be the expected behavior for other cases
- Exception handling is an expensive operation as well, so it might be that the if check will be faster than handling exceptions
- Generally relying on exceptions for the normal flow of control is not recommended: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/exception-throwing

georgii-borovinskikh-sonarsource: 1) sure
2) exception handling is not as expensive as file system lookup, false equivalence
3) it's about design of public APIs. The Json serialization method should not rethrow it, and it would not, but it would be fine to handle by using the exception IF 1. exceptions would be rare 2. the alternative is a slow file system lookup

src/Core/Persistence/JsonFileHandler.cs
georgii-borovinskikh-sonarsource: Wdym they're not hidden? Critical exceptions should not be caught and they are caught and logged here. Use ErrorHandler.IsCriticalException

src/Core/Persistence/JsonFileHandler.cs
gabriela-trutan-sonarsource: Do you have a reference that proves that "exception handling is not as expensive as file system lookup"? I think we are squeezing performance here, which is not really needed.

It is not about designing of public APIs. It's a general guideline.

georgii-borovinskikh-sonarsource: As the 1st assumption about rarity of misses is wrong, which you pointed out correctly, you can disregard this thread

georgii-borovinskikh-sonarsource: > Do you have a reference that proves that "exception handling is not as expensive as file system lookup"? I think we are squeezing performance here, which is not really needed.
> 
> It is not about designing of public APIs. It's a general guideline.

Exception handling is just some CPU-bound work, while file system look up calls win api

src/Core/Persistence/JsonFileHandler.cs
georgii-borovinskikh-sonarsource:
```
jsonSerializer.TryDeserialize(jsonContent, out content)
```

gabriela-trutan-sonarsource: Not, it is not a valid syntax:
image

georgii-borovinskikh-sonarsource: Without the T, just `out content`

PR #5643, opened 2024-08-28
DO NOT MERGE - IServerConnectionRepository

gabriela-trutan-sonarsource:
[SLVS-1399](https://sonarsource.atlassian.net/browse/SLVS-1399)

This PR targets the [file handler PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/5641), because it depends on it. 

[SLVS-1399]: https://sonarsource.atlassian.net/browse/SLVS-1399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode/Persistence/ServerConnectionJsonModel.cs
+++ b/src/ConnectedMode/Persistence/ServerConnectionJsonModel.cs
@@ -1,0 +1,50 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2024 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Newtonsoft.Json.Converters;
+using Newtonsoft.Json;
+using SonarLint.VisualStudio.Core.Binding;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Persistence;
+
+public class ServerConnectionsListJsonModel
+{
+    [JsonProperty("serverConnections")]
+    public List<ServerConnectionJsonModel> ServerConnections { get; set; } = new();
+}
+
+public record ServerConnectionJsonModel
+{
+    [JsonProperty("id")]
+    public string Id { get; set; }
+
+    [JsonProperty("setting")]
+    public ServerConnectionSettings Settings { get; set; }
+
+    [JsonProperty("organizationKey")]
+    public string OrganizationKey { get; set; }
+
+    [JsonConverter(typeof(StringEnumConverter))]
+    [JsonProperty("serverType")]
+    public ConnectionServerType ServerType { get; set; }
+
+    [JsonProperty("serverUri")]
+    public string ServerUri { get; set; }
+}
--- a/src/ConnectedMode/Persistence/ServerConnectionJsonModel.cs
+++ b/src/ConnectedMode/Persistence/ServerConnectionJsonModel.cs
@@ -41,10 +41,6 @@ public record ServerConnectionJsonModel
     [JsonProperty("organizationKey")]
     public string OrganizationKey { get; set; }
 
-    [JsonConverter(typeof(StringEnumConverter))]
-    [JsonProperty("serverType")]
-    public ConnectionServerType ServerType { get; set; }
-
     [JsonProperty("serverUri")]
     public string ServerUri { get; set; }
 }
--- a/src/ConnectedMode/Persistence/ServerConnectionModelMapper.cs
+++ b/src/ConnectedMode/Persistence/ServerConnectionModelMapper.cs
@@ -1,0 +1,97 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2024 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Binding;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Persistence;
+
+public interface IServerConnectionModelMapper
+{
+    ServerConnection GetServerConnection(ServerConnectionJsonModel jsonModel);
+    ServerConnectionsListJsonModel GetServerConnectionsListJsonModel(IEnumerable<ServerConnection> serverConnections);
+}
+
+[Export(typeof(IServerConnectionModelMapper))]
+[PartCreationPolicy(CreationPolicy.Shared)]
+public class ServerConnectionModelMapper : IServerConnectionModelMapper
+{
+    [ImportingConstructor]
+    public ServerConnectionModelMapper() { }
+
+    public ServerConnection GetServerConnection(ServerConnectionJsonModel jsonModel)
+    {
+        return jsonModel.ServerType switch
+        {
+            ConnectionServerType.SonarCloud => new ServerConnection.SonarCloud(jsonModel.OrganizationKey, jsonModel.Settings),
+            ConnectionServerType.SonarQube => new ServerConnection.SonarQube(new Uri(jsonModel.ServerUri), jsonModel.Settings),
+            _ => throw new InvalidOperationException($"Invalid {nameof(ConnectionServerType)} {jsonModel.ServerType}")
+        };
+    }
+
+    public ServerConnectionsListJsonModel GetServerConnectionsListJsonModel(IEnumerable<ServerConnection> serverConnections)
+    {
+        var model = new ServerConnectionsListJsonModel();
+        foreach (var serverConnection in serverConnections)
+        {
+            model.ServerConnections.Add(new ServerConnectionJsonModel
+            {
+                Id = serverConnection.Id,
+                ServerType = serverConnection is ServerConnection.SonarCloud ? ConnectionServerType.SonarCloud : ConnectionServerType.SonarQube,
+                Settings = serverConnection.Settings ?? throw new InvalidOperationException($"{nameof(ServerConnection.Settings)} can not be null"),
+                OrganizationKey = GetOrganizationKey(serverConnection),
+                ServerUri = GetServerUri(serverConnection)
+            });
+        }
+
+        return model;
+    }
+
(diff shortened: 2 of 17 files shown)


src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
```diff
@@ -0,0 +1,207 @@
...
+        existingServerConnection.Settings = connectionSettings;
+        return true;
+    }
+
+    private bool SafeUpdateConnectionsFile(Func<bool> updateConnectionsCache)
+    {
+        try
+        {
+            ReadServerConnectionsFile();
+            ServerConnectionsCache ??= new ConcurrentDictionary<string, ServerConnection>();
```
georgii-borovinskikh-sonarsource, line 163: The use of ??= here is a little odd. You have the method call above that is supposed to populate the cache. But since you can't distinguish between empty list (aka file doesn't exist) and failed read (corrupted file, can't access, etc.), you fix the error here by forcefully assigning empty cache here. I would either do 1) don't care about different errors AND always initialize cache in ReadServerConnectionsFile with some value 2) actually treat no connections and failed read differently, in which case the failed read would throw and there wouldn't be a need in doing this hack with ??=

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
```diff
@@ -0,0 +1,207 @@
...
+        this.serverConnectionModelMapper = serverConnectionModelMapper;
+        this.credentialsLoader = credentialsLoader;
+        this.logger = logger;
+        storageFilePath = GetStorageFilePath(environmentVariables);
+    }
+
+    public bool TryGet(string connectionId, out ServerConnection serverConnection)
+    {
+        serverConnection = null;
+        if (ServerConnectionsCache == null)
```
georgii-borovinskikh-sonarsource, line 72: 1) Technically, this is still an unsynchronized access to a field and you may try to initialize cache from different threads
2) You can move this check inside the cache population method

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
```diff
@@ -0,0 +1,207 @@
...
+        }
+    }
+
+    private static string GetStorageFilePath(IEnvironmentVariableProvider environmentVariables)
+    {
+        var appDataFolder = environmentVariables.GetSLVSAppDataRootPath();
+        return Path.Combine(appDataFolder, ConnectionsFileName);
+    }
+
+    private void ReadServerConnectionsFile()
```
georgii-borovinskikh-sonarsource, line 188: this method is used to populate cache so the name should reflect that

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
```diff
@@ -0,0 +1,207 @@
...
+    {
+        try
+        {
+            if (!jsonFileHandle.TryReadFile(storageFilePath, out ServerConnectionsListJsonModel model))
+            {
+                return;
+            }
+
+            ServerConnectionsCache = new ConcurrentDictionary<string, ServerConnection>();
+            model.ServerConnections.ForEach(c => ServerConnectionsCache.TryAdd(c.Id, serverConnectionModelMapper.GetServerConnection(c)));
```
georgii-borovinskikh-sonarsource, line 198: race condition: try adds are not executed as a single transaction. you can fix this by using the ConcurrentDictionary(IEnumerable) ctor overload

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
```diff
@@ -0,0 +1,207 @@
...
+        }
+        catch (Exception ex)
+        {
```
georgii-borovinskikh-sonarsource, line 201: This is interesting. Why are exceptions not handled here? Also, logically, if the TryReadFile never throws, there shouldn't be any exceptions here at all. Except for bugs, but If we shouldn't defend against that

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
```diff
@@ -0,0 +1,207 @@
...
+    private void ReadServerConnectionsFile()
+    {
+        try
+        {
+            if (!jsonFileHandle.TryReadFile(storageFilePath, out ServerConnectionsListJsonModel model))
+            {
+                return;
+            }
+
+            ServerConnectionsCache = new ConcurrentDictionary<string, ServerConnection>();
```
georgii-borovinskikh-sonarsource, line 197: Multiple threads can assign this variable. Maybe Interlocked.CompareExchange would work here if you want to avoid locks, not sure. Overall it might be worth it to simplify the code by just using a normal Dictionary with a lock

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
```diff
@@ -0,0 +1,207 @@
...
+
+    private bool SafeUpdateConnectionsFile(Func<bool> updateConnectionsCache)
+    {
+        try
+        {
+            ReadServerConnectionsFile();
+            ServerConnectionsCache ??= new ConcurrentDictionary<string, ServerConnection>();
+
+            var wasUpdated = updateConnectionsCache();
+            if (!wasUpdated)
```
georgii-borovinskikh-sonarsource, line 166: invert if to reduce the number of returns

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        return ServerConnectionsCache?.Values.ToList() ?? [];
+    }
+
+    public bool TryAdd(ServerConnection connection)
+    {
+        return SafeUpdateConnectionsFile(() =>
+        {
+            var wasAdded = ServerConnectionsCache.TryAdd(connection.Id, connection);
```
georgii-borovinskikh-sonarsource, line 102: All connections added here will keep holding the credentials reference. Not sure if we want that.

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
```diff
@@ -0,0 +1,207 @@
...
+
+    public bool TryGet(string connectionId, out ServerConnection serverConnection)
+    {
+        serverConnection = null;
+        if (ServerConnectionsCache == null)
+        {
+            ReadServerConnectionsFile();
+        }
+
+        if (ServerConnectionsCache != null && ServerConnectionsCache.TryGetValue(connectionId, out ServerConnection foundConnection))
```
georgii-borovinskikh-sonarsource, line 77: Always assigning Cache/throwing on cache read fails (non related to file not being on disk) wil remove the need for `ServerConnectionsCache != null` check

src/ConnectedMode/Persistence/ServerConnectionJsonModel.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    [JsonProperty("setting")]
+    public ServerConnectionSettings Settings { get; set; }
+
+    [JsonProperty("organizationKey")]
+    public string OrganizationKey { get; set; }
+
+    [JsonConverter(typeof(StringEnumConverter))]
+    [JsonProperty("serverType")]
+    public ConnectionServerType ServerType { get; set; }
```
georgii-borovinskikh-sonarsource, line 46: This is probably not needed here. Type can be deducted from OrganizationKey/ServerUri similar to BoundServerProject.FromBoundSonarQubeProject

src/ConnectedMode/Persistence/ServerConnectionJsonModel.cs
```diff
@@ -0,0 +1,50 @@
...
+public record ServerConnectionJsonModel
+{
+    [JsonProperty("id")]
+    public string Id { get; set; }
+
+    [JsonProperty("setting")]
+    public ServerConnectionSettings Settings { get; set; }
+
+    [JsonProperty("organizationKey")]
+    public string OrganizationKey { get; set; }
```
georgii-borovinskikh-sonarsource, line 42: SonarCloudOrganizationKey?

src/ConnectedMode/Persistence/ServerConnectionJsonModel.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    [JsonConverter(typeof(StringEnumConverter))]
+    [JsonProperty("serverType")]
+    public ConnectionServerType ServerType { get; set; }
+
+    [JsonProperty("serverUri")]
+    public string ServerUri { get; set; }
```
georgii-borovinskikh-sonarsource, line 49: SonarQubeUri?

src/ConnectedMode/Persistence/ServerConnectionModelMapper.cs
```diff
@@ -0,0 +1,97 @@
...
+            _ => throw new InvalidOperationException($"Invalid {nameof(ConnectionServerType)} {jsonModel.ServerType}")
+        };
+    }
+
+    public ServerConnectionsListJsonModel GetServerConnectionsListJsonModel(IEnumerable<ServerConnection> serverConnections)
+    {
+        var model = new ServerConnectionsListJsonModel();
+        foreach (var serverConnection in serverConnections)
+        {
+            model.ServerConnections.Add(new ServerConnectionJsonModel
```
georgii-borovinskikh-sonarsource, line 55: can use list constructor instead

georgii-borovinskikh-sonarsource (changes requested):

src/ConnectedMode/Persistence/ServerConnectionJsonModel.cs
gabriela-trutan-sonarsource: We said that OrganziationKey should be enough

src/ConnectedMode/Persistence/ServerConnectionJsonModel.cs
gabriela-trutan-sonarsource: Same as SonarCloud: I wouldn't use the product name, because rebranding the product might make the name awkward.

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
gabriela-trutan-sonarsource: What do you mean? The connections added here have to have the credentials. There should never be a connection added without credentials.

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
gabriela-trutan-sonarsource: This is a good point for not having the dictionary null (your comment below regarding the null check). :)) 

But if I initialize the dictionary always, then the TryGet and GetAll will always read the file when no connections exist. Unless I introduce a flag, which I personally find ugly, but I guess is the only way.

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
georgii-borovinskikh-sonarsource: The ones that are read from disk won't have credentials loaded

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
gabriela-trutan-sonarsource: What is the problem if they are not executed as a single transaction?

georgii-borovinskikh-sonarsource: Race condition. A concurrent operation may try to access a partially filled cache

src/ConnectedMode/Persistence/ServerConnectionsRepository.cs
gabriela-trutan-sonarsource: The credentials have to be passed to be able to store them. So if they are filled, you want them to be removed when calling TryAdd? That would be a very unexpected behaviour.

Furthermore, the credentials are also filled when calling TryGet, but not when GetAll. I do not think we should change anything here.

georgii-borovinskikh-sonarsource: Yeah I'm not sure what we need to do, but I find it equally confusing that on GetAll _some_ connections have credentials and some don't

gabriela-trutan-sonarsource: I know. You said in the daily: we should do it how it was done in the SolutionBindingRepository. 

Anyway, at this point, the ServerConnection object has already credentials filled. So removing them is totally unexpected behavior.

georgii-borovinskikh-sonarsource (changes requested):
1) the more I look at it, the more I realize that using the concurrent dictionary is an overkill here and we should just use the regular lock or, if we really need to, which I doubt, use a ReaderWriterLockSlim
2) the problem of not populating the concurrent dictionary in a single transaction still persist, the problem of accessing the flag without lock still exists, which adds more to the 1st point
3) we still don't distinguish between storage not being initialized (aka no file on disk) and storage inaccessible/corrupted

I suggest to remove concurrent dictionary and use a simple lock with a dictionary. Also try to think about how would we handle the errors

PR #5649, opened 2024-08-29
SLVS-1376 Binding Dialog: list connections

gabriela-trutan-sonarsource:
[SLVS-1376](https://sonarsource.atlassian.net/browse/SLVS-1376)

To test this feature in the UI, you have to adapt the DummyServerConnectionsRepository.GetAll to return some connections instead of throwing

[SLVS-1376]: https://sonarsource.atlassian.net/browse/SLVS-1376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode/UI/Resources/UiResources.resx
+++ b/src/ConnectedMode/UI/Resources/UiResources.resx
@@ -351,4 +351,10 @@
   <data name="ValidatingOrganziationSelectionFailedText" xml:space="preserve">
     <value>The connection is not valid for the chosen organization "{0}". Make sure the entered key matches exactly your organization's key.</value>
   </data>
+  <data name="LoadingConnectionsFailedText" xml:space="preserve">
+    <value>Could not load existing connections</value>
+  </data>
+  <data name="LoadingConnectionsText" xml:space="preserve">
+    <value>Loading connections...</value>
+  </data>
 </root>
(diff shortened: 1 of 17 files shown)


src/ConnectedMode/UI/Resources/UiResources.resx
```diff
@@ -351,4 +351,10 @@
   <data name="ValidatingOrganziationSelectionFailedText" xml:space="preserve">
     <value>The connection is not valid for the chosen organization "{0}". Make sure the entered key matches exactly your organization's key.</value>
   </data>
+  <data name="LoadingConnectionsFailedText" xml:space="preserve">
+    <value>Could not load existing connections</value>
```
vnaskos-sonar, line 355: Is it possible to show a message on the dropdown, when there are no connections?
Something like "No connections found. Please consider creating a new one!"

vnaskos-sonar (approved):
👍

src/ConnectedMode/UI/Resources/UiResources.resx
gabriela-trutan-sonarsource: The combobox control does not support this, unfortunately. But we do have a TextBlock that has a placeholder text when no selection exists. I could adapt this TextBlock and disable the combobox. What do you say?

gabriela-trutan-sonarsource: I also noticed that, but I didn't do it, because I was not sure if we should invest the effort.

vnaskos-sonar: Yes, at the moment, it looks confusing when there is no connection because the dropdown is opening and has a large area (the height of the pop-up is too long). In the beginning, I thought that something was not working properly and the entries were not visible. If adding the text is too difficult, can we decrease the height of the pop up..

PR #5658, opened 2024-09-02
SLVS-1431 Add server connection json model and mapper

gabriela-trutan-sonarsource:
[SLVS-1431](https://sonarsource.atlassian.net/browse/SLVS-1431)

[SLVS-1431]: https://sonarsource.atlassian.net/browse/SLVS-1431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/Binding/ServerConnection.cs
+++ b/src/Core/Binding/ServerConnection.cs
@@ -25,7 +25,7 @@ public abstract class ServerConnection
     internal static readonly ServerConnectionSettings DefaultSettings = new(true);
     
     public string Id { get; }
-    public ServerConnectionSettings Settings { get; }
+    public ServerConnectionSettings Settings { get; set; }
     public ICredentials Credentials { get; set; }
     
     public abstract Uri ServerUri { get; }
--- a/src/ConnectedMode.UnitTests/Persistence/ServerConnectionModelMapperTest.cs
+++ b/src/ConnectedMode.UnitTests/Persistence/ServerConnectionModelMapperTest.cs
@@ -1,0 +1,239 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2024 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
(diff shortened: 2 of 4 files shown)


src/Core/Binding/ServerConnection.cs
```diff
@@ -25,7 +25,7 @@ public abstract class ServerConnection
     internal static readonly ServerConnectionSettings DefaultSettings = new(true);
     
     public string Id { get; }
-    public ServerConnectionSettings Settings { get; }
+    public ServerConnectionSettings Settings { get; set; }
```
georgii-borovinskikh-sonarsource, line 28: Why was the setter added?

georgii-borovinskikh-sonarsource (approved):

src/Core/Binding/ServerConnection.cs
gabriela-trutan-sonarsource: Because in the repository we have TryUpdateSettingsById, which requires to add the settings to an existing ServerConnection model

georgii-borovinskikh-sonarsource: I don't know if we should do that. Feels like a thread safety issue. The model can be modified from another thread. Could we make it immutable? Would that have big implications on other parts of the code?

gabriela-trutan-sonarsource: In the repository, I am reading the file and then retrieving the ServerConnection model for the provided Id. I then have to update the settings: 
image


I either allow modification of the settings or I need to create a new instance of the ServerConnection (which is immutable), then remove it from the list, then re-add the new instance. This feels a lot more error-prone and unexpected.

I could make instead the IsSmartNotificationsEnabled editable (add a setter for it). Also I don't think we should be too paranoid about the multi-threading. Because all the write operations to the connections should only happen via the UI. And the UI will be blocked while anything in the connection is modified

PR #5660, opened 2024-09-02
SLVS-1433 Add credentials uri to the ServerConnection model

gabriela-trutan-sonarsource:
[SLVS-1433](https://sonarsource.atlassian.net/browse/SLVS-1433)

This PR targets this [PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/5659), because both of them are preparation for the [server connection repository](https://sonarsource.atlassian.net/browse/SLVS-1399)

[SLVS-1433]: https://sonarsource.atlassian.net/browse/SLVS-1433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core.UnitTests/Binding/ServerConnectionTests.cs
+++ b/src/Core.UnitTests/Binding/ServerConnectionTests.cs
@@ -19,6 +19,7 @@
  */
 
 using SonarLint.VisualStudio.Core.Binding;
+using static SonarLint.VisualStudio.Core.Binding.ServerConnection;
 using ICredentials = SonarLint.VisualStudio.Core.Binding.ICredentials;
 
 namespace SonarLint.VisualStudio.Core.UnitTests.Binding;
@@ -26,8 +27,8 @@ namespace SonarLint.VisualStudio.Core.UnitTests.Binding;
 [TestClass]
 public class ServerConnectionTests
 {
-    private static readonly Uri localhost = new Uri("http://localhost:5000");
-    private static readonly string org = "myOrg";
+    private static readonly Uri Localhost = new("http://localhost:5000");
+    private const string Org = "myOrg";
     
     [TestMethod]
     public void Ctor_SonarCloud_NullOrganization_Throws()
@@ -40,7 +41,7 @@ public class ServerConnectionTests
     [TestMethod]
     public void Ctor_SonarCloud_NullSettings_SetDefault()
     {
-        var sonarCloud = new ServerConnection.SonarCloud(org, null);
+        var sonarCloud = new ServerConnection.SonarCloud(Org, null);
 
         sonarCloud.Settings.Should().BeSameAs(ServerConnection.DefaultSettings);
     }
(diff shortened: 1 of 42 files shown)


src/Core/Binding/ServerConnection.cs
```diff
@@ -29,25 +29,34 @@ public abstract class ServerConnection
...
         Credentials = credentials;
     }
-    
-    public sealed class SonarCloud(string organizationKey, ServerConnectionSettings settings = null, ICredentials credentials = null)
-        : ServerConnection(organizationKey, settings, credentials)
+
+    public sealed class SonarCloud : ServerConnection
     {
-        public string OrganizationKey { get; } = organizationKey;
+        public const string Organizations = "organizations";
```
georgii-borovinskikh-sonarsource, line 43: not sure if it was necessary to create a constant just for one usage

georgii-borovinskikh-sonarsource (approved):

src/Core/Binding/ServerConnection.cs
gabriela-trutan-sonarsource: I am also referencing it in the tests. I think it is better to be a constant. If we ever update this, than it will also be updated in the tests

src/Core.UnitTests/Binding/ServerConnectionTests.cs
```diff
@@ -40,31 +41,33 @@ public void Ctor_SonarCloud_NullOrganization_Throws()
...
+        var sonarCloud = new ServerConnection.SonarCloud(Org, serverConnectionSettings, credentials);
 
-        sonarCloud.Id.Should().BeSameAs(org);
-        sonarCloud.OrganizationKey.Should().BeSameAs(org);
+        sonarCloud.Id.Should().BeSameAs(Org);
+        sonarCloud.OrganizationKey.Should().BeSameAs(Org);
         sonarCloud.ServerUri.Should().Be(new Uri("https://sonarcloud.io"));
         sonarCloud.Settings.Should().BeSameAs(serverConnectionSettings);
         sonarCloud.Credentials.Should().BeSameAs(credentials);
+        sonarCloud.CredentialsUri.Should().Be(new Uri(expectedServerUri, $"{SonarCloud.Organizations}/{sonarCloud.OrganizationKey}"));
```
georgii-borovinskikh-sonarsource, line 70: this feels like a copy paste of the implementation. using a hardcoded string here could be safer

src/Core/Binding/ServerConnection.cs
georgii-borovinskikh-sonarsource: Should it be automatically updated in the tests? I get the argument for things like UI strings, but this feels like a value that should only change intentionally and we should strictly prohibit any unintentional changes to URIs.

gabriela-trutan-sonarsource: I don't think anybody is changing "unintentionally" constants in code. But sure, I can make it hard-coded instead (although hard-coded strings are usually against clean code principles)

PR #5665, opened 2024-09-02
SLVS-1435 Support C Language standard flags

AbbasNS:


AbbasNS:
I did a manual test using this code snipped to make sure we start to take the C standard into consideration.
Even though this PR can be merged before the CFamily upgrade to 6.58, this specific test requires CFamily 6.58 to notice the difference in behavior. 

```
// * Raises only c:S820 without /std:c17
// * Raises only c:2260 with /std:clatest
implicit_int;
```

S2260 needs to be enabled in the settings.json:

```
{
  "sonarlint.rules": {
    "c:S2260": {
      "level": "On"
    }
  }
}
```
The option can be changed in the language standard of the file:
image

georgii-borovinskikh-sonarsource (approved):

PR #5669, opened 2024-09-05
SLVS-1328 Allow users to provide custom JRE location

gabriela-trutan-sonarsource:
[SLVS-1328](https://sonarsource.atlassian.net/browse/SLVS-1328)

Allow users to provide custom JRE location via an option from Options->General page.
If the path is not provided or no java.exe can be detected, the embedded JRE will be taken. 
Check for the minimum version 17 is not performed.  

[SLVS-1328]: https://sonarsource.atlassian.net/browse/SLVS-1328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Integration.Vsix/SLCore/SLCoreLocator.cs
```diff
@@ -31,28 +33,68 @@ internal class SLCoreLocator : ISLCoreLocator
     {
         private const string DefaultPathInsideVsix = "Sloop\\";
         private readonly string basePathInsideVsix;
-        private const string JreSubPath = "jre\\bin\\java.exe";
+        private readonly ISonarLintSettings slSettings;
+        private readonly ILogger logger;
+        private readonly IFileSystem fileSystem;
+        private const string SlCoreJreSubPath = "jre\\bin\\java.exe";
+        private const string WindowsJreSubPath = "bin\\java.exe";
```
georgii-borovinskikh-sonarsource, line 40: I think we should require the full path to java.exe and not rely on the bin folder to be present

src/Integration.Vsix/SLCore/SLCoreLocator.cs
```diff
@@ -31,28 +33,68 @@ internal class SLCoreLocator : ISLCoreLocator
...
+            this.fileSystem = fileSystem;
         }
 
         public SLCoreLaunchParameters LocateExecutable()
         {
             var vsixRoot = vsixRootLocator.GetVsixRoot();
 
             //This will be changed later to jre call
-            return new (Path.Combine(vsixRoot, basePathInsideVsix, JreSubPath), 
+            return new (GetJrePath(vsixRoot),
```
georgii-borovinskikh-sonarsource, line 68: I would replace this with `GetCustormJrePath() ?? GetEmbeddedJrePath(vsixRoot)` for more clarity, as it was hard to understand from the first glance that vsixRoot was only used for the embedded case. The logging can be moved from line 76 to line 92

src/Integration.Vsix/Resources/Strings.resx
```diff
@@ -293,4 +293,21 @@ Please commit this file to share it with other contributors. See documentation f
...
+  <data name="JreLocationDescriptionText" xml:space="preserve">
+    <value>SonarLint provides its own JRE to launch SonarLint Core out of process. You can provide an explicit Java installation (17 or more recent) to be used instead.
+
+E.g. C:\Program Files\Java\jdk-17</value>
+  </data>
+  <data name="JreLocationLabel" xml:space="preserve">
+    <value>JRE Location</value>
+  </data>
+  <data name="RestartIdeToApplyOptionLabel" xml:space="preserve">
+    <value>Note: The IDE must be restarted for this setting to be applied.</value>
```
georgii-borovinskikh-sonarsource, line 311: Did you test if it works for cases when SLCore cannot start initially, the setting is changed & the restart button is clicked on the gold bar

georgii-borovinskikh-sonarsource (changes requested):

src/Integration.Vsix/Resources/Strings.resx
gabriela-trutan-sonarsource: Yes, I tested it and it works.

georgii-borovinskikh-sonarsource (approved):

PR #5682, opened 2024-09-11
SLVS-1442 Fix JRE not persisted

gabriela-trutan-sonarsource:
The problem seemed to be that the binding was not updating the value of the view model property, so when the OK button was pressed in the Options page, the previous value was overwritten


vnaskos-sonar (approved):
👍

PR #5687, opened 2024-09-12
SLVS-1445 Use SecureString for token and password

gabriela-trutan-sonarsource:
[SLVS-1445](https://sonarsource.atlassian.net/browse/SLVS-1445)

This PR targets the [creating a connection PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/5686), because it is practically an improvement of the way token/password strings were dealt with in the connections repository adapter

[SLVS-1445]: https://sonarsource.atlassian.net/browse/SLVS-1445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


georgii-borovinskikh-sonarsource (commented):
There's no need to use SecureString since later on we have to convert it to a normal string all the time we establish a server connection (http client header and slcore jsonrpc response)

gabriela-trutan-sonarsource:
@georgii-borovinskikh-sonarsource, I have introduced the SecureString for compatibility reasons. The BasicAuthCredentials requires and in src/ConnectedMode/ServerConnectionsRepositoryAdapter.cs I had to manually construct the SecureString. So either I introduce the SecureString and touch these few files, or I remove it from BasicAuthCredentials and everywhere else. I decided to implement the first option.

src/ConnectedMode/ServerConnectionsRepositoryAdapter.cs
```diff
@@ -77,18 +78,13 @@ private static ServerConnection MapConnection(Connection connection)
 
     private static ICredentials MapCredentials(ICredentialsModel credentialsModel)
     {
-        var securePassword = new SecureString();
         switch (credentialsModel)
         {
             case TokenCredentialsModel tokenCredentialsModel:
-                return new BasicAuthCredentials(tokenCredentialsModel.Token, securePassword);
+                return new BasicAuthCredentials(tokenCredentialsModel.Token.ToUnsecureString(), new SecureString());
```
vnaskos-sonar, line 84: The token takes the place of the username. Is it on purpose a secure string?

vnaskos-sonar (changes requested):

src/ConnectedMode/ServerConnectionsRepositoryAdapter.cs
gabriela-trutan-sonarsource: Yes, it was on purpose. Both the Token and password are secure strings. As Georgii pointed out, that is not for security reasons, but for compatibility. For consistency reasons, I thought the Token also makes sense. Should I invest the effort to revert it and then have the doubt in the future why one is secure and one not?

vnaskos-sonar (approved):
👍

PR #5689, opened 2024-09-13
SLVS-1381 Binding Dialog: Show current binding status

vnaskos-sonar:
[SLVS-1381](https://sonarsource.atlassian.net/browse/SLVS-1381)

To test you need to have the `connections.json` file with the connection eg.
```
{
	"serverConnections": [
		{
			"id": "duncanp-sonar-test",
			"setting": {
				"IsSmartNotificationsEnabled": true
			},
			"organizationKey": "duncanp-sonar-test"
		}
}
```

For a bound project, the binding window should display that the project is bound and all the fields should be populated according to the binding info.

[SLVS-1381]: https://sonarsource.atlassian.net/browse/SLVS-1381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
```diff
@@ -492,6 +496,29 @@ await progressReporterViewModel.Received(1)
                     x.WarningText == UiResources.LoadingConnectionsFailedText &&
                     x.AfterProgressUpdated == testSubject.OnProgressUpdated));
     }
+    
+    [TestMethod]
```
gabriela-trutan-sonarsource, line 500: This test it monstrous. Please split it into multiple tests for the following cases:

- SolutionInfo is filled for Solution
- SolutionInfo is filled for FolderWorkspace
- BoundServerProject is null
- BoundServerProject is NOT null and the response succeeded
- BoundServerProject is NOT null and the response failed

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -199,4 +221,28 @@ internal bool LoadConnections()
...
     }
+    
+    private async Task DisplayBindStatusAsync()
+    {
+        var solutionName = await solutionInfoProvider.GetSolutionNameAsync();
+        var isFolderWorkspace = await solutionInfoProvider.IsFolderWorkspaceAsync();
+        SolutionInfo = new SolutionInfoModel(solutionName, isFolderWorkspace ? SolutionType.Folder : SolutionType.Solution);
+        
+        var boundServerProject = connectedModeServices.ConfigurationProvider.GetConfiguration()?.Project;
+        var connection = boundServerProject?.ServerConnection;
```
gabriela-trutan-sonarsource, line 232: Please rename the variable connection to "serverConnection", because we also have Connection type and ConnectionInfo type... And it is very easy to confuse them all.

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -199,4 +221,28 @@ internal bool LoadConnections()
...
+        if (connection == null)
+        {
+            return;
+        }
+        
+        SelectedConnectionInfo = new ConnectionInfo(
```
gabriela-trutan-sonarsource, line 238: It would be worth extracting the creation of a ConnectionInfo from a ServerConnection into a common place and reuse it, as we do that in multiple places (ServerConnectionsRepositoryAdapter, ServerSelectionViewModel)

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -106,10 +117,21 @@ public bool IsSharedBindingConfigurationDetected
...
+        this.connectedModeServices = connectedModeServices;
+        this.solutionInfoProvider = solutionInfoProvider;
+        ProgressReporter = progressReporterViewModel;
+    }
+    
     public async Task InitializeDataAsync()
     {
         var validationParams = new TaskToPerformParams<AdapterResponse>(LoadDataAsync, UiResources.LoadingConnectionsText, UiResources.LoadingConnectionsFailedText){AfterProgressUpdated = OnProgressUpdated};
         await ProgressReporter.ExecuteTaskWithProgressAsync(validationParams);
+        await DisplayBindStatusAsync();
```
gabriela-trutan-sonarsource, line 134: It would be more user-friendly to call the DisplayBindStatusAsync also by using the ProgressReporter, because we do an operation that might take a long (communicating with SlCore).This will show the progress in the UI

src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -98,6 +103,39 @@ public Task<AdapterResponseWithData<List<OrganizationDisplay>>> GetOrganizations
...
+                if (credentials == null)
+                {
+                    return failedResponse;
+                }
+                var credentialsSlCoreFormat = credentials.Password?.Length > 0
+                    ? Either<TokenDto, UsernamePasswordDto>.CreateRight(new UsernamePasswordDto(credentials.UserName, credentials.Password.ToString()))
+                    : Either<TokenDto, UsernamePasswordDto>.CreateLeft(new TokenDto(credentials.UserName));
+                var transientConnection = GetTransientConnectionDto(connectionInfo, credentialsSlCoreFormat);
+                var response = await connectionConfigurationSlCoreService.GetProjectNamesByKeyAsync(new GetProjectNamesByKeyParams(transientConnection, [serverProjectKey]));
+                return new AdapterResponseWithData<ServerProject>(true, new ServerProject(serverProjectKey, response.projectNamesByKey[serverProjectKey]));
```
gabriela-trutan-sonarsource, line 129: When I just bind a project and the SlCore can not find the ProjectKey in the server, it returns a Dictionary where the projectKey is present, but the name is null. In that case, should the response be true or false? It is actually not found...

gabriela-trutan-sonarsource (commented):
When I just bind a project and the SlCore can not find the ProjectKey in the server, it returns a Dictionary where the projectKey is present, but the name is null (not sure actually why it couldn't find the project name). 

In that case, the UI looks like in the image below. Should we rather show the project key and not the name?

image

gabriela-trutan-sonarsource (changes requested):
Please see feedback in comment below (I pressed by mistake the "comment" and not "request changes")

src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -98,6 +103,39 @@ public Task<AdapterResponseWithData<List<OrganizationDisplay>>> GetOrganizations
...
+
+            try
+            {
+                var credentials = (BasicAuthCredentials) serverConnection.Credentials;
+                if (credentials == null)
+                {
+                    return failedResponse;
+                }
+                var credentialsSlCoreFormat = credentials.Password?.Length > 0
+                    ? Either<TokenDto, UsernamePasswordDto>.CreateRight(new UsernamePasswordDto(credentials.UserName, credentials.Password.ToString()))
```
gabriela-trutan-sonarsource, line 125: Here it should be `.ToUnsecureString()` instead of `ToString()` (that's why the ProjectName was not returned when using username/password binding)

gabriela-trutan-sonarsource (changes requested):

src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -98,6 +103,39 @@ public Task<AdapterResponseWithData<List<OrganizationDisplay>>> GetOrganizations
         });
     }
 
+    public Task<AdapterResponseWithData<ServerProject>> GetServerProjectByKeyAsync(ServerConnection serverConnection, ConnectionInfo connectionInfo, string serverProjectKey)
+    {
+        var failedResponse = new AdapterResponseWithData<ServerProject>(false, null);
```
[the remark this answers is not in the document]
vnaskos-sonar: Not possible, the FailedResponseWithData is of a different type. Here it requires `AdapterResponseWithData<ServerProject>` while the other is a `AdapterResponseWithData<List<...`.

As discussed I'll keep this one as is.

gabriela-trutan-sonarsource (approved):

PR #5685, opened 2024-09-12
SLVS-1424 Integrate ISolutionBindingRepository with BoundServerProject model

georgii-borovinskikh-sonarsource:
[SLVS-1424](https://sonarsource.atlassian.net/browse/SLVS-1424)

[SLVS-1424]: https://sonarsource.atlassian.net/browse/SLVS-1424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5690, opened 2024-09-13
SLVS-1393 Delete Connection Dialog: User should be able to delete a connectionGt/delete connection

gabriela-trutan-sonarsource:
[SLVS-1393](https://sonarsource.atlassian.net/browse/SLVS-1393)

Prevent deleting the connection when there are bindings that use that connection will be done in a separate PR, because it depends on the [PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/5685) (to get the list of bindings that contain the ConnectionId info). And that PR is still work in progress

[SLVS-1393]: https://sonarsource.atlassian.net/browse/SLVS-1393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
+++ b/src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
@@ -36,12 +36,12 @@ namespace SonarLint.VisualStudio.ConnectedMode.UI.ManageConnections
             await ProgressReporterViewModel.ExecuteTaskWithProgressAsync(validationParams);
         }
 
-        public async Task DeleteConnectionWithProgressAsync(ConnectionViewModel connectionViewModel)
+        public async Task RemoveConnectionWithProgressAsync(ConnectionViewModel connectionViewModel)
         {
             var validationParams = new TaskToPerformParams<AdapterResponse>(
                 async () => await SafeExecuteActionAsync(() => RemoveConnection(connectionViewModel)),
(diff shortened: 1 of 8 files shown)


src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
```diff
@@ -36,6 +36,15 @@ public async Task LoadConnectionsWithProgressAsync()
             await ProgressReporterViewModel.ExecuteTaskWithProgressAsync(validationParams);
         }
 
+        public async Task DeleteConnectionWithProgressAsync(ConnectionViewModel connectionViewModel)
```
vnaskos-sonar, line 39: Why not `RemoveConnectionWithProgressAsync`? 🤔  (Same question for every place that we use `delete` instead of `remove`)

vnaskos-sonar (changes requested):

src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
gabriela-trutan-sonarsource: My personal preference was Delete. I can rename it to remove for consistency reasons.

vnaskos-sonar (approved):
👍

PR #5694, opened 2024-09-17
SLVS-1388 Project Selection Dialog: User should be able to view list of projects

gabriela-trutan-sonarsource:
[SLVS-1388](https://sonarsource.atlassian.net/browse/SLVS-1388)

[SLVS-1388]: https://sonarsource.atlassian.net/browse/SLVS-1388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs
```diff
@@ -58,15 +65,34 @@ public ServerProject SelectedProject
...
+
+        return await connectedModeServices.SlCoreConnectionAdapter.GetAllProjectsAsync(ConnectionInfo, serverConnection.Credentials);
     }
 
-    public void InitProjects(List<ServerProject> projects)
+    internal void InitProjects(AdapterResponseWithData<List<ServerProject>> responseData)
     {
         ProjectResults.Clear();
-        projects.ForEach(x => ProjectResults.Add(x));
+        responseData.ResponseData.ForEach(x => ProjectResults.Add(x));
```
vnaskos-sonar, line 95: Can we sort the results?

Especially because we don't support searching at the moment it's kinda hard for the user to find their projects.

src/ConnectedMode/UI/Resources/UiResources.resx
```diff
@@ -375,4 +375,13 @@
   <data name="RemovingConnectionText" xml:space="preserve">
     <value>Removing connection...</value>
   </data>
+  <data name="LoadingProjectsFailedText" xml:space="preserve">
+    <value>Loading the projects from the server failed. </value>
```
vnaskos-sonar, line 379: Super duper extra nit-picking to cure OCD 😬
 ```suggestion
     <value>Loading the projects from the server failed.</value>
```

File: src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs [hunk 3]
```diff
@@ -19,15 +19,22 @@
  */
 
 using System.Collections.ObjectModel;
+using SonarLint.VisualStudio.ConnectedMode.UI.Resources;
 using SonarLint.VisualStudio.Core.WPF;
 
 namespace SonarLint.VisualStudio.ConnectedMode.UI.ProjectSelection;
 
-public class ProjectSelectionViewModel : ViewModelBase
+public class ProjectSelectionViewModel(
```
vnaskos-sonar, line 27: I remember that this topic was brought up in our meeting about coding conventions: that we should prefer explicit constructors.

However, I can't find this point on the repo docs 🤔

I would suggest to revert that change, until it's clear that we want to go that way.

src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -136,6 +137,13 @@ public Task<AdapterResponseWithData<ServerProject>> GetServerProjectByKeyAsync(I
         });
     }
 
+    public async Task<AdapterResponseWithData<List<ServerProject>>> GetAllProjectsAsync(ConnectionInfo connectionInfo, ICredentials credentials)
+    {
+        var credentialsDto = MapCredentials(credentials);
+        var validateConnectionParams = GetAllProjectsParams(connectionInfo, credentialsDto);
+        return await GetAllProjectsAsync(validateConnectionParams);
```
vnaskos-sonar, line 144: Create a ticket to find a solution/support more than 10000 projects

vnaskos-sonar (approved):
👍

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs
gabriela-trutan-sonarsource: Makes total sense. Good idea!

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs
gabriela-trutan-sonarsource: I don't remember this discussion. I would rather skip this comment as I think using implicit constructors is not harmful, and we are already doing it in new classes.

src/ConnectedMode/SlCoreConnectionAdapter.cs
gabriela-trutan-sonarsource: https://sonarsource.atlassian.net/browse/SLVS-1456

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs
vnaskos-sonar: Implicit constructors are not harmful, after all, it's a C# feature. It's also true that there are other places in our codebase that are used (definitely they are not the majority though) and there might be good reasons to use them.

My point here, is that the whole team should be aligned on using them or not to keep the code consistent. It is a gray area as it is not documented anywhere yet. My suggestion remains, that it's better to revert the change until we are all aligned. 🤷

gabriela-trutan-sonarsource: Sorry, @vnaskos-sonar. I did not revert the change as I was not sure if it was worth the effort. If I revert it now and then, in the future, we decide implicit constructors are fine, then it would be yet additional effort to revert it back.

Also, Resharper recommends using the implicit constructor, so having the squiggly line and ignoring it is kind of hard.

PR #5696, opened 2024-09-18
SLVS-1400 Add migration mechanism to separate connections from bindings

gabriela-trutan-sonarsource:
Implemented the migration and called it in the SonarLintDaemonPackage at the very beginning, because it should be performed as early as possible. This package is also called if solutions or folders are opened, or even none of them, which is desirable.

When the migration is performed for the first time, if you open a solution super fast, it might not actually be bound. But if you re-open the solution, then everything works as expected. 

I did not find a workaround for this issue (even triggering the binding manually did not help, I guess because it was too early). But I don't think I should invest too much effort into this, as it will only happen once: when the migration is performed, after the first update.

### DIFF
--- a/src/Integration.Vsix/SonarLintDaemonPackage.cs
+++ b/src/Integration.Vsix/SonarLintDaemonPackage.cs
@@ -88,10 +88,11 @@ namespace SonarLint.VisualStudio.Integration.Vsix
         {
             try
             {
-                await MigrateBindingsToServerConnectionsIfNeededAsync();
-
                 logger = await this.GetMefServiceAsync<ILogger>();
                 logger.WriteLine(Strings.Daemon_Initializing);
+
+                // This migration should be performed before initializing other services, independent if a solution or a folder is opened.
+                await MigrateBindingsToServerConnectionsIfNeededAsync();
 
                 await MuteIssueCommand.InitializeAsync(this, logger);
                 await DisableRuleCommand.InitializeAsync(this, logger);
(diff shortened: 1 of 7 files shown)


src/Integration.Vsix/SonarLintDaemonPackage.cs
```diff
@@ -112,6 +114,15 @@ private async Task InitAsync()
             logger?.WriteLine(Strings.Daemon_InitializationComplete);
         }
 
+        /// <summary>
+        /// This migration should be performed before initializing other services, independent if a solution or a folder is opened.
```
vnaskos-sonar, line 118: I feel that this comment should be above the method usage as it's a warning about the order.

src/Integration.Vsix/SonarLintDaemonPackage.cs
```diff
@@ -78,16 +79,17 @@ public SonarLintDaemonPackage()
...
-            JoinableTaskFactory.RunAsync(InitAsync);
-            return Task.CompletedTask;
+            await JoinableTaskFactory.RunAsync(InitAsync);
         }
 
         private async Task InitAsync()
         {
             try
             {
+                await MigrateBindingsToServerConnectionsIfNeededAsync();
```
vnaskos-sonar, line 91: This should be below the logger initialization and the initialization message.

```
[ThreadId 58] [Connection Migration] Start migrating connections from existing bindings
[ThreadId 58] [Connection Migration] The server connection with the id duncanp-sonar-test was not migrated as it already exists.
[ThreadId 58] Initializing the daemon package...
```
to 
```
[ThreadId 58] Initializing the daemon package...
[ThreadId 58] [Connection Migration] Start migrating connections from existing bindings
[ThreadId 58] [Connection Migration] The server connection with the id duncanp-sonar-test was not migrated as it already exists.
```

vnaskos-sonar (approved):
Just a small ordering comment, the other is up to your preference. 👍

PR #5700, opened 2024-09-23
SLVS-1396 Migration wizard: Integrate with the new binding and connection models

gabriela-trutan-sonarsource:
[SLVS-1396](https://sonarsource.atlassian.net/browse/SLVS-1396)

The old migration was working. I just moved the method that was dealing with migrating the ServerConnection for an old binding away from UnintrusiveBindingController and into the ConnectedModeMigration. 

Also I am making sure that if the old migration is trying to migrate a server connection and the connections.json does not exist, that we throw an error. The order of the migrations is important. 

Additional changes:
- Renamed MigrateBindingToServerConnectionIfNeededAsync to MigrateAllBindingsToServerConnectionsIfNeededAsync
- Adapt the ServerConnectionsRepository to not expose the file path, but only if the file exists. In this way, the callers do not need to know the name of the file and to check each time if the file exist, instead they ask directly the repository. 

[SLVS-1396]: https://sonarsource.atlassian.net/browse/SLVS-1396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vnaskos-sonar (changes requested):
As discussed, there is
one inconsistency: When there are no credentials on the credentials manager, previously it was succeeding to migrate, but currently it fails.
and one issue: When a user has only legacy projects, the connections.json will not be created and the migration will always be skipped.

vnaskos-sonar (approved):
👍

PR #5701, opened 2024-09-24
SLVS-1448 Prevent deleting connection when in use

gabriela-trutan-sonarsource:
[SLVS-1448](https://sonarsource.atlassian.net/browse/SLVS-1448)

I did not change the existing DeleteConnectionDialog, because it would have taken too much time and I would have needed to add all kinds of special logic, that later will be thrown away. 

Instead, I introduced a PreventDeleteConnectionDialog, which can be completely thrown away once the unbind feature is implemented. 

[SLVS-1448]: https://sonarsource.atlassian.net/browse/SLVS-1448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ConnectedMode.UnitTests/UI/ManageConnections/ManageConnectionsViewModelTest.cs
```diff
@@ -18,12 +18,15 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System;
```
vnaskos-sonar, line 21: This is part of the global using, it shouldn't be necessary here.

vnaskos-sonar (approved):
👍

PR #5718, opened 2024-09-30
SLVS-1474   Refactor ServerConnectionsProvider

gabriela-trutan-sonarsource:
Update ServerConnectionsProvider to use the ServerConnectionsRepository instead of the ISolutionBindingRepository.

Now we have a separate file for storing the server connections, so we no longer need to look at the bindings. The ServerConnectionsProvider now works as an adapter: converts the ServerConnection model to the DTO expected by slCore.

Also dropped the no longer needed ConnectionIdHelper:

The connectionId is of type URI, so we don't need anymore to add the sq| and sc|, which were a workaround needed to determine if, given a connectionId, the server is of type SonarQube or SonarCloud (which, in turn, was needed to calculate the URI needed to get the credentials form the CredentialsStore)


vnaskos-sonar (approved):
👍

PR #5722, opened 2024-10-02
SLVS-1478 Create FixSuggestion scaffolding

vnaskos-sonar:


src/SLCore/Listener/FixSuggestion/ShowFixSuggestionParams.cs
```diff
@@ -0,0 +1,25 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.SLCore.Listener.FixSuggestion.Models;
+
+namespace SonarLint.VisualStudio.SLCore.Listener.FixSuggestion;
+
+public record ShowFixSuggestionParams(string configurationScopeId, string issueKey, FixSuggestionDto fixSuggestion);
```
gabriela-trutan-sonarsource, line 25: We are missing the tests for the deserialization of the params.

src/IssueViz/FixSuggestion/IFixSuggestionHandler.cs
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.IssueVisualization.FixSuggestion;
+
+public interface IFixSuggestionHandler
+{
+    void ApplyFixSuggestion();
```
gabriela-trutan-sonarsource, line 25: We should pass the parameters to this method. 
It's OK if you think it should not be done in this task. Then we can do it in the implementation task.

gabriela-trutan-sonarsource (changes requested):

src/IssueViz/FixSuggestion/IFixSuggestionHandler.cs
vnaskos-sonar: Adding them to the later task is better, so we are sure what's necessary.

src/SLCore/Listener/FixSuggestion/Models/FileEditDto.cs
```diff
@@ -0,0 +1,25 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.SLCore.Common.Models;
+
+namespace SonarLint.VisualStudio.SLCore.Listener.FixSuggestion.Models;
+
+public record FileEditDto(FileUri idePath, List<ChangesDto> changes);
```
gabriela-trutan-sonarsource, line 25: I don't think FileUri is serializable. It should probably be string.

gabriela-trutan-sonarsource (changes requested):

src/SLCore/Listener/FixSuggestion/Models/FileEditDto.cs
vnaskos-sonar: There is another FileEditDto which in the Java side had it as a URI. However, in this FileEditDto, the java side is a Path. Makes sense to change it to string 😉

gabriela-trutan-sonarsource (approved):

PR #5727, opened 2024-10-04
SLVS-1481 Show message when code does not match

gabriela-trutan-sonarsource:
[SLVS-1481](https://sonarsource.atlassian.net/browse/SLVS-1481)

[SLVS-1481]: https://sonarsource.atlassian.net/browse/SLVS-1481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/IssueViz/FixSuggestion/FixSuggestionHandler.cs
```diff
@@ -19,15 +19,167 @@
...
+    private bool ValidateConfiguration(string configurationScopeId, out string configurationScopeRoot, out string failureReason)
+    {
+        return openInIdeConfigScopeValidator.TryGetConfigurationScopeRoot(configurationScopeId, out configurationScopeRoot, out failureReason);
+    }
+
+    private void ApplyAndShowAppliedFixSuggestions(ShowFixSuggestionParams parameters, string configurationScopeRoot)
+    {
+        var absoluteFilePath = Path.Combine(configurationScopeRoot, parameters.fixSuggestion.fileEdit.idePath);
+        var textView = GetFileContent(parameters, absoluteFilePath);
+        if (!ValidateFileExists(textView, absoluteFilePath))
```
vnaskos-sonar, line 117: Minor:
The `GetFileContent` and the `ValidateFileExists` should most of the times be called together. They could be combined in a single method, as we also do for the `ValidateConfiguration`.
Something like: `ValidateFileExists(absoluteFilePath, out var textView)`

src/IssueViz/FixSuggestion/FixSuggestionHandler.cs
```diff
@@ -19,15 +19,167 @@
...
+    private ITextView GetFileContent(ShowFixSuggestionParams parameters, string absoluteFilePath)
     {
-        throw new NotImplementedException();
+        try
+        {
+            return documentNavigator.Open(absoluteFilePath);
+        }
+        catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+        {
+            logger.WriteLine(Resources.ERR_OpenDocumentException, parameters.fixSuggestion.fileEdit.idePath, ex.Message);
```
vnaskos-sonar, line 181: We could log the `absoluteFilePath` as in the notification in the above lines.

vnaskos-sonar (approved):

src/IssueViz/FixSuggestion/FixSuggestionHandler.cs
gabriela-trutan-sonarsource: I disagree. Getting the file content and validating if it exists should be separate, otherwise it violates the SRP (single responsibility principle).

The same with ValidateFileExists(absoluteFilePath, out var textView). The out parameter is already an indicator that the method does more than it should: validating and returning something. 

I will leave them separate for now.

PR #5730, opened 2024-10-07
SLVS-1491 Refactor: use ConnectedModeManager

gabriela-trutan-sonarsource:
Refactoring: use ConnectedModeManager in the other places where we instantiate the ManageBindingDialog to make code better testable


src/Integration.UnitTests/MefServices/SharedBindingSuggestionServiceTests.cs
```diff
@@ -127,6 +130,20 @@ public void Dispose_UnsubscribesFromActiveSolutionChanged()
...
 
+    [TestMethod]
+    public void ActiveSolutionChanged_SolutionIsOpened_ShowsGoldBarAndShowManageBindingDialog()
+    {
+        MockSharedBindingConfigExists();
+        MockSolutionMode(SonarLintMode.Standalone);
+        suggestSharedBindingGoldBar.When(x => x.Show(ServerType.SonarQube, Arg.Any<Action>())).Do(callInfo =>
+        {
+            callInfo.Arg<Action>()();
+            connectedModeManager.Received(1).ShowManageBindingDialog(true);
```
georgii-borovinskikh-sonarsource, line 141: Not sure if it's a good idea to put assertions inside setup code. If you wanted to assert the order of calls, you can use Received.InOrder(...)

georgii-borovinskikh-sonarsource (approved):

src/Integration.UnitTests/MefServices/SharedBindingSuggestionServiceTests.cs
gabriela-trutan-sonarsource: I don't want to test the order, but I want to know what happens when the action is called. So, I really have to store the action and call it. I can refactor this to make it more explicit.

PR #5754, opened 2024-10-16
SLVS-1522 IOpenInIdeNotification: Use INotificationService to remove duplicated code

vnaskos-sonar:
[SLVS-1522](https://sonarsource.atlassian.net/browse/SLVS-1522)

Part of SLVS-1508

[SLVS-1522]: https://sonarsource.atlassian.net/browse/SLVS-1522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/IssueViz/OpenInIde/OpenInIdeNotification.cs
```diff
@@ -37,35 +34,60 @@ internal interface IOpenInIdeNotification
...
-        infoBar.ShowAsync(toolWindowId, message, hasMoreInfo).Forget();
+
+        var moreInfoAction = new NotificationAction(OpenInIdeResources.InfoBar_Button_MoreInfo, _ => browserService.Navigate(DocumentationLinks.OpenInIdeIssueLocation), false);
+        var showLogsAction = new NotificationAction(OpenInIdeResources.InfoBar_Button_ShowLogs, _ => outputWindowService.Show(), false);
+        List<INotificationAction> actions = hasMoreInfo
+            ? [moreInfoAction, showLogsAction]
+            : [showLogsAction];
+
+        var notification = new Notification(
+            id: "OpenInIdeNotification",
```
gabriela-trutan-sonarsource, line 87: Please extract hard-coded string into a constant

gabriela-trutan-sonarsource (changes requested):
Minor requests

src/IssueViz/OpenInIde/OpenInIdeNotification.cs
gabriela-trutan-sonarsource: It is significant for the class behavior because it represents the id for the OpenInIde notification. The notification in this class should always use that and only that id.

Additionally:

1. It is cleaner. 
2. It is more performant. Having a constant that is defined once in the lifetime of the application instead of every time for the new notification instance brings a tiny performance improvement.

I incline for the reason 1, though. It is always cleaner to have constants instead of hard-coded strings

gabriela-trutan-sonarsource (approved):

PR #5756, opened 2024-10-16
SLVS-1524 Add Solution name tracking to ActiveSolutionTracker

georgii-borovinskikh-sonarsource:
[SLVS-1524](https://sonarsource.atlassian.net/browse/SLVS-1524)

Part of SLVS-1496

[SLVS-1524]: https://sonarsource.atlassian.net/browse/SLVS-1524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Integration/MefServices/ActiveSolutionBoundTracker.cs
+++ b/src/Integration/MefServices/ActiveSolutionBoundTracker.cs
@@ -56,7 +56,6 @@ namespace SonarLint.VisualStudio.Integration
         public event EventHandler<ActiveSolutionBindingEventArgs> SolutionBindingChanged;
         public event EventHandler PreSolutionBindingUpdated;
         public event EventHandler SolutionBindingUpdated;
-
         public BindingConfiguration CurrentConfiguration { get; private set; }
 
         [ImportingConstructor]
@@ -169,7 +168,7 @@ namespace SonarLint.VisualStudio.Integration
                 SolutionBindingChanged?.Invoke(this, args);
             }
             else if (isBindingCleared == false)
-            {
+            { // todo remove unreachable code & cleanup https://sonarsource.atlassian.net/browse/SLVS-1532
                 PreSolutionBindingUpdated?.Invoke(this, EventArgs.Empty);
                 SolutionBindingUpdated?.Invoke(this, EventArgs.Empty);
             }
--- a/src/Integration/MefServices/ActiveSolutionTracker.cs
+++ b/src/Integration/MefServices/ActiveSolutionTracker.cs
@@ -36,9 +36,12 @@ namespace SonarLint.VisualStudio.Integration
     internal sealed class ActiveSolutionTracker : IActiveSolutionTracker, IFolderWorkspaceMonitor, IVsSolutionEvents, IDisposable, IVsSolutionEvents7
     {
         private readonly IFolderWorkspaceService folderWorkspaceService;
+        private readonly ISolutionInfoProvider solutionInfoProvider;
         private bool isDisposed;
         private readonly IVsSolution solution;
         private readonly uint cookie;
+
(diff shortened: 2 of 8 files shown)


src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -60,6 +57,7 @@ internal sealed class ActiveSolutionBoundTracker : IActiveSolutionBoundTracker,
         public event EventHandler PreSolutionBindingUpdated;
         public event EventHandler SolutionBindingUpdated;
 
+        public string CurrentSolutionName { get; private set; }
```
vnaskos-sonar, line 60: This is only getting `set` and never `get`. It can be removed, the source of truth for the CurrentSolutionName should be the `ActiveSolutionTracker`

src/Integration/MefServices/ActiveSolutionTracker.cs
```diff
@@ -36,10 +36,13 @@ namespace SonarLint.VisualStudio.Integration
     internal sealed class ActiveSolutionTracker : IActiveSolutionTracker, IFolderWorkspaceMonitor, IVsSolutionEvents, IDisposable, IVsSolutionEvents7
     {
         private readonly IFolderWorkspaceService folderWorkspaceService;
+        private readonly ISolutionInfoProvider solutionInfoProvider;
         private bool isDisposed;
         private readonly IVsSolution solution;
         private readonly uint cookie;
 
+        public string CurrentSolutionName { get; set; }
```
vnaskos-sonar, line 44: The setter can be `private`

src/Integration/MefServices/ActiveSolutionTracker.cs
```diff
@@ -170,7 +174,16 @@ private void RaiseSolutionChangedEvent(bool isSolutionOpen)
         {
             // Note: if lightweight solution load is enabled then the solution might not
             // be fully opened at this point
-            ActiveSolutionChanged?.Invoke(this, new ActiveSolutionChangedEventArgs(isSolutionOpen));
+            if (isSolutionOpen)
```
vnaskos-sonar, line 177: The `if` is not necessary, the solutionInfoProvider is taking care to return null in case there isn't an open solution.

src/Integration.UnitTests/MefServices/VsSessionHostTests.cs
```diff
@@ -255,7 +255,7 @@ public void VsSessionHost_ResetBinding_NoOpenSolutionScenario()
             this.stepRunner.AbortAllNumberOfCalls.Should().Be(0);
 
             // Act
-            tracker.SimulateActiveSolutionChanged(isSolutionOpen: false);
```
vnaskos-sonar, line 258: Just a note: VsSessionHost has been deleted on the `hardening-8-6`.

vnaskos-sonar (changes requested):

src/Integration/MefServices/ActiveSolutionTracker.cs
georgii-borovinskikh-sonarsource: Optimization. We don't need to call SolutionInfoProvider. It will save us 1 thread check and vs hierarchy lookup

src/Integration.UnitTests/MefServices/VsSessionHostTests.cs
georgii-borovinskikh-sonarsource: Oh it has? That's great. Maybe I should rebase.

vnaskos-sonar (approved):
👍

PR #5757, opened 2024-10-17
SLVS-1517 Implement IEmbeddedAnalyzerProvider

gabriela-trutan-sonarsource:
This targets the [SLVS-1514 PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/5755), because the location to the embedded analyzers changed in that PR, so the provider should look in the updated location.

### DIFF
--- a/src/Infrastructure.VS.UnitTests/Infrastructure.VS.UnitTests.csproj
+++ b/src/Infrastructure.VS.UnitTests/Infrastructure.VS.UnitTests.csproj
@@ -9,6 +9,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="3.11.0" />
+  </ItemGroup>
+
+  <ItemGroup>
     <ProjectReference Include="..\Infrastructure.VS\Infrastructure.VS.csproj" />
     <ProjectReference Include="..\TestInfrastructure\TestInfrastructure.csproj" />
   </ItemGroup>
--- a/src/Infrastructure.VS/Roslyn/EmbeddedRoslynAnalyzerProvider.cs
+++ b/src/Infrastructure.VS/Roslyn/EmbeddedRoslynAnalyzerProvider.cs
@@ -1,0 +1,66 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2024 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.IO.Abstractions;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.Infrastructure.VS.Roslyn;
+
+[Export(typeof(IEmbeddedRoslynAnalyzerProvider))]
+[PartCreationPolicy(CreationPolicy.Shared)]
+public class EmbeddedRoslynAnalyzerProvider : IEmbeddedRoslynAnalyzerProvider
+{
+    private readonly IEmbeddedRoslynAnalyzersLocator locator;
+    private readonly IAnalyzerAssemblyLoaderFactory analyzerAssemblyLoaderFactory;
+    private readonly ILogger logger;
+    private readonly IFileSystem fileSystem;
+
+    [ImportingConstructor]
+    public EmbeddedRoslynAnalyzerProvider(IEmbeddedRoslynAnalyzersLocator locator, ILogger logger) :
+        this(locator, new AnalyzerAssemblyLoaderFactory(), logger, new FileSystem())
+    {
+    }
+
+    internal EmbeddedRoslynAnalyzerProvider(IEmbeddedRoslynAnalyzersLocator locator,
+        IAnalyzerAssemblyLoaderFactory analyzerAssemblyLoaderFactory,
+        ILogger logger,
+        IFileSystem fileSystem)
+    {
+        this.locator = locator;
+        this.analyzerAssemblyLoaderFactory = analyzerAssemblyLoaderFactory;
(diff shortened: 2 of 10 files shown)


src/Infrastructure.VS/Roslyn/EmbeddedRoslynAnalyzerProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+        IFileSystem fileSystem)
+    {
+        this.locator = locator;
+        this.analyzerAssemblyLoaderFactory = analyzerAssemblyLoaderFactory;
+        this.logger = logger;
+        this.fileSystem = fileSystem;
+    }
+
+    public ImmutableArray<AnalyzerFileReference>? Get()
+    {
```
georgii-borovinskikh-sonarsource, line 56: The return value of this method can be cached

src/Infrastructure.VS/Roslyn/EmbeddedRoslynAnalyzerProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+        var analyzerPaths = fileSystem.Directory.GetFiles(locator.GetPathToParentFolder());
```
georgii-borovinskikh-sonarsource, line 57: Maybe move the directory get files to the locator?

src/Infrastructure.VS/Roslyn/EmbeddedRoslynAnalyzerProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+        if(analyzerPaths.Length == 0)
+        {
+            logger.WriteLine(Resources.EmbeddedRoslynAnalyzersNotFound);
+            return null;
+        }
+        var loader = analyzerAssemblyLoaderFactory.Create();
```
georgii-borovinskikh-sonarsource, line 63: Not caching the assembly loader may be a problem due to this https://github.com/dotnet/roslyn/blob/a2b5cbb70c371628330a368e61d7a5cb02f72bb4/src/Compilers/Core/Portable/DiagnosticAnalyzer/AnalyzerFileReference.cs#L81

Can be solved by caching the embedded analyzers

src/Infrastructure.VS/Roslyn/IEmbeddedRoslynAnalyzerProvider.cs
```diff
@@ -26,7 +26,7 @@ namespace SonarLint.VisualStudio.Infrastructure.VS.Roslyn;
 public interface IEmbeddedRoslynAnalyzerProvider
 {
     /// <summary>
-    /// Returns SonarAnalyzer.CSharp & SonarAnalyzer.VisualBasic analyzer DLLs that are embedded in the VSIX
+    /// Returns SonarAnalyzer.CSharp & SonarAnalyzer.VisualBasic analyzer DLLs that are embedded in the VSIX or null if no analyzers are found.
```
georgii-borovinskikh-sonarsource, line 29: Not sure it's a good idea. I'm leaning towards throwing. Plus, as it is a contract change, you should've discussed that prior to committing to it

src/Integration.Vsix/Integration.Vsix.csproj
```diff
@@ -82,12 +82,20 @@
   <Import Project="..\EmbeddedSonarAnalyzer.props" />
   <ItemGroup Label="Embedded SonarLint analyzers">
     <!-- Reference the packages we want to import, but don't include them as analyzers for this project -->
-    <PackageReference Include="SonarAnalyzer.CSharp" Version="$(EmbeddedSonarAnalyzerVersion)" ExcludeAssets="all" PrivateAssets="all" />
-    <PackageReference Include="SonarAnalyzer.VisualBasic" Version="$(EmbeddedSonarAnalyzerVersion)" ExcludeAssets="all" PrivateAssets="all" />
+    <PackageReference Include="SonarAnalyzer.CSharp" Version="$(EmbeddedSonarAnalyzerVersion)" ExcludeAssets="all" PrivateAssets="all">
```
georgii-borovinskikh-sonarsource, line 85: rebase?

src/Infrastructure.VS.UnitTests/Infrastructure.VS.UnitTests.csproj
```diff
@@ -8,6 +8,10 @@
     <AssemblyName>SonarLint.VisualStudio.Infrastructure.VS.UnitTests</AssemblyName>
   </PropertyGroup>
 
+  <ItemGroup>
```
georgii-borovinskikh-sonarsource, line 11: Would `<RequireLanguageServices>true</RequireLanguageServices>`  solve the problem? https://github.com/SonarSource/sonarlint-visualstudio/blob/1fa7a8f9fa7497ee9f6ab09c482993773cac02da/src/SonarLint.VSSpecificAssemblies.props#L108

georgii-borovinskikh-sonarsource (changes requested):
Mainly requesting changes to caching the values

src/Infrastructure.VS/Roslyn/EmbeddedRoslynAnalyzerProvider.cs
gabriela-trutan-sonarsource: Why should the provider cache the value and not the factory? I think the factory should be responsible for that.

georgii-borovinskikh-sonarsource: `Create()` implies new instance. If you want the factory to cache it (which is fine), it should be either `Instance { get; }` or be more explicit with the method name like `GetShared()`

gabriela-trutan-sonarsource: Then we should do that. But it is the responsibility of the factory. Because whichever class calls the factory, the behaviour should be the same: the loader should be cached. 
So I would take this feedback out of this PR

georgii-borovinskikh-sonarsource: Yeah a new subtask

src/Infrastructure.VS/Roslyn/IEmbeddedRoslynAnalyzerProvider.cs
georgii-borovinskikh-sonarsource: They're inherently different. While for the CM provider `null` is an acceptable value (when there's no analyzer), for the embedded one it's an exceptional case that, IMO, should throw (there should be no reason to not have embedded analyzers). Moving responsibility for the error handling to a different class actually makes it more complicated

src/Infrastructure.VS/Roslyn/EmbeddedRoslynAnalyzerProvider.cs
gabriela-trutan-sonarsource: https://sonarsource.atlassian.net/browse/SLVS-1529

src/Infrastructure.VS.UnitTests/Infrastructure.VS.UnitTests.csproj
gabriela-trutan-sonarsource: It doesn't. I have a reference to IAnalyzerAssemblyLoader and to AnalyzerFileReference, which needs the reference to Microsoft.CodeAnalysis

georgii-borovinskikh-sonarsource (approved):
👍

PR #5758, opened 2024-10-17
Update LanguageServices to 3.11, reference it in Infrastructure.VS

georgii-borovinskikh-sonarsource:
Part of SLVS-1496


vnaskos-sonar (approved):
👍

PR #5769, opened 2024-10-23
SLVS-1541 Fix secrets analysis not working in connected mode for SQ 9.9.1

gabriela-trutan-sonarsource:
[SLVS-1541](https://sonarsource.atlassian.net/browse/SLVS-1541)

[SLVS-1541]: https://sonarsource.atlassian.net/browse/SLVS-1541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
+++ b/src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
@@ -18,12 +18,13 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.IO;
 using System.IO.Abstractions;
-using System.Linq;
+using System.Text.RegularExpressions;
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Integration.Vsix.Helpers;
+using SonarLint.VisualStudio.Integration.Vsix.Resources;
 using SonarLint.VisualStudio.SLCore.Configuration;
 
 namespace SonarLint.VisualStudio.Integration.Vsix.SLCore;
@@ -33,16 +34,20 @@ namespace SonarLint.VisualStudio.Integration.Vsix.SLCore;
 public class SLCoreEmbeddedPluginJarLocator : ISLCoreEmbeddedPluginJarLocator
 {
     private const string JarFolderName = "DownloadedJars";
+
+    private readonly Dictionary<string, string> connectedModePluginNamePatternByPluginKey = new(){ {"text", "sonar-text-plugin-(\\d+\\.\\d+\\.\\d+\\.\\d+)\\.jar"}};
     private readonly IVsixRootLocator vsixRootLocator;
     private readonly IFileSystem fileSystem;
+    private readonly ILogger logger;
 
     [ImportingConstructor]
-    public SLCoreEmbeddedPluginJarLocator(IVsixRootLocator vsixRootLocator) : this(vsixRootLocator, new FileSystem()) { }
+    public SLCoreEmbeddedPluginJarLocator(IVsixRootLocator vsixRootLocator, ILogger logger) : this(vsixRootLocator, new FileSystem(), logger) { }
 
-    internal SLCoreEmbeddedPluginJarLocator(IVsixRootLocator vsixRootLocator, IFileSystem fileSystem)
+    internal SLCoreEmbeddedPluginJarLocator(IVsixRootLocator vsixRootLocator, IFileSystem fileSystem, ILogger logger)
     {
         this.vsixRootLocator = vsixRootLocator;
         this.fileSystem = fileSystem;
+        this.logger = logger;
     }
 
     public List<string> ListJarFiles()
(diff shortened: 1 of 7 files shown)


src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
```diff
@@ -33,16 +34,24 @@ namespace SonarLint.VisualStudio.Integration.Vsix.SLCore;
 public class SLCoreEmbeddedPluginJarLocator : ISLCoreEmbeddedPluginJarLocator
 {
     private const string JarFolderName = "DownloadedJars";
+
+    private readonly Dictionary<string, string> connectedModePluginNamePatternByPluginKey = new()
+    {
+        { "text", "sonar-text-plugin-(\\d+\\.\\d+\\.\\d+\\.\\d+)\\.jar" },
+        { "javascript", "sonar-javascript-plugin-(\\d+\\.\\d+\\.\\d+\\.\\d+)\\.jar" }
```
georgii-borovinskikh-sonarsource, line 41: Why include js? Didn't we agree we want to use the one from the server?

src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
```diff
@@ -33,16 +34,24 @@ namespace SonarLint.VisualStudio.Integration.Vsix.SLCore;
 public class SLCoreEmbeddedPluginJarLocator : ISLCoreEmbeddedPluginJarLocator
 {
     private const string JarFolderName = "DownloadedJars";
+
+    private readonly Dictionary<string, string> connectedModePluginNamePatternByPluginKey = new()
+    {
+        { "text", "sonar-text-plugin-(\\d+\\.\\d+\\.\\d+\\.\\d+)\\.jar" },
```
georgii-borovinskikh-sonarsource, line 40: Probably would be simpler to understand if it was just `StartsWith` instead of regex. List jar files already returns only the jars and the version pattern is redundant at that point

src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
```diff
@@ -55,4 +64,36 @@ public List<string> ListJarFiles()
...
+    private string GetPathByPluginKey(List<string> pluginFilePaths, string pluginKey, string pluginNameRegexPattern)
+    {
+        var regex = new Regex(pluginNameRegexPattern);
+        var matchedFilePaths = pluginFilePaths.Where(jar => regex.IsMatch(jar)).ToList();
+        switch (matchedFilePaths.Count)
+        {
+            case 0:
+                logger.LogVerbose(Strings.ConnectedModeEmbeddedPluginJarLocator_JarsNotFound);
+                break;
+            case > 1:
```
georgii-borovinskikh-sonarsource, line 93: Should we care about these case?  Is it realistic for any of these scenarios to happen? `If`s also would look more compact than the old switch statement

src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
```diff
@@ -55,4 +64,36 @@ public List<string> ListJarFiles()
...
+                logger.LogVerbose(Strings.ConnectedModeEmbeddedPluginJarLocator_MultipleJars, pluginKey);
+                break;
+        }
+        return matchedFilePaths.FirstOrDefault();
```
georgii-borovinskikh-sonarsource, line 97: This will return null for the `0` case.

georgii-borovinskikh-sonarsource (changes requested):

src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
gabriela-trutan-sonarsource: Because the analysis is not working for typescript/javascript neither. This time I can see that the javascript embedded plugin from SlCore is used, but it doesn't seem to work at all. Not sure why

src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
gabriela-trutan-sonarsource: I think it is not redundant, but safer. If in some point we will have a sonar-text-plugin-enterprise-[*], it will match exactly the expected one. Therefore I will not change that

src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
gabriela-trutan-sonarsource: That is intended. Therefore no plugin will be added

src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
georgii-borovinskikh-sonarsource: Is it not working for SQ9.9 only? Would be nice to know the reason as well, because now we're forcing a worse UX for all users while not knowing exactly if it's the same problem or not

src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
gabriela-trutan-sonarsource: I think it is better to be secure, yes.

src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
gabriela-trutan-sonarsource: For SQ 10.x it is working. Why is it worse UX ?

georgii-borovinskikh-sonarsource (approved):

PR #5782, opened 2024-10-28
SLVS-1562 Refactor ActiveConfigScopeTrackerTests to use Nsubstitute

gabriela-trutan-sonarsource:
[SLVS-1562](https://sonarsource.atlassian.net/browse/SLVS-1562)

CaYC: refactor ActiveConfigScopeTrackerTests to use Nsubstitute and extract commin initialization logic inside TestInitialize

Part of SLVS-1497

[SLVS-1562]: https://sonarsource.atlassian.net/browse/SLVS-1562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5783, opened 2024-10-29
SLVS-1563 Raise CurrentConfigurationScopeChanged when configuration scope changed

gabriela-trutan-sonarsource:
[SLVS-1563](https://sonarsource.atlassian.net/browse/SLVS-1563)

This is a draft, because it targets the [PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/5782), which is needed for the tests. It can still be reviewed. 

[SLVS-1563]: https://sonarsource.atlassian.net/browse/SLVS-1563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SLCore.UnitTests/State/ActiveConfigScopeTrackerTests.cs
+++ b/src/SLCore.UnitTests/State/ActiveConfigScopeTrackerTests.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.ComponentModel;
 using SonarLint.VisualStudio.Core;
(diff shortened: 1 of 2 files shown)


georgii-borovinskikh-sonarsource (commented):
Looks good, except that raising the event under lock can be problematic. I think we need to move it outsider the synchronization blocks. But if we do that, we shouldn't have the configscope value in the event parameter to prevent race conditions (callers will need to call `IActiveConfigScopeTracker.Current` to get the up to date value)

georgii-borovinskikh-sonarsource (approved):

PR #5795, opened 2024-11-04
SLVS-1385 Manage Connections Dialog: User should be able to edit credentials

vnaskos-sonar:
[SLVS-1385](https://sonarsource.atlassian.net/browse/SLVS-1385)

Expectations:
* A connection should be re-triggered when the user edits the credentials
* If any error occurs, show an error message at the bottom of the dialog

To test:
1. Bind a solution with a token
2. Close the solution
3. Revoke the token
4. Open the solution again (expect that the binding is broken)
5. Update with a new token (expect that the solution is bound and you are able to view for example taints)

[SLVS-1385]: https://sonarsource.atlassian.net/browse/SLVS-1385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
+++ b/src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
@@ -21,6 +21,7 @@
 using System.Collections.ObjectModel;
 using SonarLint.VisualStudio.ConnectedMode.UI.Credentials;
 using SonarLint.VisualStudio.ConnectedMode.UI.Resources;
+using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.Core.WPF;
 
 namespace SonarLint.VisualStudio.ConnectedMode.UI.ManageConnections
@@ -95,14 +96,29 @@ namespace SonarLint.VisualStudio.ConnectedMode.UI.ManageConnections
                 UiResources.CreatingConnectionFailedText);
             await ProgressReporterViewModel.ExecuteTaskWithProgressAsync(validationParams);
         }
-        
+
         internal async Task UpdateConnectionCredentialsWithProgressAsync(Connection connection, ICredentialsModel credentialsModel)
         {
             var validationParams = new TaskToPerformParams<AdapterResponse>(
                 async () => await SafeExecuteActionAsync(() => UpdateConnectionCredentials(connection, credentialsModel)),
                 UiResources.UpdatingConnectionCredentialsProgressText,
                 UiResources.UpdatingConnectionCredentialsFailedText);
-            await ProgressReporterViewModel.ExecuteTaskWithProgressAsync(validationParams);
+            var response = await ProgressReporterViewModel.ExecuteTaskWithProgressAsync(validationParams);
+
+            if (!response.Success)
+            {
+                return;
+            }
+
+            var boundServerProject = connectedModeServices.ConfigurationProvider.GetConfiguration()?.Project;
+            if (boundServerProject != null && ConnectionInfo.From(boundServerProject.ServerConnection).Id == connection.Info.Id)
+            {
+                var refreshBinding = new TaskToPerformParams<AdapterResponse>(
+                    async () => await RebindAsync(connection, boundServerProject.ServerProjectKey),
+                    UiResources.RebindingProgressText,
+                    UiResources.RebindingFailedText);
+                await ProgressReporterViewModel.ExecuteTaskWithProgressAsync(refreshBinding);
+            }
         }
 
         internal async Task<AdapterResponse> SafeExecuteActionAsync(Func<bool> funcToExecute)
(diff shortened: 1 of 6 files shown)


src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -146,20 +151,26 @@ private async Task UpdateConnectionAsync(BindingConfiguration bindingConfigurati
...
 
-            var boundProject = bindingConfiguration.Project;
-
-            if (boundProject != null)
+            if (!bindingConfiguration.Mode.IsInAConnectedMode())
             {
-                var connectionInformation = boundProject.CreateConnectionInformation();
-                await WebServiceHelper.SafeServiceCallAsync(() => sonarQubeService.ConnectAsync(connectionInformation,
-                    CancellationToken.None), this.logger);
+                return true;
```
gabriela-trutan-sonarsource, line 156: This return feels very awckard inside this method. 
Why not having it in the caller method like so:

` if (!newBindingConfiguration.Mode.IsInAConnectedMode())
 {
     var updateConnection = await UpdateConnectionAsync(newBindingConfiguration);
     if (!updateConnection)
     {
         CurrentConfiguration = BindingConfiguration.Standalone;
         return;
     }
 }
`

We would still probably have to call the disconnect SonarQube

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -123,8 +122,14 @@ private async void OnActiveSolutionChanged(object sender, ActiveSolutionChangedE
...
                 var newBindingConfiguration = GetBindingConfiguration();
-                
-                await UpdateConnectionAsync(newBindingConfiguration);
+
+                var updateConnection = await UpdateConnectionAsync(newBindingConfiguration);
+
+                if (!updateConnection)
+                {
+                    CurrentConfiguration = BindingConfiguration.Standalone;
+                    return;
```
gabriela-trutan-sonarsource, line 131: I would document the decision behind this if

gabriela-trutan-sonarsource (changes requested):

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
vnaskos-sonar: It is closer to the context of the UpdateConnectionAsync method, it is part of updating the connection.

gabriela-trutan-sonarsource: It doesn't feel so for me. Because we are saying that "yes" the connection was updated, although it is in standalone. Feels very, very confusing. 
My proposal made it more explicit when we should connect.

src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
```diff
@@ -156,10 +172,30 @@ internal bool UpdateConnectionCredentials(Connection connection, ICredentialsMod
...
         }
 
+        internal async Task<AdapterResponse> RebindAsync(Connection connection, string serverProjectKey)
+        {
+            if (!connectedModeServices.ServerConnectionsRepositoryAdapter.TryGet(connection.Info, out var serverConnection))
+            {
+                return new AdapterResponse(false);
+            }
+
+            try
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Why on another PR? :( We introduced the code duplication right here...

src/Integration/MefServices/ActiveSolutionBoundTracker.cs
```diff
@@ -146,20 +151,26 @@ private async Task UpdateConnectionAsync(BindingConfiguration bindingConfigurati
...
-            if (boundProject != null)
+            if (!bindingConfiguration.Mode.IsInAConnectedMode())
             {
-                var connectionInformation = boundProject.CreateConnectionInformation();
-                await WebServiceHelper.SafeServiceCallAsync(() => sonarQubeService.ConnectAsync(connectionInformation,
-                    CancellationToken.None), this.logger);
+                return true;
             }
+
+            var boundProject = bindingConfiguration.Project;
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: That is assuming that the code follows the expected flow (e.g. that we call the CreateBoundConfiguration). But if the constructor of the BindingConfiguration is called, this is not guaranteed.

We can't know ahead how the runtime execution changes, so having the null check seemed safer

gabriela-trutan-sonarsource (changes requested):
Can we, please, remove the code duplication from the view model?

gabriela-trutan-sonarsource (approved):

PR #5815, opened 2024-11-08
SLVS-1589 Update RaisedFindingDto to support new MQR/Std mode with overrides from SLCore

gabriela-trutan-sonarsource:
[SLVS-1589](https://sonarsource.atlassian.net/browse/SLVS-1589)

[SLVS-1589]: https://sonarsource.atlassian.net/browse/SLVS-1589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5819, opened 2024-11-08
SLVS-1612   Update the RaiseFindingToAnalysisIssueConverter

gabriela-trutan-sonarsource:
This PR targets [the PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/5815) in which the RaisedFindingDto was updated, because the implementation of the RaiseFindingToAnalysisIssueConverter has to be updated due to that change


src/SLCore.UnitTests/Listener/Analysis/RaiseFindingToAnalysisIssueConverterTests.cs
```diff
@@ -268,7 +236,7 @@ public void GetAnalysisIssues_HotspotHasNoVulnerabilityProbability_AnalysisHotsp
                 default,
                 null,
                 HotspotStatus.SAFE,
-                default)
+                Either<StandardModeDetails, MQRModeDetails>.CreateLeft(new(default, default)))
```
georgii-borovinskikh-sonarsource, line 239: btw, https://github.com/SonarSource/sonarlint-visualstudio/commit/eb24cf22eafb9a400da6b361a8edba8d21eabf88 has added implicit casts from Left/Right to Either<Left, Right>

src/SLCore.UnitTests/Listener/Analysis/RaiseIssuesParamsTests.cs
```diff
@@ -83,15 +79,6 @@ [new ImpactDto(SoftwareQuality.MAINTAINABILITY, ImpactSeverity.LOW)])))
                          }
                        ]
                      },
-                    "severity": "BLOCKER",
```
georgii-borovinskikh-sonarsource, line 86: This is interesting. We SHOULD ignore these fields, however they've only been deprecated on the SLCore side, so they're still being sent. Should we have a test where we explicitly assert that we ignore these values?:

georgii-borovinskikh-sonarsource (approved):

src/SLCore.UnitTests/Listener/Analysis/RaiseIssuesParamsTests.cs
gabriela-trutan-sonarsource: Good point! Yes, sure. I will add one.

PR #5831, opened 2024-11-13
SLVS-1627 Graceful shutdown: Prevent exception when calling detached COM object

vnaskos-sonar:
When exiting Visual Studio we often get an Exception because SLCore is sending log messages for shutdown, but the log level settings have already been disposed and the COM object has been detached from VS.

### DIFF
--- a/src/Integration.Vsix/Settings/SonarLintSettings.cs
+++ b/src/Integration.Vsix/Settings/SonarLintSettings.cs
@@ -26,12 +26,14 @@ namespace SonarLint.VisualStudio.Integration.Vsix
 {
     [Export(typeof(ISonarLintSettings))]
     [PartCreationPolicy(CreationPolicy.Shared)]
-    internal class SonarLintSettings : ISonarLintSettings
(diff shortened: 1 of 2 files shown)


src/Integration.Vsix/Settings/SonarLintSettings.cs
```diff
@@ -20,96 +20,91 @@
...
+    }
 
-        internal /* testing purposes */ void SetValue(string key, string value)
+    public string JreLocation
+    {
+        get => GetValueOrDefault(nameof(JreLocation), string.Empty);
+        set => SetValue(nameof(JreLocation), value);
+    }
+
+    public void Dispose() => disposed = true;
```
georgii-borovinskikh-sonarsource, line 62: What calls this?

georgii-borovinskikh-sonarsource (approved):

src/Integration.Vsix/Settings/SonarLintSettings.cs
vnaskos-sonar: Visual Studio

PR #5874, opened 2024-12-05
SLVS-1404 Implement DeleteBinding method in SolutionBindingRepository

gabriela-trutan-sonarsource:
[SLVS-1404](https://sonarsource.atlassian.net/browse/SLVS-1404)

[SLVS-1404]: https://sonarsource.atlassian.net/browse/SLVS-1404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode/Persistence/PersistenceStrings.resx
+++ b/src/ConnectedMode/Persistence/PersistenceStrings.resx
@@ -122,6 +122,6 @@
     <comment>Output window message in case of deserialization error. {0} file path.</comment>
   </data>
   <data name="BindingDirectoryNotDeleted" xml:space="preserve">
-    <value>The binding configuration file '{0}' does not exist. Binding directory not deleted.</value>
+    <value>The Connected Mode configuration file '{0}' does not exist. Connected Mode configuration not deleted.</value>
   </data>
 </root>
--- a/src/ConnectedMode/Persistence/PersistenceStrings.resx
+++ b/src/ConnectedMode/Persistence/PersistenceStrings.resx
@@ -121,4 +121,7 @@
     <value>Failed loading file: '{0}'. Please rebind solution to create a new configuration file.</value>
     <comment>Output window message in case of deserialization error. {0} file path.</comment>
   </data>
+  <data name="BindingDirectoryNotDeleted" xml:space="preserve">
+    <value>The binding configuration file '{0}' does not exist. Binding directory not deleted.</value>
+  </data>
 </root>
--- a/src/ConnectedMode/Persistence/SolutionBindingRepository.cs
+++ b/src/ConnectedMode/Persistence/SolutionBindingRepository.cs
@@ -100,7 +100,11 @@ internal class SolutionBindingRepository : ISolutionBindingRepository, ILegacySo
         return true;
     }
 
-    public bool DeleteBinding(string configFilePath) => throw new NotImplementedException();
+    public bool DeleteBinding(string localBindingKey)
+    {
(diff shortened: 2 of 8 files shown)


src/ConnectedMode/Persistence/SolutionBindingRepository.cs
```diff
@@ -100,7 +100,11 @@ public bool Write(string configFilePath, BoundServerProject binding)
         return true;
     }
 
-    public bool DeleteBinding(string configFilePath) => throw new NotImplementedException();
+    public bool DeleteBinding(string localBindingKey)
+    {
+        var bindingPath = unintrusiveBindingPathProvider.GetBindingPath(localBindingKey);
```
georgii-borovinskikh-sonarsource, line 105: Not required, but a question. [Suppressions](https://github.com/SonarSource/sonarlint-visualstudio/blob/f16c3a1f8dfab8ade4df0af546a0dc2a9d976a4d/src/Roslyn.Suppressions/Roslyn.Suppressions/SuppressionExecutionContext.cs#L34) check for binding to be present and do nothing if there's no connected mode. We delete the binding here, and the suppressions will stop working (well, there might be issues with roslyn caching stuff, but that's not caused by any of this). But we still keep the suppressions file in `AppData\Local\Temp\SLVS\Roslyn`. Could you check if there are no unintended consequences on rebind? The file should get replaced instantly, but still it would be good to double check.

src/ConnectedMode/Persistence/PersistenceStrings.resx
```diff
@@ -121,4 +121,7 @@
     <value>Failed loading file: '{0}'. Please rebind solution to create a new configuration file.</value>
     <comment>Output window message in case of deserialization error. {0} file path.</comment>
   </data>
+  <data name="BindingDirectoryNotDeleted" xml:space="preserve">
+    <value>The binding configuration file '{0}' does not exist. Binding directory not deleted.</value>
```
georgii-borovinskikh-sonarsource, line 125: I would prefer using `Connected Mode configuration`. `Binding` is a bit of an internal name IMO

georgii-borovinskikh-sonarsource (approved):

src/ConnectedMode/Persistence/SolutionBindingRepository.cs
gabriela-trutan-sonarsource: I tested with the Suppressions_CSharp.cs file from our sample projects the following cases:

Case 1:
1. Bind to Samples_Bound_2019 (suppressions are there) 
2. Unbind (suppressions are removed)
3. Bind again to Samples_Bound_2019 (suppressions are there) 

=> This case seems to be working correctly

Case 2:
1. Bind to Samples_Bound_2019 (suppressions are there) 
2. Unbind (suppressions are removed)
3. Bind to another project that is NOT Samples_Bound_2019 and that does not have the suppressions in this file

=> The old suppressions seem to be applied. But they disappear as soon as analysis of the file is triggered (reopening the file)

This seems to be caused by the fact that the old file is there for a brief period of time before being replaced. Should we also consider deleting the file from %tmp%?

georgii-borovinskikh-sonarsource: yeah I think it would be safer to delete it as well. you can move it to another PR

PR #5880, opened 2024-12-09
SLVS-1377 Implement unbind in Binding Dialog

gabriela-trutan-sonarsource:
[SLVS-1377](https://sonarsource.atlassian.net/browse/SLVS-1377)

[SLVS-1377]: https://sonarsource.atlassian.net/browse/SLVS-1377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode/UI/Resources/Styles.xaml
+++ b/src/ConnectedMode/UI/Resources/Styles.xaml
@@ -30,7 +30,7 @@
         <Setter Property="Padding" Value="5,2"/>
         <Setter Property="Foreground" Value="{DynamicResource {x:Static platformUi:CommonControlsColors.ButtonTextBrushKey}}" />
         <Setter Property="Background" Value="{DynamicResource {x:Static platformUi:CommonControlsColors.ButtonBrushKey}}" />
-        <Setter Property="BorderBrush" Value="{DynamicResource {x:Static platformUi:CommonControlsColors.ButtonBorderBrushKey}}" />
+        <Setter Property="BorderBrush" Value="{DynamicResource {x:Static platformUi:CommonControlsColors.ButtonBorderFocusedBrushKey}}" />
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="Button">


src/ConnectedMode/Binding/IUnintrusiveBindingController.cs
```diff
@@ -67,6 +70,17 @@ public async Task BindAsync(BoundServerProject project, IProgress<FixedStepsProg
             await bindingProcess.SaveServerExclusionsAsync(token);
         }
 
+        public bool Unbind(string localBindingKey)
```
georgii-borovinskikh-sonarsource, line 73: this could be a separate PR, but you can leave it as-is

src/ConnectedMode/UI/ManageBinding/ManageBindingDialog.xaml
```diff
@@ -90,7 +90,7 @@
                         </Grid.ColumnDefinitions>
 
                         <Button Grid.Column="0" Content="{x:Static res:UiResources.UnbindButton}" IsEnabled="{Binding Path=IsUnbindButtonEnabled}"
```
georgii-borovinskikh-sonarsource, line 92: Discussed on slack: the button is a bit hard to see in the dark theme (same as the shared binding button)

src/ConnectedMode/UI/ManageBinding/ManageBindingDialog.xaml
georgii-borovinskikh-sonarsource, line 1: Discussed on slack: showing the Cancel button after unbind doesn't make sense. An easy fix would be to rename it to Close in general.

src/ConnectedMode/Binding/IUnintrusiveBindingController.cs
```diff
@@ -67,6 +70,17 @@ public async Task BindAsync(BoundServerProject project, IProgress<FixedStepsProg
...
+        {
+            var bindingDeleted = solutionBindingRepository.DeleteBinding(localBindingKey);
+            if (bindingDeleted)
+            {
+                sonarQubeService.Disconnect();
```
georgii-borovinskikh-sonarsource, line 78: https://github.com/SonarSource/sonarlint-visualstudio/blob/e0afc290d353c61961bf1fa9e5756372aa3c2ad6/src/Integration/MefServices/ActiveSolutionBoundTracker.cs#L119 <- this method can already handle the disconnect. I get why you're not using it, but maybe we need to generalize that for both solution change and unbind. Also it has a call to `gitEventsMonitor.Refresh();`, not sure if we need to call that here too (or in the new method)

src/ConnectedMode/Binding/IUnintrusiveBindingController.cs
georgii-borovinskikh-sonarsource: Maybe, considering a potential change to ASBT, it could actually deserve its own PR

georgii-borovinskikh-sonarsource (commented):
Overall looks good, but the controller/asbt interaction potentially needs to be changed

src/ConnectedMode/Binding/IUnintrusiveBindingController.cs
georgii-borovinskikh-sonarsource: This follows the current pattern of managing the sonar qube client in the controller, so I'm resolving this for now

georgii-borovinskikh-sonarsource (approved):

PR #5890, opened 2024-12-12
SLVS-1669 Handle the case of C/C++ files outside the projects

vnaskos-sonar:
[SLVS-1669](https://sonarsource.atlassian.net/browse/SLVS-1669)

[SLVS-1669]: https://sonarsource.atlassian.net/browse/SLVS-1669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SLCore/SLCoreStrings.resx
+++ b/src/SLCore/SLCoreStrings.resx
@@ -178,6 +178,6 @@
     <value>[CertificateChainValidator] {0}: {1}</value>
   </data>
   <data name="CompilationDatabaseNotFound" xml:space="preserve">
-    <value>[SLCoreAnalyzer] No compilation database found for file: {0}</value>
+    <value>[SLCoreAnalyzer] No compilation database found for file: {0}. Check that the file is part of a supported project type in the current solution.</value>
   </data>
 </root>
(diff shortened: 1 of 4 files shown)


src/SLCore/SLCoreStrings.resx
```diff
@@ -177,4 +177,7 @@
   <data name="CertificateValidator_FailureReasonTemplate" xml:space="preserve">
     <value>[CertificateChainValidator] {0}: {1}</value>
   </data>
+  <data name="CompilationDatabaseNotFound" xml:space="preserve">
+    <value>[SLCoreAnalyzer] No compilation database found for file: {0}</value>
```
michael-jabbour-sonarsource, line 181: Quick question: Would it make sense to add a hint about the file being possibly not associated with any project/solution for VCX? e.g. similar to what we used to have here: https://github.com/SonarSource/sonarlint-visualstudio/blob/3191dcd33da21ffd1f769d0a5dd70490572c68c2/src/CFamily/Subprocess/CFamilyStrings.resx#L182-L185

I could see that for the CMake case, we already have a very helpful error message :+1: 
https://github.com/SonarSource/sonarlint-visualstudio/blob/37eb38b0475c8f9098d78f2efedd9d79eea28630/src/CFamily/CMake/CMakeCompilationDatabaseLocator.cs#L96
https://github.com/SonarSource/sonarlint-visualstudio/blob/37eb38b0475c8f9098d78f2efedd9d79eea28630/src/CFamily/CMake/Resources.resx#L144-L147

georgii-borovinskikh-sonarsource (approved):

PR #5913, opened 2024-12-19
SLVS-1723 Introduce log contexts

georgii-borovinskikh-sonarsource:
[SLVS-1723](https://sonarsource.atlassian.net/browse/SLVS-1723)

[SLVS-1723]: https://sonarsource.atlassian.net/browse/SLVS-1723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Logging/LoggerBase.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Text;
+
+namespace SonarLint.VisualStudio.Core.Logging;
+
+internal class LoggerBase(
+    ILoggerContextManager contextManager,
+    ILoggerWriter writer,
+    ILoggerSettingsProvider settingsProvider) : ILogger
+{
+    public ILogger ForContext(params string[] context) =>
```
gabriela-trutan-sonarsource, line 31: Creating a new instance for each context feels like a very bad design decision as this will increase the memory usage and the performance overhead.

src/Core/Logging/LoggerContextManager.cs
```diff
@@ -0,0 +1,55 @@
...
+        this.contexts = contexts;
+        this.verboseContexts = verboseContexts;
+        formatedContext = new Lazy<string>(() => MergeContextsIntoSingleProperty(contexts), LazyThreadSafetyMode.PublicationOnly);
+        formatedVerboseContext = new Lazy<string>(() => MergeContextsIntoSingleProperty(verboseContexts), LazyThreadSafetyMode.PublicationOnly);
+    }
+
+    public string FormatedContext => formatedContext.Value;
+    public string FormatedVerboseContext => formatedVerboseContext.Value;
+
+    public ILoggerContextManager CreateAugmentedContext(IEnumerable<string> additionalContexts) => new LoggerContextManager(contexts.AddRange(additionalContexts), verboseContexts);
```
gabriela-trutan-sonarsource, line 50: Creating a new instance for each context feels like a very bad design decision as this will increase the memory usage and the performance overhead.

gabriela-trutan-sonarsource (changes requested):
I am not really happy with the design decisions as we are creating a lot of instances of the logger and the context, which will increase the memory usage and the performance overhead especially because logging is a very used operation.

I would rather make the LoggerBase and LoggerContextManager Mef exportable components with the creation policy non-shared.
Then the contextManager of a logger can be configured to hold a specific context or to enhance existing contexts.

src/Core/Logging/LoggerBase.cs
georgii-borovinskikh-sonarsource: This is not a correct conclusion. The instances are lightweight (partially thanks to immutable lists and lazy string property initialization, but in general they don't hold a lot of data apart from a few object references and the aforementioned property list and formatted string). This will not increase the overhead in any meaningful way.

src/Core/Logging/LoggerBase.cs
```diff
@@ -0,0 +1,94 @@
...
+            AppendProperty(builder, contextManager.FormatedVerboseContext);
+        }
+
+        return builder;
+    }
+
+    private static StringBuilder AppendProperty(StringBuilder builder, string property) =>
+        builder.Append('[').Append(property).Append(']').Append(' ');
+
+    private static StringBuilder AppendPropertyFormat(StringBuilder builder, string property, params object[] args) =>
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: What's the motivation for that?

src/Core/Logging/LoggerBase.cs
georgii-borovinskikh-sonarsource: On top of that, most of these instances will be created in user-class constructors and therefore will be long-living.

gabriela-trutan-sonarsource: It is a correct conclusion and it will bring more pressure to the garbage collector, which will impact the performance.

> On top of that, most of these instances will be created in user-class constructors and therefore will be long-living.
This is correct for the LoggerBase instances. That's why I see no reason why it shouldn't be a Mef component with a non-shared creation policy (which will basically lead to the same thing but without the need of factories).
This is not true, however for the LoggerContextManager and I think we can find better design than just creating instances for each context.

georgii-borovinskikh-sonarsource: Again, that's not true. It will, of course impact the performance, as pretty much every single added allocation or instruction does, but the question is will it impact it in any meaningful way? The answer is not. It has less of an impact on memory than the existing string allocations inside the logger.

georgii-borovinskikh-sonarsource: Uncustomized logger context manager has references to 2 empty immutable lists and 2 non-initialized lazies...

src/Core/Logging/LoggerBase.cs
```diff
@@ -0,0 +1,94 @@
...
+        builder.Append('[').AppendFormat(property, args).Append(']').Append(' ');
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Technically this is faster and has less allocations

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Chars about 2x faster than double formatting and uses less memory. I don't think it's that unreadable.

src/Core.UnitTests/Logging/LoggerBaseTests.cs
```diff
@@ -0,0 +1,359 @@
...
+    public void LogVerboseWithContext_AllContextsEnabled_AddsInCorrectOrder()
+    {
+        var messageLevelContext = new MessageLevelContext
+        {
+            Context = Substitute.For<IReadOnlyCollection<string>>(),
+            VerboseContext = Substitute.For<IReadOnlyCollection<string>>()
+        };
+        settingsProvider.IsThreadIdEnabled.Returns(true);
+        settingsProvider.IsVerboseEnabled.Returns(true);
+        contextManager.GetFormattedContextOrNull(default).Returns("context");
```
gabriela-trutan-sonarsource, line 174: The lines 174-175 are not needed for the test

src/Core.UnitTests/Logging/LoggerBaseTests.cs
```diff
@@ -0,0 +1,359 @@
...
+    public void WriteLineFormattedWithContext_AllContextsEnabled_AddsInCorrectOrder()
+    {
+        var messageLevelContext = new MessageLevelContext
+        {
+            Context = Substitute.For<IReadOnlyCollection<string>>(),
+            VerboseContext = Substitute.For<IReadOnlyCollection<string>>()
+        };
+        settingsProvider.IsThreadIdEnabled.Returns(true);
+        settingsProvider.IsVerboseEnabled.Returns(true);
+        contextManager.GetFormattedContextOrNull(default).Returns("context");
```
gabriela-trutan-sonarsource, line 350: The lines 350-351 are not needed for the test

src/Core/Logging/ILogger.cs
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+namespace SonarLint.VisualStudio.Core;
+
+public interface ILogger
+{
+    /// <summary>
+    /// Logs a message and appends a new line.
+    /// </summary>
+    void WriteLine(string message);
```
gabriela-trutan-sonarsource, line 28: Let's get rid of this overload to have less clutter in the interface

src/Core/Logging/LoggerBase.cs
```diff
@@ -0,0 +1,95 @@
...
+            writer,
+            settingsProvider);
+
+    public ILogger ForVerboseContext(params string[] context) =>
+        new LoggerBase(
+            contextManager.CreateAugmentedVerboseContext(context.Where(x => !string.IsNullOrEmpty(x))),
+            writer,
+            settingsProvider);
+
+    public void WriteLine(string message) =>
```
gabriela-trutan-sonarsource, line 44: We could actually get rid of this overload.

src/Core/Logging/ILogger.cs
```diff
@@ -0,0 +1,47 @@
...
+    void WriteLine(string messageFormat, params object[] args);
+    void WriteLine(MessageLevelContext context, string messageFormat, params object[] args);
+
+    /// <summary>
+    /// Logs a message and appends a new line if logging is set to verbose. Otherwise does nothing.
+    /// </summary>
+    void LogVerbose(string messageFormat, params object[] args);
+    void LogVerbose(MessageLevelContext context, string messageFormat, params object[] args);
+
+    ILogger ForContext(params string[] context);
```
gabriela-trutan-sonarsource, line 38: Consider renaming these two methods (lines 38 and 40) with something like GetNewWithContext, CreateNewWithContext.

gabriela-trutan-sonarsource (approved):
I still have some suggestions, but the implementation seems better now

src/Core/Logging/ILoggerContextManager.cs
```diff
@@ -18,19 +18,14 @@
...
+public interface ILoggerContextManager
 {
-    public interface ILogger
-    {
-        /// <summary>
-        /// Logs a message and appends a new line.
-        /// </summary>
-        void WriteLine(string message);
-        void WriteLine(string messageFormat, params object[] args);
+    public string GetFormattedContextOrNull(MessageLevelContext messageLevelContext);
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: If none of the contexts are available, it returns null

src/Core/Logging/ILogger.cs
georgii-borovinskikh-sonarsource: technically this overload is faster than the format overload, because string/sb format checks for `{0}` in the string even if the parameters list is empty

src/Core/Logging/ILoggerContextManager.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: https://github.com/SonarSource/sonarlint-visualstudio/blob/55280f054b444ba31a0802ecc6c305dd5455b668/src/Core.UnitTests/Logging/LoggerContextManagerTests.cs#L108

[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Yes, but not in the case I mentioned. So I think the name could be improved. Or the documentation of the method

src/Core/Logging/StringBuilderLoggingExtensions.cs
```diff
@@ -18,27 +18,15 @@
...
+{
+    public static StringBuilder AppendProperty(this StringBuilder builder, string property) =>
+        builder.Append('[').Append(property).Append(']').Append(' ');
 
-        public void LogVerbose(string messageFormat, params object[] args)
-        {
-            WriteLine(messageFormat, args);
-        }
-    }
+    public static StringBuilder AppendPropertyFormat(this StringBuilder builder, string property, params object[] args) =>
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Log properties also include thread id, debug, possibly other things. Context is a word I used to specifically describe the log properties that are set for a certain logical scope

src/Core/Logging/ILogger.cs
georgii-borovinskikh-sonarsource: I think it's enough that it has a return type of ILogger, that implies that it's a new instance

src/Core/Logging/ILoggerContextManager.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: GetCombinedContextOrNull?

src/Core/Logging/LoggerFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+using System.ComponentModel.Composition;
+
+namespace SonarLint.VisualStudio.Core.Logging;
+
+[Export(typeof(ILoggerFactory))]
+[PartCreationPolicy(CreationPolicy.NonShared)]
+[method: ImportingConstructor]
+public class LoggerFactory(ILoggerContextManager loggerContextManager) : ILoggerFactory
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: LoggerContextManager implementation is not accessible to Container.cs (as it is an internal class). I think this is a better solution than making the class public. Also it means that users of LoggerBase don't need to mef-export the manager and it also allows to use Substitute (by returning ILogger)

PR #5924, opened 2024-12-24
SLVS-1722 Log extra properties of SLCore log events

georgii-borovinskikh-sonarsource:
[SLVS-1722](https://sonarsource.atlassian.net/browse/SLVS-1722)

[SLVS-1722]: https://sonarsource.atlassian.net/browse/SLVS-1722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SLCore/Listener/Logger/LogParams.cs
```diff
@@ -26,6 +26,12 @@ namespace SonarLint.VisualStudio.SLCore.Listener.Logger;
 public class LogParams
 {
     public string message;
+    // nullable
+    public string configScopeId;
+    public string threadName;
+    public string loggerName;
+    // nullable
+    public string stackTrace;
```
gabriela-trutan-sonarsource, line 34: We are missing the `loggedAt` property. Even if we don't use it for now, I think we should add it.

src/SLCore.Listeners/Implementation/LoggerListener.cs
```diff
@@ -34,25 +34,26 @@ public class LoggerListener : ILoggerListener
     [ImportingConstructor]
     public LoggerListener(ILogger logger)
     {
-        this.logger = logger;
+        this.logger = logger.ForContext(SLCoreStrings.SLCoreName, SLCoreStrings.SLCoreInternalLogContext);
```
gabriela-trutan-sonarsource, line 37: I think it's enough to add just one context that is a string of type "SlCore Logs". It would occupies less space than [SLCore > Logs].

src/SLCore.Listeners/Implementation/LoggerListener.cs
```diff
@@ -34,25 +34,26 @@ public class LoggerListener : ILoggerListener
...
-            case LogLevel.INFO:
-            case LogLevel.DEBUG:
-            case LogLevel.TRACE:
-                logger.LogVerbose(message);
+            case LogLevel.INFO or LogLevel.DEBUG or LogLevel.TRACE:
+                contextualizedLogger.LogVerbose(parameters.message);
                 break;
         }
+
+        if (parameters.stackTrace != null)
```
gabriela-trutan-sonarsource, line 54: Move the null check inside the LogVerbose method

gabriela-trutan-sonarsource (approved):

src/SLCore.Listeners/Implementation/LoggerListener.cs
georgii-borovinskikh-sonarsource: I will leave it as it was before, just `SLCore`

src/SLCore.Listeners/Implementation/LoggerListener.cs
georgii-borovinskikh-sonarsource: Don't really want to move this logic inside the logger. It's not really the logger's business IMO

PR #5933, opened 2025-01-07
SLVS-1660 Create integration tests for CFamily

vnaskos-sonar:
[SLVS-1660](https://sonarsource.atlassian.net/browse/SLVS-1660)

[SLVS-1660]: https://sonarsource.atlassian.net/browse/SLVS-1660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SLCore.IntegrationTests/Resources/CFamilyIssues.cpp
+++ b/src/SLCore.IntegrationTests/Resources/CFamilyIssues.cpp
@@ -1,0 +1,10 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+    // TODO: Test
+    cout << "Hello CMake." << endl;
+    return 0;
+}
--- a/src/SLCore.IntegrationTests/Resources/CFamilyIssues.cpp
+++ b/src/SLCore.IntegrationTests/Resources/CFamilyIssues.cpp
@@ -4,7 +4,10 @@ using namespace std;
 
 int main()
 {
-    // TODO: Test
+    // TODO: This is an issue
     cout << "Hello CMake." << endl;
+    
+    int* a = NULL; // Some more issues
+    
     return 0;
 }
--- a/src/SLCore.IntegrationTests/SimpleAnalysisTests.cs
+++ b/src/SLCore.IntegrationTests/SimpleAnalysisTests.cs
@@ -18,8 +18,8 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.IO;
 using SonarLint.VisualStudio.SLCore.Common.Models;
-using SonarLint.VisualStudio.SLCore.Listener.Analysis.Models;
 
 namespace SonarLint.VisualStudio.SLCore.IntegrationTests;
 
@@ -67,6 +67,14 @@ public class SimpleAnalysisTests
         => DefaultRuleConfig_AnalysisProducesExpectedIssuesInFile(FileAnalysisTestsRunner.TypeScriptIssues, true);
 
     [TestMethod]
+    public Task DefaultRuleConfig_ContentFromDisk_CFamilyAnalysisProducesExpectedIssues()
+        => DefaultRuleConfig_AnalysisProducesExpectedIssuesInFile(FileAnalysisTestsRunner.CFamilyIssues, false, GenerateTestCompilationDatabase());
+
+    [TestMethod]
+    public Task DefaultRuleConfig_ContentFromRpc_CFamilyAnalysisProducesExpectedIssues()
+        => DefaultRuleConfig_AnalysisProducesExpectedIssuesInFile(FileAnalysisTestsRunner.CFamilyIssues, true, GenerateTestCompilationDatabase());
+
+    [TestMethod]
     public Task DefaultRuleConfig_ContentFromDisk_CssAnalysisProducesExpectedIssues()
         => DefaultRuleConfig_AnalysisProducesExpectedIssuesInFile(FileAnalysisTestsRunner.CssIssues, false);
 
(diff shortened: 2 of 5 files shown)


src/SLCore.IntegrationTests/SimpleAnalysisTests.cs
```diff
@@ -82,13 +90,41 @@ public Task DefaultRuleConfig_ContentFromDisk_CssAnalysisInVueProducesExpectedIs
...
+
+    private static Dictionary<string, string> GenerateTestCompilationDatabase()
+    {
+        /* The CFamily analysis apart from the source code file requires also the compilation database file.
+           The compilation database file must contain the absolute path to the source code file the compilation database json file and the compiler path.
+           For the compiler we use the MSVC which is set as an environment variable. Make sure the environment variable is set to point to the compiler path
+           (the absolute path to cl.exe). */
+        var compilerPath = Environment.GetEnvironmentVariable("MSVC")?.Replace(@"\", @"\\");
+        var cFamilyIssuesFileAbsolutePath = FileAnalysisTestsRunner.CFamilyIssues.GetFullPath().Replace(@"\", @"\\");
+        var analysisDirectory = cFamilyIssuesFileAbsolutePath.Substring(0, cFamilyIssuesFileAbsolutePath.LastIndexOf('\\') + 1);
```
georgii-borovinskikh-sonarsource, line 111: Use FileInfo/DirectoryInfo.Parent

src/SLCore.IntegrationTests/SimpleAnalysisTests.cs
```diff
@@ -82,13 +90,41 @@ public Task DefaultRuleConfig_ContentFromDisk_CssAnalysisInVueProducesExpectedIs
...
         receivedTestIssues.Should().BeEquivalentTo(testingFile.ExpectedIssues);
     }
+
+    private static Dictionary<string, string> GenerateTestCompilationDatabase()
+    {
+        /* The CFamily analysis apart from the source code file requires also the compilation database file.
+           The compilation database file must contain the absolute path to the source code file the compilation database json file and the compiler path.
+           For the compiler we use the MSVC which is set as an environment variable. Make sure the environment variable is set to point to the compiler path
+           (the absolute path to cl.exe). */
+        var compilerPath = Environment.GetEnvironmentVariable("MSVC")?.Replace(@"\", @"\\");
```
georgii-borovinskikh-sonarsource, line 109: Double check if there's a helper in Path to normalize directory separators

src/SLCore.IntegrationTests/SimpleAnalysisTests.cs
```diff
@@ -66,6 +66,14 @@ public Task DefaultRuleConfig_ContentFromDisk_TypeScriptAnalysisProducesExpected
     public Task DefaultRuleConfig_ContentFromRpc_TypeScriptAnalysisProducesExpectedIssues()
         => DefaultRuleConfig_AnalysisProducesExpectedIssuesInFile(FileAnalysisTestsRunner.TypeScriptIssues, true);
 
+    [TestMethod]
+    public Task DefaultRuleConfig_ContentFromDisk_CFamilyAnalysisProducesExpectedIssues()
```
georgii-borovinskikh-sonarsource, line 70: are these tests going to fail if the variable is not set? should we only enable them on ci then? I wonder if you can use environment variables in conditional compilation
```
#if MSVC
    [TestMethod]
#endif
```

src/SLCore.IntegrationTests/Resources/CFamilyIssues.cpp
```diff
@@ -0,0 +1,10 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+    // TODO: Test
```
georgii-borovinskikh-sonarsource, line 7: Add a bit more issues, at least one non-todo

georgii-borovinskikh-sonarsource (changes requested):

src/SLCore.IntegrationTests/SimpleAnalysisTests.cs
vnaskos-sonar: Unfortunately, I couldn't find a better way :/

src/SLCore.IntegrationTests/SimpleAnalysisTests.cs
vnaskos-sonar: The problem with this is that it requires the double backslashes at the end or else the analysis fails. The DirectoryInfo.Parent removes those and adding them back fills equally awkward.

src/SLCore.IntegrationTests/SimpleAnalysisTests.cs
vnaskos-sonar: There are ways to achieve that (eg. https://stackoverflow.com/questions/70336885/c-sharp-preprocessor-define-symbol-based-on-environment-variable). However, on the local machine you either run this tests on purpose or not, so it makes sense to see them fail if you run and haven't set the env variable.

src/SLCore.IntegrationTests/SimpleAnalysisTests.cs
georgii-borovinskikh-sonarsource: https://github.com/SonarSource/sonarlint-visualstudio/blob/9a39b7f44b699aff862af99ed96acdae17e5ca2e/src/Integration.Vsix/CFamily/VcxProject/FileConfig.cs#L70

georgii-borovinskikh-sonarsource (approved):
Still the 2 minor notes about the paths are left

PR #5948, opened 2025-01-10
SLVS-1717 Fix C++ Language standard recognition when a C language standard is also specified

georgii-borovinskikh-sonarsource:
* Only set one language standard for the file
* Always set either `/TC` or `/TP` flag for the file (`/TC` for C and `/TP` for CPP)
* Assume the file is C if `CompileAsC `or if not `CompileAsCpp` and `vcFile.ContentType` is `CCode`
* Assume the file is CPP otherwise

Note: it's entirely possible to have a file with `CompileAsCpp` and `CCode`, `CompileAsC` and `CppCode`, `Default` and `CppHeader` or any other combination of these values by manually setting the overrides in vcxproj. For consistency, `CompileAs` is chosen as the primary indicator for the language, `vcFile.ContentType` is only used if `CompileAs` is not set to `CompileAsCpp`

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/CmdBuilderTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CmdBuilderTests.cs
@@ -13,6 +13,7 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  * Lesser General Public License for more details.
  *
+ *
  * You should have received a copy of the GNU Lesser General Public License
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
@@ -235,12 +236,12 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
         [DataTestMethod]
         [DataRow(false, "", "", "")]
         [DataRow(true, "", "", "")]
-        [DataRow(false, "1", "", "1 ")]
-        [DataRow(true, "1", "", "1 ")]
-        [DataRow(false, "", "2", "2 ")]
-        [DataRow(true, "", "2", "2 ")]
-        [DataRow(false, "1", "2", "1 2 ")]
-        [DataRow(true, "1", "2", "1 2 ")]
+        [DataRow(false, "compileAsFlagValue", "", "compileAsFlagValue ")]
+        [DataRow(true, "compileAsFlagValue", "", "compileAsFlagValue ")]
+        [DataRow(false, "", "languageStandardFlagValue", "languageStandardFlagValue ")]
+        [DataRow(true, "", "languageStandardFlagValue", "languageStandardFlagValue ")]
+        [DataRow(false, "compileAsFlagValue", "compileAsFlagValue", "compileAsFlagValue compileAsFlagValue ")]
+        [DataRow(true, "compileAsFlagValue", "compileAsFlagValue", "compileAsFlagValue compileAsFlagValue ")]
         public void AddOptFromProperties_DelegatesToLanguageFlagsProvider_AddsAvailableLanguageFlags(bool isHeader, string compileAsFlag, string languageStandardFlag, string expectedCmd)
         {
             var languageFlagsProvider = Substitute.For<ILanguageFlagsProvider>();
--- a/src/Integration.Vsix.UnitTests/CFamily/CmdBuilderTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/CmdBuilderTests.cs
@@ -18,11 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System;
-using FluentAssertions;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.VCProjectEngine;
-using Moq;
 using SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject;
 
 
@@ -31,27 +27,6 @@ namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily
     [TestClass]
     public class CmdBuilderTests
     {
-
-        [TestMethod]
-        [DataRow("", "")]
-        [DataRow("Default", "")]
-        [DataRow("CompileAsC", "/TC")]
-        [DataRow("CompileAsCpp", "/TP")]
-        [DataRow("Invalid", "Unsupported CompileAs: Invalid")]
-        public void ConvertCompileAsAndGetLanguage(string input, string output)
-        {
-            if ("Invalid".Equals(input))
-            {
-                Action action = () => CmdBuilder.ConvertCompileAsAndGetLanguage(input);
-                action.Should().ThrowExactly<ArgumentException>().And.Message.Should()
-                    .StartWith(output);
-            }
-            else
-            {
-                CmdBuilder.ConvertCompileAsAndGetLanguage(input).Should().Be(output);
-            }
-        }
-
         [TestMethod]
         [DataRow("", "")]
         [DataRow("false", "")]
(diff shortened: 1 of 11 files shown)


src/Integration.Vsix.UnitTests/CFamily/VcxProject/FileConfigTests.cs
```diff
@@ -153,7 +154,120 @@ public void TryGet_Full_Cmd()
...
+        }
+
+        [TestMethod]
+        public void TryGet_Full_Cmd_ForCompileAsCFile()
+        {
+            // Arrange
+            var projectItemConfig = new ProjectItemConfig
+            {
+                ItemType = "ClCompile",
+                FileConfigProperties = new Dictionary<string, string>
```
gabriela-trutan-sonarsource, line 171: Please extract the creation of the properties in a method to prevent cluttering the test. Then reuse the method in all other tests.

src/Integration.Vsix/CFamily/VcxProject/CmdBuilder.cs
```diff
@@ -165,14 +164,22 @@ internal void AddOptFromProperties(IVCRulePropertyStorage properties)
...
-
             // Additional options
             var additionalOptions = properties.GetEvaluatedPropertyValue("AdditionalOptions");
             AddCmdOpt(additionalOptions);
         }
 
+        private void AddLanguageFlags(IVCRulePropertyStorage properties)
+        {
+            var (compileAsFlag, languageStandardFlag) = languageFlagsProvider.GetLanguageConfiguration(
+                GetPotentiallyUnsupportedPropertyValue(properties, "CompileAs", ""),
```
gabriela-trutan-sonarsource, line 175: Please use string.Empty
```suggestion
                GetPotentiallyUnsupportedPropertyValue(properties, "CompileAs", string.Empty),
```

src/Integration.Vsix/CFamily/VcxProject/CmdBuilder.cs
```diff
@@ -165,14 +164,22 @@ internal void AddOptFromProperties(IVCRulePropertyStorage properties)
...
+                GetPotentiallyUnsupportedPropertyValue(properties, "LanguageStandard_C", ""),
```
gabriela-trutan-sonarsource, line 176:
```suggestion
                GetPotentiallyUnsupportedPropertyValue(properties, "LanguageStandard_C", string.Empty),
```

src/Integration.Vsix/CFamily/VcxProject/CmdBuilder.cs
```diff
@@ -165,14 +164,22 @@ internal void AddOptFromProperties(IVCRulePropertyStorage properties)
...
+                GetPotentiallyUnsupportedPropertyValue(properties, "LanguageStandard", ""));
```
gabriela-trutan-sonarsource, line 177:
```suggestion
                GetPotentiallyUnsupportedPropertyValue(properties, "LanguageStandard", string.Empty));
```

src/Integration.Vsix/CFamily/VcxProject/CompileAsValueConverter.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject;
+
+internal static class CompileAsValueConverter
+{
+    internal /* for testing */ static string GetFlagValue(string value) =>
+        value switch
+        {
+            // https://github.com/SonarSource/sonarlint-visualstudio/issues/738
+            "" or "Default" => "",
```
gabriela-trutan-sonarsource, line 29:
```suggestion
            "" or "Default" => string.Empty,
```

src/Integration.Vsix/CFamily/VcxProject/LanguageStandardConverter.cs
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+namespace SonarLint.VisualStudio.Integration.Vsix.CFamily.VcxProject;
+
+internal static class LanguageStandardConverter
+{
+    public static string GetCppStandardFlagValue(string value) =>
+        value switch
+        {
+            null or "" or "Default" => "",
```
gabriela-trutan-sonarsource, line 28:
```suggestion
            null or "" or "Default" => string.Empty,
```

src/Integration.Vsix/CFamily/VcxProject/LanguageStandardConverter.cs
```diff
@@ -0,0 +1,46 @@
...
+            "stdcpp20" => "/std:c++20",
+            "stdcpp17" => "/std:c++17",
+            "stdcpp14" => "/std:c++14",
+            _ => throw new ArgumentException($"Unsupported LanguageStandard: {value}")
+        };
+
+    public static string GetCStandardFlagValue(string value) =>
+        value switch
+        {
+            null or "" or "Default" => "",
```
gabriela-trutan-sonarsource, line 39:
```suggestion
            null or "" or "Default" => string.Empty,
```

gabriela-trutan-sonarsource (approved):
Looks good. I only have some minor feedback.

src/Integration.Vsix/CFamily/VcxProject/LanguageFlagsProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+    public LanguageFlagsProvider(string vcFileContentType)
+    {
+        this.vcFileContentType = vcFileContentType;
+    }
+
+    public (string compileAsFlag, string languageStandardFlag) GetLanguageConfiguration(string compileAs, string cStandard, string cppStandard)
+    {
+        var compileAsFlag = CompileAsValueConverter.GetFlagValue(compileAs);
+
+        return compileAsFlag == CompileAsCFlag || (compileAsFlag != CompileAsCppFlag && vcFileContentType == ContentTypeCCode)
```
michael-jabbour-sonarsource, line 44: Would it make sense to add a comment to explain why checking `compileAsFlag` isn't sufficient?

I think we mentioned that we now need `vcFileContentType` as well for .c files when `compileAsFlag` is kept as default.

src/Integration.Vsix.UnitTests/CFamily/CmdBuilderTests.cs
```diff
@@ -316,21 +232,29 @@ public void AddOptFromProperties(string input, string output, string cmd)
...
-        [DataRow("Default", "")]
-        [DataRow("CompileAsC", "/TC ")]
-        [DataRow("CompileAsCpp", "/TP ")]
-        public void HeaderFileLang(string compileAs, string cmd)
+        [DataTestMethod]
+        [DataRow(false, "", "", "")]
+        [DataRow(true, "", "", "")]
+        [DataRow(false, "1", "", "1 ")]
+        [DataRow(true, "1", "", "1 ")]
+        [DataRow(false, "", "2", "2 ")]
```
michael-jabbour-sonarsource, line 240: Minor comment:

The real `LanguageFlagsProvider` seems to always return a non-empty string for the compile-as flag (i.e. either `/TC` or `/TP`, but never blank), as described in the PR summary. This implies that the expected command should always have either of these.

I was confused by the expected "2 " here (which means that no `/TC` or `/TP` were added). As far as I understand, this mock in this test doesn't behave similar to the real implementation.

Would it make sense to somehow align the behavior of the mock here with the real implementation by always returning a non-empty for `compileAsFlag`?

(I would leave it for you to decide if it is worth it, I am just sharing it because it took me a bit to understand what is going on)

michael-jabbour-sonarsource (approved):
LGTM! Thanks for taking care of this :pray:

It seems to work as expected. I left two minor comments, and I am letting you decide if they are worth addressing!

src/Integration.Vsix.UnitTests/CFamily/CmdBuilderTests.cs
georgii-borovinskikh-sonarsource: The idea of the test was more to show that the implementation for choosing the values is delegated and should be tested in the LanguageFlagsProviderTests. I will just change the values to be more readable than just 1 & 2.

PR #5972, opened 2025-01-29
SLVS-1786 Promote t-sql to standalone users

vnaskos-sonar:
[SLVS-1786](https://sonarsource.atlassian.net/browse/SLVS-1786)

[SLVS-1786]: https://sonarsource.atlassian.net/browse/SLVS-1786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode/Promote/PromoteGoldBar.cs
+++ b/src/ConnectedMode/Promote/PromoteGoldBar.cs
@@ -1,0 +1,100 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.ConnectedMode.UI;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Binding;
+using SonarLint.VisualStudio.Core.Notifications;
+using SonarLint.VisualStudio.Core.Telemetry;
+
+namespace SonarLint.VisualStudio.ConnectedMode.Promote;
+
+public interface IPromoteGoldBar
+{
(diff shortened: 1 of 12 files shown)


src/SLCore.Listeners/Implementation/PromoteListener.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace SonarLint.VisualStudio.SLCore.Listeners.Implementation;
+
+[Export(typeof(ISLCoreListener))]
+[PartCreationPolicy(CreationPolicy.Shared)]
+[method: ImportingConstructor]
+public class PromoteListener(IPromoteGoldBar promoteGoldBar) : IPromoteListener
+{
+    public void PromoteExtraEnabledLanguagesInConnectedMode(PromoteExtraEnabledLanguagesInConnectedModeParams parameters)
+    {
+        var languagesToPromote = string.Join(", ", parameters.languagesToPromote);
```
georgii-borovinskikh-sonarsource, line 35: Don't think that the listener should be responsible for formatting

src/SLCore.Listeners/Implementation/PromoteListener.cs
```diff
@@ -0,0 +1,38 @@
...
+        promoteGoldBar.PromoteConnectedMode(languagesToPromote);
```
georgii-borovinskikh-sonarsource, line 36: We ignore the provided config scope here. What do they do with it in Intellij?

src/ConnectedMode/Promote/PromoteGoldBar.cs
```diff
@@ -0,0 +1,100 @@
...
+        this.activeSolutionBoundTracker = activeSolutionBoundTracker;
+        this.browserService = browserService;
+        this.telemetryManager = telemetryManager;
+        this.connectedModeUiManager = connectedModeUiManager;
+
+        this.activeSolutionBoundTracker.SolutionBindingChanged += OnActiveSolutionBindingChanged;
+    }
+
+    public void PromoteConnectedMode(string languagesToPromote)
+    {
```
georgii-borovinskikh-sonarsource, line 66: Could check if we're in connected mode here. If you decide to add the config scope check, both can be combined

src/ConnectedMode/Promote/PromoteGoldBar.cs
```diff
@@ -0,0 +1,100 @@
...
+{
+    private readonly INotificationService notificationService;
+    private readonly IDoNotShowAgainNotificationAction doNotShowAgainNotificationAction;
+    private readonly IActiveSolutionBoundTracker activeSolutionBoundTracker;
+    private readonly IBrowserService browserService;
+    private readonly ITelemetryManager telemetryManager;
+    private readonly IConnectedModeUIManager connectedModeUiManager;
+
+    [ImportingConstructor]
+    public PromoteGoldBar(
```
georgii-borovinskikh-sonarsource, line 47: Elsewhere, except in one other place, we used the term `Notification` instead of `GoldBar`

georgii-borovinskikh-sonarsource (approved):
Some small issues, don't need to re-review when you fix them

src/SLCore.Listeners/Implementation/PromoteListener.cs
georgii-borovinskikh-sonarsource: Actually, convert to core language since that has user-friendly names

PR #5974, opened 2025-01-30
URL slug rebranding update

kevin-hinz:
Part of the rebranding efforts necessitates the update of the URL structure on the Docs website.
This change reflects part of that update.
A redirect has been added to address links from outside the product.


vnaskos-sonar (approved):
👍

PR #5980, opened 2025-02-03
SLVS-1822 Skip Roslyn settings sync when no solution is open

georgii-borovinskikh-sonarsource:
[SLVS-1822](https://sonarsource.atlassian.net/browse/SLVS-1822)

Part of SLVS-1766

[SLVS-1822]: https://sonarsource.atlassian.net/browse/SLVS-1822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
+++ b/src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
@@ -105,9 +105,13 @@ internal sealed class RoslynSettingsFileSynchronizer : IRoslynSettingsFileSynchr
         {
             await threadHandling.SwitchToBackgroundThread();
 
-            var sonarProjectKey = configurationProvider.GetConfiguration().Project?.ServerProjectKey;
-
             var solutionNameWithoutExtension = await GetSolutionNameWithoutExtensionAsync();
+            if (string.IsNullOrEmpty(solutionNameWithoutExtension))
+            {
+                return;
+            }
+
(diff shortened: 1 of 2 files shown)


src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
```diff
@@ -105,9 +105,13 @@ public async Task UpdateFileStorageAsync()
         {
             await threadHandling.SwitchToBackgroundThread();
 
-            var sonarProjectKey = configurationProvider.GetConfiguration().Project?.ServerProjectKey;
-
             var solutionNameWithoutExtension = await GetSolutionNameWithoutExtensionAsync();
+            if (string.IsNullOrEmpty(solutionNameWithoutExtension))
+            {
+                return;
```
vnaskos-sonar, line 111: A Verbose log message would be nice for future debugging, "No solution name found, skipping Roslyn settings sync" or something like that.

vnaskos-sonar (approved):
👍

src/Roslyn.Suppressions/Roslyn.Suppressions/InProcess/RoslynSettingsFileSynchronizer.cs
georgii-borovinskikh-sonarsource: I think this log is redundant, as many other things would break before we get to this if in the case of there-should-be-a-solution-but-the-name-is-null and there's no possibility for a race condition as the code is triggered from a solution change on the UI thread, and it queries the solution name on the same UI thread

PR #5983, opened 2025-02-04
SLVS-1825   Refactor to avoid multiple sources of truth for pluginKey

gabriela-trutan-sonarsource:
This targets the [refactoring PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/5981) of the repoKey, because it depends on those changes.


src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
```diff
@@ -72,11 +75,12 @@ public Dictionary<string, string> ListConnectedModeEmbeddedPluginPathsByKey()
         var connectedModeEmbeddedPluginPathsByKey = new Dictionary<string, string>();
         var embeddedPluginFilePaths = ListJarFiles();
 
-        foreach (var kvp in connectedModePluginNamePatternByPluginKey)
+        foreach (var group in standaloneLanguages.GroupBy(x => x.PluginInfo.PluginKey).Distinct())
```
georgii-borovinskikh-sonarsource, line 78: Select(x => x.PluginInfo).ToHashSet() & cache the result

src/Integration.Vsix/SLCore/SLCoreEmbeddedPluginJarLocator.cs
```diff
@@ -35,12 +35,15 @@ public class SLCoreEmbeddedPluginJarLocator : ISLCoreEmbeddedPluginJarLocator
 {
     private const string JarFolderName = "DownloadedJars";
 
-    private readonly Dictionary<string, string> connectedModePluginNamePatternByPluginKey = new()
+    private readonly List<Language> standaloneLanguages = new()
```
georgii-borovinskikh-sonarsource, line 38: can be made static

src/Core/PluginInfo.cs
```diff
@@ -0,0 +1,37 @@
...
+
+public record PluginInfo
+{
+    public PluginInfo(string pluginKey, string filePattern)
+    {
+        PluginKey = pluginKey ?? throw new ArgumentNullException(nameof(pluginKey));
+        FilePattern = filePattern;
+    }
+
+    public string PluginKey { get; }
```
georgii-borovinskikh-sonarsource, line 31: Key

src/Core/PluginInfo.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <summary>
+    /// Nullable, because it can be a connected mode only language so there will be no file on disk (plugin won't be embedded)
+    /// </summary>
+    public string FilePattern { get; }
```
georgii-borovinskikh-sonarsource, line 36: Do we need the plugin info for that language in such case?

georgii-borovinskikh-sonarsource (approved):

src/Core/PluginInfo.cs
gabriela-trutan-sonarsource: Yes. See [this](https://github.com/SonarSource/sonarlint-visualstudio/blob/61a8446193a34d0119cf8e9aa3fd0261a56b5a3a/src/SLCore/Common/Helpers/LanguageExtensions.cs#L52): Tsql plugin key is needed, even if it is not embedded

gabriela-trutan-sonarsource: It is used in the SLCoreInstanceHandle in the initialize method. And although it is not needed per se, because the disabled languages are just the roslyn ones, I think it will be awkward that the LanguageExtensions.GetPluginKey will return null for a known language. I think it's fine that we know the pluginKey for a language, even if it is not embedded (in case we need to provide it to SlCore for any other reason)

PR #5987, opened 2025-02-04
SLVS-1826 Introduce ILanguageProvider

gabriela-trutan-sonarsource:
[SLVS-1826](https://sonarsource.atlassian.net/browse/SLVS-1826)

Reduce the number of places that have to be adapted when adding a new language by introducing a new service, LanguageProvider, and by using it in all the places where lists of Language instances were used.

[SLVS-1826]: https://sonarsource.atlassian.net/browse/SLVS-1826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/ILanguageProvider.cs
+++ b/src/Core/ILanguageProvider.cs
@@ -1,0 +1,62 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.ComponentModel.Composition;
+
+namespace SonarLint.VisualStudio.Core;
+
+public interface ILanguageProvider
+{
+    IReadOnlyList<Language> AllKnownLanguages { get; }
+
+    IReadOnlyList<Language> SlCoreLanguages { get; }
+
+    IReadOnlyList<Language> RoslynLanguages { get; }
+
+    IReadOnlyList<Language> LanguagesInStandaloneMode { get; }
+
+    IReadOnlyList<Language> ExtraLanguagesInConnectedMode { get; }
+
+    /// <summary>
+    /// Returns the language for the specified language key, or null if it does not match a known language
+    /// </summary>
+    Language GetLanguageFromLanguageKey(string languageKey);
+}
+
+[Export(typeof(ILanguageProvider))]
+[PartCreationPolicy(CreationPolicy.Shared)]
+public class LanguageProvider : ILanguageProvider
+{
+    [ImportingConstructor]
+    public LanguageProvider()
+    {
+        SlCoreLanguages = AllKnownLanguages.Except(RoslynLanguages).ToList();
+        LanguagesInStandaloneMode = AllKnownLanguages.Except(ExtraLanguagesInConnectedMode).ToList();
+    }
+
+    public IReadOnlyList<Language> SlCoreLanguages { get; }
+    public IReadOnlyList<Language> RoslynLanguages { get; } = [Language.CSharp, Language.VBNET];
+    public IReadOnlyList<Language> AllKnownLanguages { get; } =
+        [Language.CSharp, Language.VBNET, Language.C, Language.Cpp, Language.Js, Language.Ts, Language.Css, Language.Secrets, Language.Html, Language.TSql];
+    public IReadOnlyList<Language> LanguagesInStandaloneMode { get; }
+    public IReadOnlyList<Language> ExtraLanguagesInConnectedMode { get; } = [Language.TSql];
+
+    public Language GetLanguageFromLanguageKey(string languageKey) => AllKnownLanguages.FirstOrDefault(l => languageKey.Equals(l.ServerLanguage.Key, StringComparison.OrdinalIgnoreCase));
+}
--- a/src/Core/ILanguageProvider.cs
+++ b/src/Core/ILanguageProvider.cs
@@ -44,6 +44,10 @@ public interface ILanguageProvider
 [PartCreationPolicy(CreationPolicy.Shared)]
 public class LanguageProvider : ILanguageProvider
 {
+#pragma warning disable S4277
+    public static readonly ILanguageProvider Instance = new LanguageProvider();
+#pragma warning restore S4277
+
     [ImportingConstructor]
     public LanguageProvider()
     {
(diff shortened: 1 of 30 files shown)


src/Core/ILanguageProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+
+using System.ComponentModel.Composition;
+
+namespace SonarLint.VisualStudio.Core;
+
+public interface ILanguageProvider
+{
+    IReadOnlyList<Language> AllKnownLanguages { get; }
+
+    IReadOnlyList<Language> SlCoreLanguages { get; }
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: You could then define AllKnownLanguages as Roslyn + NonRoslyn

src/SLCore/Common/Helpers/LanguageExtensions.cs
```diff
@@ -27,18 +28,63 @@ internal static class LanguageExtensions
...
+            Language.SECRETS => Secrets,
+            Language.TS => Ts,
+            Language.VBNET => VBNET,
+            Language.TSQL => TSql,
+            _ => Unknown
         };
 
+    public static Language ConvertToSlCoreLanguage(this VisualStudio.Core.Language language)
+    {
+        if (language.Id == C.Id)
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Also, it's only used in SLCoreLanguageProvider, so it probably doesn't need to be an extension method

src/Core/ILanguageProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// </summary>
+    Language GetLanguageFromLanguageKey(string languageKey);
+}
+
+[Export(typeof(ILanguageProvider))]
+[PartCreationPolicy(CreationPolicy.Shared)]
+public class LanguageProvider : ILanguageProvider
+{
+#pragma warning disable S4277
+    public static readonly ILanguageProvider Instance = new LanguageProvider();
```
georgii-borovinskikh-sonarsource, line 48: I don't like this approach. We should only go around mef exporting interfaces if it's technically impossible to do. Using this instance field makes the classes un-mockable

src/Core/ILanguageProvider.cs
```diff
@@ -0,0 +1,66 @@
+﻿/*
```
georgii-borovinskikh-sonarsource, line 1: I would split this PR in 2: one that adds the class, and another one that changes the code that should use it. It's hard to review at the moment

src/SLCore/Configuration/ISLCoreLanguageProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.SLCore.Common.Helpers;
+using Language = SonarLint.VisualStudio.SLCore.Common.Models.Language;
+
+namespace SonarLint.VisualStudio.SLCore.Configuration;
+
+public interface ISLCoreLanguageProvider
+{
+    IReadOnlyList<Language> LanguagesInStandaloneMode { get; }
```
georgii-borovinskikh-sonarsource, line 30: Not a major point, but we could use ImmutableArray because this is an immutable collection of immutable items. ImmutableArray conveys that semantic well

src/SLCore.UnitTests/Common/Helpers/LanguageExtensionsTests.cs
```diff
@@ -54,13 +55,32 @@ public void VerifyConversionToCoreLanguage()
     [DataRow(Language.TSQL, "tsql")]
     [DataRow(Language.ABAP, null)]
     [DataRow(Language.JAVA, null)]
-    public void VerifyPluginKeys(Language language, string expectedPluginKey)
+    public void VerifyPluginKeys(Language language, string expectedPluginKey) => language.GetPluginKey().Should().BeEquivalentTo(expectedPluginKey);
+
+    [TestMethod]
+    public void VerifyConvertToSlCoreLanguage()
```
georgii-borovinskikh-sonarsource, line 61: https://learn.microsoft.com/en-us/visualstudio/test/how-to-create-a-data-driven-unit-test?view=vs-2022#member-data-driven-test

src/Integration/SLCore/SLCoreConstantsProvider.cs
```diff
@@ -82,12 +54,7 @@ private static IdeVersionInformation GetVsVersion(IVsVersion vsVersion)
                 return null;
             }
 
-            return new IdeVersionInformation
-            {
-                DisplayName = vsVersion.DisplayName,
-                InstallationVersion = vsVersion.InstallationVersion,
-                DisplayVersion = vsVersion.DisplayVersion
-            };
+            return new IdeVersionInformation { DisplayName = vsVersion.DisplayName, InstallationVersion = vsVersion.InstallationVersion, DisplayVersion = vsVersion.DisplayVersion };
```
georgii-borovinskikh-sonarsource, line 57: I think it was more readable before

src/Integration.Vsix/Analysis/DisableRuleCommand.cs
```diff
@@ -72,18 +78,24 @@ internal DisableRuleCommand(
...
+            {
+                throw new ArgumentNullException(nameof(languageProvider));
+            }
 
             this.userSettingsProvider = userSettingsProvider ?? throw new ArgumentNullException(nameof(userSettingsProvider));
             this.activeSolutionBoundTracker = activeSolutionBoundTracker ?? throw new ArgumentNullException(nameof(activeSolutionBoundTracker));
             this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
             this.errorListHelper = errorListHelper ?? throw new ArgumentNullException(nameof(errorListHelper));
 
+            supportedRepos = languageProvider.LanguagesInStandaloneMode.Select(x => x.RepoInfo.Key).ToList();
```
georgii-borovinskikh-sonarsource, line 98: This should only be enabled for standalone **non-roslyn** languages

georgii-borovinskikh-sonarsource (changes requested):

src/Integration/SLCore/SLCoreConstantsProvider.cs
gabriela-trutan-sonarsource: It's Resharper's change. If the setting should be changed, please create a task outside this PR

src/Core/ILanguageProvider.cs
gabriela-trutan-sonarsource: I could split the PR, but then I would lose all your feedback from here.
I know the PR is a little bit big and I am sorry for that. But if you review it commit by commit, it should be understandable.

src/Core/ILanguageProvider.cs
gabriela-trutan-sonarsource: It is needed due the fact that the instance is needed in few classes that are not MEF components: **RuleHelpLinkProvider** (which is called by IssuesSnapshot, which in turn is called by other non-MEF-able components) and **LanguageConverter**

src/SLCore/Common/Helpers/LanguageExtensions.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: > Also, it's only used in SLCoreLanguageProvider, so it probably doesn't need to be an extension method

I don't want to have yet another place to adapt when we introduce a new language. Here it should be fine.

src/SLCore/Configuration/ISLCoreLanguageProvider.cs
gabriela-trutan-sonarsource: I think that the IReadOnlyList also conveys this as well.

src/Core/ILanguageProvider.cs
georgii-borovinskikh-sonarsource: `OutOfDateQualityProfileFinder`, `QualityProfileDownloader`, `NonRoslynDummyBindingConfigProvider` should not use it.

gabriela-trutan-sonarsource:
As requested, this PR has been split into multiple PRs and it will be closed: [introduce LanguageProvider](https://github.com/SonarSource/sonarlint-visualstudio/pull/5993), [refactor commands](https://github.com/SonarSource/sonarlint-visualstudio/pull/5994), [refactor SlCoreConstantsProvider](https://github.com/SonarSource/sonarlint-visualstudio/pull/5998).

The suggestions from here have been addressed in the corresponding PRs.

PR #5988, opened 2025-02-05
SLVS-1813 Introduce region enum

gabriela-trutan-sonarsource:
[SLVS-1813](https://sonarsource.atlassian.net/browse/SLVS-1813)

[SLVS-1813]: https://sonarsource.atlassian.net/browse/SLVS-1813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/Binding/CloudServerRegion.cs
+++ b/src/Core/Binding/CloudServerRegion.cs
@@ -1,0 +1,26 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


src/Core/Binding/CloudServerRegion.cs
```diff
@@ -0,0 +1,37 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.Core.Binding;
+
+public sealed class CloudServerRegion
```
georgii-borovinskikh-sonarsource, line 23: We can't use this in SonarQube.Client assembly. What is the plan in that regard, considering we wanted to have one source of truth for SC URL?

gabriela-trutan-sonarsource: I missed this pain point. I think we should refactor the code to change the dependency between the SonarQube.Client and Core assemblies. Otherwise we will continue having duplicate information which brings a lot of additional effort.
I will update the task accordingly.

PR #6001, opened 2025-02-07
SLVS-1807 Add Dogfooding flag service

georgii-borovinskikh-sonarsource:


src/Integration/Dogfooding/DogfoodingService.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace SonarLint.VisualStudio.Integration.Dogfooding;
+
+[Export(typeof(IDogfoodingService))]
+[PartCreationPolicy(CreationPolicy.Shared)]
+[method: ImportingConstructor]
+public class DogfoodingService(IEnvironmentVariableProvider environmentVariableProvider) : IDogfoodingService
+{
+    private const string SonarSourceDogfoodingEnvironmentVariable = "SONARSOURCE_DOGFOODING";
+
+    public bool IsDogfoodingEnvironment { get; } = !string.IsNullOrWhiteSpace(environmentVariableProvider.TryGet(SonarSourceDogfoodingEnvironmentVariable));
```
gabriela-trutan-sonarsource, line 33: It should not only check that the environment variable is set, but also that is set to 1. Reason: if we change it to 0, it means it is not dogfooding.

gabriela-trutan-sonarsource, line 33: We should not cache the variable, but always compute it. Reason for that: we should be able to test at runtime how it behaves in dogfooding/non-dogfooding

src/Integration/Dogfooding/DogfoodingService.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using System.ComponentModel.Composition;
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.Integration.Dogfooding;
+
+[Export(typeof(IDogfoodingService))]
+[PartCreationPolicy(CreationPolicy.Shared)]
+[method: ImportingConstructor]
+public class DogfoodingService(IEnvironmentVariableProvider environmentVariableProvider) : IDogfoodingService
```
gabriela-trutan-sonarsource, line 29: Why is it in the Integration assembly? I would also expect it to be in Core.

gabriela-trutan-sonarsource (changes requested):

src/Integration/Dogfooding/DogfoodingService.cs
georgii-borovinskikh-sonarsource: It doesn't need to be in core + there's already too much stuff in core

src/Integration/Dogfooding/DogfoodingService.cs
georgii-borovinskikh-sonarsource: Environment variables by default are not reloaded after the process starts

georgii-borovinskikh-sonarsource: I will change this, but what's the point of changing it to "0" as opposed to un-setting it completely?

gabriela-trutan-sonarsource: It's easier. Additionally, it would be unexpected if the value is "0" and we would return true

gabriela-trutan-sonarsource: It can be easily achieved with: Environment.GetEnvironmentVariable(variableName, EnvironmentVariableTarget.Machine)

Do we need to use the IEnvironmentVariablesProvider? If so, it can be additionally adapted to provide an overload

georgii-borovinskikh-sonarsource: > Do we need to use the IEnvironmentVariablesProvider?

src/Integration/Dogfooding/DogfoodingService.cs
gabriela-trutan-sonarsource: I don't know if the fact that there is "too much stuff" in Core is a good enough argument. I think in VS we mostly put implementations that have VS dependencies, which is not the case for this implementation

src/Integration/Dogfooding/DogfoodingService.cs
georgii-borovinskikh-sonarsource: I'm not sure we need this extra logic since we can just restart the process to see the effects of changing the variable. What's the use case? I don't see why you would need to frequently set/unset the variable, much less sense to do it when it's actually dogfooded/released

src/Integration/Dogfooding/DogfoodingService.cs
georgii-borovinskikh-sonarsource: It's not Integration.VS, that is a separate assembly

georgii-borovinskikh-sonarsource: It's not limited to vs-related stuff. In general, it doesn't make much sense to put service implementations in core. Core should be limited to models and interfaces. We've not been following that, which is a problem, but I think we should start doing that. Maybe we need a separate assembly called `Infrastructure`, but I think for now it's fine to put implementations of non-area-specific things (like ConnectedMode, IssueViz, etc) inside `Integration`

gabriela-trutan-sonarsource: I don't think Core should only contain interfaces and models. It can also contain logic/implementation of interfaces that is business logic related.

I think our assemblies are very confusing and there is no real, good architectural rules for our assemblies like Integration, Infrastucture. So adding the implementation to Integration felt like a decision that was not well thought, that's why I wanted to challenge it.

src/Core/IDogfoodingService.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    bool IsDogfoodingEnvironment { get; }
+}
+
+[Export(typeof(IDogfoodingService))]
+[PartCreationPolicy(CreationPolicy.Shared)]
+[method: ImportingConstructor]
+public class DogfoodingService(IEnvironmentVariableProvider environmentVariableProvider) : IDogfoodingService
+{
+    private const string SonarSourceDogfoodingEnvironmentVariable = "SONARSOURCE_DOGFOODING";
```
gabriela-trutan-sonarsource, line 35: Just a proposal:

```suggestion
    private const string DogfoodingEnvironmentVariable= "SONARSOURCE_DOGFOODING";
```

gabriela-trutan-sonarsource (approved):

PR #6002, opened 2025-02-07
SLVS-1817   Adapt UI model to support multi-region for SQC

gabriela-trutan-sonarsource:


src/ConnectedMode/ConnectionInfo.cs
```diff
@@ -32,19 +31,33 @@ public enum ConnectionServerType
...
+    /// <param name="id">The organization key for SonarCloud or the server uri for SonarQube</param>
+    /// <param name="serverType">The type of server (SonarCloud, SonarQube)</param>
+    /// <param name="cloudServerRegion"> Only applicable to <see cref="ConnectionServerType.SonarCloud"/>. If not provided, defaults to <see cref="Core.Binding.CloudServerRegion.Eu"/>
+    /// It is ignored for <see cref="ConnectionServerType.SonarQube"/> </param>
+    public ConnectionInfo(string id, ConnectionServerType serverType, CloudServerRegion cloudServerRegion = null)
     {
-        return serverConnection switch
+        Id = id;
+        ServerType = serverType;
+        CloudServerRegion = serverType == ConnectionServerType.SonarCloud ? (cloudServerRegion ?? CloudServerRegion.Eu) : null;
```
georgii-borovinskikh-sonarsource, line 47: Maybe we should throw instead of defaulting to null here?

georgii-borovinskikh-sonarsource (approved):

src/ConnectedMode/ConnectionInfo.cs
gabriela-trutan-sonarsource: We are defaulting to null only for SonarQube.

georgii-borovinskikh-sonarsource: Sorry, I meant throw IF it's set for SonarQube

PR #6006, opened 2025-02-10
SLVS-1846 Add reusable ConnectionIconComponent, replace duplicate layout with it

georgii-borovinskikh-sonarsource:
[SLVS-1846](https://sonarsource.atlassian.net/browse/SLVS-1846)

Part of SLVS-1804

[SLVS-1846]: https://sonarsource.atlassian.net/browse/SLVS-1846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ConnectedMode/UI/ConnectionInfoComponent.xaml
```diff
@@ -3,42 +3,20 @@
...
-                        </DataTrigger>
-                        <DataTrigger Binding="{Binding Path=ServerType}" Value="SonarQube">
-                            <Setter Property="Source" Value="{StaticResource SonarQubeIconDrawingImage}" />
-                        </DataTrigger>
-                    </Style.Triggers>
-                </Style>
-            </Image.Style>
-        </Image>
+        <ui:ConnectionIconComponent Grid.Column="0" ServerType="{Binding Path=ServerType}"
+                           Margin="-5,-5,0,-5"
```
gabriela-trutan-sonarsource: Please do not use negative values for margin. That usually indicates that something is wrong.

gabriela-trutan-sonarsource (approved):
Please make sure that the icon and text are actually vertically aligned to the previous text: (in the image they seem to be a little bit more to the bottom rather than centered)

after

src/ConnectedMode/UI/ManageConnections/ManageConnectionsDialog.xaml
```diff
@@ -66,24 +67,15 @@
                                 <ColumnDefinition Width="*" />
                                 <ColumnDefinition Width="Auto" />
                             </Grid.ColumnDefinitions>
-
-                            <Image Grid.Column="0" Width="30" Height="30" Margin="5,5,10,5" VerticalAlignment="Center"
```
gabriela-trutan-sonarsource, line 70: The icons look smaller than before (left is before and right is after your changes). I think before was better as now there is a lot of space that seems unnecessary

image

src/ConnectedMode/UI/ConnectionInfoComponent.xaml
georgii-borovinskikh-sonarsource: `public static readonly DependencyProperty ImageMarginProp = DependencyProperty.Register(nameof(ImageMargin), typeof(Thickness), typeof(ConnectionInfoComponent), new PropertyMetadata(new Thickness(-5,-5,0,-5)));` <- this is the value I used since I removed the Prop. Should I just add 5 to every side (0, 0, 5, 0)?

gabriela-trutan-sonarsource: Wouldn't 0 be acceptable?

PR #6012, opened 2025-02-11
SLVS-1810 Support multi-region in ConnectionNameComponent

georgii-borovinskikh-sonarsource:


src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
```diff
@@ -34,11 +43,16 @@ public ConnectionInfo ConnectionInfo
...
     }
 
     public string DisplayName =>
-        connectionInfo is { Id: null, ServerType: ConnectionServerType.SonarCloud}
+        connectionInfo is { Id: null, ServerType: ConnectionServerType.SonarCloud }
             ? connectionInfo.CloudServerRegion.Url.ToString()
             : connectionInfo?.Id ?? string.Empty;
+
+    public bool ShouldDisplayRegion => dogfoodingService.IsDogfoodingEnvironment && connectionInfo is { Id: not null, ServerType: ConnectionServerType.SonarCloud };
+    public string DisplayRegion => ShouldDisplayRegion ? connectionInfo.CloudServerRegion.Name : string.Empty;
```
gabriela-trutan-sonarsource, line 57: I would display it like (US) or maybe US - 
Or something to make it more easy to understand that it's about a region. For now, it's not very clear what US/EU is

src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
```diff
@@ -34,11 +43,16 @@ public ConnectionInfo ConnectionInfo
...
         }
     }
 
     public string DisplayName =>
-        connectionInfo is { Id: null, ServerType: ConnectionServerType.SonarCloud}
+        connectionInfo is { Id: null, ServerType: ConnectionServerType.SonarCloud }
             ? connectionInfo.CloudServerRegion.Url.ToString()
             : connectionInfo?.Id ?? string.Empty;
+
+    public bool ShouldDisplayRegion => dogfoodingService.IsDogfoodingEnvironment && connectionInfo is { Id: not null, ServerType: ConnectionServerType.SonarCloud };
```
gabriela-trutan-sonarsource, line 56: I would display the region also when the ID is null for consistency.

gabriela-trutan-sonarsource (changes requested):
There are a few problems:

- It's not easy to understand what US/EU text means as it gets lost in the wall of block that represents the connection. That's why I would add it in parenthesis or with a - or anything to make it distinguishable  
- When the connection name is long, it is not vertically aligned
 - I would add the region at the end, rather at the beginning as it doesn't align too well in the list of connections, when SQS connections are also present (see image)
 
 
image

src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
georgii-borovinskikh-sonarsource: There we display the full url, woudln't that be a bit redundant? Especially considering the only way the user can see the URL instead of the OrgKey is by getting to the credentials screen right after choosing the region

src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
```diff
@@ -18,13 +18,22 @@
...
+using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.WPF;
 
 namespace SonarLint.VisualStudio.ConnectedMode.UI.ConnectionDisplay;
 
 public class ConnectionNameViewModel : ViewModelBase
 {
+    public ConnectionNameViewModel() : this(DogfoodingService.Instance){ }
+
+    internal /* for testing */ ConnectionNameViewModel(IDogfoodingService dogfoodingService)
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Should I always pass the service from the UserControl then?

src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
gabriela-trutan-sonarsource: I think the fact that the difference is only in the URL, makes the information hidden (the user has to look if "us" is present). It should be visible and highlighted which region the user is connecting to in all screens. 
Also redundancy is not a problem in UX

src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: No, no. The code is fine. Just remove the  `/* for testing */` comment

src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
georgii-borovinskikh-sonarsource: I like the `US - ` approach more than braces and/or adding the region to the back

src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameComponent.xaml
```diff
@@ -3,15 +3,28 @@
...
+        </ResourceDictionary>
+    </UserControl.Resources>
+    
     <Grid DataContext="{Binding ElementName=ConnectionNameUc, Path=ViewModel}">
         <Grid.ColumnDefinitions>
+            <ColumnDefinition Width="Auto" />
             <ColumnDefinition Width="*" />
         </Grid.ColumnDefinitions>
         <TextBlock Grid.Column="0"
+                   Text="{Binding Path=DisplayRegion}"
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Fixed

src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
georgii-borovinskikh-sonarsource: Fixed

src/ConnectedMode.UnitTests/UI/ConnectionDisplay/ConnectionNameViewModelTests.cs
```diff
@@ -85,10 +92,64 @@ public void DisplayName_SonarCloud_NoId_HasRegion_ReturnsUrl(CloudServerRegion r
...
+    public void DisplayRegion_SonarQube_DoesNotDisplayRegion()
+    {
+        testSubject.ConnectionInfo = new ConnectionInfo("any id", ConnectionServerType.SonarQube);
+
+        VerifyDoesNotDisplayRegion();
+    }
+
+    [DynamicData(nameof(Regions))]
+    [DataTestMethod]
+    public void DisplayRegion_SonarCloud_NoId_DoesNotDisplayRegion(CloudServerRegion region)
```
gabriela-trutan-sonarsource, line 120:
```suggestion
    public void DisplayRegion_SonarCloud_NoId_DisplaysRegion(CloudServerRegion region)
```

gabriela-trutan-sonarsource (approved):

PR #6018, opened 2025-02-12
SLVS-1841 Fix flaky tests

georgii-borovinskikh-sonarsource:


src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
```diff
@@ -46,32 +46,42 @@ internal sealed class FileAnalysisTestsRunner : IDisposable
...
-        slCoreTestRunner.Start();
-
-        activeConfigScopeTracker = new ActiveConfigScopeTracker(slCoreTestRunner.SLCoreServiceProvider,
+    public static async Task <FileAnalysisTestsRunner> CreateInstance(string testClassName, Dictionary<string, StandaloneRuleConfigDto> initialRuleConfig = null)
+    {
+        var runner = new FileAnalysisTestsRunner(testClassName, initialRuleConfig);
+        await runner.slCoreTestRunner.Start(runner.rpcLogger);
+        runner.activeConfigScopeTracker = new ActiveConfigScopeTracker(runner.slCoreTestRunner.SLCoreServiceProvider,
             new AsyncLockFactory(),
             new NoOpThreadHandler());
```
vnaskos-sonar, line 83: The initialization of `ActiveConfigScopeTracker` could go back to the constructor and become again `readonly`

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
```diff
@@ -79,61 +83,81 @@ public void MockInitialSlCoreRulesSettings(Dictionary<string, StandaloneRuleConf
         slCoreRulesSettingsProvider.GetSLCoreRuleSettings().Returns(rulesSettings ?? []);
     }
 
-    public void Start()
+    public async Task Start(TestLogger testLogger)
+    {
+        using (await GlobalSloopStartLock.AcquireAsync())
```
vnaskos-sonar, line 88: Is the lock necessary? If so, documenting the reason would be nice for the future.

src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
georgii-borovinskikh-sonarsource: Nope, if it's done before .Start, it caches a null (uninitialized) instance of the service provider and it breaks

src/SLCore.IntegrationTests/SLCoreTestRunner.cs
georgii-borovinskikh-sonarsource: Not sure. I can try to remove it without changing the timeout inside. The idea was to decrease the chance of 2 slcore processes interfering with each other during parallel startup

src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
vnaskos-sonar: I feel like this should be either documented or enforced by the code. It's pretty easy to abuse it.

vnaskos-sonar (approved):
I've used the "Run unit tests until fail" for quite some time and I haven't got any failure.

PR #6046, opened 2025-02-20
SLVS-1883 Add RelativePathHelper for calculation of relative file paths to given root path

georgii-borovinskikh-sonarsource:
[SLVS-1883](https://sonarsource.atlassian.net/browse/SLVS-1883)

Part of SLVS-1789

[SLVS-1883]: https://sonarsource.atlassian.net/browse/SLVS-1883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SLCore.UnitTests/Common/Helpers/RelativePathHelperTests.cs
+++ b/src/SLCore.UnitTests/Common/Helpers/RelativePathHelperTests.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.IO;
 using SonarLint.VisualStudio.SLCore.Common.Helpers;
 
 namespace SonarLint.VisualStudio.SLCore.UnitTests.Common.Helpers;
@@ -28,8 +29,10 @@ public class RelativePathHelperTests
     [DataRow("C:\\", "D:\\file.json", null)]
     [DataRow("C:\\", "C:\\file.json", "file.json")]
     [DataRow("C:\\one\\", "C:\\one\\file.json", "file.json")]
+    [DataRow("C:\\one\\", "C:\\file.json", "..\\file.json")]
     [DataRow("C:\\one\\", "C:\\onetwo\\file.json", "..\\onetwo\\file.json")]
     [DataRow("C:\\one\\two\\", "C:\\one\\two\\file.json", "file.json")]
+    [DataRow("C:\\one\\two\\", "C:\\one\\file.json", "..\\file.json")]
     [DataRow("C:\\one\\two\\", "C:\\one\\twothree\\file.json", "..\\twothree\\file.json")]
     [DataRow("C:\\one\\two\\", "C:\\twothree\\file.json", "..\\..\\twothree\\file.json")]
     [DataRow("C:\\one\\two\\", "C:\\oneone\\twothree\\file.json", "..\\..\\oneone\\twothree\\file.json")]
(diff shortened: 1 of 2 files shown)


src/SLCore/Common/Helpers/RelativePathHelper.cs
```diff
@@ -0,0 +1,127 @@
...
+            throw new ArgumentException(string.Format(SLCoreStrings.RelativePathHelper_NonAbsolutePath, filePath), nameof(filePath));
+        }
+
+        if (root[root.Length - 1] != Separator)
+        {
+            throw new ArgumentException(string.Format(SLCoreStrings.RelativePathHelper_RootDoesNotEndWithSeparator, Separator), nameof(root));
+        }
+    }
+
+    private static bool IsPathFullyQualified(string path)
```
vnaskos-sonar, line 65: This is not fully correct.
Eg. `IsPathFullyQualified` for `[DataRow("network\\one\\two\\three\\", "network\\file.json", "..\\..\\..\\file.json")]` returns true, although neither the root or the filePath are fully qualified.

src/SLCore/Common/Helpers/RelativePathHelper.cs
```diff
@@ -0,0 +1,127 @@
...
+            if (root[index] == Separator)
+            {
+                depth++;
+            }
+        }
+
+        return depth;
+    }
+
+    private static int CalculateCommonPathLength(string path1, string path2)
```
vnaskos-sonar, line 100: This is quite complicated to understand and I think it could be avoided by checking if filePath starts with root.

Something like this
```
    if (!filePath.StartsWith(root)
    {
        throw new ArgumentException("Root path must be the root of the file path.", nameof(root));
    }
```

src/SLCore.UnitTests/Common/Helpers/RelativePathHelperTests.cs
```diff
@@ -0,0 +1,74 @@
...
+    [DataTestMethod]
+    public void GetRelativePathToRootFolder_ReturnsExpectedValues(string root, string file, string expected) =>
+        RelativePathHelper.GetRelativePathToRootFolder(root, file).Should().Be(expected);
+
+    [TestMethod]
+    public void GetRelativePathToRootFolder_RootPathNotEndsWithSeparator_Throws()
+    {
+        var act = () => RelativePathHelper.GetRelativePathToRootFolder("C:\\dirwithoutseparatorattheend", "C:\\dirwithoutseparatorattheend\\file.json");
+
+        act.Should().Throw<ArgumentException>();
```
vnaskos-sonar, line 48: All of the 3 validations in `Validate` are throwing `ArgumentException` but in those tests we are not sure which one failed. Given that the test should fail for the right reasons, we should also check for the message.

src/SLCore/Common/Helpers/RelativePathHelper.cs
```diff
@@ -0,0 +1,127 @@
...
+    public static string GetRelativePathToRootFolder(string root, string filePath)
+    {
+        Validate(root, filePath);
+
+        var commonPathLength = CalculateCommonPathLength(root, filePath);
+
+        if (commonPathLength == 0)
+        {
+            return null;
+        }
```
vnaskos-sonar, line 39: If I understood correctly the following should be true during validation. Good candidates for unit test as well
```
        // root is not null
        // root is longer than one character
        // root Path.IsPathRooted
        // root is a directory
        // root is fully qualified/absolute path
        // root is ending in separator
        // root is the root of filePath

        // filePath is not null
        // filePath is longer than one character
        // filePath Path.IsPathRooted
        // filePath is not a directory (not sure about that if we want a generic implementation)
        // filePath is fully qualified/absolute path
```

vnaskos-sonar (changes requested):

src/SLCore/Common/Helpers/RelativePathHelper.cs
georgii-borovinskikh-sonarsource: 1) this is not true for file paths with `..\`
2) the length itself is needed

src/SLCore/Common/Helpers/RelativePathHelper.cs
georgii-borovinskikh-sonarsource: The example you provided is not the same as the test case:
`[DataRow("\\\\network\\one\\two\\three\\", "\\\\network\\file.json", "..\\..\\..\\file.json")]`

More info: https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#unc-paths

src/SLCore/Common/Helpers/RelativePathHelper.cs
georgii-borovinskikh-sonarsource: I can add some comments for explanation

src/SLCore/Common/Helpers/RelativePathHelper.cs
vnaskos-sonar: No, it's another test case. `network\one\two\three\` It doesn't start with a drive letter neither with the `//` for network paths. So it shouldn't be considered a fully qualified name.

src/SLCore/Common/Helpers/RelativePathHelper.cs
vnaskos-sonar: I understood that filePath is also an absolute path.
Do you mean a case like this?
root = "C:\folder1\folder2"
filePath -> "C:\folder1\..\folder1\folder2"

georgii-borovinskikh-sonarsource: [DataRow("C:\\one\\two\\", "C:\\one\\twothree\\file.json", "..\\twothree\\file.json")]

paths have common prefix but this prefix is not equal to the root path

vnaskos-sonar: The root always ends with a path separator so `C:\one\twothree\file.json` does not start with `C:\one\two\` 😉

georgii-borovinskikh-sonarsource: Discussed offline: the code is correct, as it supports relative paths with `..`

src/SLCore/Common/Helpers/RelativePathHelper.cs
georgii-borovinskikh-sonarsource: Can't repro, added a test case for `_Throws` methods

PR #6049, opened 2025-02-21
SLVS-1887 Use RelativePathHelper in ClientFileDtoFactory, skip files that can't be relativized

georgii-borovinskikh-sonarsource:
[SLVS-1887](https://sonarsource.atlassian.net/browse/SLVS-1887)

Part of SLVS-1789

[SLVS-1887]: https://sonarsource.atlassian.net/browse/SLVS-1887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SLCore/Common/Helpers/ClientFileDtoFactory.cs
```diff
@@ -27,17 +27,24 @@ namespace SonarLint.VisualStudio.SLCore.Common.Helpers;
...
+[method: ImportingConstructor]
+public class ClientFileDtoFactory(ILogger logger) : IClientFileDtoFactory
 {
-    public ClientFileDto Create(string configScopeId, string rootPath, SourceFile sourceFile)
+    private readonly ILogger logger = logger.ForVerboseContext(nameof(ClientFileDtoFactory));
+
+    public ClientFileDto CreateOrNull(string configScopeId, string rootPath, SourceFile sourceFile)
     {
-        var ideRelativePath = GetRelativePath(rootPath, sourceFile.FilePath);
+        var ideRelativePath = RelativePathHelper.GetRelativePathToRootFolder(rootPath, sourceFile.FilePath);
```
gabriela-trutan-sonarsource, line 37: It is crashing when rootPath is null. Please add a check and a test for this case.

gabriela-trutan-sonarsource (changes requested):
Why should we change our implementation to provide the relative path? We used to provide the absolute path and everything was working fine. Adding our own implementation of relative paths, that might be error prone, doesn't seem to bring any benefit.

src/SLCore.UnitTests/Common/Helpers/ClientFileDtoFactoryTests.cs
```diff
@@ -28,34 +28,44 @@ namespace SonarLint.VisualStudio.SLCore.UnitTests.Common.Helpers;
...
+    [TestInitialize]
+    public void TestInitialize()
     {
-        MefTestHelpers.CheckIsSingletonMefComponent<ClientFileDtoFactory>();
+        testLogger = new TestLogger();
+        testSubject = new ClientFileDtoFactory(testLogger);
     }
 
+    [TestMethod]
+    public void MefCtor_CheckIsExported() =>
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I only added it as a verbose context, so I don't think the test is needed

src/SLCore.Listeners/Implementation/ListFilesListener.cs
```diff
@@ -59,13 +59,24 @@ public Task<ListFilesResponse> ListFilesAsync(ListFilesParams parameters)
...
         }
 
+        private void AddWorkspaceFilesDtos(
+            ListFilesParams parameters,
+            List<ClientFileDto> clientFileDtos,
+            string root,
+            IReadOnlyCollection<string> fullFilePathList) =>
+            clientFileDtos.AddRange(
+                fullFilePathList
+                    // bug
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: forgot to remove it

src/SLCore.UnitTests/Common/Helpers/ClientFileDtoFactoryTests.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Why not?

src/Integration/LocalServices/FileTracker.cs
```diff
@@ -69,15 +73,18 @@ public void RenameFiles(string[] beforeRenameFiles, SourceFile[] afterRenameFile
 
     private void NotifySlCoreFilesChanged(string[] removedFiles, SourceFile[] addedOrChangedFiles)
     {
-        if (serviceProvider.TryGetTransientService(out IFileRpcSLCoreService fileRpcSlCoreService) && activeConfigScopeTracker.Current is {} configScope)
+        if (serviceProvider.TryGetTransientService(out IFileRpcSLCoreService fileRpcSlCoreService) && activeConfigScopeTracker.Current is { RootPath: not null } configScope)
```
gabriela-trutan-sonarsource, line 76: Should we really skip the logic when RootPath is null? What if the RootPath is null, but there are some files that should be removed from SlCore tracking.

gabriela-trutan-sonarsource, line 76: In the `else ` case we log that the service is unavailable, even if the service is available, but the rootPath is null, which is wrong.

gabriela-trutan-sonarsource (approved):
I do have some feedback. Please take a look at them before merging the PR

src/Integration/LocalServices/FileTracker.cs
georgii-borovinskikh-sonarsource: If root path is null that generally means list files has not been called for this config scope, we don't need to send update events before that (as the up-to-date state will be requested and calculated later, along with the root, during list files)

PR #6050, opened 2025-02-21
SLVS-1795 Add shared binding file to the `listFiles` response

georgii-borovinskikh-sonarsource:
[SLVS-1795](https://sonarsource.atlassian.net/browse/SLVS-1795)

[SLVS-1795]: https://sonarsource.atlassian.net/browse/SLVS-1795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SLCore.Listeners/Implementation/ListFilesListener.cs
```diff
@@ -20,76 +20,69 @@
...
-                    var root = GetRoot(fullFilePathList.First());
-                    if (activeConfigScopeTracker.TryUpdateRootOnCurrentConfigScope(parameters.configScopeId, root))
-                    {
-                        AddWorkspaceFilesDtos(parameters, clientFileDtos, root, fullFilePathList);
-                    }
-                }
-            }
-            return Task.FromResult(new ListFilesResponse(clientFileDtos));
-        }
+    private IEnumerable<string> GetAllFiles(IReadOnlyCollection<string> workspaceFilePaths) =>
```
gabriela-trutan-sonarsource, line 54: The name is very confusing and difficult to distinguish between it and `GetWorkspaceFiles'. I would rather combine them together.

src/SLCore.Listeners.UnitTests/Implementation/ListFilesListenerTests.cs
```diff
@@ -174,6 +177,43 @@ public async Task ListFilesAsync_SolutionWorkSpace_UpdatesRootPath(string filePa
             activeConfigScopeTracker.Received(1).TryUpdateRootOnCurrentConfigScope(ConfigScopeId, expectedRootPath);
         }
 
+        [TestMethod]
```
gabriela-trutan-sonarsource, line 180: We are missing the case for shared binding config not present

src/SLCore.Listeners.UnitTests/Implementation/ListFilesListenerTests.cs
```diff
@@ -174,6 +177,43 @@ public async Task ListFilesAsync_SolutionWorkSpace_UpdatesRootPath(string filePa
...
+        public async Task ListFilesAsync_SharedBindingPresentButNotConverted_IgnoresSharedBinding()
+        {
+            SetUpFolderWorkSpaceService(null);
```
gabriela-trutan-sonarsource, line 183: This tests are very crowded and difficult to understand. Some of the setup can be extracted into other methods.
E.g. ` filePaths`, `clientFileDtos` and `SetUpSolutionWorkspaceService` can be a separate method

src/ConnectedMode/Shared/ISharedBindingConfigProvider.cs
```diff
@@ -22,10 +22,10 @@ namespace SonarLint.VisualStudio.ConnectedMode.Shared
 {
     public interface ISharedBindingConfigProvider
     {
-        bool HasSharedBinding();
-
         SharedBindingConfigModel GetSharedBinding();
 
         string SaveSharedBinding(SharedBindingConfigModel sharedBindingConfigModel);
+
+        string GetSharedBindingFilePathOrNull();
```
gabriela-trutan-sonarsource, line 29: Why is this not the responsibility of the `ISharedBindingConfigFileProvider`? This class does not have a dependency to the `IFileSystemService`, but we do a lot of path calculation. Also now there is the awkward method `Exists` from  `ISharedBindingConfigFileProvider` that simply wraps the call to `IFileSystemService`.

src/ConnectedMode/Shared/SharedBindingConfigProvider.cs
```diff
@@ -20,89 +20,88 @@
...
+                : null;
 
-        public SharedBindingConfigModel GetSharedBinding()
+        public string GetSharedBindingFilePathOrNull()
         {
-            var sonarLintFolder = sharedFolderProvider.GetSharedFolderPath();
-            if (sonarLintFolder == null)
+            var sharedBindingFilePath = CalculateSharedBindingPathUnderExistingFolder();
+
+            if (sharedBindingFilePath == null)
```
gabriela-trutan-sonarsource, line 51: Shouldn't we log this case?

src/ConnectedMode/Shared/SharedBindingConfigProvider.cs
```diff
@@ -20,89 +20,88 @@
...
+        {
+            var sonarLintFolder = sharedFolderProvider.GetSharedFolderPath();
+            if (sonarLintFolder == null)
             {
-                var gitRepoDir = gitWorkspaceService.GetRepoRoot();
-                if (gitRepoDir != null)
-                {
-                    fileSavePath = Path.Combine(gitRepoDir, sharedFolderName, SharedFileName);
-                }
+                logger.WriteLine(Resources.SharedBindingConfigProvider_SharedFolderNotFound);
```
gabriela-trutan-sonarsource, line 85: When this method is called by the `ChooseNewSharedBindingLocation`, why don't we just create the shared folder?

gabriela-trutan-sonarsource (changes requested):

src/ConnectedMode/Resources.resx
```diff
@@ -280,11 +280,20 @@
   <data name="TimedUpdateTriggered" xml:space="preserve">
     <value>[ConnectedMode/TimedUpdate] Triggering refresh of server settings...</value>
   </data>
-  <data name="SharedBindingConfigProvider_SavePathNotFound" xml:space="preserve">
-    <value>[SharedBindingConfigProvider] There's no .sonarlint shared folder or solution is not under git</value>
+  <data name="SharedBindingConfigProvider_LogContext" xml:space="preserve">
+    <value>Shared Binding</value>
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Too many words for a log context

src/ConnectedMode/Shared/SharedBindingConfigFileProvider.cs
```diff
@@ -20,96 +20,88 @@
...
+        catch (Exception ex)
+        {
+            logger.LogVerbose($"[SharedBindingConfigFileProvider] Unable to read shared config file: {ex.Message}");
+        }
+
+        return null;
+    }
 
-        public bool WriteSharedBindingConfigFile(string filePath, SharedBindingConfigModel sharedBindingConfigModel)
+    public bool Exists(string filePath) =>
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Path calculation is currently the responsibility of `SharedBindingConfigProvider`, `ISharedBindingConfigFileProvider` is responsible for reading/writing the file. It feels like the right place to also check for the file's existance.

src/ConnectedMode/Resources.resx
```diff
@@ -280,11 +280,20 @@
...
+  </data>
+  <data name="SharedBindingConfigProvider_GitRootNotFound" xml:space="preserve">
+    <value>Git root not found, can't create .sonarlint shared folder</value>
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: if the git root is found, it then calls `SharedBindingConfigFileProvider.Write` which then tries to create the folder

src/SLCore.Listeners.UnitTests/Implementation/ListFilesListenerTests.cs
georgii-borovinskikh-sonarsource: That's the default case for all other tests in this class. SHould I just add an assertion in one of those?

src/ConnectedMode/Shared/ISharedBindingConfigProvider.cs
georgii-borovinskikh-sonarsource: FileProvider is currently limited int it's responsibility to file system operations and deserialization. I don't want to break that split.

src/ConnectedMode/Shared/SharedBindingConfigProvider.cs
georgii-borovinskikh-sonarsource: It's logged inside the CalculateSharedBindingPathUnderExistingFolder

src/ConnectedMode/Shared/SharedBindingConfigProvider.cs
georgii-borovinskikh-sonarsource: Because it checks for an existing folder, not creates a new one. A new one can only be created if we have a git root, which is done in another method

src/ConnectedMode/Resources.resx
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: That's misleading

src/SLCore.Listeners.UnitTests/Implementation/ListFilesListenerTests.cs
gabriela-trutan-sonarsource: Sure.

src/ConnectedMode/Resources.resx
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Yes, but the log should be adapted. This is not correct with the execution path. Any method can call `CalculateSharedBindingPathUnderGitRoot` not until the SharedBindingConfigFileProvider.Write. The `CalculateSharedBindingPathUnderGitRoot` is an independent unit (it doesn't depend on its callers) so the logs should reflect that.

src/SLCore.Listeners/Implementation/ListFilesListener.cs
```diff
@@ -20,76 +20,69 @@
...
-        public ListFilesListener(IFolderWorkspaceService folderWorkspaceService, ISolutionWorkspaceService solutionWorkspaceService,
-            IActiveConfigScopeTracker activeConfigScopeTracker, IClientFileDtoFactory clientFileDtoFactory)
-        {
-            this.folderWorkspaceService = folderWorkspaceService;
-            this.solutionWorkspaceService = solutionWorkspaceService;
-            this.activeConfigScopeTracker = activeConfigScopeTracker;
-            this.clientFileDtoFactory = clientFileDtoFactory;
-        }
+    private List<ClientFileDto> GetFilesList(ListFilesParams parameters) =>
+        activeConfigScopeTracker.Current.Id == parameters.configScopeId
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: >Getting all files
>This can include the current GetWorkspaceFiles

Root calculation needs the result of GetWorkspaceFiles in solution mode

src/ConnectedMode/Shared/ISharedBindingConfigProvider.cs
gabriela-trutan-sonarsource: It is not limited. It has full access to file system operation and that's why the method should go to `ISharedBindingConfigFileProvider`. The big elephant in the room is the `Exists` method

src/ConnectedMode/Resources.resx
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: >Any method can call CalculateSharedBindingPathUnderGitRoot

I can rename it to `CalculatePotentialSharedBindingPathUnderGitRoot` to make it obvious what this method is used for

src/ConnectedMode/Shared/SharedBindingConfigProvider.cs
gabriela-trutan-sonarsource: No it is not. The Path.Combine from CalculateSharedBindingPathUnderExistingFolder can lead to a file that does not exist. That's why we have the null check here and that's why a log is missing

src/ConnectedMode/Shared/SharedBindingConfigProvider.cs
gabriela-trutan-sonarsource: But what if the project is not in a git repo? I find it silly that we create the folder for the case that is not a  git root, but we fail to do so if the shared folder does not exist. It is not at all user friendly that the user has to manually create its shared folder to be able to shared it's config file, when we could do that easily.

src/ConnectedMode/Shared/SharedBindingConfigProvider.cs
georgii-borovinskikh-sonarsource: >CalculateSharedBindingPathUnderExistingFolder can lead to a file that does not exist

No, that's not how it works. It returns null if the folder does not exist. The file existance check is done below:

```
            if (sharedBindingFilePath == null)
            {
                return null;
            }

            if (!sharedBindingConfigFileProvider.Exists(sharedBindingFilePath))
            {
```

src/ConnectedMode/Resources.resx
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: I don't get why you don't just adapt the log resource or move it to a better place

src/SLCore.Listeners/Implementation/ListFilesListener.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: I see. Anyway, please refactor those ugly conditions and consider adding logs when they fail

src/ConnectedMode/Shared/SharedBindingConfigProvider.cs
georgii-borovinskikh-sonarsource: That's a branch that was cut from the implementation. The git repo case was obvious and easy to implement. Other cases required more effort and were not included. So far there were no user requests for non-git cases

src/ConnectedMode/Resources.resx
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Because it's still true. That check checks if we can create a folder under git root, and if there's no git root, then we obviously can't create the folder.

src/ConnectedMode/Shared/SharedBindingConfigProvider.cs
gabriela-trutan-sonarsource: It is not more effort to just create a folder and I think its lack makes the usability very bad. The fact that you touched and refactored the class I found as an opportunity for improvement.

src/ConnectedMode/Shared/ISharedBindingConfigProvider.cs
georgii-borovinskikh-sonarsource: It is. ISharedBindingConfigFileProvider does not do any path calculation. That is, at the moment, the responsibility of ISharedBindingConfigProvider. ISharedBindingConfigFileProvider is a wrapper around file system operations and serialization only

src/ConnectedMode/Shared/SharedBindingConfigProvider.cs
```diff
@@ -20,89 +20,84 @@
...
+            var sharedBindingFilePath = CalculateSharedBindingPathUnderExistingFolder();
 
-        public SharedBindingConfigModel GetSharedBinding()
-        {
-            var sonarLintFolder = sharedFolderProvider.GetSharedFolderPath();
-            if (sonarLintFolder == null)
+            if (sharedBindingFilePath == null || !sharedBindingConfigFileProvider.Exists(sharedBindingFilePath))
             {
-                logger.WriteLine(Resources.SharedBindingConfigProvider_SharedFolderNotFound);
+                logger.WriteLine(Resources.SharedBindingConfigProvider_SharedFileNotFound, sharedBindingFilePath ?? Resources.SharedBindingConfigProvider_SharedFileNotFound_ProbePathDefaultValue);
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: I would say that yes, but I don't have a strong opinion. I like the new wording!

PR #6048, opened 2025-02-21
SLVS-1884 Add RPC method helpGenerateUserToken

gabriela-trutan-sonarsource:
Add the RPCs method helpGenerateUserToken and OpenUrlInBrowser, needed for generating a token

[SLVS-1884](https://sonarsource.atlassian.net/browse/SLVS-1884)

[SLVS-1884]: https://sonarsource.atlassian.net/browse/SLVS-1884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6052, opened 2025-02-24
SLVS-1885 Add GenerateToken to ISlCoreConnectionAdapter

gabriela-trutan-sonarsource:
This targets the [PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/6048) in which the RPC methods are introduced


src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -72,6 +73,7 @@ public class SlCoreConnectionAdapter(ISLCoreServiceProvider serviceProvider, ITh
 {
     private static readonly AdapterResponseWithData<List<OrganizationDisplay>> FailedResponseWithData = new(false, []);
     private static readonly AdapterResponse FailedResponse = new(false);
+    private readonly ILogger logger = logger.ForContext(nameof(SlCoreConnectionAdapter));
```
georgii-borovinskikh-sonarsource, line 76: I don't think we should use class names as non-verbose context. Could you move it to verbose context and add a more general context here like `New Server Connection`?

georgii-borovinskikh-sonarsource (approved):

src/ConnectedMode/SlCoreConnectionAdapter.cs
gabriela-trutan-sonarsource: Well, if it will become a verbose context, the class name can stay.

gabriela-trutan-sonarsource: Also this adapter is not used only for new server connection, more reason for the class name to stay.

georgii-borovinskikh-sonarsource: Sure, but you could think of another user-facing name. I understand that my suggestion might not be the best

gabriela-trutan-sonarsource: I can't come up with a good generic name. And I think the class name is OK

PR #6055, opened 2025-02-25
SLVS-1792 Show TrustConnectionDialog on assistCreatingConnection

gabriela-trutan-sonarsource:
[SLVS-1792](https://sonarsource.atlassian.net/browse/SLVS-1792)

Show `TrustConnectionDialog` on `assistCreatingConnection `and make sure the connection is added to the repository, if the server becomes trusted.
This PR deals does not include showing the `CredentialsDialog `when the received token is null.
This targets the [PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/6054) in which the dialog is created

[SLVS-1792]: https://sonarsource.atlassian.net/browse/SLVS-1792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SLCore/Service/Connection/Models/SonarCloudRegionExtensions.cs
```diff
@@ -37,4 +37,13 @@ public static SonarCloudRegion ToSlCoreRegion(this CloudServerRegion region)
         }
         throw new ArgumentOutOfRangeException(region.Name);
     }
+
+    public static CloudServerRegion ToCloudServerRegion(this SonarCloudRegion slCoreRegion)
+    {
+        if (CoreToSlCoreLanguageMap.FirstOrDefault(x => x.Value == slCoreRegion) is { Key: not null } kvp)
```
georgii-borovinskikh-sonarsource, line 43: would be cleaner to create another dictionary

src/SLCore.Listeners/Implementation/ConnectedModeSuggestionListener.cs
```diff
@@ -20,32 +20,42 @@
...
-            this.bindingSuggestionHandler = bindingSuggestionHandler;
-            this.activeConfigScopeTracker = activeConfigScopeTracker;
-        }
+        private readonly ILogger logger = logger.ForContext("Connected Mode Suggestion");
 
         public Task<AssistCreatingConnectionResponse> AssistCreatingConnectionAsync(AssistCreatingConnectionParams parameters)
         {
-            bindingSuggestionHandler.Notify();
+            var serverConnection = ConvertSeverConnection(parameters);
+            var token = parameters.connectionParams.Right?.tokenValue ?? parameters.connectionParams.Left?.tokenValue;
```
georgii-borovinskikh-sonarsource, line 47: why is token not set as credentials for ServerConnection?

src/ConnectedMode/UI/TrustConnection/TrustConnectionViewModel.cs
```diff
@@ -0,0 +1,55 @@
...
+    IConnectedModeServices connectedModeServices,
+    IProgressReporterViewModel progressReporterViewModel,
+    ServerConnection serverConnection,
+    SecureString token)
+    : ViewModelBase
+{
+    public IConnectedModeServices ConnectedModeServices { get; } = connectedModeServices;
+    public ServerConnection ServerConnection { get; } = serverConnection;
+    public SecureString Token { get; } = token;
+    public Connection Connection => ServerConnection.ToConnection();
```
georgii-borovinskikh-sonarsource, line 39: This is kind of confusing now. Is there plans to interact with this object or why do we need it at all? Adapter can be changed to accept ServerConnection as the primary (and have an extension method with Connection & .ToServerConnection() call inside)

georgii-borovinskikh-sonarsource (approved):

src/SLCore/Service/Connection/Models/SonarCloudRegionExtensions.cs
gabriela-trutan-sonarsource: I don't think it's cleaner to duplicate the data.

src/SLCore.Listeners/Implementation/ConnectedModeSuggestionListener.cs
gabriela-trutan-sonarsource: The token can be null and if set here the `TokenAuthCredentials` will throw. I think it doesn't make sense to deal with the credentials in this view, but later on, when the token can be ensured that is not null.

If I add the  `TokenAuthCredentials` here, when the token is not null, I would anyway need to check and create a new instance after the `CredentialsDialog` is instantiated and a new token is generated. So why split the logic in multiple places?

src/ConnectedMode/UI/TrustConnection/TrustConnectionViewModel.cs
gabriela-trutan-sonarsource: The object is being interacted with. The adapter accepts `Connection` model, because it adapts between UI model and business model. There is plan to drop the `Connection` model completely, so I would not change the adapter for now

PR #6059, opened 2025-02-26
SLVS-1901 Add TryGet overload to ServerConnectionsRepositoryAdapter with serverConnectionId

georgii-borovinskikh-sonarsource:
[SLVS-1901](https://sonarsource.atlassian.net/browse/SLVS-1901)

Part of SLVS-1789

[SLVS-1901]: https://sonarsource.atlassian.net/browse/SLVS-1901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode/ServerConnectionsRepositoryAdapter.cs
+++ b/src/ConnectedMode/ServerConnectionsRepositoryAdapter.cs
@@ -39,6 +39,8 @@ public interface IServerConnectionsRepositoryAdapter
     bool TryUpdateCredentials(Connection connection, ICredentialsModel credentialsModel);
 
     bool TryGet(string serverConnectionId, out ServerConnection serverConnection);
+
+    bool TryGet(ConnectionInfo connection, out ServerConnection serverConnection);
 }
 
 [Export(typeof(IServerConnectionsRepositoryAdapter))]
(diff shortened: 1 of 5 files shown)


src/ConnectedMode/ServerConnectionsRepositoryAdapter.cs
```diff
@@ -96,3 +93,9 @@ private static IConnectionCredentials MapCredentials(ICredentialsModel credentia
         }
     }
 }
+
+internal static class ServerConnectionsRepositoryAdapterExtensions
+{
+    public static bool TryGet(this IServerConnectionsRepositoryAdapter adapter, ConnectionInfo connection, out ServerConnection serverConnection) =>
```
gabriela-trutan-sonarsource, line 99: Please don't make this an extension method, but rather an overload. The purpose of the adapter is to adapt between UI models and business logic required by the `serverConnectionsRepository`. So it makes sense for the method to be part of the interface and visible, explicit.

gabriela-trutan-sonarsource (approved):

src/ConnectedMode/ServerConnectionsRepositoryAdapter.cs
georgii-borovinskikh-sonarsource: Extension methods are always visible though 🤔 Ok on the purpose argument, will change due to that

PR #6060, opened 2025-02-26
SLVS-1856 Clean up non-roslyn quality profiles

vnaskos-sonar:
[SLVS-1856](https://sonarsource.atlassian.net/browse/SLVS-1856)

Upgrading to the new version will not clean the `binding.config` to maintain backwards compatibility. The `binding.config` is not regenerated and in case one of missing entries, it will just not sync the missing entries. It will not break the Extension but there is a chance that missing languages from the `binding.config` will not be synchronized.

In this case a user can
* Update all instances of SQVS to the latest version.
* Unbind and re-bind the project to have a clean `binding.config`.

New bindings will only populate C# and VB.NET on the `binding.config`.

[SLVS-1856]: https://sonarsource.atlassian.net/browse/SLVS-1856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ConnectedMode/Binding/RoslynBindingConfigProvider.cs
```diff
@@ -0,0 +1,197 @@
...
+
+    public bool IsLanguageSupported(Language language) => languageProvider.RoslynLanguages.Contains(language);
+
+    public Task<IBindingConfig> GetConfigurationAsync(
+        SonarQubeQualityProfile qualityProfile,
+        Language language,
+        BindingConfiguration bindingConfiguration,
+        CancellationToken cancellationToken)
+    {
+        if (!IsLanguageSupported(language))
```
georgii-borovinskikh-sonarsource, line 75: There's a bit of a strange relationship between downloader and this. Feels like this check is not needed

vnaskos-sonar: True!
The Downloader, the ConfigProvider and the OutOfDateQualityProfileFinder are all of them checking for supported language. I'll keep the check just for safety, till there is a better solution.

src/ConnectedMode/Binding/IBindingConfigProvider.cs
```diff
@@ -18,39 +18,35 @@
...
 using SonarLint.VisualStudio.Core.Binding;
 using SonarQube.Client.Models;
 
-namespace SonarLint.VisualStudio.ConnectedMode.Binding
+namespace SonarLint.VisualStudio.ConnectedMode.Binding;
+
+/// <summary>
+/// Contract to provide the binding-related configuration for one or more languages
+/// </summary>
+public interface IBindingConfigProvider
```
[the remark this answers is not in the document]
vnaskos-sonar: Technically this is still generic, let's keep the interface as is.

georgii-borovinskikh-sonarsource (approved):

PR #6067, opened 2025-02-28
SLVS-1851 Drop IAnalysisIssueVisualization.IsSuppressed

vnaskos-sonar:
[SLVS-1851](https://sonarsource.atlassian.net/browse/SLVS-1851)

[SLVS-1851]: https://sonarsource.atlassian.net/browse/SLVS-1851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
+++ b/src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
@@ -220,7 +220,7 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.Vie
 
             var issueViz = ((ITaintIssueViewModel)viewModel).TaintIssueViz;
 
-            if (issueViz.IsSuppressed)
+            if (issueViz.Issue.IsResolved)
             {
                 return false;
             }
@@ -281,30 +281,15 @@ namespace SonarLint.VisualStudio.IssueVisualization.Security.Taint.TaintList.Vie
 
         private void UpdateIssues()
         {
-            foreach (var taintIssueViewModel in unfilteredIssues)
-            {
-                taintIssueViewModel.TaintIssueViz.PropertyChanged -= OnTaintIssuePropertyChanged;
-            }
-
             unfilteredIssues.Clear();
 
             foreach (var issueViz in store.GetAll())
             {
                 var taintIssueViewModel = new TaintIssueViewModel(issueViz);
                 unfilteredIssues.Add(taintIssueViewModel);
-
-                taintIssueViewModel.TaintIssueViz.PropertyChanged += OnTaintIssuePropertyChanged;
             }
 
             NotifyPropertyChanged(nameof(HasServerIssues));
-        }
-
-        private void OnTaintIssuePropertyChanged(object sender, PropertyChangedEventArgs e)
-        {
-            if (e.PropertyName == nameof(IAnalysisIssueVisualization.IsSuppressed))
-            {
-                UpdateCaptionAndListFilter();
-            }
         }
 
         private void UpdateCaptionAndListFilter()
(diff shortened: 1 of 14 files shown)


src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -281,32 +281,17 @@ private void SetCommands(ILocationNavigator locationNavigator)
 
         private void UpdateIssues()
         {
-            foreach (var taintIssueViewModel in unfilteredIssues)
```
gabriela-trutan-sonarsource, line 284: The `PropertyChanged ` of the `AnalysisIssueVisualization` is also invoked when the `CurrentFilePath` changes or the `SnapshotSpan` not only for IsResolved. Please revert this removal.

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
```diff
@@ -220,7 +220,7 @@ private bool NotSuppressedIssuesInCurrentDocumentFilter(object viewModel)
 
             var issueViz = ((ITaintIssueViewModel)viewModel).TaintIssueViz;
 
-            if (issueViz.Issue.IsResolved)
```
gabriela-trutan-sonarsource: If we anyway changed all the references to use the `Issue` why didn't we simply took the step to drop the `IsResolved` from `IAnalysisIssueVisualization` and instead we reverted the changes?  

I think we could revert the commit and simply drop the property. :)

gabriela-trutan-sonarsource (changes requested):

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
vnaskos-sonar: It was dropped because `TaintIssuesControlView` [reacts only to events with changes on the `IsSuppressed`](https://github.com/SonarSource/sonarlint-visualstudio/pull/6067/commits/b44d5ec814ae4e7f0da41b30a3be4de0b3803f75#diff-5891c711c331de7bcd2d88cfbf7baceb49ea4184b6b5e1f36252418ecc6a5955L304). However, this is not possible after the mute issue migration to sloop.

`IsSuppressed` has been replaced by `AnalysisIssue.IsResolved` which is initialized on the constructor and does not have a setter. The `IsResolved` cannot change and [the `PropertyChange` was removed](https://github.com/SonarSource/sonarlint-visualstudio/pull/6067/commits/0749d0acf704ec0deee837234e49b936c5340d49#diff-ced544057ec5c81f0116061c44d45be2746f2a70069cfb2bf3e2198931635a8eL79-L83)

gabriela-trutan-sonarsource: You are right. I missed [this ](https://github.com/SonarSource/sonarlint-visualstudio/pull/6067/commits/b44d5ec814ae4e7f0da41b30a3be4de0b3803f75#diff-5891c711c331de7bcd2d88cfbf7baceb49ea4184b6b5e1f36252418ecc6a5955L304)

src/IssueViz.Security/Taint/TaintList/ViewModels/TaintIssuesControlViewModel.cs
vnaskos-sonar: There were some cases where IAnalysisIssueVisualization has to expose the `IsResolved` (`issueViz.Issue.IsResolved`). In general this is an antipattern ([more details on object chaining](https://medium.com/@jordan.l.edmunds/stop-calling-methods-on-child-objects-2b9259bd549c)). The `IAnalysisIssueVisualization` should be responsible on what and how to expose data.

I can revert it, as our codebase tends to violate that quite often. WDYT?

gabriela-trutan-sonarsource: That's a good point about the object chaining. :) Let's make a decision together in the team.

gabriela-trutan-sonarsource: As discussed, we can leave it as it is.

gabriela-trutan-sonarsource (approved):

PR #6068, opened 2025-02-28
SLVS-1905 Adapt ManageBindingDialog for auto binding

georgii-borovinskikh-sonarsource:
[SLVS-1905](https://sonarsource.atlassian.net/browse/SLVS-1905)

Part of SLVS-1789

[SLVS-1905]: https://sonarsource.atlassian.net/browse/SLVS-1905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
+++ b/src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
@@ -26,9 +26,8 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.UI;
 public class AutomaticBindingRequestTests
 {
     [TestMethod]
-    public void Shared_Instance_NotNull()
+    public void Shared_TtpeName_ShouldBeShared()
     {
-        new AutomaticBindingRequest.Shared().Should().NotBeNull();
         new AutomaticBindingRequest.Shared().TypeName.Should().Be(Resources.AutomaticBindingType_Shared);
     }
 
--- a/src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
+++ b/src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
@@ -1,0 +1,51 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.ConnectedMode.UI;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.UI;
+
+[TestClass]
+public class AutomaticBindingRequestTests
+{
+    [TestMethod]
+    public void Shared_Instance_NotNull()
+    {
+        AutomaticBindingRequest.Shared.Current.Should().NotBeNull();
+        AutomaticBindingRequest.Shared.Current.TypeName.Should().Be(Resources.AutomaticBindingType_Shared);
+    }
+
+    public static object[][] AssistedBindingSubtypes =>
+    [
+        [true, Resources.AutomaticBindingType_Shared],
+        [false, Resources.AutomaticBindingType_Suggested]
+    ];
+    [DynamicData(nameof(AssistedBindingSubtypes))]
+    [DataTestMethod]
+    public void Assisted_ReturnsTypeNameDependingOnParameter(bool isShared, string typeName)
+    {
+        var testSubject = new AutomaticBindingRequest.Assisted("con id", "proj id", isShared);
+
+        testSubject.ServerConnectionId.Should().Be("con id");
+        testSubject.ServerProjectKey.Should().Be("proj id");
+        testSubject.TypeName.Should().Be(typeName);
+        testSubject.IsShared.Should().Be(isShared);
+    }
+}
--- a/src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
+++ b/src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
@@ -28,15 +28,10 @@ public class AutomaticBindingRequestTests
     [TestMethod]
     public void Shared_Instance_NotNull()
     {
-        AutomaticBindingRequest.Shared.Current.Should().NotBeNull();
-        AutomaticBindingRequest.Shared.Current.TypeName.Should().Be(Resources.AutomaticBindingType_Shared);
+        new AutomaticBindingRequest.Shared().Should().NotBeNull();
+        new AutomaticBindingRequest.Shared().TypeName.Should().Be(Resources.AutomaticBindingType_Shared);
     }
 
-    public static object[][] AssistedBindingSubtypes =>
-    [
-        [true, Resources.AutomaticBindingType_Shared],
-        [false, Resources.AutomaticBindingType_Suggested]
-    ];
     [DynamicData(nameof(AssistedBindingSubtypes))]
     [DataTestMethod]
     public void Assisted_ReturnsTypeNameDependingOnParameter(bool isShared, string typeName)
@@ -46,6 +41,12 @@ public class AutomaticBindingRequestTests
         testSubject.ServerConnectionId.Should().Be("con id");
         testSubject.ServerProjectKey.Should().Be("proj id");
         testSubject.TypeName.Should().Be(typeName);
-        testSubject.IsShared.Should().Be(isShared);
+        testSubject.IsFromSharedBinding.Should().Be(isShared);
     }
+
+    public static object[][] AssistedBindingSubtypes =>
+    [
+        [true, Resources.AutomaticBindingType_Shared],
+        [false, Resources.AutomaticBindingType_Suggested]
+    ];
 }
--- a/src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
+++ b/src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
@@ -26,7 +26,7 @@ namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.UI;
 public class AutomaticBindingRequestTests
 {
     [TestMethod]
-    public void Shared_TtpeName_ShouldBeShared()
+    public void Shared_TypeName_ShouldBeShared()
     {
         new AutomaticBindingRequest.Shared().TypeName.Should().Be(Resources.AutomaticBindingType_Shared);
     }
--- a/src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
+++ b/src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
@@ -33,6 +33,7 @@ using SonarLint.VisualStudio.ConnectedMode.UI.Resources;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Binding;
 using SonarLint.VisualStudio.TestInfrastructure;
+using SonarQube.Client.Models;
 
 namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.UI.ManageBinding;
 
@@ -43,9 +44,9 @@ public class ManageBindingViewModelTests
 
     private readonly SolutionInfoModel DefaultSolution = new SolutionInfoModel("Any.sln", default);
     private readonly SolutionInfoModel NoSolution = new SolutionInfoModel(null, default);
-    private readonly ServerProject serverProject = new("a-project", "A Project");
-    private readonly ConnectionInfo sonarQubeConnectionInfo = new("http://localhost:9000", ConnectionServerType.SonarQube);
-    private readonly ConnectionInfo sonarCloudConnectionInfo = new("organization", ConnectionServerType.SonarCloud);
+    private static readonly ServerProject serverProject = new("a-project", "A Project");
+    private static readonly ConnectionInfo sonarQubeConnectionInfo = new("http://localhost:9000", ConnectionServerType.SonarQube);
+    private static readonly ConnectionInfo sonarCloudConnectionInfo = new("organization", ConnectionServerType.SonarCloud);
     private readonly UsernameAndPasswordCredentials validCredentials = new("TOKEN", new SecureString());
     private readonly SharedBindingConfigModel sonarQubeSharedBindingConfigModel = new() { Uri = new Uri("http://localhost:9000"), ProjectKey = "myProj" };
     private readonly SharedBindingConfigModel sonarCloudSharedBindingConfigModel = new() { Organization = "myOrg", ProjectKey = "myProj" };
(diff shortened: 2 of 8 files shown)


src/ConnectedMode/UI/AutomaticBindingRequest.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace SonarLint.VisualStudio.ConnectedMode.UI;
+
+public abstract record AutomaticBindingRequest
+{
+    internal abstract string TypeName { get; }
+
+    private AutomaticBindingRequest() { }
+
+    public record Shared : AutomaticBindingRequest
```
gabriela-trutan-sonarsource, line 29: The name is not speaking. 

```suggestion
    public record SharedConfigurationRequest : BindingRequest
```

src/ConnectedMode/UI/AutomaticBindingRequest.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        public static readonly AutomaticBindingRequest Current = new Shared();
+        private Shared() { }
+
+        internal override string TypeName => ConnectedMode.Resources.AutomaticBindingType_Shared;
+    }
+
+    public record Assisted(
```
gabriela-trutan-sonarsource, line 37: I think here it should be "Automatic"

```suggestion
    public record AutomaticBindingRequest(
```

src/ConnectedMode/UI/AutomaticBindingRequest.cs
```diff
@@ -0,0 +1,44 @@
...
+        string ServerConnectionId,
+        string ServerProjectKey,
+        bool IsShared) : AutomaticBindingRequest
```
gabriela-trutan-sonarsource, line 40: Why is this bool here as well? This is very confusing. Consider renaming or add a documentation. By just looking at this records I didn't understand for which cases are they being used. I had to look at the callers to understand. This suggests that the namings are not so good and documentation is missing.

Also this field is not used yet.

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -366,4 +333,84 @@ private async Task<SolutionInfoModel> GetSolutionInfoModelAsync()
         var isFolderWorkspace = await connectedModeBindingServices.SolutionInfoProvider.IsFolderWorkspaceAsync();
         return new SolutionInfoModel(solutionName, isFolderWorkspace ? SolutionType.Folder : SolutionType.Solution);
     }
+
+    #region Automatic Binding Implementation
```
gabriela-trutan-sonarsource, line 337: We said we should avoid regions.

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -366,4 +333,84 @@ private async Task<SolutionInfoModel> GetSolutionInfoModelAsync()
...
+                return true;
+            default:
+                connectedModeServices.Logger.WriteLine(logContext, ConnectedMode.Resources.AutomaticBinding_ConfigurationNotAvailable);
+                serverConnectionId = null;
+                serverProjectKey = null;
+                return false;
+        }
+    }
+
+    private bool AutomaticBindingConnectionExists(MessageLevelContext logContext, string connectionId, out ServerConnection serverConnection)
```
gabriela-trutan-sonarsource, line 377: Please don't use out parameters.

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -366,4 +333,84 @@ private async Task<SolutionInfoModel> GetSolutionInfoModelAsync()
...
+        if (connectedModeServices.ServerConnectionsRepositoryAdapter.TryGet(connectionId, out serverConnection))
+        {
+            return true;
+        }
+
+        connectedModeServices.Logger.WriteLine(
+            logContext,
+            ConnectedMode.Resources.AutomaticBinding_ConnectionNotFound,
+            connectionId);
+        connectedModeServices.MessageBox.Show(
```
gabriela-trutan-sonarsource, line 388: Please don't use message box to inform the user about error, because this breaks the way we display erorrs in the connected mode dialogs.

Instead set the `ProgressReporter.WarningText` which will show a red text at the bottom of the dialog.

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -168,9 +168,12 @@ public async Task BindWithProgressAsync()
         await ProgressReporter.ExecuteTaskWithProgressAsync(bind);
     }
 
-    public async Task UseSharedBindingWithProgressAsync()
+    public async Task PerformAutomaticBindingWithProgressAsync(AutomaticBindingRequest automaticBinding)
```
gabriela-trutan-sonarsource, line 171:
```suggestion
    public async Task PerformBindingWithProgressAsync(AutomaticBindingRequest automaticBinding)
```

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
gabriela-trutan-sonarsource: I see the message box was there before. Then please create a sub-task to remove it.

src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
```diff
@@ -0,0 +1,51 @@
...
+public class AutomaticBindingRequestTests
+{
+    [TestMethod]
+    public void Shared_Instance_NotNull()
+    {
+        AutomaticBindingRequest.Shared.Current.Should().NotBeNull();
+        AutomaticBindingRequest.Shared.Current.TypeName.Should().Be(Resources.AutomaticBindingType_Shared);
+    }
+
+    public static object[][] AssistedBindingSubtypes =>
```
gabriela-trutan-sonarsource, line 35: Please move helper methods at the bottom of the class.

src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
```diff
@@ -949,29 +950,28 @@ public async Task CheckForSharedBindingAsync_WhenSharedBindingDoesNotExist_SetsN
     }
 
     [TestMethod]
-    public async Task UseSharedBindingWithProgressAsync_SharedBindingExistsAndValid_BindsProjectAndReportsProgress()
+    public async Task PerformAutomaticBindingWithProgressAsyncc_SharedBindingExistsAndValid_BindsProjectAndReportsProgress()
```
gabriela-trutan-sonarsource, line 953:
```suggestion
    public async Task PerformAutomaticBindingWithProgressAsync_SharedBindingExistsAndValid_BindsProjectAndReportsProgress()
```

src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
```diff
@@ -1099,15 +1123,98 @@ await bindingController.DidNotReceive()
...
+
+        await progressReporterViewModel.Received(1)
+            .ExecuteTaskWithProgressAsync(
+                Arg.Is<TaskToPerformParams<AdapterResponse>>(x =>
+                    x.ProgressStatus == UiResources.BindingInProgressText &&
+                    x.WarningText == UiResources.BindingFailedText &&
+                    x.AfterProgressUpdated == testSubject.OnProgressUpdated));
+    }
+
+    public static object[][] AssistedBindingParameters =>
```
gabriela-trutan-sonarsource, line 1154: Please move helper methods at the bottom of the class.

gabriela-trutan-sonarsource (changes requested):

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
georgii-borovinskikh-sonarsource: Same as this: https://github.com/SonarSource/sonarlint-visualstudio/pull/6068#discussion_r1975154049

src/ConnectedMode/UI/AutomaticBindingRequest.cs
georgii-borovinskikh-sonarsource: It should be used here: https://sonarsource.atlassian.net/browse/SLVS-1802

src/ConnectedMode/UI/AutomaticBindingRequest.cs
```diff
@@ -0,0 +1,44 @@
...
+
+public abstract record AutomaticBindingRequest
+{
+    internal abstract string TypeName { get; }
+
+    private AutomaticBindingRequest() { }
+
+    public record Shared : AutomaticBindingRequest
+    {
+        public static readonly AutomaticBindingRequest Current = new Shared();
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: It's so you don't have to type `new` each time

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
georgii-borovinskikh-sonarsource: This class is extremely big which is why I felt this is warranted here

src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
georgii-borovinskikh-sonarsource: This way it looks more like test cases

src/ConnectedMode/UI/AutomaticBindingRequest.cs
```diff
@@ -0,0 +1,44 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.ConnectedMode.UI;
+
+public abstract record AutomaticBindingRequest
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: BindingRequest is too generic. Automatic indicates that the parameters were not manually entered by the user

src/ConnectedMode/UI/AutomaticBindingRequest.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: As I said, it doesn't bring any value. Please drop it

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
georgii-borovinskikh-sonarsource: Also, in general, I disagree. `out` parameters are useful, especially when used in `bool Method(param1, out param1)` style

src/ConnectedMode/UI/AutomaticBindingRequest.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: I find it very confusing, because automatic for me means there is no user interaction. But there is user interaction as the user has to be click on the "use shared binding" button in the case of the Shared record. Please do not use automatic in the name.

src/ConnectedMode/UI/AutomaticBindingRequest.cs
gabriela-trutan-sonarsource: Please rename the bool to IsFromSharedBinding or add documentation.

src/ConnectedMode/UI/AutomaticBindingRequest.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Does it detract any value? What's the point of dropping it? It's just a singleton pattern

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
gabriela-trutan-sonarsource: But if we should decide we should avoid region, we should do so. If class is too big, you can move some logic into other class.

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
gabriela-trutan-sonarsource: They are not very easy to test and they tend to make the method parameters' list increase. You can avoid this by changing the return type to a model with multiple properties. Which is way more cleaner and testable

src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
gabriela-trutan-sonarsource: We decided to have all our helper methods at the bottom of the class. Please respect that

src/ConnectedMode/UI/AutomaticBindingRequest.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: There's no user interaction _after_. As opposed to the normal flow of the window, when it starts unbound and user clicks through options to create a binding. `BindingRequest` is a good name for a parameter of `IBindingController`, for example

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
georgii-borovinskikh-sonarsource: How is it not testable? Why does it matter for a private method? How do you chain requests with multiple return types? How do you indicate failure? Nullable tuples?

src/ConnectedMode/UI/AutomaticBindingRequest.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Yes. I have to have the knowledge that one model is a singleton and the other one not.

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
georgii-borovinskikh-sonarsource: This method is specific to automatic binding. `PerformBindingWithProgressAsync(BindingRequest request)` will make it even more confusing

src/ConnectedMode/UI/AutomaticBindingRequest.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: This is not clear from the name. This logic that you offer right now it's not obvious, it's implicit assumptions. Please rename or add documentation

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
gabriela-trutan-sonarsource: It is not easily testable, in general when using NSubstitue. I am not talking about this private method, but on a principle basis. Look how nasty the out parameters are mocked:
image


As I said, model:
`internal record ConnectionValidationResult(string connectionId, string serverProjectKey, bool result) { }`

Or you can create custom exceptions and handle the validation by catching/logging each exceptions.

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
gabriela-trutan-sonarsource: I don't find "automatic" a good naming as you could already guess from other comments. Please find another name

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
georgii-borovinskikh-sonarsource: This is just an unreasonable way to force personal preference.

Why should I create a custom return model for every method? Why should the code throw when it can not throw? Why are you suggesting to write java-style code in c# and not use the neat syntax that the language offers?

The mocking could possibly be an argument against using out parameters in interfaces, but it's definitely not an argument for creating custom models for private methods

gabriela-trutan-sonarsource: If you don't like creating custom models, you can try refactoring the logic. 

> Why should the code throw when it can not throw?
Well, if the connection is invalid, then it is a reason to fail the workflow. 

>Why are you suggesting to write java-style code in c# and not use the neat syntax that the language offers?
Not all syntax are clean or beneficial and, as mentioned, out parameters lead to long parameters list (as we have many examples in our code base) and are difficult to mock. I am just suggestion to use another syntax.

I don't think it is a personal preference, but we can discuss this in the team as well and decide together if we like the syntax or not

georgii-borovinskikh-sonarsource: It's a private method in a non-throwing method pattern. Why would I create models or insta-catch exceptions?

gabriela-trutan-sonarsource: Because you said:

Also, in general, I disagree. out parameters are useful, especially when used in bool Method(param1, out param1) style

So I brought the discussion about the general and why out parameters should be avoided

georgii-borovinskikh-sonarsource: Context matters. When it's a private method with an inherently limited number of parameters, it's fine to use out parameters.

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
georgii-borovinskikh-sonarsource: https://sonarsource.atlassian.net/browse/SLVS-1912

src/ConnectedMode/UI/AutomaticBindingRequest.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: https://sonarsource.atlassian.net/browse/SLVS-1912

gabriela-trutan-sonarsource (changes requested):
I understand you don't want to rename the methods and records. But then please add documentation to them to explain in which context and how they are used. 
The other changes look fine.

src/ConnectedMode.UnitTests/UI/AutomaticBindingRequestTests.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using SonarLint.VisualStudio.ConnectedMode.UI;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.UI;
+
+[TestClass]
+public class AutomaticBindingRequestTests
+{
+    [TestMethod]
+    public void Shared_TtpeName_ShouldBeShared()
```
gabriela-trutan-sonarsource:
```suggestion
    public void Shared_TypeName_ShouldBeShared()
```

PR #6072, opened 2025-02-28
SLVS-1911   Swap dependencies between Core and SonarQube.Client

gabriela-trutan-sonarsource:


src/CFamily/packages.lock.json
```diff
@@ -124,16 +124,6 @@
           "Microsoft.VisualStudio.Interop": "17.0.31902.203"
         }
       },
-      "Google.Protobuf": {
```
georgii-borovinskikh-sonarsource, line 127: finally

georgii-borovinskikh-sonarsource (approved):

PR #6085, opened 2025-03-05
SLVS-1816 Remove multiple source of truths for SQC Uri

gabriela-trutan-sonarsource:
It’s just important to not have the hard-coded strings that contain the SonarCloud uri split in multiple classes, because every time we have to adapt the URI we will have to touch all those classes (if it happened once, it will happen again). 

Removed all the hard-coded strings that contain the SonarCloud uri and use instead the property from `CloudServerRegion`.

[SLVS-1816](https://sonarsource.atlassian.net/browse/SLVS-1816)

[SLVS-1816]: https://sonarsource.atlassian.net/browse/SLVS-1816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Binding/CloudServerRegion.cs
```diff
@@ -25,8 +25,8 @@ public sealed class CloudServerRegion
     public const string EuRegionName = "EU";
     public const string UsRegionName = "US";
 
-    public static readonly CloudServerRegion Eu = new(EuRegionName, new("https://sonarcloud.io"));
-    public static readonly CloudServerRegion Us = new(UsRegionName, new("https://us.sonarcloud.io"));
+    public static readonly CloudServerRegion Eu = new(EuRegionName, new("https://sonarcloud.io/"));
```
georgii-borovinskikh-sonarsource, line 28: 🤔

georgii-borovinskikh-sonarsource (approved):

PR #6090, opened 2025-03-06
SLVS-1930 Show multi-region info depending on setting

gabriela-trutan-sonarsource:
This targets the [PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/6088) in which the `SonarLintSettings` service has been added to the `ConnectedModeUiServices`, which is needed to access the show region setting

### DIFF
--- a/src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
+++ b/src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
@@ -18,22 +18,27 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.WPF;
 
 namespace SonarLint.VisualStudio.ConnectedMode.UI.ConnectionDisplay;
 
 public class ConnectionNameViewModel : ViewModelBase
 {
-    public ConnectionNameViewModel() : this(DogfoodingService.Instance) { }
-
-    internal ConnectionNameViewModel(IDogfoodingService dogfoodingService)
-    {
-        this.dogfoodingService = dogfoodingService ?? throw new ArgumentNullException(nameof(dogfoodingService));
-    }
-
     private ConnectionInfo connectionInfo;
-    private readonly IDogfoodingService dogfoodingService;
+    private IConnectedModeUIServices connectedModeUiServices;
+
+    public IConnectedModeUIServices ConnectedModeUiServices
+    {
+        get => connectedModeUiServices;
+        set
+        {
+            connectedModeUiServices = value;
+            RaisePropertyChanged();
+            RaisePropertyChanged(nameof(DisplayName));
+            RaisePropertyChanged(nameof(ShouldDisplayRegion));
+            RaisePropertyChanged(nameof(DisplayRegion));
+        }
+    }
 
     public ConnectionInfo ConnectionInfo
     {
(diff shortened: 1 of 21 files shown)


src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
```diff
@@ -53,6 +58,8 @@ public ConnectionInfo ConnectionInfo
             ? connectionInfo.CloudServerRegion.Url.ToString()
             : connectionInfo?.Id ?? string.Empty;
 
-    public bool ShouldDisplayRegion => dogfoodingService.IsDogfoodingEnvironment && connectionInfo?.ServerType is ConnectionServerType.SonarCloud;
+    public bool ShouldDisplayRegion =>
+        ConnectedModeUiServices?.DogfoodingService.IsDogfoodingEnvironment == true && ConnectedModeUiServices?.SonarLintSettings.ShowCloudRegion == true &&
```
georgii-borovinskikh-sonarsource, line 62: Not sure I understand the logic. Why is it `IsDogfood && ShowRegion` and not `IsDogfood || ShowRegion`? Also, why not just remove dogfooding check entirely? Also, remove `== true`

georgii-borovinskikh-sonarsource (changes requested):

src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
gabriela-trutan-sonarsource: Removing the dogfooding check is part of another [task](https://sonarsource.atlassian.net/browse/SLVS-1805). 
I am not sure this functionality is 100% ready for production, that's why we should check if it's in the dogfooding AND the checkbox is checked, which means that the option is only working for dogfooding and not for production.

gabriela-trutan-sonarsource: I have moved the check for dogfooding inside the Options page and make sure the checkbox is hidden.

src/Integration.Vsix/Settings/GeneralOptionsDialogControlViewModel.cs
```diff
@@ -66,6 +67,8 @@ public bool IsActivateMoreEnabled
         }
     }
 
+    public bool ShouldShowCloudRegionOption => dogfoodingService.IsDogfoodingEnvironment;
+
     public bool ShowCloudRegion
```
georgii-borovinskikh-sonarsource, line 72: Not sure if this is an issue, but if you set this to true and then set dogfooding to false, it will still return true

src/ConnectedMode/UI/ConnectionDisplay/ConnectionNameViewModel.cs
```diff
@@ -59,7 +59,7 @@ public ConnectionInfo ConnectionInfo
             : connectionInfo?.Id ?? string.Empty;
 
     public bool ShouldDisplayRegion =>
-        ConnectedModeUiServices?.DogfoodingService.IsDogfoodingEnvironment == true && ConnectedModeUiServices?.SonarLintSettings.ShowCloudRegion == true &&
+        ConnectedModeUiServices?.SonarLintSettings.ShowCloudRegion == true &&
```
georgii-borovinskikh-sonarsource, line 62:
```
ConnectedModeUiServices?.SonarLintSettings.ShowCloudRegion &&
```

georgii-borovinskikh-sonarsource (approved):

src/Integration.Vsix/Settings/GeneralOptionsDialogControlViewModel.cs
gabriela-trutan-sonarsource: That is correct. But I don't think this is a problem as we are not expecting the environment variable to change on the machines. 
Additionally, we expect to remove this check anyway once we release it to production, so I wouldn't spend effort in this.

PR #6123, opened 2025-03-26
Gb/binding refactoring draft

georgii-borovinskikh-sonarsource:
Part of


src/ConnectedMode.UnitTests/SlCoreConnectionAdapterTests.cs
```diff
@@ -75,7 +76,7 @@ public async Task ValidateConnectionAsync_SwitchesToBackgroundThread()
 
         await slCoreConnectionAdapter.ValidateConnectionAsync(sonarQubeConnectionInfo, new TokenCredentialsModel("myToken".CreateSecureString()));
 
-        await threadHandlingMock.Received(1).RunOnBackgroundThread(Arg.Any<Func<Task<AdapterResponse>>>());
+        await threadHandlingMock.Received(1).RunOnBackgroundThread(Arg.Any<Func<Task<ResponseStatus>>>());
```
georgii-borovinskikh-sonarsource, line 79: Just renames

src/ConnectedMode.UnitTests/UI/BindingResultTests.cs
```diff
@@ -0,0 +1,59 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.ConnectedMode.UI;
+using SonarLint.VisualStudio.ConnectedMode.UI.Resources;
+
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests.UI;
+
+[TestClass]
+public class BindingResultTests
```
georgii-borovinskikh-sonarsource, line 27: New result tests

src/ConnectedMode.UnitTests/UI/Credentials/EditCredentialsViewModelTests.cs
```diff
@@ -125,67 +126,35 @@ public async Task UpdateConnectionCredentials_TryUpdatesProvidedConnection(bool
         serverConnectionsRepositoryAdapter.Received(1).TryUpdateCredentials(sonarQubeConnection, Arg.Is<TokenCredentialsModel>(x => x.Token == testSubject.Token));
     }
 
-    [TestMethod]
```
georgii-borovinskikh-sonarsource, line 128: This functionality is delegated to the new class

src/ConnectedMode.UnitTests/UI/ManageConnections/ManageConnectionsViewModelTest.cs
```diff
@@ -535,7 +531,7 @@ private void MockTryGetConnections(List<Connection> connections) =>
 
     private void MockDeleteBinding(string localBindingKey, bool success) => connectedModeBindingServices.SolutionBindingRepository.DeleteBinding(localBindingKey).Returns(success);
 
-    private void MockUnbind(string localBindingKey, bool success) => connectedModeBindingServices.BindingController.Unbind(localBindingKey).Returns(success);
+    private void MockUnbind(string localBindingKey, bool success) => connectedModeBindingServices.BindingControllerAdapter.Unbind(localBindingKey).Returns(success);
```
georgii-borovinskikh-sonarsource, line 534: Uses adapter to unbind now

src/ConnectedMode.UnitTests/UI/OrganizationSelection/OrganizationSelectionViewModelTests.cs
```diff
@@ -43,7 +43,7 @@ public void TestInitialize()
         credentialsModel = Substitute.For<ICredentialsModel>();
         slCoreConnectionAdapter = Substitute.For<ISlCoreConnectionAdapter>();
         progressReporterViewModel = Substitute.For<IProgressReporterViewModel>();
-        progressReporterViewModel.ExecuteTaskWithProgressAsync(Arg.Any<ITaskToPerformParams<AdapterResponse>>()).Returns(new AdapterResponse(true));
+        progressReporterViewModel.ExecuteTaskWithProgressAsync(Arg.Any<ITaskToPerformParams<ResponseStatus>>()).Returns(new ResponseStatus(true));
```
georgii-borovinskikh-sonarsource, line 46: Just renames

src/ConnectedMode.UnitTests/UI/ProjectSelection/ProjectSelectionViewModelTests.cs
```diff
@@ -116,7 +116,7 @@ public async Task ProjectSearchTerm_ExecutesInitializationWithProgress()
 
         await progressReporterViewModel.Received(1)
             .ExecuteTaskWithProgressAsync(
-                Arg.Is<TaskToPerformParams<AdapterResponseWithData<List<ServerProject>>>>(x =>
+                Arg.Is<TaskToPerformParams<ResponseStatus<List<ServerProject>>>>(x =>
```
georgii-borovinskikh-sonarsource, line 119: Just renames

src/ConnectedMode/Resources.resx
```diff
@@ -190,35 +190,16 @@
   <data name="StatefulBranchProvider_ReturnValue" xml:space="preserve">
     <value>[ConnectedMode/BranchMapping] Closest Sonar server branch: {0}</value>
   </data>
-  <data name="Store_UpdateIssue_NoMatch" xml:space="preserve">
```
georgii-borovinskikh-sonarsource, line 193: Cleaned up old unused strings

src/ConnectedMode/Resources.resx
```diff
@@ -334,20 +312,20 @@
   <data name="Binding_Fails" xml:space="preserve">
     <value>Binding failed due to: {0}</value>
   </data>
-  <data name="AutomaticBinding_ConnectionNotFound" xml:space="preserve">
+  <data name="Binding_ConnectionNotFound" xml:space="preserve">
```
georgii-borovinskikh-sonarsource, line 315: Removed `Automatic` from the name

src/ConnectedMode/Resources.resx
```diff
@@ -374,13 +352,19 @@
...
-  <data name="AutomaticBindingType_Shared" xml:space="preserve">
+  <data name="BindingType_Manual" xml:space="preserve">
+    <value>Manual</value>
+  </data>
+  <data name="BindingType_Shared" xml:space="preserve">
     <value>Shared</value>
   </data>
-  <data name="AutomaticBindingType_Suggested" xml:space="preserve">
+  <data name="BindingType_SuggestedShared" xml:space="preserve">
+    <value>Suggested Shared</value>
```
georgii-borovinskikh-sonarsource, line 365: Renamed this special case

src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -39,52 +39,37 @@ namespace SonarLint.VisualStudio.ConnectedMode;
...
-    Task<AdapterResponseWithData<List<ServerProject>>> GetAllProjectsAsync(ServerConnection serverConnection);
+    Task<ResponseStatus<List<ServerProject>>> GetAllProjectsAsync(ServerConnection serverConnection);
 
-    Task<AdapterResponseWithData<List<ServerProject>>> FuzzySearchProjectsAsync(ServerConnection serverConnection, string searchTerm);
+    Task<ResponseStatus<List<ServerProject>>> FuzzySearchProjectsAsync(ServerConnection serverConnection, string searchTerm);
 
-    Task<AdapterResponseWithData<string>> GenerateTokenAsync(ConnectionInfo connectionInfo, CancellationToken cancellationToken);
-}
-
-public class AdapterResponseWithData<T>(bool success, T responseData) : IResponseStatus
```
georgii-borovinskikh-sonarsource, line 56: Moved to a separate file and renamed to `ResponseStatus`

src/ConnectedMode/SlCoreConnectionAdapter.cs
```diff
@@ -39,52 +39,37 @@ namespace SonarLint.VisualStudio.ConnectedMode;
...
+    Task<ResponseStatus<string>> GenerateTokenAsync(ConnectionInfo connectionInfo, CancellationToken cancellationToken);
 }
 
 [Export(typeof(ISlCoreConnectionAdapter))]
 [method: ImportingConstructor]
 public class SlCoreConnectionAdapter(ISLCoreServiceProvider serviceProvider, IThreadHandling threadHandling, ILogger logger) : ISlCoreConnectionAdapter
 {
-    private static readonly AdapterResponseWithData<List<OrganizationDisplay>> FailedResponseWithData = new(false, []);
-    private static readonly AdapterResponse FailedResponse = new(false);
+    private static readonly ResponseStatus<List<OrganizationDisplay>> FailedResponseWithData = new(false, []);
```
georgii-borovinskikh-sonarsource, line 60: Just renames below

src/ConnectedMode/UI/BindingControllerAdapter.cs
```diff
@@ -0,0 +1,115 @@
...
+internal class BindingControllerAdapter(
+    IBindingController bindingController,
+    ISolutionInfoProvider solutionInfoProvider,
+    IServerConnectionsRepositoryAdapter serverConnectionsRepositoryAdapter,
+    ILogger logger)
+    : IBindingControllerAdapter
+{
+    private readonly ILogger logger = logger.ForContext(ConnectedMode.Resources.ConnectedModeLogContext, ConnectedMode.Resources.ConnectedModeBindingLogContext);
+
+    public async Task<BindingResult> ValidateAndBindAsync(BindingRequest request, IConnectedModeUIManager uiManager, CancellationToken token)
```
georgii-borovinskikh-sonarsource, line 49: New adapter which has the functionality that was previously in ManageBindingViewModel

src/ConnectedMode/UI/BindingControllerAdapter.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    public bool Unbind(string bindingKey = null)
+    {
+        bindingKey ??= solutionInfoProvider.GetSolutionName();
+        return bindingController.Unbind(bindingKey);
+    }
+
+    private async Task<Either<ServerConnection, BindingResult.ValidationFailure>> ValidateRequestAsync(
```
georgii-borovinskikh-sonarsource, line 80: Used Either as an experiment, this can be changed

src/ConnectedMode/UI/BindingResult.cs
```diff
@@ -18,14 +18,33 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using SonarLint.VisualStudio.ConnectedMode.UI.Resources;
+
 namespace SonarLint.VisualStudio.ConnectedMode.UI;
 
-internal enum BindingResult
+internal class BindingResult
```
georgii-borovinskikh-sonarsource, line 25: extended this model to better reflect the intended usage

src/ConnectedMode/UI/BindingResult.cs
```diff
@@ -18,14 +18,33 @@
...
 {
-    Success,
-    Failed,
-    ConnectionNotFound,
-    SharedConfigurationNotAvailable,
```
georgii-borovinskikh-sonarsource, line 28: This is handled by the callers in the request

src/ConnectedMode/UI/Credentials/EditCredentialsViewModel.cs
```diff
@@ -46,37 +48,26 @@ internal async Task UpdateConnectionCredentialsWithProgressAsync()
...
-            var localBindingKey = await connectedModeBindingServices.SolutionInfoProvider.GetSolutionNameAsync();
-            var boundServerProject = new BoundServerProject(localBindingKey, serverProjectKey, serverConnection);
-            await connectedModeBindingServices.BindingController.BindAsync(boundServerProject, CancellationToken.None);
-            return new AdapterResponse(true);
-        }
-        catch (Exception)
-        {
-            return new AdapterResponse(false);
-        }
+        var validateAndBindAsync = await connectedModeBindingServices.BindingControllerAdapter.ValidateAndBindAsync(
```
georgii-borovinskikh-sonarsource, line 67: logic got moved into the new class

src/ConnectedMode/UI/IResponseStatus.cs
```diff
@@ -0,0 +1,46 @@
...
+
+
+public interface IResponseStatus
+{
+    bool Success { get; }
+
+    string WarningText { get; }
+}
+
+public class ResponseStatus<T>(bool success, T responseData, string warningText = null) : IResponseStatus
```
georgii-borovinskikh-sonarsource, line 31: this class is a renamed copy of AdapterResponse/AdapterResponseWithData

src/ConnectedMode/UI/IResponseStatus.cs
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+namespace SonarLint.VisualStudio.ConnectedMode.UI;
+
+
+public interface IResponseStatus
+{
+    bool Success { get; }
+
+    string WarningText { get; }
```
georgii-borovinskikh-sonarsource, line 28: Warning text got added to support overrides based on status

src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
```diff
@@ -25,7 +25,7 @@
 
 namespace SonarLint.VisualStudio.ConnectedMode.UI.ManageConnections
 {
-    public class ManageConnectionsViewModel(
+    internal class ManageConnectionsViewModel(
```
georgii-borovinskikh-sonarsource, line 28: renames and BindingController.Unbind -> BindingControllerAdapter.Unbind

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -248,41 +277,25 @@ internal void OnProgressUpdated()
         RaisePropertyChanged(nameof(IsExportButtonEnabled));
     }
 
-    internal async Task<AdapterResponse> LoadDataAsync()
-    {
-        var succeeded = false;
```
georgii-borovinskikh-sonarsource, line 253: Thread switching and try/catch were redundant

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -167,37 +166,67 @@ public async Task InitializeDataAsync()
...
-        {
-            AfterProgressUpdated = OnProgressUpdated
-        };
+        var bind = new TaskToPerformParams<ResponseStatus<BindingResult>>(() => PerformBindingInternalAsync(binding), UiResources.BindingInProgressText,
+            UiResources.BindingFailedText) { AfterProgressUpdated = OnProgressUpdated };
         await ProgressReporter.ExecuteTaskWithProgressAsync(bind);
     }
 
-    public async Task<BindingResult> PerformAutomaticBindingWithProgressAsync(AutomaticBindingRequest automaticBinding)
+    internal async Task<ResponseStatus<BindingResult>> PerformBindingInternalAsync(BindingRequest request)
```
georgii-borovinskikh-sonarsource, line 185: Moved logic to the new adapter

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -167,37 +166,67 @@ public async Task InitializeDataAsync()
...
     {
-        var bind = new TaskToPerformParams<AdapterResponseWithData<BindingResult>>(() => PerformAutomaticBindingInternalAsync(automaticBinding), UiResources.BindingInProgressText,
-            UiResources.BindingFailedText) { AfterProgressUpdated = OnProgressUpdated };
-        var result = await ProgressReporter.ExecuteTaskWithProgressAsync(bind);
-        return result.ResponseData;
+        var bindingResult = await connectedModeBindingServices.BindingControllerAdapter.ValidateAndBindAsync(
+            request,
+            connectedModeUiManager,
+            cancellationTokenSource.Token);
+        ReloadConnectionData(); // this is to ensure that the newly added connection is added to the view model properties
```
georgii-borovinskikh-sonarsource, line 191: Refresh connections list in case adapter added a new connection

src/ConnectedMode/UI/OrganizationSelection/OrganizationSelectionViewModel.cs
```diff
@@ -61,16 +61,16 @@ public void AddOrganization(OrganizationDisplay organization)
 
     public async Task LoadOrganizationsAsync()
     {
-        var organizationLoadingParams = new TaskToPerformParams<AdapterResponseWithData<List<OrganizationDisplay>>>(
+        var organizationLoadingParams = new TaskToPerformParams<ResponseStatus<List<OrganizationDisplay>>>(
```
georgii-borovinskikh-sonarsource, line 64: Just renames

src/ConnectedMode/UI/ProgressReporterViewModel.cs
```diff
@@ -33,11 +33,6 @@ public interface IProgressReporterViewModel
     Task<T> ExecuteTaskWithProgressAsync<T>(ITaskToPerformParams<T> parameters, bool clearPreviousState = true) where T : IResponseStatus;
 }
 
-public interface IResponseStatus
```
georgii-borovinskikh-sonarsource, line 36: Moved

src/ConnectedMode/UI/ProgressReporterViewModel.cs
```diff
@@ -122,7 +117,7 @@ private void UpdateProgress<T>(string status, ITaskToPerformParams<T> parameters
 
     private void OnFailure<T>(ITaskToPerformParams<T> parameters, T response, bool clearPreviousState) where T : IResponseStatus
     {
-        UpdateWarning(parameters.WarningText, clearPreviousState);
+        UpdateWarning(response?.WarningText ?? parameters.WarningText, clearPreviousState);
```
georgii-borovinskikh-sonarsource, line 120: warning text overrides

src/ConnectedMode/UI/ProjectSelection/ProjectSelectionViewModel.cs
```diff
@@ -72,35 +72,35 @@ public ServerProject SelectedProject
 
     public async Task InitializeProjectWithProgressAsync()
     {
-        var initializeProjectsParams = new TaskToPerformParams<AdapterResponseWithData<List<ServerProject>>>(
+        var initializeProjectsParams = new TaskToPerformParams<ResponseStatus<List<ServerProject>>>(
```
georgii-borovinskikh-sonarsource, line 75: Just renames

src/ConnectedMode/UI/Resources/UiResources.resx
```diff
@@ -120,9 +120,6 @@
   <data name="AuthenticationDescription" xml:space="preserve">
     <value>Please fill the authentication details for</value>
   </data>
-  <data name="AuthenticationTypeLabel" xml:space="preserve">
```
georgii-borovinskikh-sonarsource, line 123: Cleaned up unused strings

src/ConnectedMode/UI/Resources/UiResources.resx
```diff
@@ -330,6 +321,15 @@
   <data name="FetchingBindingStatusFailedText" xml:space="preserve">
     <value>Binding could not be established. Please check the SonarQube for Visual Studio logs for more details.</value>
   </data>
+  <data name="FetchingBindingStatusFailedTextProjectNotFound" xml:space="preserve">
```
georgii-borovinskikh-sonarsource, line 324: Warning text now changes based on the failure type

src/ConnectedMode/UI/TrustConnection/TrustConnectionViewModel.cs
```diff
@@ -47,15 +47,15 @@ public class TrustConnectionViewModel(
 
     internal async Task CreateConnectionsWithProgressAsync()
     {
-        var validationParams = new TaskToPerformParams<AdapterResponse>(CreateNewConnectionAsync,
+        var validationParams = new TaskToPerformParams<ResponseStatus>(CreateNewConnectionAsync,
```
georgii-borovinskikh-sonarsource, line 50: Just renames

src/ConnectedMode/UI/BindingControllerAdapter.cs
```diff
@@ -0,0 +1,115 @@
...
+            logger.WriteLine(logContext, ConnectedMode.Resources.Binding_CredentiasNotFound, connection.Id);
+            return BindingResult.ValidationFailure.CredentialsNotFound;
+        }
+
+        return connection;
+    }
+
+    private ServerConnection GetExistingServerConnectionOrNull(string connectionId) => serverConnectionsRepositoryAdapter.TryGet(connectionId, out var connection) ? connection : null;
+
+    private async Task<ServerConnection> GetNewServerConnectionOrNullAsync(BindingRequest request, IConnectedModeUIManager uiManager) =>
```
gabriela-trutan-sonarsource, line 110: I don't think it's a good idea to hide showing the trust connection dialog inside the validation method. This is a side effect that is not obvious from the method name. Also the method is doing more than one responsibility of validating right now (it's showing a connection and creating it)
This should not be hidden.

src/ConnectedMode/UI/BindingControllerAdapter.cs
```diff
@@ -0,0 +1,115 @@
...
+            return BindingResult.Success;
+        }
+        catch (Exception ex) when (!ErrorHandler.IsCriticalException(ex))
+        {
+            logger.WriteLine(logContext, ConnectedMode.Resources.Binding_Fails, ex.Message);
+            return BindingResult.Failed;
+        }
+    }
+
+    public bool Unbind(string bindingKey = null)
```
gabriela-trutan-sonarsource, line 74: I don't see why the `bindingKey` is nullable. For this case, we have to get/calculate the solution name again, but the view model already has that information can can simply pass it.
Please change it to not-nullable

src/ConnectedMode.UnitTests/UI/BindingControllerAdapterTests.cs
```diff
@@ -0,0 +1,265 @@
...
+        var returnsList = returns
+            .Select<ServerConnection, Func<CallInfo, bool>>(x =>
+                info =>
+                {
+                    info[1] = x;
+                    return x != null;
+                })
+            .ToArray();
+        serverConnectionsAdapter.TryGet(request.ConnectionId, out Arg.Any<ServerConnection>())
+            .Returns(returnsList.First(), returnsList.Skip(1).ToArray());
```
gabriela-trutan-sonarsource, line 246: This method is a untrustworthy, because it won't work if the list of `returns` is bigger than 2
Please refactor it or change the parameter type

gabriela-trutan-sonarsource (changes requested):
I have some feedback regarding some decisions.

src/ConnectedMode.UnitTests/UI/BindingResultTests.cs
```diff
@@ -44,7 +44,6 @@ public void BindingResult_HasExpectedProperties(object result, string warningTex
     public static object[][] ResultValues =>
     [
         [BindingResult.ValidationFailure.ConnectionNotFound, BindingResult.ConnectionNotFound],
-        [BindingResult.ValidationFailure.SharedConfigurationNotAvailable, BindingResult.SharedConfigurationNotAvailable],
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: It's not possible to end up in this state - the callers check that shared binging exists before requesting binding with that shared binging provided in the request

src/ConnectedMode/UI/BindingRequest.cs
```diff
@@ -18,23 +18,42 @@
...
+    }
 
-    private AutomaticBindingRequest() { }
+    public abstract record AutomaticBindingRequest : BindingRequest;
 
     /// <summary>
     /// Indicates binding parameters derived from shared binding
     /// </summary>
-    public record Shared : AutomaticBindingRequest
+    public record Shared(SharedBindingConfigModel Model) : AutomaticBindingRequest
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Any binding is manual at some point. Automatic here refers to the binding data being selected by code, not via UI

src/ConnectedMode/UI/ManageBinding/ManageBindingViewModel.cs
```diff
@@ -167,37 +166,67 @@ public async Task InitializeDataAsync()
...
+        if (bindingResult == BindingResult.Success)
+        {
+            bindingResult = (await DisplayBindStatusAsync()).ResponseData;
+        }
+        UpdateBindingTelemetry(request, bindingResult);
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: the check would still be needed, because DisplayBindStatusAsync can still fail

src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
```diff
@@ -926,392 +892,172 @@ public void ConnectionSelectionCaptionText_NoConnectionExists_ReturnsNoConnectio
...
-
-        var response = await testSubject.PerformAutomaticBindingInternalAsync(automaticBindingRequest);
+        await testSubject.PerformManualBindingWithProgressAsync();
+        var taskToPerformParams = (TaskToPerformParams<ResponseStatus<BindingResult>>)progressReporterViewModel.ReceivedCalls().Single().GetArguments()[0];
+        await taskToPerformParams.TaskToPerform.Invoke();
 
-        await VerifyBindingNotPerformed(response, BindingResult.CredentialsNotFound, automaticBindingRequest.TypeName, Resources.AutomaticBinding_CredentiasNotFound, expectedServerConnection.Id);
-        messageBox.Received(1).Show(UiResources.NotFoundCredentialsForAutomaticBindingMessageBoxText, UiResources.NotFoundCredentialsForAutomaticBindingMessageBoxCaption, MessageBoxButton.OK,
-            MessageBoxImage.Warning);
+        connectedModeBindingServices.BindingControllerAdapter.Received(1).ValidateAndBindAsync(Arg.Is<Manual>(x => x.ProjectKey == ServerProject.Key && x.ConnectionId == connection.Id), connectedModeUIManager, Arg.Any<CancellationToken>());
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: It's a false positive. Methods after `Received` don't need awaiting.

src/ConnectedMode.UnitTests/UI/BindingControllerAdapterTests.cs
georgii-borovinskikh-sonarsource: You mean less than 2?

src/ConnectedMode/UI/BindingRequest.cs
```diff
@@ -51,8 +51,9 @@ public abstract record AutomaticBindingRequest : BindingRequest;
     public record Shared(SharedBindingConfigModel Model) : AutomaticBindingRequest
     {
         internal override string TypeName => ConnectedMode.Resources.BindingType_Shared;
-        internal override string ProjectKey => Model?.ProjectKey;
-        internal override string ConnectionId => Model?.CreateConnectionInfo().GetServerIdFromConnectionInfo();
+        internal override string ProjectKey => Model.ProjectKey;
+        internal override string ConnectionId => Model.CreateConnectionInfo().GetServerIdFromConnectionInfo();
+        public SharedBindingConfigModel Model { get; init; } = Model ?? throw new ArgumentNullException(nameof(Model));
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I think this is an impossible situation - callers check for this model to be present before attempting binding. I also want to minimize the amount of `if(request is Type && ...)` -like checks and keep the validation more generic

src/ConnectedMode/UI/InteractiveConnectionForBindingProvider.cs
```diff
@@ -29,33 +29,16 @@ public async Task<ServerConnection> GetServerConnectionAsync(BindingRequest requ
...
             && await connectedModeUiManager.ShowTrustConnectionDialogAsync(shared.Model.CreateConnectionInfo().GetServerConnectionFromConnectionInfo(), token: null) is true
             && serverConnectionsRepositoryAdapter.TryGet(shared.ConnectionId, out var newConnection))
         {
-            return await CheckCredentialsAsync(newConnection);
+            return newConnection;
         }
 
         return null;
     }
-
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Wdym?

src/ConnectedMode.UnitTests/UI/BindingResultTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    [DataTestMethod]
+    public void BindingResult_HasExpectedProperties(object result, string warningText, bool isSuccessful)
+    {
+        ((BindingResult)result).ProblemDescription.Should().Be(warningText);
+        ((BindingResult)result).IsSuccessful.Should().Be(isSuccessful);
+    }
+
+    public static object[][] ResultValues =>
+    [
+        [BindingResult.ValidationFailure.ConnectionNotFound, BindingResult.ConnectionNotFound],
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: That would partially defeat the purpose of the test, which is to show that these properties refer to the same thing. It would be confusing to use different type names to the ones that are actually tested in this file.

We should not apply coding style rules if they don't fit a particular situation.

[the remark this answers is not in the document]
gabriela-trutan-sonarsource: It does apply to this situation.
We test that a property is referencing another one. And the qualification should be speaking enough to understand the difference.

src/ConnectedMode/UI/BindingResult.cs
```diff
@@ -18,14 +18,33 @@
...
+    public static ValidationFailure CredentialsNotFound => ValidationFailure.CredentialsNotFoundFailure;
+    public static ValidationFailure ProjectKeyNotFound => ValidationFailure.ProjectKeyNotFoundFailure;
+
+    private BindingResult(string problemDescription, bool isSuccessful)
+    {
+        ProblemDescription = problemDescription;
+        IsSuccessful = isSuccessful;
+    }
+
+    internal class ValidationFailure : BindingResult
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I can try and see how it looks

src/ConnectedMode.UnitTests/UI/BindingResultTests.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: It is possible, if the method is called via another path. So the validation is (and should be agnostic) by which is its callers.

src/ConnectedMode/UI/BindingRequest.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: > I think this is an impossible situation - callers check for this model to be present before attempting binding. 

If it is impossible, than why do we throw an exception? The answer to that is that you have no control who creates an instance of this object. So it is possible.

Again, I think this makes the validation inconsistent and spreads the logic into multiple classes.

src/ConnectedMode/UI/InteractiveConnectionForBindingProvider.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: This TODO was for me. :D I will remove it

src/ConnectedMode.UnitTests/UI/BindingResultTests.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: It's a bad idea to use aliases for the component under test. Tests should use type names exactly how they're defined in the original file.

src/ConnectedMode/UI/BindingRequest.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: > Any binding is manual at some point.

Yes, but the current state is that the shared binding can be manual right now (not in a point in the future). So the inheritance makes the code confusing

src/ConnectedMode/UI/BindingRequest.cs
```diff
@@ -45,6 +64,9 @@ public record Assisted(
         string ServerProjectKey,
         bool IsFromSharedBinding) : AutomaticBindingRequest
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: > don't think it's worth it to create a separate class just for  that
I think the effort is not big.
Also we are duplicating the model from the DTO, which is not very clear for the UI purposes.

src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: I know, but there are a lot of warnings in this class which are very distracting.

src/ConnectedMode/UI/BindingRequest.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: it's not manual. The binding parameters are **not** coming from the UI. This specifically allows the UiManager to accept shared and automatic binding, but not manual. That's the distinction.

src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
```diff
@@ -926,392 +892,172 @@ public void ConnectionSelectionCaptionText_NoConnectionExists_ReturnsNoConnectio
...
-        await VerifyBindingNotPerformed(response, BindingResult.CredentialsNotFound, automaticBindingRequest.TypeName, Resources.AutomaticBinding_CredentiasNotFound,
-            automaticBindingRequest.ServerConnectionId);
-        messageBox.Received(1).Show(UiResources.NotFoundCredentialsForAutomaticBindingMessageBoxText, UiResources.NotFoundCredentialsForAutomaticBindingMessageBoxCaption, MessageBoxButton.OK,
-            MessageBoxImage.Warning);
-    }
+         response.Success.Should().BeFalse();
+         response.ResponseData.Should().BeSameAs(BindingResult.Failed);
+         connectedModeBindingServices.BindingControllerAdapter.Received(1).ValidateAndBindAsync(bindingRequest, connectedModeUIManager, Arg.Any<CancellationToken>());
+         VerifyConnectionsRefreshed();
+         VerifyBindingTelemetryNotSent();
```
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: I think the behavior before was wrong. If the binding succeeded, but we are not able to display the binding status for whatever reason, why shouldn't the telemetry reflect that the use added a binding?

src/ConnectedMode/UI/BindingRequest.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: It would require me to change the class + tests, change the conversion from DTO + tests, and change the switch in telemetry + tests. It definitely is not worth it.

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: The duplication of DTO is needed because we should not depend on DTOs in product code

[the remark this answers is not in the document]
gabriela-trutan-sonarsource: > The duplication of DTO is needed because we should not depend on DTOs in product code

Of course. I am not arguing that. I am just arguing the the UI model should be improved to be more speaking than having a flag, which makes it confusing, e.g. why is there a `Shared` model and an `Assisted` model that has a  `IsFromSharedBinding` flag.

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: There's a docstring that explains the difference: one is coming from us, the other one is from SLCore

src/ConnectedMode.UnitTests/UI/BindingControllerAdapterTests.cs
gabriela-trutan-sonarsource: No, I mean bigger than 2. If you provide 3 parameters, then in the line 246, the  `serverConnectionsAdapter` will only mock the first 2.

src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I don't think writing boilerplate code to avoid a false positive from the analyzer is a good idea. Could you look into disabling this rule for test projects?

src/ConnectedMode.UnitTests/UI/BindingControllerAdapterTests.cs
georgii-borovinskikh-sonarsource: It uses `Skip(1)`, not `Take(1)`, so it should get all of the remaining return values after the first one.

src/ConnectedMode/UI/BindingRequest.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: > This specifically allows the UiManager to accept shared and automatic binding, but not manual. That's the distinction.

I have seen that. And I understood the motivation behind adding the inheritance. It was a nice solution. 
I am just saying that the inheritance looks incomplete. Maybe the `Shared` could also inherit from the `Manual` so then it would really reflect the current state? It's just an idea.
If you don't have any other better idea how we could improve it, I think it's also fine as it is.

src/ConnectedMode/UI/BindingControllerAdapter.cs
georgii-borovinskikh-sonarsource: The idea was to remove the need to provide the current solution name by the callers, similar how to the other method in this class is called. I can see that this is confusing. I will split it into `UnbindCurrentSolution()` and `Unbind(string bindingKey)` methods

src/ConnectedMode.UnitTests/UI/BindingResultTests.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: > Tests should use type names exactly how they're defined in the original file.

It does. It's just an alias. We already do that with enums. I don't see how this situation is any different.

src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Could you look into that, please? Or create a task for it.

src/ConnectedMode/UI/BindingRequest.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Well, shared can't inherit from manual and automatic at the same time. The best solution would be to not have inheritance and just use `Shared | Assisted` as the type for the method argument in UiManager, but we can't have nice things in C#. The alternative is to have 2 interface methods in UiManager, one for Shared and one for Assisted. It would then be possible to remove the Automatic base class at the cost of having an extra method of the interface. I can do that. WDYT?

src/ConnectedMode.UnitTests/UI/BindingResultTests.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Tests should use type names exactly how they're defined in the original file. That is the difference

src/ConnectedMode.UnitTests/UI/BindingControllerAdapterTests.cs
gabriela-trutan-sonarsource: Yes, I see. And I see now why you are doing it like this: there is no available overload for the `Returns` that simply accepts an array of parameters.

src/ConnectedMode.UnitTests/UI/BindingResultTests.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: We never took such a decision.

src/ConnectedMode/UI/BindingRequest.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: We would need another overload for showing the ManageBindingDialog without any parameters.
There could be another way: creating marker interfaces for the records and abstract records.
So from my point: either make `AutomaticBindingRequest` an interface (and add a `Manual` one) or leave the code as it is now.

src/ConnectedMode/UI/BindingControllerAdapter.cs
gabriela-trutan-sonarsource: >The idea was to remove the need to provide the current solution name by the callers

There is no additional effort for the callers in the current state to provide the `bindingKey`. On the contrary, you make the adapter call unnecessarily the solution name. If that need ever arises, we could change the code in that point.

src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
```diff
@@ -926,392 +892,172 @@ public void ConnectionSelectionCaptionText_NoConnectionExists_ReturnsNoConnectio
...
+    public async Task PerformBindingInternalAsync_Manual_Succeeds()
     {
-        testSubject.SelectedProject = ServerProject; // this is to make sure the SelectedProject is ignored and the shared config is used instead
-        testSubject.SharedBindingConfigModel = sonarCloudSharedBindingConfigModel;
-        testSubject.SharedBindingConfigModel.ProjectKey = null;
-        SetupBoundProject(new ServerConnection.SonarCloud(testSubject.SharedBindingConfigModel.Organization));
+        var bindingRequest = new Manual("any", "any");
 
-        var response = await testSubject.PerformAutomaticBindingInternalAsync(new AutomaticBindingRequest.Shared());
+        await VerifySuccessfulBinding(bindingRequest);
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: The 1st method would need to return the response and pass it to the 2nd method. This will be repeated 3 times. Renamed the method instead to `TestSuccessfulBinding`

src/ConnectedMode/UI/BindingControllerAdapter.cs
georgii-borovinskikh-sonarsource: The call to GetSolutionName is done to keep the behavior consistent with the Bind implementation, which does not accept solution name

src/ConnectedMode/UI/BindingRequest.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I will leave it like it is, we can change to interfaces later if inheritance starts limiting us at some point

src/ConnectedMode/UI/BindingResult.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Looks a bit clunky to use because of the need to create an extra object + unsure how to handle the problem description

gabriela-trutan-sonarsource (approved):
Some very minor thing, otherwise the changes look OK and functionality seems to work as expected.

src/ConnectedMode/UI/BindingControllerAdapter.cs
gabriela-trutan-sonarsource: I don't understand why should we be consistent with the ValidateAndBindAsync method, as there are different use cases: binding can indeed only occur for the current solution, but unbinding can be done for multiple solutions (i.e. deleting a connection case)

src/ConnectedMode.UnitTests/UI/ManageBinding/ManageBindingViewModelTests.cs
```diff
@@ -976,54 +977,50 @@ public async Task PerformBindingInternalAsync_Assisted_Succeeds(bool isFromShare
         }
     }
 
-    private async Task VerifySuccessfulBinding(BindingRequest bindingRequest)
+    private async Task TestSuccessfulBinding(BindingRequest bindingRequest)
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I treat this more like the actual test body rather than a helper method, which is why I placed it next to the invocations (less scrolling to find out what those tests actually do)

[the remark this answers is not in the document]
gabriela-trutan-sonarsource: It is a helper method and it brings a lot of confusion to find some private method between other test methods. I think it would be a lot cleaner if it would be at the bottom, respecting not only our coding style regarding tests, but also regarding the private methods that should be after the public ones.

PR #6127, opened 2025-04-02
SLVS-1420 Connected mode: Prevent duplicate connection

vnaskos-sonar:
[SLVS-1420](https://sonarsource.atlassian.net/browse/SLVS-1420)

[SLVS-1420]: https://sonarsource.atlassian.net/browse/SLVS-1420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
+++ b/src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
@@ -103,12 +103,6 @@ internal class ManageConnectionsViewModel(
 
     internal async Task CreateConnectionsWithProgressAsync(Connection connection, ICredentialsModel credentialsModel)
     {
-        if (connectedModeServices.ServerConnectionsRepositoryAdapter.TryGet(connection.Info, out var serverConnection) && serverConnection != null)
-        {
-            ProgressReporterViewModel.Warning = UiResources.ConnectionAlreadyExistsText;
-            return;
-        }
-
         var validationParams = new TaskToPerformParams<ResponseStatus>(
             () => CreateNewConnectionAsync(connection, credentialsModel),
             UiResources.CreatingConnectionProgressText,
@@ -176,6 +170,11 @@ internal class ManageConnectionsViewModel(
 
     internal Task<ResponseStatus> CreateNewConnectionAsync(Connection connection, ICredentialsModel credentialsModel)
     {
+        if (connectedModeServices.ServerConnectionsRepositoryAdapter.TryGet(connection.Info, out var serverConnection) && serverConnection != null)
+        {
+            return Task.FromResult(new ResponseStatus(false, UiResources.ConnectionAlreadyExistsText));
+        }
+
         var succeeded = connectedModeServices.ServerConnectionsRepositoryAdapter.TryAddConnection(connection, credentialsModel);
         if (succeeded)
         {
--- a/src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
+++ b/src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
@@ -103,6 +103,12 @@ internal class ManageConnectionsViewModel(
 
     internal async Task CreateConnectionsWithProgressAsync(Connection connection, ICredentialsModel credentialsModel)
     {
+        if (connectedModeServices.ServerConnectionsRepositoryAdapter.TryGet(connection.Info, out var serverConnection) && serverConnection != null)
+        {
+            ProgressReporterViewModel.Warning = UiResources.ConnectionAlreadyExistsText;
+            return;
+        }
+
         var validationParams = new TaskToPerformParams<ResponseStatus>(
             () => CreateNewConnectionAsync(connection, credentialsModel),
             UiResources.CreatingConnectionProgressText,
--- a/src/ConnectedMode/UI/OrganizationSelection/ManualOrganizationSelectionViewModel.cs
+++ b/src/ConnectedMode/UI/OrganizationSelection/ManualOrganizationSelectionViewModel.cs
@@ -31,7 +31,7 @@ public class ManualOrganizationSelectionViewModel : ViewModelBase
         get => organizationKey;
         set
         {
-            organizationKey = value;
+            organizationKey = value?.Replace(" ", string.Empty);
             RaisePropertyChanged();
             RaisePropertyChanged(nameof(IsValidOrganizationKey));
         }
--- a/src/ConnectedMode/UI/OrganizationSelection/ManualOrganizationSelectionViewModel.cs
+++ b/src/ConnectedMode/UI/OrganizationSelection/ManualOrganizationSelectionViewModel.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using SonarLint.VisualStudio.Core.Helpers;
 using SonarLint.VisualStudio.Core.WPF;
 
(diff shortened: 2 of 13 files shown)


src/ConnectedMode/UI/ManageConnections/ManageConnectionsViewModel.cs
```diff
@@ -92,6 +92,12 @@ internal ResponseStatusWithData<List<string>> GetConnectionReferences(Connection
 
         internal async Task CreateConnectionsWithProgressAsync(Connection connection, ICredentialsModel credentialsModel)
         {
+            if (connectedModeServices.ServerConnectionsRepositoryAdapter.TryGet(connection.Info, out var serverConnection) && serverConnection != null)
```
gabriela-trutan-sonarsource, line 95: Please move the check inside the `CreateNewConnection` method. If the `TryGet` operation takes a lot of time (file is too big, another process is locking the file etc), the UI would be frozen instead of showing a progress bar. In general we should do all the operations inside the progress reporter operation.

I understand why you added the if outside, because you wanted the warning to be more specific. But the `ProgressReporterViewModel` now supports overriding the warning test, if you set it in the response. 

What needs to be done is to: 

- Remove `SafeExecuteActionAsync`
- Make `CreateNewConnection` return a `ResponseStatus`
- When the connection is duplicated, set the `WarningText` of the `ResponseStatus` for a more specific warning

gabriela-trutan-sonarsource (changes requested):
I can create a duplicate connection through manual organization if I add a whitespace at the beginning: 

image

src/ConnectedMode/UI/OrganizationSelection/ManualOrganizationSelectionViewModel.cs
```diff
@@ -31,7 +31,7 @@ public string OrganizationKey
         get => organizationKey;
         set
         {
-            organizationKey = value;
+            organizationKey = value?.Replace(" ", string.Empty);
```
gabriela-trutan-sonarsource, line 34: I don't think it's a good idea to remove the whitespaces from inside the words. I know that spaces are not allowed in the organization key, but it's usually an unexpected operation to remove white spaces from the middle of the words (e.g. "my special organization") if the user did not do so.

Anyway, if such spaces exist, our code won't crash, but the functionality will not work due to the fact the organization won't exist. So the user will receive an appropriate message for that. 

We should use `Trim` rather than `Replace`

gabriela-trutan-sonarsource, line 34: I also think it's a little bit fragile to do the correction in the UI and, more specifically, just in the `ManualOrganizationSelectionViewModel.cs`. 

I think it would be a lot safer if the trimming should happen in the `ServerConnectionsRepository.TryAdd`, which will make the functionality more maintainable (if in the future we add another information that will part of the id, which is set through UI, we will have to change yet another place, but if it's in the repository we won't)

gabriela-trutan-sonarsource (changes requested):

src/ConnectedMode/UI/OrganizationSelection/ManualOrganizationSelectionViewModel.cs
vnaskos-sonar: I went with the `Replace` because the organization key and the server uri cannot contain spaces.

vnaskos-sonar: Yes, makes sense to add the formatting on a lower level. Although I think the `ServerConnection` constructors should take care of this.

For the UI, it's more clear to show that the organization key cannot contain spaces and prevent them from adding them.

src/Core/Binding/ServerConnection.cs
```diff
@@ -48,19 +53,8 @@ public SonarCloud(
...
-            IConnectionCredentials credentials = null)
-            : base(OrganizationKeyToId(region, organizationKey), settings, credentials)
-        {
-            OrganizationKey = organizationKey;
-            Region = region;
-        }
-
-        public string OrganizationKey { get; }
-        public CloudServerRegion Region { get; }
+        public string OrganizationKey { get; } = organizationKey?.Replace(" ", "");
```
gabriela-trutan-sonarsource, line 56:
```suggestion
        public string OrganizationKey { get; } = organizationKey?.Replace(" ", string.Empty);
```

src/Core/Binding/ServerConnection.cs
```diff
@@ -78,6 +72,8 @@ private static string OrganizationKeyToId(CloudServerRegion region, string organ
     public sealed class SonarQube(Uri serverUri, ServerConnectionSettings settings = null, IConnectionCredentials credentials = null)
         : ServerConnection(serverUri?.ToString(), settings, credentials)
     {
-        public override Uri ServerUri { get; } = serverUri;
+        public override Uri ServerUri { get; } = serverUri == null
+            ? null
+            : new Uri(serverUri.ToString().Replace(" ", ""));
```
gabriela-trutan-sonarsource, line 77:
```suggestion
            : new Uri(serverUri.ToString().Replace(" ", string.Empty));
```

src/Core/Binding/ServerConnection.cs
gabriela-trutan-sonarsource, line 56: The replacing of empty strings is the same in this class and in the [view model](https://github.com/SonarSource/sonarlint-visualstudio/blob/a3d2717e3b3e74dc598ed867b388b211f365ba8a/src/ConnectedMode/UI/OrganizationSelection/ManualOrganizationSelectionViewModel.cs#L34). 

Please extract the duplicate code into one method that is reused everywhere (consider introducing a string extension method). This will improve our maintainability efforts (e.g. if we ever decide to use regex and remove additional characters, we will only do it in one place)

gabriela-trutan-sonarsource (approved):
Functionality seems to be working. Should we consider removing the white spaces also from the textbox for [SonarQubeUrl](https://github.com/SonarSource/sonarlint-visualstudio/blob/4a1701b5088b0a8f9ba48f477294b806309e8cb1/src/ConnectedMode/UI/ServerSelection/ServerSelectionViewModel.cs#L58) and from the manual `[ProjectKey](https://github.com/SonarSource/sonarlint-visualstudio/blob/4a1701b5088b0a8f9ba48f477294b806309e8cb1/src/ConnectedMode/UI/ProjectSelection/ManualProjectSelectionViewModel.cs#L29)`, so that the user have a consistent experience in the UI?

Also please refactor the duplicate logic of replacing the empty spaces (see comments).

src/Core/Binding/ServerConnection.cs
vnaskos-sonar: > (e.g. if we ever decide to use regex and remove additional characters, we will only do it in one place)

There are two options for an extension method.
1. `RemoveSpaces`, that can potentially be used in other places.
2. `RemoveSpacesFromServerUriAndOrganizationKey` 

The first option won't help in the case you mentioned above, because other places of the code may use this extension.
The second option wouldn't be applicable to all strings, so a string extension might not be the best solution.

For the time being I'll leave it as is.

gabriela-trutan-sonarsource: Why would it be dangerous option 1?

vnaskos-sonar: It depends on what changes will happen to the extension method, if they are only specific to this ticket and the extension method is already used by other places in the code.

PR #6129, opened 2025-04-02
SLVS-1997 Parse settings.json comma separated sonarlint.analysisExcludesStandalone

vnaskos-sonar:
[SLVS-1997](https://sonarsource.atlassian.net/browse/SLVS-1997)

[SLVS-1997]: https://sonarsource.atlassian.net/browse/SLVS-1997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core.UnitTests/UserRuleSettings/AnalysisSettingsTests.cs
+++ b/src/Core.UnitTests/UserRuleSettings/AnalysisSettingsTests.cs
@@ -7,6 +7,64 @@ namespace SonarLint.VisualStudio.Core.UnitTests.UserRuleSettings;
 public class AnalysisSettingsTests
 {
     [TestMethod]
+    public void AnalysisSettings_SerializesCorrectly()
+    {
+        var settings = new AnalysisSettings
+        {
+            Rules =
+            {
+                { "typescript:S2685", new RuleConfig { Level = RuleLevel.On, Parameters = new Dictionary<string, string> { { "key1", "value1" } }, } }
+            },
+            FileExclusions = ["file1.cpp", "**/obj/*", "file2.cpp"]
+        };
+        const string expectedJson =
+            """
+            {
+              "sonarlint.rules": {
+                "typescript:S2685": {
+                  "level": "On",
+                  "parameters": {
+                    "key1": "value1"
+                  }
+                }
+              },
+              "sonarlint.analysisExcludesStandalone": "file1.cpp,**/obj/*,file2.cpp"
+            }
+            """;
+
+        var json = JsonConvert.SerializeObject(settings, Formatting.Indented);
+
+        json.Should().Be(expectedJson);
+    }
+
+    [TestMethod]
+    public void AnalysisSettings_DeserializesCorrectly()
+    {
+        const string json = """
+                            {
+                              "sonarlint.rules": {
+                                "typescript:S2685": {
+                                  "level": "On",
+                                  "parameters": {
+                                    "key1": "value1"
+                                  }
+                                }
+                              },
+                              "sonarlint.analysisExcludesStandalone": "file1.cpp,**/obj/*,file2.cpp"
+                            }
+                            """;
+
+        var settings = JsonConvert.DeserializeObject<AnalysisSettings>(json);
+
+        settings.Rules.Should().BeEquivalentTo(
+            new Dictionary<string, RuleConfig>
+            {
+                { "typescript:S2685", new RuleConfig { Level = RuleLevel.On, Parameters = new Dictionary<string, string> { { "key1", "value1" } }, } }
+            });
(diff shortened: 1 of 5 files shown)


src/Core/Helpers/CommaSeparatedStringArrayConverter.cs
```diff
@@ -0,0 +1,19 @@
...
+    public override void WriteJson(JsonWriter writer, string[] value, JsonSerializer serializer) => writer.WriteValue(string.Join(",", value));
+
+    public override string[] ReadJson(
+        JsonReader reader,
+        Type objectType,
+        string[] existingValue,
+        bool hasExistingValue,
+        JsonSerializer serializer)
+    {
+        var value = reader.Value?.ToString();
```
georgii-borovinskikh-sonarsource, line 16: probably should ignore/throw on all non-strings here

src/Core/Helpers/CommaSeparatedStringArrayConverter.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Newtonsoft.Json;
+
+namespace SonarLint.VisualStudio.Core.Helpers;
+
+public class CommaSeparatedStringArrayConverter : JsonConverter<string[]>
```
georgii-borovinskikh-sonarsource, line 5: We had some problems with the generic base class on older VS versions (they use an outdated newtonsoft json version which does not have it). Run this code on the vs 17.0 to test if it still breaks. If it does, you will need to change it to the non-generic version

src/Core/UserRuleSettings/AnalysisSettings.cs
```diff
@@ -18,70 +18,67 @@
...
-            "level": "on",
-            "Parameters": {
-              "key1": "value1",
-              "key2": "value2"
-            },
-            "severity": "Critical"
-        },
-    }
-...
+    [JsonProperty("sonarlint.analysisExcludesStandalone", NullValueHandling = NullValueHandling.Ignore)]
```
georgii-borovinskikh-sonarsource, line 60: null value ignore, but the default value was not changed from [] to null

src/Core.UnitTests/UserRuleSettings/AnalysisSettingsTests.cs
```diff
@@ -0,0 +1,40 @@
+﻿using Newtonsoft.Json;
+using SonarLint.VisualStudio.Core.UserRuleSettings;
+
+namespace SonarLint.VisualStudio.Core.UnitTests.UserRuleSettings;
+
+[TestClass]
+public class AnalysisSettingsTests
```
georgii-borovinskikh-sonarsource, line 7: Why not add these tests to `AnalysisSettingsSerializerTests`? Would be nice to have a combined test as well

georgii-borovinskikh-sonarsource (changes requested):

src/Core/Helpers/CommaSeparatedStringArrayConverter.cs
vnaskos-sonar: Good to know, yup it's still an issue

src/Core.UnitTests/UserRuleSettings/AnalysisSettingsTests.cs
vnaskos-sonar: The `AnalysisSettingsSerializerTests` is testing a lot of things at the same time, this is why I created the `AnalysisSettingsTests` to test only for the serialization. I didn't want to refactor that test in this PR, the keep the scope small.

src/Core/Helpers/CommaSeparatedStringArrayConverter.cs
vnaskos-sonar: If it's not a string, it will return null (ignore)

src/Core.UnitTests/UserRuleSettings/AnalysisSettingsTests.cs
```diff
@@ -23,6 +23,22 @@ public void AnalysisSettings_FileExclusions_SerializesCorrectly()
         json.Should().Be(expectedJson);
     }
 
+    [TestMethod]
```
georgii-borovinskikh-sonarsource, line 26: still missing tests where both fields are not empty

src/Core/Helpers/CommaSeparatedStringArrayConverter.cs
georgii-borovinskikh-sonarsource: 🤔 not what I would expect

georgii-borovinskikh-sonarsource (approved):

PR #6135, opened 2025-04-04
SLVS-1966 Update Options page

gabriela-trutan-sonarsource:
Update Options page to reflect that the settings is no longer only for rules, but also for file exclusions

### DIFF
--- a/src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml
+++ b/src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml
@@ -52,11 +52,10 @@
             </TextBlock>
         </StackPanel>
 
-        <StackPanel Grid.Column="0" x:Name="VerbosityPanel" Grid.Row="2" Grid.ColumnSpan="2">
-            <TextBlock Text="Logging level:"
-                       Margin="5" />
+        <StackPanel Grid.Column="0" x:Name="VerbosityPanel" Grid.Row="2" Grid.ColumnSpan="2" Margin="5">
+            <TextBlock Text="Logging level:" />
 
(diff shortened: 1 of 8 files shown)


src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml
```diff
@@ -31,55 +31,55 @@
             <ColumnDefinition Width="*" />
         </Grid.ColumnDefinitions>
 
-        <Button Grid.Column="0" Grid.Row="0" Content="Edit rules settings"
+        <Button Grid.Column="0" Grid.Row="0" Content="Edit Settings"
```
georgii-borovinskikh-sonarsource, line 34: I know that we don't have one yet, but it would be great to include the documentation link here as well. Could you create a task for that?

georgii-borovinskikh-sonarsource (approved):

src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml
gabriela-trutan-sonarsource: I am not sure which documentation link are you talking about. We do have [this link](https://docs.sonarsource.com/sonarqube-for-ide/visual-studio/using/rules/#disabling-a-rule) in the [wiki](https://github.com/SonarSource/sonarlint-visualstudio/blob/11194e807aae56bb16e2d1448349d57b457db3b2/src/Integration.Vsix/Settings/GeneralOptionsDialogControl.xaml#L50).

Is it not good enough?

georgii-borovinskikh-sonarsource: I think this link should be good, once we add the exclusions to the docs: https://docs.sonarsource.com/sonarqube-for-ide/visual-studio/using/rules/#settingsjson-file-format-and-location

PR #6139, opened 2025-04-08
SLVS-1968 Update SonarLint.targets file to be generated at startup and to include standalone configs.

gabriela-trutan-sonarsource:
[SLVS-1968](https://sonarsource.atlassian.net/browse/SLVS-1968)

The `SonarLint.targets` file should now be generated on startup if missing, as opposed to being currently generated on first connected mode project if missing

[SLVS-1968]: https://sonarsource.atlassian.net/browse/SLVS-1968?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Integration/CSharpVB/Install/ImportBeforeFileGenerator.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+    }
+
+    private string GetTargetFileContent()
+    {
+        using var stream = GetType().Assembly.GetManifestResourceStream(ResourcePath);
+        if (stream == null)
+        {
+            return "";
```
georgii-borovinskikh-sonarsource, line 91: Should probably not overwrite the file with empty string

src/Integration/CSharpVB/Install/ImportBeforeFileGenerator.cs
```diff
@@ -0,0 +1,106 @@
...
+[method: ImportingConstructor]
+internal class ImportBeforeFileGenerator(ILogger logger, IFileSystemService fileSystem, IThreadHandling threadHandling) : IImportBeforeFileGenerator
+{
+    private readonly ILogger logger = logger.ForContext(Strings.ImportsBeforeFileGeneratorLogContext);
+    private const string TargetsFileName = "SonarLint.targets";
+    private const string ResourcePath = "SonarLint.VisualStudio.Integration.CSharpVB.Install.SonarLintTargets.xml";
+
+    private static readonly object Locker = new();
+
+    public Task WriteTargetsFileToDiskIfNotExistsAsync() => threadHandling.RunOnBackgroundThread(WriteTargetsFileToDiskIfNotExists);
```
georgii-borovinskikh-sonarsource, line 49: the name is to long + doesn't mention that we update the content of the file. move the specifics to a doc string and make the name shorter

georgii-borovinskikh-sonarsource (approved):

src/Integration/CSharpVB/Install/ImportBeforeFileGenerator.cs
gabriela-trutan-sonarsource: It was doing that [before](https://github.com/SonarSource/sonarlint-visualstudio/blob/8b401fe935bbf17bd406c85b00d56936f1671755/src/ConnectedMode/Install/ImportBeforeFileGenerator.cs#L108). I did not change the logic

src/Integration/CSharpVB/Install/ImportBeforeFileGenerator.cs
gabriela-trutan-sonarsource: I don't think a long name for a method is problematic. It's better to be descriptive

src/Integration/CSharpVB/Install/ImportBeforeFileGenerator.cs
georgii-borovinskikh-sonarsource: Should change that while we're at it

src/Integration/CSharpVB/Install/ImportBeforeFileGenerator.cs
gabriela-trutan-sonarsource: `UpdatesTargetsFileAsync` ?

PR #6157, opened 2025-04-17
SLVS-2027 Provide UI for file exclusions

gabriela-trutan-sonarsource:
[SLVS-2027](https://sonarsource.atlassian.net/browse/SLVS-2027)

As part of this PR a new page has been added in the Options dialog. Filling the page with the exclusions defined in the `settings.json` and updating the file on clicking OK will be part of another PR.

[SLVS-2027]: https://sonarsource.atlassian.net/browse/SLVS-2027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Integration.Vsix/Settings/FileExclusions/FileExclusionsDialogPage.cs
+++ b/src/Integration.Vsix/Settings/FileExclusions/FileExclusionsDialogPage.cs
@@ -21,6 +21,7 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Windows;
 using Microsoft.VisualStudio.Shell;
+using SonarLint.VisualStudio.Core;
 
 namespace SonarLint.VisualStudio.Integration.Vsix.Settings.FileExclusions;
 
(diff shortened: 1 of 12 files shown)


src/Integration.Vsix/Settings/FileExclusions/FileExclusionsDialogPage.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Windows;
+using Microsoft.VisualStudio.Shell;
+using SonarLint.VisualStudio.Core;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.Settings.FileExclusions;
+
+[ExcludeFromCodeCoverage]
+internal class FileExclusionsDialogPage : UIElementDialogPage
+{
+    public const string PageName = "File Exclusions";
```
vnaskos-sonar, line 31: `"File Exclusions"` can be a Resource

vnaskos-sonar:
This is not the case for the other IDEs at the moment. However, I think that it would be nice to warn the user that the these File Exclusions won't be applied on Connected Mode.

Another option would be to display the Connected Mode exclusions when you are in Connected Mode, so that the user won't have to check on the SQ Cloud/Server to view the file exclusions that are applied. (This is just an idea, it definitely needs a PM/UXers involvement)

georgii-borovinskikh-sonarsource (approved):

PR #6158, opened 2025-04-17
SLVS-2068 Show exclusions stored in settings.json into the UI and save them on disk on OK

gabriela-trutan-sonarsource:
[SLVS-2068](https://sonarsource.atlassian.net/browse/SLVS-2068)

This targets the [PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/6157) in which the UI for file exclusions was added.

[SLVS-2068]: https://sonarsource.atlassian.net/browse/SLVS-2068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/UserRuleSettings/UserSettingsProvider.cs
+++ b/src/Core/UserRuleSettings/UserSettingsProvider.cs
@@ -44,12 +44,16 @@ internal sealed class UserSettingsProvider : IUserSettingsProvider, IDisposable
     [ImportingConstructor]
     public UserSettingsProvider(ILogger logger, ISingleFileMonitorFactory singleFileMonitorFactory) : this(logger, singleFileMonitorFactory, new FileSystem(), UserSettingsFilePath) { }
 
-    internal /* for testing */ UserSettingsProvider(ILogger logger, ISingleFileMonitorFactory singleFileMonitorFactory, IFileSystem fileSystem, string settingsFilePath)
+    internal /* for testing */ UserSettingsProvider(
+        ILogger logger,
+        ISingleFileMonitorFactory singleFileMonitorFactory,
+        IFileSystem fileSystem,
+        string settingsFilePath)
     {
         this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
         this.fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
         var fileMonitorFactory = singleFileMonitorFactory ?? throw new ArgumentNullException(nameof(singleFileMonitorFactory));
-        this.serializer = new AnalysisSettingsSerializer(fileSystem, logger);
+        serializer = new AnalysisSettingsSerializer(fileSystem, logger);
 
         SettingsFilePath = settingsFilePath;
         settingsFileMonitor = fileMonitorFactory.Create(SettingsFilePath);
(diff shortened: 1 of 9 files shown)


src/Core/UserRuleSettings/UserSettingsProvider.cs
```diff
@@ -90,6 +94,13 @@ public void DisableRule(string ruleId)
         serializer.SafeSave(SettingsFilePath, UserSettings.AnalysisSettings);
     }
 
+    public void UpdateFileExclusions(IEnumerable<string> exclusions)
+    {
+        UserSettings.AnalysisSettings.UserDefinedFileExclusions.Clear();
```
georgii-borovinskikh-sonarsource, line 99: There's already a similar problem in `DisableRule`, but I don't think we should be modifying the collections here. The code is not thread safe. While the chance of things going sideways is low, we probably should try to fix it.

src/Integration.Vsix/Settings/FileExclusions/FileExclusionsViewModel.cs
```diff
@@ -20,13 +20,23 @@
...
 {
     private ExclusionViewModel selectedExclusion;
+    private readonly IBrowserService browserService;
+    private readonly IUserSettingsProvider userSettingsProvider;
+
+    public FileExclusionsViewModel(IBrowserService browserService, IUserSettingsProvider userSettingsProvider)
+    {
+        this.browserService = browserService;
+        this.userSettingsProvider = userSettingsProvider;
+        InitializeExclusions();
```
georgii-borovinskikh-sonarsource, line 38: this is called twice

src/Integration.Vsix/Settings/FileExclusions/FileExclusionsViewModel.cs
```diff
@@ -60,4 +70,18 @@ internal void RemoveExclusion()
...
+    {
+        Exclusions.Clear();
+        var exclusionViewModels = userSettingsProvider.UserSettings.AnalysisSettings.UserDefinedFileExclusions.Select(ex => new ExclusionViewModel(ex));
+        exclusionViewModels.ToList().ForEach(vm => Exclusions.Add(vm));
+        SelectedExclusion = Exclusions.FirstOrDefault();
+    }
+
+    internal void SaveExclusions()
+    {
+        var exclusionsToSave = Exclusions.Where(vm => vm.Error == null).Select(vm => vm.Pattern);
```
georgii-borovinskikh-sonarsource, line 84: should we instead fail the save if there are errors, possibly?

georgii-borovinskikh-sonarsource (approved):

src/Integration.Vsix/Settings/FileExclusions/FileExclusionsViewModel.cs
gabriela-trutan-sonarsource: It is validation errors from the frontend, so I don't think we should fail the save (if you modified a valid exclusion and then you added by mistake a new, empty one, I don't think the user experience would be nice if we discard everything).
 
Additionally, I am currently under discussion with the PM in how to proceed when there are validation errors and the users clicks "OK" button.

src/Core/UserRuleSettings/UserSettingsProvider.cs
gabriela-trutan-sonarsource: Created a different task for it: https://sonarsource.atlassian.net/browse/SLVS-2070

PR #6161, opened 2025-04-24
SLVS-2070 Make UserSettingsProvider thread safe

gabriela-trutan-sonarsource:


src/Core/UserRuleSettings/AnalysisSettings.cs
```diff
@@ -57,14 +58,15 @@ public class AnalysisSettings
...
-    [JsonProperty("sonarlint.rules", ObjectCreationHandling = ObjectCreationHandling.Reuse)]
-    public Dictionary<string, RuleConfig> Rules { get; set; } = new(StringComparer.OrdinalIgnoreCase);
+    [JsonProperty("sonarlint.rules")]
+    [JsonConverter(typeof(ImmutableDictionaryIgnoreCaseConverter<string, RuleConfig>))]
+    public ImmutableDictionary<string, RuleConfig> Rules { get; init; }
 
     [JsonProperty("sonarlint.analysisExcludesStandalone")]
     [JsonConverter(typeof(CommaSeparatedStringArrayConverter))]
-    public List<string> UserDefinedFileExclusions
+    public ImmutableList<string> UserDefinedFileExclusions
```
georgii-borovinskikh-sonarsource, line 69: Could be ImmutableArray

src/Core/UserRuleSettings/AnalysisSettings.cs
```diff
@@ -57,14 +58,15 @@ public class AnalysisSettings
...
     private const string AnyDirectoryWildcard = "**";
     private static readonly string AnyRootPrefix = AnyDirectoryWildcard + Path.AltDirectorySeparatorChar;
+    private readonly ImmutableList<string> userDefinedFileExclusions = ImmutableList<string>.Empty;
 
-    private readonly List<string> userDefinedFileExclusions = [];
-    [JsonProperty("sonarlint.rules", ObjectCreationHandling = ObjectCreationHandling.Reuse)]
-    public Dictionary<string, RuleConfig> Rules { get; set; } = new(StringComparer.OrdinalIgnoreCase);
+    [JsonProperty("sonarlint.rules")]
+    [JsonConverter(typeof(ImmutableDictionaryIgnoreCaseConverter<string, RuleConfig>))]
+    public ImmutableDictionary<string, RuleConfig> Rules { get; init; }
```
georgii-borovinskikh-sonarsource, line 65: Using immutables is better, so good job for doing that, but technically we could have gotten away with IReadOnlyBlaBla interfaces and copying, which would be less code changes

georgii-borovinskikh-sonarsource (approved):

src/Core/UserRuleSettings/AnalysisSettings.cs
gabriela-trutan-sonarsource: I am not sure what would be the benefit.

georgii-borovinskikh-sonarsource: Well, we don't modify this list, so it's better to use the type that better reflects that (+ allocates less memory)

georgii-borovinskikh-sonarsource:
This should be IReadOnlyDictionary/ImmutableDictionary too https://github.com/SonarSource/sonarlint-visualstudio/pull/6161/files#diff-b3eac8c92502604bf253f8d6fc6b55432bededde8350f003cf90554f8281534bR124

PR #6163, opened 2025-04-24
SLVS-2071 Add popup for adding and editing a new file exclusion

gabriela-trutan-sonarsource:
[SLVS-2071](https://sonarsource.atlassian.net/browse/SLVS-2071)

[SLVS-2071]: https://sonarsource.atlassian.net/browse/SLVS-2071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Integration.Vsix/Settings/FileExclusions/EditExclusionDialog.cs
+++ b/src/Integration.Vsix/Settings/FileExclusions/EditExclusionDialog.cs
@@ -1,0 +1,33 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.Integration.Vsix.Resources;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.Settings.FileExclusions;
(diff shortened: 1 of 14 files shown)


src/Integration.Vsix/Settings/FileExclusions/ExclusionViewModel.cs
```diff
@@ -18,15 +18,15 @@
...
-using System.ComponentModel;
 using SonarLint.VisualStudio.Core.WPF;
 using SonarLint.VisualStudio.Integration.Vsix.Resources;
 
 namespace SonarLint.VisualStudio.Integration.Vsix.Settings.FileExclusions;
 
-internal class ExclusionViewModel : ViewModelBase, IDataErrorInfo
+internal class ExclusionViewModel : ViewModelBase
 {
     private const char InvalidCharacter = ',';
```
georgii-borovinskikh-sonarsource, line 28: I know this is not the point of this PR, but as I realize now we could add more characters (in a different PR), icluding https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getinvalidfilenamechars?view=netframework-4.7.2 and whitespace

src/Integration.Vsix/Settings/FileExclusions/EditExclusionDialog.cs
```diff
@@ -0,0 +1,35 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Diagnostics.CodeAnalysis;
+using SonarLint.VisualStudio.Integration.Vsix.Resources;
+
+namespace SonarLint.VisualStudio.Integration.Vsix.Settings.FileExclusions;
+
+[ExcludeFromCodeCoverage]
+internal class EditExclusionDialog : AddExclusionDialog
```
georgii-borovinskikh-sonarsource, line 27: I think a hierarchy of ManageExclusionDialog -> AddExclusionDialog, EditExclusionDialog could be cleaner

georgii-borovinskikh-sonarsource (approved):

src/Integration.Vsix/Settings/FileExclusions/ExclusionViewModel.cs
gabriela-trutan-sonarsource: Why whitespaces? As mentioned, they can appear in file/folder names.
I would not create another task, because the other IDE flavours do not do any other validation at all (besides completely empty input). 
I added the validation for comma, because otherwise we would have to escape it to be able to save it in the json file (so less effort on our side), but I would not go any further with validation. 
From the link you posted, I expect the method to return that ? is an invalid character, but for us it's a valid character in our pattern.

PR #6164, opened 2025-04-24
SLVS-2073 Fix threading in importing constructor of ActiveSolutionTracker

georgii-borovinskikh-sonarsource:
[SLVS-2073](https://sonarsource.atlassian.net/browse/SLVS-2073)

[SLVS-2073]: https://sonarsource.atlassian.net/browse/SLVS-2073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Integration.UnitTests/MefServices/ActiveSolutionTrackerTests.cs
+++ b/src/Integration.UnitTests/MefServices/ActiveSolutionTrackerTests.cs
@@ -113,23 +113,21 @@ public class ActiveSolutionTrackerTests
     }
 
     [TestMethod]
-    public void ActiveSolutionTracker_DisposeBeforeInitilized_DisposeAndInitializeDoNothing()
+    public void ActiveSolutionTracker_DisposeBeforeInitialized_DisposeAndInitializeDoNothing()
     {
         var testSubject = CreateUninitializedTestSubject(out var barrier);
         solutionInfoProvider.ClearReceivedCalls();
         var eventHandler = Substitute.For<EventHandler<ActiveSolutionChangedEventArgs>>();
         testSubject.ActiveSolutionChanged += eventHandler;
         testSubject.Dispose();
+        barrier.SetResult(1);
 
         solutionMock.SimulateSolutionClose();
         solutionMock.SimulateSolutionOpen();
 
         eventHandler.DidNotReceiveWithAnyArgs().Invoke(default, default);
         solutionInfoProvider.DidNotReceiveWithAnyArgs().GetSolutionName();
-
-        barrier.SetResult(1);
-        var initialize = () => testSubject.InitializeAsync();
-        initialize.Should().Throw<ObjectDisposedException>();
+        solutionMock.AssertNoHierarchyEventSinks();
     }
 
     [TestMethod]
--- a/src/Integration.UnitTests/MefServices/ActiveSolutionTrackerTests.cs
+++ b/src/Integration.UnitTests/MefServices/ActiveSolutionTrackerTests.cs
@@ -19,7 +19,9 @@
  */
 
 using Microsoft.VisualStudio.Shell.Interop;
+using NSubstitute.Extensions;
 using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Initialization;
 using SonarLint.VisualStudio.TestInfrastructure;
 
 namespace SonarLint.VisualStudio.Integration.UnitTests.MefServices;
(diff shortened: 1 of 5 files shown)


src/Integration/MefServices/ActiveSolutionTracker.cs
```diff
@@ -44,15 +48,40 @@ internal sealed class ActiveSolutionTracker : IActiveSolutionTracker, IVsSolutio
...
 
+        public Task InitializeAsync() =>
+            initializationHelper.InitializeAsync(
+                nameof(ActiveSolutionTracker),
+                [],
+                async threadHandling =>
+                {
+                    if (isDisposed)
+                    {
+                        throw new ObjectDisposedException(nameof(ActiveSolutionTracker));
```
gabriela-trutan-sonarsource, line 71: Why throwing an exception and simply not exiting the initialization method? What is the gain?

src/Integration/MefServices/ActiveSolutionTracker.cs
```diff
@@ -44,15 +48,40 @@ internal sealed class ActiveSolutionTracker : IActiveSolutionTracker, IVsSolutio
...
-            this.solution = serviceProvider.GetService<SVsSolution, IVsSolution>();
-            Debug.Assert(this.solution != null, "Cannot find IVsSolution");
-            ErrorHandler.ThrowOnFailure(this.solution.AdviseSolutionEvents(this, out this.cookie));
+            this.initializationHelper = initializationHelper;
+
+            InitializeAsync().Forget();
         }
 
+        public Task InitializeAsync() =>
+            initializationHelper.InitializeAsync(
```
gabriela-trutan-sonarsource, line 64: In this case, if there are no dependencies to load, what is the gain of using the `initializationHelper`? As the creation policy for this component is shared, the constructor will anyway only be executed once, independent on how many services are going to reference it.

It would be less overhead to just run the initialization on the UI thread directly than switching to background thread, then the UI thread again as it currently happens right now.

src/Integration.UnitTests/MefServices/ActiveSolutionTrackerTests.cs
```diff
@@ -30,87 +32,165 @@ public class ActiveSolutionTrackerTests
...
+        barrier.SetResult(1);
+        solutionInfoProvider.Received(1).GetSolutionName();
+
+        solutionMock.SimulateSolutionOpen();
+        eventHandler.ReceivedWithAnyArgs(1).Invoke(default, default);
+        solutionInfoProvider.Received(2).GetSolutionName();
+    }
+
+    [TestMethod]
+    public void ActiveSolutionTracker_DisposeBeforeInitilized_DisposeAndInitializeDoNothing()
```
gabriela-trutan-sonarsource, line 104:
```suggestion
    public void ActiveSolutionTracker_DisposeBeforeInitialized_DisposeAndInitializeDoNothing()
```

src/Integration/MefServices/ActiveSolutionTracker.cs
```diff
@@ -44,15 +48,40 @@ internal sealed class ActiveSolutionTracker : IActiveSolutionTracker, IVsSolutio
...
+                nameof(ActiveSolutionTracker),
+                [],
+                async threadHandling =>
+                {
+                    if (isDisposed)
```
gabriela-trutan-sonarsource, line 69: Wouldn't there be a benefit to check if the object is `IDisposable` and do the check inside the helper? We will then don't have to do it in all classes that implement `IRequireInitialization`. 
We would have to send the owner to the helper instead just the name, which would then create a circular reference, but I think it might be worth it.

gabriela-trutan-sonarsource (changes requested):
I think the AST can skip using the `IInitializationHelper`.

src/Integration/MefServices/ActiveSolutionTracker.cs
georgii-borovinskikh-sonarsource: IDisposable does not provide an indicator of whether the interface instance has been disposed already

src/Integration/MefServices/ActiveSolutionTracker.cs
georgii-borovinskikh-sonarsource: The dependencies need to wait for it to be initalized and we can't await the task in the constructor because of thread switching

gabriela-trutan-sonarsource (approved):

PR #6169, opened 2025-04-25
SLVS-2076 File exclusion improvements

gabriela-trutan-sonarsource:
[SLVS-2076](https://sonarsource.atlassian.net/browse/SLVS-2076)

- The "AddExclusionDialog" changes background color on dark mode. The Options page does not change its color, so we should keep it consistent
- Fix exception when normalizing paths and path contains invalid characters

[SLVS-2076]: https://sonarsource.atlassian.net/browse/SLVS-2076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/UserRuleSettings/AnalysisSettings.cs
+++ b/src/Core/UserRuleSettings/AnalysisSettings.cs
@@ -95,11 +95,15 @@ public class AnalysisSettings
 
     private static string NormalizePath(string path)
     {
-        // rooted paths without drive letter are unmodified, but SLCore doesn't match them well
-        var isRooted = HasInvalidPathChars(path) || Path.IsPathRooted(path);
-
         path = path.Replace(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+        // if path is invalid, the Path.IsPathRooted check will throw an exception
+        if (HasInvalidPathChars(path))
+        {
+            return path;
+        }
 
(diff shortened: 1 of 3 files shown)


src/Core/UserRuleSettings/AnalysisSettings.cs
```diff
@@ -96,7 +96,7 @@ public AnalysisSettings()
     private static string NormalizePath(string path)
     {
         // rooted paths without drive letter are unmodified, but SLCore doesn't match them well
-        var isRooted = Path.IsPathRooted(path);
+        var isRooted = HasInvalidPathChars(path) || Path.IsPathRooted(path);
```
georgii-borovinskikh-sonarsource, line 99: I don't really understand this change. Why should that influence it?

gabriela-trutan-sonarsource: Because `Path.IsPathRooted` throws an exception when the provided parameter has an invalid path character

georgii-borovinskikh-sonarsource: `var isRooted = HasInvalidPathChars(path)` is not correct then. You need to split these cases and handle them separately

georgii-borovinskikh-sonarsource (changes requested):

PR #6172, opened 2025-04-28
SLVS-2083 Introduce InitializationProcessorFactory

georgii-borovinskikh-sonarsource:
[SLVS-2083](https://sonarsource.atlassian.net/browse/SLVS-2083)

Part of SLVS-2025

[SLVS-2083]: https://sonarsource.atlassian.net/browse/SLVS-2083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/Initialization/IInitializationProcessor.cs
+++ b/src/Core/Initialization/IInitializationProcessor.cs
@@ -34,7 +34,7 @@ public interface IInitializationProcessor
     bool IsFinalized { get; }
 
     /// <summary>
-    /// Initializes once using given dependencies and owner's initializer
+    /// Initializes once on a background thread using given dependencies and owner's initializer
     /// </summary>
     Task InitializeAsync();
 }
--- a/src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorFactoryTests.cs
+++ b/src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorFactoryTests.cs
@@ -53,8 +53,10 @@ public class InitializationProcessorFactoryTests
     public void MefCtor_CheckIsSingleton() =>
         MefTestHelpers.CheckIsSingletonMefComponent<InitializationProcessorFactory>();
 
+     /// <summary>
+     /// For more tests see <see cref="InitializationProcessorTests"/>
+     /// </summary>
     [TestMethod]
     public void Create_ReturnsNonNull() =>
-        // for more tests see InitializationProcessorTests
         testSubject.Create<InitializationProcessorFactoryTests>([], _ => Task.CompletedTask).Should().NotBeNull();
 }
--- a/src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorFactoryTests.cs
+++ b/src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorFactoryTests.cs
@@ -1,0 +1,60 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Initialization;
+using SonarLint.VisualStudio.Core.Synchronization;
+using SonarLint.VisualStudio.Infrastructure.VS.Initialization;
+
+namespace SonarLint.VisualStudio.Infrastructure.VS.UnitTests.Initialization;
+
+[TestClass]
+public class InitializationProcessorFactoryTests
+{
+    private IAsyncLockFactory asyncLockFactory;
+    private IThreadHandling threadHandling;
+    private ILogger logger;
+    private InitializationProcessorFactory testSubject;
+
+    [TestInitialize]
+    public void TestInitialize()
+    {
+        asyncLockFactory = Substitute.For<IAsyncLockFactory>();
+        threadHandling = Substitute.For<IThreadHandling>();
+        logger = Substitute.For<ILogger>();
+        testSubject = new InitializationProcessorFactory(asyncLockFactory, threadHandling, logger);
+    }
+
+    [TestMethod]
+    public void MefCtor_CheckIsExported() =>
+        MefTestHelpers.CheckTypeCanBeImported<InitializationProcessorFactory, IInitializationProcessorFactory>(
+            MefTestHelpers.CreateExport<IAsyncLockFactory>(),
+            MefTestHelpers.CreateExport<IThreadHandling>(),
+            MefTestHelpers.CreateExport<ILogger>());
+
+    [TestMethod]
+    public void MefCtor_CheckIsSingleton() =>
+        MefTestHelpers.CheckIsSingletonMefComponent<InitializationProcessorFactory>();
+
+    [TestMethod]
+    public void Create_ReturnsNonNull() =>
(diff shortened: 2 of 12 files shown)


src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorFactoryTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        MefTestHelpers.CheckTypeCanBeImported<InitializationProcessorFactory, IInitializationProcessorFactory>(
+            MefTestHelpers.CreateExport<IAsyncLockFactory>(),
+            MefTestHelpers.CreateExport<IThreadHandling>(),
+            MefTestHelpers.CreateExport<ILogger>());
+
+    [TestMethod]
+    public void MefCtor_CheckIsSingleton() =>
+        MefTestHelpers.CheckIsSingletonMefComponent<InitializationProcessorFactory>();
+
+    [TestMethod]
```
gabriela-trutan-sonarsource, line 56:
```suggestion
     /// <summary>
     /// For more tests see <see cref="InitializationProcessorTests"/>
     /// </summary>
    [TestMethod]
```

src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorTests.cs
```diff
@@ -42,31 +42,22 @@ public void TestInitialize()
...
-
     [TestMethod]
     public void IsFinalized_NotStarted_ReturnsFalse() =>
-        testSubject.IsFinalized.Should().BeFalse();
+        testSubjectFactory.Create<InitializationProcessorTests>([], _ => Task.CompletedTask).IsFinalized.Should().BeFalse();
 
     [TestMethod]
     public async Task IsFinalized_StartedButNotFinished_ReturnsFalse()
     {
         var barrier = new TaskCompletionSource<byte>();
```
gabriela-trutan-sonarsource, line 55: Both the initialization substitute and the barrier could be extracted into fields to avoid bloated tests preparation

src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorTests.cs
```diff
@@ -143,31 +137,33 @@ public void InitializeAsync_DependencyThrows_ThrowsAndExecutesOnlyOnce()
...
+        var act = () => testSubject.InitializeAsync();
         act.Should().ThrowAsync<InvalidOperationException>();
         act.Should().ThrowAsync<InvalidOperationException>();
         act.Should().ThrowAsync<InvalidOperationException>();
 
         testSubject.IsFinalized.Should().BeTrue();
         dependency.Received(1).InitializeAsync();
         initialization.DidNotReceiveWithAnyArgs().Invoke(default);
-        testLogger.OutputStrings.Last().Should().ContainAll("Owner 1", "My Failed Dependency");
+        testLogger.OutputStrings.Last().Should().ContainAll("InitializationProcessorTests", "My Failed Dependency");
```
gabriela-trutan-sonarsource, line 150:
```suggestion
        testLogger.OutputStrings.Last().Should().ContainAll(nameof(InitializationProcessorTests), "My Failed Dependency");
```

src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorTests.cs
```diff
@@ -143,31 +137,33 @@ public void InitializeAsync_DependencyThrows_ThrowsAndExecutesOnlyOnce()
...
-        var act = () => testSubject.InitializeAsync("Owner 1", [], initialization);
+        var act = () => testSubject.InitializeAsync();
         act.Should().ThrowAsync<InvalidOperationException>();
         act.Should().ThrowAsync<InvalidOperationException>();
         act.Should().ThrowAsync<InvalidOperationException>();
 
         testSubject.IsFinalized.Should().BeTrue();
         initialization.ReceivedWithAnyArgs(1).Invoke(default);
-        testLogger.OutputStrings.Last().Should().ContainAll("Owner 1", "My Failed Operation");
+        testLogger.OutputStrings.Last().Should().ContainAll("InitializationProcessorTests", "My Failed Operation");
```
gabriela-trutan-sonarsource, line 167:
```suggestion
        testLogger.OutputStrings.Last().Should().ContainAll(nameof(InitializationProcessorTests), "My Failed Operation");
```

src/Infrastructure.VS/Initialization/InitializationProcessor.cs
```diff
@@ -39,15 +38,12 @@ public class InitializationProcessor(
...
     public bool IsFinalized => state.InitializationState.IsInitialized;
 
-    public Task InitializeAsync(
-        string owner,
-        IReadOnlyCollection<IRequireInitialization> dependencies,
-        Func<IThreadHandling, Task> initialization) =>
+    public Task InitializeAsync() =>
         !CheckInitialized()
-            ? threadHandling.RunOnBackgroundThread(() => InitializeInternalAsync(owner, dependencies, initialization))
+            ? threadHandling.RunOnBackgroundThread(() => InitializeInternalAsync())
```
gabriela-trutan-sonarsource, line 43:
```suggestion
            ? threadHandling.RunOnBackgroundThread(InitializeInternalAsync)
```

src/Core/Initialization/IInitializationProcessor.cs
```diff
@@ -20,19 +20,21 @@
 
 namespace SonarLint.VisualStudio.Core.Initialization;
 
+public interface IInitializationProcessorFactory
+{
+    IInitializationProcessor Create<T>(IReadOnlyCollection<IRequireInitialization> dependencies, Func<IThreadHandling, Task> initialization);
```
gabriela-trutan-sonarsource, line 25: Enforcing the delegate to receive `IThreadHandling` parameter leads to some additional boiler plate code in the implementation of the `IRequireInitialization`: we need to have two initialize methods one with the parameter and one without. If we simply remove it will simplify the code.

```suggestion
    IInitializationProcessor Create<T>(IReadOnlyCollection<IRequireInitialization> dependencies, Func<Task> initialization);
```

gabriela-trutan-sonarsource (changes requested):
The refactoring looks good, but I have another proposal for the `IInitializationProcessorFactory` that will simplify code.

src/Core/Initialization/IInitializationProcessor.cs
```diff
@@ -20,19 +20,21 @@
...
     /// <summary>
-    /// Returns true if initialization was completed or failed, false if it's not been finished
-    /// Does not await for initialization to finish
+    /// Returns true if initialization was completed or failed, false if it's not been finished.
+    /// Does not await for initialization to finish.
     /// </summary>
     bool IsFinalized { get; }
 
     /// <summary>
     /// Initializes once using given dependencies and owner's initializer
```
gabriela-trutan-sonarsource, line 37: I think we should mention that it runs the initialization on a background thread as it is not clear

src/Core/Initialization/IInitializationProcessor.cs
georgii-borovinskikh-sonarsource: The first option would not really work because the implementation of the lambda should never be directly accessible to other classes, meaning we'll still have 2 methods `public InitializeAsync(); + private InitializeImplementationAsync()`

I have considered the 2nd option and I decided to not change it because it felt like a big change. We can discuss that and I could make a separate PR with that.

src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorTests.cs
georgii-borovinskikh-sonarsource: Created a field for the func substitute. Barrier is only used once here so it's not worth moving it to initialization

src/Core/Initialization/IInitializationProcessor.cs
gabriela-trutan-sonarsource: They were not two different options, but different proposals that can be done at the same time.
We don't need the `public InitializeAsync();` if we also do the second proposal.

> I have considered the 2nd option and I decided to not change it because it felt like a big change. We can discuss that and I could make a separate PR with that.
It is not really big. Just dropping a method from the interface. I think it's better do it as part of this PR as then we will have more places to change and the effort will be greater

georgii-borovinskikh-sonarsource: the 1stproposal does not make sense when the 2nd proposal is done. the 2nd proposal will cause more test changes and I already have a new commit based on this current PR

gabriela-trutan-sonarsource: I created a [follow up task](https://sonarsource.atlassian.net/browse/SLVS-2093).

gabriela-trutan-sonarsource (approved):

PR #6182, opened 2025-05-01
SLVS-2096 Use analysis properties with SLCore analysis

georgii-borovinskikh-sonarsource:


src/SLCore/Analysis/SLCoreAnalyzer.cs
```diff
@@ -136,7 +125,7 @@ [new FileUri(path)],
     }
 
     private IDisposable EnrichPropertiesForCFamily(
-        Dictionary<string, string> properties,
+        ref ImmutableDictionary<string, string> properties,
```
vnaskos-sonar, line 128: I can understand that is used to return a second object from this method. However, a `ref` of an `ImmutableDictionary` seems strange. 🤔

vnaskos-sonar (approved):
👍

PR #6183, opened 2025-05-01
SLVS-2061 Update SonarLint targets file to support solution-level settings

georgii-borovinskikh-sonarsource:


vnaskos-sonar (approved):
👍

PR #6191, opened 2025-05-06
SLVS-2059 Support solution-level settings.json

georgii-borovinskikh-sonarsource:
[SLVS-2059](https://sonarsource.atlassian.net/browse/SLVS-2059)

Part of 

[SLVS-2059]: https://sonarsource.atlassian.net/browse/SLVS-2059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
```diff
@@ -0,0 +1,244 @@
...
+            () =>
+            {
+                if (disposed)
+                {
+                    return;
+                }
+
+                // Note: the data is stored in the roaming profile so it will be sync across machines
+                // for domain-joined users.
+                appDataRoot = environmentVariableProvider.GetSLVSAppDataRootPath();
```
gabriela-trutan-sonarsource, line 82: Please extract the logic of subscribing to the global settings events into a method similar to the `CreateSolutionSettingsMonitorAndSubscribe`

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
```diff
@@ -0,0 +1,244 @@
...
+                GlobalAnalysisSettingsFilePath = Path.GetFullPath(Path.Combine(appDataRoot, SettingsFileName));
+                generatedGlobalSettingsFolder = Path.Combine(appDataRoot, GeneratedGlobalSettingsFolderName);
```
gabriela-trutan-sonarsource, line 84: We don't have to store the folder name. We already have the full path, so we can use the `Path.GetDirectoryName(fullPath)`

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
```diff
@@ -0,0 +1,244 @@
...
+using SonarLint.VisualStudio.Core.SystemAbstractions;
+using SonarLint.VisualStudio.Core.UserRuleSettings;
+using SonarLint.VisualStudio.Infrastructure.VS.Initialization;
+using SonarLint.VisualStudio.Integration.Resources;
+
+namespace SonarLint.VisualStudio.Integration.UserSettingsConfiguration;
+
+[Export(typeof(IUserSettingsProvider))]
+[PartCreationPolicy(CreationPolicy.Shared)]
+internal sealed class UserSettingsProvider : IUserSettingsProvider, IDisposable
```
gabriela-trutan-sonarsource, line 37: This class is getting very large. We should split the logic of creating the file monitor, the settings file, disposing etc for each of the settings (global/solution level) into two other separate classes. These classes can even have a common interface.

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
```diff
@@ -0,0 +1,244 @@
...
+                globalSettingsFileMonitor.FileChanged += OnFileChanged;
+                activeSolutionTracker.ActiveSolutionChanged += ActiveSolutionTrackerOnActiveSolutionChanged;
+            });
+    }
+
+    private void CreateSolutionSettingsMonitorAndSubscribe(string solutionName)
+    {
+        var solutionSettingsParentFolder = Path.Combine(appDataRoot, SolutionSettingsFolderName, solutionName);
+        SolutionAnalysisSettingsFilePath = Path.GetFullPath(Path.Combine(solutionSettingsParentFolder, SettingsFileName));
+        generatedSolutionSettingsFolder = solutionSettingsParentFolder;
```
gabriela-trutan-sonarsource, line 100: We don't have to store the folder name. We already have the full path, so we can use the Path.GetDirectoryName(fullPath)

gabriela-trutan-sonarsource (changes requested):

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
```diff
@@ -0,0 +1,244 @@
...
+        solutionSettingsFileMonitor.FileChanged += OnFileChanged;
+    }
+
+    private void ActiveSolutionTrackerOnActiveSolutionChanged(object sender, ActiveSolutionChangedEventArgs e)
+    {
+        DisposeSolutionSettingsMonitor();
+
+        if (!e.IsSolutionOpen)
+        {
+            ResetConfiguration();
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Sure, I can consider changing this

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Not sure I get what you're trying to describe. Could you explain?

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
```diff
@@ -0,0 +1,244 @@
...
+                    CreateSolutionSettingsMonitorAndSubscribe(solutionName);
+                }
+                globalSettingsFileMonitor.FileChanged += OnFileChanged;
+                activeSolutionTracker.ActiveSolutionChanged += ActiveSolutionTrackerOnActiveSolutionChanged;
+            });
+    }
+
+    private void CreateSolutionSettingsMonitorAndSubscribe(string solutionName)
+    {
+        var solutionSettingsParentFolder = Path.Combine(appDataRoot, SolutionSettingsFolderName, solutionName);
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: 🤔

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
georgii-borovinskikh-sonarsource: Why do we need to recalculate the unchanged value every time?

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Open/close a solution and make sure that "Common Language Runtime Exceptions" are enabled. You will see that there is an exception thrown that the `SlCoreRuleSettings` that we are trying to access the service from a UI thread.

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
gabriela-trutan-sonarsource: Not to have duplicate variables that practically store the same information (one can easily be calculated/deduced from the other)

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
georgii-borovinskikh-sonarsource: `UserSettings` property should still return the combined value of global and solution settings. I can split the interface into `IUserSettingsProvider` with just the event and the property, `IGlobalSettingsUpdater` with EnsureGlobalConfigExists, DisableRule, etc & `ISolutionSettingsUpdater` with EnsureSolutionConfigExists, etc. However, I can't think of a quick and easy way to split the implementation

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
georgii-borovinskikh-sonarsource: Introducing clearly unnecessary calculations and allocations just to not have a single field is unreasonable

gabriela-trutan-sonarsource: I find unreasonable to cache unnecessary information and then make sure to correctly update the cache of 4 fields. Feels a lot of duplicate effort.

georgii-borovinskikh-sonarsource: It's a single field. Which is updated only in one place.

georgii-borovinskikh-sonarsource: Also, it's not even true for the `generatedGlobalSettingsFolder`, as the folder is next to the settings file and not at the settings file location

gabriela-trutan-sonarsource: > It's a single field. Which is updated only in one place.
`generatedGlobalSettingsFolder`, `generatedSolutionSettingsFolder`, `SolutionAnalysisSettingsFilePath`, `GlobalAnalysisSettingsFilePath`

gabriela-trutan-sonarsource: > Also, it's not even true for the generatedGlobalSettingsFolder, as the folder is next to the settings file and not at the settings file location

You are right about that

src/Core/UserRuleSettings/IUserSettingsProvider.cs
```diff
@@ -32,6 +34,11 @@ public interface IUserSettingsProvider
     /// </summary>
     string GlobalAnalysisSettingsFilePath { get; }
 
+    /// <summary>
```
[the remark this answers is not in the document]
vnaskos-sonar: At the moment the global and solution settings are being handled by the `IUserSettingsProvider` which may look messy, which is a nice change. (If I understood this is the pain point you want to address with this comment).

On the other hand it is convenient that the user settings are handled in a single place. For example it will be easy to handle the conflicting settings. The amount of settings we have are not that many or complicated to require yet another layer of abstraction.

In case it is decided to create the `ISolutionSettingsProvider` and `IGlobalSettingsProvider`, I expect that the `IUserSettingsProvider` will remain and still be the responsible to **hide** the global and solution implementation from the rest of the code (I mean classes that need to know about the settings will only depend on the `IUserSettingsProvider`, and the `IUserSettingsProvider` will use the `ISolutionSettingsProvider` and `IGlobalSettingsProvider`).

Unless there is something that I missed or a greater need for it, this can be postponed for a future task. 🤔

src/Core.UnitTests/UserRuleSettings/UserSettingsProviderTests.cs
```diff
@@ -1,395 +0,0 @@
...
-        // Act
-        userSettingsProvider.EnsureGlobalAnalysisSettingsFileExists();
-
-        // Assert
-        fileSystem.File.Received(1).Exists(SettingsFilePath);
-        fileSystem.File.DidNotReceive().WriteAllText(Arg.Any<string>(), Arg.Any<string>());
-    }
-
-    [TestMethod]
-    public void RealFile_DisableRule_FileDoesNotExist_FileCreated()
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: There are real file tests in the serializer tests. I can see if they're missing any cases that were removed from here and add them back

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: The code here should have been correct, but this is what happens 
image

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Normally solution name should not be null when IsSolutionOpen == true

src/Core/UserRuleSettings/IUserSettingsProvider.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: @vnaskos-sonar https://github.com/SonarSource/sonarlint-visualstudio/pull/6191#discussion_r2077298829

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I think VS might be creating a dummy solution when you open settings.json without any existing solution

src/Core/UserRuleSettings/IUserSettingsProvider.cs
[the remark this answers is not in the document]
gabriela-trutan-sonarsource: > Unless there is something that I missed or a greater need for it, this can be postponed for a future task.

You understood correctly. I am not against creating another task for it.

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
gabriela-trutan-sonarsource: Can you, please, create a sub-task for it? And if we have time, we do it.

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I've added a quick if condition to the event handler, but:
https://sonarsource.atlassian.net/browse/SLVS-2112
https://sonarsource.atlassian.net/browse/SLVS-2113

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Yeah, `SlCoreRuleSettings`/`AnalysisConfigMonitor` have some incorrect implicit assumptions regarding threading. I need to fix that

src/Core.UnitTests/UserRuleSettings/UserSettingsProviderTests.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Seems like Vasilis has already added a lot of serialization tests at GlobalAnalysisSettingsTests.cs & SolutionAnalysisSettingsTests.cs. Adding more real file tests wouldn't really add any value here, because the serialization to string is equivalent

src/Core/UserRuleSettings/IUserSettingsProvider.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: https://sonarsource.atlassian.net/browse/SLVS-2114

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
georgii-borovinskikh-sonarsource: https://sonarsource.atlassian.net/browse/SLVS-2114

src/Core.UnitTests/UserRuleSettings/UserSettingsProviderTests.cs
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: The new tests were added in this sprint and they already cover the cases in deleted tests. The deleted tests are complicated and are hard to maintain. Just copy-pasting wouldn't work since I changed the serialization stuff in this PR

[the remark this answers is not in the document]
gabriela-trutan-sonarsource: Our objective this year is to add more integration tests, not less. I created another [task](https://sonarsource.atlassian.net/browse/SLVS-2115) for this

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: These are not the type of integration tests that are meant by the objective. We need end-to-end tests, not unit tests which are redundantly creating files to write exactly the same content as in the string created by the serializer.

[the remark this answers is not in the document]
gabriela-trutan-sonarsource: We can never have end-to-end tests with WPF, so it's about integration tests.
It's not about the string being serialized correctly or not (although I did mention it helped me in detecting problems), it's about testing the integration test settings being generated correctly. Taking into account that the logic is more complicated now: we have global settings, solution level settings and we expect to merge the two, why shouldn't we try to tests all that?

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: We have unit tests, which cover more cases. I don't see the point of creating an integration test which just combines multiple existing unit tests and adds a fileSystem.WriteAllText in the middle. We don't need to integration test the file system, there's no value in that. What we're interested in is integration testing services (SLCore, SQS/SQC) and UI.

src/Integration.Vsix/Analysis/AnalysisConfigMonitor.cs
```diff
@@ -91,27 +91,26 @@ public AnalysisConfigMonitor(
...
+        ).Forget();
     }
 
     #endregion Incoming notifications
 
-    private async Task OnSettingsChangedAsync()
-    {
-        await threadHandling.SwitchToBackgroundThread();
-
+    private void RequestAnalysis() =>
```
gabriela-trutan-sonarsource, line 111: We could inline the method

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
```diff
@@ -0,0 +1,248 @@
...
+    private readonly IActiveSolutionTracker activeSolutionTracker;
+    private readonly ILogger logger;
+    private readonly IAnalysisSettingsSerializer serializer;
+    private readonly ISingleFileMonitorFactory fileMonitorFactory;
+    private ISingleFileMonitor globalSettingsFileMonitor;
+    private ISingleFileMonitor solutionSettingsFileMonitor;
+    private UserSettings userSettings;
+    private string appDataRoot;
+    private bool disposed;
+    private (string settings, string generated) globalFilePaths;
```
gabriela-trutan-sonarsource, line 53: The name of the tuple is not speaking at all
```suggestion
    private (string fullPath, string folder) globalFilePaths;
```

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
```diff
@@ -0,0 +1,248 @@
...
+    private (string settings, string generated)? solutionFilePaths;
```
gabriela-trutan-sonarsource, line 54: The name of the tuple is not speaking at all
```suggestion
    private (string fullPath, string folder) solutionFilePaths;
```

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
```diff
@@ -0,0 +1,248 @@
...
+                if (disposed)
+                {
+                    return;
+                }
+
+                // Note: the data is stored in the roaming profile so it will be sync across machines
+                // for domain-joined users.
+                appDataRoot = environmentVariableProvider.GetSLVSAppDataRootPath();
+                CreateGlobalSettingsMonitorAndSubscribe();
+                if (activeSolutionTracker.CurrentSolutionName is { } solutionName)
```
gabriela-trutan-sonarsource, line 84: Why is the null check not inside `CreateSolutionSettingsMonitorAndSubscribe`?

gabriela-trutan-sonarsource (approved):

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
georgii-borovinskikh-sonarsource: it's not really fullpath and folder, those are independent paths to settings file and to generated folder. 
solutionFilePaths(to)settings, solutionFilePaths(to)generated, etc.

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
georgii-borovinskikh-sonarsource: The null check is still needed outside the method for the other case

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
gabriela-trutan-sonarsource: `Generated` doesn't tell me anything about what that is. All the folders from the bindings, settings are generated.
You can rename them better, if you can find better names. It was just a suggestion. Other suggestions: `settingsFullPath` and `solutionSettingsFolder`.

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
gabriela-trutan-sonarsource: The other null check could stay.

georgii-borovinskikh-sonarsource: I'd rather keep it consistent and let the caller check the argument that is passed

PR #6199, opened 2025-05-09
SLVS-2114 Split the UserSettingsProvider

gabriela-trutan-sonarsource:
[SLVS-2114](https://sonarsource.atlassian.net/browse/SLVS-2114)

[SLVS-2114]: https://sonarsource.atlassian.net/browse/SLVS-2114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Integration/UserSettingsConfiguration/GlobalSettingsStorage.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    public void SaveSettingsFile(GlobalAnalysisSettings settings)
+    {
+        serializer.SafeSave(SettingsFilePath, settings);
+        InvokeSettingsFileChanged();
+    }
+
+    public GlobalAnalysisSettings LoadSettingsFile() => serializer.SafeLoad<GlobalAnalysisSettings>(SettingsFilePath);
```
georgii-borovinskikh-sonarsource, line 92: If this value is cached here, AnalysisSettings.GlobalFileExclusions can be removed and GlobalSettingsProvider can use this value directly instead of going through UserSettingsProvider

src/Integration/UserSettingsConfiguration/SolutionSettingsStorage.cs
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    public void SaveSettingsFile(SolutionAnalysisSettings settings)
+    {
+        serializer.SafeSave(SettingsFilePath, settings);
+        InvokeSettingsFileChanged();
+    }
+
+    public SolutionAnalysisSettings LoadSettingsFile() => serializer.SafeLoad<SolutionAnalysisSettings>(SettingsFilePath);
```
georgii-borovinskikh-sonarsource, line 102: If this value is cached here, AnalysisSettings.SolutionFileExclusions can be removed and SolutionSettingsProvider can use this value directly instead of going through UserSettingsProvider

src/Core/UserRuleSettings/IUserSettingsProvider.cs
```diff
@@ -29,45 +30,31 @@ public interface IUserSettingsProvider : IRequireInitialization
...
-    /// Full path to the file containing the user settings
+    /// Updates the user settings to include the provided global/solution file exclusions. The value will override existing exclusions.
     /// </summary>
-    string SolutionAnalysisSettingsFilePath { get; }
+    void UpdateFileExclusions(IEnumerable<string> exclusions);
+
+    ImmutableArray<string> FileExclusions { get; }
+}
 
+public interface IGlobalSettingsProvider : IRequireInitialization, IFileExclusionsProvider
```
georgii-borovinskikh-sonarsource, line 46: IGlobalUserSettingsProvider

src/Core/UserRuleSettings/IUserSettingsProvider.cs
```diff
@@ -29,45 +30,31 @@ public interface IUserSettingsProvider : IRequireInitialization
...
-    /// Updates the user settings to include the provided global file exclusions. The value will override existing exclusions.
-    /// </summary>
-    void UpdateGlobalFileExclusions(IEnumerable<string> exclusions);
-
-    /// <summary>
-    /// Updates the user settings to include the provided solution file exclusions. The value will override existing exclusions.
-    /// </summary>
-    void UpdateSolutionFileExclusions(IEnumerable<string> exclusions);
-
+public interface ISolutionSettingsProvider : IRequireInitialization, IFileExclusionsProvider
```
georgii-borovinskikh-sonarsource, line 54: ISolutionUserSettingsProvider

src/Integration/UserSettingsConfiguration/UserSettingsProvider.cs
```diff
@@ -155,59 +80,40 @@ public UserSettings UserSettings
...
         }
     }
 
-    public void EnsureSolutionAnalysisSettingsFileExists()
+    private void OnSettingsFileChanged(object sender, EventArgs e) => ResetConfiguration();
+
+    private void ActiveSolutionTrackerOnActiveSolutionChanged(object sender, ActiveSolutionChangedEventArgs e)
     {
-        if (SolutionAnalysisSettingsFilePath == null)
+        if (!e.IsSolutionOpen || e.SolutionName == null)
```
georgii-borovinskikh-sonarsource, line 104: There is a potential issue here. This event handler needs to be executed strictly after the `SolutionSettingsStorage's` event handler. If not, then when this raises `SettingsChanged` and its event handlers are then requesting to load the new settings, the file path to the solution settings is not initialized. Currently, this is not happening because the event ordering is ensured by this class' dependency on `SolutionSettingsStorage`, however this is would be very easy to miss when modifying these classes. I suggest adding a comment here to explain that.

src/Core/UserRuleSettings/IUserSettingsStorage.cs
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.Core.Initialization;
+
+namespace SonarLint.VisualStudio.Core.UserRuleSettings;
+
+public interface IUserSettingsStorage : IRequireInitialization, IDisposable
```
georgii-borovinskikh-sonarsource, line 25: can be internal

georgii-borovinskikh-sonarsource (changes requested):
Remove duplicate properties of IXSettingsProvider from AnalysisSettings

src/Core/UserRuleSettings/IUserSettingsStorage.cs
gabriela-trutan-sonarsource: It can not as the solution/global settings storage are extending this interface. It will lead to inconsistent accessibility

src/Integration/UserSettingsConfiguration/SolutionSettingsProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public void UpdateAnalysisProperties(Dictionary<string, string> analysisProperties)
+    {
+        var userSettings = userSettingsProvider.UserSettings;
+        var solutionSettings = new SolutionAnalysisSettings(analysisProperties, userSettings.AnalysisSettings.SolutionFileExclusions);
+        solutionSettingsStorage.SaveSettingsFile(solutionSettings);
+    }
+
+    public IInitializationProcessor InitializationProcessor { get; } = processorFactory.CreateAndStart<SolutionSettingsProvider>(
```
georgii-borovinskikh-sonarsource, line 52: This primary constructor & initializer thing is confusing. Use an explicit constructor.

src/Integration/UserSettingsConfiguration/GlobalSettingsProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public void UpdateFileExclusions(IEnumerable<string> exclusions)
+    {
+        var userSettings = userSettingsProvider.UserSettings;
+        var globalSettings = new GlobalAnalysisSettings(userSettings.AnalysisSettings.Rules, exclusions.ToImmutableArray());
+        globalSettingsStorage.SaveSettingsFile(globalSettings);
+    }
+
+    public IInitializationProcessor InitializationProcessor { get; } = processorFactory.CreateAndStart<GlobalSettingsProvider>(
```
georgii-borovinskikh-sonarsource, line 55: This primary constructor & initializer thing is confusing. Use an explicit constructor.

src/Integration/UserSettingsConfiguration/GlobalSettingsProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        var userSettings = userSettingsProvider.UserSettings;
+        var newRules = userSettings.AnalysisSettings.Rules.SetItem(ruleId, new RuleConfig(RuleLevel.Off));
+        var globalSettings = new GlobalAnalysisSettings(newRules, userSettings.AnalysisSettings.GlobalFileExclusions);
+        globalSettingsStorage.SaveSettingsFile(globalSettings);
+    }
+
+    public void UpdateFileExclusions(IEnumerable<string> exclusions)
+    {
+        var userSettings = userSettingsProvider.UserSettings;
```
georgii-borovinskikh-sonarsource, line 50: we should check that this class was initialized before doing anything

src/Integration/UserSettingsConfiguration/GlobalSettingsProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+[method: ImportingConstructor]
+internal class GlobalSettingsProvider(
+    IGlobalSettingsStorage globalSettingsStorage,
+    IUserSettingsProvider userSettingsProvider,
+    IInitializationProcessorFactory processorFactory)
+    : IGlobalSettingsProvider
+{
+    public void DisableRule(string ruleId)
+    {
+        Debug.Assert(!string.IsNullOrEmpty(ruleId), "DisableRule: ruleId should not be null/empty");
```
georgii-borovinskikh-sonarsource, line 40: we should check that this class was initialized before doing anythin

src/Integration/UserSettingsConfiguration/SolutionSettingsProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+[method: ImportingConstructor]
+internal class SolutionSettingsProvider(
+    ISolutionSettingsStorage solutionSettingsStorage,
+    IUserSettingsProvider userSettingsProvider,
+    IInitializationProcessorFactory processorFactory)
+    : ISolutionSettingsProvider
+{
+    public void UpdateFileExclusions(IEnumerable<string> exclusions)
+    {
+        var userSettings = userSettingsProvider.UserSettings;
```
georgii-borovinskikh-sonarsource, line 40: we should check that this class was initialized before doing anything

src/Integration/UserSettingsConfiguration/SolutionSettingsProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+        var solutionSettings = new SolutionAnalysisSettings(userSettings.AnalysisSettings.AnalysisProperties, exclusions.ToImmutableArray());
+        solutionSettingsStorage.SaveSettingsFile(solutionSettings);
+    }
+
+    public void UpdateAnalysisProperties(Dictionary<string, string> analysisProperties)
+    {
+        var userSettings = userSettingsProvider.UserSettings;
```
georgii-borovinskikh-sonarsource, line 47: we should check that this class was initialized before doing anything

src/Integration/UserSettingsConfiguration/GlobalSettingsProvider.cs
gabriela-trutan-sonarsource: What should happen if it is not initialized? Not doing anything is unexpected, throwing an exception is also unexcpected.

gabriela-trutan-sonarsource: I made sure instead that the initialization task is awaited

PR #6216, opened 2025-05-15
SLVS-2138 Add context menu for reviewing security hotspot

gabriela-trutan-sonarsource:


jblievremont (approved):
Looks OK from a code point of view, I could not check the visibility of the context menu (no Windows box readily available) but I'll try to test once the whole feature is there 🙂

PR #6223, opened 2025-05-21
SLVS-2126 Implement filtering the hotspots by priority

gabriela-trutan-sonarsource:
[SLVS-2126](https://sonarsource.atlassian.net/browse/SLVS-2126)

This targets the [PR](https://github.com/SonarSource/sonarlint-visualstudio/pull/6222) in which the location filter was added, because the two filters should be applied together.
 
[SLVS-2126]: https://sonarsource.atlassian.net/browse/SLVS-2126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/IssueViz.Security/Hotspots/HotspotsList/ViewModels/HotspotsControlViewModel.cs
```diff
@@ -229,24 +238,33 @@ private void OnActiveDocumentChanged(object sender, ActiveDocumentChangedEventAr
...
-        private ObservableCollection<IHotspotViewModel> GetFilteredHotspots()
+        private void UpdateFilteredHotspots()
         {
             filteredHotspots.Clear();
-            var currentHotspots = GetHotspotsFilteredByLocationFilter();
-            currentHotspots.ToList().ForEach(filteredHotspots.Add);
-
-            return filteredHotspots;
+            var hotspotsToShow = GetHotspotsFilteredByLocationFilter(hotspots.ToList());
+            hotspotsToShow = GetHotspotsFilteredByPriorityFilter(hotspotsToShow.ToList());
```
georgii-borovinskikh-sonarsource, line 251: do we need to `.ToList()` twice here?

src/IssueViz.Security/Hotspots/HotspotsList/HotspotsControl.xaml
```diff
@@ -64,7 +65,66 @@
...
+                    <Trigger Property="IsMouseOver" Value="True">
+                        <Setter Property="BorderThickness" Value="1" />
+                        <Setter Property="Background" Value="{DynamicResource {x:Static vsShell:VsBrushes.CommandBarHoverKey}}" />
+                    </Trigger>
+                </Style.Triggers>
+            </Style>
+
             <Style TargetType="{x:Type ComboBox}" BasedOn="{StaticResource {x:Static vsShell:VsResourceKeys.ComboBoxStyleKey}}" />
+
+            <ControlTemplate x:Key="BorderedContentTemplate" TargetType="ContentControl">
```
georgii-borovinskikh-sonarsource, line 99: `x:Key` value is too generic

georgii-borovinskikh-sonarsource (approved):

src/IssueViz.Security/Hotspots/HotspotsList/ViewModels/HotspotsControlViewModel.cs
georgii-borovinskikh-sonarsource: They're both private methods that return IEnumerable, there's no need to be so defensive with the implementation. I think they could just accept IEnumerable themselves and only the third ToList is needed (line 252)

PR #6224, opened 2025-05-22
SLVS-2147 Add a column to display the status of a hotspot in the hotspot window

gabriela-trutan-sonarsource:


georgii-borovinskikh-sonarsource:
@gabriela-trutan-sonarsource ok, good to know

PR #6240, opened 2025-06-02
SLVS-2079 Refactor VCXCompilationDatabaseProvider

georgii-borovinskikh-sonarsource:
[SLVS-2079](https://sonarsource.atlassian.net/browse/SLVS-2079)

[SLVS-2079]: https://sonarsource.atlassian.net/browse/SLVS-2079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Integration.Vsix/CFamily/VcxProject/CompilationDatabaseEntryGenerator.cs
```diff
@@ -0,0 +1,98 @@
...
+        return environmentEntries;
+    }
+
+    private ImmutableList<EnvironmentEntry> UpdateEnvironmentWithEntry(ImmutableList<EnvironmentEntry> environmentEntries, EnvironmentEntry newEntry)
+    {
+        EnvironmentEntry oldEntry = environmentEntries.FirstOrDefault(x => x.Name == newEntry.Name);
+
+        if (oldEntry.Name != null)
+        {
+            logger.LogVerbose($"[VCXCompilationDatabaseProvider] Overwriting the value of environment variable \"{newEntry.Name}\". Old value: \"{oldEntry.Value}\", new value: \"{newEntry.Value}\"");
```
gabriela-trutan-sonarsource: As part of CaYC we should use logging context.

gabriela-trutan-sonarsource:
```suggestion
            logger.LogVerbose($"[CompilationDatabaseEntryGenerator] Overwriting the value of environment variable \"{newEntry.Name}\". Old value: \"{oldEntry.Value}\", new value: \"{newEntry.Value}\"");
```

src/Integration.Vsix/CFamily/VcxProject/CompilationDatabaseEntryGenerator.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+        else
+        {
+            logger.LogVerbose($"[VCXCompilationDatabaseProvider] Setting environment variable \"{newEntry.Name}\". Value: \"{newEntry.Value}\"");
```
gabriela-trutan-sonarsource:
```suggestion
            logger.LogVerbose($"[CompilationDatabaseEntryGenerator] Setting environment variable \"{newEntry.Name}\". Value: \"{newEntry.Value}\"");
```

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
```diff
@@ -0,0 +1,116 @@
...
+            if (databasePath is null)
+            {
+                throw new InvalidOperationException(CFamilyStrings.ActiveVcxCompilationDatabase_NotInitialized);
+            }
+
+            var compilationDatabaseEntry = generator.CreateOrNull(filePath);
+
+            if (compilationDatabaseEntry is null)
+            {
+                return;
```
gabriela-trutan-sonarsource: We should still log this, at least at the verbose level.

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+
+            storage.UpdateDatabaseEntry(databasePath, compilationDatabaseEntry);
+        }
+    }
+
+    public void RemoveFile(string filePath)
+    {
+        threadHandling.ThrowIfOnUIThread();
```
gabriela-trutan-sonarsource: We should remove the call to `ThrowIfOnUIThread`. The storage should already execute the operation on the background thread.

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
```diff
@@ -0,0 +1,116 @@
...
+                return;
+            }
+            storage.DeleteDatabase(databasePath);
+            databasePath = null;
+        }
+    }
+
+    public void AddFile(string filePath)
+    {
+        threadHandling.ThrowIfOnUIThread();
```
gabriela-trutan-sonarsource: We should remove the call to ThrowIfOnUIThread.

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
```diff
@@ -0,0 +1,110 @@
...
+        lock (lockObject)
+        {
+            if (databasePath is not null)
+            {
+                throw new InvalidOperationException(CFamilyStrings.ActiveVcxCompilationDatabase_AlreadyInitialized);
+            }
+
+            databasePath = storage.CreateDatabase();
```
gabriela-trutan-sonarsource: We should avoid performing IO operations under lock as they can be slow and unpredictable. We should consider using a `SemaphoreSlim` instead.

src/Integration.Vsix.UnitTests/CFamily/VcxProject/ActiveVcxCompilationDatabaseTests.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        VerifyThrowIfOnUIThread(2);
+        testSubject.DatabasePath.Should().BeNull();
+        storage.Received(1).DeleteDatabase(DatabasePath);
+    }
+
+    [TestMethod]
+    public void AddFile_NotInitialized_Throws()
+    {
+        storage.CreateDatabase().Returns(DatabasePath);
```
gabriela-trutan-sonarsource: This call is not needed for this test.

src/Integration.Vsix.UnitTests/CFamily/VcxProject/ActiveVcxCompilationDatabaseTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        testSubject.InitializeDatabase();
+
+        testSubject.AddFile(EntryFilePath);
+
+        storage.DidNotReceiveWithAnyArgs().UpdateDatabaseEntry(default, default);
+        VerifyThrowIfOnUIThread(2);
+    }
+
+    [TestMethod]
+    public void RemoveFile_NotInitialized_DoesNothing()
```
gabriela-trutan-sonarsource: We should log this case.
```suggestion
    public void RemoveFile_NotInitialized_DoesNothingAndLogs()
```

src/Integration.Vsix.UnitTests/CFamily/VcxProject/ActiveVcxCompilationDatabaseTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        testSubject.InitializeDatabase();
+
+        testSubject.AddFile(EntryFilePath);
+
+        storage.Received(1).UpdateDatabaseEntry(DatabasePath, entry);
+        VerifyThrowIfOnUIThread(2);
+    }
+
+    [TestMethod]
+    public void AddFile_Initialized_EntryCannotBeGenerated_DoesNothing()
```
gabriela-trutan-sonarsource: We should log this case.
```suggestion
    public void AddFile_Initialized_EntryCannotBeGenerated_DoesNothingAndLogs()
```

src/Integration.Vsix.UnitTests/CFamily/VcxProject/ActiveVcxCompilationDatabaseTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        testSubject.AddFile(EntryFilePath);
+
+        storage.DidNotReceiveWithAnyArgs().UpdateDatabaseEntry(default, default);
+        VerifyThrowIfOnUIThread(2);
+    }
+
+    [TestMethod]
+    public void RemoveFile_NotInitialized_DoesNothing()
+    {
+        storage.CreateDatabase().Returns(DatabasePath);
```
gabriela-trutan-sonarsource: This call is not needed for this test

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    public void RemoveFile(string filePath)
+    {
+        threadHandling.ThrowIfOnUIThread();
+
+        lock (lockObject)
+        {
+            if (databasePath is null)
+            {
+                return;
```
gabriela-trutan-sonarsource: We should still log this, at least at the verbose level.

src/Integration.Vsix/CFamily/VcxProject/CompilationDatabaseEntryGenerator.cs
gabriela-trutan-sonarsource: As part of CaYC we should move into resource.

src/Integration.Vsix/CFamily/VcxProject/VCXCompilationDatabaseProvider.cs
```diff
@@ -18,76 +18,27 @@
...
-        IEnvironmentVariableProvider environmentVariableProvider,
-        IFileConfigProvider fileConfigProvider,
-        ILogger logger)
-    {
-        this.storage = storage;
-        this.fileConfigProvider = fileConfigProvider;
-        this.logger = logger;
-        staticEnvironmentVariableEntries = ImmutableList.CreateRange(environmentVariableProvider.GetAll().Select(x => new EnvironmentEntry(x.name, x.value)));
-    }
+    private readonly CompilationDatabaseEntryGenerator generator = new(fileConfigProvider, environmentVariableProvider, logger);
```
gabriela-trutan-sonarsource: Why aren't we importing the service instead?

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
georgii-borovinskikh-sonarsource: SemaptoreSlim would not be much different from a lock if we're not using AcquireAsync. I did change the interface to be asynchronous, so I used that instead of the lock.

src/Integration.Vsix.UnitTests/CFamily/VcxProject/ActiveVcxCompilationDatabaseTests.cs
georgii-borovinskikh-sonarsource: Don't see the point of logging removed files if the whol DB has been removed already

src/Integration.Vsix/CFamily/VcxProject/VCXCompilationDatabaseProvider.cs
georgii-borovinskikh-sonarsource: The class is going to be removed alongside with the test. Was easier to do this than to adapt it & tests to the changes

src/Integration.Vsix.UnitTests/CFamily/VcxProject/ActiveVcxCompilationDatabaseTests.cs
gabriela-trutan-sonarsource: Well, because it might be helpful in debugging. E.g. we tried to remove an entry from the db, but the DB does not exist 
This case should not actually happen

src/CFamily/IActiveVcxCompilationDatabase.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    Task AddFileAsync(string filePath);
+    Task RemoveFileAsync(string filePath);
+}
+
+public interface IActiveVcxCompilationDatabase : IVcxCompilationDatabaseUpdater, IDisposable
+{
+    string DatabasePath { get; }
+
+    Task<string> EnsureDatabaseInitializedAsync();
```
gabriela-trutan-sonarsource, line 33: For the public method I liked the old name better as it doesn't give too much info about the internal implementation. The private method can still stay `EnsureDatabaseInitializedAsync`.

```suggestion
    Task<string> InitializeDatabaseAsync();
```

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
```diff
@@ -0,0 +1,151 @@
...
+    private string databasePath;
+    private bool disposed = false;
+
+    public string DatabasePath
+    {
+        get
+        {
+            ThrowIfDisposed();
+
+            threadHandling.ThrowIfOnUIThread();
```
gabriela-trutan-sonarsource, line 49: Please remove this call.

src/Integration.Vsix.UnitTests/CFamily/VcxProject/ActiveVcxCompilationDatabaseTests.cs
```diff
@@ -0,0 +1,304 @@
...
+
+        (await testSubject.EnsureDatabaseInitializedAsync()).Should().Be(DatabasePath);
+
+        VerifyRequiresAsynchronousBackgroundExecution<string>(1);
+        testSubject.DatabasePath.Should().Be(DatabasePath);
+    }
+
+    [TestMethod]
+    public async Task EnsureDatabaseInitialized_AlreadyInitialized_Throws()
```
gabriela-trutan-sonarsource:
```suggestion
    public async Task EnsureDatabaseInitialized_AlreadyInitialized_ReturnsExistingPath()
```

src/Integration.Vsix.UnitTests/CFamily/VcxProject/ActiveVcxCompilationDatabaseTests.cs
```diff
@@ -0,0 +1,304 @@
...
+
+    [TestMethod]
+    public async Task AddFile_NotInitialized_InitializesAndAddsFile()
+    {
+        storage.CreateDatabase().Returns(DatabasePath);
+        generator.CreateOrNull(EntryFilePath).Returns(entry);
+
+        await testSubject.AddFileAsync(EntryFilePath);
+
+        storage.Received(1).UpdateDatabaseEntry(DatabasePath, entry);
```
gabriela-trutan-sonarsource: We already check that on line 160
```suggestion
   
```

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
georgii-borovinskikh-sonarsource: The lock below will block the caller until the IO operations complete. We should not call this property from the UI thread, hence the check

src/CFamily/IActiveVcxCompilationDatabase.cs
georgii-borovinskikh-sonarsource: I was trying to name it so that it's not unexpected that it doesn't initialize a new db if one already is initialized

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
gabriela-trutan-sonarsource: I don't see that as being a problem or cause for throwing an exception. Exception should be thrown when the program is in an unrecoverable state that prevents it from normal execution, which is not the case here. 
The UI would just be a little bit slower.

src/CFamily/IActiveVcxCompilationDatabase.cs
gabriela-trutan-sonarsource: It is not unexpected. The caller is not aware how the db is initialized. It just expected that the db is initialized, which it is.

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
georgii-borovinskikh-sonarsource: UI being slower is bad and this exception signals to the developer who's using this code that they're doing something wrong. We use it like that in many places

src/CFamily/IActiveVcxCompilationDatabase.cs
georgii-borovinskikh-sonarsource: `EnsureSmth` means do Smth if it hasn't been done already, `DoSmth` means do Smth regardless

gabriela-trutan-sonarsource: I still think in our case the db is being initialized regardless, so the `Ensure` is not really needed.

src/Integration.Vsix/CFamily/VcxProject/ActiveVcxCompilationDatabase.cs
gabriela-trutan-sonarsource: We can take the discussion outside this PR.

src/CFamily/IActiveVcxCompilationDatabase.cs
gabriela-trutan-sonarsource: Anyway, the name is not wrong, so I am fine with it as well.

PR #6242, opened 2025-06-03
SLVS-2158 Remove the responsibility of ensure settings file exists in consumers

gabriela-trutan-sonarsource:
[SLVS-2158](https://sonarsource.atlassian.net/browse/SLVS-2158)

The consumers of the `GlobalSettingsStorage` and `SolutionSettingsStorage` should not need to have the knowledge about which settings file should exist as when new consumers are added this would have to be remembered, polluting the code in different places and make it more difficult to maintain.

[SLVS-2158]: https://sonarsource.atlassian.net/browse/SLVS-2158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/UserRuleSettings/IUserSettingsStorage.cs
+++ b/src/Core/UserRuleSettings/IUserSettingsStorage.cs
@@ -40,4 +40,5 @@ public interface IGlobalSettingsStorage : IUserSettingsStorage<GlobalRawAnalysis
 
 public interface ISolutionSettingsStorage : IUserSettingsStorage<SolutionRawAnalysisSettings>
 {
+    void EnsureSettingsFileExists();
 }
--- a/src/Integration.UnitTests/UserSettingsConfiguration/SolutionSettingsStorageTest.cs
+++ b/src/Integration.UnitTests/UserSettingsConfiguration/SolutionSettingsStorageTest.cs
@@ -23,6 +23,7 @@ using NSubstitute.ReturnsExtensions;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.FileMonitor;
 using SonarLint.VisualStudio.Core.Initialization;
+using SonarLint.VisualStudio.Core.SystemAbstractions;
 using SonarLint.VisualStudio.Core.UserRuleSettings;
 using SonarLint.VisualStudio.Integration.UserSettingsConfiguration;
 using SonarLint.VisualStudio.TestInfrastructure;
@@ -39,6 +40,7 @@ public class SolutionSettingsStorageTest
     private const string Solution2SettingsFilePath = @"C:\some\path\to\appdata\SonarLint for Visual Studio\SolutionSettings\Solution TWO\settings.json";
     private IActiveSolutionTracker activeSolutionTracker;
     private IEnvironmentVariableProvider environmentVariableProvider;
+    private IFileSystemService fileSystem;
     private IInitializationProcessorFactory processorFactory;
     private IAnalysisSettingsSerializer serializer;
     private ISingleFileMonitorFactory singleFileMonitorFactory;
@@ -50,6 +52,7 @@ public class SolutionSettingsStorageTest
     public void Initialize()
     {
         testLogger = new TestLogger();
+        fileSystem = Substitute.For<IFileSystemService>();
         serializer = Substitute.For<IAnalysisSettingsSerializer>();
         singleFileMonitorFactory = Substitute.For<ISingleFileMonitorFactory>();
         threadHandling = Substitute.ForPartsOf<NoOpThreadHandler>();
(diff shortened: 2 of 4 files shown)


src/Integration/UserSettingsConfiguration/GlobalSettingsStorage.cs
```diff
@@ -111,7 +103,16 @@ private void CreateGlobalSettingsMonitorAndSubscribe()
         var globalAnalysisSettingsFilePath = Path.GetFullPath(Path.Combine(appDataRoot, SettingsFileName));
         var generatedGlobalSettingsFolder = Path.Combine(appDataRoot, GeneratedGlobalSettingsFolderName);
         globalFilePaths = (globalAnalysisSettingsFilePath, generatedGlobalSettingsFolder);
+        EnsureSettingsFileExists();
```
georgii-borovinskikh-sonarsource, line 106: Groundhog day. We've discussed the fact that we shouldn't create empty files. I don't really see the need in this PR, since all it achieves is to remove the call to EnsureSettingsFileExists when we want to open an empty settings file to allow for the manual edit.

gabriela-trutan-sonarsource: It is not "empty files", it's just one. I don't see the problem of ensuring that the settings file exists in the global settings storage. 

The point is exactly that (see PR and task description). If we add more consumers (i.e. more places in the UI where we open the settings file), then we always have to remember to ensure that the file exists, otherwise it will crash. The responsibility of the file creation is all wrong

georgii-borovinskikh-sonarsource: You can create an extension method for the interface `IUserSettingsStorage` called `OpenFile` and call `Ensure` + `DocumentOpener` there

gabriela-trutan-sonarsource: I don't think this solves the issue. The `DocumentOpener` can still be called by consumers that are not aware that file should exist.

Can you explain what the problem is with the current approach?

georgii-borovinskikh-sonarsource: Can you ask Farah if she thinks it's an issue if we generate files for every solution we opened? If she says it's not an issue, I will approve the PR, as the implementation is otherwise correct

gabriela-trutan-sonarsource: We do not generate "files" for every solution we open. The implementation for the settings file of `SolutionSettingsStorage` has not changed: they are only generated if the file exclusions/analysis properties are added. 

We only generate the global `settings.json` once the `GlobalSettingsStorage` is initialized (which happens only once per VS instance)

georgii-borovinskikh-sonarsource: But this has changed https://github.com/SonarSource/sonarlint-visualstudio/pull/6242/files#diff-437e5c607cffa8cd35b0f60448c4a1707dbf04b26f99d6f03c71231d85aa21d1L73

So, is there a bug then? Or am I missing something?

gabriela-trutan-sonarsource: You are missing the fact that no file will be opened if it doesn't exist. Which is ironically leading to the advantage of creating less files, if no setting was defined.

georgii-borovinskikh-sonarsource: That button **shoud** create the file if it's missing. This is to allow the users to edit it manually without needing to first edit it via ui

gabriela-trutan-sonarsource: Added back the `EnsureSettingsFileExists` method to the `ISolutionSettingsStorage` (but not to the `IUserSettingsStorage<T>`) to allow the feature of opening the `settings.json` for a specific solution.

The current state is: 

- global `settings.json` is generated every time the implementation of `IGlobalSettingsStorage` is initialized 
- solution `settings.json` is generated only when file exclusions/analysis properties are added or when the user clicks on opening the file

An empty folder is still generated for each opened solution due to the `SingleFileMonitor` [implementation](https://github.com/SonarSource/sonarlint-visualstudio/blob/59ce2b415186a3eac4b800ae85d2f973e748ed2d/src/Core/FileMonitor/SingleFileMonitor.cs#L60). I could improve that in this PR, and dispose such folders in the `SolutionSettingsStorage` when a closed solution has an empty folder or even a folder with an empty settings file. What do you think?

georgii-borovinskikh-sonarsource: Could we have a more generic PR for the fix of SingleFileMonitor? Put the deletion code into it's dispose

georgii-borovinskikh-sonarsource (approved):

src/Integration/UserSettingsConfiguration/GlobalSettingsStorage.cs
gabriela-trutan-sonarsource: https://sonarsource.atlassian.net/browse/SLVS-2288

PR #6243, opened 2025-06-04
SLVS-2162 Fix flaky tests

gabriela-trutan-sonarsource:
[SLVS-2162](https://sonarsource.atlassian.net/browse/SLVS-2162)

[SLVS-2162]: https://sonarsource.atlassian.net/browse/SLVS-2162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorTests.cs
+++ b/src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorTests.cs
@@ -112,17 +112,22 @@ public class InitializationProcessorTests
     {
         var dependency1 = Substitute.For<IRequireInitialization>();
         var dependency2 = Substitute.For<IRequireInitialization>();
+        var manualResetEvent = new ManualResetEvent(false);
+        initialization.When(x => x.Invoke(Arg.Any<IThreadHandling>())).Do(x =>
+        {
+            // block initialization to increase the chance of multiple threads trying to initialize at the same time
+            manualResetEvent.WaitOne(50);
+        });
         var lockImplementation = new AsyncLock();
         asyncLockFactory.Create().AcquireAsync().Returns(_ => lockImplementation.AcquireAsync());
         var testSubject = testSubjectFactory.Create<InitializationProcessorTests>([dependency1, dependency2], initialization);
 
-        await Task.WhenAll(
-            Enumerable.Range(1, 20)
-                .AsParallel()
-                .Select(_ => testSubject.InitializeAsync()));
+        var tasks = Enumerable.Range(1, 20).AsParallel().Select(_ => testSubject.InitializeAsync()).ToArray();
+        manualResetEvent.Set();
+        await Task.WhenAll(tasks);
 
         testSubject.IsFinalized.Should().BeTrue();
-        asyncLockFactory.Create().Received(Quantity.Within(1, 20)).AcquireAsync();
+        asyncLockFactory.Create().Received(Quantity.Within(2, 20)).AcquireAsync();
         initialization.ReceivedWithAnyArgs(1).Invoke(default);
         dependency1.InitializationProcessor.Received(1).InitializeAsync();
         dependency2.InitializationProcessor.Received(1).InitializeAsync();


src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorTests.cs
```diff
@@ -123,7 +122,7 @@ await Task.WhenAll(
                 .Select(_ => testSubject.InitializeAsync()));
 
         testSubject.IsFinalized.Should().BeTrue();
-        asyncLockFactory.Create().Received(Quantity.Within(2, 20)).AcquireAsync();
+        asyncLockFactory.Create().Received(Quantity.Within(1, 20)).AcquireAsync();
```
georgii-borovinskikh-sonarsource: This makes the test kinda useless. The point was to force the contention and validate that it was handled correctly (initialization ran once). If there is no contention (in the case of (1) call), the test doesn't test anything beyond what other tests already do.

Please, introduce a ManualResetEvent in the `initialization` func and unblock it after launching the WhenAll task. Possibly add a Task.Run to the Select

src/Integration.UnitTests/Helpers/RetryHelperTests.cs
```diff
@@ -74,9 +68,7 @@ public void RetryOnException_WhenFailsTwiceWithAFiveMilliSecondDelay_ReturnsFals
 
             // Assert
             result.Should().BeFalse();
-            timeWatch.ElapsedMilliseconds.Should().BeInRange(
-                (delayMilliseconds * 2) - 200,
-                (delayMilliseconds * 2) + 200);
+            timeWatch.ElapsedMilliseconds.Should().BeGreaterOrEqualTo(delayMilliseconds * 2);
```
georgii-borovinskikh-sonarsource, line 71: Do we want this test to pass if it took too much time? Maybe we should put an upper limit like (2*delay, 2*delay+1sec)

src/Integration.Vsix.UnitTests/SonarLintTagger/CancellableJobRunnerTests.cs
```diff
@@ -129,7 +114,15 @@ public void CancelAfterFirstOperation()
                 testLogger.WriteLine("[Test] Executing op1");
                 op1Executed = true;
 
-                testSubject.Cancel();
+                // in some cases, the op1 is executed before the CancellableJobRunner.Start returned the runner, so testSubject can be null
```
georgii-borovinskikh-sonarsource: Could put a ManualResetEvent here and release it after Start has returned (and before waiting for the runner to complete). It took me a while to understand this change

georgii-borovinskikh-sonarsource (changes requested):

src/Integration.UnitTests/Helpers/RetryHelperTests.cs
gabriela-trutan-sonarsource: Yes, we want to. The upper limit is what the made the test flaky in the first place. Depending on the system load, the `Task.Delay` can take more time than the provided amount, so there is never a good upper limit to pick. 
I also don't see a difference between 1sec and the way it is right now

georgii-borovinskikh-sonarsource: I'd still prefer if it had some sort of upper limit, like we have in our integration tests, so that it can indicate that the test has failed without waiting indefinitely

gabriela-trutan-sonarsource: Adding the upper limit check will not prevent the test from waiting indefinitely. 
Again the upper limit is what made the test flaky, so no choice is a good one as the `Task.Delay` execution is out of our control. What matters is that the retry has happened the amount of `x` times we provided for the `delay` amount we provided, which is "at least `x` seconds"

georgii-borovinskikh-sonarsource: Yeah it will require to do something like `await Task.WhenAny(Task.Run(() => test), Task.Delay(upperLimit))`

gabriela-trutan-sonarsource: I don't see the point in doing that. There is no danger for the test running indefinitely.

src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorTests.cs
gabriela-trutan-sonarsource: We still check that the initialization is ran once. But the case of the `AcquireAsync` is a valid one: one thread acquires the lock performs the initialization and all the other ones do not even reach the  `AcquireAsync`, because the `testSubject` is initialized. This case does not happen all the time (as we see from the flakiness of the test), but it's still a valid point that can not be prevented due to the concurrency nature of running the tasks

src/Infrastructure.VS.UnitTests/Initialization/InitializationProcessorTests.cs
```diff
@@ -113,14 +112,19 @@ public async Task InitializeAsync_CompletesOnlyOnceFromMultipleThreads()
     {
         var dependency1 = Substitute.For<IRequireInitialization>();
         var dependency2 = Substitute.For<IRequireInitialization>();
+        var manualResetEvent = new ManualResetEvent(false);
+        initialization.When(x => x.Invoke(Arg.Any<IThreadHandling>())).Do(x =>
+        {
+            // block initialization to increase the chance of multiple threads trying to initialize at the same time
+            manualResetEvent.WaitOne(50);
```
georgii-borovinskikh-sonarsource, line 119: This invocation has 50 as the timeout, but the one in the other test doesn't 🤔

gabriela-trutan-sonarsource: The other one doesn't really needed as the `CancellableJobRunner.Start` will for sure return.  But the missing timeout in here is blocking the test so the signal is never invoked on line 126.

PR #6244, opened 2025-06-04
SLVS-2167 Create one compilation database per VS instance

gabriela-trutan-sonarsource:
[SLVS-2167](https://sonarsource.atlassian.net/browse/SLVS-2167)

[SLVS-2167]: https://sonarsource.atlassian.net/browse/SLVS-2167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Integration.Vsix/SonarLintDaemonPackage.cs
+++ b/src/Integration.Vsix/SonarLintDaemonPackage.cs
@@ -136,7 +136,7 @@ namespace SonarLint.VisualStudio.Integration.Vsix
 
             if (disposing)
             {
-                DisposeCompilationDatabaseStorageAsync().Forget();
+                DisposeCompilationDatabaseStorage();
                 projectDocumentsEventsListener?.Dispose();
                 projectDocumentsEventsListener = null;
                 solutionRoslynAnalyzerManager?.Dispose();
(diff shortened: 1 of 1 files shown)


src/Integration.Vsix/SonarLintDaemonPackage.cs
```diff
@@ -146,6 +146,13 @@ protected override void Dispose(bool disposing)
             }
         }
 
+        private async Task DisposeCompilationDatabaseStorageAsync()
+        {
+            await vcxCompilationDatabase.DropDatabaseAsync();
```
georgii-borovinskikh-sonarsource: Don't really have to explicitly drop the database, as the storage is going to delete the whole folder when it's disposed

src/Integration.Vsix/SonarLintDaemonPackage.cs
```diff
@@ -135,8 +136,7 @@ protected override void Dispose(bool disposing)
 
             if (disposing)
             {
-                vcxCompilationDatabaseStorage?.Dispose();
-                vcxCompilationDatabaseStorage = null;
+                DisposeCompilationDatabaseStorageAsync().Forget();
```
georgii-borovinskikh-sonarsource: This means that there is a chance we won't delete the file, because the process might shut down before this is complete

georgii-borovinskikh-sonarsource (changes requested):

src/Integration.Vsix/SonarLintDaemonPackage.cs
georgii-borovinskikh-sonarsource: Call it synchronously, or remove the DropDatabase call completely (see https://github.com/SonarSource/sonarlint-visualstudio/pull/6244#discussion_r2126144625)

src/Integration.Vsix/SonarLintDaemonPackage.cs
gabriela-trutan-sonarsource: I think it's just cleaner.

PR #6245, opened 2025-06-04
SLVS-2080 Add/remove compilation database entries when opening, renaming and closing C/C++ files

georgii-borovinskikh-sonarsource:
[SLVS-2080](https://sonarsource.atlassian.net/browse/SLVS-2080)

Part of 

[SLVS-2080]: https://sonarsource.atlassian.net/browse/SLVS-2080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/DocumentEvents.cs
+++ b/src/Core/DocumentEvents.cs
@@ -25,7 +25,7 @@ namespace SonarLint.VisualStudio.Core;
 public abstract class DocumentEventArgs(string fullPath,  IEnumerable<AnalysisLanguage> detectedLanguages) : EventArgs
 {
     /// <summary>
-    /// Full path of the document of the document for which the event was raised
+    /// Full path of the document for which the event was raised
     /// </summary>
     public string FullPath { get; } = fullPath;
     public IEnumerable<AnalysisLanguage> DetectedLanguages { get; } = detectedLanguages;


src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs
```diff
@@ -334,6 +334,19 @@ public void WhenFileIsRenamed_AnalysisIsNotRequested()
         VerifyAnalysisNotRequested();
     }
 
+    [TestMethod]
```
gabriela-trutan-sonarsource, line 324: This is a duplicate of `WhenFileIsRenamed_OpenDocumentRenamedEventIsRaised` from line [302](https://github.com/SonarSource/sonarlint-visualstudio/blob/d74a7f58c5ed88a1a18ee1d1f9a8c64720fdf5a5/src/Integration.Vsix.UnitTests/SonarLintTagger/TextBufferIssueTrackerTests.cs#L302).

Please remove one of them

src/Integration.Vsix/CFamily/VcxProject/VcxDocumentEventsHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+
+    private void DocumentEventsOnOpenDocumentRenamed(object sender, DocumentRenamedEventArgs e)
+    {
+        if (e.DetectedLanguages.Contains(AnalysisLanguage.CFamily))
+        {
+            ReplaceFileAsync(e.OldFilePath, e.FullPath).Forget();
+        }
+
+
+        async Task ReplaceFileAsync(string oldFilePath, string fullPath)
```
gabriela-trutan-sonarsource: Is there a reason for this being a local function? We should [prefer methods](https://github.com/SonarSource/sonarlint-visualstudio/blob/master/docs/coding_style.md#local-functions) over local function.

src/Integration.Vsix.UnitTests/CFamily/VcxProject/VcxDocumentEventsHandlerTests.cs
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    [TestMethod]
+    public void Dispose_UnsubscribesFromOpenDocumentRenamed()
+    {
+        var renameArgs = new DocumentRenamedEventArgs("old.cpp", "new.cpp", [AnalysisLanguage.CFamily]);
+
+        testSubject.Dispose();
+
+        documentEvents.OpenDocumentRenamed += Raise.EventWith(documentEvents, renameArgs);
```
gabriela-trutan-sonarsource: I find the empty lines above/bellow this confusing. Is there a reason for them? 
Also we don't need to actually raise the event. Checking that the event was unsubscribed should be enough, so we could even remove it.

src/Integration.Vsix.UnitTests/CFamily/VcxProject/VcxDocumentEventsHandlerTests.cs
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    [TestMethod]
+    public void Dispose_UnsubscribesFromDocumentClosed()
+    {
+        var closeArgs = new DocumentClosedEventArgs("file.cpp", [AnalysisLanguage.CFamily]);
+
+        testSubject.Dispose();
+
+        documentEvents.DocumentClosed += Raise.EventWith(documentEvents, closeArgs);
```
gabriela-trutan-sonarsource: I find the empty lines above/bellow this confusing. Is there a reason for them? 
Also we don't need to actually raise the event. Checking that the event was unsubscribed should be enough, so we could even remove it.

src/Integration.Vsix.UnitTests/CFamily/VcxProject/VcxDocumentEventsHandlerTests.cs
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    [TestMethod]
+    public void Dispose_UnsubscribesFromDocumentOpened()
+    {
+        var openArgs = new DocumentOpenedEventArgs("file.cpp", [AnalysisLanguage.CFamily]);
+
+        testSubject.Dispose();
+
+        documentEvents.DocumentOpened += Raise.EventWith(documentEvents, openArgs);
```
gabriela-trutan-sonarsource: I find the empty lines above/bellow this confusing. Is there a reason for them? 
Also we don't need to actually raise the event. Checking that the event was unsubscribed should be enough, so we could even remove it.

src/Integration.Vsix.UnitTests/CFamily/VcxProject/VcxDocumentEventsHandlerTests.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        documentEvents.DocumentOpened += Raise.EventWith(documentEvents, args);
+
+        vcxCompilationDatabaseUpdater.DidNotReceive().AddFileAsync(Arg.Any<string>());
+    }
+
+    [TestMethod]
+    public void DocumentClosed_NonCFamily_NoRemoveFileAsyncCalled()
+    {
+        var args = new DocumentClosedEventArgs("file.cs", [AnalysisLanguage.Javascript]);
```
gabriela-trutan-sonarsource: We could extract the `args` into fields or, at least, the file names into constants.

src/Integration.Vsix.UnitTests/CFamily/VcxProject/VcxDocumentEventsHandlerTests.cs
```diff
@@ -0,0 +1,154 @@
...
+    {
+        var renameArgs = new DocumentRenamedEventArgs("old.cpp", "new.cpp", [AnalysisLanguage.CFamily]);
+
+        testSubject.Dispose();
+
+        documentEvents.OpenDocumentRenamed += Raise.EventWith(documentEvents, renameArgs);
+
+        vcxCompilationDatabaseUpdater.DidNotReceive().RemoveFileAsync(Arg.Any<string>());
+        vcxCompilationDatabaseUpdater.DidNotReceive().AddFileAsync(Arg.Any<string>());
+        documentEvents.Received().OpenDocumentRenamed -= Arg.Any<EventHandler<DocumentRenamedEventArgs>>();
```
gabriela-trutan-sonarsource: We are also missing a test that the events were subscribed to in the constructor.

src/Core/DocumentEvents.cs
```diff
@@ -22,27 +22,25 @@
 
 namespace SonarLint.VisualStudio.Core;
 
-public abstract class DocumentEventArgs(string fullPath) : EventArgs
+public abstract class DocumentEventArgs(string fullPath,  IEnumerable<AnalysisLanguage> detectedLanguages) : EventArgs
 {
     /// <summary>
     /// Full file path to the document being closed
```
gabriela-trutan-sonarsource: CaYC: please update wrong comment

gabriela-trutan-sonarsource (changes requested):
There is sometimes an exception thrown when a file is renamed and it is not being added to the compilation database. 

Steps to reproduce: 
1. Rename "Hotspot1.cpp" to "Hotspot12.cpp"
2. Rename "Hotspot12.cpp to "Hotspot1.cpp"

image


If you think we should fix this in another task, then I can approve this PR, because it looks good.

src/Integration.Vsix.UnitTests/CFamily/VcxProject/VcxDocumentEventsHandlerTests.cs
georgii-borovinskikh-sonarsource: Yeah I think the smart AI messed up my request to separate Arrange, Act and Assert. I will remove unnecessary empty lines

georgii-borovinskikh-sonarsource:
> There is sometimes an exception thrown when a file is renamed (in which case, it is not being added to the compilation database).
> 
> Steps to reproduce:
> 
> 1. Rename "Hotspot1.cpp" to "Hotspot12.cpp"
> 2. Rename "Hotspot12.cpp to "Hotspot1.cpp"
> 
> image
> 
> If you think we should fix this in another task, then I can approve this PR, because it looks good.

@gabriela-trutan-sonarsource 
I can put a try/catch in the CompilationDatabaseEntryGenerator, but I don't think the underlying issue is caused by this PR, as I'm passing the correct file name to AddFile.

gabriela-trutan-sonarsource:
> @gabriela-trutan-sonarsource
> I can put a try/catch in the CompilationDatabaseEntryGenerator, but I don't think the underlying issue is caused by this PR, as I'm passing the correct file name to AddFile.

It is not caused by your PR, but it should be fixed as there is a real problem. That's why I mentioned it can be addressed in another task. A try/catch will not fix the issue it would just treat a symptom, but the problem will remain: the file won't be added to the compilation database.

src/Integration.Vsix/CFamily/VcxProject/VcxDocumentEventsHandler.cs
georgii-borovinskikh-sonarsource: It's not really useful as a private method, I only needed it so I can use `.Forget()` on it. I can remove the parameters to make it more like an actual local function

PR #6252, opened 2025-06-06
Fix threading in mef-importing constructor of DiffViewService

georgii-borovinskikh-sonarsource:
Part of SLVS-1987


gabriela-trutan-sonarsource:
@georgii-borovinskikh-sonarsource, we should target master in this PR

PR #6254, opened 2025-06-06
SLVS-2292 Fix threading in mef-importing ctor of BoundSolutionGitMonitor

georgii-borovinskikh-sonarsource:
[SLVS-2292](https://sonarsource.atlassian.net/browse/SLVS-2292)

Part of SLVS-1986

[SLVS-2292]: https://sonarsource.atlassian.net/browse/SLVS-2292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ConnectedMode.UnitTests/BoundSolutionGitMonitorTests.cs
+++ b/src/ConnectedMode.UnitTests/BoundSolutionGitMonitorTests.cs
@@ -18,192 +18,246 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System;
-using SonarLint.VisualStudio.ConnectedMode.UnitTests.Extensions;
 using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.Initialization;
 using SonarLint.VisualStudio.TestInfrastructure;
-using static SonarLint.VisualStudio.ConnectedMode.BoundSolutionGitMonitor;
 
-namespace SonarLint.VisualStudio.ConnectedMode.UnitTests
+namespace SonarLint.VisualStudio.ConnectedMode.UnitTests;
+
+[TestClass]
+public class BoundSolutionGitMonitorTests
 {
-    [TestClass]
-    public class BoundSolutionGitMonitorTests
+    private IGitWorkspaceService gitWorkspaceService;
+    private BoundSolutionGitMonitor.GitEventFactory factory;
+    private IGitEvents gitEvents;
+    private TestLogger logger;
+    private IThreadHandling threadHandling;
+    private IInitializationProcessorFactory initializationProcessorFactory;
+
+    [TestInitialize]
+    public void TestInitialize()
     {
-        [TestMethod]
-        public void MefCtor_CheckIsExported()
+        gitEvents = Substitute.For<IGitEvents>();
+        gitWorkspaceService = Substitute.For<IGitWorkspaceService>();
+        factory = Substitute.For<BoundSolutionGitMonitor.GitEventFactory>();
+        factory.Invoke(Arg.Any<string>()).Returns(gitEvents);
+        logger = Substitute.ForPartsOf<TestLogger>();
+        threadHandling = Substitute.ForPartsOf<NoOpThreadHandler>();
+    }
+
+    private BoundSolutionGitMonitor CreateUninitializedTestSubject(out TaskCompletionSource<byte> barrier)
+    {
+        var tcs = barrier = new TaskCompletionSource<byte>();
+        initializationProcessorFactory = MockableInitializationProcessor.CreateFactory<BoundSolutionGitMonitor>(threadHandling, logger, proc => MockableInitializationProcessor.ConfigureWithWait(proc, tcs));
+        return new BoundSolutionGitMonitor(gitWorkspaceService, logger, initializationProcessorFactory, factory);
+    }
+
+
+    private BoundSolutionGitMonitor CreateAndInitializeTestSubject()
+    {
+        initializationProcessorFactory = MockableInitializationProcessor.CreateFactory<BoundSolutionGitMonitor>(threadHandling, logger);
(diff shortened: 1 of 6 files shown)


src/ConnectedMode/BoundSolutionGitMonitor.cs
```diff
@@ -18,130 +18,137 @@
...
-                if (disposing)
-                {
-                    CleanupLocalGitEventResources();
-                }
+            logger.LogVerbose(Resources.GitMonitor_NoRepo);
+        }
+        else
+        {
+            logger.LogVerbose(Resources.GitMonitor_MonitoringRepoStarted, rootPath);
+            currentRepoEvents = createLocalGitMonitor(rootPath);
```
gabriela-trutan-sonarsource: `currentRepoEvents` should be thread safe, because the instance might be overridden and we would not unsubscribe from the event, which would lead to memory leaks.

src/ConnectedMode.UnitTests/BoundSolutionGitMonitorTests.cs
```diff
@@ -18,192 +18,246 @@
...
+        logger = Substitute.ForPartsOf<TestLogger>();
+        threadHandling = Substitute.ForPartsOf<NoOpThreadHandler>();
+    }
 
-        [TestMethod]
-        public void Initialize_NoRepo_FactoryNotCalledAndNoError()
-        {
-            var gitWorkspaceService = CreateGitWorkSpaceService(null);
-            var factory = new Mock<GitEventFactory>();
+    private BoundSolutionGitMonitor CreateUninitializedTestSubject(out TaskCompletionSource<byte> barrier)
```
gabriela-trutan-sonarsource: Helper methods should go at the bottom.

src/ConnectedMode.UnitTests/BoundSolutionGitMonitorTests.cs
```diff
@@ -18,192 +18,246 @@
...
+        initializationProcessorFactory = MockableInitializationProcessor.CreateFactory<BoundSolutionGitMonitor>(threadHandling, logger, proc => MockableInitializationProcessor.ConfigureWithWait(proc, tcs));
+        return new BoundSolutionGitMonitor(gitWorkspaceService, logger, initializationProcessorFactory, factory);
+    }
 
-            var testSubject = CreateTestSubject(gitWorkspaceService.Object, factory.Object);
 
-            gitWorkspaceService.Verify(x => x.GetRepoRoot(), Times.Once);
-            factory.Invocations.Should().HaveCount(0);
-        }
+    private BoundSolutionGitMonitor CreateAndInitializeTestSubject()
```
gabriela-trutan-sonarsource: Helper methods should go at the bottom.

gabriela-trutan-sonarsource (changes requested):
We should make `currentRepoEvents` thread safe

src/ConnectedMode/BoundSolutionGitMonitor.cs
georgii-borovinskikh-sonarsource: I don't think there's a race condition here. `RefreshInternal` is only called by:

1) the initializer, before the initialization is finished
2) ASBT from the UI thread after it is finished (and not before)

gabriela-trutan-sonarsource: There is no race condition yet. But there might be and we should defend our code against that

src/ConnectedMode/BoundSolutionGitMonitor.cs
```diff
@@ -112,19 +113,22 @@ private void OnHeadChanged(object sender, EventArgs e)
 
     private void RefreshInternal()
     {
-        CleanupLocalGitEventResources();
+        lock (lockObject)
```
gabriela-trutan-sonarsource: The `gitWorkspaceService.GetRepoRoot` could be a very expensive operation, therefore I don't think it should be under lock

gabriela-trutan-sonarsource: The `currentRepoEvents` can still be set to null in a non-thread safe manner, so line 130 might throw.

gabriela-trutan-sonarsource: We should rather have a property that controls the event subscription/un-subscription. Something like:

```
        private IGitEvents CurrentRepoEvents
        {
            get => currentRepoEvents;
            set
            {
                lock (lockObject)
                {
                    if (currentRepoEvents != null)
                    {
                        currentRepoEvents.HeadChanged -= OnHeadChanged;
                        currentRepoEvents.Dispose();
                    }
                    currentRepoEvents = value;
                    if (currentRepoEvents != null)
                    {
                        currentRepoEvents.HeadChanged += OnHeadChanged;
                    }
                }
            }
        }
```

georgii-borovinskikh-sonarsource: Too much code for a property

gabriela-trutan-sonarsource: I wouldn't say so.

georgii-borovinskikh-sonarsource: That can only happen on dispose, I think that's fine

gabriela-trutan-sonarsource: Why would it be fine? It's not really thread safe

georgii-borovinskikh-sonarsource: There's no impact really after the vs starts to shutdown

gabriela-trutan-sonarsource: There might be if the method is `RefreshInternal()` is also run in the same time.

vnaskos-sonar (approved):
👍

gabriela-trutan-sonarsource (approved):

PR #6264, opened 2025-06-11
SLVS-2041, SLVS-2166 Update compilation database on config scope change

georgii-borovinskikh-sonarsource:
[SLVS-2041](https://sonarsource.atlassian.net/browse/SLVS-2041)
[SLVS-2166](https://sonarsource.atlassian.net/browse/SLVS-2166)

[SLVS-2041]: https://sonarsource.atlassian.net/browse/SLVS-2041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SLVS-2166]: https://sonarsource.atlassian.net/browse/SLVS-2166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/CFamily/IActiveCompilationDatabaseTracker.cs
+++ b/src/Core/CFamily/IActiveCompilationDatabaseTracker.cs
@@ -1,0 +1,28 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.Core.Initialization;
+
+namespace SonarLint.VisualStudio.Core.CFamily;
+
+public interface IActiveCompilationDatabaseTracker : IRequireInitialization, IDisposable
+{
+    string DatabasePath { get; }
+}
--- a/src/Integration.Vsix.UnitTests/CFamily/ActiveCompilationDatabaseTrackerTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/ActiveCompilationDatabaseTrackerTests.cs
@@ -1,0 +1,387 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using NSubstitute.ReturnsExtensions;
+using SonarLint.VisualStudio.CFamily;
+using SonarLint.VisualStudio.CFamily.CMake;
+using SonarLint.VisualStudio.Core;
+using SonarLint.VisualStudio.Core.CFamily;
+using SonarLint.VisualStudio.Core.ConfigurationScope;
+using SonarLint.VisualStudio.Core.Initialization;
+using SonarLint.VisualStudio.Core.Synchronization;
+using SonarLint.VisualStudio.Integration.Vsix.CFamily;
+using SonarLint.VisualStudio.SLCore.Core;
+using SonarLint.VisualStudio.SLCore.Service.Analysis;
+
+namespace SonarLint.VisualStudio.Integration.UnitTests.CFamily;
+
+[TestClass]
+public class ActiveCompilationDatabaseTrackerTests
+{
+    private const string DefaultConfigScope = "scope1";
+    private const string CmakeJson = "cmake.json";
+    private const string VcxJson = "vcx.json";
+    private ICMakeCompilationDatabaseLocator cMakeCompilationDatabaseLocator;
+    private IActiveVcxCompilationDatabase activeVcxCompilationDatabase;
+    private IActiveConfigScopeTracker activeConfigScopeTracker;
+    private IInitializationProcessorFactory initializationProcessorFactory;
+    private ICFamilyAnalysisConfigurationSLCoreService cFamilyAnalysisConfigurationSlCore;
+    private IAsyncLockFactory asyncLockFactory;
+    private IThreadHandling threadHandling;
+    private TestLogger testLogger;
+    private ISLCoreServiceProvider serviceProvider;
+    private IAsyncLock asyncLock;
+
+    private readonly IRequireInitialization[] initializationDependencies = [];
+
+    [TestInitialize]
+    public void TestInitialize()
+    {
+        cMakeCompilationDatabaseLocator = Substitute.For<ICMakeCompilationDatabaseLocator>();
+        activeVcxCompilationDatabase = Substitute.For<IActiveVcxCompilationDatabase>();
+        activeConfigScopeTracker = Substitute.For<IActiveConfigScopeTracker>();
+        initializationProcessorFactory = Substitute.For<IInitializationProcessorFactory>();
+        cFamilyAnalysisConfigurationSlCore = Substitute.For<ICFamilyAnalysisConfigurationSLCoreService>();
+        asyncLockFactory = Substitute.For<IAsyncLockFactory>();
+        threadHandling = Substitute.ForPartsOf<NoOpThreadHandler>();
+        testLogger = Substitute.ForPartsOf<TestLogger>();
+        serviceProvider = Substitute.For<ISLCoreServiceProvider>();
+        SetUpServiceProvider();
+        asyncLock = Substitute.For<IAsyncLock>();
+        asyncLockFactory.Create().Returns(asyncLock);
+    }
+
+    [TestMethod]
+    public void MefCtor_CheckIsExported() =>
+        MefTestHelpers.CheckTypeCanBeImported<ActiveCompilationDatabaseTracker, IActiveCompilationDatabaseTracker>(
+            MefTestHelpers.CreateExport<ICMakeCompilationDatabaseLocator>(),
+            MefTestHelpers.CreateExport<IActiveVcxCompilationDatabase>(),
+            MefTestHelpers.CreateExport<IActiveConfigScopeTracker>(),
+            MefTestHelpers.CreateExport<IInitializationProcessorFactory>(),
+            MefTestHelpers.CreateExport<IAsyncLockFactory>(),
+            MefTestHelpers.CreateExport<IThreadHandling>(),
+            MefTestHelpers.CreateExport<ISLCoreServiceProvider>());
+
+    [TestMethod]
+    public void MefCtor_CheckIsSingleton() => MefTestHelpers.CheckIsSingletonMefComponent<ActiveCompilationDatabaseTracker>();
+
+    [TestMethod]
+    public void Ctor_RunsInitialization()
+    {
+        var testSubject = CreateAndInitializeTestSubject();
+
+        Received.InOrder(() =>
+        {
+            initializationProcessorFactory.Create<ActiveCompilationDatabaseTracker>(Arg.Is<IReadOnlyCollection<IRequireInitialization>>(x => x.SequenceEqual(initializationDependencies)),
+                Arg.Any<Func<IThreadHandling, Task>>());
+            testSubject.InitializationProcessor.InitializeAsync();
+            activeVcxCompilationDatabase.EnsureDatabaseInitializedAsync();
+            activeConfigScopeTracker.CurrentConfigurationScopeChanged += Arg.Any<EventHandler<ConfigurationScopeChangedEventArgs>>();
+            var releaser = asyncLock.AcquireAsync().GetAwaiter().GetResult();
+            releaser.Dispose();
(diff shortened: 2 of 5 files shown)


src/Core/CFamily/IActiveCompilationDatabaseTracker.cs
```diff
@@ -0,0 +1,28 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarLint.VisualStudio.Core.Initialization;
+
+namespace SonarLint.VisualStudio.Core.CFamily;
```
gabriela-trutan-sonarsource, line 23: Why is this in `Core` assembly? It can be moved to `SonarLint.VisualStudio.Integration.Vsix.CFamily`

src/Integration.Vsix.UnitTests/CFamily/ActiveCompilationDatabaseTrackerTests.cs
```diff
@@ -0,0 +1,340 @@
...
+            var releaser = asyncLock.AcquireAsync().GetAwaiter().GetResult();
+            releaser.Dispose();
+            testSubject.InitializationProcessor.InitializeAsync(); // called by CreateAndInitializeTestSubject
+        });
+        testSubject.DatabasePath.Should().BeNull();
+        VerifyOnBackgroundThreadLockTakenAndReleased(1);
+    }
+
+    [TestMethod]
+    public void WhenUninitialized_DoesNotReactToConfigScopeChangedEvents()
```
gabriela-trutan-sonarsource:
```suggestion
    public void Ctor_WhenUninitialized_DoesNotReactToConfigScopeChangedEvents()
```

src/Integration.Vsix.UnitTests/CFamily/ActiveCompilationDatabaseTrackerTests.cs
```diff
@@ -0,0 +1,340 @@
...
+        activeVcxCompilationDatabase.DidNotReceiveWithAnyArgs().DropDatabaseAsync();
+        activeConfigScopeTracker.DidNotReceiveWithAnyArgs().CurrentConfigurationScopeChanged += Arg.Any<EventHandler>();
+        activeConfigScopeTracker.DidNotReceiveWithAnyArgs().CurrentConfigurationScopeChanged -= Arg.Any<EventHandler>();
+        _ = activeConfigScopeTracker.DidNotReceiveWithAnyArgs().Current;
+    }
+
+    [TestMethod]
+    public void Ctor_ConfigScopeExists_InitializesWithCMakeDatabase()
+    {
+        SetCurrentConfiguration("scope1","cmake.json", null);
```
gabriela-trutan-sonarsource: Extract all hard-coded strings in constants

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
```diff
@@ -0,0 +1,133 @@
...
+                {
+                    return;
+                }
+                await activeVcxCompilationDatabase.EnsureDatabaseInitializedAsync();
+                activeConfigScopeTracker.CurrentConfigurationScopeChanged += ActiveConfigScopeTrackerOnCurrentConfigurationScopeChanged;
+                await HandleConfigScopeEventAsync();
+            });
+    }
+
+    private void ActiveConfigScopeTrackerOnCurrentConfigurationScopeChanged(object sender, EventArgs e) =>
```
gabriela-trutan-sonarsource: The name is very long and adding the event in the name doesn't really bring any more info

```suggestion
    private void OnCurrentConfigurationScopeChanged(object sender, EventArgs e) =>
```

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
```diff
@@ -0,0 +1,133 @@
...
+            if (activeConfigScopeTracker.Current is { Id: { } currentConfigScopeId })
+            {
+                if (currentConfigScopeId == lastConfigScopeId)
+                {
+                    return;
+                }
+
+                if (!serviceProvider.TryGetTransientService(out ICFamilyAnalysisConfigurationSLCoreService cFamilyAnalysisConfiguration))
+                {
+                    return;
```
gabriela-trutan-sonarsource: Please log this case, because it means SlCore is not informed about the path to the compilation database, although we expect to.

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
```diff
@@ -0,0 +1,133 @@
...
+    private void ActiveConfigScopeTrackerOnCurrentConfigurationScopeChanged(object sender, EventArgs e) =>
+        threadHandling.RunOnBackgroundThread(HandleConfigScopeEventAsync).Forget();
+
+    private async Task HandleConfigScopeEventAsync()
+    {
+        using (await asyncLock.AcquireAsync())
+        {
+            if (activeConfigScopeTracker.Current is { Id: { } currentConfigScopeId })
+            {
+                if (currentConfigScopeId == lastConfigScopeId)
```
gabriela-trutan-sonarsource: I can imagine a scenario where this optimization is breaking the logic: when opening a solution and then a folder that has the same name as the solution, where one is a VCX project and the other is a CMake project. 

I think it's a risk we should not take. The only downside of removing the optimization is additional calls to SlCore, which should actually not happen that often.

gabriela-trutan-sonarsource (approved):
I have some feedback, but the implementation looks OK

src/Core/CFamily/IActiveCompilationDatabaseTracker.cs
georgii-borovinskikh-sonarsource: If we ever need to do something else with the compilation database file path, it's useful to have it in core. For example, if we want to show it in the ui at some point

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
georgii-borovinskikh-sonarsource: It can only realistically happen if the slcore is dead and it will be called again once it's restarted and re-initialized

georgii-borovinskikh-sonarsource: 🤔  but probably should reset the path to null then, so that re-initialization is done properly

src/Core/CFamily/IActiveCompilationDatabaseTracker.cs
gabriela-trutan-sonarsource: Shouldn't we move the interface in that point to Core and not before? At this point it's not needed.

georgii-borovinskikh-sonarsource: I don't think it hurts to put the interface in Core

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
georgii-borovinskikh-sonarsource: This should not happen. It should first transition to `null`, which will reset the cached id. The optimization is there because config scope event is triggered by every update to the config scope

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
georgii-borovinskikh-sonarsource: Actually, nevermind, the same argument about `null` works here, as when SLCore dies it transitions config scope to `null` via Reset method

src/Core/CFamily/IActiveCompilationDatabaseTracker.cs
gabriela-trutan-sonarsource: My only concern is the fact that `Core` is already quite big and it contains things that are not really needed there. If we ever decide the split it, it would be difficult to sort out what is really a "Core" logic.

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
gabriela-trutan-sonarsource: It would still be nice to log, in my opinion.

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
gabriela-trutan-sonarsource: You are right. :+1:

gabriela-trutan-sonarsource: I understood why you added the optimization. I think we do this in multiple places. In some point, it might make sense to introduce different events for the config scope (e.g. when the root path changes could be a different event)

georgii-borovinskikh-sonarsource: yes, I was thinking about that. Maybe as the first step I can add a flag to event args like `is miscellaneous update` and ignore those kinds of updates in this class

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
georgii-borovinskikh-sonarsource: IMO adds no value, if slcore is dead we already know about that

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
gabriela-trutan-sonarsource: You don't necessarily have to do it in this PR. Could be a task in the backlog as well. Either would be fine for me

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
georgii-borovinskikh-sonarsource: I kinda like the `OwnerName_EventName` format for event handlers, so I've updated it to be like that

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
```diff
@@ -0,0 +1,129 @@
...
+internal sealed class ActiveCompilationDatabaseTracker : IActiveCompilationDatabaseTracker
+{
+    private readonly ICMakeCompilationDatabaseLocator cMakeCompilationDatabaseLocator;
+    private readonly IActiveVcxCompilationDatabase activeVcxCompilationDatabase;
+    private readonly IActiveConfigScopeTracker activeConfigScopeTracker;
+    private readonly ISLCoreServiceProvider serviceProvider;
+    private readonly IThreadHandling threadHandling;
+    private readonly IAsyncLock asyncLock;
+    private bool isDisposed;
+    private string lastConfigScopeId;
```
gabriela-trutan-sonarsource: The field is not used anymore

src/Integration.Vsix.UnitTests/CFamily/ActiveCompilationDatabaseTrackerTests.cs
```diff
@@ -0,0 +1,387 @@
...
+        testSubject.DatabasePath.Should().Be(CmakeJson); // sanity check
+        VerifyOnBackgroundThreadLockTakenAndReleased(1);
+
+        const string scope2 = "scope2";
+        SetCurrentConfiguration(scope2, null, VcxJson);
+        RaiseEventNewConfigScope();
+
+        testSubject.DatabasePath.Should().Be(VcxJson);
+        VerifyCalledService(scope2, VcxJson);
+        VerifyOnBackgroundThreadLockTakenAndReleased(1 + 1);
```
gabriela-trutan-sonarsource, line 243: I think the `1+1` and `1+1+1` does not bring any better clarity then `2` or `3`. 

```suggestion
        VerifyOnBackgroundThreadLockTakenAndReleased(2);
```

src/Integration.Vsix.UnitTests/CFamily/ActiveCompilationDatabaseTrackerTests.cs
```diff
@@ -0,0 +1,387 @@
...
+        cFamilyAnalysisConfigurationSlCore.ClearReceivedCalls();
+        cMakeCompilationDatabaseLocator.ClearReceivedCalls();
+        activeVcxCompilationDatabase.ClearReceivedCalls();
+        activeConfigScopeTracker.CurrentConfigurationScopeChanged += Raise.EventWith<CurrentConfigurationScopeChangedEventArgs>(new(false));
+
+        testSubject.DatabasePath.Should().Be(CmakeJson);
+        cFamilyAnalysisConfigurationSlCore.DidNotReceiveWithAnyArgs().DidChangePathToCompileCommands(default);
+        cMakeCompilationDatabaseLocator.DidNotReceiveWithAnyArgs().Locate();
+        _ = activeVcxCompilationDatabase.DidNotReceiveWithAnyArgs().DatabasePath;
+        VerifyOnBackgroundThreadLockTakenAndReleased(1); // still one invocation, as the event is just ignored
```
gabriela-trutan-sonarsource, line 263: If you clear the received calls for the `threadHandling`, then you could check for  `DidNotReceiveWithAnyArgs`

src/Integration.Vsix.UnitTests/CFamily/ActiveCompilationDatabaseTrackerTests.cs
```diff
@@ -0,0 +1,387 @@
...
+        VerifyCalledService(scope2, VcxJson);
+        VerifyOnBackgroundThreadLockTakenAndReleased(1 + 1);
+    }
+
+    [TestMethod]
+    public void CurrentConfigurationScopeChanged_SameConfigurationScopeId_NotRedeclaredButUpdated_DoesNothing()
+    {
+        SetCurrentConfiguration(DefaultConfigScope,CmakeJson, null);
+        var testSubject = CreateAndInitializeTestSubject();
+        testSubject.DatabasePath.Should().Be(CmakeJson); // sanity check
```
gabriela-trutan-sonarsource, line 251: Do we need sanity checks in all the tests? In my opinion, they could be skipped.

src/Integration.Vsix.UnitTests/CFamily/ActiveCompilationDatabaseTrackerTests.cs
georgii-borovinskikh-sonarsource: The idea is to indicate that they are separate operations and not taken 2 times within one call

src/Integration.Vsix.UnitTests/CFamily/ActiveCompilationDatabaseTrackerTests.cs
georgii-borovinskikh-sonarsource: I would need to clear the lock too so the list just grows, I think it's ok as it i

src/Integration.Vsix.UnitTests/CFamily/ActiveCompilationDatabaseTrackerTests.cs
georgii-borovinskikh-sonarsource: I added single line checks to complicated tests so it's easier to know that it was not set up correctly

PR #6265, opened 2025-06-11
SLVS-2036 Inform SlCore about opened and closed files

gabriela-trutan-sonarsource:
[SLVS-2036](https://sonarsource.atlassian.net/browse/SLVS-2036)

[SLVS-2036]: https://sonarsource.atlassian.net/browse/SLVS-2036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
+++ b/src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
@@ -106,7 +106,7 @@ public sealed class DocumentEventsHandler : IDocumentEventsHandler
                 await vcxCompilationDatabaseUpdater.RemoveFileAsync(args.Document.FullPath);
             }
 
-            NotifySlCoreFileClosed(args.Document.FullPath);
+            NotifySlCoreFileClosed(args.Document.FullPath, activeConfigScopeTracker.Current);
         }).Forget();
 
     private void OnDocumentOpened(object sender, DocumentEventArgs args) =>
@@ -114,7 +114,7 @@ public sealed class DocumentEventsHandler : IDocumentEventsHandler
         {
             await AddFileToCompilationDatabaseAsync(args.Document);
 
-            NotifySlCoreFileOpened(args.Document.FullPath);
+            NotifySlCoreFileOpened(args.Document.FullPath, activeConfigScopeTracker.Current);
         }).Forget();
 
     /// <summary>
@@ -142,9 +142,8 @@ public sealed class DocumentEventsHandler : IDocumentEventsHandler
         }
     }
 
-    private void NotifySlCoreFileClosed(string filePath, ConfigurationScope configScope = null)
+    private void NotifySlCoreFileClosed(string filePath, ConfigurationScope currentConfigurationScope)
     {
-        var currentConfigurationScope = configScope ?? activeConfigScopeTracker.Current;
         if (VerifyConfigurationScopeInitialized(currentConfigurationScope))
         {
             GetFileRpcSlCoreService()?.DidCloseFile(new DidCloseFileParams(currentConfigurationScope.Id, new FileUri(filePath)));


src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
```diff
@@ -98,4 +126,38 @@ private void AddFileToCompilationDatabase(Document document)
             vcxCompilationDatabaseUpdater.AddFileAsync(document.FullPath).Forget();
         }
     }
+
+    private void NotifySlCoreFileOpened(string filePath)
+    {
+        if (activeConfigScopeTracker.Current?.Id is null)
+        {
+            logger.WriteLine(SLCoreStrings.ConfigScopeNotInitialized);
```
georgii-borovinskikh-sonarsource: How do we solve this problem architecturally? We need to register all open files even they were opened before config scope was set

gabriela-trutan-sonarsource: Good point. I guess we can subscribe to `CurrentConfigurationScopeChanged` event and inform SlCore about all opened files. We already exposed the list in the `IDocumentTracker`

georgii-borovinskikh-sonarsource: It needs to happen after ListFiles. Maybe we can put it into that listener

georgii-borovinskikh-sonarsource: Alternatively, listFiles is the thing responsible for setting the RootPath of ConfigurationScope, so the event can be used with a `configScope?.RootPath != null` check

gabriela-trutan-sonarsource: I don't think it has to. SlCore simply adds the list of opened files in its repository and then publishes the FileOpenedEvent (see [here](https://github.com/SonarSource/sonarlint-core/blob/65259f98f739ad36448a0a262e76eb8a85a1ada4/backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFileSystemService.java#L201)), which schedules an analysis.

georgii-borovinskikh-sonarsource: For visibility: https://sonarsource.slack.com/archives/C07QZKVQQ0L/p1749719344444859

src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
```diff
@@ -98,4 +126,38 @@ private void AddFileToCompilationDatabase(Document document)
...
+            return;
+        }
+        GetFileRpcSlCoreService()?.DidOpenFile(new DidOpenFileParams(activeConfigScopeTracker.Current.Id, new FileUri(filePath)));
+    }
+
+    private void NotifySlCoreFileClosed(string filePath)
+    {
+        if (activeConfigScopeTracker.Current?.Id is null)
```
georgii-borovinskikh-sonarsource: Following the discussion in the other thread, I think this needs to change to `is not { Id: {} currentId, RootPath: {} }`

src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
```diff
@@ -98,4 +126,38 @@ private void AddFileToCompilationDatabase(Document document)
...
+        {
+            logger.WriteLine(SLCoreStrings.ConfigScopeNotInitialized);
+            return;
+        }
+        GetFileRpcSlCoreService()?.DidCloseFile(new DidCloseFileParams(activeConfigScopeTracker.Current.Id, new FileUri(filePath)));
```
georgii-borovinskikh-sonarsource: `.Current` config scope needs to be cached into a local variable

georgii-borovinskikh-sonarsource (changes requested):
Small changes needed to how config scope is used + a ticket to register the open files that were opened before config scope was initialized

src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
```diff
@@ -73,6 +91,11 @@ private void OnOpenDocumentRenamed(object sender, DocumentRenamedEventArgs args)
         {
             vcxCompilationDatabaseUpdater.RenameFileAsync(args.OldFilePath, args.Document.FullPath).Forget();
         }
+        threadHandling.RunOnBackgroundThread(() =>
```
georgii-borovinskikh-sonarsource: One of the reasons I wanted to not do thread switching in `vcxCompilationDatabaseUpdater` and do it in the callers was because they will mort likely need to do that anyways. Here I would prefer if `vcxCompilationDatabaseUpdater` was also called inside `RunOnBackgroundThread` for consistency

src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
gabriela-trutan-sonarsource: It doesn't have to. If the configuration scope is not initialized and SlCore receives the event, it won't perform the analysis. So I don't see why would we check for the root in the closing files method.

georgii-borovinskikh-sonarsource: Won't it cause errors in logs? Opening an unknown (to SLCore) should result in something like that

src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
georgii-borovinskikh-sonarsource: The property can be updated concurrently, so it's better to work with the same instance of the ConfigurationScope in the scope of the method

src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
gabriela-trutan-sonarsource: Your comment is in the close file method.

gabriela-trutan-sonarsource: Which only [updates the repository](https://github.com/SonarSource/sonarlint-core/blob/65259f98f739ad36448a0a262e76eb8a85a1ada4/backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFileSystemService.java#L207) and not even re-triggers analysis.

src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
gabriela-trutan-sonarsource: [Here](https://sonarsource.atlassian.net/browse/SLVS-2305) is the sub-task.

src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
georgii-borovinskikh-sonarsource: Anyway, I don't think we should send rpc notifications when it can be avoided by slightly updating the pattern in the if

gabriela-trutan-sonarsource: I did the change, but this might actually be an effort that will be thrown out of the window as SlCore plans to drop the `ListFiles` listener at startup

src/Integration.Vsix/Analysis/DocumentEventsHandler.cs
```diff
@@ -85,66 +85,70 @@ public void Dispose()
...
+            GetFileRpcSlCoreService()?.DidOpenFile(new DidOpenFileParams(currentConfigurationScope.Id, new FileUri(filePath)));
         }
-        GetFileRpcSlCoreService()?.DidOpenFile(new DidOpenFileParams(activeConfigScopeTracker.Current.Id, new FileUri(filePath)));
     }
 
-    private void NotifySlCoreFileClosed(string filePath)
+    private void NotifySlCoreFileClosed(string filePath, ConfigurationScope configScope = null)
     {
-        if (activeConfigScopeTracker.Current?.Id is null)
+        var currentConfigurationScope = configScope ?? activeConfigScopeTracker.Current;
```
georgii-borovinskikh-sonarsource: I don't think it makes sense to have this as an optional parameter

georgii-borovinskikh-sonarsource (approved):

PR #6266, opened 2025-06-12
SLVS-2301 add IUserAnalysisPropertiesService & DidSetUserAnalysisProperties

georgii-borovinskikh-sonarsource:
[SLVS-2301](https://sonarsource.atlassian.net/browse/SLVS-2301)

[SLVS-2301]: https://sonarsource.atlassian.net/browse/SLVS-2301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SLCore/Service/Analysis/DidChangeAnalysisPropertiesParams.cs
+++ b/src/SLCore/Service/Analysis/DidChangeAnalysisPropertiesParams.cs
@@ -1,0 +1,23 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


src/SLCore/Service/Analysis/DidChangeAnalysisPropertiesParams.cs
```diff
@@ -0,0 +1,23 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarLint.VisualStudio.SLCore.Service.Analysis;
+
+public record DidChangeAnalysisPropertiesParams(string configurationScopeId, IReadOnlyDictionary<string, string> extraProperties);
```
vnaskos-sonar: I see that on the sonarlint-core `DidChangeAnalysisPropertiesParams` the field is called `properties` but on the constructor it's `extraProperties`.

Not sure which on should be used here. Probably `extraProperties` here is correct, because there is no setter apart from the constructor. Can you double check that?

vnaskos-sonar (approved):
👍

src/SLCore/Service/Analysis/DidChangeAnalysisPropertiesParams.cs
georgii-borovinskikh-sonarsource: https://sonarsource.slack.com/archives/C02E6L5C01H/p1749804365469989

PR #6267, opened 2025-06-12
SLVS-2050 Adapt `IAnalyzer` to match the new API points for SlCore automatic analysis.

gabriela-trutan-sonarsource:
[SLVS-2050](https://sonarsource.atlassian.net/browse/SLVS-2050)

[SLVS-2050]: https://sonarsource.atlassian.net/browse/SLVS-2050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Core/Analysis/AnalysisService.cs
```diff
@@ -52,7 +53,8 @@ public void ScheduleAnalysis(string filePath,
                 if (!token.IsCancellationRequested)
                 {
                     issueConsumerStorage.Set(filePath, analysisId, issueConsumer);
-                    analyzerController.ExecuteAnalysis(filePath, analysisId, detectedLanguages, analyzerOptions, token);
+                    // TODO by https://sonarsource.atlassian.net/browse/SLVS-2051 Adapt the call
+                    analyzerController.ExecuteAnalysis([filePath]);
```
georgii-borovinskikh-sonarsource, line 57: Do we even need this service at all anymore?

src/Core/Analysis/IAnalyzer.cs
```diff
@@ -22,10 +22,10 @@ namespace SonarLint.VisualStudio.Core.Analysis
...
-        void ExecuteAnalysis(string path,
-            Guid analysisId,
-            IEnumerable<AnalysisLanguage> detectedLanguages,
-            IAnalyzerOptions analyzerOptions,
-            CancellationToken cancellationToken);
+        Task<Guid?> ExecuteAnalysis(List<string> paths);
+
+        Task<Guid?> ExecuteAnalysisForOpenedFiles();
+
+        void CancelAnalysis(Guid analysisId);
```
georgii-borovinskikh-sonarsource, line 29: Just a question: where would we get this id from to call this method?

src/Core/Analysis/IAnalyzer.cs
```diff
@@ -22,10 +22,10 @@ namespace SonarLint.VisualStudio.Core.Analysis
...
     public interface IAnalyzer
     {
-        void ExecuteAnalysis(string path,
-            Guid analysisId,
-            IEnumerable<AnalysisLanguage> detectedLanguages,
-            IAnalyzerOptions analyzerOptions,
-            CancellationToken cancellationToken);
+        Task<Guid?> ExecuteAnalysis(List<string> paths);
+
+        Task<Guid?> ExecuteAnalysisForOpenedFiles();
```
georgii-borovinskikh-sonarsource: Opened -> Open

georgii-borovinskikh-sonarsource: Opened kinda includes all files that were ever open, even if they were closed

src/SLCore/Analysis/SLCoreAnalyzer.cs
```diff
@@ -41,77 +36,95 @@ public class SLCoreAnalyzer(
...
+            return false;
+        }
+        if (analysisSlCoreService != null)
+        {
+            return true;
+        }
+        if (serviceProvider.TryGetTransientService(out IAnalysisSLCoreService analysisService))
         {
-            EnrichPropertiesForCFamily(out var analysisProperties, path, detectedLanguages, analyzerOptions);
+            analysisSlCoreService = analysisService;
```
georgii-borovinskikh-sonarsource: The service is called transient for a reason: we should not cache it in the field because the instance of the service can change if the rpc connection changes

georgii-borovinskikh-sonarsource (changes requested):
Instances of ISLCoreService should not be cached. Other comments are minor.

src/Core/Analysis/AnalysisService.cs
gabriela-trutan-sonarsource: Do you mean the `AnalysisService`? For now, yes. There is bigger refactoring [task](https://sonarsource.atlassian.net/browse/SLVS-2051) for it, but I set it to minor, because the feature can be delivered without this. 

Or do you mean the `AnalyzerController`? The `AnalyzerController` will be dropped by [this task](https://sonarsource.atlassian.net/browse/SLVS-2047).

src/Core/Analysis/IAnalyzer.cs
gabriela-trutan-sonarsource: The new API [endpoints](https://github.com/SonarSource/sonarlint-visualstudio/blob/ed5041ebebb40deb148f557809e3fb9fd1102f6d/src/SLCore/Analysis/SLCoreAnalyzer.cs#L45) always return the analysis Id.

The idea in one of the bigger refactoring [tasks](https://sonarsource.atlassian.net/browse/SLVS-2051) is to return an `IAnalysisRequest` object that could store the received analysis id that the consumers can use to try to cancel a request

src/SLCore/Analysis/SLCoreAnalyzer.cs
gabriela-trutan-sonarsource: Thank you for catching this!

src/SLCore/Analysis/SLCoreAnalyzer.cs
```diff
@@ -65,7 +65,8 @@ public class SLCoreAnalyzer(
         analysisStatusNotifier.AnalysisStarted();
 
         var configurationScope = activeConfigScopeTracker.Current;
```
georgii-borovinskikh-sonarsource, line 67: could make the config scope an `out` parameter instead to not have this line twice

georgii-borovinskikh-sonarsource: out parameter of `VerifyConfigScopeInitialized`

gabriela-trutan-sonarsource: I try to avoid out parameters.

georgii-borovinskikh-sonarsource: I think it's a bit irrational to avoid it when it works great

PR #6280, opened 2025-06-17
SLVS-2049 Adapt AnalysisStatusNotifier and its factory

gabriela-trutan-sonarsource:
[SLVS-2049](https://sonarsource.atlassian.net/browse/SLVS-2049)

[SLVS-2049]: https://sonarsource.atlassian.net/browse/SLVS-2049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
```diff
@@ -18,86 +18,72 @@
...
 using System.IO;
 using SonarLint.VisualStudio.Core;
 using SonarLint.VisualStudio.Core.Analysis;
 using SonarLint.VisualStudio.Integration.Vsix.Helpers;
 
 namespace SonarLint.VisualStudio.Integration.Vsix.Analysis
 {
-    internal class AnalysisStatusNotifier : IAnalysisStatusNotifier
+    internal class AnalysisStatusNotifier(
+        string analyzerName,
```
georgii-borovinskikh-sonarsource: Do we still need the `analyzerName` parameter?

src/Integration.Vsix/Analysis/AnalysisStrings.resx
```diff
@@ -158,19 +158,19 @@
     <value>SonarQube for Visual Studio: Analyzing {0}</value>
   </data>
   <data name="MSG_AnalysisComplete" xml:space="preserve">
-    <value>Finished analyzing {0} with id {1}, analysis time: {2}s</value>
+    <value>Analysis {1} finished for {0}, analysis time: {2}s</value>
```
georgii-borovinskikh-sonarsource: If id is null then this will look strange with the double space
Also the reversed order of placeholders is confusing

src/Integration.Vsix/Analysis/AnalysisStrings.resx
```diff
@@ -158,19 +158,19 @@
...
-    <value>Failed to analyze {0} with id {1}: {2}</value>
+    <value>Analysis {0} failed for files {1}. Reason: {2}</value>
   </data>
   <data name="MSG_AnalysisStarted" xml:space="preserve">
-    <value>Analyzing {0} with id {1}</value>
+    <value>Analyzing {0}</value>
   </data>
   <data name="MSG_FoundIssues" xml:space="preserve">
-    <value>Found {0} {1}(s) in {2} [id: {3}, final: {4}]</value>
+    <value>Analysis {3} found {0} {1}(s) in {2} [final: {4}]</value>
```
georgii-borovinskikh-sonarsource: 3 0 1 2 4

gabriela-trutan-sonarsource: It was easier to change. :D

src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
gabriela-trutan-sonarsource: It's a way to provide context to the logs that the status notifier is creating

src/Integration.Vsix/Analysis/AnalysisStrings.resx
gabriela-trutan-sonarsource: I think it's flexible enough to have a readable error message also for the case when analysisId is null

gabriela-trutan-sonarsource: The reverse order was because the change was easier.

src/Integration.Vsix/Analysis/AnalysisStatusNotifier.cs
georgii-borovinskikh-sonarsource: We can keep the context, sure, but we don't need the parameter - it can be an internal constant

gabriela-trutan-sonarsource: I removed it and used "Analysis" as context as the `AnalysisStatusNotifier` can be called from other parts than `SlCoreAnalyzer`.
Additionally, I find confusing that our logs that really come from SlCore and have the context `[SlCore]` And then we have `[SlCoreAnalyzer]` context, which is coming from the client side

PR #6282, opened 2025-06-18
SLVS-2053 Drop the `analyzeFilesAndTrack`

gabriela-trutan-sonarsource:
[SLVS-2053](https://sonarsource.atlassian.net/browse/SLVS-2053)

[SLVS-2053]: https://sonarsource.atlassian.net/browse/SLVS-2053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
```diff
@@ -39,6 +38,8 @@ namespace SonarLint.VisualStudio.SLCore.IntegrationTests;
 
 internal sealed class FileAnalysisTestsRunner : IDisposable
 {
+    private const double AnalysisReadinessWaitTimeout = 60;
```
georgii-borovinskikh-sonarsource: specify the unit of measurement in the variable name or, even better, change it to TimeSpan

src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
```diff
@@ -124,15 +125,12 @@ public async Task<Dictionary<FileUri, List<RaisedIssueDto>>> RunFileAnalysis(
...
         string configScope,
-        bool sendContent = false,
-        Dictionary<string, string> extraProperties = null)
+        bool sendContent = false)
     {
         try
         {
-            var (analysisId, analysisRaisedIssues) = await SetUpAnalysis(testingFile, configScope, sendContent);
-
-            var rawIssueDtos = await RunSlCoreFileAnalysis(configScope, testingFile.GetFullPath(), analysisId, extraProperties);
```
georgii-borovinskikh-sonarsource, line 134: The old analysis method used to complete the await only after the analysis had been finished. Does the new method behave the same? If not, aren't we making the tests that use this method flaky?

gabriela-trutan-sonarsource: The old analysis used to complete the await after the analysis finished, but did not await for the `RaiseIssues` to complete. In fact, it was expecting that `RaiseIssues` was never invoked.

We are also awaiting for `AnalyzeFileListAsync` to complete.

georgii-borovinskikh-sonarsource: That's why I had the question about the behavior of the new method. The old method used to await the analysis completion and return the list of failed files: the test method then asserted that the analysis method has returned, the list of files with errors was empty and no issues were raised => the file was excluded from analysis. Does the new method guarantee the same? What if the exclusions are broken, but we don't wait long enough for the RaiseIssue to be sent and the test passes?

gabriela-trutan-sonarsource: > the list of files with errors was empty

I don't see that being checked.

> no issues were raised

The problem is that it was not waiting for the issues being raised. What if they might have been raised at a later point?

> What if the exclusions are broken, but we don't wait long enough for the RaiseIssue to be sent and the test passes?

You have a point. We can reuse the same normal method that is used for the analysis, which has a timeout of 60 seconds. The problem here is that this will increase the waiting time by 60 seconds for each test, because we don't expect the `RaiseIssues` to be invoked neither. So for each new integration test we add, the time will increase again

georgii-borovinskikh-sonarsource:
@gabriela-trutan-sonarsource new feature branch, rebase please

src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
gabriela-trutan-sonarsource: I instead awaited the call for file exclusions listener and added a small timeout to make sure `raiseIssues` was not invoked (before we were not waiting at all for RaiseIssues to be called)

src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
```diff
@@ -141,22 +154,34 @@ public async Task VerifyAnalysisSkipped(
...
+    }
+
+    private void SetupCompilationDatabase(string configScope, string compilationDatabasePath)
+    {
+        if (compilationDatabasePath is null)
+        {
+            return;
+        }
+        slCoreTestRunner.SLCoreServiceProvider.TryGetTransientService(out ICFamilyAnalysisConfigurationSLCoreService slCoreService).Should().BeTrue();
+        slCoreService.DidChangePathToCompileCommands(new DidChangePathToCompileCommandsParams(configScope, compilationDatabasePath));
```
georgii-borovinskikh-sonarsource, line 184: This is definitely better than it was originally, however I'm still not a fan of the 'short time out to verify it's not called' because it may be too short in some scenarios. Is there no way to receive an event from SLCore that analysis processing has been completed? For example, via reportProgress methods?

georgii-borovinskikh-sonarsource (approved):

src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
```diff
@@ -121,18 +126,26 @@ public async Task<Dictionary<FileUri, List<RaisedIssueDto>>> RunFileAnalysis(
...
+            {
+                var fileExclusionsParams = callInfo.Arg<GetFileExclusionsParams>();
+                if (fileExclusionsParams.configurationScopeId == configScope)
+                {
+                    fileExclusionListenerCompletionSource.SetResult(1);
+                }
+            });
+            NotifyDidOpenFile(configScope, testingFile.GetFullPath());
+            await ConcurrencyTestHelper.WaitForTaskWithTimeout(fileExclusionListenerCompletionSource.Task, "file exclusions listener");
+            await Task.WhenAny(analysisRaisedIssues.Task, Task.Delay(TimeSpan.FromSeconds(2))); // wait for a short time to see if any issues are raised
```
georgii-borovinskikh-sonarsource, line 148: This is definitely better than it was originally, however I'm still not a fan of the 'short time out to verify it's not called' because it may be too short in some scenarios. Is there no way to receive an event from SLCore that analysis processing has been completed? For example, via reportProgress methods?

gabriela-trutan-sonarsource: > For example, via reportProgress methods?

I can check. But I feel this would be a hack rather than an improvement

gabriela-trutan-sonarsource: Could actually work. But as this depends on a work in progress [task](https://sonarsource.atlassian.net/browse/SLVS-2054), I updated the follow up [task](https://sonarsource.atlassian.net/browse/SLVS-2334) for the integration tests to implement this suggestion

PR #6283, opened 2025-06-18
SLVS-2045 Remove no longer needed triggering of manual analysis

gabriela-trutan-sonarsource:
[SLVS-2045](https://sonarsource.atlassian.net/browse/SLVS-2045)

[SLVS-2045]: https://sonarsource.atlassian.net/browse/SLVS-2045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SLCore.Listeners/Implementation/Analysis/AnalysisListener.cs
+++ b/src/SLCore.Listeners/Implementation/Analysis/AnalysisListener.cs
@@ -33,7 +33,6 @@ namespace SonarLint.VisualStudio.SLCore.Listeners.Implementation.Analysis;
 [method: ImportingConstructor]
 internal class AnalysisListener(
     IActiveConfigScopeTracker activeConfigScopeTracker,
-    IAnalysisRequester analysisRequester,
     IRaisedFindingProcessor raisedFindingProcessor,
     IIssuePublisher issuePublisher,
     IHotspotPublisher hotspotPublisher,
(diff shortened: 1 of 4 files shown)


src/SLCore.Listeners/Implementation/Analysis/AnalysisListener.cs
```diff
@@ -47,10 +46,6 @@ public void DidChangeAnalysisReadiness(DidChangeAnalysisReadinessParams paramete
         if (activeConfigScopeTracker.TryUpdateAnalysisReadinessOnCurrentConfigScope(configScopeId, parameters.areReadyForAnalysis))
         {
             logger.WriteLine(SLCoreStrings.AnalysisReadinessUpdate, parameters.areReadyForAnalysis);
-            if (parameters.areReadyForAnalysis)
```
gabriela-trutan-sonarsource, line 50: With the removal of this manual analysis request, the [bug](https://sonarsource.atlassian.net/browse/SLCORE-907) for the BOM files is now visible once you start VS with a file that was previously opened ( it will be visible after [this task is done](https://sonarsource.atlassian.net/browse/SLVS-2305), because in the current branch the SlCore is not even aware that the file is opened). That is because we no longer inform SlCore of the content of the opened files, which this manual analysis was ensuring.

We can consider reverting this change and add a comment to remove it once the bug on SlCore is fixed. Or we can accept the drawback (the issue will be fixed on the next analysis, e.g. reopening the file, on save etc)

image

PR #6287, opened 2025-06-20
SLVS-2324 QuickFix can be split into multiple InputFileEdits

vnaskos-sonar:


src/SLCore.UnitTests/Listener/Analysis/RaiseFindingToAnalysisIssueConverterTests.cs
```diff
@@ -356,6 +356,35 @@ public void GetAnalysisIssues_TextRangeDtoIsNull_ConvertsCorrectly()
...
+            "ruleKey",
+            "PrimaryMessage",
+            DateTimeOffset.Now,
+            true,
+            false,
+            new TextRangeDto(1, 2, 3, 4),
+            [],
+            [
+                new QuickFixDto([
+                    new FileEditDto(new FileUri("C:\\IssueFile.cs"), [new TextEditDto(new TextRangeDto(5, 6, 7, 8), "new text")]),
```
gabriela-trutan-sonarsource, line 373: This can be extracted into a variable

gabriela-trutan-sonarsource (approved):
Looks good!

src/SLCore.UnitTests/Listener/Analysis/RaiseFindingToAnalysisIssueConverterTests.cs
vnaskos-sonar: On one hand this is quite extensive and bloats the test with many setup lines. On the other hand it's really specific to only this exact test and it's visible that it contains 3 `FileEditDto`s and one of them should be skipped (because it refers to another file). I am leaning towards keeping it as is 🤔

gabriela-trutan-sonarsource: I only meant that the path of the `FileEditDto` that is common should be extracted as a variable. So

`var currentFile ="C:\\IssueFile.cs";`

It's also fine to keep it as it is

PR #6289, opened 2025-06-20
Fix: SLCoreServiceProvider returning true before backend is initialized

georgii-borovinskikh-sonarsource:
Part of SLVS-1986

### DIFF
--- a/src/SLCore.UnitTests/Core/SLCoreServiceProviderTests.cs
+++ b/src/SLCore.UnitTests/Core/SLCoreServiceProviderTests.cs
@@ -1,237 +1,237 @@
-﻿/*
- * SonarLint for Visual Studio
- * Copyright (C) 2016-2025 SonarSource SA
- * mailto:info AT sonarsource DOT com
- *
- * This program is free software; you can redistribute it and/or
- * modify it under the terms of the GNU Lesser General Public
- * License as published by the Free Software Foundation; either
- * version 3 of the License, or (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
- * Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public License
- * along with this program; if not, write to the Free Software Foundation,
- * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
- */
-
-using SonarLint.VisualStudio.Core;
-using SonarLint.VisualStudio.SLCore.Core;
-
-namespace SonarLint.VisualStudio.SLCore.UnitTests.Core;
-
-[TestClass]
-public class SLCoreServiceProviderTests
-{
-    [TestMethod]
-    public void MefCtor_CheckIsExported()
-    {
-        MefTestHelpers.CheckTypeCanBeImported<SLCoreServiceProvider, ISLCoreServiceProvider>(
-            MefTestHelpers.CreateExport<IThreadHandling>(),
-            MefTestHelpers.CreateExport<ILogger>());
-    }
-
-    [TestMethod]
-    public void MefCtor_WriterInterface_CheckIsExported()
-    {
-        MefTestHelpers.CheckTypeCanBeImported<SLCoreServiceProvider, ISLCoreServiceProviderWriter>(
-            MefTestHelpers.CreateExport<IThreadHandling>(),
-            MefTestHelpers.CreateExport<ILogger>());
-    }
-
-    [TestMethod]
-    public void Mef_CheckIsSingleton()
-    {
-        MefTestHelpers.CheckIsSingletonMefComponent<SLCoreServiceProvider>();
-    }
-
-    [TestMethod]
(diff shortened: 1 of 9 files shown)


src/SLCore.UnitTests/Core/SLCoreServiceProviderTests.cs
```diff
@@ -197,26 +203,213 @@ public void SetCurrentConnection_ClearsAllCachedServices()
...
+        SetConnectionAndInitialize(rpcMock, false);
+
+        var result = testSubject.TryGetTransientService(out ITestSLcoreService1 service);
+
+        result.Should().BeFalse();
+        service.Should().BeNull();
+    }
+
+    [TestMethod]
+    public void TryGetTransientService_WithoutInitialize_AndConnectionNotAlive_ReturnsFalse()
```
gabriela-trutan-sonarsource, line 232: This test and `TryGetTransientService_WithoutInitialize_AndConnectionAlive_ReturnsFalse` can be converted into one with `isAlive` as a parameter

src/SLCore/Core/ISLCoreServiceProvider.cs
```diff
@@ -44,69 +55,123 @@ internal interface ISLCoreServiceProviderWriter : ISLCoreServiceProvider
...
             }
+
+            service = (TService)cachedService;
+            return true;
         }
 
         public void SetCurrentConnection(ISLCoreJsonRpc newRpcInstance)
         {
-            lock (cacheLock)
+            threadHandling.ThrowIfOnUIThread();
```
gabriela-trutan-sonarsource: I don't see why this check is needed. If we are on the UI thread there are no side effects.

src/SLCore/Core/ISLCoreServiceProvider.cs
```diff
@@ -31,8 +33,17 @@ public interface ISLCoreServiceProvider
...
-        bool TryGetTransientService<TService>(out TService service) where TService : class, ISLCoreService;
+        /// <returns>True if the underlying connection is alive, False if the connection is unavailable at the moment or the backend hasn't been initialized</returns>
+        bool TryGetTransientService<TService>([NotNullWhen(returnValue: true)]out TService? service) where TService : class, ISLCoreService;
+    }
+
+    public interface ISLCoreRpcManager
+    {
+        void Initialize(InitializeParams parameters);
+
+        bool IsInitialized { get; }
```
gabriela-trutan-sonarsource: This doesn't have to be exposed. Nobody seems to be using it

src/SLCore/Core/ISLCoreServiceProvider.cs
```diff
@@ -44,69 +55,123 @@ internal interface ISLCoreServiceProviderWriter : ISLCoreServiceProvider
...
+                    return false;
+                }
+                cache.Add(serviceType, cachedService);
             }
+
+            service = (TService)cachedService;
+            return true;
         }
 
         public void SetCurrentConnection(ISLCoreJsonRpc newRpcInstance)
```
gabriela-trutan-sonarsource: I find the current name unintuitive

```suggestion
        public void SetCurrentRpcInstance(ISLCoreJsonRpc newRpcInstance)
```

src/SLCore/ISLCoreInstanceHandle.cs
```diff
@@ -99,15 +99,12 @@ public void Initialize()
 
         SLCoreRpc = slCoreRpcFactory.StartNewRpcInstance();
 
-        if (!serviceProvider.TryGetTransientService(out ILifecycleManagementSLCoreService lifecycleManagementSlCoreService))
-        {
-            throw new InvalidOperationException(SLCoreStrings.ServiceProviderNotInitialized);
-        }
+
```
gabriela-trutan-sonarsource, line 102: Please remove not needed empty lines

src/SLCore.UnitTests/Core/SLCoreServiceProviderTests.cs
```diff
@@ -18,41 +18,59 @@
...
+    public void MefCtor_CheckIsExported() =>
         MefTestHelpers.CheckTypeCanBeImported<SLCoreServiceProvider, ISLCoreServiceProvider>(
             MefTestHelpers.CreateExport<IThreadHandling>(),
             MefTestHelpers.CreateExport<ILogger>());
-    }
 
     [TestMethod]
-    public void MefCtor_WriterInterface_CheckIsExported()
-    {
+    public void MefCtor_LifecycleManagerInterface_CheckIsExported() =>
```
gabriela-trutan-sonarsource:
```suggestion
    public void MefCtor_SLCoreRpcManagerInterface_CheckIsExported() =>
```

gabriela-trutan-sonarsource (approved):
Looks OK, except for some suggestions

src/SLCore/Core/ISLCoreServiceProvider.cs
georgii-borovinskikh-sonarsource: I think it's fine to keep it, doesn't hurt

src/SLCore/Core/ISLCoreServiceProvider.cs
georgii-borovinskikh-sonarsource: The side effect is the lock, so it's better to not do it (and anyway it is only called on the bg thread)

PR #6290, opened 2025-06-20
Map new C++ LanguageStandard ('c++23preview' and 'c++23')

guillem-bartrina-sonarsource:
Part of 
CPP-6599

### DIFF
--- a/src/Integration.Vsix.UnitTests/CFamily/VcxProject/LanguageStandardConverterTests.cs
+++ b/src/Integration.Vsix.UnitTests/CFamily/VcxProject/LanguageStandardConverterTests.cs
@@ -31,6 +31,8 @@ public class LanguageStandardConverterTests
     [DataRow("Default", "")]
     [DataRow(null, "")]
     [DataRow("stdcpplatest", "/std:c++latest")]
+    [DataRow("stdcpp23", "/std:c++23")]
+    [DataRow("stdcpp23preview", "/std:c++23preview")]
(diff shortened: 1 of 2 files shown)


edward-gonzales-sonarsource:
@guillem-bartrina-sonarsource @mpaladin -- what are next steps for this PR?

guillem-bartrina-sonarsource:
Hello, @edward-gonzales-sonarsource, it should be merged as soon as possible, but some local validation is still needed. @mpaladin has already conveyed these same concerns to me. I will try to finalize it this week without fail.

guillem-bartrina-sonarsource:
I have tested this locally on my Windows VM.

When setting std c++23preview in the user interface, we currently get the value _stdcpp23_ (and not _stdcpp23preview_) for `LanguageStandard`. However, this is not documented anywhere. I would keep the mapping of the value _stdcpp23preview_ just in case.

michael-jabbour-sonarsource (approved):
LGTM!

@guillem-bartrina-sonarsource Could you create an SLVS ticket for this change with a title like "Add support for the 'c++23preview' flag", to make the impact clear for users in the release notes?

The current CPP ticket can then be linked or closed as duplicate IMO. WDYT?

guillem-bartrina-sonarsource:
> Could you create an SLVS ticket for this change with a title like "Add support for the 'c++23preview' flag", to make the impact clear for users in the release notes?

Here you go: SLVS-2463

PR #6298, opened 2025-06-25
SLVS-2326 Don't send file content for flushed files to SLCore

georgii-borovinskikh-sonarsource:
[SLVS-2326](https://sonarsource.atlassian.net/browse/SLVS-2326)

[SLVS-2326]: https://sonarsource.atlassian.net/browse/SLVS-2326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/DocumentEvents.cs
+++ b/src/Core/DocumentEvents.cs
@@ -27,15 +27,9 @@ public class DocumentEventArgs(Document document) : EventArgs
     public Document Document { get; } = document;
 }
 
-public class DocumentSavedEventArgs(Document document, string newContent) : DocumentEventArgs(document)
-{
-    public string NewContent { get; } = newContent;
-}
+public class DocumentSavedEventArgs(Document document) : DocumentEventArgs(document);
 
-public class DocumentOpenedEventArgs(Document document, string content) : DocumentEventArgs(document)
-{
-    public string Content { get; } = content;
-}
+public class DocumentOpenedEventArgs(Document document) : DocumentEventArgs(document);
 
 public class DocumentRenamedEventArgs(Document document, string oldFilePath) : DocumentEventArgs(document)
 {
--- a/src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
+++ b/src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
@@ -129,7 +129,7 @@ internal sealed class TaggerProvider : ITaggerProvider, IDocumentTracker
                 .Select<IIssueTracker, Action>(it => it.UpdateAnalysisState)
                 .ToList(); // create a fixed list - the user could close a file before the reanalysis completes which would cause the enumeration to change
             var documentsToAnalyzeCount = operations.Count;
-            operations.Add(() => NotifyFileTracker(filteredIssueTrackers));
+            operations.Add(() => NotifyFileTrackerOpenFilesCurrentState(filteredIssueTrackers));
             operations.Add(() => ExecuteAnalysisAsync(filteredIssueTrackers).Forget());
 
             reanalysisProgressHandler = new StatusBarReanalysisProgressHandler(vsStatusBar, logger);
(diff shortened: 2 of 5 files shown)


src/Core/DocumentEvents.cs
```diff
@@ -27,15 +27,9 @@ public class DocumentEventArgs(Document document) : EventArgs
     public Document Document { get; } = document;
 }
 
-public class DocumentSavedEventArgs(Document document, string newContent) : DocumentEventArgs(document)
-{
-    public string NewContent { get; } = newContent;
-}
+public class DocumentSavedEventArgs(Document document) : DocumentEventArgs(document);
```
gabriela-trutan-sonarsource, line 30: As now the args have no special logic, they can just be removed and the `DocumentEventArgs` can be used instead.

src/Core/DocumentEvents.cs
```diff
@@ -27,15 +27,9 @@ public class DocumentEventArgs(Document document) : EventArgs
...
 
-public class DocumentOpenedEventArgs(Document document, string content) : DocumentEventArgs(document)
-{
-    public string Content { get; } = content;
-}
+public class DocumentOpenedEventArgs(Document document) : DocumentEventArgs(document);
```
gabriela-trutan-sonarsource, line 32: As now the args have no special logic, they can just be removed and the `DocumentEventArgs` can be used instead.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
```diff
@@ -154,15 +154,16 @@ private async Task ExecuteAnalysisAsync(IEnumerable<IIssueTracker> filteredIssue
...
     }
 
-    private void NotifyFileTracker(string filePath, string content) => fileTracker.AddFiles(CreateSourceFile(filePath, content));
+    private void NotifyFileTrackerFileChanged(string filePath) => fileTracker.AddFiles(CreateSourceFile(filePath));
 
-    private void NotifyFileTracker(IEnumerable<IIssueTracker> filteredIssueTrackers)
+    private void NotifyFileTrackerOpenFilesCurrentState(IEnumerable<IIssueTracker> filteredIssueTrackers)
     {
-        var sourceFilesToUpdate = filteredIssueTrackers.Select(it => CreateSourceFile(it.LastAnalysisFilePath, it.GetText()));
+        // still sending the content here because the files may be dirty in this case
```
gabriela-trutan-sonarsource, line 161: The comment should go on line 132. The method itself just notifies file tracker for a list of changed files. It is not aware of why it is being used

gabriela-trutan-sonarsource (approved):
Minor feedback

src/Core/DocumentEvents.cs
georgii-borovinskikh-sonarsource: I like it more that they're separate, it's easier to modify if we need to in the future

src/Core/DocumentEvents.cs
georgii-borovinskikh-sonarsource: I like it more that they're separate, it's easier to modify if we need to in the future

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
georgii-borovinskikh-sonarsource: I renamed the method to reflect that it's intended to send the Current State of the document, hence the comment should go inside the method to explain what the current stat is in this case

src/Core/DocumentEvents.cs
gabriela-trutan-sonarsource: We don't need it now, so it would be cleaner to remove it :)

gabriela-trutan-sonarsource: It's also not difficult to create in the future if it will ever be needed.

src/Integration.Vsix/SonarLintTagger/TaggerProvider.cs
gabriela-trutan-sonarsource: You added the comment on the line 162, which doesn't bring any clarity of the content of the method. We call `fileTracker.AddFiles` in other methods, what is the distinction between those methods and this one?

The intention of why calling something should go to the place where it's being called

georgii-borovinskikh-sonarsource: The distinction is explained in the comment and the method name. The intention is obvious from the context of RequestAnalysis (`NotifyFileTrackerOpenFilesCurrentState` vs `NotifyFileTrackerFileChanged`)

gabriela-trutan-sonarsource: But the content of those methods are identical (just that one of them adds a file and the other one a list of files). The intention comes from the caller, not from the content of the method itself.

gabriela-trutan-sonarsource: The method can be reused in other places with a different intention and reason. It's just a coincidence that is used just a single time now

georgii-borovinskikh-sonarsource: No, the content is not identical. One of them adds files without content, and one with content. The context for that was added to the method names

gabriela-trutan-sonarsource: I missed that. 🫨

PR #6299, opened 2025-06-25
SLVS-2054 Implement “startProgress” and “reportProgress”

gabriela-trutan-sonarsource:
[SLVS-2054](https://sonarsource.atlassian.net/browse/SLVS-2054)

[SLVS-2054]: https://sonarsource.atlassian.net/browse/SLVS-2054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SLCore/Protocol/EitherJsonConverter.cs
+++ b/src/SLCore/Protocol/EitherJsonConverter.cs
@@ -105,12 +105,12 @@ namespace SonarLint.VisualStudio.SLCore.Protocol
             {
                 if (leftProperties.Contains(jsonProperty))
                 {
-                    return Either<TLeft, TRight>.CreateLeft(jToken.ToObject<TLeft>());
+                    return Either<TLeft, TRight>.CreateLeft(jToken.ToObject<TLeft>()!);
                 }
 
                 if (rightProperties.Contains(jsonProperty))
                 {
-                    return Either<TLeft, TRight>.CreateRight(jToken.ToObject<TRight>());
+                    return Either<TLeft, TRight>.CreateRight(jToken.ToObject<TRight>()!);
                 }
             }
             return null;
@@ -122,11 +122,11 @@ namespace SonarLint.VisualStudio.SLCore.Protocol
             {
                 if (!leftProperties.Any())
                 {
-                    return Either<TLeft, TRight>.CreateLeft(jToken.ToObject<TLeft>());
+                    return Either<TLeft, TRight>.CreateLeft(jToken.ToObject<TLeft>()!);
                 }
                 if (!rightProperties.Any())
                 {
-                    return Either<TLeft, TRight>.CreateRight(jToken.ToObject<TRight>());
+                    return Either<TLeft, TRight>.CreateRight(jToken.ToObject<TRight>()!);
                 }
             }
             return null;
--- a/src/SLCore/Protocol/Either.cs
+++ b/src/SLCore/Protocol/Either.cs
@@ -32,9 +32,9 @@ namespace SonarLint.VisualStudio.SLCore.Protocol
         public TLeft? Left { get; private init; }
         public TRight? Right { get; private init; }
 
-        public static Either<TLeft, TRight> CreateLeft(TLeft? left) => new() { Left = left ?? throw new ArgumentNullException(nameof(left)) };
+        public static Either<TLeft, TRight> CreateLeft(TLeft left) => new() { Left = left ?? throw new ArgumentNullException(nameof(left)) };
 
-        public static Either<TLeft, TRight> CreateRight(TRight? right) => new() { Right = right ?? throw new ArgumentNullException(nameof(right)) };
+        public static Either<TLeft, TRight> CreateRight(TRight right) => new() { Right = right ?? throw new ArgumentNullException(nameof(right)) };
 
         public static implicit operator Either<TLeft, TRight>(TLeft left) => CreateLeft(left);
 


src/SLCore/Protocol/EitherJsonConverter.cs
```diff
@@ -110,6 +112,18 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
                 }
             }
 
+            if (!jsonProperties.Any())
```
georgii-borovinskikh-sonarsource, line 121: it would be easier to read if conditions were inverted

georgii-borovinskikh-sonarsource: This block can also be put before the foreach

src/SLCore.UnitTests/Protocol/EitherJsonConverterTests.cs
```diff
@@ -44,34 +44,18 @@ public void SerializeObject_SerializesEitherAsSingleObject()
...
         JsonConvert.SerializeObject(left).Should().BeEquivalentTo("""{"Property":{"Left":"lll"}}""");
         JsonConvert.SerializeObject(right).Should().BeEquivalentTo("""{"Property":{"Right":10}}""");
     }
 
-    [TestMethod]
-    public void DeserializeObject_PrimitiveNotAnObject_Throws()
-    {
-        var str = """
-                  {
-                    "Property" : "ThisIsExpectedToBeAnObjectButItIsAString"
```
georgii-borovinskikh-sonarsource: Why was this removed? It's still true and it still passes

src/SLCore.UnitTests/Protocol/EitherJsonConverterTests.cs
```diff
@@ -201,4 +248,34 @@ public class RightOption
...
     }
+
+    public class ObjectWithRightOptionNoProperties
+    {
+        [JsonConverter(typeof(EitherJsonConverter<LeftOption, RightOption>))]
+        public Either<LeftOption, RightOption> Property { get; set; }
+
+        public class LeftOption
+        {
+            public string Left { get; set; }
```
georgii-borovinskikh-sonarsource: Use LeftProperty instead of Left (same for Right) as it is clearer that it refers to a property of the Left type not the Left type itself. Update SimpleObject class as well

src/SLCore.UnitTests/Protocol/EitherJsonConverterTests.cs
```diff
@@ -126,13 +109,78 @@ public void DeserializeObject_UnresolvableTypes_Throws()
                 "Types SonarLint.VisualStudio.SLCore.UnitTests.Protocol.EitherJsonConverterTests+ConflictingObject+ConflictingLeft and SonarLint.VisualStudio.SLCore.UnitTests.Protocol.EitherJsonConverterTests+ConflictingObject+ConflictingRight have equivalent sets of properties and fields");
     }
 
+    [TestMethod]
+    public void DeserializeObject_ObjectWithRightOptionNoProperties_Right_ChoosesCorrectSide()
```
georgii-borovinskikh-sonarsource, line 163: need a test where both left and right are empty objects

georgii-borovinskikh-sonarsource: and another one where they're not empty, but the values for them are omitted in json (for example, if NullValueHandling is set to Ignore)

src/SLCore.Listeners/Implementation/ProgressListener.cs
```diff
@@ -19,32 +19,37 @@
...
-        public Task ReportProgressAsync(object parameters)
+        /// <inheritdoc />
+        public void ReportProgress(ReportProgressParams parameters)
         {
-            return Task.CompletedTask;
+            if (parameters.notification.Left is { } updateNotification)
+            {
+                slCoreStatusNotifier.Notify(FormatNotification(updateNotification.message), false);
+            }
+            else if (parameters.notification.Right != null)
```
georgii-borovinskikh-sonarsource, line 47: Rider complains that this code is unreachable because Left is not nullable. We need to adapt the EIther class like this

```
public TLeft? Left { get; private init; }
public TRight? Right { get; private init; }
```

src/SLCore.Listeners/Implementation/ProgressListener.cs
```diff
@@ -19,32 +19,37 @@
...
-        /// <summary>
-        /// Stub method for compability with SLCore. We do not support progress
-        /// </summary>
-        /// <param name="parameters">Parameter's here for compability we discard it</param>
-        public Task ReportProgressAsync(object parameters)
+        /// <inheritdoc />
+        public void ReportProgress(ReportProgressParams parameters)
         {
-            return Task.CompletedTask;
+            if (parameters.notification.Left is { } updateNotification)
```
georgii-borovinskikh-sonarsource: Personal preference: `parameters.notification is {Left: {} updateNotification}`

src/SLCore.Listeners/Implementation/ProgressListener.cs
```diff
@@ -19,32 +19,37 @@
...
 using SonarLint.VisualStudio.SLCore.Core;
 using SonarLint.VisualStudio.SLCore.Listener.Progress;
 
 namespace SonarLint.VisualStudio.SLCore.Listeners.Implementation
 {
     [Export(typeof(ISLCoreListener))]
     [PartCreationPolicy(CreationPolicy.Shared)]
-    public class ProgressListener : IProgressListener
+    [method: ImportingConstructor]
+    public class ProgressListener(IStatusBarNotifier slCoreStatusNotifier) : IProgressListener
```
georgii-borovinskikh-sonarsource: why is it called `slCoreStatusNotifier`?

src/SLCore.Listeners/Implementation/ProgressListener.cs
```diff
@@ -19,32 +19,37 @@
...
+            {
+                slCoreStatusNotifier.Notify(FormatNotification(updateNotification.message), false);
+            }
+            else if (parameters.notification.Right != null)
+            {
+                slCoreStatusNotifier.Notify(string.Empty, false);
+            }
         }
+
+        private string FormatNotification(string message) => $"{SLCoreStrings.LongProductName}: {message}";
```
georgii-borovinskikh-sonarsource: Does the `LongProductName` fit well into the task bar?

src/SLCore.Listeners/Implementation/ProgressListener.cs
```diff
@@ -19,32 +19,37 @@
...
-        /// </summary>
-        /// <param name="parameters">Parameter's here for compability we discard it</param>
-        public Task ReportProgressAsync(object parameters)
+        /// <inheritdoc />
+        public void ReportProgress(ReportProgressParams parameters)
         {
-            return Task.CompletedTask;
+            if (parameters.notification.Left is { } updateNotification)
+            {
+                slCoreStatusNotifier.Notify(FormatNotification(updateNotification.message), false);
```
georgii-borovinskikh-sonarsource: Have you looked into this? https://sonarsource.atlassian.net/browse/SLVS-2054?focusedCommentId=800661 Seems like a better alternative, becuase we may have multiple progresses from slcore and they might overwrite one another

src/SLCore/Protocol/EitherJsonConverter.cs
gabriela-trutan-sonarsource: I chose to put it after the foreach, because this is a special, edge case. It's the first DTO encountered that reached this case. So for performance reasons, after the foreach is better

src/SLCore.UnitTests/Protocol/EitherJsonConverterTests.cs
gabriela-trutan-sonarsource: It was not intentional. :O

src/SLCore.Listeners/Implementation/ProgressListener.cs
gabriela-trutan-sonarsource: Because of a previous local implementation

src/SLCore.Listeners/Implementation/ProgressListener.cs
gabriela-trutan-sonarsource: It does. The notifications received from SlCore are quite short

gabriela-trutan-sonarsource: It's also similar to the analysis [notifications](https://github.com/SonarSource/sonarlint-visualstudio/blob/eb9a58a375d3ddc2595768ad8c98dc8f20351464/src/Integration.Vsix/Analysis/AnalysisStrings.resx#L151)

src/SLCore.Listeners/Implementation/ProgressListener.cs
gabriela-trutan-sonarsource: I think it's a nice feature, but unfortunately it does not apply here. It shows a status while you run a task on the background, so it can be cancelled etc. 
In our case, we just receive the notification from SlCore. Of course we can create a hack to run a task and wait until we receive the notification that the task with the same id ended, but I don't think it's the intended use for this case. Also what happens if, for whatever reason, SlCore never notifies us about a task being ended? I find it too risky

georgii-borovinskikh-sonarsource (approved):

PR #6301, opened 2025-06-25
SLVS-2287 Improve DocumentEventsHandler to not do unecessary operations for CMake projects

gabriela-trutan-sonarsource:
[SLVS-2287](https://sonarsource.atlassian.net/browse/SLVS-2287)

[SLVS-2287]: https://sonarsource.atlassian.net/browse/SLVS-2287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Core/CFamily/IActiveCompilationDatabaseTracker.cs
+++ b/src/Core/CFamily/IActiveCompilationDatabaseTracker.cs
@@ -25,5 +25,11 @@ namespace SonarLint.VisualStudio.Core.CFamily;
 public interface IActiveCompilationDatabaseTracker : IRequireInitialization, IDisposable
 {
     string DatabasePath { get; }
-    bool IsCmake { get; }
+    CompilationDatabaseType DatabaseType { get; }
+}
+
+public enum CompilationDatabaseType
+{
+    CMake,
+    VCX
 }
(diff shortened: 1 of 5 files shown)


src/Core/CFamily/IActiveCompilationDatabaseTracker.cs
```diff
@@ -25,4 +25,5 @@ namespace SonarLint.VisualStudio.Core.CFamily;
 public interface IActiveCompilationDatabaseTracker : IRequireInitialization, IDisposable
 {
     string DatabasePath { get; }
+    bool IsCmake { get; }
```
georgii-borovinskikh-sonarsource: should be IsVcx, because the fact that Vcx = CFamily - CMake may change if we support another configuration. Or, even better, have an Enum to make it clear

georgii-borovinskikh-sonarsource (changes requested):

src/Core/CFamily/IActiveCompilationDatabaseTracker.cs
gabriela-trutan-sonarsource: I like the idea of the enum better.

georgii-borovinskikh-sonarsource (approved):

PR #6308, opened 2025-06-27
SLVS-2334 Add integration tests for multiple files analysis

gabriela-trutan-sonarsource:
[SLVS-2334](https://sonarsource.atlassian.net/browse/SLVS-2334)

[SLVS-2334]: https://sonarsource.atlassian.net/browse/SLVS-2334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
+++ b/src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
@@ -112,7 +112,7 @@ internal sealed class FileAnalysisTestsRunner : IDisposable
         return getFileExclusionsListener;
     }
 
-    public async Task<Dictionary<FileUri, List<RaisedIssueDto>>> RunAutomaticFileAnalysis(
+    public async Task<Dictionary<FileUri, List<RaisedIssueDto>>> RunAnalysisOnOpenFile(
         ITestingFile testingFile,
         string configScope,
         bool sendContent = false,
@@ -131,7 +131,7 @@ internal sealed class FileAnalysisTestsRunner : IDisposable
         }
     }
 
-    public async Task<Dictionary<FileUri, List<RaisedIssueDto>>> RunAutomaticMultipleFileAnalysis(
+    public async Task<Dictionary<FileUri, List<RaisedIssueDto>>> RunAnalysisOnUpdateFiles(
         List<ITestingFile> testingFiles,
         string configScope,
         string compilationDatabasePath = null)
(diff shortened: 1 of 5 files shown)


src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
```diff
@@ -127,14 +131,34 @@ public async Task<Dictionary<FileUri, List<RaisedIssueDto>>> RunAutomaticFileAna
         }
     }
 
+    public async Task<Dictionary<FileUri, List<RaisedIssueDto>>> RunAutomaticMultipleFileAnalysis(
```
georgii-borovinskikh-sonarsource: I think we should remove the `RunAutomaticFileAnalysis` method and make this current one use `params` for files if nothing prevents it

georgii-borovinskikh-sonarsource (approved):

src/SLCore.IntegrationTests/FileAnalysisTestsRunner.cs
gabriela-trutan-sonarsource: The idea was to test both `DidUpdateFileSystem` and `DidOpenFile`, that's why the two methods are separate

georgii-borovinskikh-sonarsource: 🤔 that's unexpected. Could we somehow make it a flag or a separate method (while the removing the single file method?)

gabriela-trutan-sonarsource: It is already a separate method. Or am I missing your point?
The presence of a flag would imply two separate methods are needed (as the current state is)

georgii-borovinskikh-sonarsource: The methods are called `AnalyzeSingle` & `AnalyzeMultiple`, but should be called `AnalyzeOnUpdate` & `AnalyzeOnOpen`. Both of them should be able receive multiple files

gabriela-trutan-sonarsource: The `DidOpenFile` parameter receives only a file name. I understand it can be called multiple times, but I don't see the point on doing so, when `DidUpdateFileSystem` is more appropriate.

The proposed renaming sound good.

PR #6315, opened 2025-07-04
SLVS-1275   1     Implement `getBaseDir` listener with non default-value

georgii-borovinskikh-sonarsource:


src/Core/GitWorkSpaceService.cs
```diff
@@ -21,65 +21,80 @@
...
-    public interface IGitWorkspaceService
+    private readonly ILogger logger;
+    private readonly ISolutionInfoProvider solutionInfoProvider;
+    private readonly IFileSystem fileSystem;
+
+    private const string GitFolder = ".git";
+
+    private record struct GitWorkspaceCache(string WorkspaceRoot, string GitRoot);
+    private readonly object lockObject = new();
+    private GitWorkspaceCache value;
```
vnaskos-sonar: np: `value` could be renamed to a more declarative name

src/Core/GitWorkSpaceService.cs
```diff
@@ -21,65 +21,80 @@
...
+{
+    /// <summary>
+    /// Returns the file path to the current Git repository root
+    /// Or null if not a git repo
+    /// </summary>
+    string GetRepoRoot();
+}
+
+[Export(typeof(IGitWorkspaceService))]
+internal class GitWorkSpaceService : IGitWorkspaceService
```
vnaskos-sonar: np: for consistency
```suggestion
internal class GitWorkspaceService : IGitWorkspaceService
```

src/SLCore.Listeners.UnitTests/Implementation/Analysis/AnalysisConfigurationProviderListenerTests.cs
```diff
@@ -28,41 +30,71 @@ namespace SonarLint.VisualStudio.SLCore.Listeners.UnitTests.Implementation.Analy
...
     {
-        var testSubject = new AnalysisConfigurationProviderListener();
+        folderWorkspaceService.FindRootDirectory().ReturnsNull();
+        gitWorkspaceService.GetRepoRoot().ReturnsNull();
 
-        var result = testSubject.GetBaseDirAsync(new GetBaseDirParams(configScopeId)).Result;
+        var result = testSubject.GetBaseDirAsync(new GetBaseDirParams("any")).Result;
 
         result.baseDir.Should().BeNull();
     }
```
vnaskos-sonar, line 95: Missing a test for the case when workspace and git roots exist.

src/SLCore.Listeners/Implementation/Analysis/AnalysisConfigurationProviderListener.cs
```diff
@@ -19,18 +19,21 @@
...
 [Export(typeof(ISLCoreListener))]
 [PartCreationPolicy(CreationPolicy.Shared)]
-internal class AnalysisConfigurationProviderListener : IAnalysisConfigurationProviderListener
+[method: ImportingConstructor]
+internal class AnalysisConfigurationProviderListener(IFolderWorkspaceService folderWorkspaceService, IGitWorkspaceService gitWorkspaceService) : IAnalysisConfigurationProviderListener
 {
     public Task<GetBaseDirResponse> GetBaseDirAsync(GetBaseDirParams parameters)
     {
-        return Task.FromResult(new GetBaseDirResponse(null));
+        var rootDirectory = folderWorkspaceService.FindRootDirectory() ?? gitWorkspaceService.GetRepoRoot();
```
vnaskos-sonar: Shouldn't this be the other way around?

If a workspace is open and is tracked by git, but the git root is one level above the workspace, then the git blame actions will fail, because it reported the workspace root. 🤔

vnaskos-sonar: Logging which path (eg. workspace root or git root), would be nice to have for debugging purposes.

vnaskos-sonar (changes requested):

src/SLCore.Listeners/Implementation/Analysis/AnalysisConfigurationProviderListener.cs
georgii-borovinskikh-sonarsource: above or below? if the git root is the same or the parent path of workspace root, then it will work. if the git root is somehow a child of the workspace root then I guess it's not something we should support. I think having this set to the same folder as opened in vs is more predictable

src/Core/GitWorkSpaceService.cs
georgii-borovinskikh-sonarsource: 👨‍💼 👨‍🚀  work space

src/SLCore.Listeners/Implementation/Analysis/AnalysisConfigurationProviderListener.cs
georgii-borovinskikh-sonarsource: This made me think: currently slcore calls this method way too often (partially why I added the cache to git workspace). I will consider adding this baseDir as part of the ListFiles & active ConfigurationScope logic

PR #6321, opened 2025-07-10
SLVS-2350 Fix: Race condition with CFamily & compilation database

georgii-borovinskikh-sonarsource:
[SLVS-2350](https://sonarsource.atlassian.net/browse/SLVS-2350)

[SLVS-2350]: https://sonarsource.atlassian.net/browse/SLVS-2350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Integration.Vsix.UnitTests/Analysis/DocumentEventsHandlerTests.cs
+++ b/src/Integration.Vsix.UnitTests/Analysis/DocumentEventsHandlerTests.cs
@@ -43,7 +43,7 @@ public class DocumentEventsHandlerTests
     private IActiveConfigScopeTracker activeConfigScopeTracker;
     private IThreadHandling threadHandling;
     private IInitializationProcessorFactory initializationProcessorFactory;
-    private readonly IRequireInitialization[] initializationDependencies = [];
+    private IRequireInitialization[] initializationDependencies;
     private IActiveCompilationDatabaseTracker activeCompilationDatabaseTracker;
     private const string CFamilyOldFile = "file:///tmp/SLVS/old.cpp";
     private const string CFamilyNewFile = "file:///tmp/SLVS/new.cpp";
@@ -59,10 +59,11 @@ public class DocumentEventsHandlerTests
     public void TestInitialize()
     {
         documentTracker = Substitute.For<IDocumentTracker>();
-        documentTracker.GetOpenDocuments().Returns(new List<Document>());
+        documentTracker.GetOpenDocuments().Returns([]);
         vcxCompilationDatabaseUpdater = Substitute.For<IVcxCompilationDatabaseUpdater>();
         activeConfigScopeTracker = Substitute.For<IActiveConfigScopeTracker>();
         activeCompilationDatabaseTracker = Substitute.For<IActiveCompilationDatabaseTracker>();
+        initializationDependencies = [activeCompilationDatabaseTracker];
         threadHandling = Substitute.For<IThreadHandling>();
         MockThreadHandling();
         MockCurrentConfigScope(ConfigurationScope);
@@ -101,6 +102,7 @@ public class DocumentEventsHandlerTests
                 Arg.Any<Func<IThreadHandling, Task>>());
             testSubject.InitializationProcessor.InitializeAsync();
             activeConfigScopeTracker.CurrentConfigurationScopeChanged += Arg.Any<EventHandler<ConfigurationScopeChangedEventArgs>>();
+            activeCompilationDatabaseTracker.DatabaseChanged += Arg.Any<EventHandler>();
             documentTracker.DocumentOpened += Arg.Any<EventHandler<DocumentOpenedEventArgs>>();
             documentTracker.DocumentClosed += Arg.Any<EventHandler<DocumentEventArgs>>();
             documentTracker.DocumentSaved += Arg.Any<EventHandler<DocumentSavedEventArgs>>();
@@ -517,6 +519,7 @@ public class DocumentEventsHandlerTests
         subject.Dispose();
 
         activeConfigScopeTracker.Received(1).CurrentConfigurationScopeChanged -= Arg.Any<EventHandler<ConfigurationScopeChangedEventArgs>>();
+        activeCompilationDatabaseTracker.Received(1).DatabaseChanged -= Arg.Any<EventHandler>();
         documentTracker.Received(1).DocumentClosed -= Arg.Any<EventHandler<DocumentEventArgs>>();
         documentTracker.Received(1).DocumentOpened -= Arg.Any<EventHandler<DocumentOpenedEventArgs>>();
         documentTracker.Received(1).DocumentSaved -= Arg.Any<EventHandler<DocumentSavedEventArgs>>();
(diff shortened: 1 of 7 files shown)


src/Integration.Vsix.UnitTests/Analysis/DocumentEventsHandlerTests.cs
```diff
@@ -576,6 +579,50 @@ public void ActiveConfigScopeTracker_CurrentConfigurationScopeChanged_SlCoreServ
...
+
+        activeCompilationDatabaseTracker.DatabaseChanged += Raise.EventWith(EventArgs.Empty);
+
+        vcxCompilationDatabaseUpdater.DidNotReceiveWithAnyArgs().AddFileAsync(default);
+    }
+
+    [DataTestMethod]
+    [DataRow(CompilationDatabaseType.CMake)]
+    [DataRow(null)]
+    public void ActiveCompilationDatabaseTracker_DatabaseChanged_WhenDatabaseNotVcx_DoesNotUpdateCompilationDatabase(CompilationDatabaseType? databaseType)
```
gabriela-trutan-sonarsource, line 616: We should also have a test for the case that the  `CurrentDatabase` is nulll

src/Integration.Vsix.UnitTests/Analysis/DocumentEventsHandlerTests.cs
```diff
@@ -576,6 +579,50 @@ public void ActiveConfigScopeTracker_CurrentConfigurationScopeChanged_SlCoreServ
...
+    {
+        CreateAndInitializeTestSubject();
+        MockCompilationDatabaseType(databaseType);
+        documentTracker.GetOpenDocuments().Returns([CFamilyDocument, NonCFamilyDocument, CFamilyDocument2]);
+        vcxCompilationDatabaseUpdater.ClearReceivedCalls();
+
+        activeCompilationDatabaseTracker.DatabaseChanged += Raise.EventWith(EventArgs.Empty);
+
+        vcxCompilationDatabaseUpdater.DidNotReceiveWithAnyArgs().AddFileAsync(default);
+    }
```
gabriela-trutan-sonarsource, line 626: We don't have any test for the unsubscribing from the `DatabaseChanged` event

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
```diff
@@ -94,15 +106,17 @@ private async Task HandleConfigScopeEventAsync()
...
+                cFamilyAnalysisConfiguration.DidChangePathToCompileCommands(new(currentConfigScopeId, database?.FullPath));
             }
             else
             {
-                DatabasePath = null;
+                database = null;
             }
         }
+
+        DatabaseChanged?.Invoke(this, EventArgs.Empty);
```
gabriela-trutan-sonarsource, line 119: It's very unexpected that the event is raised even when no changes happened (e.g. we call `HandleConfigScopeEventAsync` during initialization and the config scop might still not be initialized). The consumers of the event will perform operations on all these cases. 

I know that there are no side-effects in the current consumer (`DocumentEventsHandler`), but it's not a good practice to raise unnecessary events. I think we can improve this now at a very low effort.

gabriela-trutan-sonarsource (changes requested):
It's very unexpected that the event `DatabaseChanged` is raised even when no changes happened (e.g. we call HandleConfigScopeEventAsync during initialization and the config scop might still not be initialized). The consumers of the event will perform operations on all these cases.

I know that there are no side-effects in the current consumer (DocumentEventsHandler), but it's not a good practice to raise unnecessary events. I think we can improve this now at a very low effort.

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
georgii-borovinskikh-sonarsource: I could fix the null to null case easily, but changes to config scopes are a bit tricky. Considering that we always have the same vcx db path, we can have 2 different solutions with the same name opened one after another and, technically, we could miss the config scope state of null, that tells us the first one was closed, because of concurrency (side note: normally this is not an issue, as we're relying on eventual consistency here). In this case, if the new event is not raised, the files are not going to be analyzed. I think that the optimization `if (!e.DefinitionChanged)` is enough in this case and discarding false positive events could lead to false negatives if not done carefully. Since there's no negative impact currently, I think we should leave the implementation simple

gabriela-trutan-sonarsource: > technically, we could miss the config scope state of null, that tells us the first one was closed, because of concurrency

I don't see why would we miss the state of the config in the null case. Also the `DocumentEventsHandler` tries to add the opened documents to the compilation database for the null case as well, so nothing would be lost.

georgii-borovinskikh-sonarsource: > why would we miss the state of the config in the null case

Race condition. The event payload does not contain the state. The state is fetched later via property. The property can be modified concurrently, but that raises an event, which is why raising on every config scope event (except for `!e.DefinitionChanged`) is the safest option
> Also the DocumentEventsHandler tries to add the opened documents to the compilation database for the null case as well

That's not the point. It's complicated to have event deduplication logic in this class due to concurrency - it's hard to define when one config scope changed to another if you miss intermediate events.

src/Integration.Vsix.UnitTests/Analysis/DocumentEventsHandlerTests.cs
georgii-borovinskikh-sonarsource: I added it to existing tests

src/Integration.Vsix.UnitTests/Analysis/DocumentEventsHandlerTests.cs
georgii-borovinskikh-sonarsource: `[DataRow(null)]` already tests that

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
georgii-borovinskikh-sonarsource: >That's not the point. It's complicated to have event deduplication logic in this class due to concurrency - it's hard to define when one config scope changed to another if you miss intermediate events.

Sorry, if you miss intermediate state. The events can't be missed, but they don't contain the state in the payload

src/Integration.Vsix.UnitTests/Analysis/DocumentEventsHandlerTests.cs
gabriela-trutan-sonarsource: Missed that 🙄

src/Integration.Vsix/CFamily/ActiveCompilationDatabaseTracker.cs
gabriela-trutan-sonarsource: Ok, we can leave the implementation like that. But can we, at least, update the documentation of the event to mention that the event could be raised multiple times for the same path, so that consumers can be aware of that?

georgii-borovinskikh-sonarsource: Sure, makes sense

PR #6324, opened 2025-07-15
SLVS-2385 Fix threading in mef-importing ctor of AnalysisIssueSelectionService

georgii-borovinskikh-sonarsource:
[SLVS-2385](https://sonarsource.atlassian.net/browse/SLVS-2385)

Part of SLVS-2341

[SLVS-2385]: https://sonarsource.atlassian.net/browse/SLVS-2385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/IssueViz.UnitTests/Selection/AnalysisIssueSelectionServiceTests.cs
+++ b/src/IssueViz.UnitTests/Selection/AnalysisIssueSelectionServiceTests.cs
@@ -42,7 +42,6 @@ public class AnalysisIssueSelectionServiceTests
         selectionService = Substitute.For<IIssueSelectionService>();
         uiServiceOperation = Substitute.For<IVsUIServiceOperation>();
 
-        // Setup the UIServiceOperation to execute the action with the monitor selection
         uiServiceOperation.When(x => x.Execute<SVsShellMonitorSelection, IVsMonitorSelection>(Arg.Any<Action<IVsMonitorSelection>>()))
             .Do(x =>
             {
--- a/src/IssueViz.UnitTests/Selection/AnalysisIssueSelectionServiceTests.cs
+++ b/src/IssueViz.UnitTests/Selection/AnalysisIssueSelectionServiceTests.cs
@@ -18,104 +18,95 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using System;
-using FluentAssertions;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
-using Moq;
-using SonarLint.VisualStudio.TestInfrastructure;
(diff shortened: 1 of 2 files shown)


src/IssueViz.UnitTests/Selection/AnalysisIssueSelectionServiceTests.cs
```diff
@@ -19,339 +19,344 @@
...
+    private AnalysisIssueSelectionService testSubject;
+
+    [TestInitialize]
+    public void TestInitialize()
+    {
+        monitorSelection = Substitute.For<IVsMonitorSelection>();
+        selectionService = Substitute.For<IIssueSelectionService>();
+        uiServiceOperation = Substitute.For<IVsUIServiceOperation>();
+
+        // Setup the UIServiceOperation to execute the action with the monitor selection
```
gabriela-trutan-sonarsource: Instead of the comment you could extract into a method with speaking name. 
Also `setup` as a verb seems to be two words.

src/IssueViz/Selection/AnalysisIssueSelectionService.cs
```diff
@@ -18,118 +18,114 @@
...
-        }
+    private void RaiseSelectionChanged(SelectionChangeLevel changeLevel) =>
+        SelectionChanged?.Invoke(this, new SelectionChangedEventArgs(changeLevel, SelectedIssue, SelectedFlow, SelectedLocation));
 
-        private void UpdateUiContext()
+    private void UpdateUiContext() =>
+        uiServiceOperation.Execute<Microsoft.VisualStudio.Shell.Interop.SVsShellMonitorSelection, Microsoft.VisualStudio.Shell.Interop.IVsMonitorSelection>(monitorSelection =>
         {
-            if (monitorSelection.GetCmdUIContextCookie(ref uiContextGuid, out uint cookie) == VSConstants.S_OK)
+            if (monitorSelection.GetCmdUIContextCookie(ref uiContextGuid, out uint cookie) == Microsoft.VisualStudio.VSConstants.S_OK)
```
gabriela-trutan-sonarsource, line 96: We can consider caching the `monitorSelection` as the `Execute` always switches to the UI thread, which has an impact on performance as compared to the previous implementation.

src/IssueViz.UnitTests/Selection/AnalysisIssueSelectionServiceTests.cs
georgii-borovinskikh-sonarsource: it's probably an AI comment 🙈

src/IssueViz/Selection/AnalysisIssueSelectionService.cs
georgii-borovinskikh-sonarsource: I think monitorSelection's methods should be called on UI thread. The previous implementation assumed that the current thread is the ui thread, but now it's explicit. And since it's already on UI thread then Execute will not switch

PR #6325, opened 2025-07-15
SLVS-2370 UI: Add dependency risks to SonarQube panel

gabriela-trutan-sonarsource:


src/IssueViz.Security/DependencyRisks/DependencyRiskImpactSeverity.xaml
```diff
@@ -0,0 +1,49 @@
+﻿<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+>
+    
+    <!-- ***********************************************************************************
+         Auto-generated from the SVG files by SvgToXaml. See Readme.txt for more information.
```
georgii-borovinskikh-sonarsource: the readme is src/Education/UI/Readme.txt, right?

src/IssueViz.Security/DependencyRisks/DependencyRiskImpactSeverityToImageSourceConverter.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            return resourceFinder.TryFindResource(element, $"{severity}{Suffix}");
+        }
+
+        public object[] ConvertBack(
+            object value,
+            Type[] targetTypes,
+            object parameter,
+            CultureInfo culture) =>
+            throw new NotImplementedException();
```
georgii-borovinskikh-sonarsource, line 53: should it be NotImplementedException? I know this shouldn't ever be called, but then InvalidOperationException may fit the purpose better?

src/IssueViz.Security/ReportView/ReportViewModel.cs
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+using SonarLint.VisualStudio.Core.WPF;
+using SonarLint.VisualStudio.IssueVisualization.Security.DependencyRisks;
+
+namespace SonarLint.VisualStudio.IssueVisualization.Security.ReportView;
+
+internal class ReportViewModel : ViewModelBase
+{
+    public GroupDependencyRiskViewModel GroupDependencyRisk { get; } = new();
```
georgii-borovinskikh-sonarsource, line 28: In the future this will be a list, right? We may have multiple groups

src/IssueViz.Security/ReportView/ReportViewControl.xaml
```diff
@@ -1,6 +1,199 @@
...
+                          Data="M 0 0 L 8 4 L 0 8 Z" VerticalAlignment="Center" HorizontalAlignment="Center" />
+                </Grid>
+                <ControlTemplate.Triggers>
+                    <Trigger Property="IsChecked" Value="True">
+                        <Setter TargetName="Arrow" Property="Data" Value="M 0 0 L 8 0 L 4 6 Z" />
+                    </Trigger>
+                </ControlTemplate.Triggers>
+            </ControlTemplate>
+
+            <Style TargetType="TreeViewItem">
```
georgii-borovinskikh-sonarsource, line 50: Did you create this yourself or extracted it from VS templates? Looks good

src/IssueViz.Security/ReportView/ReportViewControl.xaml
```diff
@@ -1,6 +1,199 @@
...
+                <TreeViewItem.Header>
+                    <Grid>
+                        <Grid.ColumnDefinitions>
+                            <ColumnDefinition Width="Auto" />
+                            <ColumnDefinition Width="Auto" />
+                        </Grid.ColumnDefinitions>
+
+                        <TextBlock Grid.Column="0" Text="{Binding Title}" Style="{StaticResource GroupTextBlockStyle}" />
+                        <TextBlock Grid.Column="1" Margin="5,0,0,0"
+                                   Text="{Binding Path=Risks.Count, StringFormat={}({0} issues)}"
```
georgii-borovinskikh-sonarsource: `1 issues`

src/IssueViz.Security/ReportView/ReportViewControl.xaml
```diff
@@ -1,6 +1,199 @@
...
+                                    <RowDefinition Height="Auto" />
+                                </Grid.RowDefinitions>
+                                <Border Name="Bd" Grid.ColumnSpan="2"
+                                        Background="{TemplateBinding Background}"
+                                        BorderThickness="{TemplateBinding BorderThickness}"
+                                        BorderBrush="{TemplateBinding BorderBrush}"
+                                        SnapsToDevicePixels="true"
+                                        Margin="{TemplateBinding Margin}">
+                                    <StackPanel Orientation="Horizontal" VerticalAlignment="Center">
+                                        <ToggleButton x:Name="ExpanderArrow" Width="16" Height="16" Margin="2,0,2,0"
```
georgii-borovinskikh-sonarsource, line 72: There's a small issue with expanding: single clicking the Arrow toggles the state, however, if you miss the Arrow even by 1px, you need to double click to toggle it

src/IssueViz.Security.UnitTests/DependencyRisks/DependencyRiskTypeToLocalizedTypeConverterTest.cs
```diff
@@ -0,0 +1,72 @@
...
+    public void ConvertBack_NotImplementedException()
+    {
+        Action act = () => testSubject.ConvertBack(null, null, null, null);
+
+        act.Should().Throw<NotImplementedException>();
+    }
+
+    public static IEnumerable<object[]> GetDependencyRiskTypesToLocalization() =>
+    [
+        [DependencyRiskType.Vulnerability, DependencyRiskType.Vulnerability.ToString()],
```
georgii-borovinskikh-sonarsource: we probably should have consistent localization for different risk types

georgii-borovinskikh-sonarsource (approved):

src/IssueViz.Security/DependencyRisks/DependencyRiskImpactSeverityToImageSourceConverter.cs
gabriela-trutan-sonarsource: I would leave it as it is for consistency with the other already existing converters

src/IssueViz.Security/ReportView/ReportViewModel.cs
gabriela-trutan-sonarsource: No, I don't think it will be a list. Each group will have its own particularity, so the `TreeViewItem` will have a separate, manual entry in the [xaml](https://github.com/SonarSource/sonarlint-visualstudio/blob/4ffe31417a5515a1a7e31103f55b017b96e686ec/src/IssueViz.Security/ReportView/ReportViewControl.xaml#L131)  (as opposed to being directly bound via `ItemsSource` property).

But if it could be made a list, the change should not be too difficult.

src/IssueViz.Security/ReportView/ReportViewControl.xaml
gabriela-trutan-sonarsource: I used Copilot, but I spent quite some time to tweak it and to make it more consistent with the style from VS's Error List...

src/IssueViz.Security/ReportView/ReportViewModel.cs
georgii-borovinskikh-sonarsource: How are we going to implement grouping by file if we don't have the list?

src/IssueViz.Security/ReportView/ReportViewControl.xaml
gabriela-trutan-sonarsource: This is the default behaviour of a `TreeView`: when clicking on the arrow, it toggles the child nodes by one click; when clicking on the parent node, it toggles the child nodes via double click. 

I struggled to change this behaviour by making sure that the child nodes are toggled when clicking just once on the parent node. I couldn't make it, because it seems to be difficult to add an event just for the `TreeViewItem` that is the parent node. I might give it another try.

src/IssueViz.Security.UnitTests/DependencyRisks/DependencyRiskTypeToLocalizedTypeConverterTest.cs
gabriela-trutan-sonarsource: I don't think it brings a lot of advantages of localizing single words. With this approach, adding new types will just work by default and we would only need to localize the values that have two or more words.

src/IssueViz.Security/ReportView/ReportViewModel.cs
gabriela-trutan-sonarsource: By just having another group `GroupIssueViewModel`. But we might actually need a different `TreeView` for it as it might be difficult to mix the different styles for different parent `TreeViewItem`s

I think it's not worth investing time now in speculating this as we won't know for sure what the requirements will be

src/IssueViz.Security/ReportView/ReportViewControl.xaml
gabriela-trutan-sonarsource: I managed to implement the functionality.

src/IssueViz.Security.UnitTests/DependencyRisks/DependencyRiskTypeToLocalizedTypeConverterTest.cs
georgii-borovinskikh-sonarsource: That's not really the point of having ui strings put in the resources. It will also be confusing to anyone who modifies this code in the future why some values re converted directly and why some are looked up in the resources

src/IssueViz.Security/ReportView/ReportViewControl.xaml
gabriela-trutan-sonarsource: I created a `EnglishPluralizationConverter` as we might use it in the future for other cases as well

src/IssueViz.Security.UnitTests/DependencyRisks/DependencyRiskTypeToLocalizedTypeConverterTest.cs
gabriela-trutan-sonarsource: Then I will probably just modify the converter to make sure it adds white space between words, so it's generic and there is no need to always localize enum values (lot of effort to maintain)

